PR #935, opened 2022-01-08
Add new make targets to build custom release binaries

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

This builds a binary with a different tag and also zips it with the bundle embedded, and can only be used by `eksctl anywhere` plugin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ endif
 RELEASE_MANIFEST_URL?=https://dev-release-prod-pdx.s3.us-west-2.amazonaws.com/eks-a-release.yaml
 
 DEV_GIT_VERSION:=v0.0.0-dev
+CUSTOM_GIT_VERSION:=v0.0.0-custom
 
 AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
 AWS_REGION=us-west-2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -141,6 +142,32 @@ eks-a-cross-platform-embed-latest-config: ## Build cross platform dev release ve
 	$(MAKE) eks-a-embed-config GO_OS=linux GO_ARCH=arm64 OUTPUT_FILE=bin/linux/arm64/eksctl-anywhere
 	rm pkg/cluster/config/bundle-release.yaml
 
+.PHONY: eks-a-custom-embed-config
+eks-a-custom-embed-config:
+	$(MAKE) eks-a-binary GIT_VERSION=$(CUSTOM_GIT_VERSION) RELEASE_MANIFEST_URL=embed:///config/releases.yaml LINKER_FLAGS='-s -w -X github.com/aws/eks-anywhere/pkg/eksctl.enabled=true' BUILD_TAGS='$(BUILD_TAGS) spec_embed_config'
+
+.PHONY: eks-a-cross-platform-custom-embed-latest-config
+eks-a-cross-platform-custom-embed-latest-config: ## Build custom binary with latest dev release bundle that embeds config and builds it as a release binary for linux/amd64
```
mrajashree, line 150: is the only difference between this target and `eks-a-cross-platform-embed-latest-config` the use of the custom git version for embed config? in that case we can just modify the existing `eks-a-cross-platform-embed-latest-config` target to use the custom git version. Also just curious why do we need to add `v0.0.0-custom` and can't use `v0.0.0-dev`?

vivek-koppuru: The other main difference is that this can only be called by eksctl due to the eksctl enabled flag being built in. I want to keep the tags separate for that reason as this was how I shared the binary to external members, so that we don't accidentally share a dev version that can be invoked without the eksctl plugin. Other than that it's true that the embedding is the same. This is just more like the release target with a proper git tag than our regular dev binaries we use to test.

abhay-krishna (approved):
/lgtm

PR #948, opened 2022-01-11
Readd cilium helm chart to bundle

g-gaston:
*Description of changes:*
Readds the chart to the bundle after the build tooling repo has been updated with support for it

Also make a distcition between name and image for the image artifacts since the chart and the cilium image have the same image name (just a different tag). When using the image as the map key, we got some overlapping.


abhay-krishna (approved):
/lgtm

PR #953, opened 2022-01-12
Remove logic for patching cloud-controller-manager with tolerations

mrajashree:
With CAPV 1.0.1, cloud-controller-manager can be specified in the spec
and is not deployed from within capv. So we can add tolerations for CP taints
on this daemonset within the spec. This change is already merged, and logic
for patching the daemonset with tolerations post upgrade has been removed.
This commit removes the post controlplane creation action of applying these
tolerations to the daemonset since that's also not needed anymore.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -38,26 +38,25 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


vivek-koppuru (approved):
/lgtm

PR #962, opened 2022-01-13
Remove analyzers for `capi-webhook-system` namespace, as it's been re…

danbudris:
…moved in CAPI v1beta1

Additionally, there is a bug in upstream Troubleshoot which casues a panic if the namespace for a deplyoment it's analyzing does not exist. This issue was revealed by the removal of the `capi-webhook-system` namespace by the migration to v1beta1 CAPI.
Working on an upstream fix to Troubleshoot support-bundle.

*Issue #, if available:*

*Description of changes:*


g-gaston (approved):
/lgtm

PR #969, opened 2022-01-13
Move `upgrade plan` functionality to `upgrade plan cluster`

### LINKED ISSUE #822: New cobra command for `plan`
**What would you like to be added**:
Create new cobra command for `plan` instead of using the `upgrade plan` as a single cobra command.

g-gaston:
Resolves #822 

*Description of changes:*
This is more in sync with our 'verb + resource' pattern for commands
plus it opens the possibility to write upgrade plan commands for other
types of resources.


danbudris (approved):
/lgtm

PR #971, opened 2022-01-14
Provide datastore for deploying the os template

abhinavmpandey08:
*Description of changes:*
Currently, during auto-import we don't provide a datastore to deploy the template to. This results in the template being deployed in a random datastore. This PR specifies the datastore for the template to be deployed in.


g-gaston (approved):
Beautiful work
/lgtm

PR #975, opened 2022-01-14
Revert "Create v1beta1 manifests for vsphere provider (#930)"

vivek-koppuru:
This reverts commit 1f90618c94d61efd8cdd985dffd2c9cd57146040.

*Issue #, if available:*

*Description of changes:*

This is causing issues with upgrades/scaling because the vsphere username and password is not accessible by the controller, so it is filling up the values as empty strings, causing the capv integration to break. Will need to move the secrets back to a separate step or retrieve those values through the controller, but because that would require some rework on these changes, I will revert this for now.


jaxesn (approved):
/lgtm

abhay-krishna (approved):
/lgtm

PR #982, opened 2022-01-18
More options to eks-a test tool fetch logs command

g-gaston:
*Description of changes:*
Added a flag to filter the tests by name

Added a flag to only print tests logs to stdout instead of writing to disk. This prints different logs with different colors depending on the origin and info: e2e test logs, go test logs, V5-9 CLi logs, error/warning logs, etc.

Added bin to gitignore

This introduces two new concepts:
* Test filter: allows to filter which test logs are retrieved. For now we have two: failed tests (default behavior) and filtr by test name
* Consumers: these take some kind of test log artifact and process them. For now we have writers to disk or a printer to stdout. there are different processors for the different "log artifacts": codebuild, build messages and test logs

These are all configurable through variadic options


danbudris (approved):
/lgtm

PR #984, opened 2022-01-19
Update controller unit tests

taneyland:
*Issue #, if available:*

*Description of changes:*
Controller unit tests were using two different specs as the source of truth for the machine configs. This lead to names of objects being mismatched throughout the reconciler workflow. 

These changes set the specs from the testdata file as the only source of truth for the reconciler unit tests.


vivek-koppuru (approved):
/lgtm

PR #990, opened 2022-01-19
Import v1beta1 apiversion for capi, capv and external etcd

mrajashree:
*Issue #, if available:*

*Description of changes:*


go.mod
```diff
@@ -9,31 +9,31 @@ require (
...
+	sigs.k8s.io/cluster-api v1.0.2
+	sigs.k8s.io/cluster-api-provider-vsphere v1.0.1
 	sigs.k8s.io/controller-runtime v0.10.3
 	sigs.k8s.io/yaml v1.3.0
 )
 
-// exclude un-required transitive dependency from cluster-api-provider-vsphere v0.7.8
-exclude sigs.k8s.io/cluster-api v0.3.14
+// exclude un-required transitive dependency from cluster-api-provider-vsphere v1.0.1
+exclude sigs.k8s.io/cluster-api v0.0.0-00010101000000-000000000000
```
vivek-koppuru, line 36: Why is this not `v1.0.0` as it shows in upstream?

controllers/main.go
```diff
@@ -5,15 +5,22 @@ import (
...
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	vspherev1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
-	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1alpha3"
+	vspherev1alpha3 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3"
+	vspherev1alpha4 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4"
```
vivek-koppuru, line 16: Why do we need the `v1alpha4` ones?

mrajashree: I thought we might need them for conversion webhooks, but neither v1alpha3 nor v1alpha4 are needed, so removing them

go.mod
mrajashree: otherwise `go mod tidy` throws this error
```
go: sigs.k8s.io/cluster-api-provider-vsphere@v1.0.1 requires
	sigs.k8s.io/cluster-api/test@v1.0.0 requires
	sigs.k8s.io/cluster-api@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

vivek-koppuru: So I tried to checkout this change and remove this line and add a replace similar to underneath like how it is done here, and then `go mod tidy` succeeds. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/v1.0.1/go.mod#L37

I am not sure what is different, but plan on taking a look more into it tomorrow. Wondering if that is better than excluding a dependency? But may not be that different tbh, just looking at what upstream does.

mrajashree: I too tried this out locally with capi tag v1.0.2 `sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.0.2` under replace and go mod tidy worked

vivek-koppuru (approved):
/lgtm

PR #1009, opened 2022-01-21
Removing clusterID from AWSIamConfig spec and generating uuid

pokearu:
*Description of changes:*
Removing clusterID field from AWSIamConfig CRD. By default the clusterID is set to a UUID during cluster creation and does not support upgrade/gitops to change the value.


danbudris:
So, my understanding from the convo the other day is that we'd allow the customer to override the default value if they explicitly wanted to

danbudris:
> So, my understanding from the convo the other day is that we'd allow the users to override the default `clusterId` value if they explicitly wanted to

spoke with @pokearu and we are moving forward with the removal of the clusterID configuration; we can always open up the option to configure it later based on user feedback.

pokearu:
> > So, my understanding from the convo the other day is that we'd allow the users to override the default `clusterId` value if they explicitly wanted to
> 
> spoke with @pokearu and we are moving forward with the removal of the clusterID configuration; we can always open up the option to configure it later based on user feedback.

Yup! Currently we are enforcing a default uuid 👍  and we can address this again based on feedback.

danbudris (approved):

danbudris:
sorting out integration tests prior to merge

PR #1010, opened 2022-01-21
Add additional worker node groups during upgrade

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR supports adding additional worker node groups during upgrade. 

It also adds support to specify node labels for different worker node groups

Removed the `machineConfigUniqueness` check for worker nodes because we allow multiple groups to point to the same machine config if desired


pkg/clustermanager/cluster_manager.go
```diff
@@ -890,8 +892,10 @@ func (c *ClusterManager) PauseEKSAControllerReconcile(ctx context.Context, clust
 		if clusterSpec.Spec.ControlPlaneConfiguration.MachineGroupRef == nil {
 			return fmt.Errorf("machineGroupRef for control plane is not defined")
 		}
-		if len(clusterSpec.Spec.WorkerNodeGroupConfigurations) <= 0 || clusterSpec.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef == nil {
-			return fmt.Errorf("machineGroupRef for worker nodes is not defined")
+		for _, workerNodeGroupConfiguration := range clusterSpec.Spec.WorkerNodeGroupConfigurations {
+			if workerNodeGroupConfiguration.MachineGroupRef == nil {
```
g-gaston, line 896: nit: this almost seems like a validation, if we don't have it at the api validations level, we could add it. And remove it from here

pkg/clustermanager/cluster_manager.go
```diff
@@ -934,41 +938,23 @@ func (c *ClusterManager) ResumeEKSAControllerReconcile(ctx context.Context, clus
 		if clusterSpec.Spec.ControlPlaneConfiguration.MachineGroupRef == nil {
 			return fmt.Errorf("machineGroupRef for control plane is not defined")
 		}
-		if len(clusterSpec.Spec.WorkerNodeGroupConfigurations) <= 0 || clusterSpec.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef == nil {
-			return fmt.Errorf("machineGroupRef for worker nodes is not defined")
+		for _, workerNodeGroupConfiguration := range clusterSpec.Spec.WorkerNodeGroupConfigurations {
```
g-gaston, line 941: Is this the exact same code as in line 895? In that case, I vote to remove it

pkg/providers/docker/docker.go
```diff
@@ -460,3 +467,11 @@ func (p *provider) UpgradeNeeded(_ context.Context, _, _ *cluster.Spec) (bool, e
 func (p *provider) RunPostControlPlaneCreation(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error {
 	return nil
 }
+
+func prevWorkerNodeGroups(prevWorkerNodeGroups []v1alpha1.WorkerNodeGroupConfiguration) map[string]v1alpha1.WorkerNodeGroupConfiguration {
```
g-gaston, line 471: nit: `buildMapForWorkerNodeGroupsByName` (or something else that tell us what the method is doing

pkg/providers/docker/docker.go
```diff
@@ -270,17 +270,24 @@ func (p *provider) generateCAPISpecForUpgrade(ctx context.Context, bootstrapClus
...
 	for _, workerNodeGroupConfiguration := range newClusterSpec.Spec.WorkerNodeGroupConfigurations {
-		needsNewWorkloadTemplate := NeedsNewWorkloadTemplate(currentSpec, newClusterSpec)
-		if !needsNewWorkloadTemplate {
-			machineDeploymentName := fmt.Sprintf("%s-%s", newClusterSpec.Name, workerNodeGroupConfiguration.Name)
-			md, err := p.providerKubectlClient.GetMachineDeployment(ctx, workloadCluster, machineDeploymentName, executables.WithCluster(bootstrapCluster), executables.WithNamespace(constants.EksaSystemNamespace))
-			if err != nil {
-				return nil, nil, err
+		if _, ok := previousWorkerNodeGroupConfigs[workerNodeGroupConfiguration.Name]; ok {
+			needsNewWorkloadTemplate := NeedsNewWorkloadTemplate(currentSpec, newClusterSpec)
+			if !needsNewWorkloadTemplate {
```
g-gaston, line 279: nit: you could avoid nested ifs with
```if _, ok := previousWorkerNodeGroupConfigs[workerNodeGroupConfiguration.Name]; ok && !NeedsNewWorkloadTemplate(currentSpec, newClusterSpec) {```

File: pkg/workflows/upgrade.go [hunk 5]
```diff
@@ -393,7 +393,7 @@ func (s *updateClusterAndGitResources) Run(ctx context.Context, commandContext *
 	}
 
 	logger.Info("Resuming EKS-A controller reconciliation")
-	err = commandContext.ClusterManager.ResumeEKSAControllerReconcile(ctx, target, commandContext.ClusterSpec, commandContext.Provider)
+	err = commandContext.ClusterManager.ResumeEKSAControllerReconcile(ctx, target, commandContext.CurrentClusterSpec, commandContext.Provider)
```
g-gaston, line 396: doesn't this happen after the upgrade? and in that case, shouldn't it be `ClusterSpec`?

pkg/workflows/upgrade.go
```diff
@@ -259,7 +259,7 @@ func (s *pauseEksaAndFluxReconcile) Run(ctx context.Context, commandContext *tas
 	target := getManagementCluster(commandContext)
 
 	logger.Info("Pausing EKS-A cluster controller reconcile")
-	err := commandContext.ClusterManager.PauseEKSAControllerReconcile(ctx, target, commandContext.ClusterSpec, commandContext.Provider)
+	err := commandContext.ClusterManager.PauseEKSAControllerReconcile(ctx, target, commandContext.CurrentClusterSpec, commandContext.Provider)
```
g-gaston, line 262: good catch!

pkg/providers/vsphere/testdata/cluster_main_multiple_worker_node_groups.yaml
```diff
@@ -23,11 +23,6 @@ spec:
         name: test-wn
         kind: VSphereMachineConfig
       name: md-1
-  externalEtcdConfiguration:
-    count: 3
-    machineGroupRef:
-      name: test-etcd
-      kind: VSphereMachineConfig
```
g-gaston, line 30: why are we changing this? :)

pkg/providers/vsphere/vsphere.go
```diff
@@ -885,23 +874,30 @@ func (p *vsphereProvider) generateCAPISpecForUpgrade(ctx context.Context, bootst
...
+		if _, ok := previousWorkerNodeGroupConfigs[workerNodeGroupConfiguration.Name]; ok {
+			workerVmc, err := p.providerKubectlClient.GetEksaVSphereMachineConfig(ctx, workerNodeGroupConfiguration.MachineGroupRef.Name, workloadCluster.KubeconfigFile, newClusterSpec.Namespace)
 			if err != nil {
 				return nil, nil, err
 			}
-			workloadTemplateName = md.Spec.Template.Spec.InfrastructureRef.Name
-			workloadTemplateNames[workerNodeGroupConfiguration.Name] = workloadTemplateName
+
+			needsNewWorkloadTemplate := NeedsNewWorkloadTemplate(currentSpec, newClusterSpec, vdc, p.datacenterConfig, workerVmc, workerMachineConfig)
+			if !needsNewWorkloadTemplate {
```
g-gaston, line 889: same here with nested ifs
in this case it might be a bit too lengthy so you could abstract the condition in a method

pkg/providers/vsphere/testdata/cluster_main_multiple_worker_node_groups.yaml
taneyland: I'm using this file for multiple tests & it required more configurations for the tests outside of the scope of the feature that we're actually testing so I removed it

g-gaston (approved):

PR #1018, opened 2022-01-21
Don't specify kubeadm control plane api version

g-gaston:
*Description of changes:*
When running validations over old clusters that haven't yet received the
CAPI upgrade, only v1alpha3 objects exit. By not specifying the version,
we retrieve whatever object version is available.

This has a downside: the old version of the object can be serialized in
the new api go struct. In that case, there could fields that are not
unmarshalled correctly, depending on the control plane spec changes. IN
this case, this doesn't affect any of our code.

This should also fix the problems with `LatestMinorRelease` e2e tests. I tested it locally with `TestDockerKubernetes121UpgradeFromLatestMinorRelease`


danbudris (approved):

mrajashree (approved):
/lgtm

PR #1021, opened 2022-01-23
Specify worker node group name in default cluster config

taneyland:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generateclusterconfig.go
+++ b/cmd/eksctl-anywhere/cmd/generateclusterconfig.go
@@ -66,6 +66,7 @@ func generateClusterConfig(clusterName string) error {
 			v1alpha1.ControlPlaneConfigCount(1),
 			v1alpha1.ExternalETCDConfigCount(1),
 			v1alpha1.WorkerNodeConfigCount(1),
+			v1alpha1.WorkerNodeConfigName("md-0"),
 		)
 		dcyaml, err := yaml.Marshal(datacenterConfig)
 		if err != nil {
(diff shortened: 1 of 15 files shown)


cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -66,6 +66,7 @@ func generateClusterConfig(clusterName string) error {
 			v1alpha1.ControlPlaneConfigCount(1),
 			v1alpha1.ExternalETCDConfigCount(1),
 			v1alpha1.WorkerNodeConfigCount(1),
+			v1alpha1.WorkerNodeConfigName("md-0"),
```
danbudris, line 69: we should put this in a constant

danbudris (changes requested):

danbudris (approved):

pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -43,11 +42,3 @@ func setRegistryMirrorConfigDefaults(clusterConfig *Cluster) error {
 	}
 	return nil
 }
-
-func setWorkerNodeGroupDefaults(cluster *Cluster) error {
```
mrajashree, line 47: let's keep this in, because this is the part that actually sets the default, in case the config is not created using generate cluster config command. Since we're removing the same logic from [vsphere validator](https://github.com/aws/eks-anywhere/pull/1021/files#diff-238036d622b0e49427d5c5115ba7fa29145e3741beb69f67aa345a6094280deeL91) it won't print the same msg twice

mrajashree (approved):
looks good!

PR #1022, opened 2022-01-24
Remove DNS config from default cluster config

taneyland:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -244,7 +244,7 @@ func (n *ClusterNetwork) Equal(o *ClusterNetwork) bool {
 	}
 	return SliceEqual(n.Pods.CidrBlocks, o.Pods.CidrBlocks) &&
 		SliceEqual(n.Services.CidrBlocks, o.Services.CidrBlocks) &&
-		n.CNI == o.CNI && n.DNS == o.DNS
+		n.CNI == o.CNI && n.DNS.ResolvConf.Equal(o.DNS.ResolvConf)
 }
 
 func SliceEqual(a, b []string) bool {
@@ -305,6 +305,16 @@ type DNS struct {
 type ResolvConf struct {
 	// Path defines the path to the file that contains the DNS resolver configuration
 	Path string `json:"path,omitempty"`
+}
+
+func (n *ResolvConf) Equal(o *ResolvConf) bool {
+	if n == o {
+		return true
+	}
+	if n == nil || o == nil {
+		return false
+	}
+	return n.Path == o.Path
 }
 
 type KubernetesVersion string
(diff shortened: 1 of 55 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -231,8 +231,8 @@ type ClusterNetwork struct {
 	Pods     Pods     `json:"pods,omitempty"`
 	Services Services `json:"services,omitempty"`
 	// CNI specifies the CNI plugin to be installed in the cluster
-	CNI CNI `json:"cni,omitempty"`
-	DNS DNS `json:"dns,omitempty"`
+	CNI CNI  `json:"cni,omitempty"`
+	DNS *DNS `json:"dns,omitempty"`
```
g-gaston, line 235: What do you think about making `DNS` a struct but `ResolvConf` a pointer?
I think that express better the design of the api:
* DNS is just a grouping of configuration, it doesn't need to be a pointer. This follows the same pattern as other fields like `Networking`
* `ResolvConf` is an optional configuration for an specific sense. It makes sense that it's a pointer since that's the way the user can express they don't want to configure it.

If you decide to keep `DNS` as a pointer, then I would make `ResolvConf` a pointer as well

pkg/clustermarshaller/clustermarshaller.go
```diff
@@ -40,6 +41,13 @@ func MarshalClusterSpec(clusterSpec *cluster.Spec, datacenterConfig providers.Da
 		if err != nil {
 			return nil, fmt.Errorf("failed marshalling resource for cluster spec: %v", err)
 		}
+		if clusterSpec.Spec.ClusterNetwork.DNS.ResolvConf == nil {
```
g-gaston, line 44: I'm no opposed to this but I'm not sure we need it

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -299,14 +299,24 @@ type Services struct {
...
 
+func (n *ResolvConf) Equal(o *ResolvConf) bool {
+	if n == o {
+		return true
+	}
+	if n == nil || o == nil {
+		return false
+	}
+	return n.Path == o.Path
+}
```
g-gaston, line 318: Do we have tests for all these cases?

g-gaston (approved):

pkg/clustermarshaller/clustermarshaller.go
taneyland: Without this, it always removes the dns path, even when we specify a resolvConf file.

pkg/api/v1alpha1/cluster_types.go
taneyland: Tests added

PR #1025, opened 2022-01-24
Delete worker node group name in old CLI E2E tests

g-gaston:
*Description of changes:*
The CLI does not allow extra fields in the cluster configuration file.
Old versions of the CLI don't support names for worker node groups,
since that a new field. This means we need to remove that field in the
E2E tests before passing the cluster config file to the old CLI.

Moved helpers to delete paths from a yaml object to the internal api
package and added tests.

### DIFF
--- a/internal/pkg/api/yaml.go
+++ b/internal/pkg/api/yaml.go
@@ -1,0 +1,67 @@
+package api
+
+import (
+	"strings"
+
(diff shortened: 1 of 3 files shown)


danbudris (approved):

mrajashree (approved):
/lgtm

PR #1027, opened 2022-01-25
Fix URLs for CAPI manifests

abhay-krishna:
This check was causing the incorrect URL to be assigned to CAPI metadata manifests.


vivek-koppuru (approved):
/lgtm

PR #1028, opened 2022-01-25
Fix linting errors

ptrivedi:
*Issue #, if available:*

*Description of changes:*


pkg/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
g-gaston, line 1: Can we remove this change?
It's an autogenerated file so next time we generate it will show up as a new change

g-gaston (approved):
/lgtm

PR #1032, opened 2022-01-25
Modifying eks-a controller error message

pokearu:
*Description of changes:*
Nit fix.
Modifying error message to reflect the CRD kind that fetcher is trying to do an eksa cluster lookup.

### DIFF
--- a/controllers/controllers/resource/fetcher.go
+++ b/controllers/controllers/resource/fetcher.go
@@ -162,7 +162,7 @@ func (r *CapiResourceFetcher) fetchClusterForRef(ctx context.Context, refId type
 			}
 		}
 	}
-	return nil, fmt.Errorf("eksa cluster not found for datacenterRef %v", refId)
+	return nil, fmt.Errorf("eksa cluster not found for %s: %v", kind, refId)
 }
 
 func (r *CapiResourceFetcher) machineDeploymentsMap(ctx context.Context, c *anywherev1.Cluster) (map[string]*clusterv1.MachineDeployment, error) {


controllers/controllers/resource/fetcher.go
```diff
@@ -162,7 +162,7 @@ func (r *CapiResourceFetcher) fetchClusterForRef(ctx context.Context, refId type
 			}
 		}
 	}
-	return nil, fmt.Errorf("eksa cluster not found for datacenterRef %v", refId)
+	return nil, fmt.Errorf("eksa cluster not found for %s: %v", kind, refId)
```
g-gaston, line 165: I think `%s` would be better for `refId` as well

g-gaston (approved):

controllers/controllers/resource/fetcher.go
pokearu: Actually `refId` is a struct of type `types.NamespacedName`
That's why i left it as `%v`. It gets printed as `<namespace>/<name>`

g-gaston: Interesting, in this case it behaves the same way, I didn't know this
For the record:
When a struct implements the `Stringer` interface (method `String() string`), Go will use it for both `%v` and `%s` (I tough it was only for `%s`)
Which is what is happening [in this case](https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L37)

g-gaston:
Leaving the :lgtm for @danbudris

danbudris:
> Leaving the :lgtm for @danbudris

I abdicate my responsibility now that the release is done 👑 ⬇️

PR #1035, opened 2022-01-25
Add Tinkerbell assets only to dev release manifest

abhay-krishna:
Tinkerbell is still in development phase behind feature flag, so we don't want these images to feature in the production bundle.


danbudris (approved):
/lgtm

PR #1036, opened 2022-01-25
Add feature gate to node labels

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Adding a feature flag to node labels to address issue with upgrading node labels in a following release. It has been added the same way as the taints feature flag.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1038, opened 2022-01-26
Refactor preflight validations and change bool name for bypassing LogPass call

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Refactoring preflight validations so we have common code instead of repeating validation code in both create and upgrade. Also changed bool name to `Silent` so that it can be more flexible than just for feature flags and explains what is actually happening. 

This PR addresses comments from https://github.com/aws/eks-anywhere/pull/1036

### DIFF
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -1,0 +1,31 @@
+package validations
+
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


g-gaston (approved):
👏🏻

PR #1066, opened 2022-01-27
Pass branch name from codebuild to e2e tests

g-gaston:
*Description of changes:*
`BRANCH_NAME` env var is already available in codebuild runs and the e2e tests understand `T_BRANCH_NAME` and are able to use to determine the prior available release.

This just closes the bridge, passing the value to the e2e test runner binary and from them to the e2e tests through the ssm command. It defaults to `main` when not set.


g-gaston:
Testing the cherrypick command

/cherrypick release-0.7

PR #1074, opened 2022-01-28
Adding replace for patched CAPT in go.mod

pokearu:
*Description of changes:*
We are maintaining a patched version of [CAPT](https://github.com/pokearu/cluster-api-provider-tinkerbell/tree/release-v0.1.0) which has modified APIs. 
Added a replace and ran `go mod tidy`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,4 +42,5 @@ replace (
 	github.com/containerd/containerd => github.com/containerd/containerd v1.5.9
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.0.0
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc95
+	github.com/tinkerbell/cluster-api-provider-tinkerbell => github.com/pokearu/cluster-api-provider-tinkerbell v0.0.0-20220128001529-79d851d0861f
 )


jaxesn:
Have we had to do this for capi or the other deps like we have patches for? Just checking if this is the right way to go Cc @g-gaston @vignesh-goutham @mrajashree

abhinavmpandey08 (approved):

g-gaston:
> Have we had to do this for capi or the other deps like we have patches for? Just checking if this is the right way to go Cc @g-gaston @vignesh-goutham @mrajashree

If this is something temporary and we expect the struct changes to be merged upstream eventually, this is ok. It allows us to not have to change the code (the imports in particular) once we are ready to start using the upstream module. If we intend to always use our own fork, then there is no need to mask it and we should import it directly.

For the record: I don't think we did it for capi because we didn't need in the CLI any of the changes we made. We only use the the api structs from that module and in particular we only care about standard fields and none of the custom ones.

However, I'm curious about all those indirect dependencies. Why do we need them?

Also, would you mind adding a comment next to your `replace`, indicating why we need it? I would like for us to start documenting all the replaces. They are supposed to be temporary and we should start to try to get rid of them from time to time. This becomes easier if we know why there were there in the first place.

pokearu:
@g-gaston 

We do hope this is temporary. Eventually we plan to change the way PBNJ interacts with Tink and that point these new API types in CAPT might not be required.

About the indirect dependencies, they actually popped up when i ran `go mod tidy`?

And yes you are right, I just did an initial PR to add the `replace` and @abhinavmpandey08 is going to do a follow up PR that consumes the new structs in the eks-a-tool.

pokearu:
> About the indirect dependencies, they actually popped up when i ran `go mod tidy`?

My bad. I think those popped up when I was testing adding `github.com/tinkerbell/cluster-api-provider-tinkerbell` as an import and ran a `go mod tidy`. I can take them out.

pokearu:
Closing in favor of #1103

PR #1076, opened 2022-01-28
Adding cloudstack cluster config and CRD's

maxdrib:
*Issue #, if available:*

*Description of changes:*
First PR of many to add support for Cloudstack CAPI provider (https://github.com/aws/cluster-api-provider-cloudstack). This PR initially defines the new EKS-A objects and adds some tests for them, and we are able to generate the cloudstack cluster config behind a feature gate. Future PR's will include the preflight checks with CloudMonkey, provider code for creation/upgrade/deletion, and the controller code. Much of this code was originally developed by @wanyufe 


config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdeploymentconfigs.yaml
```diff
@@ -0,0 +1,78 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.1
+  creationTimestamp: null
+  name: cloudstackdeploymentconfigs.anywhere.eks.amazonaws.com
+spec:
+  group: anywhere.eks.amazonaws.com
+  names:
+    kind: CloudStackDeploymentConfig
```
vivek-koppuru, line 13: What is a `CloudStackDeploymentConfig`? Can we make this `CloudStackDatacenterConfig` to make it consistent with the other providers?

vivek-koppuru: Also maybe need to go through a design doc of what fields we are adding in terms of custom resources

pkg/api/v1alpha1/cloudstackdeploymentconfig_types.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (v *CloudStackDeploymentConfig) ClearPauseAnnotation() {
+	if v.Annotations != nil {
+		delete(v.Annotations, pausedAnnotation)
+	}
+}
+
+func (v *CloudStackDeploymentConfig) ConvertConfigToConfigGenerateStruct() *CloudStackDeploymentConfigGenerate {
+	config := &CloudStackDeploymentConfigGenerate{
```
vivek-koppuru, line 85: We might be missing this condition here? https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheredatacenterconfig_types.go#L107-L110

pkg/api/v1alpha1/cloudstackdeploymentconfig_webhook.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(GroupVersion.WithKind(CloudStackDeploymentKind).GroupKind(), r.Name, allErrs)
+}
+
+func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDeploymentConfig) field.ErrorList {
+	if old.IsReconcilePaused() {
```
vivek-koppuru, line 75: We can add this in `ValidateUpdate` itself right? Like here? https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheredatacenterconfig_webhook.go#L68-L71

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// CloudStackMachineConfigSpec defines the desired state of CloudStackMachineConfig
+type CloudStackMachineConfigSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	Template         string              `json:"template"`
```
vivek-koppuru, line 28: It would be easier to understand these fields if we revisit any design doc regarding these

pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
```diff
@@ -0,0 +1,71 @@
...
+	allErrs = append(allErrs, validateImmutableFieldsCloudStackMachineConfig(r, oldCloudStackMachineConfig)...)
+
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(GroupVersion.WithKind(CloudStackMachineConfigKind).GroupKind(), r.Name, allErrs)
+}
+
+func validateImmutableFieldsCloudStackMachineConfig(new, old *CloudStackMachineConfig) field.ErrorList {
```
vivek-koppuru, line 57: We don't really need this function unless it is modifying control plane or etcd, as shown here: https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheremachineconfig_webhook.go#L92

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -135,7 +135,8 @@ type ControlPlaneConfiguration struct {
 	// Count defines the number of desired control plane nodes. Defaults to 1.
 	Count int `json:"count,omitempty"`
 	// Endpoint defines the host ip and port to use for the control plane.
-	Endpoint *Endpoint `json:"endpoint,omitempty"`
+	Endpoint       *Endpoint `json:"endpoint,omitempty"`
+	ExtraEndpoints []string  `json:"extraEndpoints,omitempty"`
```
vivek-koppuru, line 139: Can you explain this addition as we would need to support this for other providers then, unless we will have specific validation for this based on this provider?

pkg/api/v1alpha1/machine_types.go
```diff
@@ -0,0 +1,15 @@
+package v1alpha1
+
+type OSFamily string
+
+const (
+	Ubuntu       OSFamily = "ubuntu"
+	Bottlerocket OSFamily = "bottlerocket"
+	Redhat       OSFamily = "redhat"
```
vivek-koppuru, line 8: If we refactor this, does this mean we support `Redhat` for VSphere too potentially?

vivek-koppuru (commented):
Nice start!

Can we put this all underneath a feature flag until we finish this effort, similar to how we do it for tinkerbell? Unless nothing is using this code as of now, which is what I see for the most part unless it would be enabled by a different PR.

pkg/api/v1alpha1/machine_types.go
vignesh-goutham: We are exploring an idea of supporting a yum based distro. Till we have something solid, should we also put this behind a feature flag?

vivek-koppuru: I see, I think it's okay to not need a feature flag for this because these aren't values we display in the CLI, and we specifically validate against Ubuntu and Bottlerocket here: https://github.com/aws/eks-anywhere/blob/39d53e7d1e49252d27d14367db9048e770f4128b/pkg/providers/vsphere/validator.go#L90

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: @vignesh-goutham do you remember if we ever wrote something like this? If not, I can put together a doc quickly that describes all these attributes and what they represent and why they're needed

pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
maxdrib: Will remove until this is needed at a later date then

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdeploymentconfigs.yaml
maxdrib: The reason we decided to go with CloudStackDeploymentConfig originally was that there is not necessarily a 1:1 mapping between a datacenter and a particular CloudStack deployment/endpoint. A CloudStack deployment may manage multiple datacenters, aka CloudStack "zones", and when we are executing API's, it is against a particular CloudStack deployment.

The fields we are adding align pretty closely with what is present in CAPC (https://github.com/aws/cluster-api-provider-cloudstack) but I'll have to look for a doc that outlines each one specifically.

pkg/api/v1alpha1/cluster_types.go
maxdrib: Our customer has a usecase where they inject their own load balancer while the cluster is being created, as explained in https://quip-amazon.com/TwUbAhygIjuP/Expose-extra-SANs-attribute-for-Control-Plane-Load-Balancing. 

I see your point about needing to support this for other providers, however. @vignesh-goutham do you have any inputs about whether this is still needed and/or how we should address it moving forward?

pkg/api/v1alpha1/cloudstackdeploymentconfig_webhook.go
maxdrib: Yep, we certainly can

pkg/api/v1alpha1/cloudstackdeploymentconfig_types.go
maxdrib: Yes, looks like we are, good catch.

pkg/api/v1alpha1/cluster_types.go
vivek-koppuru: Let's call this `certSANs` to keep that consistent with kubeadm configuration

maxdrib: Discussed offline - decided that this was not a blocker. If we want to move forward and add it, then add some validation for other providers like https://github.com/aws/eks-anywhere/blob/bfaa4d37237cdc3c8aeef58545a9bec212fc0a55/pkg/providers/docker/docker.go#L99, and also create an issue to add support for other providers

maxdrib:
Discussed offline with Vivek, who conferred with Guillermo, and they both understand the current and future limitations and implications of a multi-zone deployment, and feel that the Datacenter nomenclature is more applicable for the time being

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,160 @@
...
+
+// CloudStackMachineConfigSpec defines the desired state of CloudStackMachineConfig
+type CloudStackMachineConfigSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Template refers to a VM image template which has been previously registered in CloudStack. It can either be specified as a UUID or name
+	Template string `json:"template"`
+	// ComputeOffering refers to a compute offering which has been previously registered in CloudStack. It represents a VM’s instance size including number of CPU’s, memory, and CPU speed. It can either be specified as a UUID or name
+	ComputeOffering string `json:"computeOffering"`
```
g-gaston, line 32: curious: what is the reason this can be either an ID or a name? Why not two fields, one for name and another one for ID? Or just simply accept only one of them.
Is this something idiosyncratic to cloud stack?

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// DiskOffering refers to a disk offering which has been previously registered in CloudStack. It represents a VM’s disk size and IOPS (Quality of Service) for primary data storage
+	DiskOffering string `json:"diskOffering,omitempty"`
```
g-gaston, line 34: Is this also name or ID? in that case, same question

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,160 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// CloudStackMachineConfigSpec defines the desired state of CloudStackMachineConfig
+type CloudStackMachineConfigSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Template refers to a VM image template which has been previously registered in CloudStack. It can either be specified as a UUID or name
+	Template string `json:"template"`
```
g-gaston, line 30: If this is required, do we actually need the OSFamily?
VSphere uses OSFamily to indicate the OS but let the CLI choose and import the right template

Here it seems like we always have to pass the template

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// ComputeOffering refers to a compute offering which has been previously registered in CloudStack. It represents a VM’s instance size including number of CPU’s, memory, and CPU speed. It can either be specified as a UUID or name
+	ComputeOffering string `json:"computeOffering"`
+	// DiskOffering refers to a disk offering which has been previously registered in CloudStack. It represents a VM’s disk size and IOPS (Quality of Service) for primary data storage
+	DiskOffering string `json:"diskOffering,omitempty"`
+	// OSFamily is an EKS-A required field which represents the family of the guest OS image. Currently only “redhat“ is supported for CloudStack, which is also the default
+	OSFamily OSFamily `json:"osFamily,omitempty"`
+	// Details represents an arbitrary mapping of strings to strings, which are passed through to the CloudStack VM. Some example details which are commonly set on the VM include cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag, and rootdisksize
+	Details map[string]string `json:"details,omitempty"`
```
g-gaston, line 38: I could be missing something here but I would vote against this and in favor of something more structured. For example, if we want to allow users to configure cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag, and rootdisksize, then create 4 more fields, one for each parameter.

This follows a bit better the patterns we have in the current API. It also reduces our testing surface since we control/restrict the number of possible configurations.

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// Users utilizes cloud-init to configure ssh users which will be allowed to log in to the VM using private-key authentication. It consists of an array of objects containing the user name, as well as a list of their authorized public keys
```
g-gaston, line 39: super nit: I would advise not use the term cloud-init here, since some OS's might not support cloud-init and could need a different format.

pkg/api/v1alpha1/cluster_types.go
g-gaston: Just to clarify, are we keeping this or removing it?

pkg/api/v1alpha1/machine_types.go
g-gaston: I would suggest creating a different type for cloud stack OS's
That allows us to maintain two separate "collections" of values (is more semantic that anything else)

pkg/api/v1alpha1/cluster_types.go
vivek-koppuru: Looks like we are removing this, but I am fine with adding it as `certSANs` and having similar validations too.

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
vivek-koppuru: I guess another question here is: Is it possible to support auto-importing the template for the user similar to how we have it for VSphere?

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
vivek-koppuru: I would be curious to hear what combinations we can expect a user to specify here. If there really are too many to track, I think since these details are supported as simple key/value pairs, I think it's okay. But if there are the main ones and we don't anticipate the user wanting fine-grain control of these fields, then I agree with creating new fields for structure.

pkg/api/v1alpha1/machine_types.go
maxdrib: I'm not sure I agree - that'll lead to redundancy. Fundamentally, the os family concept is exactly the same between the providers, we just need to have proper validation for each one to make sure only the supported ones are allowed, right? 

If you feel strongly otherwise, I'm happy to disagree and commit though

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: I have refined the list of attributes which support UUID/name resolution in the next commit

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: is "os-specific startup scripts" a better way to capture that?

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: I'll ask some CloudStack experts and get back to you on this

pkg/api/v1alpha1/cluster_types.go
maxdrib: Will re-add

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: It sounds like the intention behind this field, even within CloudStack itself, is to support general-purpose arguments that needed to be added and available in the VM "object" in CloudStack's database. It's used as a catch-all API. I think it adds value in that if a customer wants to start using some new field specific to their use case, there's no code changes required in eks-a, or new release that has to be cut

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
g-gaston: Would it be enough to mention that these will be the users and keys authorized to ssh into the machines?

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
g-gaston: I get what you are saying
However, I would lean on the other side of the argument, which is that we want to control which fields a user can and can't set and that goes against the concept of a catch all API.

Part of the added value that EKS-A offers is the easiness of use and the reliability (achieved through extra validations and testing). And we get there by simplifying the CAPI and cloud provider APIs, adding our own abstraction on top of them. If we were gonna open this API to be able to set any kind of configuration, we would either have to:
* Exponentially increase our validations and testing matrix. I don't think we can realistically tackle this and even if we could, I don't think it will scale.
* Renounce to be able to cover and validate all the possible configuration. This would reduce the value of EKS-A.

That obviously doesn't mean the APIs have to be static. The idea is that we start with an opinionated proposal of what we think is best and iterate on it (adding, changing and removing) based on user's feedback and real case scenarios.

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
g-gaston: Interesting. What kind of validations need the `OSFamily`?

If this is not being used right now, I would vote to remove it. We can always add it later if we implement the auto-import functionality or something else that can make use of it.

pkg/api/v1alpha1/machine_types.go
g-gaston: Putting this convo on pause until [we decide](https://github.com/aws/eks-anywhere/pull/1076#discussion_r801148963) if we even need `OSFmaily` in the cloud stack machines :)

pkg/api/v1alpha1/cluster_types.go
g-gaston: I'm confused
Do we need it?
If we don't, I vote to remove
If we do, then I think we should support it for all providers. But we should be aware that this increases the scope and adds more work on the testing side

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -114,6 +114,47 @@ func generateClusterConfig(clusterName string) error {
...
+				return fmt.Errorf("error outputting yaml: %v", err)
+			}
+			etcdMcYaml, err := yaml.Marshal(etcdMachineConfig)
+			if err != nil {
+				return fmt.Errorf("error outputting yaml: %v", err)
+			}
+			machineGroupYaml = append(machineGroupYaml, cpMcYaml, workerMcYaml, etcdMcYaml)
+		} else {
+			return fmt.Errorf("The cloudstack infrastructure provider is still in development!")
+		}
```
g-gaston, line 157: nit: you could avoid nesting (and make cleanup easier) if you move this to the top, like
```go
if !features.IsActive(features.CloudStackProvider()) {
   return fmt.Errorf("The cloudstack infrastructure provider is still in development!")
}
```

g-gaston (approved):
Looks great!

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
g-gaston: We are gonna do more research about this and decide later if we need it or not
Tracking that work in #1187

PR #1084, opened 2022-01-31
Adding CloudMonkey wrapper

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds a wrapper for CloudMonkey (cmk) to enable pre-flight check calls, along with unit tests. Cmk is a CLI tool for interacting with a cloudstack environment - more details at https://github.com/apache/cloudstack-cloudmonkey. This functionality can be utilized once the cmk executable is being built into the cli-tools image. 


pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func (c *Cmk) SearchTemplate(ctx context.Context, domain string, zone string, account string, template string) (string, error) {
+	templates, err := c.ListTemplates(ctx, template)
+	if err != nil {
+		return "", fmt.Errorf("search template %s error: %v", template, err)
+	} else if len(templates) > 1 {
+		return "", fmt.Errorf("duplicate templates %s found", template)
+	} else if len(templates) == 0 {
+		return "", fmt.Errorf("template %s not found", template)
```
g-gaston, line 55: nit: no need to change but you could return `"", nil` here and let the caller decide if not finding the the template is an error or not

If you prefer to leave it as is, maybe you could rename the method to `FindTemplate`. I think that's a bit more explicit about expecting the template to exist.

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return templates[0].Name, nil
```
g-gaston, line 57: What is the difference between the input `template` and this? Is it like full path or something like that?

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+const (
+	maxRetriesCmk    = 5
+	backOffPeriodCmk = 5 * time.Second
+)
+
+type Cmk struct {
+	writer     filewriter.FileWriter
+	executable Executable
+	retrier    *retrier.Retrier
+	execConfig *cmkExecConfig
```
g-gaston, line 45: Why does this need to be stored in the struct? If it's transient, can we just create a new one on each command execution?

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	logger.MarkPass("Connected to CloudStack server")
+	return nil
+}
+
+// TODO: Add support for domain, account filtering
+func (c *Cmk) ListTemplates(ctx context.Context, template string) ([]types.CmkTemplate, error) {
+	result, err := c.execWithNameAndIdFilters(ctx, template, "list", "templates", "templatefilter=all", "listall=true")
+	if err != nil {
+		return make([]types.CmkTemplate, 0), fmt.Errorf("error getting templates info: %v", err)
```
g-gaston, line 131: you can just return
```suggestion
		return nil, fmt.Errorf("error getting templates info: %v", err)
```

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+	buffer, err := c.exec(ctx, "sync")
+	if err != nil {
+		return fmt.Errorf("error validating cloudstack connection for cmk config %s: %v", buffer.String(), err)
+	}
+	logger.MarkPass("Connected to CloudStack server")
+	return nil
+}
+
+// TODO: Add support for domain, account filtering
+func (c *Cmk) ListTemplates(ctx context.Context, template string) ([]types.CmkTemplate, error) {
```
g-gaston, line 128: I find a bit confusing this method
If it's listing templates, why does it take a template as input?
I think I'm missing something :)

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+	err = json.Unmarshal(result.Bytes(), &response)
+	if err != nil {
+		return make([]types.CmkAffinityGroup, 0), fmt.Errorf("failed to parse response into json: %v", err)
+	}
+	return response.CmkAffinityGroups, nil
+}
+
+func (c *Cmk) execWithNameAndIdFilters(ctx context.Context, parameterValue string, genericArgs ...string) (stdout bytes.Buffer, err error) {
+	argsWithNameFilterArg := append(genericArgs, fmt.Sprintf("name=\"%s\"", parameterValue))
+	argsWithIdFilterArg := append(genericArgs, fmt.Sprintf("id=\"%s\"", parameterValue))
```
g-gaston, line 249: nit: can we build this only if we need it?

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+		return fmt.Errorf("failed to generate absolute filepath for generated config file at %s", writtenFileName)
+	}
+
+	return nil
+}
+
+func (c *Cmk) getEnvMap() (map[string]string, error) {
+	envMap := make(map[string]string)
+	for _, key := range requiredEnvsCmk {
+		if env, ok := os.LookupEnv(key); ok && len(env) > 0 {
```
g-gaston, line 333: Could we avoid passing info in env vars? I know we have used this pattern in the past but it has proven to be a bit inflexible and cumbersome to expand

I would suggest passing all the necessary info as arguments to the `Cmk` constructor
If there are too many to be separate args, it's totally fine to create a `CmkConf` struct and pass it as input to the constructor

pkg/executables/cmk_test.go
```diff
@@ -0,0 +1,564 @@
...
+
+			if resultCount != tt.wantResultCount {
+				t.Fatalf("Cmk call returned = %d results, want %d", resultCount, tt.wantResultCount)
+			}
+		})
+	}
+	restoreEnv()
+}
+
+type TestCmkClient interface {
```
g-gaston, line 556: nit: curious about what's the purpose of this interface

pkg/types/resources.go
```diff
@@ -57,6 +57,44 @@ type ClusterStatus struct {
 	Phase string
 }
 
+type CmkTemplate struct {
```
g-gaston, line 60: Curious about all these structs
Do they need to be public? And do they need to be here? They seem very much tied to the cloudstack api, so shouldn't they be encapsulated with cmk?

pkg/executables/cmk.go
maxdrib: The confusion is understandable. The method naming is based on how the cloudmonkey CLI itself is implemented. They don't expose get operations, only list with filters. I was trying to keep as much business logic as possible outside of these methods so they could be used easily by functions implementing business logic. If that's more confusing that valuable, I'll refactor to just have something like "ValidateTemplate" which makes sure there is exactly one template matching the criteria, otherwise return an error. And same for the other resource types.

pkg/executables/cmk.go
maxdrib: you're suggesting I move this to line 255? I can do that, sure

pkg/executables/cmk.go
maxdrib: I'll re-evaluate the contract between this library and the business logic where it'll be consumed to determine what to return here in the next revision

pkg/executables/cmk.go
maxdrib: input template may be either a template name or ID. This will always return the human readable name

pkg/executables/cmk.go
maxdrib: Agreed, the current contract/naming are unclear and confusing. I'm leaning towards renaming these methods to FindXXXX and re-defining the contract accordingly

pkg/executables/cmk.go
```diff
@@ -0,0 +1,363 @@
...
+)
+
+type Cmk struct {
+	writer     filewriter.FileWriter
+	executable Executable
+	retrier    *retrier.Retrier
+	execConfig *cmkExecConfig
+}
+
+func (c *Cmk) SearchTemplate(ctx context.Context, domain string, zone string, account string, template string) (string, error) {
```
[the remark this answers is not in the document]
maxdrib: Thanks for the tip about repeated types - I'm still a bit new to Go :)

pkg/executables/cmk.go
maxdrib: Thanks for the review! This aligns with the pattern established by govc. Is that a sufficient reason? I'm not entirely sure what the alternative is here, when you suggest "create a new one on each command execution

pkg/types/resources.go
maxdrib: They are tied to the cloudstack API. I thought I recalled seeing some vsphere structs in here and thought it would be fitting to add cloudstack structs here as well. I'll find somewhere else to put them

pkg/executables/cmk.go
maxdrib: To be clear, passing in data to the eks-a executable should still be done with environment variables, correct? You're just suggesting passing data through to Cmk specifically not be done with environment variables?

pkg/executables/cmk.go
g-gaston: Sorry, my bad, I didn't explain it correctly.
I was only referring to `execConfig`

I can see a comment:
```go
// It must be cleaned after each execution to prevent side effects from past executions options
```

This makes me think it's transient, hence we don't need to store it in the `Cmk` struct and we can just build it ad-hoc for each command. Which it seems like we are doing it now with `setupExecConfig`. So I'm suggesting doing the same thing but without storing it in `Cmk`.

pkg/executables/cmk.go
g-gaston: Yeah, totally agree, we should pass the "secret" data to the eks-a cli through env vars. But that doesn't mean we need to read them from `Cmk`. We can just read them at a higher level (eg. the `cmd` package) and pass them down to the `Cmk` constructor.

My main reasons to suggest this:
* Making testing easier (instead of setting env vars in the test, we just pass strings to the constructor). This facilitates parallel testing.
* Portability: we will be able to use this from a controller, where the secrets won't come from env vars (most probably from something like kubernetes secrets)
* Flexibility: it will allow us to have multiple `Cmk` instances, each with a different set of credentials (this could be useful when having to contact multiple "datacenters")

I know none of the things I just mentioned are necessarily requirements for this current usecase. But we learnt with `govc` that using env vars at the executable level presents all those problems and we are gonna need to refactor it. So trying to avoid that same scenario here 🙂

pkg/executables/cmk.go
g-gaston: Oh I 100% support leaving the business logic out of this file. In fact out of `executables` completely (this is a mistake we have made in the past).

I think writing methods that follow `cmk` api is totally fine. Maybe we can just rename things a bit to avoid confusion among people who don't know a lot about the API (like me). For example, instead of `ListTemplates` we could go with `FilterTemplates`. And instead of `template` we could use `filterValue`.

That said, what is the reason the filter value can be name or ID? Doesn't the caller know what is the value they have when calling this method? We could just have different methods, like `FilterTemplatesByName` and `FilterTemplatesByID`.

pkg/executables/cmk.go
```diff
@@ -0,0 +1,371 @@
...
+	buffer, err := c.exec(ctx, "sync")
+	if err != nil {
+		return fmt.Errorf("error validating cloudstack connection for cmk config %s: %v", buffer.String(), err)
+	}
+	logger.MarkPass("Connected to CloudStack server")
+	return nil
+}
+
+func (c *Cmk) execWithNameAndIdFilters(ctx context.Context, parameterValue string, genericArgs ...string) (stdout bytes.Buffer, err error) {
+	argsWithNameFilterArg := append(genericArgs, fmt.Sprintf("name=\"%s\"", parameterValue))
```
vivek-koppuru, line 235: I am a little confused by the name `parameterValue`. Can we just call it `name` or `resourceName`? Also, I think this function is better named `execWithNameWithIdFiltersRetry` to describe what it is doing, but maybe that is confusing too.

pkg/executables/cmk.go
```diff
@@ -0,0 +1,371 @@
...
+	apiUrl, err := section.GetKey("api-url")
+	if err != nil {
+		return fmt.Errorf("failed to extract value of 'api-url' from %s: %v", cloudStackb64EncodedSecretKey, err)
+	}
+	cmkInsecure, err := strconv.ParseBool(envMap[cloudmonkeyInsecureKey])
+	if err != nil {
+		return fmt.Errorf("failed to parse boolean value from %s: %v", cloudmonkeyInsecureKey, err)
+	}
+	cmkVerifyCert := strconv.FormatBool(!cmkInsecure)
+	t := templater.New(c.writer)
```
vivek-koppuru, line 298: So this file would be written under the `generated` folder right?

pkg/executables/cmk.go
maxdrib: Thanks for the clarification. I believe the comment may have been copied/pasted from the govc implementation unintentionally so I'll take a closer look at this.

pkg/executables/cmk.go
maxdrib: The reasoning for this naming is we're unsure at this point if the input value is a resource name or unique ID. I thought "parameter" might be the right word to convey that ambiguity, so I'm inclined to push back against a variable name including `name`. I'm happy to rename the function, however - I think that's a good idea. Let me know if you disagree

pkg/executables/cmk.go
```diff
@@ -0,0 +1,371 @@
...
+func (c *Cmk) execWithNameAndIdFilters(ctx context.Context, parameterValue string, genericArgs ...string) (stdout bytes.Buffer, err error) {
+	argsWithNameFilterArg := append(genericArgs, fmt.Sprintf("name=\"%s\"", parameterValue))
+	result, err := c.exec(ctx, argsWithNameFilterArg...)
+	if err != nil {
+		return result, fmt.Errorf("error getting resource info filtering by id %s: %v", parameterValue, err)
+	}
+	if result.Len() == 0 {
+		msg := fmt.Sprintf("No resources found with name %s. Trying again filtering by id instead", parameterValue)
+		argsWithIdFilterArg := append(genericArgs, fmt.Sprintf("id=\"%s\"", parameterValue))
+		logger.V(6).Info(msg)
```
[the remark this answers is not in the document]
maxdrib: Will make these changes, thanks for the tip

pkg/executables/cmk.go
maxdrib: That's right!

pkg/executables/cmk.go
maxdrib: I see. So you're suggesting we parse the input environment variable at a higher level into some meaningful struct containing CloudStack credentials and other information, and then pass that struct in when executing cloudmonkey?

pkg/executables/cmk.go
```diff
@@ -0,0 +1,308 @@
...
+)
+
+type Cmk struct {
+	writer     filewriter.FileWriter
+	executable Executable
+	retrier    *retrier.Retrier
+}
+
+// TODO: Add support for domain, account filtering
+func (c *Cmk) ValidateTemplatePresent(ctx context.Context, config CmkExecConfig, domain, zone, account, template string) error {
```
g-gaston, line 35: If you want to simplify these interfaces and you don't mind tiding an instance of `Cmk` to a particular set of CloudStack Cluster and creds (which I think it's ok), you could just pass this to the constructor and have store in `Cmk`

pkg/executables/cmk.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	if result.Len() == 0 {
+		return fmt.Errorf("template %s not found", template)
+	}
+
+	response := struct {
+		CmkTemplates []CmkTemplate `json:"template"`
+	}{}
+	err = json.Unmarshal(result.Bytes(), &response)
+	if err != nil {
```
g-gaston, line 48: nit: you can do
```suggestion
    if err = json.Unmarshal(result.Bytes(), &response); err != nil {
```

pkg/executables/cmk.go
```diff
@@ -0,0 +1,308 @@
...
+const (
+	cmkConfigFileName = "cmk_tmp.ini"
+	maxRetriesCmk     = 5
+	backOffPeriodCmk  = 5 * time.Second
+)
+
+type Cmk struct {
+	writer     filewriter.FileWriter
+	executable Executable
+	retrier    *retrier.Retrier
```
g-gaston, line 31: Are we using this? If not, we can just remove it

pkg/executables/cmk.go
```diff
@@ -0,0 +1,308 @@
...
+	ApiKey        string // Api Key for CloudMonkey to access CloudStack Cluster
+	SecretKey     string // Secret Key for CloudMonkey to access CloudStack Cluster
+	ManagementUrl string // Management Endpoint Url for CloudMonkey to access CloudStack Cluster
+	VerifyCert    bool   // boolean indicating if CloudMonkey should verify the cert presented by the CloudStack Management Server
+}
+
+// TODO: Add support for passing in domain from Deployment Config Spec
+func (c *Cmk) buildCmkConfigFile(config CmkExecConfig) (configFile string, err error) {
+	t := templater.New(c.writer)
+	data := map[string]string{
```
g-gaston, line 254: If you name the "variables" in the template with the same name as the struct fields, you can pass it directly without having to create a map

pkg/executables/cmk.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	configFile, err = filepath.Abs(writtenFileName)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate absolute filepath for generated config file at %s", writtenFileName)
+	}
+
+	return configFile, nil
+}
+
+type CmkTemplate struct {
```
g-gaston, line 272: nit: I think we can make these structs private (starting with lowercase)

pkg/executables/cmk.go
g-gaston: @maxdrib do all the `Validate...` methods take a name or ID? Or some of them only take names and others only take IDs?

pkg/executables/cmk.go
maxdrib: Good point, refactored the struct

pkg/executables/cmk.go
maxdrib: I'll make these one-liners wherever possible - thanks for the hint

maxdrib:
>@maxdrib do all the Validate... methods take a name or ID? Or some of them only take names and others only take IDs?

It varies based on what CloudMonkey supports and what usecases we support. For example with affinity groups, CAPC only supports passing in a list of UUID's, not names, so we only query cmk by ID. I think the opposite is true if/when we validate the network (which is a TODO). There's [no parameter for "name" when listing networks](https://cloudstack.apache.org/api/apidocs-4.14/apis/listNetworks.html)

pkg/executables/cmk.go
maxdrib: It varies based on what CloudMonkey supports and what usecases we support. For example with affinity groups, CAPC only supports passing in a list of UUID's, not names, so we only query cmk by ID. I think the opposite is true if/when we validate the network (which is a TODO). There's [no parameter for "name" when listing networks](https://cloudstack.apache.org/api/apidocs-4.14/apis/listNetworks.html), so we can only query by ID

pkg/executables/cmk.go
g-gaston: Interesting. Not big deal either way but it should work. Not sure if that was the case but even though the structs can be private, the fields need to be public.

pkg/executables/cmk.go
g-gaston: Got it. In that case I would suggest having separate methods for ID and name. If there is a type of resource that needs both, I think that's totally ok.

pkg/executables/cmk.go
```diff
@@ -0,0 +1,265 @@
...
+const (
+	cmkConfigFileName = "cmk_tmp.ini"
+	maxRetriesCmk     = 5
+	backOffPeriodCmk  = 5 * time.Second
+)
+
+type Cmk struct {
+	writer     filewriter.FileWriter
+	executable Executable
+	retrier    *retrier.Retrier
```
g-gaston, line 31: Are we using this? I can't find any references :)

pkg/executables/cmk.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	configFile, err = filepath.Abs(writtenFileName)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate absolute filepath for generated config file at %s", writtenFileName)
+	}
+
+	return configFile, nil
+}
+
+type CmkTemplate struct {
```
g-gaston, line 235: nit: did you get to try and make these structs private but keep the fields public?

g-gaston (approved):
A couple of nits, but looks great

pkg/executables/cmk.go
maxdrib: Not yet - will use this in the provider factory when integrating with the provider

pkg/executables/cmk.go
maxdrib: I did try that, and deserialization failed :(

pkg/executables/cmk.go
g-gaston: ok, do you might leaving a comment there indicating just that? just so we don't lose track of the reason why this is here :)

vivek-koppuru (approved):
/lgtm

PR #1118, opened 2022-02-01
Design for new vcenter template import command

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

These changes came from users having the ability to import templates that we vend beforehand as well as make it easier to test with templates provided by our bundles in our e2e workflows.

### DIFF
--- a/designs/vcenter-template-import.md
+++ b/designs/vcenter-template-import.md
@@ -1,0 +1,91 @@
+# Import vcenter template before create
+
+## Introduction
+
+**Problem:** If a user doesn't have a template defined in the cluster spec, the create command 
+imports the latest available OVA for the specific version of the CLI. However, there is no way
+to import this OVA before running create without manually importing the OVA into a template
+in vcenter and specifying that template name in the cluster spec, even though we have that 
+automation logic in the CLI already. 
+
+### Goals and Objectives
+
+As an EKS Anywhere user:
+
+* I want to have the ability to pre-import the template before I run create
+* I want to specify the template I want to import in a simple manner
+* I want the create command to know to use the pre-imported template
+
+
+## Overview of Solution
+
+With this feature, a user can specify to import the latest OVA template before create by
+running a command. They would specify the cluster spec as input so that there is nothing 
+new that they need to know in terms of flags or different configuration. We would be introducing new
+subcommands as follows:
+
+```
+eksctl anywhere import vsphere template -f <cluster_spec.yaml>
+```
+
+
+### Solution Details
+
+With this feature, users will be able to import the latest OVA available based on the Kubernetes version
+and OS family defined in the cluster spec. We would introduce a new 
+`import` subcommand. Since this feature is specific to vsphere, we will introduce a `vsphere` 
+subcommand to give flexibility of the types of resources that we can import under 
+vsphere. This also gives us flexibility to import provider-agnostic resources under the `import`
+subcommand. Since this is importing the OVA into a vcenter template, we will be calling
+the final subcommand `template` based on the resource name. 
+
+Currently for the `create` subcommand, users can specify either a template if they already
+manually imported the OVA into a template in vcenter, or not specify the template field or leave the
+field empty if they want it imported via a generated name that is based on the Kubernetes
+version, OS, and hash of the build. 
+
+With this command, if a user has the template name defined, we will check to see if a template of that 
+name already exists and the appropriate tags are set. If the template does not exist, we will import the template under
+that name with the appropriate tags. If the template exists but the tags don't match the bundle 
+that the CLI is configured with, we will error saying that the appropriate tags are not set, allowing
+the user to decide if they want to manually fix the tags or delete the template first before running
+the command again. 
+
+If a user does not have the template name defined, then we will do what we already do today in the `create` command, 
+which is importing the template under the name that we have predefined and setting the appropriate tags 
+for the user. 
+
+For both cases, we will have output saying that the template has been imported under the 
+specific name set in vcenter. If the user did not have the template name defined, the `create`
+cluster will still use the template that was imported before because the name would match.
+It would only change if the user ran `import`, upgraded the CLI (or the bundle got updated), and then 
+ran `create`.
+
+### Alternatives
+
+We could also solve this by not passing in the cluster spec file and just take in all the values
+as flags, but having a consistent cluster spec driven approach for cluster related actions
+will give the user a less confusing experience when trying to interact with all the commands.
+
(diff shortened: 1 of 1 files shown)


jaxesn (commented):
Just some thoughts, nothing blocking

designs/vcenter-template-import.md
```diff
@@ -0,0 +1,91 @@
...
+
+For both cases, we will have output saying that the template has been imported under the 
+specific name set in vcenter. If the user did not have the template name defined, the `create`
+cluster will still use the template that was imported before because the name would match.
+It would only change if the user ran `import`, upgraded the CLI (or the bundle got updated), and then 
+ran `create`.
+
+### Alternatives
+
+We could also solve this by not passing in the cluster spec file and just take in all the values
```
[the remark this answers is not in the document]
vivek-koppuru: Sure, I didn't really think about it from the ops admin vs someone who is familiar with the cluster spec. My other main point of reference however is the `import-images` command which I would assume would be done by a similar user, unless we are saying that use case is different?

designs/vcenter-template-import.md
```diff
@@ -0,0 +1,91 @@
...
+the final subcommand `template` based on the resource name. 
+
+Currently for the `create` subcommand, users can specify either a template if they already
+manually imported the OVA into a template in vcenter, or not specify the template field or leave the
+field empty if they want it imported via a generated name that is based on the Kubernetes
+version, OS, and hash of the build. 
+
+With this command, if a user has the template name defined, we will check to see if a template of that 
+name already exists and the appropriate tags are set. If the template does not exist, we will import the template under
+that name with the appropriate tags. If the template exists but the tags don't match the bundle
```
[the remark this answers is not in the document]
vivek-koppuru: I think that is something that with the cluster spec we can loop through and document, so I wouldn't be concerned with the complexity. But I understand the lack of clarity and it would depend on who our end user is. I can add that section to this doc too, but I want to make sure we are in alignment with `import-images` or at least how they differ.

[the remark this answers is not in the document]
vivek-koppuru: Updated doc to include both types of users

designs/vcenter-template-import.md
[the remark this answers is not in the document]
vivek-koppuru: Updated doc to show that we will support both cluster spec and flags

designs/vcenter-template-import.md
```diff
@@ -16,18 +16,27 @@ As an EKS Anywhere user:
 * I want to specify the template I want to import in a simple manner
 * I want the create command to know to use the pre-imported template
 
+### Users
+
+There are two types of users who would use this functionality. 
+
+* EKS-A maintainers who are in charge of creating the cluster
```
g-gaston, line 23: nit: EKS-A users or maybe EKS-A cluster admins
Not big deal, but EKS-A maintainers sounds more like it's us :)

designs/vcenter-template-import.md
```diff
@@ -16,18 +16,27 @@ As an EKS Anywhere user:
...
+running a command. They would either specify the cluster spec or the appropriate flags as input. 
+We would be introducing new subcommands as follows:
 
 ```
-eksctl anywhere import vsphere template -f <cluster_spec.yaml>
+eksctl anywhere import vsphere template -f <cluster_spec.yaml> --name <template_name>
+```
+or
+```
+eksctl anywhere import vsphere template --name <template_name> --path <vsphere_template_path> --os <os_family> --server <server_url>
```
g-gaston, line 38: nit: we might need a couple more flags here like datastore. We could just add `...` at the end to represent this.

Also, maybe we could simplify the interface by combining `--path` and `--name` into `--full-path`?

designs/vcenter-template-import.md
```diff
@@ -44,28 +53,33 @@ manually imported the OVA into a template in vcenter, or not specify the templat
...
-With this command, if a user has the template name defined, we will check to see if a template of that 
-name already exists and the appropriate tags are set. If the template does not exist, we will import the template under
-that name with the appropriate tags. If the template exists but the tags don't match the bundle 
-that the CLI is configured with, we will error saying that the appropriate tags are not set, allowing
-the user to decide if they want to manually fix the tags or delete the template first before running
-the command again. 
+If the user specifies the cluster spec as input, we will get most of the template configuration
+from there except for the name. If there is a name defined, we will throw an error mentioning
+that they have a template name defined and to remove the template name if they want to import as
+we want to get the name either from the flag. If they choose not to specify a flag for the template name,
```
g-gaston, line 59:
```suggestion
we want to get the name from the flag. If they choose not to specify a flag for the template name,
```

designs/vcenter-template-import.md
```diff
@@ -44,28 +53,33 @@ manually imported the OVA into a template in vcenter, or not specify the templat
...
 version, OS, and hash of the build. 
 
-With this command, if a user has the template name defined, we will check to see if a template of that 
-name already exists and the appropriate tags are set. If the template does not exist, we will import the template under
-that name with the appropriate tags. If the template exists but the tags don't match the bundle 
-that the CLI is configured with, we will error saying that the appropriate tags are not set, allowing
-the user to decide if they want to manually fix the tags or delete the template first before running
-the command again. 
+If the user specifies the cluster spec as input, we will get most of the template configuration
+from there except for the name. If there is a name defined, we will throw an error mentioning
```
g-gaston, line 57: Should this be only if the full path from the flags doesn't match the one in the spec?
Like the user could just pass the same path to both in spec and flags, and that would work.

designs/vcenter-template-import.md
```diff
@@ -44,28 +53,33 @@ manually imported the OVA into a template in vcenter, or not specify the templat
...
+name or not. 
 
 For both cases, we will have output saying that the template has been imported under the 
 specific name set in vcenter. If the user did not have the template name defined, the `create`
 cluster will still use the template that was imported before because the name would match.
 It would only change if the user ran `import`, upgraded the CLI (or the bundle got updated), and then 
-ran `create`.
+ran `create`. The template would also be imported with the appropriate tags with any of the options.
+
+If the user specifies both cluster spec and flags, we will throw an error mentioning to choose
```
g-gaston, line 73: Maybe we could just say here than flags take precedence over values in cluster spec, instead of throwing an error?

g-gaston (approved):

designs/vcenter-template-import.md
vivek-koppuru: Good point

g-gaston (approved):

PR #1119, opened 2022-02-01
Remove worker node groups during upgrade

taneyland:
*Issue #, if available:*

*Description of changes:*
These changes add the ability to remove worker node groups during upgrade. The user can simply remove the unwanted worker node group configuration from the cluster config file, run upgrade, and the machine deployments & nodes associated with that worker node group will be deleted. 

These changes also update the defaulting method for the first worker node group. It now defaults the first worker node group if one *or more* groups are defined. This removes the need for users to manually set the first worker node group to `md-0` when adding new worker node groups to a cluster that was created with only one worker node group with no name specified.


pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -45,8 +45,8 @@ func setRegistryMirrorConfigDefaults(clusterConfig *Cluster) error {
 }
 
 func setWorkerNodeGroupDefaults(cluster *Cluster) error {
-	if len(cluster.Spec.WorkerNodeGroupConfigurations) == 1 && cluster.Spec.WorkerNodeGroupConfigurations[0].Name == "" {
-		logger.V(1).Info("Worker node group name not specified. Defaulting name to md-0.")
+	if len(cluster.Spec.WorkerNodeGroupConfigurations) >= 1 && cluster.Spec.WorkerNodeGroupConfigurations[0].Name == "" {
```
g-gaston, line 48: I think I would vote against this
I think if users are going to use the new feature, they should use the new syntax

Curious, what's the reason to propose this change?

pkg/clustermanager/cluster_manager.go
```diff
@@ -833,6 +838,21 @@ func (c *ClusterManager) waitForAllControlPlanes(ctx context.Context, cluster *t
 	return nil
 }
 
+func (c *ClusterManager) removeOldMachineDeployments(ctx context.Context, workloadCluster *types.Cluster, provider providers.Provider, currentSpec, newSpec *cluster.Spec) error {
+	workerConfigs := provider.BuildMapForWorkerNodeGroupsByName(newSpec.Spec.WorkerNodeGroupConfigurations)
+	for _, prevWorkerNodeGroupConfig := range currentSpec.Spec.WorkerNodeGroupConfigurations {
+		if _, ok := workerConfigs[prevWorkerNodeGroupConfig.Name]; !ok {
+			workerNodeGroupName := fmt.Sprintf("%s-%s", workloadCluster.Name, prevWorkerNodeGroupConfig.Name)
+			err := c.clusterClient.DeleteMachineDeployment(ctx, workloadCluster.KubeconfigFile, workerNodeGroupName)
```
g-gaston, line 846: Does CAPI take care of deleting the `KubeadmConfigTemplate` and the provider machine template?

pkg/clustermanager/cluster_manager.go
```diff
@@ -833,6 +838,21 @@ func (c *ClusterManager) waitForAllControlPlanes(ctx context.Context, cluster *t
 	return nil
 }
 
+func (c *ClusterManager) removeOldMachineDeployments(ctx context.Context, workloadCluster *types.Cluster, provider providers.Provider, currentSpec, newSpec *cluster.Spec) error {
+	workerConfigs := provider.BuildMapForWorkerNodeGroupsByName(newSpec.Spec.WorkerNodeGroupConfigurations)
```
g-gaston, line 842: curious why does this need to come from the provider?
Are you thinking the provider might have some hidden logic that you want to abstract here?

Asking because it seems that this is only creating a set of strings and we are using to speed up the "search" of deleted worker node groups.

It would be different if the provider returned the name of the machine deployments.

pkg/api/v1alpha1/cluster_defaults.go
taneyland: I proposed this change because if the user created the cluster with a version before 0.7.0, they would have to set the `name` of their first worker node group to `md-0` to have the node group supported during upgrade. If they set it to any other value, the original worker node group would be deleted during upgrade & it would create a new worker node group based on that new name.

pkg/clustermanager/cluster_manager.go
taneyland: Yes, good point. Nothing provider-specific is happening here so it doesn't *need* to come from the provider. I added it because we need to access this method from the provider & cluster manager, so I added it here avoid redefining this method in cluster manger since we already have access to the provider.

pkg/clustermanager/cluster_manager.go
taneyland: No it took care of deleting the nodes & machine deployments, but it left the old resources in the other two. Updated the method to delete all resources related to unwanted node group.

abhay-krishna:
Wow, Prow has become sophisticated, showing us [servals](https://en.m.wikipedia.org/wiki/Serval) and all now!

pkg/clustermanager/cluster_manager.go
```diff
@@ -833,6 +838,21 @@ func (c *ClusterManager) waitForAllControlPlanes(ctx context.Context, cluster *t
 	return nil
 }
 
+func (c *ClusterManager) removeOldWorkerNodeGroups(ctx context.Context, workloadCluster *types.Cluster, provider providers.Provider, currentSpec, newSpec *cluster.Spec) error {
+	workerConfigs := provider.BuildMapForWorkerNodeGroupsByName(newSpec.Spec.WorkerNodeGroupConfigurations)
+	for _, prevWorkerNodeGroupConfig := range currentSpec.Spec.WorkerNodeGroupConfigurations {
+		if _, ok := workerConfigs[prevWorkerNodeGroupConfig.Name]; !ok {
+			workerNodeGroupName := fmt.Sprintf("%s-%s", workloadCluster.Name, prevWorkerNodeGroupConfig.Name)
```
g-gaston, line 845: This is making an assumption that all providers use the same naming. It's a correct assumption, but I think it goes a little bit against the idea of a provider interface.

What do you think about just returning the objects that need to be deleted from the provider? Like some kind of struct (there might be something already in one of the kubernetes api/runtime packages) that has the type, name and namespace.

That way we don't need to make naming assumptions and we simplify a lot the kubectl code. That keeps cluster manager and kubectl agnostic from the objects created/deleted by the providers.

We could move this method to "select" the worker nodes to delete to a shared package and use it from each provider to get the "to delete" worker nodes and return the associated capi objects.

pkg/executables/kubectl.go
```diff
@@ -318,6 +318,24 @@ func (k *Kubectl) ValidateNodes(ctx context.Context, kubeconfig string) error {
...
+	providerMachineTemplateName := md.Spec.Template.Spec.InfrastructureRef.Name
+	params := []string{"delete", fmt.Sprintf("machinedeployments.%s", clusterv1.GroupVersion.Group), md.Name, "--kubeconfig", kubeconfig, "--namespace", constants.EksaSystemNamespace}
+	if _, err := k.Execute(ctx, params...); err != nil {
+		return err
+	}
+	params = []string{"delete", fmt.Sprintf("kubeadmconfigtemplates.bootstrap.%s", clusterv1.GroupVersion.Group), kubeadmConfigTemplateName, "--kubeconfig", kubeconfig, "--namespace", constants.EksaSystemNamespace}
+	if _, err := k.Execute(ctx, params...); err != nil {
+		return err
+	}
+	params = []string{"delete", fmt.Sprintf("%smachinetemplates.infrastructure.%s", providerName, clusterv1.GroupVersion.Group), providerMachineTemplateName, "--kubeconfig", kubeconfig, "--namespace", constants.EksaSystemNamespace}
```
g-gaston, line 332: Do we need the provider name? The machine deployment should have the type it references right? So we maybe extract it from there to pass it to kubectl instead of passing the provider name.

pkg/providers/docker/docker.go
```diff
@@ -463,10 +463,26 @@ func (p *provider) RunPostControlPlaneCreation(ctx context.Context, clusterSpec
...
+		if _, ok := workerConfigs[prevWorkerNodeGroupConfig.Name]; !ok {
+			workerNodeGroupName := fmt.Sprintf("%s-%s", workloadCluster.Name, prevWorkerNodeGroupConfig.Name)
+			machineDeployment, err := p.providerKubectlClient.GetMachineDeployment(ctx, workerNodeGroupName, executables.WithKubeconfig(workloadCluster.KubeconfigFile), executables.WithNamespace(constants.EksaSystemNamespace))
+			if err != nil {
+				return nil, err
+			}
+			machineDeployments = append(machineDeployments, machineDeployment)
+		}
+	}
+	return machineDeployments, nil
```
g-gaston, line 487: There is a lot of duplicated code in between providers for this method. Since the only difference (if any) should be the naming, what do you think about:
* Move the logic to determine the removed worker node groups somewhere else. Maybe in the `cluster` package. Just a standalone method that takes a cluster config and returns a slice of worker node group configs.
* Return only the machine deployment names from this method and move the `GetMachineDeployment` logic to cluster manager.

pkg/providers/docker/docker.go
```diff
@@ -463,10 +463,26 @@ func (p *provider) RunPostControlPlaneCreation(ctx context.Context, clusterSpec
...
-	return prevConfigs
+	return workerNodeGroupConfigs
+}
+
+func (p *provider) NodeGroupsToDelete(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newSpec *cluster.Spec) ([]*clusterv1.MachineDeployment, error) {
+	workerConfigs := buildMapForWorkerNodeGroupsByName(newSpec.Spec.WorkerNodeGroupConfigurations)
+	machineDeployments := make([]*clusterv1.MachineDeployment, 0, len(currentSpec.Spec.WorkerNodeGroupConfigurations))
+	for _, prevWorkerNodeGroupConfig := range currentSpec.Spec.WorkerNodeGroupConfigurations {
+		if _, ok := workerConfigs[prevWorkerNodeGroupConfig.Name]; !ok {
+			workerNodeGroupName := fmt.Sprintf("%s-%s", workloadCluster.Name, prevWorkerNodeGroupConfig.Name)
```
g-gaston, line 479: Do you think we could move this to a method so we can reuse it from the two places we need it (building the template and here)?

pkg/providers/provider.go
```diff
@@ -36,6 +38,7 @@ type Provider interface {
 	UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec) (bool, error)
 	DeleteResources(ctx context.Context, clusterSpec *cluster.Spec) error
 	RunPostControlPlaneCreation(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error
+	NodeGroupsToDelete(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newSpec *cluster.Spec) ([]*clusterv1.MachineDeployment, error)
```
g-gaston, line 41: Should this be `MachineDeploymentsToDelete`?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -425,3 +437,19 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, workerNodeGroupMachineSpec v1
 	}
 	return values
 }
+
+func (p *tinkerbellProvider) NodeGroupsToDelete(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newSpec *cluster.Spec) ([]*clusterv1.MachineDeployment, error) {
+	workerConfigs := buildMapForWorkerNodeGroupsByName(newSpec.Spec.WorkerNodeGroupConfigurations)
+	machineDeployments := make([]*clusterv1.MachineDeployment, 0, len(currentSpec.Spec.WorkerNodeGroupConfigurations))
+	for _, prevWorkerNodeGroupConfig := range currentSpec.Spec.WorkerNodeGroupConfigurations {
+		if _, ok := workerConfigs[prevWorkerNodeGroupConfig.Name]; !ok {
+			mdName := fmt.Sprintf("%s-md-0", workloadCluster.Name)
```
g-gaston, line 446: Why is this `md-0`?

taneyland: Tinkerbell provider uses a `workerPoolName` to name the machineDeployment, and it is currently hard-coded to `md-0`. I need to check with the providers team to see how they want to handle naming for multiple groups, but for now it's always `md-0`

g-gaston: Does that mean that multiple worker node groups is not supported for tinkerbell?

pkg/providers/docker/docker.go
```diff
@@ -466,10 +466,12 @@ func (p *provider) RunPostControlPlaneCreation(ctx context.Context, clusterSpec
 	return nil
 }
 
-func (p *provider) BuildMapForWorkerNodeGroupsByName(workerNodeGroups []v1alpha1.WorkerNodeGroupConfiguration) map[string]v1alpha1.WorkerNodeGroupConfiguration {
-	workerNodeGroupConfigs := make(map[string]v1alpha1.WorkerNodeGroupConfiguration, len(workerNodeGroups))
-	for _, config := range workerNodeGroups {
-		workerNodeGroupConfigs[config.Name] = config
+func (p *provider) MachineDeploymentsToDelete(workloadCluster *types.Cluster, currentSpec, newSpec *cluster.Spec) []string {
+	nodeGroupsToDelete := cluster.NodeGroupsToDelete(currentSpec, newSpec)
+	machineDeployments := make([]string, 0, len(currentSpec.Spec.WorkerNodeGroupConfigurations))
```
g-gaston, line 471:
```suggestion
	machineDeployments := make([]string, 0, len(nodeGroupsToDelete))
```

pkg/providers/tinkerbell/tinkerbell.go
taneyland: Yes, correct. The logic is in place for it to work except for the naming in the tinkerbell `template-md.yaml`. I have to discuss with the providers team to understand how they want to go about it since the tinkerbell template has the extra `pool` field that is currently being used to name those CAPI objects

pkg/providers/docker/docker.go
g-gaston: gentle bump :)

pkg/providers/tinkerbell/tinkerbell.go
g-gaston: Cool. Do we have ticket for this? It doesn't matter if it's in the tinkerbell project as long as we have one

taneyland: I have it on my personal TODO list but I'll create a ticket on here to track it

g-gaston (approved):

PR #1129, opened 2022-02-01
Update docsy and change it to use static js imports

TerryHowe:
Update docsy and change it to use static js imports rather than CDN

It looks like all those js imports are MIT licenses


jaxesn (approved):
/lgtm

PR #1134, opened 2022-02-03
wngc comparison takes name, taints into proper account

danbudris:
*Issue #, if available:*

*Description of changes:*

Simplify and correct worker node group configuration slice comparisons:
- take `name` into account when comparing worker node groups
- ignore order of taints when comparing worker node group configuration taints
- use existing, correct method for taint slice comparison

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -148,9 +148,9 @@ func TaintsSliceEqual(s1, s2 []corev1.Taint) bool {
 	if len(s1) != len(s2) {
 		return false
 	}
-	taints := make(map[corev1.Taint]bool)
+	taints := make(map[corev1.Taint]struct{})
 	for _, taint := range s1 {
-		taints[taint] = true
+		taints[taint] = struct{}{}
 	}
 	for _, taint := range s2 {
 		_, ok := taints[taint]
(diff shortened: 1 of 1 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -200,6 +200,7 @@ type WorkerNodeGroupConfiguration struct {
 }
 
 func generateWorkerNodeGroupKey(c WorkerNodeGroupConfiguration) (key string) {
+	key = c.Name
 	if c.MachineGroupRef != nil {
 		key = c.MachineGroupRef.Kind + c.MachineGroupRef.Name
```
g-gaston, line 205: Can we remove this? Now all groups should have a name
I feel like this method can just be `return c.Name`

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -228,14 +229,7 @@ func WorkerNodeGroupConfigurationsSliceEqual(a, b []WorkerNodeGroupConfiguration
 		return false
 	}
```
g-gaston, line 230: I think we can simplify this logic
we can build a map[key]workernodegroup with `a`
Then iterate over b. We check every workernodegroup in `b` to see if it's in `a`. If it's not, return false. If it is, compare the two worker node groups (probably with a new method?)

pkg/api/v1alpha1/cluster_types.go
danbudris: if we're using this key we need the name to be a part of it; otherwise, a wngc with the same machine ref and count but a different name would return the same key.

pkg/api/v1alpha1/cluster_types.go
danbudris: do you mean map[name]workernodegroup?

danbudris: so I get what you're saying but this already does what we need it to

danbudris: with the only modification being the addition of the `name` to the key generation method

jiayiwang7 (approved):

PR #1142, opened 2022-02-04
Update the docs to link the demo

TerryHowe:
The docs still link to a live demo that happened four months ago. New look:

Screen Shot 2022-02-04 at 05 13 51


rothgar:
Thanks! I forgot I wanted to update that last year.

/lgtm

PR #1147, opened 2022-02-04
Pre-flight vallidation: Validate tinkerbell stack connection

panktishah26:
*Issue #, if available:*

*Description of changes:*
Added pre-flight validation for tinkerbell stack connection. 


pkg/providers/tinkerbell/tinkerbell_test.go
```diff
@@ -24,6 +24,16 @@ const (
 	expectedTinkerbellPBnJGRPCAuthority = "1.2.3.4:42000"
 )
 
+type DummyProviderTinkClient struct{}
+
+func NewDummyProviderTinkClient() *DummyProviderTinkClient {
```
chrisdoherty4, line 29: Does this need to be a pointer return?

pkg/providers/tinkerbell/tinkerbell_test.go
```diff
@@ -24,6 +24,16 @@ const (
 	expectedTinkerbellPBnJGRPCAuthority = "1.2.3.4:42000"
 )
 
+type DummyProviderTinkClient struct{}
```
g-gaston, line 27: nit: why build your own dummy if you are already generating the mocks for this interface?

pkg/executables/tink.go
```diff
@@ -18,6 +18,11 @@ type Tink struct {
 	envMap                  map[string]string
 }
 
+func (t *Tink) Close(ctx context.Context) error {
+	// TODO: implement close/logout functionality
```
g-gaston, line 22: curious: does tink maintain some kind of persistent session?

pkg/dependencies/factory.go
```diff
@@ -252,6 +253,25 @@ func (f *Factory) WithGovc() *Factory {
 	return f
 }
 
+func (f *Factory) WithTink(clusterConfigFile string) *Factory {
+	f.WithExecutableBuilder()
+	tinkerbellDatacenterConfig, err := v1alpha1.GetTinkerbellDatacenterConfig(clusterConfigFile)
```
g-gaston, line 258: nit: we could move this to inside the build step, that way we avoid calling it more than once. It also allows you to handle the error properly instead of just return nil and letting it explode

pkg/dependencies/factory.go
```diff
@@ -159,14 +159,14 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 	return f
 }
 
-func (f *Factory) WithProviderFactory(clusterConfig *v1alpha1.Cluster) *Factory {
+func (f *Factory) WithProviderFactory(clusterConfig *v1alpha1.Cluster, clusterConfigFile string) *Factory {
```
g-gaston, line 162: nit: we could simplify this and just pass the clusterConfigFile, since that one already contains the cluster definition

pkg/dependencies/factory.go
```diff
@@ -180,6 +180,7 @@ func (f *Factory) WithProviderFactory(clusterConfig *v1alpha1.Cluster) *Factory
 			VSphereGovcClient:         f.dependencies.Govc,
 			VSphereKubectlClient:      f.dependencies.Kubectl,
 			TinkerbellKubectlClient:   f.dependencies.Kubectl,
+			TinkerbellTinkClient:      f.dependencies.Tink,
```
g-gaston, line 183: No need to make any changes to this PR about this, it is more for future work
Do y'all plan on adding more `executables` and/or other dependencies to tinkerbell?

Having this factory and the provider factory together makes a bit cumbersome to inject new dependencies (as you saw, you need to define them and pass them through 3 methods). I think this is totally ok for now, but we should look into unifying both if we want to add more of them

Let me know what you think 🙂

g-gaston (commented):
Looks good!

pkg/executables/tink.go
```diff
@@ -37,3 +42,19 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
...
+	params := []string{"hardware", "list"}
+	if _, err := t.Command(ctx, params...).WithEnvVars(t.envMap).Run(); err != nil {
+		return fmt.Errorf("error getting hardware list: %v", err)
+	}
+	return nil
+}
+
+func (t *Tink) ValidateTinkerbellAccess(ctx context.Context) error {
+	if err := t.ListHardware(ctx); err != nil {
+		return fmt.Errorf("failed validating connection to tinkerbell stack")
```
mdsgabriel, line 56: should we return "err" here?

pkg/dependencies/factory.go
panktishah26: Good point! I don't think we have any plans to add more dependencies to Tinkerbell. I would definitely keep this in mind if we plan in the future. Thanks!

pkg/dependencies/factory.go
```diff
@@ -159,7 +159,9 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 	return f
 }
 
-func (f *Factory) WithProviderFactory(clusterConfig *v1alpha1.Cluster, clusterConfigFile string) *Factory {
+func (f *Factory) WithProviderFactory(clusterConfigFile string) *Factory {
+	clusterConfig, _ := v1alpha1.GetClusterConfig(clusterConfigFile)
```
g-gaston, line 163: I don't love ignoring this error
Is there something we can do about it?

pkg/dependencies/factory.go
```diff
@@ -255,16 +257,16 @@ func (f *Factory) WithGovc() *Factory {
...
-		return nil
-	}
+
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		if f.dependencies.Tink != nil {
 			return nil
 		}
+		tinkerbellDatacenterConfig, err := v1alpha1.GetTinkerbellDatacenterConfig(clusterConfigFile)
+		if err != nil {
+			return nil
```
g-gaston, line 267:
```suggestion
			return err
```

pkg/dependencies/factory.go
panktishah26: I can return error but for that I'll have to change return values for WithProviderFactory, WithProvider and all the related places. Pl correct me if we can do that without making changes at other places.

pkg/executables/tink.go
```diff
@@ -37,3 +37,11 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
...
 	return nil
 }
+
+func (t *Tink) ListHardware(ctx context.Context) error {
+	params := []string{"hardware", "list"}
+	if _, err := t.Command(ctx, params...).WithEnvVars(t.envMap).Run(); err != nil {
+		return fmt.Errorf("error getting hardware list: %v", err)
+	}
+	return nil
+}
```
abhinavmpandey08, line 47: Shouldn't this function return a list of Hardware?

pkg/executables/tink.go
```diff
@@ -37,3 +40,26 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
...
+		return nil, fmt.Errorf("error getting hardware list: %v", err)
+	}
+	var hardwareList []*hardware.Hardware
+	hardwareString := data.String()
+
+	if len(hardwareString) > 0 {
+		hardwareListData := map[string][]*hardware.Hardware{}
+
+		if err = json.Unmarshal([]byte(hardwareString), &hardwareListData); err != nil {
+			return nil, fmt.Errorf("error unmarshling json: %v", err)
```
ptrivedi, line 57: would be good to be more specific about the error and say "error unmarshalling hardware json..."

pkg/providers/tinkerbell/validator.go
```diff
@@ -6,13 +6,26 @@ import (
...
+}
 
 func (v *Validator) ValidateTinkerbellConfig(ctx context.Context, datacenterConfig *anywherev1.TinkerbellDatacenterConfig) error {
-	// TODO: add validations for tinkerbellAccess
+	if err := v.validateTinkerbellAccess(ctx); err != nil {
+		return err
+	}
+	logger.MarkPass("Connected to tinkerbell stack")
+
 	if err := v.validateTinkerbellIP(ctx, datacenterConfig.Spec.TinkerbellIP); err != nil {
```
ptrivedi, line 29: Should validate the tinkerbell IP before we try to connect to it.

pkg/providers/tinkerbell/config/template-cp.yaml
```diff
@@ -70,6 +70,8 @@ spec:
                 value: "true"
               - name: port
                 value: "6443"
+              - name: vip_interface
```
ptrivedi, line 73: Ideally we should do unrelated things like this in a separate PR. But then again, this will go away anyway with the new kube-vip version, so it doesn't matter.

panktishah26: Agreed! As this is a temp change I thought to include it in this PR. I will keep this in mind from the next time.

PR #1153, opened 2022-02-05
Doc to explain Tinkerbell config and manifests

abhinavmpandey08:
*Issue #, if available:*

*Description of changes:*
https://github.com/aws/eks-anywhere/pull/1103

### DIFF
--- a/designs/tinkerbell-provider-input.md
+++ b/designs/tinkerbell-provider-input.md
@@ -1,0 +1,358 @@
+# Required configuration for EKS Anywhere Tinkerbell provider
+
+## Introduction
+
+The purpose of this document is to explain the different input files required to create an EKS Anywhere Tinkerbell cluster and how to generate them.
+
+### Goals and objectives
+
+As an EKS Anywhere Tinkerbell provider user, I want to:
+
+* Generate the required cluster configuration file with minimal customizations
+* Generate the required hardware files using CSV files that contain hardware information
+
+### State of Scope
+
+**In Scope**
+* Explain the user configuration required for setting up an EKS Anywhere cluster using Tinkerbell
+* Provide users a way to easily generate the required cluster config and hardware manifests
+* Push the hardware directly to the Tinkerbell stack so users don't have to do it manually
+
+## Solution details
+
+### Cluster Configuration 
+
+The first file users will need is the clusterconfig file, which they can use to configure their EKS Anywhere cluster.
+
+To generate the clusterconfig for Tinkerbell provider, we use the existing `generate` command and specify `tinkerbell` as the provider
+```bash
+eksctl anywhere generate clusterconfig <cluster-name> -p tinkerbell
+```
+
+For Tinkerbell provider, we have introduced two new CRDs:
+* TinkerbellDatacenterConfig
+* TinkerbellMachineConfig
+
+**TinkerbellDatacenterConfig**
+
+TinkerbellDatacenterConfig describes the configuration details for Tinkerbell stack.
+
+In this config, users must specify the following fields:
+* spec.tinkerbellIP: defines the IP of the machine running the Tinkerbell stack.
+* spec.tinkerbellCertURL: defines the endpoint where the Tinkerbell stack hosts its certificate. This is used to communicate with the stack.
+* spec.tinkerbellGRPCAuth: defines the GRPC endpoint used by Tinkerbell stack. This is also used to communicate with the stack.
+* spec.tinkerbellPBnJGRPCAuth: defines the GRPC endpoint used by PNBJ for hardware remote management. This is used to change the power state and the boot order of the servers.
+
+Here's an example `TinkerbellDatacenterConfig`
+
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: TinkerbellDatacenterConfig
+metadata:
+  name: eks-a-tinkerbell-cluster
+spec:
+  tinkerbellIP: "192.168.0.20"
+  tinkerbellCertURL: "http://192.168.0.20:42114/cert"
+  tinkerbellGRPCAuth: "192.168.0.20:42113"
+  tinkerbellPBnJGRPCAuth: "192.168.0.20:50051"
+```
+
+**TinkerbellMachineConfig**
+
+TinkerbellMachineConfig describes the configuration for machines used for creating EKS Anywhere cluster on Tinkerbell.
+
+This config supports the following fields
+* spec.osFamily (required): defines the operating system to be installed on this machine. We currently only support ubuntu for this but this list is expected to grow in the future.
+* spec.users (required): defines the users to be created on this machine along with the ssh keys.
+  * spec.users[].name (required): defines the name of the user.
+  * spec.users[].sshAuthorizedKeys (required): defines the SSH authorized keys to be installed on this machine to be used for SSH'ing into it.
+* spec.templateOverride (optional): defines the [Tinkerbell template](https://docs.tinkerbell.org/templates/) to be used to provision the OS on the machine. This field is passed as a YAML encoded as string.
+
+Here is an example `TinkerbellMachineConfig`
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: TinkerbellMachineConfig
+metadata:
+  name: eks-a-tinkerbell-cluster-cp
+spec:
+  osFamily: ubuntu
+  users:
+  - name: ec2-user
+    sshAuthorizedKeys:
+    - ssh-rsa AAAA...
+  templateOverride: |
+    <>
+```
+
+Here is the default `templateOverride` that is generated in the clusterconfig
+```yaml
+version: "0.1"
+name: eks-a-tinkerbell-cluster-cp-test
+global_timeout: 6000
+tasks:
+  - name: "eks-a-tinkerbell-cluster-cp-ndrl8"
+    worker: "{{.device_1}}"
+    volumes:
+      - /dev:/dev
+      - /dev/console:/dev/console
+      - /lib/firmware:/lib/firmware:ro
+    actions:
+      - name: "stream-image"
+        image: image2disk:v1.0.0
+        timeout: 360
+        environment:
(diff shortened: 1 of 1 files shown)


designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,358 @@
...
+Here is an example hardware CSV file with the required information for two hardwares:
+```csv
+ip_address,gateway,nameservers,netmask,mac,hostname,vendor,bmc_ip,bmc_username,bmc_password
+192.168.1.10,192.168.0.1,8.8.8.8|8.8.4.4,255.255.255.0,00:00:00:00:00:00,worker1,supermicro,192.168.0.10,admin,Admin
+192.168.1.11,192.168.0.1,8.8.8.8|8.8.4.4,255.255.255.0,00:00:00:00:00:01,worker2,supermicro,192.168.0.11,admin,Admin
+```
+
+Once the CSV file is set up, users can use the `generate hardware` command to generate the hardware manifests
+
+(Recommended) Command to generate the hardware manifests and push it to the Tinkerbell stack
```
g-gaston, line 333: not blocking: tbh I don't really like this. I think it breaks the pattern of the `generate` command, where it doesn't have side effects, it just writes files to local disk and lets the user decide what to do with those files.

My vote would be to make "not pushing" the default. Or even better, move the "push" functionality to a separate command.

g-gaston (commented):
I really like this, thanks a lot!

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,358 @@
...
+### Goals and objectives
+
+As an EKS Anywhere Tinkerbell provider user, I want to:
+
+* Generate the required cluster configuration file with minimal customizations
+* Generate the required hardware files using CSV files that contain hardware information
+
+### State of Scope
+
+**In Scope**
```
vivek-koppuru, line 16: Are we separating the scope into two parts? I recall us talking about starting off with these files as input and then moving towards only specifying the id's with all the hardware information embedded in the tinkerbell side.

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,358 @@
...
+        environment:
+          BLOCK_DEVICE: /dev/sda1
+          FS_TYPE: ext4
+```
+
+### Hardware manifests
+
+Tinkerbell provider also requires hardware manifest files along with the clusterconfig to provide configuration details about the hardware. There are two hardware manifest files involved: a hardware JSON and a hardware YAML. 
+
+>**_Note_**: There is some ongoing work here and this may change post-beta
```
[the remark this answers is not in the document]
vivek-koppuru: It might help to have the scope defined above for this. Let's call it mvp scope and final or something along those lines :)

designs/tinkerbell-provider-input.md
vivek-koppuru: I agree with `generate` and `push` being separate commands, or with a flag. Maybe we have a `setup` that does both

abhinavmpandey08: The reason we wanted to push hardware by default as part of generate is mainly that it is tedious to push hardware manually to Tinkerbell. We also wanted to keep it part of the same command so users wouldn't have to run two different commands that can be part of just one. 

This is also just a beta thing and most likely the push part will go away at GA (still have to finalize a few details for GA)
If you guys prefer, I could move it to a separate `setup` command like Vivek suggested. Let me know what you guys think.

designs/tinkerbell-provider-input.md
abhinavmpandey08: Those details aren't finalized yet, so I don't know what this will to look like post-beta. I would prefer to keep this doc just for beta and maybe update/add a new one once we have full knowledge of what we want for GA.

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,358 @@
...
+For Tinkerbell provider, we have introduced two new CRDs:
+* TinkerbellDatacenterConfig
+* TinkerbellMachineConfig
+
+**TinkerbellDatacenterConfig**
+
+TinkerbellDatacenterConfig describes the configuration details for Tinkerbell stack.
+
+In this config, users must specify the following fields:
+* spec.tinkerbellIP: defines the IP of the machine running the Tinkerbell stack.
```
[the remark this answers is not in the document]
abhinavmpandey08: You mean like change `tinkerbellIP` to just `IP` and similar for other fields?

designs/tinkerbell-provider-input.md
g-gaston: If it's not a lot of work, I like the separate command idea

abhinavmpandey08: I have updated the doc to use `setup` instead of `generate`

designs/tinkerbell-provider-input.md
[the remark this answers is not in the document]
abhinavmpandey08: Added a small section about future work at the end. Let me know if that will suffice

designs/tinkerbell-provider-input.md
abhinavmpandey08: Made the scope specific to just beta as post-beta details are still tentative.

designs/tinkerbell-provider-input.md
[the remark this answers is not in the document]
vivek-koppuru: Yes exactly

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,366 @@
...
+  osFamily: ubuntu
+  users:
+  - name: ec2-user
+    sshAuthorizedKeys:
+    - ssh-rsa AAAA...
+  templateOverride: |
+    <>
+```
+
+Here is the default `templateOverride` that is generated in the clusterconfig
```
vivek-koppuru, line 89: So the `templateOverride` gets populated in `generate clusterconfig` or do these values get filled in from datacenter and machine configs?

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,366 @@
...
+    volumes:
+      - /dev:/dev
+      - /dev/console:/dev/console
+      - /lib/firmware:/lib/firmware:ro
+    actions:
+      - name: "stream-image"
+        image: image2disk:v1.0.0
+        timeout: 360
+        environment:
+          IMG_URL: "https://<some-s3-endpoint>/ubuntu-2004-kube-v1.21.5.gz"
```
vivek-koppuru, line 106: Doesn't have to be s3 right?

designs/tinkerbell-provider-input.md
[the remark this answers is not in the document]
vivek-koppuru:
```suggestion
>**_Note_**: There is some ongoing work here and this may change post-beta, more details below under "Future Changes"
```

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,366 @@
...
+
+```bash
+eksctl anywhere create cluster -f <clusterconfig> -w <hardware YAML>
+```
+
+## Future changes
+
+Most of the design details described above are only valid for beta release of Tinkerbell provider as for beta, we are using the default installation of Tinkerbell. This default installation runs Tinkerbell components as microservices on docker containers. We have plans to eliminate Tinkerbell's dependency on docker and instead run it on Kubernetes as controllers. You can read the proposal for this [here](https://github.com/tinkerbell/proposals/tree/main/proposals/0026).
+
+Since this is a work-in-progress, we still have not finalized how our design will look like when we leverage Kubernetes backend for Tinkerbell.
```
vivek-koppuru, line 366: Can we talk about what we discussed in a meeting earlier about how these commands are going to change? For example changes that can be made to upstream to decrease the number of files we need to maintain and changing our abstraction of these objects?

designs/tinkerbell-provider-input.md
```diff
@@ -0,0 +1,366 @@
...
+### Goals and objectives
+
+As an EKS Anywhere Tinkerbell provider user, I want to:
+
+* Generate the required cluster configuration file with minimal customizations
+* Generate the required hardware files using CSV files that contain hardware information
+
+### Statement of Scope
+
+The scope of this document is mainly for beta release of Tinkerbell provider. Some of the details in the design will change once Tinkerbell is [migrated completely to Kubernetes](https://github.com/tinkerbell/proposals/tree/main/proposals/0026).
```
g-gaston, line 16: This is good, but could we add a few things:

* Could we specify what are the parts that might change?
* Could we explain why migrating to kubernetes might change this design?
* I think we mentioned previously that the GA design will depend on ongoing conversations/designs. The linked proposal has already been approved. Does that mean we can already design the GA solution even if we can't implement it? Or are there actual unknowns pending on separate discussions?

g-gaston (commented):
Only one suggestion, rest looks good

designs/tinkerbell-provider-input.md
abhinavmpandey08: right can be any endpoint. I said s3 because that's where we host our images

designs/tinkerbell-provider-input.md
abhinavmpandey08: It gets generated when you run `generate clusterconfig`

designs/tinkerbell-provider-input.md
abhinavmpandey08: The proposal has been approved but we don't know if all of that will be implemented before GA. I will add a few more details around what we are expecting to change but that may also change later

designs/tinkerbell-provider-input.md
abhinavmpandey08: Added a few more specific things that we expect will change

designs/tinkerbell-provider-input.md
abhinavmpandey08: Updated the doc. Let me know if i missed something

g-gaston (approved):
I like this a lot, thank @abhinavmpandey08!

PR #1158, opened 2022-02-05
Deleting a cluster with GitOpsConfig enabled fails unless branch is set explicitly

kingdonb:
*Issue #, if available:* There is no issue filed for this report.

I just tried out eks-anywhere on ARM64 architecture with a Docker configuration and was able to make everything work, except for deleting the cluster cleanly at the end of the experiment.

I figured out that when I set a branch explicitly in the GitOpsConfig it succeeded. This change might help; the git clone is smart enough to pick the default branch, but my working hypothesis anyway is that the CR will not have that data and when it comes time to push a change representing the cluster delete, that's where it fails.

I have since lost track of where I found the code that emitted the branch error, but it was clear enough from reading it what happened. We just didn't have a string "main" and so whatever routine that expected the string branch name failed.

I guess it works up til that point because git repos have a default branch, and you can already clone them without naming it.

*Description of changes:*

I added a kubebuilder annotation that really makes the default mentioned in the adjacent comment official. I also ran `make generate-manifests` so it appears I did that correctly. I'm new at kubebuilder so, I wanted to mention that I don't know if this will solve the issue, or how to test it myself without running it through CI here.

---
(By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.)

### DIFF
--- a/config/crd/bases/anywhere.eks.amazonaws.com_gitopsconfigs.yaml
+++ b/config/crd/bases/anywhere.eks.amazonaws.com_gitopsconfigs.yaml
@@ -44,6 +44,7 @@ spec:
                       Git repo.
                     properties:
                       branch:
+                        default: main
                         description: Git branch. Defaults to main.
                         type: string
                       clusterConfigPath:
--- a/pkg/api/v1alpha1/gitopsconfig_types.go
+++ b/pkg/api/v1alpha1/gitopsconfig_types.go
@@ -26,6 +26,7 @@ type Github struct {
 	FluxSystemNamespace string `json:"fluxSystemNamespace,omitempty"`
 
 	// Git branch. Defaults to main.
+	// +kubebuilder:default:="main"
 	Branch string `json:"branch,omitempty"`
 
 	// ClusterConfigPath relative to the repository root, when specified the cluster sync will be scoped to this path.


kingdonb:
Here's the line of code that bombed out, IIRC:

https://github.com/aws/eks-anywhere/blob/eed94fa13c0ca580b683858a071cb029fbe50c93/pkg/addonmanager/addonclients/fluxaddonclient.go#L422

jaxesn:
Thanks for the contribution! Curious were you working  on a Linux arm machine or a m1 Mac?

TerryHowe (commented):
/ok-to-test

kingdonb:
It was a branch not found error. No crash dump.

g-gaston (changes requested):
Thanks a lot for bringing this up @kingdonb!

Unfortunately, I don't think this change would fix the problem. This sets the default in the open api spec of the CRD, which is only ran at the kube api server level (if I'm not mistaken).

However, most of the code of our CLI reads the cluster configuration exclusively from the provided file in local disk. This makes the open api defaults, although correct, ineffective.

I have checked and we have some logic to set defaults for gitops locally but unfortunately, it runs for create and upgrade but doesn't run in the delete flow, hence this issue. To be fair, that code is not the greatest and is not very obvious how/when it needs to be called.

I'll create an issue to clean it up. In the meantime, what do you think about adding it to the gitops delete flow? If you could add a `clusterSpec.SetDefaultGitOps()` [here](https://github.com/aws/eks-anywhere/blob/main/pkg/addonmanager/addonclients/fluxaddonclient.go#L303), that should do the trick. It's a bit hacky but it should work until we improve this code.

Also, even if it doesn't have any effects for now, I would like to keep your kubebuilder changes. Would you mind running `make release-manifests`? That should update one more yaml file.

kingdonb:
Thanks for the feedback, I added more commits. If that looks ok, and you think it will address the issue, welcome to merge!

jiayiwang7 (approved):

g-gaston (approved):
Thanks a lot!
/lgtm

kingdonb:
Awesome, I've re-tested after the release and this is working.

I can now delete clusters that were provisioned with `GitOpsConfig` even when the disk YAMLs miss the `branch` field.

```
$ eksctl-anywhere delete cluster dev-cluster-4 -f dev-cluster-4.yaml --kubeconfig dev-mgmt-cluster/dev-mgmt-cluster-eks-a-cluster.kubeconfig
Checking Github Access Token permissions
✅ Github personal access token has the required repo permissions
Performing provider setup and validations
Deleting workload cluster
Clean up Git Repo
Finalized commit and committed to local repository	{"hash": "2c22d4effd34ec96930b86ea5e05b7462400f19a"}
Bootstrap cluster information missing - skipping delete kind cluster
🎉 Cluster deleted!
```

PR #1161, opened 2022-02-07
Refactor artifact table generation for release assets

abhay-krishna:
Refactor artifact table generation logic for EKS-A releases into a single file. This will reduce code duplication and make it easier to add more assets.

/hold

### Additions

- [x] Artifact table logic for
  - [x] Images
  - [x] Manifests
  - [x] Archives
- [x] Documentation ([Design doc](https://github.com/aws/eks-anywhere/pull/1173), [Guide to add new project assets](https://github.com/aws/eks-anywhere/pull/1161/files#diff-d8f684dc36925e0f03f40b48ae871c6ef387eb8d5bef96da5af4d4c56f38c4df))

### Modifications
- [x] Update spec generation methods to use new asset methods
- [x] Release branch integrations (`1-20`, `1-21`, ...)
- [x] Move remaining files existing under `pkg` to sub-packages based on functionality

### Deletions
- [x] Remove all `assets_X.go` files
- [x] Cleanup methods

### Testing
- [x] Dev-release dry-run succeeds
- [x] Source S3 and ECR URIs are preserved
- [x] Release S3 and ECR URIs are preserved
- [x] Testing generated bundle against golden file (#1743)


release/pkg/utils/utils.go
```diff
@@ -16,12 +16,14 @@ package utils
```
abhay-krishna, line 16: **Changes:** moved utility methods from other packages here

release/pkg/version/versions.go
```diff
@@ -1,4 +1,4 @@
-package pkg
```
abhay-krishna, line 1: **Changes:** moved versioner to its own package and added other import changes.

release/pkg/filereader/file_reader.go
```diff
@@ -342,6 +334,20 @@ func getEksdRelease(eksdReleaseURL string) (*eksdv1alpha1.Release, error) {
 	return eksd, nil
 }
 
+// Retrieve values from https://github.com/aws/eks-anywhere-build-tooling/blob/main/EKSD_LATEST_RELEASES
+func GetEksdReleaseValues(release interface{}) (string, bool) {
```
g-gaston, line 338: I think it would be easier to just have struct to unmarshall this instead of using `interface{}`

release/pkg/assets/assets.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+
+			artifacts = append(artifacts, releasetypes.Artifact{Archive: archiveArtifact})
+		}
+	}
+
+	return artifacts, nil
+}
+
+func GetBundleReleaseAssets(supportedK8sVersions []string, eksDReleaseMap map[string]interface{}, rc *releasetypes.ReleaseConfig) (map[string][]releasetypes.Artifact, error) {
```
g-gaston, line 298: The fact that we are using a map of empty interface here makes me think we might need to rethink this a bit

release/pkg/assets/assets.go
```diff
@@ -0,0 +1,329 @@
...
+	projectName := ac.ProjectName
+	projectPath := ac.ProjectPath
+	sourcedFromBranch := rc.BuildRepoBranchName
+	gitTagPath := projectPath
+	if ac.HasSeparateTagPerReleaseBranch {
+		gitTagPath = filepath.Join(projectPath, eksDReleaseChannel)
+	}
+	// Get git tag for project if exists
+	if !ac.NoGitTag {
+		if projectName == "eks-anywhere-cluster-controller" {
```
g-gaston, line 33: This looks like an anti-pattern to me
We are trying to dry out common logic, moving from individual implementations to a common shared implementation, from concrete to abstract. But this is including the concrete (a particular project name) in the abstract (the shared "oracle"

release/pkg/assets/assets.go
```diff
@@ -0,0 +1,329 @@
...
+			repoName, assetName := image.RepoName, image.RepoName
+			if image.AssetName != "" {
+				assetName = image.AssetName
+			}
+
+			if ac.ImageRepoPrefix != "" {
+				repoName = fmt.Sprintf("%s/%s", ac.ImageRepoPrefix, repoName)
+			}
+			sourceRepoName, releaseRepoName := repoName, repoName
+			if strings.HasPrefix(repoName, "eks-anywhere") {
```
g-gaston, line 66: same here about the concrete vs abstract
I think we should able to rethink this a little bit and encode it as some kind of option

release/pkg/assets/assets.go
```diff
@@ -0,0 +1,329 @@
...
+
+			imageTagOverrides = append(imageTagOverrides, releasetypes.ImageTagOverride{
+				Repository: sourceRepoName,
+				ReleaseUri: imageArtifact.ReleaseImageURI,
+			})
+		}
+	}
+
+	// Add manifests to artifacts list
+	if len(ac.Manifests) > 0 {
```
g-gaston, line 131: nit: we can skip this check, the for loop does this for us

release/pkg/images/images.go
```diff
@@ -83,3 +91,276 @@ func CopyToDestination(sourceAuthConfig, releaseAuthConfig *docker.AuthConfigura
 
 	return nil
 }
+
+func GetSourceImageURI(r *releasetypes.ReleaseConfig, name, repoName string, tagOptions map[string]string) (string, string, error) {
+	var sourceImageUri string
+	sourcedFromBranch := r.BuildRepoBranchName
+	if r.DevRelease || r.ReleaseEnvironment == "development" {
+		latestTag := utils.GetLatestUploadDestination(r.BuildRepoBranchName)
+		if strings.Contains(name, "bottlerocket-bootstrap") {
```
g-gaston, line 100: Similar thing here and in `GetReleaseImageURI`
I think we should avoid doing this and codify this information in the asset/image declaration

g-gaston (commented):
Most of the comments go towards a similar issue: adding "exceptions" in the abstracted code to deal with the differences between concrete implementations

I still think the "declarative" approach and the abstractions used here are the right ones. But we need to take a mini step back and reconsider how we want to implement them in go code

A good starting point could be to document (no need for a design doc or review at all, just for designing the code) what are all the projects and the commonalities and differences between them. That would probably help determine what should be the common logic and what needs to be independent per project. And based on that we can come up with the high level code design

release/pkg/filereader/file_reader.go
abhay-krishna: I agree, but with the current format of the [releases file](https://github.com/aws/eks-anywhere-build-tooling/blob/main/EKSD_LATEST_RELEASES), it would be hard to make it a struct since it uses release numbers as keys. I will get back to this as a followup after we change that file's format.
I prefer something like
```
releases:
- branch: 1-18
  kubeVersion: v1.18.20
  number: 14
- branch: 1-19
  kubeVersion: v1.19.15
  number: 13
  ...
- branch: 1-22
  kubeVersion: v1.22.4
  number: 2
  dev: true
latest: 1-22
```

release/pkg/assets/assets.go
abhay-krishna: agreed, we will be using a struct like
```go
type EksDLatestRelease struct {
    Dev         boolean `json:"dev"`
    Number      int     `json:"number`
    KubeVersion string  `json:"kubeVersion`
}
but the change to the releases file would need to happen first, as in the above comment.

File: release/pkg/assets/assets.go [hunk 9]
```diff
@@ -0,0 +1,323 @@
...
+	assetconfig "github.com/aws/eks-anywhere/release/pkg/assets/config"
+	assettypes "github.com/aws/eks-anywhere/release/pkg/assets/types"
+	"github.com/aws/eks-anywhere/release/pkg/filereader"
+	"github.com/aws/eks-anywhere/release/pkg/git"
+	"github.com/aws/eks-anywhere/release/pkg/images"
+	releasetypes "github.com/aws/eks-anywhere/release/pkg/types"
+	"github.com/aws/eks-anywhere/release/pkg/utils"
+)
+
+func getAssetsFromConfig(ac *assettypes.AssetConfig, rc *releasetypes.ReleaseConfig, eksDReleaseChannel, eksDReleaseNumber, kubeVersion string) ([]releasetypes.Artifact, error) {
```
chrisdoherty4, line 19: This is a 270 line function with 3 major for loops. At a glance its doing lots of different things, do you think we could break this down _and_ write unit tests.

release/pkg/aws/ecr/ecr.go
```diff
@@ -9,7 +9,7 @@ import (
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/pkg/errors"
 
-	"github.com/aws/eks-anywhere/release/pkg/utils"
+	"github.com/aws/eks-anywhere/release/pkg/aws/utils"
```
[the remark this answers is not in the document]
chrisdoherty4: Is this a smell? Rearranging to avoid a cyclic dep is generally a sign of incorrectly modelled dependencies. What caused the cycle?

release/pkg/filereader/file_reader.go
```diff
@@ -76,15 +83,15 @@ func readShaFile(filename string) (string, error) {
 	return "", errors.Errorf("Error parsing shasum file %s", filename)
 }
 
-func readFile(filename string) (string, error) {
+func ReadFile(filename string) (string, error) {
```
chrisdoherty4, line 86: Exporting this symbol is offers it up as an API to be used from the package. It does more than read a file, it trims space. Unknowing maintainers could use it and have undesirable side effects. I suggest its refactored to only read the file and you wrap an unexported function around the exported `ReadFile` to prevent the undesirable behavior.

release/pkg/filereader/file_reader.go
```diff
@@ -356,3 +362,18 @@ func ReadHttpFile(uri string) ([]byte, error) {
 
 	return data, nil
 }
+
+func ReadGitTag(projectPath, gitRootPath, branch string) (string, error) {
```
chrisdoherty4, line 366: The function name isn't representative of the behavior. Reading a git tag is a reasonably trivial operation, checking out an entire repository is a very different thing entirely: `CheckoutAndReadGitTag()` would be better and I'd suggest breaking it into 2 functions with this being a meta function.

release/pkg/filereader/file_reader.go
```diff
@@ -356,3 +362,18 @@ func ReadHttpFile(uri string) ([]byte, error) {
...
+	_, err := git.CheckoutRepo(gitRootPath, branch)
+	if err != nil {
+		return "", errors.Cause(err)
+	}
+
+	tagFile := filepath.Join(gitRootPath, projectPath, "GIT_TAG")
+	gitTag, err := ReadFile(tagFile)
+	if err != nil {
+		return "", errors.Cause(err)
```
chrisdoherty4, line 375: Why strip context from the error?

release/pkg/utils/utils.go
```diff
@@ -83,3 +77,31 @@ func GetManifestFilepaths(devRelease bool, bundleNumber int, kind, branch string
 	}
 	return manifestFilepath
 }
+
+func SortArtifactsMap(m map[string][]releasetypes.Artifact) []string {
```
chrisdoherty4, line 81: I don't see any relationship between the new and existing functions. As with many `common`, `utils` and `base` packages that represent arbitrary logic, this looks like its set to become Sakaar. 

Suggest we don't add to this and instead create a nested packaged with a concrete purpose. E.g. `/utils/git` and `/utils/uri`.

chrisdoherty4:
I think we should really have some unit tests around the code to model expected behavior. How is this being tested at present?

release/pkg/filereader/file_reader.go
abhay-krishna: Resolved by #1246 and aws/eks-anywhere-build-tooling#450.

release/pkg/filereader/file_reader.go
abhay-krishna: Done

abhay-krishna:
> I think we should really have some unit tests around the code to model expected behavior. How is this being tested at present?

I will look into adding more tests. The purpose of the release module is to migrate artifacts from build (dev) accounts to staging/prod destinations (with the appropriate tags/S3 paths) and also generate a bundle manifest that is an exhaustive list of components (container images, OS images, manifests, tarballs, etc) required to create an EKS-A cluster. This bundle will be parsed during cluster creation and corresponding components (like CAPI, cert-manager, kind) will be pulled and used in various cluster creation workflows.

Currently changes to the release tooling are tested by the `eks-anywhere-release-tooling-presubmit`, which performs a dry-run of the release process and also prints out a bundle manifest at the end, with close-to-reality fake URIs, with the only difference being the S3 bucket names and ECR endpoints.

release/pkg/utils/utils.go
abhay-krishna: I had these as miscellaneous functions which served as nothing more than utilities to other functions. I can do the nested approach if that would be more idiomatic.

release/pkg/images/images.go
abhay-krishna: I will work on making this as a configurable field, and have a default to just use git tag if not provided.

release/pkg/assets/assets.go
abhay-krishna: Will do

release/pkg/utils/utils.go
chrisdoherty4: As a rule of thumb we should avoid `util` packages that have undefined purpose. 

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_avoid_package_names_like_base_common_or_util
https://github.com/go-kit/kit/tree/master/util/conn

release/pkg/filereader/file_reader.go
abhay-krishna: Fixed

release/pkg/utils/utils.go
abhay-krishna: Split into nested packages, thanks for the references!

release/go.mod
```diff
@@ -15,6 +13,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.7.1
+	gopkg.in/yaml.v2 v2.4.0
```
g-gaston, line 16: what was the problem with `sigs.k8s.io/yaml`?

release/pkg/assets/archives.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/pkg/errors"
+
+	assettypes "github.com/aws/eks-anywhere/release/pkg/assets/types"
+	"github.com/aws/eks-anywhere/release/pkg/filereader"
+	releasetypes "github.com/aws/eks-anywhere/release/pkg/types"
+	artifactutils "github.com/aws/eks-anywhere/release/pkg/util/artifacts"
+	sliceutils "github.com/aws/eks-anywhere/release/pkg/util/slices"
+)
+
+func getArchiveAssets(rc *releasetypes.ReleaseConfig, archive *assettypes.Archive, projectPath, gitTag, eksDReleaseChannel, eksDReleaseNumber, kubeVersion string) (*releasetypes.ArchiveArtifact, error) {
```
g-gaston, line 16: How difficult would it be to add tests to this method?

release/pkg/assets/archives.go
```diff
@@ -0,0 +1,142 @@
...
+	os := "linux"
+	arch := "amd64"
+	var sourceS3Key string
+	var sourceS3Prefix string
+	var releaseS3Path string
+	var releaseName string
+	sourcedFromBranch := rc.BuildRepoBranchName
+	latestPath := artifactutils.GetLatestUploadDestination(sourcedFromBranch)
+
+	if archive.Name == "eks-distro" {
```
g-gaston, line 26: I would vote to try to avoid this kind of thing
Instead of using branching, could we codify this in the archive definition?
Like for example, using something like `ArchiveS3KeyGenerator`, `ArchiveS3PathGenerator`... (they can be closures or interfaces)
Those could just be part of each `Archive` declaration and then we can do `archive.getS3Path()`

The same thing applies to all the if/else for `rc.DevRelease` and `rc.ReleaseEnvironment`, could we avoid them with dependency injection?

release/pkg/assets/assets.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+		} else {
+			gitTag, err = filereader.ReadGitTag(gitTagPath, rc.BuildRepoSource, rc.BuildRepoBranchName)
+			if err != nil {
+				return nil, errors.Cause(err)
+			}
+		}
+	} else {
+		gitTag = "non-existent"
+	}
```
g-gaston, line 51: This is what I was suggetsing at the beggining to try to implement options through closures/interfaces instead of boolean. The complexity of this block of code is significant, up to 4 nested ifs.

And there is very little we can do about it with boolean options: the more options we add, the more permutations.

release/go.mod
abhay-krishna: Removed this in favor of sigs yaml.

release/pkg/assets/archives.go
abhay-krishna: Done!

release/pkg/assets/assets.go
abhay-krishna: Fixed this.

g-gaston (approved):

Issue #410, opened 2021-10-07
Embed EKS Distro manifests in eksa-controller-manager


abhinavmpandey08:
**What would you like to be added**:
Embed eks-distro manifest in eksa-controller-manager so it doesn't try it to pull it dynamically

**Why is this needed**:
Currently `eksa-controller-manager` pulls eks-distro manifest which requires access to the internet. It would be nice to have to embedded in the controller so it won't need to pull it during runtime. This will eventually be needed for air-gapped support

-------------------
Design doc: #1162 

- [x] Update EKS-D bundle with new fields #1198 
- [x] Add `eksdReleaseRef` to cluster status #1217 
- [x] Add tasks during create & upgrade to install EKS-D components #1391

PR #1162, opened 2022-02-07
Design doc for embedding EKSD resources in EKSA cluster

taneyland:
*Issue #, if available:*
#410 

*Description of changes:*
Design doc for embedding EKSD resources in EKSA cluster

### DIFF
--- a/designs/embed_eksd_manifest.md
+++ b/designs/embed_eksd_manifest.md
@@ -25,10 +25,6 @@ In the upgrade workflow, we will also add a step to apply the EKSD resources as 
 We currently don’t pull in the EKSD release CRD in the bundle, but we need to apply this CRD before applying the release manifest.
 I propose introducing a new field `components` in the EKSD bundle to refer to the file in S3 that stores this CRD.
 
-While most of our bundles have a `version` field, the EKS-D bundle does not have this field.
-Instead, the version is typically decided by the `releaseChannel` and the `releaseNumber`.
-We already have the `channel` field in the EKS-D bundle, so I propose adding the `number` field to this bundle to have an effective way to check the EKS-D version during upgrades.
-
 Example bundles manifest with proposed changes:
 
 ```
@@ -49,12 +45,6 @@ spec:
       kindNode:
         arch:
           ...
-      kubeVersion: v1.21.5
-      manifestUrl: https://distro.eks.amazonaws.com/kubernetes-1-21/kubernetes-1-21-eks-8.yaml
-      name: kubernetes-1-21-eks-8
-      number: 8
-      ova:
-         ...
 ```
 
 From here, we can use `kubectl apply` to apply the content from the EKSD release CRD & manifest to the cluster.
@@ -62,7 +52,7 @@ From here, we can use `kubectl apply` to apply the content from the EKSD release
 #### Cluster status changes
 
 We will need a way to refer to the EKSD release object on the cluster when calling it from the controller.
-One way to achieve this is by adding a field `EkdsReleaseRef` to the cluster status.
+One way to achieve this is by adding a field `eksdReleaseRef` to the cluster status.
 ```
 apiVersion: anywhere.eks.amazonaws.com/v1alpha1
 kind: Cluster
@@ -73,8 +63,10 @@ spec:
     ...
 status:
   eksdReleaseRef:
+    apiGroup: distro.eks.amazonaws.com/v1alpha1
     kind: Release
-    name: "name of resource" 
+    name: "name of resource"
+    namespace: "namespace for resource" (default: eksa-system)
 ```
 
 From here, we can call `kubectl get release` using the `eksdReleaseRef.Name` to access the EKSD release manifest from the controller.
@@ -83,6 +75,6 @@ From here, we can call `kubectl get release` using the `eksdReleaseRef.Name` to 
 
 Now that the EKSD components are embedded in the cluster, we have to ensure that those components are supported during upgrade.
 We will update the upgrader in cluster manager to not only check for updates to EKSA, but it will also check if updates are needed for the EKSD objects on the cluster.
-We will have a method `EksdChangeDiff` to check the EKSD version in the bundle to decide if we need to update the EKSD manifest objects that are applied to the cluster.
+We will have a method `EksdChangeDiff` to check the EKSD version (based on the `name` in the bundle) to decide if we need to update the EKSD manifest objects that are applied to the cluster.
 
 


designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,59 @@
...
+One way to achieve this is by adding a field to the cluster status.
+We will introduce a field named `EksdRelease` of type `*v1alpha1.Release`.
+This is how we will attain that direct access to the release object when calling it from the controller.
+
+
+### Upgrading EKSD components
+
+Now that the EKSD components are embedded in the cluster, we have to ensure that those components are supported during upgrade.
+We will update the upgrader in cluster manager to not only check for updates to EKSA, but it will also check for updates to EKSD.
+If there is a newer version, we will upgrade those components.
```
jaxesn, line 58: What do we mean here by `components`?  Dont we already upgrade eks-d components?  Is this referring to updating the CRD/CR for the to the new EKS-D bundle if the user was to run `anywhere upgrade`?

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,59 @@
+# Embed EKS-D resources in EKS-A cluster
+
+## Introduction
+
+**Problem:** Currently, the EKS Anywhere controller pulls the EKS Distro manifest from the release URL provided in the versions bundle, which requires access to the internet.
+It would be ideal to have this manifest and its CRDs embedded in the cluster to allow the controller to operate without the requirement to make that external call.
+This will eventually be needed for air-gapped support.
```
TerryHowe, line 7: Can't we just dump all this stuff in the private registry?

designs/embed_eksd_manifest.md
taneyland: Yes. I guess another way to put it is if there's a new version in the EKSD bundle, we will update the EKSD object on the cluster to that new version by calling `kubectl apply` with the updated release & CRD URLs.

designs/embed_eksd_manifest.md
vivek-koppuru: We could but I think this is more of a smaller step to fix this specific problem of downloading on the fly from within the controller, which was something that we wanted to fix from a previous issue

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Overview of Solution
+
+The EKS Distro manifest and the release CRD will be applied directly to the cluster.
+This will allow the controller to grab the EKSD release resources from within the cluster instead of reading in the EKSD manifest from the internet when fetching the applied spec.
+
+### Solution Details
+
+With this feature, the EKSD manifest & its CRD will be applied directly to the cluster.
+We will do this similarly to how we apply the EKSA components to the cluster, introducing a new task to the create workflow called `InstallEksdComponentsTask`.
```
vivek-koppuru, line 24: This would need to be kept in upgrade workflow as well, especially in upgrading K8s

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,59 @@
...
+We currently don’t pull in the EKSD release CRD in the bundle, so I propose introducing a new field `crdUrl` in the EKSD Bundle to refer to the file in S3 that stores this CRD.
```
vivek-koppuru, line 25: Actually in practice, this is something that is stored as part of the actual manifest URL that we have linked in the eksD. I would be curious to see what the effort required is to just edit our current manifest files to include the CRDs themselves instead of maintaining a separate URL

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,59 @@
...
+        fcddf59a516102e20f75b6c672e3aa74cf2da877
+      kindNode:
+        arch:
+```
+
+From here, we can use `kubectl apply` to apply the content from the EKSD release URL and the CRD URL to the cluster.
+
+We will need a way to refer to the EKSD release object on the cluster.
+One way to achieve this is by adding a field to the cluster status.
+We will introduce a field named `EksdRelease` of type `*v1alpha1.Release`.
```
vivek-koppuru, line 50: Can you show an example of this? Also we have talked about this in the past in some of our design docs, but what about something like `eksdReleaseRef`?

```
eksdReleaseRef:
    name: "name of resource" 
    url: "url we downloaded from"
```

designs/embed_eksd_manifest.md
taneyland: I plan to have an `EksdChangeDiff` method in the upgrade flow similar to how we have for EKS-A components. I could check both the k8s version & ESK-D version in that method when deciding whether or not we need to install an updated version of the EKS-D manifest.

designs/embed_eksd_manifest.md
taneyland: Yes I like this. Will update the doc

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,72 @@
...
+kind: Cluster
+metadata:
+  name: test_cluster
+spec:
+  controlPlaneConfiguration:
+    ...
+status:
+  eksdReleaseRef:
+    name: "name of resource" 
+    url: "url we downloaded from"
```
g-gaston, line 61: I would vote to have the `apiGroup` and `kind` here since the eksd release doesn't belong to the eks-a group.
I would even have the namespace, but if we want to assume it will always be eksa-system, that's ok.

What is the url for? I would vote to remove it of there is no an specific need yet.

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+From here, we can call `kubectl get release` using the `eksdReleaseRef.Name` to access the EKSD release manifest from the controller.
+
+### Upgrading EKSD components
+
+Now that the EKSD components are embedded in the cluster, we have to ensure that those components are supported during upgrade.
+We will update the upgrader in cluster manager to not only check for updates to EKSA, but it will also check if updates are needed for the EKSD objects on the cluster.
+We will have a method `EksdChangeDiff` to check both the EKSD version in the bundle and the Kubernetes version in the cluster config to decide if we need to update the EKSD manifest objects that are applied to the cluster.
```
g-gaston, line 70: nit: it should be enough to check the eks-d version, since a different kubernetes minor version guarantees a new eks-d release.

g-gaston (commented):
I like this!

designs/embed_eksd_manifest.md
taneyland: I was thinking to add this in the case where the user, for example, creates a cluster with kubeVersion 1.20 & upgrades the cluster to use 1.21

designs/embed_eksd_manifest.md
taneyland: Looking into this option!

designs/embed_eksd_manifest.md
taneyland: This will be in the `default` namespace. (only CAPI resources are automatically placed in `eksa-system`)
True, there isnt a specific need for the `url` at the moment. The only thing actually needed is the `name`, so are you proposing the other two fields to be more so informative and show the distinction that it's not in that eks-a group?

g-gaston: Why would it be in `default`?
This is not something aimed for users to freely modify. 

From our product perspective, I think we decided that anything that is not part of the api we expose it would be in its specific namespace, so we communicate those are "internal" resources.

From a security perspective, this can lead to problems since it references components of the cluster (like the api server container). So basically having access to update this resource gives an attacker access to decide which are the core components of kubernetes.

vivek-koppuru: I agree this should be a part of eksa-system namespace by default, but from the perspective of potentially allowing users to specify it in the future, maybe we should should still ask for a namespace field, and default it to eksa-system.

Url helps determine where we got it from, which I personally think is useful information for a status especially as it is optional. But I am not super set on that so if you feel like it is better not to, that's fine with me

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,88 @@
...
+    name: "name of resource" 
+```
+
+From here, we can call `kubectl get release` using the `eksdReleaseRef.Name` to access the EKSD release manifest from the controller.
+
+### Upgrading EKSD components
+
+Now that the EKSD components are embedded in the cluster, we have to ensure that those components are supported during upgrade.
+We will update the upgrader in cluster manager to not only check for updates to EKSA, but it will also check if updates are needed for the EKSD objects on the cluster.
+We will have a method `EksdChangeDiff` to check the EKSD version in the bundle to decide if we need to update the EKSD manifest objects that are applied to the cluster.
```
vivek-koppuru, line 86: This would just be checking `channel` and `number` fields right?

taneyland: Yes, exactly!

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,90 @@
...
+          ...
+      kubeVersion: v1.21.5
+      manifestUrl: https://distro.eks.amazonaws.com/kubernetes-1-21/kubernetes-1-21-eks-8.yaml
+      name: kubernetes-1-21-eks-8
+      number: 8
+      ova:
+         ...
+```
+
+From here, we can use `kubectl apply` to apply the content from the EKSD release CRD & manifest to the cluster.
```
jiayiwang7, line 60: what `manifest` are we applying?

taneyland: the eks-d manifest linked in the eks-d bundle, for example, `https://distro.eks.amazonaws.com/kubernetes-1-21/kubernetes-1-21-eks-8.yaml`

designs/embed_eksd_manifest.md
```diff
@@ -0,0 +1,90 @@
...
+We will do this similarly to how we apply the EKSA components to the cluster, introducing a new task to the create workflow called `InstallEksdComponentsTask`.
+In the upgrade workflow, we will also add a step to apply the EKSD resources as we update other cluster resources.
+
+#### EKS-D bundle changes
+
+We currently don’t pull in the EKSD release CRD in the bundle, but we need to apply this CRD before applying the release manifest.
+I propose introducing a new field `components` in the EKSD bundle to refer to the file in S3 that stores this CRD.
+
+While most of our bundles have a `version` field, the EKS-D bundle does not have this field.
+Instead, the version is typically decided by the `releaseChannel` and the `releaseNumber`.
```
g-gaston, line 29: I think here `name` is our `version`
It uniquely identifies the eksd `Release` object. And it will also be the name of the object in the cluster once we do the "apply"
Can we use that instead of using version channel and number? In fact, `name` contains both the release channel and release number.

taneyland: Yes from the way that we construct the `name`, we could use that as our version here. The only reason why I'm thinking to use release channel & number instead is to depend on the true source of those values rather than what we construct for the name. Thinking of a scenario where the name is constructed differently, but it's definitely an edge case.

g-gaston: That name should be the same name that we find in `metadata.name` in the eks-d manifest, right?

taneyland: Yes that’s the same name

g-gaston: Cool, then my vote would be to rely on this name

g-gaston (approved):

Issue #1150, opened 2022-02-04
Controller Support for Worker Node Taints


danbudris:
The EKS-A Controller should reconcile modifications to the `taints` field of the `workerNodeGroupConfigurations` in such a way that new nodes are rolled out and tainted appropriately. 

In order to provide controller support for taints, we'll need to add a mechanism by which we generate new `kubeadmconfigtemplates` with an update `joinConfiguration` containing the requisite taints, and then modify the `machineDeployemnts` which reference those `kubeadmconfigtemplates` to reference the new `kubeadmconfigtemplate`; this way we can trigger a rollout of the `machinedeployments`, causing new machines to be launched, and the updated `joinConfiguration` in the `kubeadmconfigtemplate` to go into effect. This will also address SSH key rolling. 

As of right now, upstream intentionally does not support 'immutable' `kubeadmconfigtemplates`, but the official guidance is to treat them in the same way we treat the immutable `machineTemplates` -- e.g. create a new one with the updates, modify the `machineDeployment` to use it, trigger rollout & node recreation.

See:
- https://github.com/kubernetes-sigs/cluster-api/issues/4910
- https://cluster-api.sigs.k8s.io/tasks/updating-machine-templates.html

_Originally posted by @danbudris in https://github.com/aws/eks-anywhere/issues/189#issuecomment-1030421998_

danbudris:
support for reconciliation of taints we merged in with #1175

PR #1175, opened 2022-02-08
Reconcile taints

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1150

*Description of changes:*
Allow the EKS-A Controller to reconcile the `Taints` field of the `WokerNodeGroupConfiguration`.

This is done by updating the `joinconfiguration` of the `kubeadmconfigtemplate` associated with the worker nodes `machinedeployment`.

We map a `WorkerNodeGroupConfiguration` to the corresponding `kubeadmconfigtemplate` associated with its `machinedeployment`, compare the taints on the existing worker node to the taints associated with the template, and, if they differ, generate a new `kubeadmconfigtemplate` template; we update the `machinedeployment` to use this template. When we change the machine template associated with the machine deployment, a machine deployment rollout is trigged; new nodes are rolled out; and the new `joinconfiguration` in the `kubeadmconfigtemplate` (e.g. taints) are used during node registration. We therefore go from an updated worker node group configuration in the cluster spec to rolled-out nodes with the new settings.

This pattern can also be used to update node labels, useres/sshkeys, and other values which are set at node registration time in the join configuration (other kubelet extraargs, etc).

## Testing
Create/Update/Delete via CLI

Taints reconciliation via GitOps:
- no taints, add a single taint
    - taint added
- no taints, add an empty slice of taints "taints: []"
    - no change (to node)
- taints an empty slice "taints: []", remove it
    - no change (to node)
- one taint, remove it and add an empty slice of taints "taints: []"
    - taint removed, taints "<none>"
- one taint, remove it and add nothing
    - taint removed, taints "<none>"
- no taints, add 3
    - taints added
- 3 taints, remove 1
    - removes the taint (edited) 


pkg/providers/docker/docker.go
```diff
@@ -139,6 +139,11 @@ func (d *DockerTemplateBuilder) EtcdMachineTemplateName(clusterName string) stri
 	return fmt.Sprintf("%s-etcd-template-%d", clusterName, t)
 }
 
+func (vs *DockerTemplateBuilder) KubeadmConfigTemplateName(clusterName, workerNodeGroupName string) string {
```
danbudris, line 142: these can be moved to a common method to be reused across providers

controllers/controllers/resource/template.go
```diff
@@ -76,23 +76,40 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
...
 	for _, workerNodeGroupConfiguration := range clusterSpec.Spec.WorkerNodeGroupConfigurations {
-		oldVmc, err := r.ExistingVSphereWorkerMachineConfig(ctx, eksaCluster, workerNodeGroupConfiguration)
+		existingKubeadmConfigTemplate, err := r.KubeadmConfigTemplate(ctx, eksaCluster, workerNodeGroupConfiguration)
 		if err != nil {
 			return nil, err
 		}
+		oldWn := MapKubeadmConfigTemplateToWorkerNodeGroupConfiguration(*existingKubeadmConfigTemplate)
+		if vsphere.NeedsNewKubeadmConfigTemplate(&workerNodeGroupConfiguration, oldWn) {
+			kubeadmconfigTemplateNames[workerNodeGroupConfiguration.Name] = templateBuilder.KubeadmConfigTemplateName(clusterName, workerNodeGroupConfiguration.Name)
+		} else {
```
taneyland, line 89: Can we just do
`kubeadmconfigTemplateNames[workerNodeGroupConfiguration.Name] = existingKubeadmConfigTemplate.Name`?

pkg/providers/vsphere/vsphere.go
```diff
@@ -663,13 +672,21 @@ func (vs *VsphereTemplateBuilder) GenerateCAPISpecWorkers(clusterSpec *cluster.S
 	workerSpecs := make([][]byte, 0, len(clusterSpec.Spec.WorkerNodeGroupConfigurations))
 	for _, workerNodeGroupConfiguration := range clusterSpec.Spec.WorkerNodeGroupConfigurations {
 		values := buildTemplateMapMD(clusterSpec, *vs.datacenterSpec, vs.workerNodeGroupMachineSpecs[workerNodeGroupConfiguration.MachineGroupRef.Name], workerNodeGroupConfiguration)
-		_, ok := templateNames[workerNodeGroupConfiguration.Name]
-		if templateNames != nil && ok {
-			values["workloadTemplateName"] = templateNames[workerNodeGroupConfiguration.Name]
+
+		_, ok := workloadTemplateNames[workerNodeGroupConfiguration.Name]
```
taneyland, line 676: nit: one liner?
`if _, ok := workloadTemplateNames[workerNodeGroupConfiguration.Name]; ok {`
same for the other one below

taneyland (commented):
Very nice!

controllers/controllers/resource/template.go
danbudris: oh yeah good call will change that

controllers/controllers/resource/template.go
```diff
@@ -76,23 +76,40 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
 		controlPlaneTemplateName = cp.Spec.MachineTemplate.InfrastructureRef.Name
 	}
 
+	kubeadmconfigTemplateNames := make(map[string]string, len(clusterSpec.Spec.WorkerNodeGroupConfigurations))
 	workloadTemplateNames := make(map[string]string, len(clusterSpec.Spec.WorkerNodeGroupConfigurations))
 	for _, workerNodeGroupConfiguration := range clusterSpec.Spec.WorkerNodeGroupConfigurations {
-		oldVmc, err := r.ExistingVSphereWorkerMachineConfig(ctx, eksaCluster, workerNodeGroupConfiguration)
+		existingKubeadmConfigTemplate, err := r.KubeadmConfigTemplate(ctx, eksaCluster, workerNodeGroupConfiguration)
```
jiayiwang7, line 82: do u think if we can keep `KubeadmConfigTemplate` and `MapKubeadmConfigTemplateToWorkerNodeGroupConfiguration` in the fetcher.go, and just call a method like `ExistingWorkerNodeGroupConfig` to get the `oldWn`? just to keep it consistent w the rest of the code in this method

pkg/providers/docker/docker.go
```diff
@@ -139,6 +139,11 @@ func (d *DockerTemplateBuilder) EtcdMachineTemplateName(clusterName string) stri
...
+	t := vs.now().UnixNano() / int64(time.Millisecond)
+	return fmt.Sprintf("%s-%s-kubeadmconfig-template-%d", clusterName, workerNodeGroupName, t)
```
jiayiwang7, line 144: the name feels too long, maybe just adding a timestamp?

pkg/providers/vsphere/vsphere.go
```diff
@@ -663,13 +672,21 @@ func (vs *VsphereTemplateBuilder) GenerateCAPISpecWorkers(clusterSpec *cluster.S
...
+		if ok {
```
jiayiwang7, line 677: a bit confused here (might be im not understanding it all): isn't we already calculating `workloadTemplateNames` map before calling this method? what is the case that `workerNodeGroupConfiguration.Name` key does not exist in the map?

controllers/controllers/resource/template.go
danbudris: yeah we should be able to do that

danbudris: https://github.com/aws/eks-anywhere/pull/1175/commits/0697232a99fef364b34fcdd25c784bac4473c2f2

controllers/controllers/resource/template.go
danbudris: we'll no longer have access to this same information here, so leaving it as-is

pkg/providers/docker/docker.go
danbudris: removing the `kubeadmconfig` and just adding `template`

danbudris: https://github.com/aws/eks-anywhere/pull/1175/commits/11c985eaa284af33896e87c73aa29e55be453b1e

pkg/providers/vsphere/vsphere.go
danbudris: `GenerateCAPISpecWorkers` is called directly during `create`, and is passed `nil` kubeamconfigTemplateNames and `nil` workernodegroupconfiguration names; it then generates default values for them if they don't exist. This could be simplified

danbudris: https://github.com/aws/eks-anywhere/pull/1175/commits/41432003fb13dec72278a6d322c47c9051c4c974

danbudris: remove redundant defaulting behavior and adjust the create flow to set names

jiayiwang7 (approved):

Issue #396, opened 2021-10-06
Flux E2E tests enhancement


jiayiwang7:
#395 

**What would you like to be added**:

- [ ] Add workload cluster upgrade through Flux tests (version, datastore, folder, resourcepool)
- [x] Refactor existing flux test framework to use cluster filler on git file rewrite (#1176)
- [ ] Update existing Flux upgrade test to remove original CLI upgrade step
- [ ] Support for multiple worker node groups when `validateWorkerNodeUpdates` and `validateWorkerNodeMachineSpec`
- [ ] #1233

**Why is this needed**:

PR #1176, opened 2022-02-08
Refactor GitOps e2e tests

jiayiwang7:
*Issue #, if available:*

#396 

*Description of changes:*

Originally the specific Flux test are "hardcoded" in the `flux.go` under test framework, and all the Flux e2e tests run the same update: scale up the worker node number to 3 and change certain field in the machineconfig in Git repo.

This PR refactors the existing Flux e2e framework and makes it flexible to add GitOps tests by passing in multiple `ClusterE2ETestOpt` similar to the CLI upgrade tests. It writes the new cluster spec that is configurable in the test level, and push it to the remote Git repo. Then verify the cluster status after. The framework also works for both management and workload cluster.

- Refactor flux test framework to accept `ClusterE2ETestOpt` as arg
- Add GitOps e2e tests to workload cluster

*Tests:*

Ran 
```
make e2e
./bin/e2e.test -test.v -test.run TestVSphereUpgradeWorkloadClusterWithGitOps
./bin/e2e.test -test.v -test.run TestDockerUpgradeWorkloadClusterWithGitOps
```
locally and all succeeded.

*Future Work:*

- Once the newly added tests are stabilized, we are going to update the existing Flux related tests to use the same GitOpsUpgrade method used here.
- Add more checks to verify the new cluster is upgraded successfully with all the correct spec. (right now I'm reusing the existing checks from flux framework)
 


test/framework/flux.go
```diff
@@ -274,7 +346,7 @@ func (e *ClusterE2ETest) validateWorkerNodeReplicaUpdates(ctx context.Context) e
 	return e.validateWorkerNodeUpdates(ctx)
 }
 
-func (e *ClusterE2ETest) validateWorkerNodeUpdates(ctx context.Context) error {
+func (e *ClusterE2ETest) validateWorkerNodeUpdates(ctx context.Context, opts ...CommandOpt) error {
```
danbudris, line 349: TODO for the future work around validation: support for multiple worker node groups; this method currently only takes the 0 index wngc into account

danbudris: same with `validateWorkerNodeMachineSpec`

test/framework/flux.go
```diff
@@ -64,13 +64,77 @@ func WithFlux(opts ...api.GitOpsConfigOpt) ClusterE2ETestOpt {
...
+	}
+
+	e.buildClusterConfigFileForGit()
+
+	if err := e.pushConfigChanges(); err != nil {
+		e.T.Errorf("Error pushing local changes to remote git repo: %v", err)
+	}
+	e.T.Logf("Successfully updated version controlled cluster configuration")
+
+	if err := e.validateWorkerNodeUpdates(ctx); err != nil {
```
danbudris, line 110: did you have plans for these specific validation methods in future work? I feel like these should be moved out of `upgradeThroughGitOps` and into their own separate validation methods, separate from the initialization of the repository/cluster config generation/git push

thinking that as a test user I may want to modify the cluster in some way after performing the upgrade via git, before performing the validation (for example, modifying the cluster directly and then validating that it returns to the expected state, just as one idea; or running a CLI upgrade flow and then validating it)

for now, even just consider bumping the validations up to the public `UpgradeThroughGitops` method so we set up the distinction between the initialization/push and the validation

test/framework/workload.go
```diff
@@ -21,3 +21,7 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
 	opts = append(opts, withKubeconfig(w.managementClusterKubeconfigFile()))
 	w.deleteCluster(opts...)
 }
+
+func (w *WorkloadCluster) UpgradeClusterThroughGitOps(clusterOpts []ClusterE2ETestOpt) {
```
danbudris, line 25: `UpgradeClusterWithGitOps`?

test/framework/multicluster.go
```diff
@@ -47,3 +53,7 @@ func (m *MulticlusterE2ETest) CreateManagementCluster() {
 func (m *MulticlusterE2ETest) DeleteManagementCluster() {
 	m.ManagementCluster.DeleteCluster()
 }
+
+func (m *MulticlusterE2ETest) UpgradeManagementClusterThroughGitOps(clusterOpts []ClusterE2ETestOpt) {
```
danbudris, line 57: `UpgradeManagementClusterWithGitOps`

test/framework/flux.go
```diff
@@ -64,13 +64,77 @@ func WithFlux(opts ...api.GitOpsConfigOpt) ClusterE2ETestOpt {
 	}
 }
 
+func WithClusterUpgradeGit(fillers ...api.ClusterFiller) ClusterE2ETestOpt {
+	return func(e *ClusterE2ETest) {
+		e.ClusterConfigB = e.customizeClusterConfig(e.clusterConfigGitPath(), fillers...)
+
+		// TODO: the gitopsconfig object is generated from api.NewGitOpsConfig each time
+		// instead of marshalling from the file in git repo. Flux requires namespace always
```
danbudris, line 72: not sure I understand this comment

test/e2e/workload_clusters_test.go
```diff
@@ -20,6 +20,17 @@ func runWorkloadClusterFlow(test *framework.MulticlusterE2ETest) {
 	test.DeleteManagementCluster()
 }
 
+func runWorkloadClusterFlowWithGitOps(test *framework.MulticlusterE2ETest, clusterOpts ...framework.ClusterE2ETestOpt) {
+	test.CreateManagementCluster()
+	test.RunInWorkloadClusters(func(w *framework.WorkloadCluster) {
```
danbudris, line 25: do you have thoughts about how we could upgrade a single workload cluster via gitops in a test which has multiple workload clusters?

danbudris (commented):
curious about your plans for the validation mentioned in the 'future work' section of the description...it seems like we'll want to provide a way to map cluster opts to validations in order to easily generate validations based on the difference between the initial cluster state and the opts provide to the gitops flow

test/e2e/workload_clusters_test.go
jiayiwang7: we can potentially have a method similar to https://github.com/aws/eks-anywhere/blob/main/test/framework/multicluster.go#L35
Something like
```
RunInWorkloadCluster(clusterName string, flow func(*WorkloadCluster))
```
and use it in the GitOps flow.

Just seems not a significant test case to test a single workload cluster in multiple workload clusters settings.

danbudris (approved):

test/framework/flux.go
jiayiwang7: updated the comment, is it better?

test/framework/flux.go
jiayiwang7: added in https://github.com/aws/eks-anywhere/issues/396

g-gaston: On that topic for future work:

I would love us to build a package that takes a kubernetes client (kubectl or go-client or an interface that abstracts them) and a cluster spec and validates it. Something like `ValidateEKSACluster(spec *cluster.Spec, client KubeClient) error`

I think we have a bunch of very useful validations sparse across different places, all with different cases. I'm sure we can reused most of them. But I believe the idea of a central place to answer the question "does my cluster conform to how I specked it" is very powerful. Moreover, it will help make a bunch of parts of our system more robust, not only the tests.

test/framework/cluster.go
```diff
@@ -388,6 +392,15 @@ func (e *ClusterE2ETest) kubeconfigFilePath() string {
 	return filepath.Join(e.ClusterName, fmt.Sprintf("%s-eks-a-cluster.kubeconfig", e.ClusterName))
 }
 
+func (e *ClusterE2ETest) managementKubeconfigFilePath() string {
```
g-gaston, line 395: Curious, why do we need this?
Ideally `ClusterE2ETest` should stay agnostic from the management vs workload distinction. It's just represents a test over an eksa cluster.

The multi cluster test is the one that should hold those concerns.

test/framework/flux.go
jiayiwang7: great point and I agree we should. This validation method should be shared among e2e cli create/upgrade and gitops.

added in https://github.com/aws/eks-anywhere/issues/396

test/framework/multicluster.go
```diff
@@ -23,6 +23,12 @@ func NewMulticlusterE2ETest(t *testing.T, managementCluster *ClusterE2ETest, wor
 	for i, c := range workloadClusters {
 		c.clusterFillers = append(c.clusterFillers, api.WithManagementCluster(managementCluster.ClusterName))
 		c.ClusterName = fmt.Sprintf("%s-w-%d", managementCluster.ClusterName, i)
+
+		if c.GitOpsConfig != nil {
+			c.GitOpsConfig.SetName(c.ClusterName)
+			c.clusterFillers = append(c.clusterFillers, api.WithGitOpsRef(c.ClusterName))
```
g-gaston, line 29: Could this leave in the `ClusterE2ETest`? (constructor or any of the other "setup" methods)

It seems to be something not tied to the concept of workload clusters and it only needs the `ClusterE2ETest` itself

test/framework/workload.go
```diff
@@ -21,3 +21,7 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
 	opts = append(opts, withKubeconfig(w.managementClusterKubeconfigFile()))
 	w.deleteCluster(opts...)
 }
+
+func (w *WorkloadCluster) UpgradeClusterWithGitOps(clusterOpts []ClusterE2ETestOpt) {
```
g-gaston, line 25: Do we need this method?
We should be able to call `UpgradeWithGitOps` directly on a `WorkloadCluster`

g-gaston (approved):

test/framework/flux.go
jiayiwang7: added in https://github.com/aws/eks-anywhere/issues/396

test/framework/flux.go
```diff
@@ -64,13 +64,77 @@ func WithFlux(opts ...api.GitOpsConfigOpt) ClusterE2ETestOpt {
...
 	}
 }
 
+func (e *ClusterE2ETest) UpgradeThroughGitOps(clusterOpts []ClusterE2ETestOpt) {
+	e.upgradeThroughGitOps(clusterOpts)
+}
+
+func (e *ClusterE2ETest) upgradeThroughGitOps(clusterOpts []ClusterE2ETestOpt) {
+	ctx := context.Background()
+	e.initGit(ctx)
```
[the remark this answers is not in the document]
jiayiwang7: `e.initGit(ctx)` is only needed when upgrading with GitOps. Installing Flux in a create cluster test does not require it. That's why I don't think we should move it to WithFlux()

test/framework/cluster.go
jiayiwang7: To validate the workload cluster's CAPI objects in management cluster. It's going to be much more refactoring work and I prefer to do it in separate PR. We can refactor this when we modularize the cluster validation.

test/framework/multicluster.go
jiayiwang7: Not sure if i can. The workload cluster name is generated after the flux setup. I need to update GitOpsConfig name after that.

jiayiwang7: avoided overriding gitopsconfig name by always generating new names

Issue #726, opened 2021-11-30
Cilium policy enforcement configuration


g-gaston:
Add support for Cilium's [policy enforcement mode](https://docs.cilium.io/en/stable/policy/intro/) through the cluster spec api

PR #1183, opened 2022-02-10
Add design doc for configuring cilium

mrajashree:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/726

*Description of changes:*
This design doc covers the API and implementation changes required for supporting policy enforcement mode in Cilium

### DIFF
--- a/designs/cilium-configuration-policy.md
+++ b/designs/cilium-configuration-policy.md
@@ -1,0 +1,146 @@
+# Configure Cilium CNI
+
+## Introduction
+
+**Problem:** The Cilium CNI plugin (built-in default for EKS Anywhere) supports defining and enforcement of network policies. Network policies determine how pods within a cluster are able to communicate with one another. Cilium accepts policy enforcement modes from the users to determine the allowed traffic between endpoints. Currently there is no way to configure Cilium launched by EKS Anywhere to use any mode apart from the default mode, which allows all traffic.
+
+### Goals and Objectives
+
+As an EKS Anywhere user:
+
+* I want to have the ability to configure Cilium's [supported policy enforcement modes](https://docs.cilium.io/en/v1.8/policy/intro/).
+* I want that, depending on the policy enforcement mode I select, EKS Anywhere should configure the network policies required for its core components to ensure that the cluster gets provisioned successfully.
+
+### Non-Goals/Future Work
+
+* Accepting network policies for user workloads through EKS Anywhere spec.
+
+
+## Overview of Solution
+
+With this feature, a user can launch Cilium with one of its supported policy enforcement modes: default, always, never. The default mode will be Cilium's default, which is also named `default`. The user can specify the mode through the EKS Anywhere cluster spec, and EKS Anywhere will use that while launching the Cilium Helm chart in the workload cluster.
+
+### Implementation details
+
+#### API Changes
+
+A new type will be added to accept the cni plugin name and additional options from the user:
+
+```go
+type CNIConfig struct {
+  Cilium *CiliumConfig `json:"cilium,omitempty"`
+}
+
+type CiliumConfig struct {
+  PolicyEnforcementMode string `json:"policyEnforcementMode,omitempty"`
+}
+```
+
+The user can then configure cilium, starting with the policy enforcement mode as follows:
+```yaml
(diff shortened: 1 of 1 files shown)


designs/cilium-configuration-policy.md
```diff
@@ -0,0 +1,119 @@
...
+With this feature, a user can launch Cilium with one of its supported policy enforcement modes: default, always, never. The default mode will be Cilium's default, which is also named `default`. The user can specify the mode through the EKS Anywhere cluster spec, and EKS Anywhere will use that while launching the Cilium Helm chart in the workload cluster.
+
+### Implementation details
+
+#### API Changes
+
+A new type will be added to accept the cni plugin name and additional options from the user:
+
+```go
+type CNIConfig struct {
```
jiayiwang7, line 30: kinda want to know the possibility of using`CNIProviderRef` here. Especially when we want to support more CNIs

designs/cilium-configuration-policy.md
```diff
@@ -0,0 +1,119 @@
...
+  + capi-system
+  + capi-kubeadm-bootstrap-system
+  + capi-kubeadm-control-plane-system
+  + etcdadm-bootstrap-provider-system
+  + etcdadm-controller-system
+  + cert-manager
+- Infrastruture provider's namespace (for instance, capd-system OR capv-system)
+
+For a workload cluster managed by another cluster, EKS Anywhere will create NetworkPolicy resource only in the following namespace by default:
+- kube-system
```
jaxesn, line 104: Are you thinking of opening ingress to the kube-system pods only to other kube-system pods by default, or all pods across all namespaces?

jiayiwang7 (approved):

designs/cilium-configuration-policy.md
vignesh-goutham: I think ingress from only `kube-system` pods to other `kube-system` pods should suffice on the workload cluster. On the management cluster though, we should be allowing from/to across all the listed namespaces.

mrajashree: >On the management cluster though, we should be allowing from/to across all the listed namespaces.

Agreed. But I was thinking the same for workload clusters, since the user apps running on workload clusters would also communicate with kube-apiserver from kube-system. Or should we only allow ingress egress for kube-apiserver across all namespaces on both clusters?

designs/cilium-configuration-policy.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Introduction
+
+**Problem:** The Cilium CNI plugin (built-in default for EKS Anywhere) supports defining and enforcement of network policies. Network policies determine how pods within a cluster are able to communicate with one another. Cilium accepts policy enforcement modes from the users to determine the allowed traffic between endpoints. Currently there is no way to configure Cilium launched by EKS Anywhere to use any mode apart from the default mode, which allows all traffic.
+
+### Goals and Objectives
+
+As an EKS Anywhere user:
+
+* I want to have the ability to configure Cilium's [supported policy enforcement modes](https://docs.cilium.io/en/v1.8/policy/intro/).
```
abhay-krishna, line 11: nit: should we reference v1.9 or v1.10 or latest (least preferred) here?
```suggestion
* I want to have the ability to configure Cilium's [supported policy enforcement modes](https://docs.cilium.io/en/v1.9/policy/intro/).
```

designs/cilium-configuration-policy.md
vignesh-goutham: I'd rather have networking open only within the `kube-system` namespace for workload cluster. If a user is explicitly selecting the non-default `always` policyEnforcementMode they should expect that they need to open additional networking for their workload pods. With `always` mode, eks-a should only open up networking for cluster ops.

mrajashree: modified the doc to show which networkPolicy we'd use in that case. If we find anything wrong with this approach during implementation I'll update the design doc again

designs/cilium-configuration-policy.md
mrajashree: I think either of the versions should be fine since they all contain policy enforcement modes

vignesh-goutham (approved):
/lgtm

PR #1193, opened 2022-02-10
Update newly added GitOps e2e test name to include "Flux"

jiayiwang7:
*Issue #, if available:*

The Flux e2e test envs are only set up when the test name includes "Flux" in it (https://github.com/aws/eks-anywhere/blob/main/internal/test/e2e/flux.go#L12). 

*Description of changes:*

Change the newly added GitOps test name to include "Flux" so that the envs can be properly set. Also to keep it consistent with the existing tests. Later we can change all flux tests name to "GitOps" which makes more sense.


danbudris (approved):
/lgtm

PR #1194, opened 2022-02-10
E2E tests for multiple worker node groups

g-gaston:
*Description of changes:*
Expands current e2e testing framework to support multiple worker node
groups. It allows to add/remove/update worker node groups to the initial
spec and during upgrades.

Adds one simple vSphere tests that starts with two worker node groups.
During the upgrade, it scales down one of them to 1 replica, it deletes
the other one and creates a new one.

This will need to be refactored eventually. The framework api is becoming a bit convoluted and expanding the framework itself is quite difficult. We will create tickets to refactor this in the near future. We don't do it now because it requires to rethink a bit the framework internals and in particular the way we apply api object updates and that will take some time.

We are postponing that refactor because at least this adds the ability to test multiple worker node groups and unblocks subsequent work.

### DIFF
--- a/internal/pkg/api/cluster.go
+++ b/internal/pkg/api/cluster.go
@@ -2,18 +2,29 @@ package api
 
 import (
 	"fmt"
+	"io/ioutil"
 
 	"sigs.k8s.io/yaml"
 
-	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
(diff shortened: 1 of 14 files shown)


danbudris:
all of the checks have passed, and then were re-trigged by commentary. Overriding. 
/override eks-anywhere-attribution-files-presubmit
/override eks-anywhere-presubmit

PR #1202, opened 2022-02-12
Add support for 1-22 with feature flag

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Adding support for using the EKS-D dev release of 1-22 and putting that behind a feature flag. Added tests specifically for Ubuntu as our Bottlerocket builds aren't ready yet. Also added an ability to set env vars in the tests for being able to run tests that require feature flags.


danbudris (approved):
/lgtm

jaxesn (approved):
/lgtm

pkg/validations/cluster.go
```diff
@@ -28,13 +28,22 @@ func ValidateTaintsSupport(clusterSpec *cluster.Spec) error {
...
+				return fmt.Errorf("Node labels feature is not enabled. Please set the env variable %v.", features.NodeLabelsSupportEnvVar)
 			}
 		}
 	}
 	return nil
 }
+
+func ValidateK8s122Support(clusterSpec *cluster.Spec) error {
+	if !features.IsActive(features.K8s122Support()) {
+		if clusterSpec.Cluster.Spec.KubernetesVersion == "1.22" {
```
g-gaston, line 44: `== anywherev1.Kube122`

test/e2e/upgrade_test.go
```diff
@@ -47,6 +47,22 @@ func TestVSphereKubernetes120UbuntuTo121Upgrade(t *testing.T) {
...
+	test := framework.NewClusterE2ETest(
+		t,
+		provider,
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube121)),
+	)
+	runSimpleUpgradeFlow(
+		test,
+		v1alpha1.Kube122,
+		framework.WithClusterUpgrade(api.WithKubernetesVersion(v1alpha1.Kube122)),
+		framework.WithEnvVar("K8S_1_22_SUPPORT", "true"),
```
g-gaston, line 61: We have some stuff already in the e2e setup binary that set env vars depending on the test name. Or at least we used to for IAM tests

I would vote to do the same for these tests (if test name has "122" then set flag env var)

The main advantages are there will be only one place to cleanup and we keep all the feature flags for tests in one place

PR #1203, opened 2022-02-12
Check uniqueness of control plane endpoint IP for provider tinkerbell

ptrivedi:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/providers/tinkerbell/validator.go
+++ b/pkg/providers/tinkerbell/validator.go
@@ -11,12 +11,14 @@ import (
 )
 
 type Validator struct {
-	tink ProviderTinkClient
+	tink      ProviderTinkClient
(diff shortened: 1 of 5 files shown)


pkg/providers/tinkerbell/validator.go
```diff
@@ -94,6 +102,14 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, tinkerbel
 	return nil
 }
 
+func (v *Validator) validateControlPlaneIpUniqueness(tinkerBellClusterSpec *spec) error {
```
panktishah26, line 105: Should we move validateControlPlaneIpUniqueness to one common place from where each provider can use the same validation if possible?

PR #1205, opened 2022-02-14
Replace EKS with EKS-A in docs

abhay-krishna:
Using EKS in these contexts (atleast the second) would be a bit misleading.


TerryHowe (approved):
/lgtm

PR #1213, opened 2022-02-14
Update e2e test flow so cluster ip pool is not required

gwesterfieldjr:
no issue

Update e2e test flow so cluster ip pool is not required

Ran local e2e tests

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -41,7 +41,7 @@ VSPHERE_USERNAME
 VSPHERE_PASSWORD
 T_VSPHERE_CIDR
 GOVC_URL
-T_CLUSTER_IP_POOL
+T_CLUSTER_IP_POOL # comma-separated list of CP ip addresses
 ```
 
 # OIDC tests requisites
--- a/test/framework/vsphere.go
+++ b/test/framework/vsphere.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/eks-anywhere/internal/pkg/api"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/networkutils"
 )
 
 const (
@@ -58,7 +59,6 @@ var requiredEnvVars = []string{
 	cidrVar,
 	privateNetworkCidrVar,
 	govcUrlVar,
-	ClusterIPPoolEnvVar,
 }
 
 type VSphere struct {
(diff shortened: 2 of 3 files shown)


test/e2e/README.md
```diff
@@ -40,7 +40,6 @@ VSPHERE_USERNAME
 VSPHERE_PASSWORD
 T_VSPHERE_CIDR
 GOVC_URL
-T_CLUSTER_IP_POOL
```
vivek-koppuru, line 43: I think we can still add a section below as to how this env var can be used.

test/framework/vsphere.go
```diff
@@ -247,13 +242,34 @@ func (v *VSphere) WithNewVSphereWorkerNodeGroup(name string, workerNodeGroup *Wo
 }
 
 func (v *VSphere) ClusterConfigFillers() []api.ClusterFiller {
-	return v.clusterFillers
+	_, ok := os.LookupEnv(ClusterIPPoolEnvVar)
+	var clusterIP string
+	var err error
+	if ok {
+		clusterIP, err = PopIPFromEnv(ClusterIPPoolEnvVar)
```
vivek-koppuru, line 249: It should be if it is not set or if the string is empty right?

test/framework/vsphere.go
```diff
@@ -114,11 +114,6 @@ func NewVSphere(t *testing.T, opts ...VSphereOpt) *VSphere {
 	}
 
 	v.cidr = os.Getenv(cidrVar)
-	clusterIP, err := PopIPFromEnv(ClusterIPPoolEnvVar)
```
gwesterfieldjr, line 117: @g-gaston , I moved this back to ClusterConfigFillers otherwise only 1 ip address gets set for all clusters in a multi cluster test since the provider is created once and passed to each cluster test.

test/framework/vsphere.go
gwesterfieldjr: Pop ip catches it if its not empty, but I'll add it here as well

test/framework/vsphere.go
```diff
@@ -247,13 +242,34 @@ func (v *VSphere) WithNewVSphereWorkerNodeGroup(name string, workerNodeGroup *Wo
...
+	if ok {
+		clusterIP, err = PopIPFromEnv(ClusterIPPoolEnvVar)
+		if err != nil {
+			v.t.Fatalf("failed to pop cluster ip from test environment: %v", err)
+		}
+	} else {
+		clusterIP = v.generateUniqueIp()
+	}
+
+	return []api.ClusterFiller{api.WithControlPlaneEndpointIP(clusterIP)}
```
g-gaston, line 257: `v.clusterFillers` can have more than just `WithControlPlaneEndpointIP`, this will break tests with multiple worker node groups

test/framework/vsphere.go
g-gaston: In that case I would vote to create multiple copies of the provider instead
I will start thinking how to make those tests better but I'm not sure the provider can be shared with the current design

test/framework/vsphere.go
gwesterfieldjr: I was gonna ask if I needed to append it! thanks for the info. I will update

g-gaston (approved):
/lgtm

PR #1215, opened 2022-02-15
Hardware to bmcRef validation

pokearu:
*Description of changes:*
Adding support to parse tinkerbell hardware config file and perform hardware to bmc validations.

1. Every hardware has a unique BmcRef
2. Hardware's BmcRef is present in the hardwareConfig

*Testing:*
Tested by running `eksctl-anywhere create cluster`

### DIFF
--- a/pkg/hardware/config.go
+++ b/pkg/hardware/config.go
@@ -1,0 +1,98 @@
+package hardware
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	pbnjv1alpha1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/pbnj/api/v1alpha1"
+	tinkv1alpha1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/tink/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type HardwareConfig struct {
+	hardwareList []tinkv1alpha1.Hardware
+	bmcList      []pbnjv1alpha1.BMC
+	secretList   []corev1.Secret
+}
+
+func (hc *HardwareConfig) GetHardwareConfig(hardwareFileName string) error {
+	err := hc.parseHardwareConfig(hardwareFileName)
+	if err != nil {
+		return fmt.Errorf("unable to parse hardware file %s: %v", hardwareFileName, err)
+	}
+	return nil
+}
+
+func (hc *HardwareConfig) parseHardwareConfig(hardwareFileName string) error {
+	content, err := ioutil.ReadFile(hardwareFileName)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/hardware/config.go
```diff
@@ -0,0 +1,98 @@
...
+	bmcList      []pbnjv1alpha1.BMC
+	secretList   []corev1.Secret
+}
+
+func (hc *HardwareConfig) GetHardwareConfig(hardwareFileName string) error {
+	err := hc.parseHardwareConfig(hardwareFileName)
+	if err != nil {
+		return fmt.Errorf("unable to parse hardware file %s: %v", hardwareFileName, err)
+	}
+	return nil
```
panktishah26, line 28: should we return the hardwareConfig struct here?

pokearu: Actually since its a method of a struct, the changes are kinda made by pass by reference.
So `hc` would have the parsed hardware config.

pkg/hardware/config.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		bmcRefMap[hw.Spec.BmcRef] = &hw
+	}
+
+	return nil
+}
+
+func (hc *HardwareConfig) getBmcRefMap() map[string]*tinkv1alpha1.Hardware {
+	bmcRefMap := make(map[string]*tinkv1alpha1.Hardware, len(hc.bmcList))
```
abhinavmpandey08, line 92: looks like `bmcRefMap` is being used to validate that the bmc is not already referenced by another hardware. 
Currently its mapping `string -> Hardware` but I see the hardware just being checked it see if its nil to determine if BMC is already being used by another hardware.

Correct me if im wrong but we can achieve the same with `map[string]bool`? This would save memory as well mainly because the Hardware/BMC list can be huge

abhinavmpandey08 (commented):
looks good, just a couple minor comments

pkg/hardware/config.go
pokearu: Oh yea we can do the same with bool. But i am actually just storing pointers.
Like `map[string]*tinkv1alpha1.Hardware` so i am initializing a nil pointer map.

So I think it would just reference the hardware and not really copy it and use up memory?

pkg/hardware/config.go
pokearu: Yea sure 👍

pkg/hardware/config.go
ptrivedi: I would call this makeBmcRefMap or initBmcRefMap. This will never return a populated list, so it could be confusing to call it 'getBmcRefMap'

ptrivedi:
Looks good to me. Assuming that BMC to SecretRef mapping validation is going to be in a separate PR? Also, what about the validation where two BMCs have same IPMI IP? Also assuming validation of IP format within BMC would be in a different PR.

pokearu:
> Looks good to me. Assuming that BMC to SecretRef mapping validation is going to be in a separate PR? Also, what about the validation where two BMCs have same IPMI IP? Also assuming validation of IP format within BMC would be in a different PR.

Yes and yes. This initial PR mainly has the parsing logic, which would unblock others who are adding similat validations. Will do follow ups!

PR #1216, opened 2022-02-15
Add Kind Haproxy image to bundle

abhay-krishna:
Adding Kind Haproxy image to bundle to be used by CAPD as loadbalancer image.


release/pkg/assets_haproxy.go
```diff
@@ -0,0 +1,86 @@
...
+	releaseImageUri, err := r.GetReleaseImageURI(name, repoName, tagOptions)
+	if err != nil {
+		return nil, errors.Cause(err)
+	}
+
+	imageArtifact := &ImageArtifact{
+		AssetName:         name,
+		SourceImageURI:    sourceImageUri,
+		ReleaseImageURI:   releaseImageUri,
+		Arch:              []string{"amd64"},
```
jaxesn, line 52: Do we only do amd64 still in these places?

abhay-krishna: yeah for now, but i'm changing that in my refactor PR to include both amd64 and arm64. That would make it easier to change, like updating it in one place vs 20 files.

jaxesn: cool

jaxesn (approved):
/lgtm

PR #1217, opened 2022-02-15
Add EKS-D release ref to cluster status

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR adds `eksdReleaseRef` to the cluster status. This will allow us to know which eksd release object to grab from the cluster.

More details on why this change is needed in design doc #1162 


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -372,6 +372,19 @@ type ClusterStatus struct {
...
+
+type EksdReleaseRef struct {
+	// ApiGroup refers to the EKS-D API grouo
+	ApiGroup string `json:"apiGroup,omitempty"`
+	// Kind refers to the Release kind for the EKS-D object
+	Kind string `json:"kind,omitempty"`
+	// Name refers to the name of the EKS-D object on the cluster
+	Name string `json:"name,omitempty"`
+	// Namespace refers to the namespace for the EKS-D release resources
+	Namespace string `json:"namespace,omitempty"`
```
g-gaston, line 387: I would vote to make all these fields inside `EksdReleaseRef` required
```suggestion
	ApiGroup string `json:"apiGroup"`
	// Kind refers to the Release kind for the EKS-D object
	Kind string `json:"kind"`
	// Name refers to the name of the EKS-D object on the cluster
	Name string `json:"name"`
	// Namespace refers to the namespace for the EKS-D release resources
	Namespace string `json:"namespace"`
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -50,6 +53,13 @@ func NewClusterGenerate(clusterName string, opts ...ClusterGenerateOpt) *Cluster
 				CNI: Cilium,
 			},
 		},
+		Status: ClusterStatus{
```
g-gaston, line 56: I would vote to not generate this here and just set in the cli during create/upgrade
I don't think we should concern the user with this

pkg/api/v1alpha1/cluster.go
```diff
@@ -14,15 +14,18 @@ import (
...
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 )
 
 const (
 	ClusterKind         = "Cluster"
+	ReleaseKind         = "Release"
 	YamlSeparator       = "\n---\n"
 	RegistryMirrorCAKey = "EKSA_REGISTRY_MIRROR_CA"
+	EksDistroApiGroup   = "distro.eks.amazonaws.com/v1alpha1"
```
g-gaston, line 28: This probably belongs in the `cluster` package or maybe even just `constants`

pkg/api/v1alpha1/cluster.go
taneyland: Sure, makes sense. If we want to allow the user the configure the namespace for this resource in the future, we will have to expose this.

pkg/api/v1alpha1/cluster_types.go
taneyland: Updated! Right now the fields are empty in those test files but it will be populated when I implement the logic in the create/upgrade workflow in the next PR

pkg/addonmanager/addonclients/testdata/cluster-config-default-path-management.yaml
```diff
@@ -15,6 +15,12 @@ spec:
   kubernetesVersion: "1.19"
   managementCluster:
     name: management-cluster
+status:
+  eksdReleaseRef:
```
g-gaston, line 19: Curious why this is needed here?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -458,7 +471,8 @@ type ClusterGenerate struct {
 	metav1.TypeMeta `json:",inline"`
 	ObjectMeta      `json:"metadata,omitempty"`
 
-	Spec ClusterSpec `json:"spec,omitempty"`
+	Spec   ClusterSpec   `json:"spec,omitempty"`
+	Status ClusterStatus `json:"status,omitempty"`
```
g-gaston, line 475: Interesting, curious why you are adding this here
I would vote to not include it in the generate struct since this should never be changed manually by a human (neither in disk with the cli or through gitops)

g-gaston (commented):
Sorry, one more question that I missed the first time :)

pkg/api/v1alpha1/cluster_types.go
taneyland: Ah yes. Thanks for catching this. I meant to remove this when removing the other generate logic for this

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
```diff
@@ -277,6 +277,30 @@ spec:
           status:
             description: ClusterStatus defines the observed state of Cluster
             properties:
+              eksdReleaseRef:
+                description: EksdReleaseRef defines the properties of the EKS-D object
```
chrisdoherty4, line 281: Yaml is case sensitive so you probably want to use the _actual_ key in the description: `eksdReleaseRef`. Ditto for the other keys.

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
```diff
@@ -277,6 +277,30 @@ spec:
...
+                  on the cluster
+                properties:
+                  apiGroup:
+                    description: ApiGroup refers to the EKS-D API grouo
```
chrisdoherty4, line 285: `grouo` -> `group`.

config/manifest/eksa-components.yaml
```diff
@@ -2803,6 +2803,30 @@ spec:
           status:
             description: ClusterStatus defines the observed state of Cluster
             properties:
+              eksdReleaseRef:
+                description: EksdReleaseRef defines the properties of the EKS-D object
+                  on the cluster
+                properties:
+                  apiGroup:
+                    description: ApiGroup refers to the EKS-D API grouo
```
chrisdoherty4, line 2811: `grouo` -> `group`.

g-gaston (approved):

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -373,6 +373,19 @@ type ClusterStatus struct {
 	// Descriptive message about a fatal problem while reconciling a cluster
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+	// EksdReleaseRef defines the properties of the EKS-D object on the cluster
+	EksdReleaseRef EksdReleaseRef `json:"eksdReleaseRef,omitempty"`
+}
+
+type EksdReleaseRef struct {
+	// ApiGroup refers to the EKS-D API group
+	ApiGroup string `json:"apiGroup"`
```
vivek-koppuru, line 382: Should we call this ApiVersion to be more in line with the cluster-api templates?

vivek-koppuru (approved):
Just one comment, but feel free to merge

vivek-koppuru (approved):
/lgtm

PR #1218, opened 2022-02-15
Use Kind Haproxy image for CAPD loadbalancing

abhay-krishna:
CAPD was initially hardcoded to use `kindest/haproxy` as the loadbalancer image. They made this configurable in `v1alpha4` by adding a `LoadBalancer` field to the `DockerClusterSpec` type. This PR sets this field to use the HAProxy image we build and distribute.

### DIFF
--- a/pkg/providers/docker/docker.go
+++ b/pkg/providers/docker/docker.go
@@ -212,6 +212,8 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec) map[string]interface{} {
 		"auditPolicy":                common.GetAuditPolicy(),
 		"podCidrs":                   clusterSpec.Spec.ClusterNetwork.Pods.CidrBlocks,
 		"serviceCidrs":               clusterSpec.Spec.ClusterNetwork.Services.CidrBlocks,
+		"haproxyImageRepository":     getHAProxyImageRepo(bundle.Haproxy.Image),
+		"haproxyImageTag":            bundle.Haproxy.Image.Tag(),
 	}
 
 	if clusterSpec.Spec.ExternalEtcdConfiguration != nil {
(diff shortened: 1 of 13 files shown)


pkg/providers/docker/docker.go
```diff
@@ -212,6 +213,8 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec) map[string]interface{} {
 		"auditPolicy":                common.GetAuditPolicy(),
 		"podCidrs":                   clusterSpec.Spec.ClusterNetwork.Pods.CidrBlocks,
 		"serviceCidrs":               clusterSpec.Spec.ClusterNetwork.Services.CidrBlocks,
+		"haproxyImageRepository":     strings.ReplaceAll(bundle.Haproxy.Image.Image(), "/haproxy", ""),
```
abhay-krishna, line 216: CAPD expects only the partial repository URI prefix here, as it appends `haproxy` to the prefix to get the final URI. By default this field is set to `kindest`, so the full URI becomes `kindest/haproxy`. We override it by setting it to `public.ecr.aws/<registry alias>/kubernetes-sigs/kind`, to form the final URI `public.ecr.aws/<registry alias>/kubernetes-sigs/kind/haproxy`.

abhay-krishna: I tried using filepath.Dir(), but it returned a `.` for empty string input, which isn't ideal. If using that and having a check in the template to replace `.` with empty string is more preferrable, I can change to using that.

jiayiwang7: instead of using replaceAll, can we possibly check if the image follows the patten (regex) and then remove the last `/haproxy` part?

jaxesn:
Ran this locally, looks good. thanks!

pkg/providers/docker/docker.go
abhay-krishna: Updated.

PR #1219, opened 2022-02-15
Adding kubectl for cloudstack

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR is a continuation of the integration of the CloudStack provider into EKS-A. It adds support for querying cloudstack items using the kubectl executable. Most of the code is copied/pasted from vsphere.

*Testing (if applicable):*
Cluster creation succeeded locally (with other changes). EKS-A builds, and all older unit tests succeed.

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -33,22 +33,112 @@ const (
 )
 
 var (
-	capiClustersResourceType          = fmt.Sprintf("clusters.%s", clusterv1.GroupVersion.Group)
-	eksaClusterResourceType           = fmt.Sprintf("clusters.%s", v1alpha1.GroupVersion.Group)
-	eksaVSphereDatacenterResourceType = fmt.Sprintf("vspheredatacenterconfigs.%s", v1alpha1.GroupVersion.Group)
-	eksaVSphereMachineResourceType    = fmt.Sprintf("vspheremachineconfigs.%s", v1alpha1.GroupVersion.Group)
-	eksaAwsResourceType               = fmt.Sprintf("awsdatacenterconfigs.%s", v1alpha1.GroupVersion.Group)
-	eksaGitOpsResourceType            = fmt.Sprintf("gitopsconfigs.%s", v1alpha1.GroupVersion.Group)
-	eksaOIDCResourceType              = fmt.Sprintf("oidcconfigs.%s", v1alpha1.GroupVersion.Group)
-	eksaAwsIamResourceType            = fmt.Sprintf("awsiamconfigs.%s", v1alpha1.GroupVersion.Group)
-	etcdadmClustersResourceType       = fmt.Sprintf("etcdadmclusters.%s", etcdv1.GroupVersion.Group)
-	bundlesResourceType               = fmt.Sprintf("bundles.%s", releasev1alpha1.GroupVersion.Group)
-	clusterResourceSetResourceType    = fmt.Sprintf("clusterresourcesets.%s", addons.GroupVersion.Group)
-	kubeadmControlPlaneResourceType   = fmt.Sprintf("kubeadmcontrolplanes.controlplane.%s", clusterv1.GroupVersion.Group)
+	capiClustersResourceType             = fmt.Sprintf("clusters.%s", clusterv1.GroupVersion.Group)
+	eksaClusterResourceType              = fmt.Sprintf("clusters.%s", v1alpha1.GroupVersion.Group)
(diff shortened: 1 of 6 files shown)


pkg/executables/kubectl.go
```diff
@@ -33,24 +33,114 @@ const (
...
+	bundlesResourceType                  = fmt.Sprintf("bundles.%s", releasev1alpha1.GroupVersion.Group)
+	clusterResourceSetResourceType       = fmt.Sprintf("clusterresourcesets.%s", addons.GroupVersion.Group)
+	kubeadmControlPlaneResourceType      = fmt.Sprintf("kubeadmcontrolplanes.controlplane.%s", clusterv1.GroupVersion.Group)
 )
 
 type Kubectl struct {
 	Executable
 }
 
+func (k *Kubectl) SearchCloudStackMachineConfig(ctx context.Context, name string, kubeconfigFile string, namespace string) ([]*v1alpha1.CloudStackMachineConfig, error) {
```
g-gaston, line 56: Could you explain the difference in usecase between "search" and "get" methods?
The "get" only returns one item and the search returns a slice. However the "search" is by name and in a specific namespace, which means (I think) that there should only at most only one object.

Aren't they basically doing the same thing?

pkg/executables/kubectl.go
```diff
@@ -33,24 +33,114 @@ const (
...
+func (k *Kubectl) DeleteEksaCloudStackDatacenterConfig(ctx context.Context, cloudstackDeploymentConfigName string, kubeconfigFile string, namespace string) error {
+	params := []string{"delete", eksaCloudStackDatacenterResourceType, cloudstackDeploymentConfigName, "--kubeconfig", kubeconfigFile, "--namespace", namespace, "--ignore-not-found=true"}
+	_, err := k.Execute(ctx, params...)
+	if err != nil {
+		return fmt.Errorf("error deleting cloudstackdeploymentconfig cluster %s apply: %v", cloudstackDeploymentConfigName, err)
+	}
+	return nil
+}
+
+func (k *Kubectl) GetEksaCloudStackDatacenterConfig(ctx context.Context, cloudstackDatacenterConfigName string, kubeconfigFile string, namespace string) (*v1alpha1.CloudStackDatacenterConfig, error) {
```
g-gaston, line 84: You might be able to simplify this method by using this https://github.com/aws/eks-anywhere/blob/678d8bca08065030a7abc5123f8d0c486e66e372/pkg/executables/kubectl.go#L1195-L1206

And here is an example on how we use it https://github.com/aws/eks-anywhere/blob/678d8bca08065030a7abc5123f8d0c486e66e372/pkg/executables/kubectl.go#L1208-L1215

pkg/executables/kubectl.go
```diff
@@ -33,24 +33,114 @@ const (
...
+	params := []string{
+		"get", eksaCloudStackMachineResourceType, "-o", "json", "--kubeconfig",
+		kubeconfigFile, "--namespace", namespace, "--field-selector=metadata.name=" + name,
+	}
+	stdOut, err := k.Execute(ctx, params...)
+	if err != nil {
+		return nil, fmt.Errorf("error searching eksa CloudStackMachineConfigResponse: %v", err)
+	}
+
+	response := &CloudStackMachineConfigResponse{}
```
g-gaston, line 66: There should be already a struct for this
like `v1alpha1.CloudStackMachineConfigList`

g-gaston (commented):
Could we add some unit tests? :)

pkg/executables/kubectl.go
maxdrib: I'll have a peek at this - seems very useful, almost like generics

pkg/executables/kubectl.go
maxdrib: I see a difference here in the presence of the omitempty tag. I'm not sure currently how significant that is, but given that it's defined in both places for VSphere, I wonder if it makes sense to keep both

maxdrib (commented):
Will add unit tests for the newly defined methods

vivek-koppuru (approved):

pkg/executables/kubectl_test.go
```diff
@@ -621,6 +621,135 @@ func TestKubectlGetMachines(t *testing.T) {
 	}
 }
 
+func TestKubectlGetEksaCloudStackMachineConfig(t *testing.T) {
```
g-gaston, line 624: This might help with these tests :)

https://github.com/aws/eks-anywhere/blob/main/pkg/executables/kubectl_test.go#L1404

g-gaston (approved):

PR #1221, opened 2022-02-15
Make new bundle fields optional

g-gaston:
*Description of changes:*
Without this, old `Bundle`'s become invalid since the api server makes them required


config/crd/bases/anywhere.eks.amazonaws.com_bundles.yaml
```diff
@@ -1659,8 +1659,6 @@ spec:
                               description: The image repository, name, and tag
                               type: string
                           type: object
-                        version:
```
g-gaston, line 1662: I didn't change this, I assume this is because we never regenerated the manifests after removing that field
@abhay-krishna

TerryHowe (approved):

TerryHowe (approved):
/lgtm

PR #1223, opened 2022-02-15
Use helm template to generate cilium manifest

mrajashree:
*Issue #, if available:*

*Description of changes:* Right now we use the manifest included in bundle-release to install cilium. This PR switches to using the existing [GenerateManifest](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/cilium/templater.go#L63) method that uses helm template to generate the cilium manifest, this way the values for the helm chart can be configured based on what we accept from the user.

*Testing (if applicable):*
Ran the e2e test `TestVSphereKubernetes120UbuntuTo121Upgrade` on this branch locally


pkg/networking/cilium/cilium_test.go
```diff
@@ -33,34 +32,19 @@ func newCiliumTest(t *testing.T) *ciliumtest {
 		client: *client,
 		h:      h,
 		spec: test.NewClusterSpec(func(s *cluster.Spec) {
-			s.VersionsBundle.Cilium = v1alpha1.CiliumBundle{
-				Cilium: v1alpha1.Image{
-					URI: "public.ecr.aws/isovalent/cilium:v1.9.10-eksa.1",
```
g-gaston, line 38: Why removing this?

g-gaston (approved):

pkg/networking/cilium/cilium_test.go
mrajashree: because it wasn't getting used, and referring the spec for [upgrader test](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/cilium/upgrader_test.go#L41)

g-gaston: I'm ok with that
However, I think the images are being used by the templater when generating the values for helm

PR #1224, opened 2022-02-15
End to end tests for taints

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1121

*Update*
Ok, so after discussion with @jiayiwang7 , I've improved the validation logic.

We're now going through the whole process of deducing the `node` from the `workerNodeGroupConfiguraton`, by:
- map wngc to machine deployment
- map machineDeployment to machineSpec
- iterate through the nodes, checking their 'owner' annotations, and mapping that back to the mahcineSpec.

This way, we get a robust 1-1 validation of taint to node.  This mapping should be trivial but is not (espeically on BR, where the node names are the IP of the hosts, and not related to the node group configuration name), so I've filed a ticket to simplify it (perhaps by putting an annotation on the nodes which has their wngc?) -- https://github.com/aws/eks-anywhere/issues/1236

So this means we get a better validation of node-to-wngc taints mapping than the original implementation.

*Description of changes:*
Add an initial end-to-end test which validates the taints feature works.

We add taints to worker nodes, validate that the taints we expect to be present are present after cluster creation; then, we update the cluster, removing one taint, and adding two more -- one to a node that had no taints, and one to a node that already had taints. After the upgrade finishes, we then validate once again that the taints we expect to be present are present on the set of nodes.

Due to the relative complexity of converting a given worker-node-group name to a node name, we're deducing the taints present by creating a map of `taint` to expected number of that taint, and using that to confirm the expected taints are present in the cluster.

So, this test covers:
- create a node with a taint
- remove a taint from a node
- add a taint to a node which already has another taint
- add a taint to a node which had no taints

Currently running into some snags with BR, so that test coverage will come at a later time.

*Testing (if applicable):*
Ran the tests and they work as expected.

*Follow Up Work*

https://github.com/aws/eks-anywhere/issues/1225 unit tests for new methods of kubectl
https://github.com/aws/eks-anywhere/issues/1226 e2e tests for CP node taints
https://github.com/aws/eks-anywhere/issues/1227 e2e tests for BR node taints

### DIFF
--- a/internal/pkg/api/cluster.go
+++ b/internal/pkg/api/cluster.go
@@ -174,7 +174,7 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
 		position := -1
 		for i, w := range c.Spec.WorkerNodeGroupConfigurations {
 			if w.Name == name {
-				logger.Info("updating worker node group", "name", name)
+				logger.Info("Updating worker node group", "name", name)
 				nodeGroup = &w
 				position = i
 				break
--- a/internal/pkg/api/cluster.go
+++ b/internal/pkg/api/cluster.go
@@ -174,6 +174,7 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
 		position := -1
 		for i, w := range c.Spec.WorkerNodeGroupConfigurations {
 			if w.Name == name {
+				logger.Info("updating worker node group", "name", name)
 				nodeGroup = &w
 				position = i
 				break
@@ -181,6 +182,7 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
 		}
 
 		if nodeGroup == nil {
+			logger.Info("adding worker node group", "name", name)
 			nodeGroup = &anywherev1.WorkerNodeGroupConfiguration{Name: name}
 			c.Spec.WorkerNodeGroupConfigurations = append(c.Spec.WorkerNodeGroupConfigurations, *nodeGroup)
 			position = len(c.Spec.WorkerNodeGroupConfigurations) - 1
--- a/internal/pkg/api/cluster.go
+++ b/internal/pkg/api/cluster.go
@@ -182,7 +182,7 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
 		}
 
 		if nodeGroup == nil {
-			logger.Info("adding worker node group", "name", name)
+			logger.Info("Adding worker node group", "name", name)
 			nodeGroup = &anywherev1.WorkerNodeGroupConfiguration{Name: name}
 			c.Spec.WorkerNodeGroupConfigurations = append(c.Spec.WorkerNodeGroupConfigurations, *nodeGroup)
 			position = len(c.Spec.WorkerNodeGroupConfigurations) - 1
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -293,6 +293,18 @@ func (k *Kubectl) ListCluster(ctx context.Context) error {
 	return nil
 }
 
+func (k *Kubectl) GetNodes(ctx context.Context, kubeconfig string) ([]corev1.Node, error) {
+	params := []string{"get", "nodes", "--kubeconfig", kubeconfig}
+	stdOut, err := k.Execute(ctx, params...)
+	if err != nil {
+		return nil, fmt.Errorf("error getting nodes: %v", err)
+	}
+	response := &corev1.NodeList{}
+	err = json.Unmarshal(stdOut.Bytes(), response)
+
+	return response.Items, err
+}
+
 func (k *Kubectl) ValidateNodes(ctx context.Context, kubeconfig string) error {
 	template := "{{range .items}}{{.metadata.name}}\n{{end}}"
 	params := []string{"get", "nodes", "-o", "go-template", "--template", template, "--kubeconfig", kubeconfig}
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -294,7 +294,7 @@ func (k *Kubectl) ListCluster(ctx context.Context) error {
 }
 
 func (k *Kubectl) GetNodes(ctx context.Context, kubeconfig string) ([]corev1.Node, error) {
-	params := []string{"get", "nodes", "--kubeconfig", kubeconfig}
+	params := []string{"get", "nodes", "-o", "json", "--kubeconfig", kubeconfig}
 	stdOut, err := k.Execute(ctx, params...)
 	if err != nil {
 		return nil, fmt.Errorf("error getting nodes: %v", err)
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/version"
 	vspherev1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	kubeadmv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	addons "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -495,6 +496,31 @@ func (k *Kubectl) GetMachines(ctx context.Context, cluster *types.Cluster, clust
 	return response.Items, nil
 }
 
+type machineSetResponse struct {
+	Items []clusterv1.MachineSet `json:"items,omitempty"`
+}
+
+func (k *Kubectl) GetMachineSets(ctx context.Context, machineDeploymentName string, cluster *types.Cluster) ([]clusterv1.MachineSet, error) {
+	params := []string{
+		"get", "machinesets", "-o", "json", "--kubeconfig", cluster.KubeconfigFile,
+		"--selector=cluster.x-k8s.io/deployment-name=" + machineDeploymentName,
+		"--namespace", constants.EksaSystemNamespace,
+	}
+
+	stdOut, err := k.Execute(ctx, params...)
+	if err != nil {
+		return nil, fmt.Errorf("error getting machineset associated with deployment %s: %v", machineDeploymentName, err)
+	}
+
+	response := &machineSetResponse{}
+	err = json.Unmarshal(stdOut.Bytes(), response)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing get machinesets response: %v", err)
+	}
+
+	return response.Items, nil
+}
+
 type ClustersResponse struct {
 	Items []types.CAPICluster `json:"items,omitempty"`
 }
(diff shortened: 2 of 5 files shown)


test/framework/taints.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/aws/eks-anywhere/pkg/retrier"
+)
+
+func (e *ClusterE2ETest) ValidateNodeTaints(expectedTaints map[corev1.Taint]int) {
```
gwesterfieldjr, line 13: Why put receiver method here instead of in cluster.go? Is there a general practice we follow for when we place receivers outside of the the file containing the type the receiver belongs too?

danbudris: in the case of the test framework we're using the file to differentiate and organize the framework functionality (e.g. taints specific test functionality in `taints.go`, OIDC in `oidc.go`) while still making it really straight forward to run the validations on the given test execution

internal/pkg/api/cluster.go
```diff
@@ -174,13 +174,15 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
 		position := -1
 		for i, w := range c.Spec.WorkerNodeGroupConfigurations {
 			if w.Name == name {
+				logger.Info("updating worker node group", "name", name)
```
chrisdoherty4, line 177: Do we practice injecting dependencies?

test/e2e/taints_test.go
```diff
@@ -0,0 +1,110 @@
...
+	test.CreateCluster()
+	test.ValidateNodeTaints(preUpgradeNodeTaints)
+	test.UpgradeCluster(clusterOpts)
+	test.ValidateCluster(updateVersion)
+	test.ValidateNodeTaints(postUpgradeNodeTaints)
+	test.StopIfFailed()
+	test.DeleteCluster()
+}
+
+func TestVSphereKubernetes121TaintsWorkerNodeGroups(t *testing.T) {
```
danbudris, line 26:
```suggestion
// this test covers the following scenarios:
// create a node with a taint
// remove a taint from a node
// add a taint to a node which already has another taint
// add a taint to a node which had no taints
func TestVSphereKubernetes121TaintsWorkerNodeGroups(t *testing.T) {
```

test/e2e/taints_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func noScheduleWorkerNodeGroup(name string, count int) *framework.WorkerNodeGroup {
+		return framework.WithWorkerNodeGroup(name, api.WithCount(count), api.WithTaint(noScheduleTaint()))
+}
+
+func preferNoScheduleWorkerNodeGroup(name string, count int) *framework.WorkerNodeGroup {
+	return framework.WithWorkerNodeGroup(name, api.WithCount(count), api.WithTaint(preferNoScheduleTaint()))
+}
+
+func noExecuteTaint() corev1.Taint {
```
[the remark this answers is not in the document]
danbudris: done

test/framework/taints.go
```diff
@@ -0,0 +1,54 @@
...
+			return fmt.Errorf("error validating nodes taint presence: %v", err)
+		}
+		return nil
+	})
+	if err != nil {
+		e.T.Fatal(err)
+	}
+	for _, node := range nodes {
+		for _, taint := range node.Spec.Taints {
+			if taint == masterNodeTaint() {
```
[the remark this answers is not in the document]
danbudris: Yes, this allows the CP node taints(node-role.kubernetes.io/master:NoSchedule) to be expected by default; we're iterating over all of the nodes in this method, not just the worker nodes. It just so happens that we haven't tainted the CP nodes with any custom taints.

test/framework/taints.go
jiayiwang7: So `ValidateNodeTaints` does not validate if the taints count and value are correct on the worker node group level / control plane level?

test/e2e/taints_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/test/framework"
+)
+
+func runTaintsUpgradeFlow(test *framework.ClusterE2ETest, updateVersion v1alpha1.KubernetesVersion, preUpgradeNodeTaints, postUpgradeNodeTaints map[corev1.Taint]int, clusterOpts ...framework.ClusterE2ETestOpt) {
```
jiayiwang7, line 15: I feel to some point, we should not create new flows if unnecessary. Taints upgrade should be part of the upgrade workflow. The only difference is this adds two `test.ValidateNodeTaints` steps. To me it should be part of the `ValidateCluster`. And refactor the `ValidateCluster` method to take arbitrary validation checks (in this case, check node taints). So that we dont keep adding new upgrade flow when we simply want to validate different aspects of the cluster.

chrisdoherty4: Are you suggesting there should be a single workflow with assertions placed throughout and that these could just be made a part of that? Is that how e2e workflows are currently modeled?

danbudris: yeah I don't disagree....my motivation at this moment is to clearly validate the taints in isolation with as few additional factors as possible so I can make sure the feature works as expected

jiayiwang7: added in https://github.com/aws/eks-anywhere/issues/1233

chrisdoherty4: E2E tests tend to be bigger with steps and intermediate assertions but I'm wondering if following that pattern would result in obnoxiously large e2e test workflows? Have you got a link to existing workflows?

danbudris: https://github.com/aws/eks-anywhere/tree/main/test/e2e

you can find the various flows related to various functionalities spread out through the tests in this dir

test/framework/taints.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	wn := c.Spec.WorkerNodeGroupConfigurations
+	// deduce the worker node group configuration to node mapping via the machine deployment and machine set
+	for _, w := range wn {
+		mdName := fmt.Sprintf("%v-%v", e.ClusterName, w.Name)
+		md, err := e.KubectlClient.GetMachineDeployment(ctx, mdName, executables.WithKubeconfig(e.cluster().KubeconfigFile), executables.WithNamespace(constants.EksaSystemNamespace))
+		if err != nil {
+			e.T.Fatal(fmt.Errorf("failed to get machine deployment for worker node %s when validating taints: %v", w.Name, err))
+		}
+		ms, err := e.KubectlClient.GetMachineSets(ctx, md.Name, e.cluster())
```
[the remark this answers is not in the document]
danbudris: not after upgrade; we end up with multiple MS and the old ones are not cleaned up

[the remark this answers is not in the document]
danbudris: and they still have the labels on them which associate them with the machine deployment

test/framework/taints.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		for _, node := range nodes {
+			ownerName, ok := node.Annotations[ownerAnnotation]
+			if ok {
+				// there will be multiple machineSets present on a cluster following an upgrade.
+				// find the one that is associated with this worker node, and compare the taints.
+				for _, machineSet := range ms {
+					if ownerName == machineSet.Name {
+						if !v1alpha1.TaintsSliceEqual(node.Spec.Taints, w.Taints) {
```
jiayiwang7, line 51: won't worker node carries some default/system taints? or is it just for control plane nodes?

danbudris: pod security policies are the exception, but as far as I know we don't currently support that or apply them in our tests; for future tests which did, we could append them to the expected taints or ignore them when we compare. The CP tests wil lhave a mechanism to do this, in order to ignore the CP taint

pkg/executables/kubectl.go
```diff
@@ -293,6 +294,18 @@ func (k *Kubectl) ListCluster(ctx context.Context) error {
 	return nil
 }
 
+func (k *Kubectl) GetNodes(ctx context.Context, kubeconfig string) ([]corev1.Node, error) {
+	params := []string{"get", "nodes", "-o", "json", "--kubeconfig", kubeconfig}
+	stdOut, err := k.Execute(ctx, params...)
```
g-gaston, line 299: nit: you could use the `getObject`method for this

pkg/executables/kubectl.go
```diff
@@ -483,6 +496,31 @@ func (k *Kubectl) GetMachines(ctx context.Context, cluster *types.Cluster, clust
 	return response.Items, nil
 }
 
+type machineSetResponse struct {
```
g-gaston, line 499: Isn't there a `clusterv1.MachineSetList`?

pkg/executables/kubectl.go
```diff
@@ -483,6 +496,31 @@ func (k *Kubectl) GetMachines(ctx context.Context, cluster *types.Cluster, clust
...
+	Items []clusterv1.MachineSet `json:"items,omitempty"`
+}
+
+func (k *Kubectl) GetMachineSets(ctx context.Context, machineDeploymentName string, cluster *types.Cluster) ([]clusterv1.MachineSet, error) {
```
g-gaston, line 503: nit: `GetMachineSetsForMachineDeployment` or `GetMachineSetsByMachineDeploymentName`

test/framework/taints.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/executables"
+)
+
+const ownerAnnotation = "cluster.x-k8s.io/owner-name"
+
+func (e *ClusterE2ETest) VaidateWorkerNodeTaints() {
```
g-gaston, line 17: No need to change, just starting the convo:
Should we make this more generic, like `VaidateWorkerNodes`? (I'm cool if that's divided into multiple sub validations)

The reason I'm suggesting this is because these mappings from eks-a to actual kube node objects is universal, independent from using taints or not. In fact these validations should pass even if all worker node groups lack taints. And once we have mapped eks-a api to nodes, adding more validations (labels, number of replicas...) becomes very easy. So I wonder if this is a validation worth running in all tests or at least other tests not taints related.

I know this discussion is not related to your work for taints and that's why I don't want this to be a blocker. But I think you wrote a very valuable piece of code and I would love for us to take advantage of it.

test/framework/taints.go
```diff
@@ -0,0 +1,92 @@
...
+					if ownerName == machineSet.Name {
+						if !v1alpha1.TaintsSliceEqual(node.Spec.Taints, w.Taints) {
+							e.T.Fatal(fmt.Errorf("taints on node %v and corresponding worker node group configuration %v do not match", node.Name, w.Name))
+						}
+					}
+				}
+			}
+		}
+	}
+	e.T.Log("validated that expected taints are present on the workload cluster nodes")
```
g-gaston, line 59: I'm gonna be that person, sorry
```suggestion
	e.T.Log("Validated that expected taints are present on the workload cluster nodes")
```

internal/pkg/api/cluster.go
g-gaston, line 177:
```suggestion
				logger.Info("Updating worker node group", "name", name)
```

internal/pkg/api/cluster.go
```diff
@@ -174,13 +174,15 @@ func WithWorkerNodeGroup(name string, fillers ...WorkerNodeGroupFiller) ClusterF
...
 				nodeGroup = &w
 				position = i
 				break
 			}
 		}
 
 		if nodeGroup == nil {
+			logger.Info("adding worker node group", "name", name)
```
g-gaston, line 185:
```suggestion
			logger.Info("Adding worker node group", "name", name)
```

g-gaston (commented):
Most comments are nits
But, only thing is, could we add some unit tests for the new kubectl methods?

test/framework/taints.go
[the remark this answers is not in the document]
jiayiwang7: ah i see

test/framework/taints.go
danbudris: totally. 

To confirm I understand -- when you say

> In fact these validations should pass even if all worker node groups lack taints

are you saying, we should be able to do a more universal node group validation that the wngc has created the proper nodes PERIOD (e.g. the node matches the wngc -- taints, labels, number present, etc)?  

Yes I think we should abstract this mapping out from this specific method to be re-used elsewhere. It was a pain to do and should be re-used. I'm going to do a bit of that when I push the CP taints e2e tests but we should move it forward more.

And also agree with you, and Joey mentioned this as well, that we should be running these kinds of validations in all our tests to confrim the nodes are right and not just isolate them to specific small sub-feautres (like taints or node lables)

pkg/executables/kubectl.go
danbudris: will change this when I add the unit tests

pkg/executables/kubectl.go
danbudris: will look at this when doing the unit tests

test/framework/taints.go
g-gaston: Yeah, exactly. Cool, let's continue this convo when you push the CP taints tests then.

PR #1237, opened 2022-02-17
Adding support for cloudstack bundles

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds the CloudStack bundles to the list of bundles. After adding the types to the `release` directory, I ran `make manifests` to update the CRD's under `config/crd/bases/anywhere.eks.amazonaws.com_bundles.yaml`.

I have not yet been able to determine if kubeproxy is required or not

*Testing (if applicable):*
Successfully created CloudStack cluster when using this change (along with others)


release/api/v1alpha1/bundle_types.go
```diff
@@ -196,6 +197,15 @@ type DockerBundle struct {
 	Metadata        Manifest `json:"metadata"`
 }
 
+type CloudStackBundle struct {
+	Version              string   `json:"version"`
+	ClusterAPIController Image    `json:"clusterAPIController"`
+	KubeProxy            Image    `json:"kubeProxy"`
+	Manager              Image    `json:"manager"`
```
abhay-krishna, line 204: I think we don't need the `Manager` here. We only have the ClusterAPIController. For vSphere we had the `cloud-provider-vsphere` which was the cloud-controller-manager component.

release/pkg/assets_capc.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	imageTagOverride := ImageTagOverride{
+		Repository: repoName,
+		ReleaseUri: imageArtifact.ReleaseImageURI,
+	}
+	imageTagOverrides = append(imageTagOverrides, imageTagOverride, kubeRbacProxyImageTagOverride)
+
+	manifestList := []string{
+		"infrastructure-components.yaml",
```
abhay-krishna, line 85: Don't we also need the cluster template manifest?

release/api/v1alpha1/bundle_types.go
maxdrib: Got it, will remove!

release/api/v1alpha1/bundle_types.go
```diff
@@ -66,6 +66,7 @@ type VersionsBundle struct {
 	ControlPlane           KubeadmControlPlaneBundle   `json:"controlPlane"`
 	Aws                    AwsBundle                   `json:"aws"`
 	VSphere                VSphereBundle               `json:"vSphere"`
+	CloudStack             CloudStackBundle            `json:"cloudStack"`
```
g-gaston, line 69:
```suggestion
	CloudStack             CloudStackBundle            `json:"cloudStack,omitempty"`
```
New fields should be optional to keep backwards compatibility :)

g-gaston (commented):
One minor comment about json tags
I would leave up to @abhay-krishna to give the final `lgtm` since he is the release code expert 🥇

abhay-krishna (approved):
/lgtm

PR #1238, opened 2022-02-17
Validate requested node provisioninging against available cr hardware

chrisdoherty4:
## Description
Add pre-flight validation to assert there are sufficient Tinkerbell CRs to satisfy the requested provisioning of control plane, etcd and worker nodes.

## Future work

There is a tight coupling between a few methods as the `ValidateHardwareConfig()` assumes responsibility for loading and parsing the hardware config that's required by `ValidateMinimumRequiredTinkerbellHardwareAvailable()`. This made it difficult to declare the `ValidateMinimumRequiredTinkerbellHardwareAvailable()` dependencies explicitly as parameters; we should really decoupled that to create much better testability.

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -179,6 +179,12 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context, 
 		return fmt.Errorf("failed setup and validations: %v", err)
 	}
 
+	// ValidateHardwareConfig performs a lazy load of hardware configuration. Given subsequent steps need the hardware
+	// read into memory it needs to be done first.
+	if err := p.validator.ValidateHardwareConfig(ctx, p.hardwareConfigFile); err != nil {
+		return err
+	}
+
 	tinkerbellClusterSpec := newSpec(clusterSpec, p.machineConfigs, p.datacenterConfig)
 
 	if err := p.validator.ValidateTinkerbellConfig(ctx, tinkerbellClusterSpec.datacenterConfig); err != nil {
@@ -187,6 +193,10 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context, 
 
 	if err := p.validator.ValidateClusterMachineConfigs(ctx, tinkerbellClusterSpec); err != nil {
 		return err
+	}
+
+	if err := p.validator.ValidateMinimumRequiredTinkerbellHardwareAvailable(tinkerbellClusterSpec.Spec.Cluster.Spec); err != nil {
+		return fmt.Errorf("minimum hardware not available: %v", err)
 	}
 
 	p.controlPlaneSshAuthKey = p.machineConfigs[p.clusterConfig.Spec.ControlPlaneConfiguration.MachineGroupRef.Name].Spec.Users[0].SshAuthorizedKeys[0]
@@ -199,10 +209,6 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context, 
 		}
 	} else {
 		logger.Info("Skipping check for whether control plane ip is in use")
-	}
-
-	if err := p.validator.ValidateHardwareConfig(ctx, p.hardwareConfigFile); err != nil {
-		return err
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


pkg/providers/tinkerbell/validator.go
```diff
@@ -5,12 +5,20 @@ import (
...
-	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/hardware"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 )
 
+// NodeMinimum defines the minimum number of nodes for the respective
+// node types.
+const (
```
chrisdoherty4, line 16: Need to remove this, it was part of an additional validation I'd done.

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -179,6 +179,12 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context,
 		return fmt.Errorf("failed setup and validations: %v", err)
 	}
 
+	// ValidateHardwareConfig performs a lazy load of hardware configuration. Given subsequent steps need the hardware
```
pokearu, line 182: That's a good catch 😄  
Will do a follow up PR to decouple this.

chrisdoherty4: I have some ideas around how you can do this that would make dependencies more explicit for each validation method if you'd like to chat.

pokearu: Definitely! If you already started on it feel free to raise a PR with what you had in mind 👍

pkg/providers/tinkerbell/validator_helper_test.go
```diff
@@ -0,0 +1,8 @@
+package tinkerbell
+
+import "github.com/aws/eks-anywhere/pkg/hardware"
+
+// SetValidatorHardwareConfig sets v internal hardwareConfig member to cfg.
+func SetValidatorHardwareConfig(v *Validator, cfg hardware.HardwareConfig) {
+	v.hardwareConfig = cfg
+}
```
abhinavmpandey08, line 8: i would vote to just do this inline instead of having a separate function for this

abhinavmpandey08 (commented):
awesome work! Just a few minor comments on convention and stuff

pkg/providers/tinkerbell/validator_helper_test.go
chrisdoherty4: If you look at the packages they're different - I've used the `_test` package idiom for the unit tests. This is a helper function used in testing only, because it shouldn't be a part of the package API, for hacking a `Validator` instance removing the need to construct the `Validator` in full. It exists largely because of a tight coupling between reading hardware configuration and one of the validate methods that lazily loads the hardware config. As a rule of thumb you don't want io occurring in unit tests if you can help it. 

In short, this can't be done in-line using the `_test` idiom which is best practice for unit testing with some small exceptions.

pkg/providers/tinkerbell/validator_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	tinkv1alpha1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/tink/api/v1alpha1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/hardware"
+	. "github.com/aws/eks-anywhere/pkg/providers/tinkerbell"
```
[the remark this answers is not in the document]
chrisdoherty4: Sure, I'm indifferent. Its common to use dot notation in unit tests when following the `_test` package idiom but it doesn't really matter which way we do it.

abhinavmpandey08 (approved):
/lgtm

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: Working on it. I found some other oddities so I might propose something a little more elaborate.

PR #1239, opened 2022-02-17
Add multiple-zone support in cloudstack - part 1 (pre-flight checking)

wanyufe:
*Add multiple-zone support for CloudStack provider*

*Description of changes:*
This is multiple-zone support part 1 which updates CloudStack datacenter config and updates cmk (CloudMonkey) pre-flight checking logic.

Also, domain and account are added as filter parameters when pre-flight checking CloudStack resources.
*Testing (if applicable):*


pkg/api/v1alpha1/cloudstackdatacenterconfig.go
```diff
@@ -18,13 +18,9 @@ func NewCloudStackDatacenterConfigGenerate(clusterName string) *CloudStackDatace
...
-			Zone: CloudStackResourceRef{
-				Value: "zone1",
-				Type:  "name",
-			},
-			Network: CloudStackResourceRef{
-				Value: "net1",
-				Type:  "name",
+			Zones: []Zone{
+				{ZoneRef: CloudStackResourceRef{Type: "name", Value: "zone1"}, Network: CloudStackResourceRef{Type: "name", Value: "GuestNet1"}},
+				{ZoneRef: CloudStackResourceRef{Type: "name", Value: "zone2"}, Network: CloudStackResourceRef{Type: "name", Value: "GuestNet2"}},
```
maxdrib, line 23: I feel like the default should only be one zone. Most customers would only use one, right?

pkg/executables/cmk.go
```diff
@@ -26,17 +27,71 @@ type Cmk struct {
...
+	zones []cmkZone
+	domain cmkDomain
+	account cmkAccount
+
 }
 
-// TODO: Add support for domain, account filtering
-func (c *Cmk) ValidateTemplatePresent(ctx context.Context, domain string, zone v1alpha1.CloudStackResourceRef, account string, template v1alpha1.CloudStackResourceRef) error {
-	var filterArg string
+func (c *Cmk) ValidateDatacenterConfig(ctx context.Context, datacenterConfig v1alpha1.CloudStackDatacenterConfig) error {
```
maxdrib, line 36: these functions introduce some business logic which should exist outside cmk, and instead live in cloudstack.go or validator.go. Are you just putting it here as a placeholder?

pkg/executables/cmk.go
```diff
@@ -91,11 +147,16 @@ func (c *Cmk) ValidateServiceOfferingPresent(ctx context.Context, domain string,
...
+func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domainId string, account string, affinityGroupIds []string) error {
+	var filterArgs []string
 	for _, affinityGroupId := range affinityGroupIds {
-		idFilterParam := fmt.Sprintf("id=\"%s\"", affinityGroupId)
-		result, err := c.exec(ctx, "list", "affinitygroups", idFilterParam)
+		filterArgs = []string{"list", "affinitygroups"}
+		filterArgs = append(filterArgs, fmt.Sprintf("id=\"%s\"", affinityGroupId))
+		if len(domainId) > 0 || len(account) > 0 {
+			filterArgs = append(filterArgs,  fmt.Sprintf("domainid=\"%s\"", domainId))
+			filterArgs = append(filterArgs,  fmt.Sprintf("account=\"%s\"", account))
```
maxdrib, line 157: I don't think this logic will work. Domain can be defined without an account specified, so we can't just always add the account filter

pkg/executables/cmk.go
```diff
@@ -119,39 +180,132 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+	}
+	if zone.Network.Type == v1alpha1.Id {
+		filterArgs = append(filterArgs, fmt.Sprintf("id=\"%s\"", zone.Network.Value))
+	}
+	if len(shared) > 0 {
+		filterArgs = append(filterArgs, fmt.Sprintf("type=\"%s\"", shared))
+	}
+	if len(domainId) > 0 || len(account) > 0 {
+		filterArgs = append(filterArgs, fmt.Sprintf("domainid=\"%s\"", domainId))
+		filterArgs = append(filterArgs, fmt.Sprintf("account=\"%s\"", account))
```
maxdrib, line 234: here as well

pkg/executables/cmk.go
```diff
@@ -119,39 +180,132 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+	if err = json.Unmarshal(result.Bytes(), &response); err != nil {
+		return fmt.Errorf("failed to parse response into json: %v", err)
+	}
+	domains := response.CmkDomains
+	if len(domains) > 1 {
+		return fmt.Errorf("duplicate domain %s found", domain)
+	} else if len(domains) == 0 {
+		return fmt.Errorf("domain %s not found", domain)
+	}
+	c.domain = cmkDomain{Id: domains[0].Id, Name: domains[0].Name}
```
maxdrib, line 303: this is clever, I like it

pkg/executables/cmk.go
wanyufe: yes, as a place holder since our cloudstack.go and validator have not been merged yet. This logic will be moved to them later.

pkg/executables/cmk.go
wanyufe: good point. will update with a comment to explain the situation.

pkg/api/v1alpha1/cloudstackdatacenterconfig.go
wanyufe: I agree that single zone should be the common case, the point here is to let customer know multiple zone is supported. I also recognize that the array grammar in yaml indicates it support multiple zones.

pkg/executables/cmk.go
chrisdoherty4: For something like this you might want to declare your intention explicitly by creating a `to_be_moved.go` and just using functions rather than polluting the `Cmk` struct.

wanyufe: I removed these two functions for this PR since they are not used for now. It will add back in coming commit when CloudMonkey wrapper is added.

pkg/api/v1alpha1/cloudstackdatacenterconfig_test.go
```diff
@@ -39,14 +39,24 @@ func TestGetCloudStackDatacenterConfig(t *testing.T) {
...
-						Value: "zone1",
-						Type:  Name,
-					},
 					Account: "admin",
-					Network: CloudStackResourceRef{
-						Value: "net1",
-						Type:  Name,
+					Zones: []Zone{
+						{
+							ZoneRef: CloudStackResourceRef{
```
maxdrib, line 45: spacing is off here. Have you run `make lint`?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -50,6 +48,16 @@ const (
 	Name CloudStackResourceRefType = "Name"
 )
 
+// Zone is an organizational construct typically used to represent a single datacenter, and all its physical and virtual resources exist inside that zone. It can either be specified as a UUID or name
+type Zone struct {
```
maxdrib, line 52: Since this is in the v1alpha1 package, it would make sense to add the CloudStack prefix to the type name

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -50,6 +48,16 @@ const (
...
+	// ZoneRef is the name or UUID of the CloudStack zone in which clusters should be created. Zones should be managed by a single CloudStack Management endpoint.
+	ZoneRef CloudStackResourceRef `json:"zone"`
```
maxdrib, line 54: Adding "ref" suffix here makes it inconsistent I think. If we add the prefix to Zone above, then we can change ZoneRef -> Zone

pkg/executables/cmk.go
```diff
@@ -200,7 +318,12 @@ func (c *Cmk) exec(ctx context.Context, args ...string) (stdout bytes.Buffer, er
 	if err != nil {
 		return bytes.Buffer{}, fmt.Errorf("failed cmk validations: %v", err)
 	}
-	argsWithConfigFile := append([]string{"-c", configFile}, args...)
+	argsWithConfigFile := []string{"-c", configFile}
+	for _, arg := range args {
```
maxdrib, line 322: what's the point of filtering out the spaces?

pkg/executables/cmk_test.go
```diff
@@ -374,6 +483,10 @@ func TestCmkListOperations(t *testing.T) {
 			defer tctx.RestoreContext()
 
 			executable := mockexecutables.NewMockExecutable(mockCtrl)
+			//argumentsExecCall := make([]interface{}, len(tt.argumentsExecCall))
```
maxdrib, line 486: let's remove this if it's unused

maxdrib (changes requested):
Could you check to see why unit tests are failing?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
g-gaston, line 52: Should we find a different name for this struct?
It seems with this like a zone contains a zone ref and a network
That makes me think that `Zone` should be something else
What could be the construct that groups a zone and a network?
```go
type SomethingElse struct {
  Zone CloudStackResourceRef
  Network CloudStackResourceRef
}
```

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -86,10 +87,10 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 		)
 	}
 
-	if old.Spec.Zone != new.Spec.Zone {
+	if !reflect.DeepEqual(old.Spec.Zones, new.Spec.Zones) {
```
g-gaston, line 90: We have been trying to avoid using deep equal aside from in tests
I would suggest creating an `Equal` to compare `[]Zone`
I think there already other api structs that implement this pattern

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook_test.go
```diff
@@ -21,15 +21,25 @@ func TestCloudStackDatacenterValidateUpdateDomainImmutable(t *testing.T) {
 
 func TestCloudStackDatacenterValidateUpdateZoneImmutable(t *testing.T) {
```
g-gaston, line 22:
```suggestion
func TestCloudStackDatacenterValidateUpdateZonesImmutable(t *testing.T) {
```

pkg/executables/cmk.go
```diff
@@ -26,17 +27,26 @@ type Cmk struct {
 	writer     filewriter.FileWriter
 	executable Executable
 	config     CmkExecConfig
+	zones []cmkZone
+	domain cmkDomain
+	account cmkAccount
```
g-gaston, line 32: Why are we storing these in the cmk struct?

pkg/executables/cmk_test.go
```diff
@@ -30,9 +32,11 @@ var execConfig = executables.CmkExecConfig{
...
-	Type:  v1alpha1.Name,
-	Value: "zone1",
+var zones = []v1alpha1.Zone{
+	{ZoneRef: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Name, Value: "TEST_RESOURCE"}, Network: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Name, Value: "TEST_RESOURCE"}},
+	{ZoneRef: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Name, Value: "TEST_RESOURCE"}, Network: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Id, Value: "TEST_RESOURCE"}},
+	{ZoneRef: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Id, Value: "TEST_RESOURCE"}, Network: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Name, Value: "TEST_RESOURCE"}},
+	{ZoneRef: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Id, Value: "TEST_RESOURCE"}, Network: v1alpha1.CloudStackResourceRef{Type: v1alpha1.Id, Value: "TEST_RESOURCE"}},
 }
 
 var resourceName = v1alpha1.CloudStackResourceRef{
```
g-gaston, line 42: Could we add tests for the new methods?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -28,12 +28,10 @@ type CloudStackDatacenterConfigSpec struct {
 
 	// Domain contains a grouping of accounts. Domains usually contain multiple accounts that have some logical relationship to each other and a set of delegated administrators with some authority over the domain and its subdomains
 	Domain string `json:"domain"`
-	// Zone is an organizational construct typically used to represent a single datacenter, and all its physical and virtual resources exist inside that zone. It can either be specified as a UUID or name
-	Zone CloudStackResourceRef `json:"zone"`
+	// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+	Zones []Zone `json:"zones"`
```
vivek-koppuru, line 32: If we are changing this to multiple zones, should we introduce a new type called `CloudStackMultiDatacenterConfig` or `CloudStackMultiZoneDatacenterConfig`?

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook_test.go
```diff
@@ -21,15 +21,25 @@ func TestCloudStackDatacenterValidateUpdateDomainImmutable(t *testing.T) {
...
-		Type:  v1alpha1.Name,
-		Value: "oldCruftyZone",
+	vOld.Spec.Zones = []v1alpha1.Zone{
+		{ZoneRef: v1alpha1.CloudStackResourceRef{
+			Type:  v1alpha1.Name,
+			Value: "oldCruftyZone"},
+		 Network: v1alpha1.CloudStackResourceRef{
+			Type: v1alpha1.Name,
+			Value: "GuestNet1"},
+		},
```
vivek-koppuru, line 31: nit
```suggestion
	vOld.Spec.Zones = []v1alpha1.Zone{{
		ZoneRef: v1alpha1.CloudStackResourceRef{
			Type:  v1alpha1.Name,
			Value: "oldCruftyZone"},
		 Network: v1alpha1.CloudStackResourceRef{
			Type: v1alpha1.Name,
			Value: "GuestNet1",
		}},
```

pkg/executables/cmk.go
```diff
@@ -26,17 +27,26 @@ type Cmk struct {
...
+
 }
 
-// TODO: Add support for domain, account filtering
-func (c *Cmk) ValidateTemplatePresent(ctx context.Context, domain string, zone v1alpha1.CloudStackResourceRef, account string, template v1alpha1.CloudStackResourceRef) error {
-	var filterArg string
+func (c *Cmk) ValidateTemplatePresent(ctx context.Context, domainId string, zoneId string, account string, template v1alpha1.CloudStackResourceRef) error {
+	filterArgs := []string{"list", "templates", "templatefilter=all", "listall=true"}
```
vivek-koppuru, line 37: This gives the impression that `list` and `templates` are filter args too. Should we have `commandArgs` and `filterArgs` as separate variables?

pkg/executables/cmk.go
```diff
@@ -119,39 +139,136 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+	filterArgs := []string{"list", "networks"}
+	shared := ""
+	if multipleZone {
+		shared = "Shared"
+	}
+	if zone.Network.Type == v1alpha1.Id {
+		filterArgs = append(filterArgs, fmt.Sprintf("id=\"%s\"", zone.Network.Value))
+	}
+	if len(shared) > 0 {
+		filterArgs = append(filterArgs, fmt.Sprintf("type=\"%s\"", shared))
```
vivek-koppuru, line 189: This can go in the `multipleZone` if condition above right?

pkg/api/v1alpha1/cloudstackdatacenterconfig.go
```diff
@@ -18,13 +18,9 @@ func NewCloudStackDatacenterConfigGenerate(clusterName string) *CloudStackDatace
...
 		Spec: CloudStackDatacenterConfigSpec{
 			Domain: "domain1",
-			Zone: CloudStackResourceRef{
-				Value: "zone1",
-				Type:  "name",
-			},
-			Network: CloudStackResourceRef{
-				Value: "net1",
-				Type:  "name",
+			Zones: []Zone{
```
maxdrib, line 21: Following up on our conversation earlier, how will multi-zone interact with WorkerNodeGroup in terms of worker node placement?

pkg/api/v1alpha1/cloudstackdatacenterconfig.go
chrisdoherty4: Keep it simple, multiple zones is probably unnecessary. Its just more for the end-user to have to delete/think about.

pkg/api/v1alpha1/cloudstackdatacenterconfig_test.go
```diff
@@ -39,14 +39,24 @@ func TestGetCloudStackDatacenterConfig(t *testing.T) {
...
+								Value: "zone1",
+								Type:  Name,
+						}, Network: CloudStackResourceRef{
+								Value: "net1",
+								Type:  Name,
+						},},
```
chrisdoherty4, line 51: Blip?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
chrisdoherty4: How do you see that being advantageous vs this approach? Thats a little loaded because I can't see the value and would vote KISS.

pkg/executables/cmk.go
```diff
@@ -119,39 +139,136 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
 	return nil
 }
 
-func (c *Cmk) ValidateZonePresent(ctx context.Context, zone v1alpha1.CloudStackResourceRef) error {
+func (c *Cmk) ValidateZonesPresent(ctx context.Context, zones []v1alpha1.Zone) error {
```
chrisdoherty4, line 142: This signature seems off. Why does the `Cmk` executable care about multiple zones? I would think the fact there are multiple zones that need validating is a property of some validation logic and this signature is just for 1 zone reducing its scope.

```go
func (c *Cmk) ZoneExists(ctx context.Context, zone v1alpha1.Zone) error
```

pkg/executables/cmk.go
```diff
@@ -119,39 +139,136 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
 	var filterArg string
-	if zone.Type == v1alpha1.Id {
-		filterArg = fmt.Sprintf("id=\"%s\"", zone.Value)
+
+	for _, z := range zones {
```
chrisdoherty4, line 145: [Ditto](https://github.com/aws/eks-anywhere/pull/1239/files#r816833565)

pkg/executables/cmk.go
```diff
@@ -119,39 +139,136 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+			return fmt.Errorf("zone %s not found", filterArg)
+		} else {
+			c.zones = append(c.zones, cmkZones[0])
+		}
+	}
+
+	return nil
+}
+
+func (c *Cmk) ValidateNetworkPresent(ctx context.Context, domainId string, zone v1alpha1.Zone, account string, multipleZone bool) error {
```
chrisdoherty4, line 179: Is it possible to specify multiple zones with different networks?

pkg/executables/cmk.go
chrisdoherty4: OK, I've just noticed there are a bunch of existing `Validate*` methods on this struct. Whats the intent behind this struct because it contains both rules on whats OK and not OK as well as information retrieval from CloudStack so conflates several concerns.

@g-gaston is this on your radar?

pkg/executables/cmk.go
```diff
@@ -119,39 +135,147 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+	if zoneRef.Zone.Type == v1alpha1.Id {
+		zoneId = zoneRef.Zone.Value
 	} else {
-		filterArg = fmt.Sprintf("name=\"%s\"", zone.Value)
+		for _, zoneIdentifier := range zones {
+			if zoneRef.Zone.Value == zoneIdentifier.Name {
+				zoneId = zoneIdentifier.Id
+				break
+			}
+		}
```
maxdrib, line 232: it seems like the retrieval of zoneID should be extracted into a separate function in case we need to do it elsewhere

pkg/executables/cmk.go
```diff
@@ -119,39 +135,147 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
 	}
 	return nil
 }
 
-// TODO: Add support for domain filtering
-func (c *Cmk) ValidateAccountPresent(ctx context.Context, account string) error {
-	result, err := c.exec(ctx, "list", "accounts", fmt.Sprintf("name=\"%s\"", account))
+func (c *Cmk) ValidateAccountPresent(ctx context.Context, account string, domainId string) error {
+	result, err := c.exec(ctx, "list", "accounts", fmt.Sprintf("name=\"%s\"", account), fmt.Sprintf("domainid=\"%s\"", domainId))
```
maxdrib, line 278: what happens if domain isn't set? Does that break this function/cmk? Or is it required in the cluster spec?

pkg/providers/cloudstack/validator.go
```diff
@@ -26,27 +26,54 @@ func NewValidator(cmk ProviderCmkClient) *Validator {
...
+	ValidateZonesPresent(ctx context.Context, zones []anywherev1.CloudStackZoneRef) ([]anywherev1.CloudStackResourceIdentifier, error)
+	ValidateNetworkPresent(ctx context.Context, domainId string, zoneRef anywherev1.CloudStackZoneRef, zones []anywherev1.CloudStackResourceIdentifier, account string, multipleZone bool) error
+	ValidateDomainPresent(ctx context.Context, domain string) (anywherev1.CloudStackResourceIdentifier, error)
+	ValidateAccountPresent(ctx context.Context, account string, domainId string) error
 }
 
 func (v *Validator) validateCloudStackAccess(ctx context.Context) error {
 	if err := v.cmk.ValidateCloudStackConnection(ctx); err != nil {
-		return fmt.Errorf("failed validating connection to vCenter: %v", err)
+		return fmt.Errorf("failed validating connection to cloudStack: %v", err)
```
maxdrib, line 40: nice catch

pkg/providers/cloudstack/validator.go
```diff
@@ -148,17 +166,32 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, cloudStac
...
+	}
+	domain, errDomain := v.cmk.ValidateDomainPresent(ctx, datacenterConfigSpec.Domain)
+	if errDomain != nil {
+		return fmt.Errorf("error while checking domain %v", errDomain)
+	}
+
+	zones, err := v.cmk.ValidateZonesPresent(ctx, datacenterConfigSpec.Zones)
+	if err != nil {
+		return fmt.Errorf("error while checking zones %v", err)
 	}
```
maxdrib, line 180: these validations are on the datacenter config, not the machine config. Shouldn't they be handled elsewhere, not here?

pkg/providers/cloudstack/validator.go
```diff
@@ -26,27 +26,54 @@ func NewValidator(cmk ProviderCmkClient) *Validator {
...
-		return err
+	if len(datacenterConfig.Spec.Domain) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
+	}
+	domain, errDomain := v.cmk.ValidateDomainPresent(ctx, datacenterConfig.Spec.Domain)
+	if errDomain != nil {
+		return fmt.Errorf("error while checking domain %v", errDomain)
+	}
+
+	err := v.cmk.ValidateAccountPresent(ctx, datacenterConfig.Spec.Account, domain.Id)
```
maxdrib, line 56: we should check to make sure account is set before doing this validation

pkg/providers/cloudstack/validator.go
```diff
@@ -26,27 +26,54 @@ func NewValidator(cmk ProviderCmkClient) *Validator {
...
 	return nil
 }
 
 func (v *Validator) ValidateCloudStackDatacenterConfig(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
-	if err := v.cmk.ValidateZonePresent(ctx, datacenterConfig.Spec.Zone); err != nil {
-		return err
+	if len(datacenterConfig.Spec.Domain) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
+	}
+	domain, errDomain := v.cmk.ValidateDomainPresent(ctx, datacenterConfig.Spec.Domain)
```
maxdrib, line 51: is domain required? Maybe we should check to see if domain is set before doing this check

maxdrib (commented):
minor comments, overall approach lgtm

pkg/providers/cloudstack/validator.go
wanyufe: yes, domain is required per cloudstackdatacenterconfig_types.go file.

maxdrib (approved):

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
vivek-koppuru: I guess what I am confused by is that initially we went with the naming `CloudStackDatacenterConfig` because a zone is kind of considered a single datacenter when compared to vsphere. So I thought the plan was that when we supported multiple zones, it would be a different type of deployment. But I guess we are going to treat one management endpoint as a datacenter instead of one zone as a datacenter? @maxdrib

pkg/api/v1alpha1/cloudstackdatacenterconfig_test.go
wanyufe: run gofumpt after make lint, should be good now.

pkg/executables/cmk.go
wanyufe: removed. 

The reason it was added there is for unit testing which compares command args with expected values.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
wanyufe: refactored.

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
wanyufe: Equal could not be applied directly to []Zone, I added one to apply Zone

pkg/executables/cmk.go
wanyufe: removed/refactored after merging with latest main commits

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
wanyufe: yes, one cloudstack management endpoint is a datacenter.

pkg/executables/cmk.go
wanyufe: separated into commandArgs and filterArgs.

pkg/executables/cmk.go
wanyufe: yes, each zone can have its network which can be different with network from other zones.

pkg/executables/cmk.go
wanyufe: @chrisdoherty4 I agree with your concerns. 
CMK should only provider some functions to call cloudstack to retrieve some info back and should not do extra, like validation. 
The validation should be inside validator.go file. 

I am thinking do the refactor in next PR since Max and I are both touching this cmk file in on-going PRs.

pkg/providers/cloudstack/validator.go
vivek-koppuru: I am a bit confused how unit testing is the reason why we are adding datacenter validation here. We can always ensure the order of testing all of validate if that is an issue instead of do it this way as it is effectively the same thing.

pkg/executables/cmk.go
```diff
@@ -119,39 +139,156 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+}
+
+func (c *Cmk) ValidateNetworkPresent(ctx context.Context, domainId string, zoneRef v1alpha1.CloudStackZoneRef, zones []v1alpha1.CloudStackResourceIdentifier, account string, multipleZone bool) error {
+	commandArgs := []string{"list", "networks"}
+	var filterArgs []string
+	if zoneRef.Network.Type == v1alpha1.Id {
+		filterArgs = append(filterArgs, fmt.Sprintf("id=\"%s\"", zoneRef.Network.Value))
+	}
+	if multipleZone {
+		filterArgs = append(filterArgs, fmt.Sprintf("type=\"%s\"", "Shared"))
```
chrisdoherty4, line 217: Concatenating string literal with string literal?

pkg/executables/cmk.go
```diff
@@ -119,39 +139,156 @@ func (c *Cmk) ValidateAffinityGroupsPresent(ctx context.Context, domain string,
...
+	commandArgs = append(commandArgs, filterArgs...)
+	result, err := c.exec(ctx, commandArgs...)
 	if err != nil {
-		return fmt.Errorf("error getting zones info: %v", err)
+		return fmt.Errorf("error getting network info: %v", err)
 	}
 	if result.Len() == 0 {
-		return fmt.Errorf("zone %s not found", zone)
+		if multipleZone {
+			return fmt.Errorf("%s network %s not found in zoneRef %s", "Shared", zoneRef.Network.Value, zoneRef.Zone.Value)
```
chrisdoherty4, line 245: [Ditto](https://github.com/aws/eks-anywhere/pull/1239/files#r819665207)

pkg/executables/cmk.go
chrisdoherty4: In-fact, "Shared" should be a constant. It has semantics in the context of cloud monkey so deserves dedicated constant.

pkg/providers/cloudstack/validator.go
```diff
@@ -26,27 +26,56 @@ func NewValidator(cmk ProviderCmkClient) *Validator {
...
+	if errDomain != nil {
+		return fmt.Errorf("error while checking domain %v", errDomain)
+	}
+
+	if len(datacenterConfig.Spec.Account) > 0 {
+		err := v.cmk.ValidateAccountPresent(ctx, datacenterConfig.Spec.Account, domain.Id)
+		if err != nil {
+			return fmt.Errorf("error while checking account %v", err)
```
chrisdoherty4, line 59: No need to prepend 'error' to errors. Thats done a lot in these checks, can we strip that out.

```
invalid account: %v
```

pkg/providers/cloudstack/validator.go
```diff
@@ -148,17 +168,32 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, cloudStac
 }
 
 func (v *Validator) validateMachineConfig(ctx context.Context, datacenterConfigSpec anywherev1.CloudStackDatacenterConfigSpec, machineConfig *anywherev1.CloudStackMachineConfig) error {
-	domain := datacenterConfigSpec.Domain
-	zone := datacenterConfigSpec.Zone
-	account := datacenterConfigSpec.Account
-	if err := v.cmk.ValidateTemplatePresent(ctx, domain, zone, account, machineConfig.Spec.Template); err != nil {
-		return fmt.Errorf("validating template: %v", err)
+	if len(datacenterConfigSpec.Domain) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
```
chrisdoherty4, line 172: If there's no formatting to be done on the string just use `errors.New()`. `fmt.Errorf()` invokes a bunch of formatting logic.

pkg/providers/cloudstack/validator.go
```diff
@@ -26,27 +26,56 @@ func NewValidator(cmk ProviderCmkClient) *Validator {
...
 	logger.MarkPass("Connected to server")
 
 	return nil
 }
 
 func (v *Validator) ValidateCloudStackDatacenterConfig(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
-	if err := v.cmk.ValidateZonePresent(ctx, datacenterConfig.Spec.Zone); err != nil {
-		return err
+	if len(datacenterConfig.Spec.Domain) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
```
chrisdoherty4, line 49: [Ditto](https://github.com/aws/eks-anywhere/pull/1239/files#r819772280)

chrisdoherty4 (approved):
Change the `cmkCommandArgs.go` to snake case and you're good to go.

vignesh-goutham:
Thanks for reviewing everyone!

/lgtm

PR #1244, opened 2022-02-18
create tinkerbell template override struct

gwesterfieldjr:
Create a new type to represent TemplateOverride

Build and Unit test


cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -170,8 +170,14 @@ func generateClusterConfig(clusterName string) error {
 			}
 			datacenterYaml = dcyaml
 
-			cpMachineConfig := v1alpha1.NewTinkerbellMachineConfigGenerate(clusterName + "-cp")
-			workerMachineConfig := v1alpha1.NewTinkerbellMachineConfigGenerate(clusterName)
+			cpMachineConfig, err := v1alpha1.NewTinkerbellMachineConfigGenerate(clusterName + "-cp")
```
chrisdoherty4, line 173: The `clusterName + "-cp"` really ought to be in a function somewhere, probably not in this package, so it can be recreated as needed (testing etc).

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -170,8 +170,14 @@ func generateClusterConfig(clusterName string) error {
...
+			if err != nil {
+				return fmt.Errorf("error outputting yaml: %v", err)
```
chrisdoherty4, line 175: There's no need to add 'error' to error messages. As errors are wrapped with increasing amounts of context you end up with lots of 'error' and we fundamentally know its an error because of how it gets presented to the end user (e.g. `error exuting cmd: error with hardware yaml: error writing file: error file already exists`)

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -0,0 +1,197 @@
...
+						"BLOCK_DEVICE": "/dev/sda1",
+						"FS_TYPE":      "ext4",
+					},
+				},
+			},
+		}},
+	}
+}
+
+func (c *TinkerbellMachineConfigGenerate) APIVersion() string {
```
chrisdoherty4, line 162: Do you think we need this? `config.TypeMeta.APIVersion` isn't particularly verbose. Ditto for the other accessor methods.

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -0,0 +1,197 @@
...
+				Name:              "ec2-user",
+				SshAuthorizedKeys: []string{"ssh-rsa AAAA..."},
+			}},
+			TemplateOverride: templateOverride,
+		},
+	}, nil
+}
+
+func GetTinkerbellDefaultTemplateOverride(name string) *TinkerbellTemplate {
+	return &TinkerbellTemplate{
```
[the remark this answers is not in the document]
gwesterfieldjr: Good question, I wondered the same thing for all the string literals. Rest of team will need to chime in as I don't have the context.

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
gwesterfieldjr: Do you recommend which package to put this in?

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
gwesterfieldjr: I'll clean all of these up

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
chrisdoherty4: Having skimmed the relevant code, the function would be relevant to all providers so `/pkg/providers` seems appropriate. I would then substitute the vsphere naming with your new function also :)

gwesterfieldjr: Yeah and I'll create one for etcd as well

pokearu (approved):
/lgtm
/woof

PR #1249, opened 2022-02-21
Connectivity and auth credentials check for BMC IPMI

panktishah26:
*Issue #, if available:*

*Description of changes:*
This PR is to add functionality to check connectivity and auth credentials for BMC IPMI. This is the part of pre-flight validation for create cluster using Tinkerbell provider.

*Testing (if applicable):*


pkg/dependencies/factory.go
```diff
@@ -31,6 +32,7 @@ type Dependencies struct {
 	Kubectl                   *executables.Kubectl
 	Govc                      *executables.Govc
 	Tink                      *executables.Tink
+	PBNJ                      *pbnj.Pbnj
```
abhinavmpandey08, line 35: nit: I see `Pbnj` and `PBNJ` throughout the PR. Let's stick to CamelCase for consistency
```suggestion
	Pbnj                      *pbnj.Pbnj
```

pkg/dependencies/factory.go
```diff
@@ -266,6 +269,9 @@ func (f *Factory) WithTink(clusterConfigFile string) *Factory {
 		}
 		f.dependencies.Tink = f.executableBuilder.BuildTinkExecutable(tinkerbellDatacenterConfig.Spec.TinkerbellCertURL, tinkerbellDatacenterConfig.Spec.TinkerbellGRPCAuth)
 
+		pbnjClient, _ := pbnj.NewPBNJClient(tinkerbellDatacenterConfig.Spec.TinkerbellPBnJGRPCAuth)
```
abhinavmpandey08, line 272: why ignore the error here?

pkg/hardware/config.go
```diff
@@ -22,6 +22,13 @@ type HardwareConfig struct {
 	Secrets   []corev1.Secret
 }
 
+type BmcSecretConfig struct {
+	Host     string
+	Username string
+	Password string
+	Vendor   string
+}
```
abhinavmpandey08, line 30: wouldn't this struct make more sense in pbnj's client.go since it's being used in that file but not in this?

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,60 @@
+package pbnj
```
abhinavmpandey08, line 1: Correct me if im wrong, but didn't we decide to keep it under `pkg/clients/pbnj`?

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,60 @@
...
+func NewPBNJClient(pbnjGrpcAuth string) (*Pbnj, error) {
+	os.Setenv(PbnjGrpcAuth, pbnjGrpcAuth)
+
+	conn, _ := pbnjClient.SetupConnection()
+
+	mClient := pbnjv1.NewMachineClient(conn)
+	tClient := pbnjv1.NewTaskClient(conn)
+	pbnjObj := &Pbnj{pbnjClient.NewPbnjClient(mClient, tClient)}
+
+	os.Unsetenv(PbnjGrpcAuth)
```
abhinavmpandey08, line 30: why is this needed?

pkg/providers/tinkerbell/validator.go
```diff
@@ -132,7 +134,28 @@ func (v *Validator) ValidateHardwareConfig(ctx context.Context, hardwareConfigFi
...
+func (v *Validator) ValidateBMCSecretCreds(ctx context.Context, hc hardware.HardwareConfig) error {
+	for index, bmc := range hc.Bmcs {
+		bmcInfo := hardware.BmcSecretConfig{
+			Host:     bmc.Spec.Host,
+			Username: string(hc.Secrets[index].Data["username"]),
+			Password: string(hc.Secrets[index].Data["password"]),
+			Vendor:   bmc.Spec.Vendor,
+		}
+		if err := v.pbnj.ValidateBMCSecretCreds(ctx, bmcInfo); err != nil {
+			return fmt.Errorf("failed validating Hardware BMC credentials for the node %v, host %v : %v", hc.Hardwares[index].Name, bmcInfo.Host, err)
```
abhinavmpandey08, line 155: nit: use `%s` format specifier for strings
```suggestion
			return fmt.Errorf("failed validating Hardware BMC credentials for the node %s, host %s : %v", hc.Hardwares[index].Name, bmcInfo.Host, err)
```

pkg/providers/tinkerbell/pbnj/client.go
pokearu: Actually that's needed because SetupConnection looks for `PBNJ_GRPC_AUTHORITY` env var for getting the IP and port to connect to.

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,60 @@
+package pbnj
+
+import (
+	"context"
+	"os"
+
+	pbnjClient "github.com/tinkerbell/cluster-api-provider-tinkerbell/pbnj/client"
+	pbnjv1 "github.com/tinkerbell/pbnj/api/v1"
+
```
pokearu, line 9: nit: I think we can remove the names, `pbnjClient` and `pbnjv1` for the imports.
Names make sense when there are imports with the same name. 
Here we could directly do `client.` and `v1.`

pkg/hardware/config.go
pokearu: I agree. This is actually only used as a function parameter input for `ValidateBMCSecretCreds` so it can be in the pbnj package.

pkg/providers/tinkerbell/validator.go
```diff
@@ -132,7 +134,28 @@ func (v *Validator) ValidateHardwareConfig(ctx context.Context, hardwareConfigFi
 		return fmt.Errorf("failed validating Secrets in hardware config: %v", err)
 	}
 
-	logger.MarkPass("Hardware Config is valid")
+	if err := v.ValidateBMCSecretCreds(ctx, v.hardwareConfig); err != nil {
```
pokearu, line 137: I feel this can be its own validation, inside `SetupAndValidateCreateCluster`. Yes, we are using information from hardwareConfig but we are also doing API calls and connectivity checks so I think it can be moved out of this function.

pkg/providers/tinkerbell/pbnj/client.go
pokearu: I am fine either way.
But i think we decided this might be specific to tinkerbell provider and not a more generic "client" and hence we wanted to keep it in providers package?

pokearu:
Looks good 👍 
I had a question, I remember we had discussed to wrap `TinkerbellTinkClient` and `TinkerbellPbnjClient` into a common struct so we could limit the number of individual tinkerbell dependencies being injected into the provider?
And since we use `WithTink` to build both the dependencies would that make more sense?

pkg/providers/tinkerbell/pbnj/client.go
panktishah26: yeah, I am also fine either way but I thought we decided to put it under Tinkerbell provider considering it is only for Tinkerbell.

pkg/dependencies/factory.go
```diff
@@ -253,19 +255,21 @@ func (f *Factory) WithGovc() *Factory {
...
-			return nil
-		}
 		tinkerbellDatacenterConfig, err := v1alpha1.GetTinkerbellDatacenterConfig(clusterConfigFile)
 		if err != nil {
 			return err
 		}
-		f.dependencies.Tink = f.executableBuilder.BuildTinkExecutable(tinkerbellDatacenterConfig.Spec.TinkerbellCertURL, tinkerbellDatacenterConfig.Spec.TinkerbellGRPCAuth)
+		tinkClient := f.executableBuilder.BuildTinkExecutable(tinkerbellDatacenterConfig.Spec.TinkerbellCertURL, tinkerbellDatacenterConfig.Spec.TinkerbellGRPCAuth)
 
+		pbnjClient, err := pbnj.NewPBNJClient(tinkerbellDatacenterConfig.Spec.TinkerbellPBnJGRPCAuth)
```
chrisdoherty4, line 268: I think this should be broken down into atomic constructs, i.e. 1 for the tink executable and 1 for pbnj. At present it forces consumers to create both even if they don't want them. 

I'd even like to propose a way to get away from this construct all together as its basically got knowledge of how to construct everything in the program which is a lot of responsibility.

pkg/providers/tinkerbell/pbnj/client.go
chrisdoherty4: The `client` package is a bit of an organizational thing and not strictly necessary. The `pbnj` package provides constructs and functions related to pbnj within the context of tinkerbell so its totally appropriate to have it here.

pkg/dependencies/factory.go
pokearu: Actually it was originally atomic, but we wanted to make it combined as they are tinkerbell specific dependencies and would be needed together.
Also, hopefully in the near future we can eliminate the need for these 2 clients as everything would become `kubectl` driven.

And I think there have been discussions on maybe refactoring this approach, like reducing responsibility of the dependencies factory as its becoming hard to maintain. We could discuss on that more :)

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -75,21 +81,32 @@ type ProviderTinkClient interface {
...
+		providerTinkbellClient.ProviderPbnjClient,
 		&networkutils.DefaultNetClient{},
 		now,
 		skipIpCheck,
 		hardwareConfigFile,
 	)
 }
 
-func NewProviderCustomNet(datacenterConfig *v1alpha1.TinkerbellDatacenterConfig, machineConfigs map[string]*v1alpha1.TinkerbellMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, providerTinkClient ProviderTinkClient, netClient networkutils.NetClient, now types.NowFunc, skipIpCheck bool, hardwareConfigFile string) *tinkerbellProvider {
+func NewProviderCustomNet(datacenterConfig *v1alpha1.TinkerbellDatacenterConfig, machineConfigs map[string]*v1alpha1.TinkerbellMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, providerTinkClient ProviderTinkClient, pbnjClient ProviderPbnjClient, netClient networkutils.NetClient, now types.NowFunc, skipIpCheck bool, hardwareConfigFile string) *tinkerbellProvider {
```
pokearu, line 109: nit: Can we rename this `NewProviderCustomNet`? I think it was orginally named this because the only additional dependency injected was `networkutils.NetClient`. 
We now inject a few more.

pkg/dependencies/factory.go
chrisdoherty4: Glad you chimed in, I saw your other comment. Whether or not the provider accepts individual parameters or a single struct for clients is orthogonal to how they're constructed. I.e. you can construct the individual clients as separate entities, then combine them when the provider is created. 

If I paint a different picture for you, there could be consuming constructs that want _only_ a Tink client or _only_ a PBnJ client. Having individual constructor functions with a _single responsibility_ enables that world. The current approach here _couples_ creating a Tink client with a PBnJ client when they are _logically distinct_ and 1 doesn't strictly need the other.

chrisdoherty4: Re refactoring of `dependencies`, yes that sounds good - lets chat at some point.

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -120,7 +137,7 @@ func NewProviderCustomNet(datacenterConfig *v1alpha1.TinkerbellDatacenterConfig,
 			now:                         now,
 		},
 		hardwareConfigFile: hardwareConfigFile,
-		validator:          NewValidator(providerTinkClient, netClient, hardware.HardwareConfig{}),
+		validator:          NewValidator(providerTinkClient, netClient, hardware.HardwareConfig{}, pbnjClient),
```
chrisdoherty4, line 140: Why is the validator constructed here and not injected?

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: This function has too many clients being injected. It suggests the construct it returns, given this is an initializing constructor func, has too much responsibility. Have we thought about how we can break the provider down?

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,65 @@
...
+	tClient := v1.NewTaskClient(conn)
+	pbnjObj := &Pbnj{client.NewPbnjClient(mClient, tClient)}
+
+	os.Unsetenv(PbnjGrpcAuth)
+
+	return pbnjObj, nil
+}
+
+func (p *Pbnj) ValidateBMCSecretCreds(ctx context.Context, bmcInfo BmcSecretConfig) error {
+	powerRequest := &v1.PowerRequest{
```
chrisdoherty4, line 41: This construction could do with living in its own function,

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,65 @@
...
+func NewPBNJClient(pbnjGrpcAuth string) (*Pbnj, error) {
+	os.Setenv(PbnjGrpcAuth, pbnjGrpcAuth)
+
+	conn, _ := client.SetupConnection()
+
+	mClient := v1.NewMachineClient(conn)
+	tClient := v1.NewTaskClient(conn)
+	pbnjObj := &Pbnj{client.NewPbnjClient(mClient, tClient)}
+
+	os.Unsetenv(PbnjGrpcAuth)
```
chrisdoherty4, line 35: To make it clear for maintainers this could do with being deferred immediately after the `os.Setenv()`.

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"os"
+
+	"github.com/tinkerbell/cluster-api-provider-tinkerbell/pbnj/client"
+	"github.com/tinkerbell/pbnj/api/v1"
+)
+
+const (
+	PbnjGrpcAuth = "PBNJ_GRPC_AUTHORITY"
```
chrisdoherty4, line 12: Do we expect package consumers to need to use this at all? Seems unnecessary if they explicitly pass in an authorization parameter to the PBnJ client on l. 26.

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type Pbnj struct {
+	*client.PbnjClient
```
chrisdoherty4, line 16: Are you trying to model `Pbnj` as a client? I.e. is this an 'is a' relationship or should it be a 'has a'

pkg/dependencies/factory.go
pokearu: Ah I see what you meant. I’m fine with breaking down the factory methods and then injecting them with a parent struct.

My original intention was to have a wrapper struct that represents both the interfaces that can be injected into the provider. Just to simplify the number of individual objects being passed in.

pkg/providers/tinkerbell/tinkerbell.go
pokearu: Yah I agree. We could possibly break out all the external clients and have them injected somewhere else and the constructor would only target it’s members. 
But yes again another discussion 👍🏻

pkg/dependencies/factory.go
chrisdoherty4: There would still be a struct for the provider representing clients. But, as I said, that's orthogonal to how those clients are constructed. I spoke with Pankti and pointed out the `WithProviderFactory()` function l. 185 can create the client struct inline. The point there is that the struct is a property of the provider factory and it doesn't need to impact the construction of the individual clients. 

Happy to chat to tomorrow to clarify further.

pkg/providers/tinkerbell/pbnj/client.go
chrisdoherty4: @panktishah26 does this need to be exported?

pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+type BmcSecretConfig struct {
+	Host     string
+	Username string
+	Password string
+	Vendor   string
+}
+
+func NewPBNJClient(pbnjGrpcAuth string) (*Pbnj, error) {
```
chrisdoherty4, line 26: Its very difficult to know what valid values for this are. Can we define some constants to help consumers?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -75,21 +81,26 @@ type ProviderTinkClient interface {
 	GetHardware(ctx context.Context) ([]*tink.Hardware, error)
 }
 
-func NewProvider(datacenterConfig *v1alpha1.TinkerbellDatacenterConfig, machineConfigs map[string]*v1alpha1.TinkerbellMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, providerTinkClient ProviderTinkClient, now types.NowFunc, skipIpCheck bool, hardwareConfigFile string) *tinkerbellProvider {
-	return NewProviderCustomNet(
+type ProviderPbnjClient interface {
```
chrisdoherty4, line 84: I don't think this interface name adequately describes the intent. Specifically, PBnJ is an implementation to achieve validating BMC credentials but as far as this interface is concerned, that point is moot. Suggest `BMCCredentialValidator` instead.

pkg/providers/tinkerbell/tinkerbell_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
-	"github.com/tinkerbell/tink/protos/hardware"
+	tinkHardware "github.com/tinkerbell/tink/protos/hardware"
```
chrisdoherty4, line 10: Imports should be all lower case.

chrisdoherty4:
I don't see any tests that model the behavior when BMC credentials are invalid. Are you adding that separately?

pkg/providers/tinkerbell/pbnj/client.go
panktishah26: Customers are providing this value in the clusterconfig file (tinkerbellPBnJGRPCAuth) and I am passing it's value in this function only.

pkg/providers/tinkerbell/tinkerbell.go
panktishah26: We might use PBnJ for powering on/off machines after beta. Also, I feel we should keep the name consistent here same like ProviderTinkClient, ProviderKubectlClient, ProviderGovcClient, etc.

panktishah26:
> I don't see any tests that model the behavior when BMC credentials are invalid. Are you adding that separately?

Yes, I will add that separately.

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: I'm gonna back track here: for some reason I thought the method was non pbnj specific but it accepts a pbnj type so ignore this comment.

pkg/providers/tinkerbell/pbnj/client.go
chrisdoherty4: Understood. If we forget about how this package is being used externally for a moment, how would we know whats valid and whats not valid? Does PbnJ define some consts upstream? If so, could we have a code comment that links to it.

Issue #189, opened 2021-09-13
Support node taints in cluster config


jiayiwang7:
**What would you like to be added**:

Add node taints option in cluster config for both control plane and worker node groups. So that the defined taints can be attached to the node(s) automatically during cluster creation/upgrade.

**cluster config**

We add taints option at the `cluster` level, eg:

```
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: Cluster
metadata:
  name: cluster-1
spec:
  ...
  controlPlaneConfiguration:
    count: 2
    endpoint:
      host: "198.18.100.27"
    machineGroupRef:
      kind: VSphereMachineConfig
      name: cluster-1-cp
    taints: # optional
    - key: key1
      value: val1 # optional
      effect: NoSchedule # Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
  workerNodeGroupConfigurations:
  - count: 2
    machineGroupRef:
      kind: VSphereMachineConfig
      name: cluster-1
    taints: # optional
    - key: key2
      value: val2 # optional
      effect: PreferNoSchedule # Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
```

**implementation**

Cluster API already supports taints for both control plane and machine deployment:
- CP: https://github.com/kubernetes-sigs/cluster-api/blob/master/controlplane/kubeadm/config/crd/bases/controlplane.cluster.x-k8s.io_kubeadmcontrolplanes.yaml#L708 and https://github.com/kubernetes-sigs/cluster-api/blob/master/controlplane/kubeadm/config/crd/bases/controlplane.cluster.x-k8s.io_kubeadmcontrolplanes.yaml#L895
- MD: https://github.com/kubernetes-sigs/cluster-api/blob/master/bootstrap/kubeadm/config/crd/bases/bootstrap.cluster.x-k8s.io_kubeadmconfigtemplates.yaml#L656

We might easily pass down the taints option to the CAPI spec before applying.

The CAPI taints option does not work for worker nodes (kubeadmconfigtemplates). We https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/controllers/controller.go#L267might add taints for worker nodes in the CLI workflow, or as a `postKubeadmCommands` defined in the `kubeadmconfigtemplates`. 

**Why is this needed**:

[...]

PR #1254, opened 2022-02-21
Control Plane Taints: end-to-end tests; account for applied taints

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1226
https://github.com/aws/eks-anywhere/issues/189

*Description of changes:*
- Account for control plane taints in end-to-end tests
  - map a CP configuration back to a node, and compare their taints
  - if no taints are provided by the end-user, ensure the default taint is present
- Add methods to make validating node-to-configuration mappings much easier
  - add method `ValidateWorkerNodes` and `ValidateControlPlaneNodes`, which accept var arg validation functions
  - these methods then deduce the config-to-node mapping and run the provided validation functions, passing the node/config pair for each node
  - these can be later added to the internals of methods like `ValidateCluster` to call a variety of validation methods on the underlying nodes and abstract away the need to specifiy all of these validations at the test level. This refactoring is a future extension of this work.
- Update some existing configurations to ensure they tolerate custom CP taints
  - AWS IAM Auth now tolerate custom CP taints in addition to the default one
  - VSphere CSI drivers now tolerate custom CP taints in addition to the default one

As discussed in pervious PRs, unit tests for the new kubectl methods are coming in a separate PR to keep this one from getting too too out of hand.

*Testing (if applicable):*
Ran the taints tests on both BottleRocket and Ubuntu


test/framework/workerNodes.go
```diff
@@ -0,0 +1,68 @@
...
+								e.T.Errorf("Worker node %v, member of Worker Node Group Configuration %v, is not valid: %v", node.Name, w.Name, err)
+							}
+						}
+					}
+				}
+			}
+		}
+		e.StopIfFailed()
+	}
+	e.T.Log("validated that expected taints are present on the workload cluster nodes")
```
danbudris, line 67:
```suggestion
```

test/e2e/taints_test.go
```diff
@@ -26,26 +34,35 @@ func runTaintsUpgradeFlow(test *framework.ClusterE2ETest, updateVersion v1alpha1
 // remove a taint from a node
 // add a taint to a node which already has another taint
 // add a taint to a node which had no taints
-func TestVSphereKubernetes121TaintsWorkerNodeGroups(t *testing.T) {
-	worker0 := "worker-0"
-	worker1 := "worker-1"
-	worker2 := "worker-2"
+func TestVSphereKubernetes121Taints(t *testing.T) {
```
vivek-koppuru, line 37: Is the feature flag being removed with these changes? If not, you need to pass that in too right?

danbudris: yes true, good call

danbudris: done https://github.com/aws/eks-anywhere/pull/1254/commits/64727114bf2dbb536274b62fee8e4a91b08ef0d5

vivek-koppuru (approved):
/lgtm

PR #1255, opened 2022-02-21
EKS Anywhere get packages command

a-cool-train:
*Description of changes:*
As part of including support for curated packages, I am including the start of the get curated packages command in this pull request.


TerryHowe (approved):

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var getPackageCommand = &cobra.Command{
+	Use:     "package(s) [flags]",
+	Aliases: []string{"package", "packages"},
+	Short:   "Get package(s)",
+	Long:    "This command is used to display the installed packages",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if !features.IsActive(features.CuratedPackagesSupport()) {
+			return fmt.Errorf("curated Packages not supported in this release")
```
vivek-koppuru, line 35:
```suggestion
			return fmt.Errorf("This command is currently not supported")
```

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+	getCmd.AddCommand(getPackageCommand)
+	getPackageCommand.Flags().StringVarP(&gpo.output, "output", "o", "", "Specifies the output format (valid option: json, yaml)")
+}
+
+var getPackageCommand = &cobra.Command{
+	Use:     "package(s) [flags]",
+	Aliases: []string{"package", "packages"},
+	Short:   "Get package(s)",
+	Long:    "This command is used to display the installed packages",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
```
vivek-koppuru, line 33: Let's move this to a separate function like we do here: https://github.com/aws/eks-anywhere/blob/main/cmd/eksctl-anywhere/cmd/supportbundle.go#L80

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+				log.Fatalf("Error initializing flags: %v", err)
+			}
+		})
+		return nil
+	},
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		packageInstanceName := ""
+		if len(args) > 0 {
+			packageInstanceName = args[0]
```
vivek-koppuru, line 48: Let's have a separate validate function that is a bit more robust and can fail if more than one arg is specified

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		return getPackages(cmd.Context(), packageInstanceName, gpo.output)
+	},
+}
+
+func getPackages(ctx context.Context, packageInstanceName string, output string) error {
+	kubeConfig := os.Getenv(kubeconfigEnvVariable)
+	executableBuilder, _, err := executables.NewExecutableBuilder(ctx, executables.DefaultEksaImage())
```
vivek-koppuru, line 56: Can we use this instead to keep our dependencies logic in sync with how we build dependencies elsewhere? I think we are moving away from directly calling NewExecutableBuilder. https://github.com/aws/eks-anywhere/blob/main/pkg/dependencies/factory.go#L85

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+	kubectl := executableBuilder.BuildKubectlExecutable()
+	packages, err := kubectl.GetPackagesFromKubectl(ctx, packageInstanceName, kubeConfig, output)
+	if err != nil {
+		return fmt.Errorf("error executing kubectl: %v", err)
+	}
+	fmt.Println(packages)
```
vivek-koppuru, line 65: How does this output look like? Do we want to print it out simpler than what kubectl returns?

pkg/constants/constants.go
```diff
@@ -32,4 +32,5 @@ const (
 
 	VSphereCredentialsName = "vsphere-credentials"
 	EksaLicenseName        = "eksa-license"
+	EksaPackagesName       = "eksa-packages"
```
vivek-koppuru, line 35: What is this used for? I see it is passed into the kubectl flag, but it isn't clear how it is different from just packages. If it is a namespace, is there a reason why we aren't putting it under `eksa-system`?

pkg/executables/kubectl.go
```diff
@@ -1222,3 +1222,27 @@ func (k *Kubectl) GetDaemonSet(ctx context.Context, name, namespace, kubeconfig
...
+	params := []string{
+		"get", "packages", kubeconfigFlag, kubeconfig, namespaceFlag, constants.EksaPackagesName,
+	}
+	params = append(params, withPackageInstanceName(packageInstanceName)...)
+	params = append(params, withOutput(output)...)
+	stdOut, err := k.Execute(ctx, params...)
+	return stdOut.String(), err
+}
+
+func withOutput(output string) []string {
```
vivek-koppuru, line 1236: This helpers don't seem like they are adding much apart from putting the if conditional in the function. Is there any other reason we want to do it this way? If output and packageInstanceName can be empty strings, then we should follow the following type of format: https://github.com/aws/eks-anywhere/blob/main/pkg/executables/kubectl.go#L682

vivek-koppuru (commented):
What is the yaml binary under `cmd/eksctl-anywhere/yaml`?

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,67 @@
...
+	getPackageCommand.Flags().StringVarP(&gpo.output, "output", "o", "", "Specifies the output format (valid option: json, yaml)")
+}
+
+var getPackageCommand = &cobra.Command{
+	Use:     "package(s) [flags]",
+	Aliases: []string{"package", "packages"},
+	Short:   "Get package(s)",
+	Long:    "This command is used to display the installed packages",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if !features.IsActive(features.CuratedPackagesSupport()) {
```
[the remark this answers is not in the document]
a-cool-train: We are adding the functionality of providing mechanisms for managing curated packages (previously referred to as addons). This is the first command (get) to start on this journey. There will be more commands but in order to make the PR simpler, I have started with the get packages command in this PR.

pkg/constants/constants.go
a-cool-train: We are planning to store all the installed packages under the eksa-packages namespace. The packages we are installing don't qualify as system level packages hence the reason of having a separate namespace.

cmd/eksctl-anywhere/cmd/getpackage.go
a-cool-train: For this initial command, we don't plan on filtering the output from kubectl. There will be other commands where we will be filtering out the response.

cmd/eksctl-anywhere/cmd/getpackage.go
a-cool-train: Let me look into that.

cmd/eksctl-anywhere/cmd/getpackage.go
a-cool-train: Is there any reason we want to fail in case more than the necessary args are specified? The validation for our use case is to a minimal and if there is a necessity for more validation, it is a good idea to move it to a function but I don't see that here in this case.

pkg/executables/kubectl.go
a-cool-train: I think it makes the code easily readable.

a-cool-train: Ignore my comment.

TerryHowe:
It kind of looks like you did a `git pull` instead of a `git rebase` to resolve merge conflicts. Not a huge deal since we squash, but looks a bit bit messy.

pkg/executables/kubectl.go
```diff
@@ -1311,3 +1311,27 @@ func (k *Kubectl) GetDaemonSet(ctx context.Context, name, namespace, kubeconfig
 
 	return obj, nil
 }
+
+func (k *Kubectl) GetPackagesFromKubectl(ctx context.Context, packageInstanceName, kubeconfig string, output string) (string, error) {
+	params := []string{
+		"get", "packages", kubeconfigFlag, kubeconfig, namespaceFlag, constants.EksaPackagesName,
```
g-gaston, line 1317: nit: I understand the desire to dry out hardcoded strings but I think that doing this for the flags hurts a little bit the readability. And I hardly doubt that those flags are ever going to change.

pkg/executables/kubectl.go
g-gaston: I'm down for these methods but it does seem like they are the same as `KubectlOpt`
But I do like that they are abstracted from the public interface

pkg/executables/kubectl.go
g-gaston, line 1317: is packages a CRD?

g-gaston (commented):
Looks good, just a couple nits
Waiting for the design discussion before the `lgtm` but no major concerns 🚀

pkg/executables/kubectl.go
a-cool-train: Ok so I have resolved the packageInstanceName but I don't see anyway around output. I have looked into using KubectlOpt but it doesn't fit the needs for this specific case. The reason I say this is because -o flag needs to be set if and only if there is an output of type yaml or any other set. Because of that, it doesn't quite fit the KubectlOpt suggestion here. Please correct me if I am missing something.

vivek-koppuru: You only set `WithOutput` if you need it, and that would be called from the caller of `GetPackagesFromKubectl`, not from within the function.

pkg/executables/kubectl.go
a-cool-train: Yes there is a CRD associated with it but this will be a CR that will be fetched.

pkg/executables/kubectl.go
```diff
@@ -695,6 +695,17 @@ func WithOverwrite() KubectlOpt {
 	return appendOpt("--overwrite")
 }
 
+func WithOutput(output string) KubectlOpt {
+	if output == "" {
+		return appendOpt()
```
vivek-koppuru, line 700: You can just return an empty func right? But this check is redundant because the user would not specify `WithOutput` if the string is empty.

vivek-koppuru (commented):
Also you should rebase your branch with upstream instead to clean up the commits even for a squash

pkg/executables/kubectl.go
a-cool-train: I am not sure whether the suggestion is to move the check on empty output to the client calling the function? Either ways, the check needs to happen somewhere, I have preferred it in here since the output from the customer might be empty and we can just return an empty func if that is the case.

a-cool-train:
I did rebase from main since the last discussion on rebase but the commits you are currently seeing are from the initial merge I had done. Not sure if it can be undone without much effort which might not be necessary since the commits will be squashed anyways.

pkg/executables/kubectl.go
vivek-koppuru: What I am saying is that the check is redundant and not needed because the user should not be specifying this option with an empty string, and if they did they may be on purpose. If you take a look at the others like `WithKubeconfig`, that's what we do.

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,84 @@
...
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+	kubectl := deps.Kubectl
+
+	var packages string
+	if output != "" {
+		packages, err = kubectl.GetPackages(ctx, executables.WithKubeconfig(kubeConfig), executables.WithOutput(output), executables.WithArgs(args))
+	} else {
+		packages, err = kubectl.GetPackages(ctx, executables.WithKubeconfig(kubeConfig), executables.WithArgs(args))
+	}
```
vivek-koppuru, line 67:
```suggestion
        params := []KubectlOpt{executables.WithKubeconfig(kubeConfig), executables.WithArgs(args)}
	if output != "" {
		params = append(params, executables.WithOutput(output)
	}
	packages, err = kubectl.GetPackages(ctx, params...)
```

cmd/eksctl-anywhere/cmd/getpackage.go
```diff
@@ -0,0 +1,83 @@
...
+
+	params := []executables.KubectlOpt{executables.WithKubeconfig(kubeConfig), executables.WithArgs(args)}
+	if output != "" {
+		params = append(params, executables.WithOutput(output))
+	}
+	packages, err := kubectl.GetPackages(ctx, params...)
+
+	if err != nil {
+		fmt.Print(packages)
+		return fmt.Errorf("error executing kubectl: %v", err)
```
vivek-koppuru, line 70: Can we make the output of `GetPackages` part of the error instead of doing a print here? It should be there as part of `err`

a-cool-train: I separated them out since kubectl returns the current resources, then prints an error. I separated to follow this standard from kubectl.

vivek-koppuru (approved):

PR #1256, opened 2022-02-21
Adding cloudstack validator with unit tests

maxdrib:
*Issue #, if available:*

*Description of changes:*
In alignment with other providers, we have implemented a validator utility class which will be used by the Cloudstack provider to simplify the logic within. 

*Testing (if applicable):*
Unit tests, lint succeeds


pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+	if len(controlPlaneMachineConfig.Spec.Users) <= 0 {
+		controlPlaneMachineConfig.Spec.Users = []anywherev1.UserConfiguration{{}}
+	}
+	if len(workerNodeGroupMachineConfig.Spec.Users) <= 0 {
+		workerNodeGroupMachineConfig.Spec.Users = []anywherev1.UserConfiguration{{}}
+	}
+	if len(controlPlaneMachineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
+		controlPlaneMachineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
+	}
+	if len(workerNodeGroupMachineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
```
g-gaston, line 107: It seems like we have a lot of duplicated code for validating the machine configs
Could we form a slice of cloud stack machine configs and iterate over them, running the common validations?

pkg/providers/cloudstack/validator_test.go
```diff
@@ -0,0 +1,90 @@
...
+	Value: "zone1",
+}
+
+func givenCloudMonkeyMock(t *testing.T) *mocks.MockProviderCmkClient {
+	ctrl := gomock.NewController(t)
+	return mocks.NewMockProviderCmkClient(ctrl)
+}
+
+func givenDatacenterConfig(t *testing.T, fileName string) *v1alpha1.CloudStackDatacenterConfig {
+	deploymentConfig, err := v1alpha1.GetCloudStackDatacenterConfig(path.Join(testDataDir, fileName))
```
g-gaston, line 31: I would suggest avoiding this and declaring the datacenter and machine configs structs in code directly

g-gaston (commented):
Looks good, two minor suggestions

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func NewValidator(cmk ProviderCmkClient, machineConfigs map[string]*anywherev1.CloudStackMachineConfig) *Validator {
+	return &Validator{
+		cmk:            cmk,
+		machineConfigs: machineConfigs,
+	}
+}
+
+type ProviderCmkClient interface {
```
vivek-koppuru, line 29: Let's create a `cloudstack.go` and keep the interface there even if that file doesn't have any other content right now. That would allow us to not have to move it back in a later PR. Same with any other constants or methods that would reside in that file.

pkg/providers/cloudstack/validator.go
vivek-koppuru: Yea we added support for worker node groups when we refactored this a little which doesn't seem to be included here. I forget, we aren't planning on supporting multiple worker node groups from the start right?

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return nil
+}
+
+func (v *Validator) validateControlPlaneHost(pHost *string) error {
+	_, port, err := net.SplitHostPort(*pHost)
+	if err != nil {
+		if strings.Contains(err.Error(), "missing port") {
+			port = controlEndpointDefaultPort
+			*pHost = fmt.Sprintf("%s:%s", *pHost, port)
```
vivek-koppuru, line 187: I am not sure how I feel about editing a value from the parameter, but I think we do something like this with ssh auth keys in vsphere. Maybe we can just call this `validateAndParseControlPlaneHost` so that it is clear that the value is being modified.

pkg/providers/cloudstack/validator.go
maxdrib: I'm not sure I agree. At this point, the provider itself doesn't need to interact with Cmk - its only interactions are for validations, which should presumably all occur in validator.go, so I'm inclined to keep the CloudMonkey structs isolated to the validator if possible. What do you think?

pkg/providers/cloudstack/validator.go
maxdrib: Correct, that would make the initial merge more complicated, so I'd prefer to keep worker node groups out if possible. I'll see what I can do about iterating/validating, thanks for the tip :)

pkg/providers/cloudstack/validator.go
chrisdoherty4: I think you're on the right track, Vivek. This method shouldn't really modify the input, even if we do it elsewhere. Its more idiomatic to return the value and set it explicitly in the calling function. The function could do with reflecting the fact it will apply a default and a short code comment would be useful to clarify it: `validateControlPlaneHostAndApplyDefaultPort(host string) (string, error)`.

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+		} else {
+			return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host is invalid: %s (%s)", *pHost, err.Error())
```
chrisdoherty4, line 189: This function isn't operating on a `controlPlaneConfiguration.Endpoint.Host` so it shouldn't really add that context. The struct path context can be added in the caller instead.

pkg/providers/cloudstack/testdata/cluster_main.yaml
```diff
@@ -0,0 +1,109 @@
...
+  name: test-etcd
+  namespace: test-namespace
+spec:
+  computeOffering:
+    value: "m4-large"
+    type: "name"
+  users:
+    - name: "mySshUsername"
+      sshAuthorizedKeys:  # The key below was manually generated and not used in any production systems
+        - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC1BK73XhIzjX+meUr7pIYh6RHbvI3tmHeQIXY5lv7aztN1UoX+bhPo3dwo2sfSQn5kuxgQdnxIZ/CTzy0p0GkEYVv3gwspCeurjmu0XmrdmaSGcGxCEWT/65NtvYrQtUE5ELxJ+N/aeZNlK2B7IWANnw/82913asXH4VksV1NYNduP0o1/G4XcwLLSyVFB078q/oEnmvdNIoS61j4/o36HVtENJgYr0idcBvwJdvcGxGnPaqOhx477t+kfJAa5n5dSA5wilIaoXH5i1Tf/HsTCM52L+iNCARvQzJYZhzbWI1MDQwzILtIBEQCJsl2XSqIupleY8CxqQ6jCXt2mhae+wPc3YmbO5rFvr2/EvC57kh3yDs1Nsuj8KOvD78KeeujbR8n8pScm3WDp62HFQ8lEKNdeRNj6kB8WnuaJvPnyZfvzOhwG65/9w13IBl7B1sWxbFnq2rMpm5uHVK7mAmjL0Tt8zoDhcE1YJEnp9xte3/pvmKPkST5Q/9ZtR9P5sI+02jY0fvPkPyC03j2gsPixG7rpOCwpOdbny4dcj0TDeeXJX8er+oVfJuLYz0pNWJcT2raDdFfcqvYA0B0IyNYlj5nWX4RuEcyT3qocLReWPnZojetvAG/H8XwOh7fEVGqHAKOVSnPXCSQJPl6s0H12jPJBDJMTydtYPEszl4/CeQ== testemail@test.com"
```
chrisdoherty4, line 103: We probably don't want to check in public keys, right?

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+		return err
+	}
+
+	for _, machineConfig := range v.machineConfigs {
+		if machineConfig.Namespace != clusterSpec.Namespace {
+			return errors.New("CloudStackMachineConfig and Cluster objects must have the same namespace specified")
+		}
+	}
+	if clusterSpec.datacenterConfig.Namespace != clusterSpec.Namespace {
+		return errors.New("CloudStackDatacenterConfig and Cluster objects must have the same namespace specified")
```
chrisdoherty4, line 122: These errors could be more informative. Tell the user what you received so they know, explicitly, what the program doesn't like.

```go
fmt.Errorf("CloudStackDatacenterConfig and Cluster objects must have the same namespace: CloudstackDatacenterConfig namespace=%s; Cluster namespace=%s", clusterSpec.datacenterConfig.Namespace, clusterSpec.Namespace)
```

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (v *Validator) validateMachineConfig(ctx context.Context, datacenterConfigSpec anywherev1.CloudStackDatacenterConfigSpec, machineConfig *anywherev1.CloudStackMachineConfig) error {
+	domain := datacenterConfigSpec.Domain
+	zone := datacenterConfigSpec.Zone
+	account := datacenterConfigSpec.Account
+	if err := v.cmk.ValidateTemplatePresent(ctx, domain, zone, account, machineConfig.Spec.Template); err != nil {
+		return fmt.Errorf("error validating template: %v", err)
+	}
+	if err := v.cmk.ValidateServiceOfferingPresent(ctx, domain, zone, account, machineConfig.Spec.ComputeOffering); err != nil {
+		return fmt.Errorf("error validating template: %v", err)
```
chrisdoherty4, line 171: The context here is wrong, I think its just a copy-paste error.

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return nil
+}
+
+func (v *Validator) validateMachineConfig(ctx context.Context, datacenterConfigSpec anywherev1.CloudStackDatacenterConfigSpec, machineConfig *anywherev1.CloudStackMachineConfig) error {
+	domain := datacenterConfigSpec.Domain
+	zone := datacenterConfigSpec.Zone
+	account := datacenterConfigSpec.Account
+	if err := v.cmk.ValidateTemplatePresent(ctx, domain, zone, account, machineConfig.Spec.Template); err != nil {
+		return fmt.Errorf("error validating template: %v", err)
```
chrisdoherty4, line 168: As a general comment, applying the word `error` when adding context is superfluous. We know its an error, that's how its reported to the user. The context need only shed light on the circumstances: `validating template: %v`.

I know we use the word `error` in lots of places, but think about how that presents: `error running create: error validating template: error with foo: error with bar`. It becomes a little obnoxious 😄

pkg/providers/cloudstack/validator_test.go
maxdrib: Sounds good

pkg/providers/cloudstack/testdata/cluster_main.yaml
maxdrib: See the comment above - this is the same exact key used across the codebase, and I confirmed with vignesh that it was generated for test purposes. The code does validate the format of the key, so it has to at least resemble one

pkg/providers/cloudstack/validator.go
g-gaston: My suggestion was independent from supporting worker node groups or not
There is duplicated code here when validating machine configs for cp, etcd and workers (it really doesn't matter if we support multiple worker node groups here, since all the worker node groups could be pointing to the same cloud stack machine config).
All of these machine configs can be grouped in a slice and run the common validations from there

As far as aligning with other provider patterns, I would suggest to make a judgment call case by case. Not all the code present for other providers is ideal. Most of it has been built over time and the patterns that made sense in the past might not make sense anymore.  We should refactor them if they don't, of course. But that's a different topic. 

I think we should write this new code in the way that makes the most sense now.

pkg/providers/cloudstack/validator.go
g-gaston: I would advise to keep defaults outside of the validator
For vsphere we had a "validator" and a "defaulter" (no need to follow the same)
A couple of reasons:
* Changing the input in a validator is an unexpected side effect. This could lead to hidden bugs
* When running this in a validation webhook, we can't really modify the api object

maxdrib: I agree - would it be sufficient to add a TODO to move this to the cloudstack provider, after this code is merged?

pkg/providers/cloudstack/validator.go
maxdrib: Refactored to perform more checks inside the machineConfig for loop, while still trying to keep some valuable error messaging in place

pkg/providers/cloudstack/validator.go
vivek-koppuru: Ohh I didn't realize that the provider doesn't need the cmk, which if that is the case you can disregard my comment. I was relating it to govc

pkg/providers/cloudstack/validator.go
vivek-koppuru: That's a good point that we didn't really consider before, I would say `unable to validate template` or something is fine too, a bit more descriptive of the action. I think there is value in bubbling up with some context as to what kind of error validating template is, because it could make it seem like there was an error validating after finding the template or something.

vivek-koppuru (approved):
I think these changes are great overall now!

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,181 @@
...
+	hostWithPort, err := v.validateControlPlaneHostAndApplyDefaultPort(cloudStackClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host)
+	if err != nil {
+		return fmt.Errorf("failed to validate controlPlaneConfiguration.Endpoint.Host: %v", err)
+	}
+	cloudStackClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host = hostWithPort
+
+	for _, machineConfig := range v.machineConfigs {
+		if machineConfig.Namespace != cloudStackClusterSpec.Namespace {
+			return fmt.Errorf(
+				"CloudStackMachineConfig and Cluster objects must have the same namespace: CloudStackMachineConfig namespace=%s; Cluster namespace=%s",
```
g-gaston, line 123: It might be easier for the user to correct this mistake if we include the name of the invalid `CloudStackMachineConfig`

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if cloudStackClusterSpec.datacenterConfig.Namespace != cloudStackClusterSpec.Namespace {
+		return fmt.Errorf(
+			"CloudStackDatacenterConfig and Cluster objects must have the same namespace: CloudstackDatacenterConfig namespace=%s; Cluster namespace=%s",
+			cloudStackClusterSpec.datacenterConfig.Namespace,
+			cloudStackClusterSpec.Namespace,
+		)
+	}
+
+	workerNodeGroupMachineConfig, ok := v.machineConfigs[cloudStackClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef.Name]
```
g-gaston, line 104: I think we should avoid this kind of thing
Even if we don't support multiple worker node groups (if we don't, we should probably add a validation for that), there is no reason why we can't run the validations for all the worker node group machine configs in a loop. It takes the same amount of time and it facilitates adding support later

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,181 @@
...
+		if controlPlaneMachineConfig.Spec.Template != workerNodeGroupMachineConfig.Spec.Template {
+			return fmt.Errorf("control plane and worker nodes must have the same template specified")
+		}
+	}
+
+	hostWithPort, err := v.validateControlPlaneHostAndApplyDefaultPort(cloudStackClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host)
+	if err != nil {
+		return fmt.Errorf("failed to validate controlPlaneConfiguration.Endpoint.Host: %v", err)
+	}
+	cloudStackClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host = hostWithPort
```
g-gaston, line 118: I would vote to move this to a separate standalone method and keep the validator free from "updates"
We can still run the host port validation here, just not assign the default port
So two methods `validateControlPlaneEndpoint` and `setDefaultControlPlanePort`

maxdrib: will define separate methods for the two usages

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,181 @@
...
+	if len(cloudStackClusterSpec.datacenterConfig.Spec.Domain) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
+	}
+	if len(cloudStackClusterSpec.datacenterConfig.Spec.Zone.Value) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig zone is not set or is empty")
+	}
+	if len(cloudStackClusterSpec.datacenterConfig.Spec.Network.Value) <= 0 {
+		return fmt.Errorf("CloudStackDatacenterConfig network is not set or is empty")
+	}
+	if cloudStackClusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef == nil {
```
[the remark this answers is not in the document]
vivek-koppuru: > I still think this should be run in the machine config validation loop. Same thing applies to the template validation. What's the reason to have it separate?
> 
> Also, we should probably run this validation at the API level, this is not exclusive to cloud stack. @vivek-koppuru thoughts?

I think that's fine, but I also think we can probably have a tracking github issue to fix it for all the providers now that we have a bunch of providers coming in at the same time.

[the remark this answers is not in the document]
g-gaston: > > I still think this should be run in the machine config validation loop.
> 
> Are you referring to the machineGroupRef nil check? How can we run that in a loop? Each path to the respective ref is unique
> 

We could have an array of machine group refs and check that none of them are nil

But I'm ok leaving this for another ticket and moving it to the api package

pkg/providers/cloudstack/validator.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("must specify machineGroupRef for control plane")
+	}
+
+	if cloudStackClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef == nil {
+		return fmt.Errorf("must specify machineGroupRef for worker nodes")
+	}
```
g-gaston, line 75: Should this be a loop as well?
The one you have for
`for _, workerNodeGroupConfiguration := range cloudStackClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {`

g-gaston (approved):

chrisdoherty4 (approved):

pkg/providers/cloudstack/validator.go
maxdrib: Yes! Good catch.

g-gaston (approved):
/lgtm

PR #1265, opened 2022-02-22
added --skip-bmc-check validation flag for create cluster command

panktishah26:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/hardware/config.go
```diff
@@ -69,7 +69,7 @@ func (hc *HardwareConfig) setHardwareConfigFromFile(hardwareFileName string) err
 	return nil
 }
 
-func (hc *HardwareConfig) ValidateHardware() error {
+func (hc *HardwareConfig) ValidateHardware(skipBmcCheck bool) error {
 	bmcRefMap := hc.initBmcRefMap()
```
ptrivedi, line 73: It would be nice to not call this initBmcRefMap in the skipBmcCheck=true case since it will not get populated/used

ptrivedi: Going to lgtm this PR for now, since all looks good. You can address this minor comment in a follow-up later.

chrisdoherty4 (approved):
/lgtm

The signature for creating the Tinkerbell provider keeps growing. Thats a pretty big code smell, I'm gonna raise a ticket to rework that.

PR #1274, opened 2022-02-23
create tinkerbell e2e test provider and test cases

gwesterfieldjr:
e2e test's for tinkerbell provider

*Description of changes:*
Update e2e test framework to account for tinkerbell provider. For now test's are only intended to be ran locally. Therefore, I updated CI e2e test run to skip the Tinkerbell tests until lab is ready.

Add simple flow test cases for Ubuntu with k8s 1.20 and 1.21

Tinkerbell simple flow is as follows

1. Generate cluster config yaml
2. Generate hardware yaml based on developer provided inventory.csv.
3. Create cluster
4. Delete cluster

*Testing:*

Build, unit, lint, and ran the new e2e test cases for tinkerbell locally against our port99 bare metal lab


cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -130,7 +130,13 @@ func (hOpts *hardwareOptions) generateHardware(ctx context.Context) error {
 			return fmt.Errorf("failed reading csv: %v", err)
 		}
 
-		id := uuid.New().String()
+		var id string
```
gwesterfieldjr, line 133: I had to update hardware to support optionally passing the hardware guid. This is required if you want to generate hardware.yaml for hardware that already exists in the tinkerbell stack which is required for the e2e test's since all the hardware already exists. Otherwise, the hardware generate command creates new guids for the hardware that does not match what is in our tinkerbell sandbox.

cmd/integration_test/build/buildspecs/conformance-eks-a-cli.yml
```diff
@@ -5,6 +5,7 @@ env:
     INTEGRATION_TEST_MAX_EC2_COUNT: 25
     T_VSPHERE_CIDR: "198.18.0.0/16"
     T_VSPHERE_PRIVATE_NETWORK_CIDR: "10.1.0.0/24"
+    SKIPPED_TESTS: "TestTinkerbellKubernetes120SimpleFlow,TestTinkerbellKubernetes121SimpleFlow"
```
gwesterfieldjr, line 8: Skipping Tinkerbell tests in CI until lab is ready. Until then tests are to be ran locally by developers against the port99 lab.

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -87,17 +99,18 @@ func GetTinkerbellDefaultTemplateOverride(name string) *TinkerbellTemplate {
...
-  ethernets:
-    eno1:
-        dhcp4: true
-    eno2:
-        dhcp4: true
-    eno3:
-        dhcp4: true
-    eno4:
-        dhcp4: true`,
+    version: 2
```
gwesterfieldjr, line 102: This is to clean up formatting of the outputted cluster.yaml

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -113,18 +126,19 @@ func GetTinkerbellDefaultTemplateOverride(name string) *TinkerbellTemplate {
 						"FS_TYPE":   "ext4",
 						"DEST_PATH": "/etc/cloud/cloud.cfg.d/10_tinkerbell.cfg",
 						"CONTENTS": fmt.Sprintf(`datasource:
-  Ec2:
-    metadata_urls: ["%s"]
-    strict_id: false
+    Ec2:
```
gwesterfieldjr, line 129: Clean up formatting.

test/framework/vsphere.go
```diff
@@ -31,6 +30,7 @@ const (
 	vsphereTlsThumbprintVar     = "T_VSPHERE_TLS_THUMBPRINT"
 	vsphereUsernameVar          = "EKSA_VSPHERE_USERNAME"
 	vspherePasswordVar          = "EKSA_VSPHERE_PASSWORD"
+	VsphereClusterIPPoolEnvVar  = "T_VSPHERE_CLUSTER_IP_POOL"
```
gwesterfieldjr, line 33: Renamed this back to vSphere specific as provider tests run on different networks so this can't be a shared property as I originally thought

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -78,6 +78,18 @@ func GetTinkerbellDefaultTemplateOverride(name string) *TinkerbellTemplate {
 						"CMD_LINE":            "apt -y update && apt -y install openssl",
 					},
 				},
+				{
+					Name:    "create-user",
```
gwesterfieldjr, line 82: I added this task back to the default TinkerbellTemplate as I realized I missed it in the struct PR.

gwesterfieldjr:
@abhinavmpandey08 @pokearu , I'm thinking I can demo e2e during office hours next week to get feedback and hopefully that will help with the review as well.

cmd/integration_test/build/buildspecs/conformance-eks-a-cli.yml
pokearu: Actually we don’t need to skip here. So the tests run here are compiled with different a build tag and hence only the tests added here are run.
https://github.com/aws/eks-anywhere/blob/main/test/e2e/conformance_test.go#L1

internal/pkg/api/tinkerbell.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return nil
+	}
+}
+
+func WithTinkerbellHegelURL(value string) TinkerbellFiller {
+	return func(config TinkerbellConfig) error {
+		for _, t := range config.templateConfigs {
+			for _, task := range t.Spec.Template.Tasks {
+				if task.Name == t.Name {
```
pokearu, line 120: Why are we checking the name here? It doesn’t matter if it’s the same name or not right?

internal/pkg/api/tinkerbell.go
```diff
@@ -0,0 +1,162 @@
...
+						}
+					}
+				}
+			}
+		}
+		return nil
+	}
+}
+
+func WithTemplateForAllTinkerbellMachines(value string) TinkerbellFiller {
```
pokearu, line 134: I think this can be made into something like `WithImageUrl`? I think the intention here is to set the image depending on osFamily.

pkg/providers/tinkerbell/hardware/csvparser.go
```diff
@@ -6,13 +6,18 @@ import (
 	"os"
 )
 
+const (
+	HeaderNotExist = -1
+)
+
```
pokearu, line 12: nit: Since we have a single constant.

```suggestion
const HeaderNotExist = -1
```

pokearu (approved):
Overall looks great 👍 Just left some minor comments.

internal/pkg/api/tinkerbell.go
gwesterfieldjr: Right now, the default template only has one task with a bunch of actions. I was just trying to account for future tasks being introduced. Therefore, I check name to ensure I'm only updating the required task. I will remove for now since we only have one task and it's not technically needed yet.

abhinavmpandey08 (approved):

PR #1299, opened 2022-02-24
Introduce minimal cloudstack provider with cluster create functionality

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds the remaining changes to support minimal cluster creation with the CloudStack provider. Note that at this stage, worker node groups are not supported, and supported will be added later. Additionally, only a single zone is supported (although the input is a list of zones), and account/domain specifications will also not be applied in the created cluster.

*Testing (if applicable):*
Successfully created cluster against ec2 test environment with shared network using 1 CP and 1 worker node with tilt, using bundle-overrides which include CAPC artifacts from https://github.com/aws/cluster-api-provider-cloudstack/pull/31 (multizone)


pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -108,6 +108,10 @@ type CloudStackMachineConfig struct {
 	Status CloudStackMachineConfigStatus `json:"status,omitempty"`
 }
 
+func (c *CloudStackMachineConfig) OSFamily() OSFamily {
+	return ""
```
vivek-koppuru, line 112: Why is this returning an empty string?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,765 @@
...
+
+//go:embed config/machine-health-check-template.yaml
+var mhcTemplate []byte
+
+var requiredEnvs = []string{cloudStackCloudConfigB64SecretKey}
+
+var (
+	eksaCloudStackDeploymentResourceType = fmt.Sprintf("cloudstackdatacenterconfigs.%s", v1alpha1.GroupVersion.Group)
+	eksaCloudStackMachineResourceType    = fmt.Sprintf("cloudstackmachineconfigs.%s", v1alpha1.GroupVersion.Group)
+	noProxyDefaults                      = []string{
```
vivek-koppuru, line 61: Can these defaults be shared by the providers if it's there for vsphere.go too?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,765 @@
...
+
+func (p *cloudstackProvider) DatacenterResourceType() string {
+	return eksaCloudStackDeploymentResourceType
+}
+
+func (p *cloudstackProvider) MachineResourceType() string {
+	return eksaCloudStackMachineResourceType
+}
+
+func (p *cloudstackProvider) setupSSHAuthKeysForCreate() error {
```
vivek-koppuru, line 207: ssh keys work the same way for cloudstack as they do for vmware right? Is there some wireguard like setup required to ssh?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,765 @@
...
+	datacenterConfigSpec       *v1alpha1.CloudStackDatacenterConfigSpec
+	controlPlaneMachineSpec    *v1alpha1.CloudStackMachineConfigSpec
+	workerNodeGroupMachineSpec *v1alpha1.CloudStackMachineConfigSpec
+	etcdMachineSpec            *v1alpha1.CloudStackMachineConfigSpec
+	now                        types.NowFunc
+}
+
+func (vs *CloudStackTemplateBuilder) WorkerMachineTemplateName(clusterName string) string {
+	t := vs.now().UnixNano() / int64(time.Millisecond)
+	return fmt.Sprintf("%s-worker-node-template-%d", clusterName, t)
```
vivek-koppuru, line 407: This looks like logic that can be shared too?

vivek-koppuru (commented):
This PR is huge because of the templates, so should we separate those into a different PR? That way we can focus on the yaml first and then the code after?

Also a lot of the code seems to be duplicate of vsphere logic, I am wondering if we can have a shared file, like providers.go, that contain the business logic and then we can just call those functions from the actual providers. And also utilize interfaces where possible.

maxdrib:
> This PR is huge because of the templates, so should we separate those into a different PR? That way we can focus on the yaml first and then the code after?

Created https://github.com/aws/eks-anywhere/pull/1314 to review the templates separately

maxdrib:
>Also a lot of the code seems to be duplicate of vsphere logic, I am wondering if we can have a shared file, like providers.go, that contain the business logic and then we can just call those functions from the actual providers. And also utilize interfaces where possible.

There seems to already be a package under providers/ called common, with a file called [common.go](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/common/common.go). Would this be a possible contender for somewhere to put the shared logic?

vivek-koppuru:
> common.go

I would say so yes.

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: Oh so how do you log into the node to run commands? I was assuming you can use ssh to log in just like we do for vsphere.

Also, this looks like the same as vsphere, so maybe we can move this to common too.

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: I thought the naming conventions were the same even without worker node group support?

pkg/providers/cloudstack/cloudstack.go
maxdrib: You would ssh using the user and corresponding private key to the ones passed in in the cluster spec. I can take a swing at extracting out this duplicate logic. It's a little messy since it's not read-only but maybe the common function can return a list of ssh keys and we'll set them in the invoking function

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: I think we can use the same naming convention but just not support multiple worker node groups. Unless you are saying we don't support worker nodes at all?

maxdrib: Yep, that's what I've done in the latest couple commits

pkg/decoder/cloudstack.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"os"
+
+	"gopkg.in/ini.v1"
+)
+
+const EksacloudStackCloudConfigB64SecretKey = "EKSA_CLOUDSTACK_B64ENCODED_SECRET"
+
+// ParseCloudStackSecret parses the input b64 string into the ini object to extract out the api key, secret key, and url
+func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
```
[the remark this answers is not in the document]
vivek-koppuru: Oh I think I remember this discussion, thanks for reminding me.

vivek-koppuru (approved):
Everything else looks good on my end. I am assuming we are waiting on #1239 for refactoring though right?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,692 @@
...
+		"nodeDriverRegistrarImage":               bundle.KubeDistro.NodeDriverRegistrar.VersionedImage(),
+		"livenessProbeImage":                     bundle.KubeDistro.LivenessProbe.VersionedImage(),
+		"externalAttacherImage":                  bundle.KubeDistro.ExternalAttacher.VersionedImage(),
+		"externalProvisionerImage":               bundle.KubeDistro.ExternalProvisioner.VersionedImage(),
+		"cloudstackManagementApiEndpoint":        managementApiEndpoint,
+		"managerImage":                           bundle.CloudStack.ClusterAPIController.VersionedImage(),
+		"verifySsl":                              verifySsl,
+		"cloudstackNetwork":                      datacenterConfigSpec.Zones[0].Network.Value,
+		"cloudstackDomain":                       datacenterConfigSpec.Domain,
+		"cloudstackZone":                         datacenterConfigSpec.Zones[0].Zone.Value,
```
maxdrib, line 433: todo: Add support for passing multiple zones to CAPC

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
```diff
@@ -52,39 +52,52 @@ spec:
...
+                  properties:
+                    network:
+                      description: Network is the name or UUID of the CloudStack network
+                        in which clusters should be created. It can either be an isolated
+                        or shared network. If it doesn’t already exist in CloudStack,
+                        it’ll automatically be created by CAPC as an isolated network.
+                        It can either be specified as a UUID or name In multiple-zones
+                        situation, only 'Shared' network is supported.
+                      properties:
+                        type:
```
vivek-koppuru, line 72: Just to confirm, we have an open Github issue for this right?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -501,7 +500,7 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
 	values := map[string]interface{}{
 		"clusterName":                clusterSpec.ObjectMeta.Name,
 		"kubernetesVersion":          bundle.KubeDistro.Kubernetes.Tag,
-		"cloudstackNetwork":          datacenterConfigSpec.Network.Value,
+		"cloudstackNetwork":          datacenterConfigSpec.Zones[0].Network.Value,
```
vivek-koppuru, line 503: This will eventually change in a follow up PR to support multiple networks based on the zone? Or do we only support the first network defined?

pkg/providers/common/common.go
```diff
@@ -67,13 +68,20 @@ func ParseSSHAuthKey(key *string) error {
 }
 
 func GenerateSSHAuthKey(username string, writer filewriter.FileWriter) (string, error) {
-	keygenerator, _ := crypto.NewKeyGenerator(writer)
-	sshAuthorizedKeyBytes, err := keygenerator.GenerateSSHKeyPair("", "", privateKeyFileName, publicKeyFileName, username)
-	if err != nil || sshAuthorizedKeyBytes == nil {
+	privateKeyPath, sshAuthorizedKeyBytes, err := crypto.NewSshKeyPairUsingFileWriter(writer, privateKeyFileName, publicKeyFileName)
```
vivek-koppuru, line 71: You may need to pull the latest changes we made here to make this a persistent file. 

https://github.com/aws/eks-anywhere/pull/1395

vivek-koppuru (approved):
Everything else looks good to me, my main concern is the ssh key persistence. The other stuff I was just curious if it will get changed.

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
maxdrib: you mean regarding only supporting shared networks? I think the solution to that across zones would be implementing kube-vip , which is described in https://github.com/aws/eks-anywhere/issues/1396

pkg/providers/common/common.go
maxdrib: merged latest changes with newest commit

pkg/decoder/cloudstack.go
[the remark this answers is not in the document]
chrisdoherty4: If its cloud stack specific I suggest putting it in `/pkg/providers/cloudstack/decoder`. 

On a different note, having seen multiple PRs where we tout circular dependencies as the need to put code in odd places I'm feeling like we have a significant problem. Have there been any holistic efforts to restructure our code base?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -127,6 +127,10 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command) error {
 		}
 	}
 
+	if !features.IsActive(features.CloudStackProvider()) && deps.Provider.Name() == "cloudstack" {
```
chrisdoherty4, line 130: [Ditto](https://github.com/aws/eks-anywhere/pull/1299/files#r821664422).

pkg/executables/kubectl.go
```diff
@@ -53,6 +53,16 @@ type Kubectl struct {
 	Executable
 }
 
+func (k *Kubectl) SearchCloudStackMachineConfig(ctx context.Context, name string, kubeconfigFile string, namespace string) ([]*v1alpha1.CloudStackMachineConfig, error) {
```
chrisdoherty4, line 56: Does it make sense for `Kubectl` to have these methods? I.e. will they ever be implemented?

pkg/providers/cloudstack/spec.go
```diff
@@ -11,6 +11,23 @@ type spec struct {
 	machineConfigsLookup map[string]*anywherev1.CloudStackMachineConfig
 }
 
+func NewSpec(clusterSpec *cluster.Spec, machineConfigs map[string]*anywherev1.CloudStackMachineConfig, datacenterConfig *anywherev1.CloudStackDatacenterConfig) *spec {
```
chrisdoherty4, line 14: Exported functions should return exported types. Any reason `spec` is unexported?

pkg/providers/common/common.go
```diff
@@ -2,11 +2,105 @@ package common
...
+	return nil
+}
+
+func GenerateSSHAuthKey(username string, writer filewriter.FileWriter) (string, error) {
+	privateKeyPath, sshAuthorizedKeyBytes, err := crypto.NewSshKeyPairUsingFileWriter(writer, privateKeyFileName, publicKeyFileName)
+	if err != nil {
+		return "", fmt.Errorf("VSphereMachineConfig error generating sshAuthorizedKey: %v", err)
+	}
+
+	logger.Info(fmt.Sprintf(
```
chrisdoherty4, line 76: Loggers should be injected.

pkg/providers/common/common.go
```diff
@@ -2,11 +2,105 @@ package common
...
+	}
+	return []bootstrapper.BootstrapClusterOption{bootstrapper.WithEnv(env)}, nil
+}
+
+func ParseSSHAuthKey(key *string) error {
+	if len(*key) > 0 {
+		// When public key is entered by user in provider config, it may contain email address (or any other comment) at the end. ssh-keygen allows users to add comments as suffixes to public key in
+		// public key file. When CLI generates the key pair, no comments will be present. So we get rid of the comment from the public key to ensure unit tests that do string compare on the sshAuthorizedKey
+		// will pass
+		parts := strings.Fields(strings.TrimSpace(*key))
```
chrisdoherty4, line 58: I know you got this from vSphere but its pretty convoluted and could be much simpler. 

The function name is unrepresentative of whats happening. The function is  normalizing keys for our use-case by stripping the comment from a valid ssh authorized key as opposed to parsing an ssh auth key. It also returns error messages that make assumptions about the context under which it was called. It also has a subtle, and in my view hidden, behavior in that if the string is len(0) the call is a noop. That's cunning and code shouldn't be cunning.

```
func StripSshAuthorizedKeyComment(key string) (string, error) { 
  public, _, _, _, err := ssh.ParseAuthorizedKey([]byte(*key))
  if err != nil {
    return err
  }
  return string(ssh.MarshalAuthorizedKey(public))
}
```

pkg/providers/common/common.go
```diff
@@ -2,11 +2,105 @@ package common
...
+}
+
+func GenerateSSHAuthKey(username string, writer filewriter.FileWriter) (string, error) {
+	privateKeyPath, sshAuthorizedKeyBytes, err := crypto.NewSshKeyPairUsingFileWriter(writer, privateKeyFileName, publicKeyFileName)
+	if err != nil {
+		return "", fmt.Errorf("VSphereMachineConfig error generating sshAuthorizedKey: %v", err)
+	}
+
+	logger.Info(fmt.Sprintf(
+		"DatacenterConfig private key saved to %[1]s. Use 'ssh -i %[1]s %s@<VM-IP-Address>' to login to your cluster VM",
```
chrisdoherty4, line 77: The message assumes VMs are being used which isn't true of all providers.

pkg/providers/common/common.go
maxdrib: Will implement this function instead and remove the hidden behavior, requiring the caller to check whether to strip or not (depending on key presence). Also will remove the default trailing newline added by ssh.MarshalAuthorizedKey

chrisdoherty4 (approved):
/lgtm

PR #1310, opened 2022-02-25
Get branchname from Prow env var instead of hardcoding to main

abhay-krishna:
Hardcoding to main fails when trying to get artifacts.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1311, opened 2022-02-25
Get branchname from Prow env var instead of hardcoding to main

abhay-krishna:
Same as #1310.


mrajashree (approved):

TerryHowe (commented):
/lgtm

PR #1314, opened 2022-02-25
Adding CAPC templates

maxdrib:
*Issue #, if available:*

*Description of changes:*
As requested per https://github.com/aws/eks-anywhere/pull/1299/files, splitting out the CAPC templates into a separate PR for analysis before adding the cloudstack provider code

*Testing (if applicable):*
Tested in conjunction with https://github.com/aws/eks-anywhere/pull/1299/files and successfully created a cluster


pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -0,0 +1,305 @@
...
+    spec:
+      offering: {{.cloudstackEtcdComputeOffering}}
+      template: {{.cloudstackEtcdTemplateOffering}}
+{{- if .cloudstackEtcdAffinityGroupIds }}
+      affinitygroupids:
+      {{- range .cloudstackEtcdAffinityGroupIds }}
+      - {{.}}
+      {{- end }}
+{{- end }}
+---
```
vivek-koppuru, line 304: Is this extra?

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -0,0 +1,305 @@
...
+    users:
+    - name: {{.controlPlaneSshUsername}}
+      sshAuthorizedKeys:
+      - '{{.cloudstackControlPlaneSshAuthorizedKey}}'
+      sudo: ALL=(ALL) NOPASSWD:ALL
+    format: {{.format}}
+  replicas: {{.controlPlaneReplicas}}
+  version: {{.kubernetesVersion}}
+---
+{{- if .externalEtcd }}
```
vivek-koppuru, line 241:
```suggestion
{{- if .externalEtcd }}
---
```
Similar to the comment below, but do we need these trailing delimiters if our templater will add them anyways?

maxdrib: I'm not sure if I follow. I was under the impression the templater just evaluates the logic and fills in variable as specified. Are you concerned about having duplicate delimiters, such as in the case where .exteralEtcd is undefined?

vivek-koppuru: Yup mainly about duplicate delimiters, which might be a problem in vsphere too. Might be better to catch in the review that we have here, which was why I listed it as a comment

pkg/providers/cloudstack/config/template-cp.yaml
maxdrib: removed

pkg/providers/cloudstack/config/template-cp.yaml
maxdrib: Refactored :)

vivek-koppuru (approved):
/lgtm

PR #1322, opened 2022-02-26
Validate given hardware uuid exists on the tinkerbell stack

panktishah26:
This PR is for validating given hardware uuids against the tinkerbell stack as a step of pre-flight validation of the hardware yaml file.

### DIFF
--- a/pkg/executables/tink.go
+++ b/pkg/executables/tink.go
@@ -41,13 +41,13 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
 	return nil
 }
 
-func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
+func (t *Tink) GetHardware(ctx context.Context) (map[string]*hardware.Hardware, error) {
 	params := []string{"hardware", "get", "--format", "json"}
 	data, err := t.Command(ctx, params...).WithEnvVars(t.envMap).Run()
 	if err != nil {
 		return nil, fmt.Errorf("error getting hardware list: %v", err)
 	}
-	var hardwareList []*hardware.Hardware
+	hardwareList := make(map[string]*hardware.Hardware)
 	hardwareString := data.String()
 
 	if len(hardwareString) > 0 {
@@ -57,7 +57,9 @@ func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
 			return nil, fmt.Errorf("error unmarshling hardware json: %v", err)
 		}
 		if len(hardwareListData["data"]) > 0 {
-			hardwareList = append(hardwareList, hardwareListData["data"]...)
+			for _, data := range hardwareListData["data"] {
+				hardwareList[data.GetId()] = data
+			}
 		}
 	}
 
--- a/pkg/executables/tink.go
+++ b/pkg/executables/tink.go
@@ -41,13 +41,13 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
 	return nil
 }
 
-func (t *Tink) GetHardware(ctx context.Context) (map[string]*hardware.Hardware, error) {
+func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
 	params := []string{"hardware", "get", "--format", "json"}
 	data, err := t.Command(ctx, params...).WithEnvVars(t.envMap).Run()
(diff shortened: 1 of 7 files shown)


pkg/executables/tink.go
```diff
@@ -41,13 +41,13 @@ func (t *Tink) PushHardware(ctx context.Context, hardware []byte) error {
 	return nil
 }
 
-func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
+func (t *Tink) GetHardware(ctx context.Context) (map[string]*hardware.Hardware, error) {
```
chrisdoherty4, line 44: Do we need a map of _pointers_?

pkg/executables/tink.go
```diff
@@ -57,7 +57,9 @@ func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
 			return nil, fmt.Errorf("error unmarshling hardware json: %v", err)
```
chrisdoherty4, line 57: Is the indentation correct here?

pkg/providers/tinkerbell/hardware/config.go
```diff
@@ -80,6 +81,10 @@ func (hc *HardwareConfig) ValidateHardware() error {
 			return fmt.Errorf("hardware: %s ID is required", hw.Name)
 		}
 
+		if _, ok := hardwares[hw.Spec.ID]; !ok {
+			return fmt.Errorf("hardware with the ID: %s is not present on the tinkerbell stack", hw.Spec.ID)
```
chrisdoherty4, line 85: The error seems a little confusing. `fmt.Errorf("found hardware uuid not registered with the tinkerbell stack: %v", id)`?

pkg/executables/tink.go
```diff
@@ -57,7 +57,9 @@ func (t *Tink) GetHardware(ctx context.Context) ([]*hardware.Hardware, error) {
...
 		}
 		if len(hardwareListData["data"]) > 0 {
-			hardwareList = append(hardwareList, hardwareListData["data"]...)
+			for _, data := range hardwareListData["data"] {
+				hardwareList[data.GetId()] = data
```
pokearu, line 61: I think we can leave this function as it was and form the `uuid: hardware` map while doing validations.
This method was intended to make the `tink get hardware` call and return a list of hardwares.

chrisdoherty4:
Is there a negative test case somewhere? I can't see anything on this PR.

pkg/providers/tinkerbell/hardware/config.go
ptrivedi: IMHO, starting with "found hardware uuid ... " seems more confusing to me. We could stick with the original or keep it simple and say "Hardware ID %s not registered with tinkerbell... "

chrisdoherty4: Sure. It was less about the words and more about `hardware with the ID:` not quite making sense. `hardware id '%v' is not registered with tinkerbell` sounds good.

pkg/executables/tink.go
panktishah26: I guess yes otherwise l would have got linting errors.

pkg/providers/tinkerbell/validator.go
```diff
@@ -119,11 +116,18 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, tinkerbel
 }
 
 func (v *Validator) ValidateHardwareConfig(ctx context.Context, hardwareConfigFile string) error {
+	hardwares, err := v.validateTinkerbellAccess(ctx)
```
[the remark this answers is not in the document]
panktishah26: I moved it from ValidateTinkerbellConfig() to ValidateHardwareConfig() as I needed it's return value to compare hardware uuids.

pkg/providers/tinkerbell/hardware/config.go
```diff
@@ -80,6 +81,10 @@ func (hc *HardwareConfig) ValidateHardware() error {
 			return fmt.Errorf("hardware: %s ID is required", hw.Name)
 		}
 
+		if _, ok := hardwares[hw.Spec.ID]; !ok {
```
chrisdoherty4, line 84: Why did we opt to move this here? I don't think the error makes sense now because it infers stuff about being registered with the tinkerbell stack but does do anything with the tinkerbell stack.

PR #1336, opened 2022-03-01
Rework tinkerbell hardware generation to lay ground work for extensible validation

chrisdoherty4:
As part of the Tinkerbell provider end-user workflow we generate files for registration in the Tinkerbell stack and management clusters. This is the start of a PR stack that introduces a more decoupled and cohesive set of capabilities that lay the groundwork for extensible validation of machine entries.

The approach adopts a ports and adapters style implementation that isolates reading machine configuration from a source and writing that configuration elsewhere. At a high level the `generate hardware` command will leverage `hardware.Translate(reader, writer, validator)` to translate machine configurations from 1 format to another. 

This PR focuses on the `Machine` value objects and `Translate*` functions.


pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/networkutils"
+)
+
+// Machine is a machine configuration with optional BMC interface configuration.
+type Machine struct {
+	// Id uniquely identifies this machine. It is programmatically generated so is not readable from CSV.
+	Id          string      `csv:"-"`
```
gwesterfieldjr, line 13: So during e2e tests the Id already exists in Tink stack so we do a dry run and pass the id that matches the id in tink. So in that case guid header will be in the csv. Will this account for that?

gwesterfieldjr: I ask so I can use the same struct in e2e vs creating a different one. If I need to keep it different that is fine too I guess. Whatever you recommend.

chrisdoherty4: This is where tags become a little annoying. In some contexts you want it to be required and others you don't but tags aren't dynamic like that.

Do you need the ID in the CSV or could you generate it?

abhinavmpandey08 (commented):
looks good

pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	if m.Netmask == "" {
+		return newEmptyFieldError("Netmask")
+	}
+
+	if m.MacAddress == "" {
+		return newEmptyFieldError("MacAddress")
+	}
```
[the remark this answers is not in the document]
chrisdoherty4: Added the validation.

pkg/providers/tinkerbell/hardware/machine.go
chrisdoherty4: We decided the tests can use a wrapping construct to read the CSV and require a UUID.

PR #1343, opened 2022-03-01
Cluster config parsing

g-gaston:
## What are we solving and why
* We lack an entity that represents a full cluster configuration.
  * We have `cluster.Spec` but it doesn't have provider specific api structs (like machine configs or datacenter). 
  * This has created duplicated logic, pushing providers into implementing the same thing again and again.
  * It complicates providers: since they don't receive their specific api structs as method arguments, they need to store them as members. This introduces complexity for both data management and construction. Moreover, it tides providers to a specific cluster config as a side effect.
* The api package has grown a lot, assuming too many responsibilities. It unmarshals, sets defaults and validates. And all of this tied to an inflexible public api that only takes paths to files in local disk as input.
  * This makes it difficult to construct, validate and set defaults for these structs in alternative ways.
  * This has been a patent problem in our E2E tests, which would benefit significantly from more flexibility on this area.
  * In order to move to a controller based workflow, we need to be able to run some of this logic in the validation and mutation webhooks.

## Proposal
I propose to introduce a new entity:
```go
type Config struct {
	Cluster               *anywherev1.Cluster
	VSphereDatacenter     *anywherev1.VSphereDatacenterConfig
	DockerDatacenter      *anywherev1.DockerDatacenterConfig
	VSphereMachineConfigs map[string]*anywherev1.VSphereMachineConfig
	OIDCConfigs           map[string]*anywherev1.OIDCConfig
	AWSIAMConfigs         map[string]*anywherev1.AWSIamConfig
	GitOpsConfig          *anywherev1.GitOpsConfig
}
```
`cluster.Config` represents the whole cluster API definition, including all necessary objects specified by the user and linked to the `Cluster` object.

`cluster.Spec` will wrap around `cluster.Config`, adding other external data objects that complement the cluster config to form the full cluster specification

Why exported fields? This facilitates the mutation of the cluster config: eg. defaults being set by other packages like providers. It also facilitates testing, since any test package can set the data appropriately. This is supposed to be a data container and doesn’t need to enforce anything else. As a drawback, any piece of code receiving `cluster.Spec` can modify the cluster configuration. It is up to us to define when it’s ok and when it’s not to modify it. The only decision this proposal make in that regard is that validations should never modify the cluster config.

The implementation should allow to construct a `cluster.Config` from a:
* Yaml file
  * When parsing from a yaml file, and due to the fact that most flows in the CLI happen without going through the kube api server, we need to “manually” take care of validation and defaults. This should be independent from parsing and the logic should be reusable even when not building `cluster.Config` from a file.
* Kube api server (using a Kubernetes client)
  * When using the api server, defaults and validation should already be taken care of (by the validation and mutation webhooks), so this implementation should assume so.

### Validations

There are two types of validations:
* Context agnostic validations: there are no external dependencies, having the API object is enough to validate it: mutually exclusive fields, string format, empty fields…
  * These should live in the pkg/api/version package
  * These should be the ones called in the validation webhook
* Context aware validations: these are dependent on other entities like clients or even other API objects
  * These could leave in different places, I’m not making a decision right now: providers, validations package… This is probably worth a separate discussion.
  * Try to keep these modular and reusable (eg. Don’t make them part of the providers struct’s api)
  * If they only need other API objects from the same package, they can live in the api package
  * Controller: these shouldn’t be run in a webhook and must be part of the main reconciliation loop

### Defaults
These follow the same categorization and rules as the validations: 
* Context agnostic
    * In api package
    * Run un mutation webhook
* Context aware
    * Keep them modular
    * In controller reconcile loop

## Implementation plan
* **Add new `cluster.Config` struct (this PR)**
* **Add utils to build `Config` from yaml (this PR)**
* Make sure all validations in the `api/v1alpha1` package comply with the proposed definition of context unaware validations
* Make sure all defaults in the `api/v1alpha1` package comply with the proposed definition of context unaware defaults
* Add some validate tooling to `cluster.Config` to be able to run api level validations
* Embed `cluster.Config` in `cluster.Spec`

*Testing (if applicable):*
The PR includes unit tests

### DIFF
--- a/pkg/cluster/config.go
+++ b/pkg/cluster/config.go
@@ -1,0 +1,27 @@
+package cluster
+
+import (
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type Config struct {
+	Cluster               *anywherev1.Cluster
+	VSphereDatacenter     *anywherev1.VSphereDatacenterConfig
+	DockerDatacenter      *anywherev1.DockerDatacenterConfig
+	VSphereMachineConfigs map[string]*anywherev1.VSphereMachineConfig
+	OIDCConfigs           map[string]*anywherev1.OIDCConfig
+	AWSIAMConfigs         map[string]*anywherev1.AWSIamConfig
+	GitOpsConfig          *anywherev1.GitOpsConfig
+}
+
+func (c *Config) VsphereMachineConfig(name string) *anywherev1.VSphereMachineConfig {
+	return c.VSphereMachineConfigs[name]
+}
+
+func (c *Config) OIDCConfig(name string) *anywherev1.OIDCConfig {
+	return c.OIDCConfigs[name]
+}
+
+func (c *Config) AWSIamConfig(name string) *anywherev1.AWSIamConfig {
+	return c.AWSIAMConfigs[name]
+}
--- a/pkg/cluster/parse.go
+++ b/pkg/cluster/parse.go
@@ -1,0 +1,210 @@
+package cluster
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type parsed struct {
+	objects        objects
+	cluster        *anywherev1.Cluster
+	datacenter     apiObject
+	machineConfigs []runtime.Object
+}
+
+type basicAPIObject struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+}
+
+func (k *basicAPIObject) empty() bool {
+	return k.APIVersion == "" && k.Kind == ""
+}
+
+type apiObject interface {
+	runtime.Object
+	GetName() string
+}
+
+func keyForObject(o apiObject) string {
+	return key(o.GetObjectKind().GroupVersionKind().GroupVersion().String(), o.GetObjectKind().GroupVersionKind().Kind, o.GetName())
+}
+
+type objects map[string]apiObject
+
+func (o objects) add(obj apiObject) {
+	o[keyForObject(obj)] = obj
+}
+
+func (o objects) getFromRef(apiVersion string, ref anywherev1.Ref) apiObject {
+	return o[keyForRef(apiVersion, ref)]
+}
+
+func key(apiVersion, kind, name string) string {
+	// this assumes we don't allow to have objects in multiple namespaces
+	return fmt.Sprintf("%s%s%s", apiVersion, kind, name)
+}
+
+func keyForRef(apiVersion string, ref anywherev1.Ref) string {
+	return key(apiVersion, ref.Kind, ref.Name)
+}
+
+func ParseConfigFromFile(path string) (*Config, error) {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading cluster config file: %v", err)
+	}
+
+	return ParseConfig(content)
+}
+
+func ParseConfig(yamlManifest []byte) (*Config, error) {
+	parsed := &parsed{
+		objects: objects{},
+	}
+	yamlObjs := strings.Split(string(yamlManifest), "---")
+
+	for _, yamlObj := range yamlObjs {
+		k := &basicAPIObject{}
+		err := yaml.Unmarshal([]byte(yamlObj), k)
+		if err != nil {
+			return nil, err
+		}
+
+		// Ignore empty objects.
+		// Empty objects are generated if there are weird things in manifest files like e.g. two --- in a row without a yaml doc in the middle
+		if k.empty() {
+			continue
+		}
+
+		var obj apiObject
+
+		switch k.Kind {
+		case anywherev1.ClusterKind:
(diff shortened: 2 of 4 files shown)


pkg/cluster/config.go
```diff
@@ -0,0 +1,27 @@
...
+	Cluster               *anywherev1.Cluster
+	VSphereDatacenter     *anywherev1.VSphereDatacenterConfig
+	DockerDatacenter      *anywherev1.DockerDatacenterConfig
+	VSphereMachineConfigs map[string]*anywherev1.VSphereMachineConfig
+	OIDCConfigs           map[string]*anywherev1.OIDCConfig
+	AWSIAMConfigs         map[string]*anywherev1.AWSIamConfig
+	GitOpsConfig          *anywherev1.GitOpsConfig
+}
+
+func (c *Config) VsphereMachineConfig(name string) *anywherev1.VSphereMachineConfig {
```
g-gaston, line 17: I'm still on the fence about this.

I like the idea of hiding the fact that we store this as a map. It allow us that change that later and it also stops other packages from adding more elements.

But if we want to give the ability to iterate over the whole collection, we would need another method that returns it, which adds "extra work" when adding new providers.

Opinions?

pkg/cluster/config.go
```diff
@@ -0,0 +1,27 @@
+package cluster
+
+import (
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type Config struct {
+	Cluster               *anywherev1.Cluster
+	VSphereDatacenter     *anywherev1.VSphereDatacenterConfig
```
maxdrib, line 9: what about cloudstack?

g-gaston: Those can be added later as the need arises
Or if everyone wants that, I can add it to the scope of this feature and implement it separately

My goal here was to introduce the new pattern and tooling that can be reused by everyone, not to refactor all existing providers

pkg/cluster/parse.go
```diff
@@ -0,0 +1,210 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type parsed struct {
+	objects        objects
+	Cluster        *anywherev1.Cluster
```
jiayiwang7, line 18: why does `Cluster` field need to be exported?

pkg/cluster/config.go
```diff
@@ -0,0 +1,27 @@
...
+	DockerDatacenter      *anywherev1.DockerDatacenterConfig
+	VSphereMachineConfigs map[string]*anywherev1.VSphereMachineConfig
+	OIDCConfigs           map[string]*anywherev1.OIDCConfig
```
jiayiwang7, line 12: why do we need a map for oidc? thought we only allow one

pkg/cluster/config.go
```diff
@@ -0,0 +1,27 @@
+package cluster
+
+import (
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+type Config struct {
```
jiayiwang7, line 7: does bundle considered "external data objects" and will be only in `cluster.spec`?

pkg/cluster/parse.go
```diff
@@ -0,0 +1,210 @@
...
+		var obj apiObject
+
+		switch k.Kind {
+		case anywherev1.ClusterKind:
+			if parsed.Cluster != nil {
+				return nil, errors.New("only one Cluster per yaml manifest is allowed")
+			}
+			parsed.Cluster = &anywherev1.Cluster{}
+			obj = parsed.Cluster
+		case anywherev1.VSphereDatacenterKind:
```
jiayiwang7, line 97: currently only one vcenter, dockerdatacenter is allowed right?

pkg/cluster/config.go
g-gaston: This struct tries not to make any assumptions about what is valid and what is not
It simply tries to group everything that can be linked to a `Cluster` object

pkg/cluster/config.go
g-gaston: Yeah, correct

pkg/cluster/parse.go
g-gaston: Correct

pkg/cluster/parse.go
g-gaston: It doesn't, changed

pkg/cluster/parse.go
g-gaston: I'm not trying to validate anything here. That was one of the goals of this proposal, separating validation from parsing/unmarshalling

The only reason I check that there is only one cluster is because this code doesn't know how to handle multiple clusters. It's all built around the notion of one cluster per file, because all the processing is done from the one central cluster that links the other objects

jiayiwang7 (approved):

pkg/cluster/parse.go
```diff
@@ -0,0 +1,210 @@
...
+		var obj apiObject
+
+		switch k.Kind {
+		case anywherev1.ClusterKind:
+			if parsed.cluster != nil {
+				return nil, errors.New("only one Cluster per yaml manifest is allowed")
+			}
+			parsed.cluster = &anywherev1.Cluster{}
+			obj = parsed.cluster
+		case anywherev1.VSphereDatacenterKind:
```
chrisdoherty4, line 97: This is gonna grow so it would be good if we could make it dynamic and extensible by retrieving a marshalling func from a registry and asking the func to perform the unmarshall and return the `apiObj` so it can be cached.

PR #1349, opened 2022-03-02
TinkerbellTemplateConfig type for defining Tinkerbell workflow

pokearu:
*Description of changes:*
The changes involve adding a new type TinkerbellTemplateConfig. This type is used to define Tinkerbell workflow tasks and actions. 

*Key changes:*
- Adding new TinkerbellTemplateConfig type
- Modifying TinkerbellMachineConfig to use a `templateRef` instead of large string.
- Parsing the bundle during to `generate clusterconfig` to populate default images.
- Updating marshaller and CRD config files


config/manifest/eksa-components.yaml
```diff
@@ -3526,6 +3531,144 @@ status:
...
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: TinkerbellTemplateConfigSpec defines the desired state of
+              TinkerbellTemplateConfig
+            properties:
+              template:
+                description: Workflow represents a workflow to be executed.
```
chrisdoherty4, line 3571: > Workflow represents ...

What do you mean here? This is a template.

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -203,15 +206,23 @@ func generateClusterConfig(clusterName string) error {
 				return fmt.Errorf("failed to generate cluster yaml: %v", err)
 			}
 			datacenterYaml = dcyaml
-
-			cpMachineConfig, err := v1alpha1.NewTinkerbellMachineConfigGenerate(providers.GetControlPlaneNodeName(clusterName))
+			versionBundle, err := cluster.GetVersionsBundleForVersion(version.Get(), v1alpha1.GetDefaultKubernetesVersion())
```
chrisdoherty4, line 209: What does a 'default Kubernetes version' mean in this context? I.e. What API is the default version tied to and how is that defined?

controllers/controllers/resource/template.go
```diff
@@ -161,6 +161,7 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
 	return generateTemplateResources(templateBuilder, clusterSpec, workloadTemplateNames, kubeadmconfigTemplateNames, cpOpt)
 }
 
+// TODO: This method is currently not used. Need to add logic in reconciler for TinkerbellDatacenterKind
```
chrisdoherty4, line 164: I suggest you throw your name against the to do: `todo(pokearu): ...`. It acts as a BOM so you can search for todo's you're responsible for in the code base super easy.

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -37,128 +30,14 @@ func NewTinkerbellMachineConfigGenerate(name string) (*TinkerbellMachineConfigGe
 				Name:              "ec2-user",
```
chrisdoherty4, line 30: the indentation here has gone funky because of the double `{{` which, at a glance, makes it look like `Name` and `SshAuhotirzedKeys` are part of `TinkerbellMachineConfigSpec`. Would you mind re-formatting so the indentation helps glance-ability.

pkg/api/v1alpha1/tinkerbelltemplateconfig.go
```diff
@@ -0,0 +1,84 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1/thirdparty/tinkerbell"
+)
+
+const TinkerbellTemplateConfigKind = "TinkerbellTemplateConfig"
+
+func NewTinkerbellTemplateConfigGenerate(name string, opts ...ActionOpt) *TinkerbellTemplateConfigGenerate {
```
chrisdoherty4, line 16: Why is this limited to action customization's only?

pkg/api/v1alpha1/thirdparty/tinkerbell/workflow_types.go
```diff
@@ -0,0 +1,34 @@
+// Package represents https://pkg.go.dev/github.com/tinkerbell/tink@v0.6.0/workflow#pkg-types with json tags.
+//+kubebuilder:object:generate=true
+package tinkerbell
+
```
gwesterfieldjr, line 4: Why recreate these types vs just using existing types from https://github.com/tinkerbell/tink/blob/main/workflow/types.go?

pkg/api/v1alpha1/thirdparty/tinkerbell/workflow_types.go
```diff
@@ -0,0 +1,34 @@
+// Package represents https://pkg.go.dev/github.com/tinkerbell/tink@v0.6.0/workflow#pkg-types with json tags.
+//+kubebuilder:object:generate=true
+package tinkerbell
```
gwesterfieldjr, line 3: Should this not go under providers/tinkerbell?

pkg/api/v1alpha1/tinkerbelltemplateconfig_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			got, err := GetTinkerbellTemplateConfig(tt.fileName)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("GetTinkerbellTemplateConfig() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if !reflect.DeepEqual(got, tt.wantConfigs) {
+				t.Fatalf("GetTinkerbellTemplateConfig() = %#v, want %#v", got, tt.wantConfigs)
+			}
```
chrisdoherty4, line 180: Suggest the use of an assertion library. Assertion libraries really help document intent in a succinct way. 

```suggestion
            g := gomega.NewWithT(t)
			got, err := GetTinkerbellTemplateConfig(tt.fileName)
			g.Expect(err).To(gomega.BeEquivalentTo(tt.wantErr))
			g.Expect(gpt).To(gomega.BeEquivalentTo(tt.wanrConfigs))
```

config/manifest/eksa-components.yaml
pokearu: I think this got auto generated from the types comment. I just copied the upstream types in.
I can make this more descriptive 👍

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
pokearu: Actually i meant for this to define the default k8s version EKS-A uses.
I can make the name more descriptive 👍

controllers/controllers/resource/template.go
pokearu: Sure! But actually this isn’t related to the current changes, I noticed we hadn’t implemented so I just added a comment in case we forget. This would come in when we introduce update/upgrade.

pkg/api/v1alpha1/tinkerbelltemplateconfig_test.go
pokearu: Sure! 
I just followed some tests we had defined earlier. I can change it to use asserts.

pkg/api/v1alpha1/thirdparty/tinkerbell/workflow_types.go
pokearu: So two reasons actually.
1. k8s CRD definition require json tags on structs, or else `controller-gen` tool fails.
2. We also needed  DeepCopy method definitions for those types.

hence I had to redefine them. I thought about maybe upstreaming the changes by adding the tags :)

gwesterfieldjr: Ok cool thanks for background.

pkg/api/v1alpha1/tinkerbelltemplateconfig.go
pokearu: Yea I agree with what you are saying.

So I think the idea here was `NewTinkerbellTemplateConfigGenerate` is meant to be used while generating the "default" TemplateConfig as part of the `generate clusterconfig`. In that context we want to keep a single default task and add the default actions. Hence i added options only for actions, since we might want to add/remove those as we develop.

1. With a higher level `NewTinkerbellTemplateConfigGenerateOption` my concern was how do we feed it an action option? Like then we might need a TasksOpt and an ActionOpt to specify what task and what action to add.
2. I could do that. But do you mean like get the config object and then apply defaults?

pkg/api/v1alpha1/tinkerbellmachineconfig.go
pokearu: Nice catch! I did not notice that.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+const (
+	netplan = `network:
+  version: 2
+  renderer: networkd
+  ethernets:
+      eno1:
+          dhcp4: true
+      eno2:
```
panktishah26, line 15: Didn't we decide to add only one netplan Ethernet network to reduce the cluster creation time?

pokearu: Oh I was not sure we had tested that and made the change? I am sure it would work though.

pokearu: I updated the netplan as discussed 👍

PR #1355, opened 2022-03-02
Simplify and generalize ssh key generation

chrisdoherty4:
Improve ssh key generation algorithm ahead. The ssh key generator was being used by the vSphere provider only and the vSphere provider created it inline to generate a key. This change converts key generation to a much simpler function while maintaining backward compatibility with the file writer. Additionally, it removes the vSphere specific logging from the key generator into the vSphere provider so the key generation function can be re-used.

This change shed light on how inflexible the `filewriter.FileWriter` interface is. It defines a path context for the application's io operations but doesn't compliment the stdlib's `io` package making it difficult to isolate io. Watch this space for a design proposal.


ptrivedi:
FileWriter refactoring discussions have been brought up in the past as well. I don't see the changes for the Tinkerbell provider in here. Assuming that's coming in a different PR?

ptrivedi:
Also, IIRC, the previous key generator was not specific to vSphere, except for a recent log message that had been added. This looks fine too.

PR #1356, opened 2022-03-02
Move context aware validation out of api package

g-gaston:
This is part of the efforts started by https://github.com/aws/eks-anywhere/pull/1343:

...
- [x] Make sure all validations in the api/v1alpha1 package comply with the proposed definition of context unaware validations

...

**Changes on this PR:**

Validations made at the api package level should be lean, fast, free of
dependencies and context unaware.

This moves the registry mirror cert
validation out of the api package because it needs network connection
(this is done in order to check if the cert is self signed). It also
adds unit tests since the original code was missing them.

*Testing (if applicable):*
Added unit tests


pkg/crypto/validator.go
```diff
@@ -8,30 +8,24 @@ import (
...
-func (tv *DefaultTlsValidator) HasSelfSignedCert() (bool, error) {
+func (tv *DefaultTlsValidator) HasSelfSignedCert(host, port string) (bool, error) {
 	conf := &tls.Config{
 		InsecureSkipVerify: false,
 	}
 
-	_, err := tls.Dial("tcp", net.JoinHostPort(tv.host, tv.port), conf)
+	_, err := tls.Dial("tcp", net.JoinHostPort(host, port), conf)
 	if err != nil {
 		// If the error is x509.UnknownAuthorityError, this means the url is using self-signed certs
```
chrisdoherty4, line 30: I know this isn't the point of this PR, but I don't think this is accurate. Generally a certificate is self signed if the issuer and subject are the same but I think there are more complicated semantics to self-signed certs so using `x509` lib would be good.

We ought to verify its self signed by: (1) connect to acquire the cert ignoring any cert verification errors; (2) add the cert to a `CertPool`; (3) Use the `CertPool` as an option to `Certifiate.Verify()`.

See https://stackoverflow.com/a/63317898

g-gaston: Makes sense. To be honest, I just moved the the struct members to method arguments to it's easier to use. I didn't stop to look at the code.

Should we create an issue for this and put it in the pipeline? Maybe @abhinavmpandey08 would have more context about this

chrisdoherty4: Yea I think a GH issue would be good.

chrisdoherty4: Of course there's the possibility that this method isn't named appropriately and its actual name is more akin to `IsAuthorityKnown()`.

chrisdoherty4 (approved):

pkg/crypto/validator.go
g-gaston: @chrisdoherty4 feel free to add/change anything https://github.com/aws/eks-anywhere/issues/1368

PR #1360, opened 2022-03-03
Feature get package bundle

a-cool-train:
Implement get PackageBundle as part of the curated packages feature. 

### DIFF
--- a/cmd/eksctl-anywhere/cmd/get.go
+++ b/cmd/eksctl-anywhere/cmd/get.go
@@ -1,6 +1,18 @@
 package cmd
 
-import "github.com/spf13/cobra"
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/features"
+)
 
 var getCmd = &cobra.Command{
 	Use:   "get",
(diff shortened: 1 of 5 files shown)


cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -11,3 +25,49 @@ var getCmd = &cobra.Command{
...
+	}
+	kubectl := deps.Kubectl
+
+	params := []executables.KubectlOpt{executables.WithKubeconfig(kubeConfig), executables.WithArgs(args), executables.WithNamespace(constants.EksaPackagesName)}
+	if output != "" {
+		params = append(params, executables.WithOutput(output))
+	}
+	packages, err := kubectl.GetResources(ctx, resourceType, params...)
+	if err != nil {
+		fmt.Print(packages)
```
TerryHowe, line 56: I guess packages will never be nil?

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -11,3 +25,49 @@ var getCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(getCmd)
 }
+
+func preRunGetPackages(cmd *cobra.Command, args []string) error {
```
TerryHowe, line 29:
```suggestion
func preRunPackages(cmd *cobra.Command, args []string) error {
```

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -11,3 +25,49 @@ var getCmd = &cobra.Command{
...
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if err := viper.BindPFlag(flag.Name, flag); err != nil {
+			log.Fatalf("Error initializing flags: %v", err)
+		}
+	})
+	return nil
+}
+
+func getResources(ctx context.Context, resourceType string, output string, args []string) error {
+	kubeConfig := os.Getenv(kubeconfigEnvVariable)
```
TerryHowe, line 42: Chris just had a PR this week about a new kubeconfig manager

cmd/eksctl-anywhere/cmd/get.go
a-cool-train: Yes that's correct. It might be nil only if err is not nil in which case, we are already returning the err.

cmd/eksctl-anywhere/cmd/get.go
a-cool-train: Ok let me check on it.

a-cool-train: Ok moved the config setup to kubeconfig file from Chris's PR.

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -11,3 +25,49 @@ var getCmd = &cobra.Command{
...
+	}
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if err := viper.BindPFlag(flag.Name, flag); err != nil {
+			log.Fatalf("Error initializing flags: %v", err)
+		}
+	})
+	return nil
+}
+
+func getResources(ctx context.Context, resourceType string, output string, args []string) error {
```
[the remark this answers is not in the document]
a-cool-train: Yes the reason is because packageBundles and packages have similar code except the resourcetype. This is consistent with any future get operations. All the operations will differ by only the resourcetype to get.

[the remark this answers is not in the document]
vivek-koppuru: But when you run `-h`, we want the supported resources to get printed right? Which we make use of cobra's functionality to do that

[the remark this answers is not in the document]
vivek-koppuru: Disregard above comment, misunderstood that we were just refactoring

vivek-koppuru (approved):
/lgtm

PR #1363, opened 2022-03-03
Add harbor images under dev release

vincentni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


vincentni:
This is the temporary solution so we just need to take care of manual promotion of harbor helm chart. It will be taken out once our own pipeline is ready.

PR #1364, opened 2022-03-03
Cluster Config validation

g-gaston:
This continues the work started on #1343

*Description of changes:*
Context unaware validations rely on having a Validate method in the api
structs.

This adds the option to run context aware validations at the cluster
Config level. These should have no external dependencies and depend
solely on the data in cluster Config.

### DIFF
--- a/pkg/api/v1alpha1/awsiamconfig.go
+++ b/pkg/api/v1alpha1/awsiamconfig.go
@@ -17,10 +17,16 @@ func GetAndValidateAWSIamConfig(fileName string, refName string, clusterConfig *
 	if err != nil {
 		return nil, err
 	}
-	err = validateAWSIamConfig(config, refName, clusterConfig)
-	if err != nil {
+	if err = validateAWSIamConfig(config); err != nil {
 		return nil, err
 	}
+	if err = validateAWSIamRefName(config, refName); err != nil {
+		return nil, err
+	}
+	if err = validateAWSIamNamespace(config, clusterConfig); err != nil {
+		return nil, err
+	}
+
 	return config, nil
 }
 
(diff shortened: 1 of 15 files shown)


pkg/api/v1alpha1/awsiamconfig.go
```diff
@@ -98,3 +98,28 @@ func validateMapUsers(mapUsers []MapUsers) error {
...
 	return nil
 }
+
+func validateAWSIamRefName(config *AWSIamConfig, refName string) error {
+	if config == nil {
+		return nil
+	}
+
+	if config.Name != refName {
+		return fmt.Errorf("AWSIamConfig retrieved with name %v does not match name (%v) specified in "+
```
jiayiwang7, line 108: nit:
```suggestion
		return fmt.Errorf("AWSIamConfig retrieved with name %s does not match name (%s) specified in "+
```

pkg/api/v1alpha1/gitopsconfig.go
```diff
@@ -79,3 +78,28 @@ func validateGitRepoName(repoName string) error {
...
 	return nil
 }
+
+func validateGitOpsRefName(config *GitOpsConfig, refName string) error {
+	if config == nil {
+		return nil
+	}
+
+	if config.Name != refName {
+		return fmt.Errorf("GitOpsConfig retrieved with name %v does not match name (%v) specified in "+
```
jiayiwang7, line 88: nit:
```suggestion
		return fmt.Errorf("GitOpsConfig retrieved with name %s does not match name (%s) specified in "+
```

pkg/api/v1alpha1/oidcconfig.go
```diff
@@ -64,3 +64,28 @@ func validateOIDCConfig(config *OIDCConfig, refName string, clusterConfig *Clust
...
 	return nil
 }
+
+func validateOIDCRefName(config *OIDCConfig, refName string) error {
+	if config == nil {
+		return nil
+	}
+
+	if config.Name != refName {
+		return fmt.Errorf("OIDCConfig retrieved with name %v does not match name (%v) specified in "+
```
jiayiwang7, line 74: nit: `%s`

pkg/api/v1alpha1/awsiamconfig.go
```diff
@@ -98,3 +98,28 @@ func validateMapUsers(mapUsers []MapUsers) error {
...
+
+	if config.Name != refName {
+		return fmt.Errorf("AWSIamConfig retrieved with name %s does not match name (%s) specified in "+
+			"identityProviderRefs", config.Name, refName)
+	}
+
+	return nil
+}
+
+func validateAWSIamNamespace(config *AWSIamConfig, clusterConfig *Cluster) error {
```
jiayiwang7, line 115: this only reason this function exists is to keep current validation logic working? eventually we dont need this right? since it's taken care of in cluster.validation

g-gaston: Yeah
My plan was to deprecate all the "parsing" methods in this package once we start using the `cluster.Config` in the `cluster.Spec`
And once that's done, the RefName and Namespace validations won't be necessary in this package

jiayiwang7 (approved):

PR #1366, opened 2022-03-04
Tinkerbell hardware workflow mapping - pre flight validation

panktishah26:
*Description of changes:*

This PR is a part of pre-flight validation for Tinkerbell provider. Before creating a cluster our CLI will check whether a workflow exists in the Tinkerbell stack or not. It will fail the cluster creation if there is any workflows associated with any of the hardwares provided in the hardware yaml file.

### DIFF
--- a/pkg/executables/tink_test.go
+++ b/pkg/executables/tink_test.go
@@ -53,7 +53,7 @@ func TestTinkGetHardware(t *testing.T) {
 	}
 }
 
-func TestTinkWorkflow(t *testing.T) {
+func TestTinkGetWorkflow(t *testing.T) {
 	tink, ctx, e := newTink(t)
 	expectedParam := []string{"workflow", "get", "--format", "json"}
 	expectCommand(e, ctx, expectedParam...).withEnvVars(envMap).to().Return(bytes.Buffer{}, nil)
--- a/pkg/executables/tink_test.go
+++ b/pkg/executables/tink_test.go
@@ -52,3 +52,12 @@ func TestTinkGetHardware(t *testing.T) {
 		t.Errorf("Tink.GetHardware() error = %v, want nil", err)
 	}
 }
+
+func TestTinkWorkflow(t *testing.T) {
+	tink, ctx, e := newTink(t)
+	expectedParam := []string{"workflow", "get", "--format", "json"}
+	expectCommand(e, ctx, expectedParam...).withEnvVars(envMap).to().Return(bytes.Buffer{}, nil)
+	if _, err := tink.GetWorkflow(ctx); err != nil {
+		t.Errorf("Tink.GetWorkflow() error = %v, want nil", err)
+	}
+}
--- a/pkg/providers/tinkerbell/hardware/config.go
+++ b/pkg/providers/tinkerbell/hardware/config.go
@@ -92,9 +92,11 @@ func (hc *HardwareConfig) ValidateHardware(skipPowerActions bool, tinkHardwareMa
 
 		hardwareInterface := tinkHardwareMap[hw.Spec.ID].GetNetwork().GetInterfaces()
 		if len(hardwareInterface) > 0 {
-			mac := hardwareInterface[0].GetDhcp()
-			if _, ok := tinkWorkflowMap[mac.Mac]; ok {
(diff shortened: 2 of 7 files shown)


pkg/providers/tinkerbell/validator.go
```diff
@@ -283,3 +295,22 @@ func hardwareMap(hardwareList []*tinkhardware.Hardware) map[string]*tinkhardware
...
+	workflowMap := make(map[string]*tinkworkflow.Workflow)
+
+	for _, data := range workflowList {
+		var macAddress map[string]string
+
+		if err := json.Unmarshal([]byte(data.GetHardware()), &macAddress); err != nil {
+			return nil, fmt.Errorf("error unmarshling workflow data: %v", err)
+		}
+
+		if _, ok := macAddress["device_1"]; ok {
```
pokearu, line 310: Just to be sure, this would always be `device _1` right? Even if we had different templates for CP and worker?

pkg/providers/tinkerbell/hardware/config.go
```diff
@@ -89,6 +90,14 @@ func (hc *HardwareConfig) ValidateHardware(skipPowerActions bool, tinkHardwareMa
 			return fmt.Errorf("hardware id '%s' is not registered with tinkerbell stack", hw.Spec.ID)
 		}
 
+		hardwareInterface := tinkHardwareMap[hw.Spec.ID].GetNetwork().GetInterfaces()
+		if len(hardwareInterface) > 0 {
+			mac := hardwareInterface[0].GetDhcp()
```
pokearu, line 95: Maybe this is a broader discussion, but can we assume that the Mac of the first interface is what the workflow will map with? Not sure what happens if we specify more interfaces.

pkg/executables/tink_test.go
```diff
@@ -52,3 +52,12 @@ func TestTinkGetHardware(t *testing.T) {
 		t.Errorf("Tink.GetHardware() error = %v, want nil", err)
 	}
 }
+
+func TestTinkWorkflow(t *testing.T) {
```
pokearu, line 56: nit:

```suggestion
func TestTinkGetWorkflow(t *testing.T) {
```

pkg/providers/tinkerbell/validator.go
```diff
@@ -283,3 +295,22 @@ func hardwareMap(hardwareList []*tinkhardware.Hardware) map[string]*tinkhardware
 
 	return hardwareMap
 }
+
+// workflowMap returns all the workflows on the tinkerbell stack in the form of map with mac address as a key
+func workflowMap(workflowList []*tinkworkflow.Workflow) (map[string]*tinkworkflow.Workflow, error) {
```
pokearu, line 300: nit: do the same for hardwareMap? Since both are returning a map it might add more clarity. 
```suggestion
func getWorkflowMap(workflowList []*tinkworkflow.Workflow) (map[string]*tinkworkflow.Workflow, error) {
```

pokearu:
Overall looks good 🚀

pkg/providers/tinkerbell/validator.go
panktishah26: I am not getting the mac address with the key now. So we are good here. Thanks for the suggestion.

pkg/providers/tinkerbell/validator.go
```diff
@@ -283,3 +295,21 @@ func hardwareMap(hardwareList []*tinkhardware.Hardware) map[string]*tinkhardware
 
 	return hardwareMap
 }
+
+// getWorkflowMap returns all the workflows on the tinkerbell stack in the form of map with mac address as a key
+func getWorkflowMap(workflowList []*tinkworkflow.Workflow) (map[string]*tinkworkflow.Workflow, error) {
+	workflowMap := make(map[string]*tinkworkflow.Workflow)
+
+	for _, data := range workflowList {
+		var macAddress map[string]string
```
abhinavmpandey08, line 304: nit: maybe call it `macAddressMap` or `macAddresses`
```suggestion
		var macAddressMap map[string]string
```

panktishah26: Agree, I will change the name in my next PR as this one is merged. Thanks!

PR #1388, opened 2022-03-07
Remove superfluous cluster name param in kubeconfig error

chrisdoherty4:
The `missingFileError` struct doesn't use the cluster name and it doesn't need to. This removes it.


TerryHowe (approved):
/lgtm

PR #1391, opened 2022-03-07
Embed eksd manifest to cluster

taneyland:
*Issue #, if available:*
#410 

*Description of changes:*
These changes adds a step to install eksd components to the cluster during cluster creation & upgrade. This will allow the controller to pull the eksd manifest directly from the cluster 

[Design doc](https://github.com/aws/eks-anywhere/blob/main/designs/embed_eksd_manifest.md)

*Testing (if applicable):*
`create cluster` & `upgrade cluster`. Viewed controller logs using Tilt


internal/test/cluster.go
```diff
@@ -88,6 +89,34 @@ func Bundles(t *testing.T) *releasev1alpha1.Bundles {
 	return bundles
 }
 
+func EksdRelease(t *testing.T) *eksdv1alpha1.Release {
```
g-gaston, line 92: nit: you can add this to improve logs

```suggestion
func EksdRelease(t *testing.T) *eksdv1alpha1.Release {
t.Helper()
```

g-gaston, line 92: nit: you can add this to improve logs

```suggestion
func EksdRelease(t *testing.T) *eksdv1alpha1.Release {
  t.Helper()
```

internal/test/cluster.go
```diff
@@ -88,6 +89,34 @@ func Bundles(t *testing.T) *releasev1alpha1.Bundles {
...
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("Failed getting path to current file")
+	}
+
+	templateValues := map[string]string{
+		"TestPath": filepath.Dir(filename),
+	}
+
+	eksdContent, err := templater.Execute(string(content), templateValues)
```
g-gaston, line 107: Do we need this?

g-gaston, line 107: Do we need this?

pkg/cluster/spec.go
```diff
@@ -277,19 +277,14 @@ func NewSpecFromClusterConfig(clusterConfigPath string, cliVersion version.Info,
 	return s, nil
 }
 
-func BuildSpecFromBundles(cluster *eksav1alpha1.Cluster, bundles *v1alpha1.Bundles, opts ...SpecOpt) (*Spec, error) {
+func BuildSpecFromBundles(cluster *eksav1alpha1.Cluster, bundles *v1alpha1.Bundles, eksd *eksdv1alpha1.Release, opts ...SpecOpt) (*Spec, error) {
```
g-gaston, line 280: Can we pass the eksd release in an option?

And then have a nil check and if the eksd release is not there, then retrieve it

g-gaston, line 280: Can we pass the eksd release in an option?

And then have a nil check and if the eksd release is not there, then retrieve it

pkg/cluster/spec.go
```diff
@@ -491,6 +487,28 @@ func userAgent(eksAComponent, version string) string {
 	return fmt.Sprintf("eks-a-%s/%s", eksAComponent, version)
 }
 
+type EksdRelease struct {
+	ReleaseManifestContent []byte
+	ReleaseCrdContent      []byte
+}
+
+func (s *Spec) GetEksdRelease(release v1alpha1.EksDRelease) (*EksdRelease, error) {
```
g-gaston, line 495: This method signature is a bit confusing

Should this be something like

```suggestion
func (s *Spec) ReadEksdManifests(release v1alpha1.EksDRelease) (*EksdManifests, error) {
```

g-gaston, line 495: This method signature is a bit confusing

Should this be something like

```suggestion
func (s *Spec) ReadEksdManifests(release v1alpha1.EksDRelease) (*EksdManifests, error) {
```

pkg/executables/kubectl.go
```diff
@@ -1142,6 +1143,22 @@ func (k *Kubectl) ValidateNodesVersion(ctx context.Context, kubeconfig string, k
 	return nil
 }
 
+func (k *Kubectl) GetEksdRelease(ctx context.Context, name, namespace, kubeconfigFile string) (*eksdv1alpha1.Release, error) {
+	params := []string{"get", "release", name, "-o", "json", "--kubeconfig", kubeconfigFile, "--namespace", namespace}
```
g-gaston, line 1147: can we use the `getObject` helper?

g-gaston, line 1147: can we use the `getObject` helper?

pkg/workflows/create.go
```diff
@@ -247,13 +249,30 @@ func (s *MoveClusterManagementTask) Run(ctx context.Context, commandContext *tas
...
+	return &InstallEksdComponentsTask{}
 }
 
 func (s *MoveClusterManagementTask) Name() string {
 	return "capi-management-move"
 }
 
+// InstallEksdComponentsTask implementation
+
+func (s *InstallEksdComponentsTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
```
g-gaston, line 261: Does this need to be a new task?

In fact, can this just be part of install eksa components in cluster manager? This not different than installing the bundle

g-gaston, line 261: Does this need to be a new task?

In fact, can this just be part of install eksa components in cluster manager? This not different than installing the bundle

pkg/workflows/create.go
```diff
@@ -247,13 +249,30 @@ func (s *MoveClusterManagementTask) Run(ctx context.Context, commandContext *tas
...
+	if !commandContext.BootstrapCluster.ExistingManagement {
```
g-gaston, line 262: this is problematic

What happens if this is a workload cluster but we are installing a new eksd version?

g-gaston, line 262: this could be problematic

What happens if this is a workload cluster but we are installing a new eksd version?

pkg/cluster/fetch.go
```diff
@@ -4,17 +4,22 @@ import (
...
 type BundlesFetch func(ctx context.Context, name, namespace string) (*v1alpha1release.Bundles, error)
 
 type GitOpsFetch func(ctx context.Context, name, namespace string) (*v1alpha1.GitOpsConfig, error)
 
+type EksdReleaseFetch func(ctx context.Context, name, namespace string) (*eksdv1alpha1.Release, error)
+
 type OIDCFetch func(ctx context.Context, name, namespace string) (*v1alpha1.OIDCConfig, error)
 
-func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, gitOpsFetch GitOpsFetch, oidcFetch OIDCFetch) (*Spec, error) {
+func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, oidcFetch OIDCFetch) (*Spec, error) {
```
g-gaston, line 22: We might want to refactor this to take an interface, the number of arguments is starting to be quite big. This could be done separately

g-gaston, line 22: We might want to refactor this to take an interface, the number of arguments is starting to be quite big

controllers/controllers/clusters/vsphere.go
```diff
@@ -110,8 +112,19 @@ func (v *VSphereClusterReconciler) bundles(ctx context.Context, name, namespace
 	return clusterBundle, nil
 }
 
+func (v *VSphereClusterReconciler) eksdRelease(ctx context.Context, name, namespace string) (*eksdv1alpha1.Release, error) {
```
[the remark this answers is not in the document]
taneyland: Sure. This change will affect the `BuildSpecForCluster` method. Since refactoring that is a follow-up task here, I would prefer to make that change in that separate PR

pkg/cluster/fetch.go
```diff
@@ -51,6 +60,28 @@ func GetGitOpsForCluster(ctx context.Context, cluster *v1alpha1.Cluster, fetch G
...
+	}
+	eksd, err := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed fetching EKS-D release for cluster: %v", err)
+	}
+
+	return eksd, nil
+}
+
+func getVersionsBundle(clusterConfig *v1alpha1.Cluster, bundles *v1alpha1release.Bundles) (*v1alpha1release.VersionsBundle, error) {
```
g-gaston, line 76: nit: could we make this method public and use it from the controller?

pkg/executables/kubectl.go
```diff
@@ -1323,6 +1324,15 @@ func (k *Kubectl) getObject(ctx context.Context, resourceType, name, namespace,
 	return nil
 }
 
+func (k *Kubectl) GetEksdRelease(ctx context.Context, name, namespace, kubeconfigFile string) (*eksdv1alpha1.Release, error) {
+	obj := &eksdv1alpha1.Release{}
+	if err := k.getObject(ctx, "release", name, namespace, kubeconfigFile, obj); err != nil {
```
g-gaston, line 1329: nit: maybe we should specify the group as well with "release" so it know how to distinguish from eks-a and eks-d releases

pkg/clustermanager/retrier_client.go
```diff
@@ -23,6 +24,31 @@ func NewRetrierClient(client *client, retrier *retrier.Retrier) *retrierClient {
...
 
+func (c *retrierClient) installEksdComponents(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error {
+	eksdComponents, err := clusterSpec.ReadEksdManifests(clusterSpec.VersionsBundle.EksD)
+	if err != nil {
+		return fmt.Errorf("failed loading manifest for eksd components: %v", err)
+	}
+
+	if err = c.Retry(
+		func() error {
+			return c.ApplyKubeSpecFromBytesWithNamespace(ctx, cluster, eksdComponents.ReleaseCrdContent, constants.EksaSystemNamespace)
```
g-gaston, line 35: I wonder if this should be part of the "upgrade core components" or not. wdy?

taneyland: I wondered the same thing. Since this method is called in both create and upgrade, if there's an update to the eksd manifest in the bundle, it'll be applied to the cluster here. So i don't think it's necessary to add it to the core components upgrades

g-gaston: Oh, I'm not saying it is necessary. But it's more about concern separation.
The CRDs for eks-a get updated in the core component upgrades. And then we later apply the changes to the eks-a objects.

I was wondering if we should follow the same pattern here: upgrade the CRDs in core component upgrades and then later after the cluster upgrade has been done, create the new ekd-d release object (if there is a new one).

g-gaston (approved):
/lgtm

pkg/clustermanager/retrier_client.go
taneyland: #1436

PR #1394, opened 2022-03-07
Use bundlesRef instead of eksdReleaseRef to get eksd bundle

taneyland:
*Issue #, if available:*

*Description of changes:*
Instead of using `eksdReleaseRef` in the cluster status, we will fetch the eks-d object using the eks-d bundle name from the versions bundle. We already have access to this bundle. However, we currently grab it based on the cluster name, which is usually the case, but a more accurate way to get this object is by introducing a `bundlesRef` & directly referring to that object instead of assuming the name based on the cluster

*Testing (if applicable):*

### DIFF
--- a/designs/embed_eksd_manifest.md
+++ b/designs/embed_eksd_manifest.md
@@ -71,10 +71,10 @@ spec:
     apiGroup: anywhere.eks.amazonaws.com/v1alpha1
     kind: Bundles
     name: "bundle-name"
-    namespace: "namespace for resource" 
+    namespace: eksa-system 
 ```
 
-This bundles ref will allow us accurately fetch the bundle from the cluster, access the EKSD bundle, and grab the correct release manifest in the controller.
+This bundles ref will allow us to accurately fetch the bundle from the cluster, access the EKSD bundle, and grab the correct release manifest in the controller.
 
 ### Upgrading EKSD components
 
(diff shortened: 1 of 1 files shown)


designs/embed_eksd_manifest.md
```diff
@@ -61,15 +67,14 @@ metadata:
...
-    namespace: "namespace for resource" (default: eksa-system)
+  bundlesRef:
+    apiGroup: anywhere.eks.amazonaws.com/v1alpha1
+    kind: Bundles
+    name: "bundle-name"
+    namespace: "namespace for resource" 
 ```
 
-From here, we can call `kubectl get release` using the `eksdReleaseRef.Name` to access the EKSD release manifest from the controller.
+This bundles ref will allow us accurately fetch the bundle from the cluster, access the EKSD bundle, and grab the correct release manifest in the controller.
```
abhay-krishna, line 77:
```suggestion
This bundles ref will allow us to accurately fetch the bundle from the cluster, access the EKSD bundle, and grab the correct release manifest in the controller.
```

designs/embed_eksd_manifest.md
```diff
@@ -61,15 +67,14 @@ metadata:
...
-  eksdReleaseRef:
-    apiGroup: distro.eks.amazonaws.com/v1alpha1
-    kind: Release
-    name: "name of resource"
-    namespace: "namespace for resource" (default: eksa-system)
+  bundlesRef:
+    apiGroup: anywhere.eks.amazonaws.com/v1alpha1
+    kind: Bundles
+    name: "bundle-name"
+    namespace: "namespace for resource"
```
g-gaston, line 74: This should be `eksa-system` as well imo

g-gaston (approved):

PR #1395, opened 2022-03-07
Declare ssh keys persistent

chrisdoherty4:
Fix generated key path so it doesn't appear int he `generated` file and is considered 'permanent'.


vivek-koppuru (approved):
/lgtm

PR #1403, opened 2022-03-08
Add tinkerbell test cases for supported hardware vendors, cp replicas…

gwesterfieldjr:
…, and worker replicas, and external etcd

- Add hardware specific tests to check for vendor 
- Add test for multiple control planes and worker nodes
- Add test case for external etcd
- Fix for external etcd found by external etcd e2e test
- Add option to provide output location of hardware-manifests
- Output hardware manifests to cluster config folder per test so tests can be ran in parallel from separate shells if desired in the future

*Testing (if applicable):*

E2E Tests

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generatehardware.go
+++ b/cmd/eksctl-anywhere/cmd/generatehardware.go
@@ -20,6 +20,7 @@ import (
 
 type hardwareOptions struct {
 	csvPath      string
+	outputPath   string
 	tinkerbellIp string
 	grpcPort     string
 	certPort     string
@@ -50,6 +51,7 @@ var generateHardwareCmd = &cobra.Command{
 func init() {
 	generateCmd.AddCommand(generateHardwareCmd)
 	generateHardwareCmd.Flags().StringVarP(&hOpts.csvPath, "filename", "f", "", "path to csv file")
+	generateHardwareCmd.Flags().StringVarP(&hOpts.outputPath, "output", "o", "", "directory path to output hardware files")
 	generateHardwareCmd.Flags().StringVar(&hOpts.tinkerbellIp, "tinkerbell-ip", "", "Tinkerbell stack IP, required unless --dry-run flag is set")
 	generateHardwareCmd.Flags().StringVar(&hOpts.grpcPort, "grpc-port", defaultGrpcPort, "Tinkerbell GRPC Authority port")
 	generateHardwareCmd.Flags().StringVar(&hOpts.certPort, "cert-port", defaultCertPort, "Tinkerbell Cert URL port")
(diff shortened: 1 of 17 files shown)


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -465,6 +465,18 @@ func (p *tinkerbellProvider) MachineConfigs() []providers.MachineConfig {
 			p.machineConfigs[workerMachineName].SetManagedBy(p.clusterConfig.ManagedBy())
 		}
 	}
+
+	if p.clusterConfig.Spec.ExternalEtcdConfiguration != nil {
```
gwesterfieldjr, line 469: This is to fix external etcd cluster creation failure

cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -50,6 +51,7 @@ var generateHardwareCmd = &cobra.Command{
 func init() {
 	generateCmd.AddCommand(generateHardwareCmd)
 	generateHardwareCmd.Flags().StringVarP(&hOpts.csvPath, "filename", "f", "", "path to csv file")
+	generateHardwareCmd.Flags().StringVarP(&hOpts.outputPath, "output", "o", "", "directory path to output hardware files")
```
gwesterfieldjr, line 54: Added option to export hardware bundle to the directory provided. This is needed to export the hardware.yaml to cluster config directory during test run.

gwesterfieldjr: Also, I think this is a nice to have option as customer may desire to export to another directory as well

abhinavmpandey08 (approved):
/lgtm

cmd/eksctl-anywhere/cmd/generatehardware.go
chrisdoherty4: Should we go the full mile and let them specify the full file path (i.e. including file name)? Feels like the flexibility we're going for here.

internal/pkg/api/hardware.go
```diff
@@ -0,0 +1,45 @@
...
+	Netmask      string `csv:"netmask"`
+	MacAddress   string `csv:"mac"`
+	Hostname     string `csv:"hostname"`
+	Vendor       string `csv:"vendor"`
+	BmcIpAddress string `csv:"bmc_ip"`
+	BmcUsername  string `csv:"bmc_username"`
+	BmcPassword  string `csv:"bmc_password"`
+}
+
+func NewHardwareSlice(csvFile string) ([]*Hardware, error) {
```
chrisdoherty4, line 30: I would name this something other than `New*` which are generally just initializing the return. 

```
func HardwareFromFile(...)
```

test/framework/tinkerbell.go
```diff
@@ -62,13 +63,20 @@ func NewTinkerbell(t *testing.T, opts ...TinkerbellOpt) *Tinkerbell {
 		},
 	}
 
+	tink.cidr = os.Getenv(tinkerbellNetworkCidrEnvVar)
```
chrisdoherty4, line 66: Not something to be addressed here as I've seen it throughout our code base but we want to avoid arbitrarily reading the environment, including in tests. 

Mutating and retrieving the env mid program can result in accidental behavior and make the code inflexible. Configuration is a concern in its own right and should be handled as such.

PR #1405, opened 2022-03-08
Increase e2e instance count to 85

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Our sequential e2e logic doesn't work as expected right now and fails to upload the JUnit reports due to how the test name is constructed. It could also be that some of the flux tests are failing due to resources not being cleaned up. Changing it back to a higher number to get a test run in and see if we are going to run into the resource limits again.

*Testing (if applicable):*


jaxesn (approved):
/lgtm

PR #1412, opened 2022-03-08
Fix junit file upload/download

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Latest changes here cause the files to be uploaded/downloaded with the string quotes. https://github.com/aws/eks-anywhere/pull/1400

There will still be an issue running multiple tests in the same instance as the test name would be `test1|test2`, which will result in the junit reports to not upload properly, so will be following up with a fix for that.

*Testing (if applicable):*


abhay-krishna (approved):
/lgtm

jaxesn (approved):
/lgtm

PR #1413, opened 2022-03-09
Compare only semver and prerelease information when overwriting releases

abhay-krishna:
This comparison is done to determine whether to overwrite an existing release or append a new one to the list of current EKS-A releases. It won't be proper to compare the whole release version, for example, `v0.0.0-dev+build.100` since the build metadata keeps monotonically increasing. Instead it would suffice to just compare the part before the `+`, i.e. semver and pre-release information.


g-gaston (approved):

vivek-koppuru (approved):
/lgtm

PR #1415, opened 2022-03-09
Improve missing hardwarefile flag error speed

chrisdoherty4:
The `hardwarefile` flag is required when creating a cluster with a tinkerbell provider. We noticed the 'missing flag' error was taking unnecessarily long to appear.

The flag can't be modeled as required using Cobra because its only required if the cluster configuration uses a `TinkerbellDatacenterConfig`. Consequently, we need to establish what provider is in use with the `create cluster` command via the yaml before we can make the flag required. The cause for the delay (approximately 2s) stems from the app performing disk io, network io, sub process management and validation activities.

The launching of child processes and performing network io (the majority of the wait) was performed by a bulk `commonValidations` function that's shared between sub commands. This function makes it difficult to model a succinct bootstrapping process for sub command implementations so I took the liberty to break away from it to model the steps in a more atomic fashion. In doing so I could rejig the order of events the command executes and got the error down to milliseconds.

The change here isn't the most idiomatic. We need a follow up to define bootstrapping for CLI commands. As it stands there doesn't seem to be anything detailing a clear bootstrapping order anywhere.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -1,16 +1,18 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"log"
+	"runtime"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 	"github.com/aws/eks-anywhere/pkg/types"
@@ -35,15 +37,7 @@ var createClusterCmd = &cobra.Command{
 	Long:         "This command is used to create workload clusters",
 	PreRunE:      preRunCreateCluster,
 	SilenceUsage: true,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		if err := cc.validate(cmd.Context()); err != nil {
-			return err
-		}
-		if err := cc.createCluster(cmd); err != nil {
-			return fmt.Errorf("failed to create cluster: %v", err)
-		}
-		return nil
-	},
+	RunE:         cc.createCluster,
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -73,12 +71,51 @@ func preRunCreateCluster(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("the cluster config file provided is invalid: %v", err)
+	}
+
+	if clusterConfig.Spec.DatacenterRef.Kind == v1alpha1.TinkerbellDatacenterKind {
+		flag := cmd.Flags().Lookup("hardwarefile")
+
+		// If no flag was returned there is a developer error as the flag has been removed
+		// from the program.
+		if flag == nil {
+			panic("'hardwarefile' flag not configured")
```
abhinavmpandey08, line 93: why panic instead of returning an error here?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -114,24 +145,6 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command) error {
...
-		}
-		hardwareConfigFileExist := validations.FileExists(cc.hardwareFileName)
-		if !hardwareConfigFileExist {
-			return fmt.Errorf("Error: hardware config file %s does not exist", cc.hardwareFileName)
-		}
-	}
-
-	if !features.IsActive(features.CloudStackProvider()) && deps.Provider.Name() == constants.CloudStackProviderName {
-		return fmt.Errorf("error: provider cloudstack is not supported in this release")
-	}
```
abhinavmpandey08, line 133: why remove this?

abhinavmpandey08 (commented):
looks great! just a couple of comments

cmd/eksctl-anywhere/cmd/createcluster.go
chrisdoherty4: If the flag is missing and we expect it there's a programmer error. I.e the program is invalid which is cause for a panic. To paint it with a different perspective, the user can't do anything about this problem. The binary is totally useless to them.

cmd/eksctl-anywhere/cmd/createcluster.go
chrisdoherty4: Oops. Looks like a merge mistake, my bad. Will fix in the morning.

chrisdoherty4: Fixed

abhinavmpandey08 (approved):
/lgtm

PR #1418, opened 2022-03-09
Fix deadlock in e2e tests

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
The way we were using two channels caused a deadlock as demonstrated in this example here: https://go.dev/play/p/Vpyv-cGmoSD

I went ahead and changed it to make our queue be the only channel as we don't really need a results channel anymore to keep track of go routines. 

*Testing (if applicable):*
Tested with example and ran `make integration-test-binary`

Example: https://go.dev/play/p/XEuUlyBQ0jq


abhay-krishna (approved):
/lgtm

PR #1419, opened 2022-03-09
Modify resizing disk for Bottlerocket templates based on Kube version

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
The latest version of Bottlerocket for 1.22 only has a single disk, which removes our need to resize the second disk, but we found out we still need to resize the first disk. We are adding this conditional as a work around for now until we figure out the root cause of this issue.

This shouldn't merge until we test with Bottlerocket once merging and getting new bundle from https://github.com/aws/eks-anywhere-build-tooling/pull/536

/hold

*Testing (if applicable):*
`make`
Running functional testing

### DIFF
--- a/pkg/executables/govc.go
+++ b/pkg/executables/govc.go
@@ -295,17 +295,20 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
 			}
 		}
 		diskName := ""
+		var diskSizeInGB int
 		if disk2 != "" {
 			logger.V(4).Info("Resizing disk 2 of template to 20G")
 			diskName = disk2
+			diskSizeInGB = 20
 		} else if disk1 != "" {
-			logger.V(4).Info("Resizing disk 1 of template to 20G")
+			logger.V(4).Info("Resizing disk 1 of template to 22G")
 			diskName = disk1
+			diskSizeInGB = 22
 		} else {
 			return fmt.Errorf("template %v is not valid as there are no associated disks", templateName)
 		}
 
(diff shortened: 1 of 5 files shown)


vivek-koppuru:
/unhold

Finished testing

jaxesn (approved):
/lgtm

pkg/executables/govc.go
```diff
@@ -263,33 +263,52 @@ func (g *Govc) CreateLibrary(ctx context.Context, datastore, library string) err
...
+			logger.V(4).Info("Resizing disk 2 of template to 20G")
+			diskName = disk2
+		} else if disk1 != "" {
+			logger.V(4).Info("Resizing disk 1 of template to 20G")
+			diskName = disk1
+		} else {
+			return fmt.Errorf("template %v is not valid as there are no associated disks", templateName)
+		}
+
+		err = g.ResizeDisk(ctx, datacenter, templateName, diskName, 20)
```
vignesh-goutham, line 308: If this is `disk1` aka `unified disk` we might need to resize to 22G vs 20G. I'm not entirely sure if this will affect the OS operation, probably not. But this might help to stay consistent with OVF on the image. If not, having a comment here should suffice.

vignesh-goutham (approved):

pkg/executables/govc.go
vivek-koppuru: Good point, I would rather be in line with the ovf. Later we definitely need to read the ovf file to get the values, but I think the main solution to all of this is figure out how to make VSphere honor the ovf values.

vignesh-goutham (approved):
Looks great! Thanks

/lgtm

PR #1420, opened 2022-03-09
Adding account/domain to the capc template

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR enables us to pass the specified account/domain down to CAPC. We were already consuming it in the EKSA cluster spec and doing preflight checks on it, but never passing it to CAPC

*Testing (if applicable):*
Successfully created cluster with all machines in specified account/domain, unlike the prior behavior where they defaulted to admin and root. 


vignesh-goutham (approved):
Looks good. 

/lgtm

Issue #1105, opened 2022-01-31
CAPAS - CLI create workflow


jiayiwang7:
Issue to track Snow support for CLI create workflow

- [x] Define the CAPAS templates EKS-A spec converts to
- [x] Extra objects to be installed that are specific to Snow as post-creation installation (private registry, etc.)
- [x] bare minimum create workflow 
- [x] #1573

PR #1421, opened 2022-03-09
Add unit tests for snow provider and apiBuilder

jiayiwang7:
*Issue #, if available:*

#1105 

*Description of changes:*

*Testing (if applicable):*


pkg/providers/snow/snow_test.go
```diff
@@ -0,0 +1,200 @@
...
+	tt := newSnowTest(t)
+	tt.Expect(tt.provider.Name()).To(Equal(expectedSnowProviderName))
+}
+
+func TestSetupAndValidateCreateClusterSuccess(t *testing.T) {
+	tt := newSnowTest(t)
+	setupContext(t, credsFileEnvVar)
+	setupContext(t, certsFileEnvVar)
+	os.Setenv(credsFileEnvVar, credsFilePath)
+	os.Setenv(certsFileEnvVar, certsFilePath)
```
g-gaston, line 125: Do we need to call these two methods twice?

pkg/providers/snow/snow_test.go
```diff
@@ -0,0 +1,200 @@
...
+	err := tt.provider.SetupAndValidateCreateCluster(tt.ctx, tt.clusterSpec)
+	tt.Expect(err).To(Succeed())
+}
+
+func TestSetupAndValidateCreateClusterNoCredsEnv(t *testing.T) {
+	tt := newSnowTest(t)
+	setupContext(t, credsFileEnvVar)
+	setupContext(t, certsFileEnvVar)
+	os.Setenv(certsFileEnvVar, certsFilePath)
```
g-gaston, line 134: Should this be called inside `setupContext`?

pkg/providers/snow/snow_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+func TestSetupAndValidateCreateClusterNoCertsEnv(t *testing.T) {
+	tt := newSnowTest(t)
+	setupContext(t, credsFileEnvVar)
+	os.Setenv(credsFileEnvVar, credsFilePath)
+	err := tt.provider.SetupAndValidateCreateCluster(tt.ctx, tt.clusterSpec)
+	tt.Expect(err).To(MatchError(ContainSubstring("EKSA_SNOW_DEVICES_CA_BUNDLES_FILE is not set or is empty")))
+}
+
+// TODO: add more tests (multi worker node groups, unstacked etcd, etc.)
```
g-gaston, line 147: Suggestion: maybe keep this method tests at a minimum and just test the methods that return api structs? That avoids having to create a ton of yaml files and probably makes tests more "unitary"

pkg/providers/snow/snow_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+type apiObjects struct {
+	cluster          *v1alpha1.Cluster
+	clusterSpec      *cluster.Spec
+	datacenterConfig *v1alpha1.SnowDatacenterConfig
+	machineConfigs   map[string]*v1alpha1.SnowMachineConfig
+}
+
+func givenFullAPIObjects(t *testing.T, fileName string) apiObjects {
+	clusterSpec := test.NewFullClusterSpec(t, path.Join(testDataDir, fileName))
```
g-gaston, line 65: Can we build this without using a yaml file? Just defining structs in code
I know that vsphere does it this way but I think structs will be more flexible

g-gaston (commented):
I would add some tests for the snow apibuilder methods in the same way you did for the generic clusterapi ones

pkg/providers/snow/snow_test.go
jiayiwang7: not sure if i 100% agree. this method returns cp and md yaml templates in []byte. it makes sense to test against yamls. 
also the api struct test is covered in the two apibuilder_test

jiayiwang7:
> I would add some tests for the snow apibuilder methods in the same way you did for the generic clusterapi ones

added.

g-gaston (approved):
/lgtm

PR #1426, opened 2022-03-09
Update client.authorization to v1beta1

mdsgabriel:
*Issue #, if available:*

*Description of changes:*
Fix TestDockerKubernetes122AWSIamAuth" and "TestVSphereKubernetes122AWSIamAuth

client.authentication v1alpha1 has been deprecated in K8s 1.22.  

*Testing (if applicable):*


pkg/awsiamauth/config/aws-iam-authenticator-kubeconfig.yaml
```diff
@@ -16,7 +16,7 @@ users:
 - name: {{.clusterName}}-aws
   user:
     exec:
-      apiVersion: client.authentication.k8s.io/v1alpha1
+      apiVersion: client.authentication.k8s.io/v1beta1
```
jaxesn, line 19: I'm not positive you can do this. The older versions still use the older iam auth which I believe expects the v1alpha1. Worth a sort to verify tho.

mdsgabriel: V1beta1 is [available since k8s 1.11](https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/0221afb8e2a5d14a20a93473edc4c2aa7676ce95)

This was tested by manually creating 1.21 and 1.22 clusters and using the generated kubeconfig (cluster name-aws.kubeconfig) to get the list of pods (similar to what's done by the tests that are failing).

What other scenarios should be tested?

abhay-krishna: `v1beta1` apiversion has been available since k8s `v1.11`, but only since aws-iam-auth `v0.5.4`. Since 1.21 and below use aws-iam-auth versions previous to `v0.5.4`, we want to confirm if this change will work on them in a backward compatible manner.

jaxesn: This is the iam auth pr where it was changed

https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/386/files

Like Abhay said, eksd kube versions less than 1.22 are still using the older iam auth which was parsing it as v1alpha. I wasn't sure if they would also work if the version was v1beta1 in the config file. If you tried a 1.21 cluster with this change and it worked, then it must be ok with it. Let's just make sure we still have the 121 test in addition to the new 122 one and then we can merge and make sure tests still pass.

jaxesn:
This is also interesting, looks Jyoti is trying to add a migration path to iam auth upstream for this situation

https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/439

scripts/e2e_test_docker.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 REPO_ROOT=$(git rev-parse --show-toplevel)
 BIN_FOLDER=$REPO_ROOT/bin
-TEST_REGEX="${1:-TestDockerKubernetes121SimpleFlow}"
+TEST_REGEX="${1:-TestDockerKubernetes122AWSIamAuth}"
```
vivek-koppuru, line 33: You didn't mean to check this in right?

pkg/awsiamauth/config/aws-iam-authenticator-kubeconfig.yaml
vivek-koppuru: I wonder if doing this migration will affect our upgrade use case as well.

jaxesn: We should make sure there is an upgrade test for this.

jaxesn (approved):
/lgtm

Issue #1208, opened 2022-02-14
Make ssh keys on MachineConfig mutable 


vignesh-goutham:
- [x] Enable updates to ssh keys using CLI
- [x] Enable updates to ssh keys using GitOps

We should make sure the ssh keys are mutable across all providers.

mitalipaygude:
#1461

PR #1427, opened 2022-03-10
Update ssh keys and user name using CLI

mitalipaygude:
*Issue #, if available:*
#1208 

*Description of changes:*
These code changes allows an user to update the ssh keys and the user name for vsphere machine configs.
If the user updates the `etcd` or the `control plane` machine configs, we update the `template-cp` and this change triggers CAPI to roll out new nodes.
In case of worker node machine config updates, we create a new `kubeadm config template` and update the worker nodes with the new user/ssh keys.

*Testing (if applicable):*
- Upgrade management cluster's `etcd`, `control plane` and `worker node` machine configs with new users and ssh keys using `upgrade cluster` (CLI only). Logged into the vms with the updated keys.
- Upgrade workload cluster's `etcd`, `control plane` and `worker node` machine configs with new users and ssh keys using `upgrade cluster` (CLI only). Logged into the vms with the updated keys

### DIFF
--- a/controllers/controllers/resource/reconciler_test.go
+++ b/controllers/controllers/resource/reconciler_test.go
@@ -152,13 +152,23 @@ func TestClusterReconcilerReconcile(t *testing.T) {
 				if err := yaml.Unmarshal([]byte(machineDeploymentFile), machineDeployment); err != nil {
 					t.Errorf("unmarshalling machinedeployment failed: %v", err)
 				}
+				workerNodeMachineConfig := &anywherev1.VSphereMachineConfig{
+					Spec: anywherev1.VSphereMachineConfigSpec{
+						Users: []anywherev1.UserConfiguration{
+							{
+								Name:              "capv",
+								SshAuthorizedKeys: []string{"ssh-rsa ssh_key_value"},
+							},
+						},
+					},
+				}
 				fetcher.EXPECT().MachineDeployment(ctx, gomock.Any(), gomock.Any()).Return(machineDeployment, nil)
 
 				fetcher.EXPECT().Etcd(ctx, gomock.Any()).Return(etcdadmCluster, nil)
 				fetcher.EXPECT().ExistingVSphereDatacenterConfig(ctx, gomock.Any(), gomock.Any()).Return(&anywherev1.VSphereDatacenterConfig{}, nil)
 				fetcher.EXPECT().ExistingVSphereControlPlaneMachineConfig(ctx, gomock.Any()).Return(&anywherev1.VSphereMachineConfig{}, nil)
 				fetcher.EXPECT().ExistingVSphereEtcdMachineConfig(ctx, gomock.Any()).Return(&anywherev1.VSphereMachineConfig{}, nil)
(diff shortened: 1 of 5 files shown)


controllers/controllers/resource/reconciler_test.go
```diff
@@ -152,13 +152,23 @@ func TestClusterReconcilerReconcile(t *testing.T) {
 				if err := yaml.Unmarshal([]byte(machineDeploymentFile), machineDeployment); err != nil {
 					t.Errorf("unmarshalling machinedeployment failed: %v", err)
 				}
+				workerNodeMachineConfig := &anywherev1.VSphereMachineConfig{
```
jiayiwang7, line 155: this pr also covers controller updates? if not fully, let's maybe move the controller changes into separate pr

mitalipaygude: There is common code that calls the adding new kubeadm config template, which is why these changes were necessary. However the immutable changes for ssh keys in the webhook haven't been updated, so the updates to the ssh keys wont go through.

pkg/providers/vsphere/vsphere.go
```diff
@@ -1182,23 +1182,23 @@ func (p *vsphereProvider) ValidateNewSpec(ctx context.Context, cluster *types.Cl
 	return nil
 }
 
-func (p *vsphereProvider) needsNewMachineTemplate(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newClusterSpec *cluster.Spec, workerNodeGroupConfiguration v1alpha1.WorkerNodeGroupConfiguration, vdc *v1alpha1.VSphereDatacenterConfig, prevWorkerNodeGroupConfigs map[string]v1alpha1.WorkerNodeGroupConfiguration) (bool, error) {
+func (p *vsphereProvider) needsNewMachineTemplate(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newClusterSpec *cluster.Spec, workerNodeGroupConfiguration v1alpha1.WorkerNodeGroupConfiguration, vdc *v1alpha1.VSphereDatacenterConfig, prevWorkerNodeGroupConfigs map[string]v1alpha1.WorkerNodeGroupConfiguration) (bool, error, *v1alpha1.VSphereMachineConfig, *v1alpha1.VSphereMachineConfig) {
```
jiayiwang7, line 1185: this method should not return objects

mitalipaygude: Added new method

pkg/api/v1alpha1/vspheremachineconfig_types.go
```diff
@@ -20,6 +20,25 @@ type VSphereMachineConfigSpec struct {
 	Users             []UserConfiguration `json:"users,omitempty"`
 }
 
+func UsersSliceEqual(a, b []UserConfiguration) bool {
```
jiayiwang7, line 23: can we add unit test?

jiayiwang7 (approved):

mitalipaygude:
#1440

pkg/api/v1alpha1/vspheremachineconfig_types.go
mitalipaygude: #1440

PR #1432, opened 2022-03-10
Dynamic config of api structs in cluster Config

g-gaston:
*Description of changes:*
This introduces the concept of Cluster Config Manager and Manager Entry.

* The manager holds the configuration to parse a Config from yaml, set its
defaults and validate it. It allows to dynamically register all of that
config, oferring a public api to Parse, SetDefaults and Validate.
* The entry is just a Manager api object to allow registering mappings,
processors, defaults and validations all at once. When adding new
content to the Config, we will probably need to register all of those
for the new type/s. This just makes it a bit easier.

The package public api to parse, set defaults and validate will continue
to be the same. The cluster package has a package level manager that is
used from ParseConfig, SetConfigDefaults and ValidateConfig.

Parsing has already been moved to the new manager framework. Defaults
and validations will follow in the next PRs.

Now when the parsing method finds an unknown Kind in the yaml content,
it ignores it instead of returning an error. This, aside for following a
bit better the idea of registering types, allows us to add new providers
as we go, without breaking the existing parsing until the new provider
is registered in the default manager.


pkg/cluster/config_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	return &ConfigManager{
+		entry: NewConfigManagerEntry(),
+	}
+}
+
+func (c *ConfigManager) Register(entries ...*ConfigManagerEntry) error {
+	return c.entry.Merge(entries...)
+}
+
+func (c *ConfigManager) RegisterMapping(kind string, generator APIObjectGenerator) error {
```
chrisdoherty4, line 38: I had a thought. If the intent is for consumers to model `ConfigManagerEntry`s as everything needed for mapping, defaulting and validating, why allow consumers to register things individually? 

It just feels odd to encourage that and then make the statement that they're individual concepts on the same construct that wants to model them as a single concept?

chrisdoherty4 (approved):

chrisdoherty4:
Need code comments.

pkg/cluster/config_manager.go
g-gaston: How I see it, they are just two different options to do the same thing. `ConfigManagerEntry` exists only for convenience, not to enforce a particular pattern.
If you are going to register validations and defaults at the same time, you can use `ConfigManagerEntry`. But if you prefer, you can register them in separate calls.

chrisdoherty4: Ok. I think we could add something to `ConfigManagerEntry` docs to explicitly state its a way to conveniently model everything associated with a particular type and is semantically equivalent to performing individual registrations to avoid any confusion.

chrisdoherty4 (approved):

pkg/cluster/config_manager.go
g-gaston: Done

PR #1434, opened 2022-03-10
Add CNIConfig type

mrajashree:
*Issue #, if available:*

*Description of changes:*
This PR adds the new field CNIConfig to specify cni plugin proposed in this design doc: https://github.com/aws/eks-anywhere/blob/main/designs/cilium-configuration-policy.md#api-changes

The exiting CNI field will be deprecated.
The reason behind adding this new type is to allow users to specify more configuration options for the different CNI plugins where possible. For instance, eks anywhere installs cilium through a helm chart, which can accept values from the user. So the new CNIConfig type has separate fields to configure each plugin, like CiliumConfig that can accept values from the user and configure the cilium helm chart with it.

*Testing (if applicable):*
1. Cluster creation & upgrade
2. Upgrade existing cluster from release-0.7
3. Upgrade existing mgmt & workload cluster with flux enabled from release-0.7

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -47,7 +47,7 @@ func NewClusterGenerate(clusterName string, opts ...ClusterGenerateOpt) *Cluster
 				Services: Services{
 					CidrBlocks: []string{"10.96.0.0/12"},
 				},
-				CNI: Cilium,
+				CNIConfig: &CNIConfig{Cilium: &CiliumConfig{}},
 			},
 		},
 	}
@@ -433,12 +433,44 @@ func validateNetworking(clusterConfig *Cluster) error {
 	if err != nil {
 		return fmt.Errorf("invalid CIDR block for Services: %s. Please specify a valid CIDR block for service subnet", clusterConfig.Spec.ClusterNetwork.Services)
 	}
-	if clusterConfig.Spec.ClusterNetwork.CNI == "" {
-		return errors.New("cni not specified or empty")
+
+	return validateCNIPlugin(clusterConfig.Spec.ClusterNetwork)
+}
+
+func validateCNIPlugin(network ClusterNetwork) error {
+	if network.CNI != "" {
+		if network.CNIConfig != nil {
+			return fmt.Errorf("invalid format for cni plugin: both old and new formats used, use only the CNIConfig field")
+		}
+		logger.Info("Warning: CNI field is deprecated. Provide CNI information through CNIConfig")
+		if _, ok := validCNIs[network.CNI]; !ok {
+			return fmt.Errorf("cni %s not supported", network.CNI)
+		}
+		return nil
 	}
-	if _, ok := validCNIs[clusterConfig.Spec.ClusterNetwork.CNI]; !ok {
-		return fmt.Errorf("cni %s not supported", clusterConfig.Spec.ClusterNetwork.CNI)
+	return validateCNIConfig(network.CNIConfig)
+}
+
+func validateCNIConfig(cniConfig *CNIConfig) error {
+	if cniConfig == nil {
+		return fmt.Errorf("cni not specified")
 	}
+	var cniPluginSpecified int
+
+	if cniConfig.Cilium != nil {
+		cniPluginSpecified++
+	}
+
+	if cniConfig.Kindnetd != nil {
+		cniPluginSpecified++
+	}
+
+	if cniPluginSpecified == 0 {
+		return fmt.Errorf("no cni plugin specified")
+	} else if cniPluginSpecified > 1 {
+		return fmt.Errorf("cannot specify more than one cni plugins")
+	}
+
 	return nil
 }
 
--- a/pkg/api/v1alpha1/cluster_defaults.go
+++ b/pkg/api/v1alpha1/cluster_defaults.go
@@ -12,6 +12,7 @@ import (
 var clusterDefaults = []func(*Cluster) error{
 	setRegistryMirrorConfigDefaults,
 	setWorkerNodeGroupDefaults,
+	setCNIConfigDefault,
 }
 
 func setClusterDefaults(cluster *Cluster) error {
@@ -49,5 +50,22 @@ func setWorkerNodeGroupDefaults(cluster *Cluster) error {
 		logger.V(1).Info("First worker node group name not specified. Defaulting name to md-0.")
 		cluster.Spec.WorkerNodeGroupConfigurations[0].Name = "md-0"
 	}
+	return nil
+}
+
+func setCNIConfigDefault(cluster *Cluster) error {
+	if cluster.Spec.ClusterNetwork.CNIConfig != nil {
+		return nil
+	}
+
+	cluster.Spec.ClusterNetwork.CNIConfig = &CNIConfig{}
+	switch cluster.Spec.ClusterNetwork.CNI {
+	case Cilium, CiliumEnterprise:
+		cluster.Spec.ClusterNetwork.CNIConfig.Cilium = &CiliumConfig{}
+	case Kindnetd:
+		cluster.Spec.ClusterNetwork.CNIConfig.Kindnetd = &KindnetdConfig{}
+	}
+
+	cluster.Spec.ClusterNetwork.CNI = ""
 	return nil
 }
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -311,9 +311,11 @@ type ClusterNetwork struct {
 	// Defaults to 192.168.0.0/16 for pod subnet.
 	Pods     Pods     `json:"pods,omitempty"`
 	Services Services `json:"services,omitempty"`
-	// CNI specifies the CNI plugin to be installed in the cluster
+	// Deprecated. Use CNIConfig
 	CNI CNI `json:"cni,omitempty"`
-	DNS DNS `json:"dns,omitempty"`
+	// CNIConfig specifies the CNI plugin to be installed in the cluster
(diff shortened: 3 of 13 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -47,7 +47,7 @@ func NewClusterGenerate(clusterName string, opts ...ClusterGenerateOpt) *Cluster
 				Services: Services{
 					CidrBlocks: []string{"10.96.0.0/12"},
 				},
-				CNI: Cilium,
+				CNIConfig: &CNIConfig{Cilium: &CiliumConfig{}},
```
mrajashree, line 50: This is going to make all e2e tests run with the new field

pkg/api/v1alpha1/cluster.go
```diff
@@ -433,12 +433,45 @@ func validateNetworking(clusterConfig *Cluster) error {
...
+		return fmt.Errorf("cni not specified")
+	}
+	var cniPluginSpecified bool
+
+	if cniConfig.Cilium != nil {
+		cniPluginSpecified = true
+	}
+
+	if cniConfig.Kindnetd != nil {
+		if cniPluginSpecified {
```
g-gaston, line 465: nit: you can make `cniPluginSpecified` a counter and just return error at the end if it's more than 1

That avoids nested `if` and allows to add more CNIs easily

pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -51,3 +52,20 @@ func setWorkerNodeGroupDefaults(cluster *Cluster) error {
 	}
 	return nil
 }
+
+func setCNIConfigDefault(cluster *Cluster) error {
```
g-gaston, line 56: test for this method?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -323,9 +325,48 @@ func (n *ClusterNetwork) Equal(o *ClusterNetwork) bool {
 	if n == nil || o == nil {
 		return false
 	}
+
+	if n.CNI != "" {
```
g-gaston, line 329: shouldn't be enough to only check `CNIConfig`? at this point the cluster should already have the defaults assigned

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -323,9 +325,48 @@ func (n *ClusterNetwork) Equal(o *ClusterNetwork) bool {
...
+	}
+
 	return SliceEqual(n.Pods.CidrBlocks, o.Pods.CidrBlocks) &&
 		SliceEqual(n.Services.CidrBlocks, o.Services.CidrBlocks) &&
-		n.CNI == o.CNI && n.DNS.ResolvConf.Equal(o.DNS.ResolvConf)
+		n.DNS.ResolvConf.Equal(o.DNS.ResolvConf)
+}
+
+func (n *CNIConfig) Equal(o *CNIConfig) bool {
+	if (n.Cilium != nil && o.Cilium == nil) || (n.Cilium == nil && o.Cilium != nil) {
```
g-gaston, line 363: This is not checking the value of `PolicyEnforcementMode`

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -410,6 +451,17 @@ const (
 
 type CNI string
 
+type CNIConfig struct {
+	Cilium   *CiliumConfig   `json:"cilium,omitempty"`
+	Kindnetd *KindnetdConfig `json:"kindnetd,omitempty"`
+}
+
+type CiliumConfig struct {
+	PolicyEnforcementMode string `json:"policyEnforcementMode,omitempty"`
```
g-gaston, line 460: doc comment for this field?

pkg/dependencies/factory.go
```diff
@@ -423,7 +423,7 @@ func (f *Factory) WithHelm() *Factory {
 
 func (f *Factory) WithNetworking(clusterConfig *v1alpha1.Cluster) *Factory {
 	var networkingBuilder func() clustermanager.Networking
-	if clusterConfig.Spec.ClusterNetwork.CNI == v1alpha1.Kindnetd {
+	if clusterConfig.Spec.ClusterNetwork.CNI == v1alpha1.Kindnetd || (clusterConfig.Spec.ClusterNetwork.CNIConfig != nil && clusterConfig.Spec.ClusterNetwork.CNIConfig.Kindnetd != nil) {
```
g-gaston, line 426: I think we can assume defaults have been set and just check `CNIConfig`
Imo that's the advantage of setting defaults, we can rely on that in the rest of the code, just ignoring the old field

pkg/api/v1alpha1/cluster_defaults.go
mrajashree: I thought [these changes](https://github.com/aws/eks-anywhere/pull/1434/files#diff-c05e092d1a8232f079da26d00f68592a14a768e67bde0423768134e57b3815e0) would cover it, but I just realized in that I need to update the `in` cluster to use the deprecated field, will do that

pkg/api/v1alpha1/cluster_types.go
mrajashree: yeah that's what I was thinking, but when I was testing upgrade from an existing cluster, the eks-a controller at that point didn't have cniConfig set by the cluster defaults so it was in an error loop. I get that it will eventually progress once cluster defaults are set, but I thought it'll be better to have these additional checks

pkg/api/v1alpha1/cluster_types.go
mrajashree: I left that on purpose, I think switching between different policy enforcement modes should be allowed since it's all going through the helm chart, what do you think? also @vignesh-goutham would like your opinion on this

pkg/api/v1alpha1/cluster_types.go
g-gaston: Oh, is this because we don't have a mutation webhook, isn't it?

pkg/api/v1alpha1/cluster_types.go
g-gaston: Oh yeah I agree with that
However, this is an `Equal` method, so it should test equality I think

Also, I think there is a nuance. We can allow to update this from the cli but not from the controller (since we don't update/upgrade the cni from the controller)

pkg/api/v1alpha1/cluster_types.go
mrajashree: yeah I think so

g-gaston: cool, makes sense
what do you think about adding a comment here explaining exactly that?

pkg/dependencies/factory.go
```diff
@@ -423,7 +423,7 @@ func (f *Factory) WithHelm() *Factory {
 
 func (f *Factory) WithNetworking(clusterConfig *v1alpha1.Cluster) *Factory {
 	var networkingBuilder func() clustermanager.Networking
-	if clusterConfig.Spec.ClusterNetwork.CNI == v1alpha1.Kindnetd {
+	if clusterConfig.Spec.ClusterNetwork.CNIConfig != nil && clusterConfig.Spec.ClusterNetwork.CNIConfig.Kindnetd != nil {
```
g-gaston, line 426: nit: if this has been validated, this shouldn't be necessary right? we can just check one

pkg/api/v1alpha1/cluster_defaults.go
g-gaston: I think we should have a test that asserts that the structs are being set (they are not nil)
So add two testcases to the existing test

mrajashree: the changes in the existing testcase do assert that they are not nil, since CNIConfig and CiliumConfig in the `wantCluster` are initialized. So I think that is being tested. But still I separated out the test cases, so the original test for worker node group name does just that, and takes in a non-nil CNIConfig. and the new test case takes in nil CNIConfig and the `wantCluster` contains non-nil CNIConfig & CiliumConfig

pkg/api/v1alpha1/cluster_types.go
mrajashree: changed this so that clusterNetwork.Equal checks the entire object including configuration options for the cni plugins like cilium.
added separate Equal checks for individual clusterNetwork fields so that CLI can call those, and allow changing policyEnforcementMode.
webhook will continue to call clusterNetwork.Equal so no changes will be allowed through the webhook

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -323,9 +325,102 @@ func (n *ClusterNetwork) Equal(o *ClusterNetwork) bool {
...
 		return false
 	}
-	return SliceEqual(n.Pods.CidrBlocks, o.Pods.CidrBlocks) &&
-		SliceEqual(n.Services.CidrBlocks, o.Services.CidrBlocks) &&
-		n.CNI == o.CNI && n.DNS.ResolvConf.Equal(o.DNS.ResolvConf)
+
+	if !CNIPluginSame(*n, *o) {
+		return false
+	}
+	if o.CNIConfig != nil {
```
g-gaston, line 332: nit: we can remove this, `CNIConfig.Equal` handles this case

g-gaston (approved):

pkg/api/v1alpha1/cluster_types.go
mrajashree: this is needed in case the older cluster spec does not have cniConfig set, and is using the deprecated field

g-gaston (commented):
/lgtm

PR #1439, opened 2022-03-10
Snow unstacked etcd support

jiayiwang7:
*Issue #, if available:*

#1105 

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/clusterapi/apibuilder.go
+++ b/pkg/clusterapi/apibuilder.go
@@ -3,6 +3,8 @@ package clusterapi
 import (
 	"fmt"
 
+	etcdbootstrapv1 "github.com/mrajashree/etcdadm-bootstrap-provider/api/v1beta1"
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -27,6 +29,7 @@ var (
 	clusterAPIVersion             = clusterv1.GroupVersion.String()
 	kubeadmControlPlaneAPIVersion = controlplanev1.GroupVersion.String()
 	bootstrapAPIVersion           = bootstrapv1.GroupVersion.String()
+	etcdAPIVersion                = etcdv1.GroupVersion.String()
 	etcdClusterAPIVersion         = fmt.Sprintf("etcdcluster.%s/%s", clusterv1.GroupVersion.Group, clusterv1.GroupVersion.Version)
 )
 
@@ -95,6 +98,9 @@ func KubeadmControlPlane(clusterSpec *cluster.Spec, infrastructureObject APIObje
 	if clusterSpec.Spec.ExternalEtcdConfiguration != nil {
 		etcd.External = &bootstrapv1.ExternalEtcd{
 			Endpoints: []string{},
+			CAFile:    "/etc/kubernetes/pki/etcd/ca.crt",
+			CertFile:  "/etc/kubernetes/pki/apiserver-etcd-client.crt",
+			KeyFile:   "/etc/kubernetes/pki/apiserver-etcd-client.key",
 		}
 	} else {
 		etcd.Local = &bootstrapv1.LocalEtcd{
@@ -239,6 +245,43 @@ func MachineDeployment(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1
 				},
 			},
 			Replicas: &replicas,
+		},
+	}
+}
+
+func EtcdadmCluster(clusterSpec *cluster.Spec, infrastructureObject APIObject) *etcdv1.EtcdadmCluster {
+	replicas := int32(clusterSpec.Spec.ExternalEtcdConfiguration.Count)
+	return &etcdv1.EtcdadmCluster{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: etcdAPIVersion,
+			Kind:       etcdadmClusterKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterSpec.GetName(),
+			Namespace: constants.EksaSystemNamespace,
+		},
+		Spec: etcdv1.EtcdadmClusterSpec{
+			Replicas: &replicas,
+			EtcdadmConfigSpec: etcdbootstrapv1.EtcdadmConfigSpec{
+				EtcdadmBuiltin: true,
+				Format:         etcdbootstrapv1.Format("cloud-config"),
+				CloudInitConfig: &etcdbootstrapv1.CloudInitConfig{
+					Version:    clusterSpec.VersionsBundle.KubeDistro.EtcdVersion,
+					InstallDir: "/usr/bin",
+				},
+				PreEtcdadmCommands: []string{
+					"hostname \"{{`{{ ds.meta_data.hostname }}`}}",
+					"echo \"::1         ipv6-localhost ipv6-loopback\" >/etc/hosts",
+					"echo \"127.0.0.1   localhost\" >>/etc/hosts",
+					"echo \"127.0.0.1   {{`{{ ds.meta_data.hostname }}`}}\" >>/etc/hosts",
+					"echo \"{{`{{ ds.meta_data.hostname }}`}}\" >/etc/hostname",
+				},
+			},
+			InfrastructureTemplate: v1.ObjectReference{
+				APIVersion: infrastructureObject.GetObjectKind().GroupVersionKind().GroupVersion().String(),
+				Kind:       infrastructureObject.GetObjectKind().GroupVersionKind().Kind,
+				Name:       infrastructureObject.GetName(),
+			},
 		},
 	}
 }
--- a/pkg/clusterapi/apibuilder_test.go
+++ b/pkg/clusterapi/apibuilder_test.go
@@ -3,6 +3,8 @@ package clusterapi_test
 import (
 	"testing"
 
+	etcdbootstrapv1 "github.com/mrajashree/etcdadm-bootstrap-provider/api/v1beta1"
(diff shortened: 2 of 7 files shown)


mitalipaygude (approved):

pkg/clusterapi/apibuilder_test.go
```diff
@@ -48,6 +51,7 @@ func (c *providerMachineTemplate) DeepCopyObject() runtime.Object {
 }
 
 func newApiBuilerTest(t *testing.T) apiBuilerTest {
+	format.MaxLength = 0
```
taneyland, line 54: What does this do?

jiayiwang7: it prevents truncation of the objects being compared in unit tests. It helps debug the tests

cxbrowne1207: Can we remove it if it's not needed anymore then?

jiayiwang7: i kinda want to keep it for all the apibuilder tests - it helps w troubleshooting in general

g-gaston: Instead of `MaxLength`, you can swap `Equal` by `BeComparableTo`
That will print only the diff between the two objects as opposed to the whole structs. It makes it way easier to debug tests when comparing two big structs

pkg/clusterapi/apibuilder.go
```diff
@@ -298,3 +280,36 @@ func MachineDeployment(clusterSpec *cluster.Spec, workerNodeGroupConfig anywhere
 
 	return *md
 }
+
+// EtcdadmCluster builds a etcdadmCluster based on an eks-a cluster spec and infrastructureObject.
+func EtcdadmCluster(clusterSpec *cluster.Spec, infrastructureObject APIObject) *etcdv1.EtcdadmCluster {
```
g-gaston, line 285: nit
```suggestion
func EtcdadmCluster(clusterSpec *cluster.Spec, machineTemplate APIObject) *etcdv1.EtcdadmCluster {
```

pkg/clusterapi/bottlerocket.go
```diff
@@ -85,3 +87,12 @@ func SetBottlerocketAdminContainerImageInKubeadmConfigTemplate(kct *bootstrapv1.
 func SetBottlerocketControlContainerImageInKubeadmConfigTemplate(kct *bootstrapv1.KubeadmConfigTemplate, versionsBundle *cluster.VersionsBundle) {
 	kct.Spec.Template.Spec.JoinConfiguration.BottlerocketControl = bottlerocketControl(versionsBundle.BottleRocketHostContainers.Control)
 }
+
+// SetBottlerocketInEtcdCluster adds bottlerocket config in etcdadmCluster.
+func SetBottlerocketInEtcdCluster(etcd *etcdv1.EtcdadmCluster, versionsBundle *cluster.VersionsBundle) {
```
g-gaston, line 92: does this method need to be public or can we just test it through `EtcdadmCluster`?

pkg/clusterapi/etcd.go
```diff
@@ -0,0 +1,86 @@
...
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+// SetEtcdConfigInEtcdCluster sets up the etcd config in EtcdadmCluster.
+func SetEtcdConfigInEtcdCluster(etcd *etcdv1.EtcdadmCluster, version string) {
```
g-gaston, line 17: Same comment about all these "Set" methods, I would vote to make them unexported and test them through the main method that uses them

g-gaston, line 17:
```suggestion
func SetUbuntuConfigInEtcdCluster(etcd *etcdv1.EtcdadmCluster, version string) {
```

pkg/clusterapi/etcd.go
```diff
@@ -0,0 +1,86 @@
...
+		"hostname \"{{`{{ ds.meta_data.hostname }}`}}",
+		"echo \"::1         ipv6-localhost ipv6-loopback\" >/etc/hosts",
+		"echo \"127.0.0.1   localhost\" >>/etc/hosts",
+		"echo \"127.0.0.1   {{`{{ ds.meta_data.hostname }}`}}\" >>/etc/hosts",
+		"echo \"{{`{{ ds.meta_data.hostname }}`}}\" >/etc/hostname",
+	}
+}
+
+// SetEtcdConfigInCluster sets up the etcd config in CAPI Cluster.
+func SetEtcdConfigInCluster(cluster *clusterv1.Cluster, clusterSpec *cluster.Spec) {
```
g-gaston, line 32: why not pass the `EtcdadmCluster` here to set the ref? I believe we do that with other objects that are referenced

pkg/clusterapi/etcd.go
```diff
@@ -0,0 +1,86 @@
...
+		APIVersion: etcdClusterAPIVersion,
+		Kind:       etcdadmClusterKind,
+		Name:       EtcdClusterName(clusterSpec),
+		Namespace:  constants.EksaSystemNamespace,
+	}
+}
+
+// SetUnstackedEtcdConfigInKubeadmControlPlaneForBottlerocket sets up unstacked etcd configuration in kubeadmControlPlane for bottlerocket.
+func SetUnstackedEtcdConfigInKubeadmControlPlaneForBottlerocket(kcp *controlplanev1.KubeadmControlPlane, externalEtcdConfig *v1alpha1.ExternalEtcdConfiguration) {
+	if externalEtcdConfig == nil {
```
g-gaston, line 47: why not perform this check in the caller and that way you don't need to pass `externalEtcdConfig`?

pkg/clusterapi/etcd.go
```diff
@@ -0,0 +1,86 @@
...
+		return
+	}
+
+	kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.External = &bootstrapv1.ExternalEtcd{
+		Endpoints: []string{},
```
g-gaston, line 52: do we need to set this? I think it should be empty right? The capi controllers take care of populating this

pkg/clusterapi/etcd.go
```diff
@@ -0,0 +1,86 @@
...
+		CAFile:    "/etc/kubernetes/pki/etcd/ca.crt",
+		CertFile:  "/etc/kubernetes/pki/apiserver-etcd-client.crt",
+		KeyFile:   "/etc/kubernetes/pki/apiserver-etcd-client.key",
+	}
+}
+
+// SetStackedEtcdConfigInKubeadmControlPlane sets up stacked etcd configuration in kubeadmControlPlane.
+func SetStackedEtcdConfigInKubeadmControlPlane(kcp *controlplanev1.KubeadmControlPlane, clusterSpec *cluster.Spec) {
+	if clusterSpec.Cluster.Spec.ExternalEtcdConfiguration != nil {
```
g-gaston, line 75: yeah, I feel like all these checks should probably be in the caller

pkg/clusterapi/name.go
```diff
@@ -76,6 +76,11 @@ func KubeadmControlPlaneName(clusterSpec *cluster.Spec) string {
 	return clusterSpec.Cluster.GetName()
 }
 
+// EtcdClusterName sets the default EtcdCluster object name.
+func EtcdClusterName(clusterSpec *cluster.Spec) string {
```
g-gaston, line 80: can we pass only the eks-a cluster? it makes this easier to reuse

pkg/clusterapi/registry_mirror.go
```diff
@@ -53,6 +55,18 @@ func SetRegistryMirrorInKubeadmConfigTemplateForUbuntu(kct *bootstrapv1.KubeadmC
 	return addRegistryMirrorInKubeadmConfigSpecFiles(&kct.Spec.Template.Spec, mirrorConfig)
 }
 
+// SetRegistryMirrorInEtcdCluster sets up registry mirror configuration in etcdadmCluster.
+func SetRegistryMirrorInEtcdCluster(etcd *etcdv1.EtcdadmCluster, mirrorConfig *v1alpha1.RegistryMirrorConfiguration) {
```
g-gaston, line 59: isn't this only for BR?

pkg/clusterapi/apibuilder.go
jiayiwang7: why? it is referencing to a `InfrastructureTemplate` and i dont see why not naming it `infrastructureObject`.

pkg/clusterapi/etcd.go
jiayiwang7: same as above

pkg/clusterapi/etcd.go
jiayiwang7: why not passing this in the method so the caller does not need to have the if condition? we use same pattern for all methods in apibuilder and imo it is clear: updating etcd config in kcp based on externalEtcdConfig arg.

pkg/clusterapi/etcd.go
jiayiwang7: why? does it make big difference?

pkg/clusterapi/registry_mirror.go
jiayiwang7: no for both BR and Ubuntu

pkg/clusterapi/apibuilder.go
g-gaston: IMO `infrastructureObject` is less clear. What does object mean in this context, a kubernetes object? But then what is an infrastructure kubernetes object?

If you prefer to use `infrastructure` since that's how etcdadmcluster names the field, then I think `infrastructureTemplate` is better than `object`.

PR #1445, opened 2022-03-11
Minor change to docs contrib guide

chrisnegus:
This was one comment that didn't get it when the new "Contributing to documentation" guide before it was published, so I'm making the fix here.


TerryHowe (approved):
/lgtm

PR #1446, opened 2022-03-11
Ensure Tinkerbell Hardware objects have unique IDs

chrisdoherty4:
When submitting `Hardware` manifests during `create cluster` for the Tinkerbell provider, we need to ensure the hardware ID they reference is unique.

### DIFF
--- a/pkg/providers/tinkerbell/hardware/config.go
+++ b/pkg/providers/tinkerbell/hardware/config.go
@@ -77,6 +77,9 @@ func (hc *HardwareConfig) ValidateHardware(skipPowerActions bool, tinkHardwareMa
 		bmcRefMap = hc.initBmcRefMap()
 	}
 
+	// A database of observed hardware IDs so we can check for uniqueness.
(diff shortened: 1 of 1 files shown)


pkg/providers/tinkerbell/hardware/config.go
```diff
@@ -86,6 +89,11 @@ func (hc *HardwareConfig) ValidateHardware(skipPowerActions bool, tinkHardwareMa
 			return fmt.Errorf("hardware: %s ID is required", hw.Name)
 		}
 
+		if _, found := hardwareIdsDb[hw.Spec.ID]; found {
+			return fmt.Errorf("duplicate hardware id: %v", hw.Spec.ID)
+		}
+		hardwareIdsDb[hw.Spec.ID] = struct{}{}
```
ptrivedi, line 95: Are you planning to do anything with this hardwareIdsDb later? I won't have thought of sticking an empty struct, possibly would have gone for bool, more intuitive IMO. I miss having c++'s unordered_set here, where I could just store the keys and not have to worry about values.

PR #1448, opened 2022-03-11
Pull CAPAS image from release bundle

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

CAPAS image was hardcoded in the snow provider. Now that the image is built in the release bundle, we can fetch it from there instead.

Notice this is still a tmp solution before we automate the build for snow `infrastructure-components.yaml` with proper image  version.

*Testing (if applicable):*


jaxesn (approved):
/lgtm

PR #1452, opened 2022-03-14
Create Hardware CSV per cluster 

gwesterfieldjr:
Create a hardware CSV per test cluster so we can easily reference/identify what hardware was associated with the test.


internal/pkg/api/hardware.go
```diff
@@ -3,32 +3,26 @@ package api
...
 )
 
 const (
-	Dell       = "dell"
-	HP         = "hp"
-	SuperMicro = "supermicro"
+	HardwareVendorDell       = "dell"
+	HardwareVendorHP         = "hp"
+	HardwareVendorSuperMicro = "supermicro"
+	HardwareVendorAgnostic   = "agnostic"
```
chrisdoherty4, line 16: You might want to use 'unspecified'. 'Agnostic' suggests it should 'just work' and that the underlying technology is irrelevant. I'm not sure we  know that hence it might be safer to use 'unspecified' which doesn't make any propositions.

internal/pkg/api/hardware.go
```diff
@@ -3,32 +3,26 @@ package api
...
-	Gateway      string `csv:"gateway"`
-	Nameservers  string `csv:"nameservers"`
-	Netmask      string `csv:"netmask"`
-	MacAddress   string `csv:"mac"`
-	Hostname     string `csv:"hostname"`
-	Vendor       string `csv:"vendor"`
-	BmcIpAddress string `csv:"bmc_ip"`
-	BmcUsername  string `csv:"bmc_username"`
-	BmcPassword  string `csv:"bmc_password"`
+	Id string `csv:"guid"`
```
chrisdoherty4, line 20: This is awesome.

internal/pkg/api/hardware.go
```diff
@@ -3,32 +3,26 @@ package api
...
+	hardware.Machine
 }
 
 func NewHardwareSlice(csvFile string) ([]*Hardware, error) {
```
chrisdoherty4, line 24: I tend to make it explicit that we're reading from a file when performing initialization from a file (e.g. `NewHardwareSliceFromFile()`). The standard `New` func shouldn't really be doing anything fancy like opening files. It _could_ be given an `io.Reader` though allowing you to deal with opening files as a separate concern from initialization. Then you can do stuff like:

```go
func NewHardwareSlice(r io.Reader) ... {...}

func NewHardwareSliceFromFile(file string) ... {
  fh, _ := os.Open(file, ...)
  return NewHardwareSlice(fh)
}
```

internal/pkg/api/hardware.go
```diff
@@ -43,3 +37,57 @@ func NewHardwareSlice(csvFile string) ([]*Hardware, error) {
...
+
+func NewHardwareMap(csvFile string) (map[string]*Hardware, error) {
+	slice, err := NewHardwareSlice(csvFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create hardware map from hardware csv file: %v", err)
+	}
+	return HardwareSliceToMap(slice), nil
+}
+
+func HardwareSliceToMap(slice []*Hardware) map[string]*Hardware {
```
chrisdoherty4, line 49: The function could do with a code comment indicating the behavior when 2 hardware with the same ID but different data are supplied in the slice. The behavior you've got is to use the first instance. 

In general, comment exported functions. Even if its intended for use in test because its still just code.

pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -110,13 +110,22 @@ const NameserversSeparator = "|"
...
+}
+
 // UnmarshalCSV unmarshalls s where is is a list of nameservers separated by NameserversSeparator.
 func (n *Nameservers) UnmarshalCSV(s string) error {
 	servers := strings.Split(s, NameserversSeparator)
 	*n = append(*n, servers...)
 	return nil
 }
 
+// MarshalCSV marshalls Nameservers into a string list of nameservers separated by NameserversSeparator.
```
chrisdoherty4, line 124: I've just added this on a different PR for the CSV parsing as I also needed it in test FYI in-case mine goes in before.

test/framework/cluster.go
```diff
@@ -90,6 +94,38 @@ func NewClusterE2ETest(t *testing.T, provider Provider, opts ...ClusterE2ETestOp
 	return e
 }
 
+func WithHardware(vendor string, requiredCount int) ClusterE2ETestOpt {
+	return func(e *ClusterE2ETest) {
+		csvFilePath := os.Getenv(tinkerbellInventoryCsvFilePathEnvVar)
```
chrisdoherty4, line 99: Can we deal with reading the environment separately?

chrisdoherty4 (commented):
See comments.

test/framework/cluster.go
gwesterfieldjr: We can but then every test would need to read the envvar and pass the value to the WithHardware filler when creating the ClusterE2ETest object. In principle I agree it should be an input, but in practice all the env vars required by ClusterE2ETest just read them within the implementation when required. Maybe this is a good option for a larger clean up task?

PR #1454, opened 2022-03-14
clean up ec2 instances after each build job

danbudris:
*Issue #, if available:*

*Description of changes:*
Clean up the e2e test instances after each test run. Follow up to this is going to be tagging instances based on the build ID and using that tag for the cleanup.

*Testing (if applicable):*

### DIFF
--- a/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
+++ b/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
@@ -91,6 +91,11 @@ phases:
         ./bin/test e2e cleanup vsphere
         -n i-
         -v 4
+      - >
+        ./cmd/test e2e cleanup aws
+        -s ${INTEGRATION_TEST_STORAGE_BUCKET}
+        -t EKSA-E2E
+        -v 4
 reports:
   e2e-reports:
     files:


cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
```diff
@@ -91,6 +91,11 @@ phases:
         ./bin/test e2e cleanup vsphere
         -n i-
         -v 4
+      - >
+        ./cmd/test e2e cleanup aws
```
jaxesn, line 95: Actually, this may affect running presubmits...

jaxesn: Looks like we tag the instances [here](https://github.com/aws/eks-anywhere/blob/c839dee8be3448b98ffd2bb00aa55faa0a36eed4/internal/test/e2e/setup.go#L74), but both e2e and docker e2e are going thru here.  If we could pass this tag through, we could override it for the docker/presubmit and avoid cleaning those up too soon?

As a side note, the prow job has the ami-id in it, so if/when we update the ami well probably have to update them as well?
https://github.com/aws/eks-anywhere-prow-jobs/blob/main/jobs/aws/eks-anywhere/eks-anywhere-e2e-presubmits.yaml#L40

danbudris: I'm going to just adjust this to use a tag right off the bat, so we don't end up with the concurrent run issues. Let's hold this for now.

danbudris:
this is stale, closing

PR #1455, opened 2022-03-14
Added Kubernetes validation for Tinkerbell hardware name field

panktishah26:
*Description of changes:*
Added Kubernetes validation for Tinkerbell hardware name field for the generate hardware and create cluster command.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/hardware/validator.go
+++ b/pkg/providers/tinkerbell/hardware/validator.go
@@ -2,8 +2,6 @@ package hardware
 
 import (
 	"fmt"
-
-	apimachineryvalidation "k8s.io/apimachinery/pkg/util/validation"
 )
 
 // MachineAssertion defines a condition that Machine must meet.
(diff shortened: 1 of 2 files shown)


cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -132,6 +132,13 @@ func (hOpts *hardwareOptions) generateHardware(ctx context.Context) error {
 			return fmt.Errorf("failed reading csv: %v", err)
 		}
 
+		if !hOpts.dryRun {
+			err := validation.IsDNS1123Subdomain(items[csv.HostnameIndex])
+			if err != nil {
+				return fmt.Errorf("error validating hardware hostname, hostname %s is invalid %v", items[csv.HostnameIndex], err)
```
mitalipaygude, line 138: We don't really need the keyword `error` while already returning an `Errorf`

pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -93,6 +97,11 @@ func UniqueHostnames() MachineAssertion {
 			return fmt.Errorf("duplicate Hostname: %v", m.Hostname)
 		}
 
+		err := apimachineryvalidation.IsDNS1123Subdomain(m.Hostname)
```
chrisdoherty4, line 100: This check is already being done on the `Machine`s `Validate` method. I think the additional check you added was when we read in the yamls again - I've not touched that yet. 

Also, side note, this function is self contained and single responsibility. Ideally we shouldn't add to these instead adding new validation functions.

panktishah26: I used to think that you have added validation for hardware yaml file and not the generate hardware file. I have corrected the code.

chrisdoherty4 (approved):

PR #1459, opened 2022-03-15
clean up test ec2 instances on post build

danbudris:
*Issue #, if available:*

*Description of changes:*
Clean up test instances associated with a given build ID during the post-build stage. This change will tag instances with the parent job ID (the codebuild ID), and then in the post-build phase clean up all the e2e instances tagged with the build ID of the executing codebuild job. This will help us to keep from accumulating stranded instances. 

We are cleaning up the instances before cleaning up the VMs so that any orphaned instances with management clusters left behind are not reconciling and spawning new VMs.

- tag e2e test instances with the codebuild Job ID which spawned the test run
- add a 'buildId' parameter to the AWS cleanup command which takes the build id and cleans up all e2e instances that have that build ID.

*Testing (if applicable):*

### DIFF
--- a/cmd/integration_test/cmd/cleanupaws.go
+++ b/cmd/integration_test/cmd/cleanupaws.go
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	maxAgeFlagName = "max-age"
-	tagFlagName    = "tag"
+	maxAgeFlagName  = "max-age"
+	tagFlagName     = "tag"
+	buildIdFlagName = "buildId"
 )
 
 var cleanUpAwsCmd = &cobra.Command{
@@ -49,6 +50,7 @@ func init() {
 	cleanUpAwsCmd.Flags().StringP(storageBucketFlagName, "s", "", "Name of s3 bucket used for e2e testing")
 	cleanUpAwsCmd.Flags().StringP(maxAgeFlagName, "a", "0", "Instance age in seconds after which it should be deleted")
 	cleanUpAwsCmd.Flags().StringP(tagFlagName, "t", "", "EC2 instance tag")
+	cleanUpAwsCmd.Flags().StringP(buildIdFlagName, "b", "", "Build ID of CodeBuild build; will clean up instances associated with the build")
 
 	for _, flag := range requiredAwsCleanUpFlags {
 		if err := cleanUpAwsCmd.MarkFlagRequired(flag); err != nil {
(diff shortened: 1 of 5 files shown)


internal/test/e2e/setup.go
```diff
@@ -81,6 +84,11 @@ func (e *E2ESession) setup(regex string) error {
 	logger.V(1).Info("Instance created", "instance-id", instanceId)
 	e.instanceId = instanceId
 
+	err = ec2.TagInstance(e.session, instanceId, jobIdTagKey, e.parentJobId)
```
g-gaston, line 87: I would vote to accept a `map[string]string` in `CreateInstance` so the instance is created with the tag from the beginning.

internal/test/e2e/cleanup.go
```diff
@@ -13,19 +13,29 @@ import (
 	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
-func CleanUpAwsTestResources(storageBucket string, maxAge string, tag string) error {
+const (
+	integrationTestTagKey = "Integration-Test"
+	codebuildIdTagKey     = "CodeBuildId"
+)
+
+func CleanUpAwsTestResources(storageBucket string, maxAge string, tag string, codebuildId string) error {
```
g-gaston, line 21: What do you think about taking `map[string]string` in the same way you do for `ListInstances`?
I would suggest passing that from the command flags. That way we can clean by any combination of tags and we don't have to keep adding arguments here when we add more tags.

cmd/integration_test/cmd/cleanupaws.go
```diff
@@ -49,6 +50,7 @@ func init() {
 	cleanUpAwsCmd.Flags().StringP(storageBucketFlagName, "s", "", "Name of s3 bucket used for e2e testing")
 	cleanUpAwsCmd.Flags().StringP(maxAgeFlagName, "a", "0", "Instance age in seconds after which it should be deleted")
 	cleanUpAwsCmd.Flags().StringP(tagFlagName, "t", "", "EC2 instance tag")
+	cleanUpAwsCmd.Flags().StringP(buildIdFlagName, "b", "", "Build ID of CodeBuild build; will clean up instances associated with the build")
```
g-gaston, line 53: Following the other comment, I would remove these two tag flags and just take `--tags`

danbudris:
this is stale; closing

PR #1464, opened 2022-03-15
Remove expired 1.20 conformance

TerryHowe:
This badge was already removed from the README. It is no longer supported because 1.20 is too old.


chrisnegus (commented):
/lgtm

PR #1476, opened 2022-03-15
Update eksdReleaseFetcher to support previous versions

taneyland:
*Issue #, if available:*

*Description of changes:*
E2E tests for prev release versions failed during upgrade because we now embed the eksd release objects to the cluster, so they were failing with this error: 
`Error: failed to upgrade cluster: failed fetching EKS-D release for cluster: error getting releases.distro.eks.amazonaws.com with kubectl: error: the server doesn't have a resource type "releases"`

We already have the logic in place to get the eksd release object if it isn't on the cluster, so this PR updates the eksdReleaseFetcher method to only return the `nil` eksd release object if it isn't on the cluster (instead of returning an error). From here, if it sees the eksd object as nil, it will install it & continue with the upgrade.

*Testing (if applicable):*
Created cluster with 7.0, upgraded with main

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -65,10 +65,7 @@ func GetEksdReleaseForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bu
 	if err != nil {
 		return nil, fmt.Errorf("failed fetching versions bundle: %v", err)
 	}
-	eksd, err := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
-	if err != nil {
-		return nil, fmt.Errorf("failed fetching EKS-D release for cluster: %v", err)
-	}
+	eksd, _ := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
 
 	return eksd, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/cluster/fetch.go
```diff
@@ -65,10 +65,7 @@ func GetEksdReleaseForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bu
 	if err != nil {
 		return nil, fmt.Errorf("failed fetching versions bundle: %v", err)
 	}
-	eksd, err := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
-	if err != nil {
-		return nil, fmt.Errorf("failed fetching EKS-D release for cluster: %v", err)
-	}
+	eksd, _ := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
```
danbudris, line 68: two thoughts:
- we should validate that the error is the expected error before passing on it; if it's not the expected does not exist error, we shoudl make sure to throw it
- add a debug log statement when it does pass on the error so we know what's happening and why in the future

vivek-koppuru: Agreed

danbudris (approved):

Issue #1166, opened 2022-02-07
Build eks-a admin AMI for Snow


g-gaston:
This is the image that Snow users will use to run the eks-a CLI to manage eks-a clusters

- [x] Allow to pull all images needed for eks-a and store them locally. Ideally without cluster spec, so the downloaded images can cover a `Bundle` completely, including all supported kubernetes versions. This can reuse the `import-images` commands with new flags or add a new command. The images can be stored either in a local tar ball or in the local "docker registry" (by just pulling them). Whatever is the "storage" mechanism should be accepted as input by some command to be able to populate a local registry later on.
- [x] Allow to download all artifacts locally, including manifests. Make sure that manifests referencing other manifests always point to local disk. This is to ensure the CLI won't fail in disconnected environments.
- [x] #3269

PR #1478, opened 2022-03-15
Import images command

g-gaston:
Part of #1166 

## What are we doing
In order to facilitate disconnected environments for Snow devices, we need to be able to store all container images needed for EKS-A in an AMI (Snow admin AMI), which will be deployed to a device and later used to populate a private registry.

The idea is that we will we use the CLI to both download and store the images and later populate the registry with them.

## Current state
The CLI already provides an `import-images` command. However, this one performs the pulling (downloading images from their origin) and pushing (populating the private registry) in one step. This means that internet connection is required to populate the registry. As explained previously, this won't work for Snow.

### Summary
* Add a new command `import`
* Add a sub-command `import images`
* Add a sub-command `download images`
* Deprecate the existing command `import-images`

### How do these command work
A couple of examples that are the most relevant to the Snow usecase:

* This will download the images from the original source (mostly ECR) and store them in a tarball. This command will require internet connection.
```console
eksctl anywhere download images --output path/file_images.tar
```

* And then to populate the registry:
```console
eksctl anywhere import images --input path/file_images.tar --registry https://djfkdks.xy/whatever
```

All the necessary input will be provided directly through flags and env vars and not through the cluster config file (like the current `import-image` and other commands do). This is more flexible, more straight forward, easier to script and it doesn't require users to create and understand the syntax of a `Cluster` object, which in my opinion belongs to a different subdomain.

### Why storing images on disk
An alternative could have been to just rely on docker to store the container images. Since the "pull" command will be run before "committing" the AMI, once this one is deployed to a device and the VM is started, we should be able to just do `docker push` against the private registry since the images should still be there.

While this works for the Snow design, it is a bit inflexible and it doesn't allow for other disconnected environments where the push and the pull are not run in the same machine. Using a tarball allows for easier portability.

### Why a new command
The current behavior of `import-images` is not flexible enough. Since the use of this command is not very extended (I don't have data to support this though), it is easier to deprecate it and create a new one, incorporating what have learned since we first introduce it. As opposed to trying to add more flags to alter how it works while keeping backwards compatibility.

Moreover, the CLI command tries to follow a `verb + resource` pattern, so `import images` follows this better than `import-images`. [Previous literature on this topic](https://github.com/aws/eks-anywhere/blob/main/designs/vcenter-template-import.md#side-effects).

### Implementation notes
* The eks-a tools image needs to be imported separetly since it's needed
in order to initialize the helm executable, which is needed to push the
helm charts to the registry.


TerryHowe:
Maybe we could add a `packages` subcommand to this at some point @a-cool-train @vincentni

jiayiwang7:
> This will download the images from the original source (mostly ECR) and store them in a tarball. This command will require internet connection. (Note: I still haven't decided how to pass the location of the tarball, maybe something like --to disk=mypath/to/image.tar)
eksctl anywhere import images --from origin --to disk

> And then to populate the registry (again, here the command is missing the registry endpoint info, so TBD if it's a new flag or part of the to value):
eksctl anywhere import images --from disk --to registry

In this proposal, the type of data passed into `--from` and `--to` can be either a container registry or a tarball (or more formats in the future). Then it relies on the CLI code determine the format and parse out the content. This may add ambiguity from both UX and code. But I understand it's more flexible and we control what to support from the cli code.

An alternative is to have args like `--from-registry`, `--to-registry`, `--from-disk`, `--to-disk` to distinguish the type of arg passed in. it's more verbose but also introduces more arguments.

What do you think about this alternative?

g-gaston:
> > This will download the images from the original source (mostly ECR) and store them in a tarball. This command will require internet connection. (Note: I still haven't decided how to pass the location of the tarball, maybe something like --to disk=mypath/to/image.tar)
> > eksctl anywhere import images --from origin --to disk
> 
> > And then to populate the registry (again, here the command is missing the registry endpoint info, so TBD if it's a new flag or part of the to value):
> > eksctl anywhere import images --from disk --to registry
> 
> In this proposal, the type of data passed into `--from` and `--to` can be either a container registry or a tarball (or more formats in the future). Then it relies on the CLI code to determine the format and parse out the content. This may add ambiguity from both UX and code. But I understand it's more flexible and we control what to support from the cli code.
> 
> An alternative is to have args like `--from-registry`, `--to-registry`, `--from-disk`, `--to-disk` to distinguish the type of arg passed in. it's more verbose but also introduces more arguments.
> 
> What do you think about this alternative?

That's a fair point.

However, my concern with that is that we now need to figure out what happens if multiple `from`s and multiple `to`s are provided. Do we support all of them? if so, what combinations? And whatever is the answer, it won't be clear by just looking at the command interface, you would need to read the documentation. Which is not a huge problem, but I think the ambiguity is higher.

Maybe there is a third option that we can iterate towards. When you say `This may add ambiguity from both UX and code`, what kind of ambiguity are you thinking about?

jiayiwang7:
we can validate the args and fail the cmd if `from-registry` and `from-disk` are passed in at the same time (same thing for `to`). Ambiguity is mainly that the args are less verbose since the contents can be different types (regi vs tarball). I understand we find similar use case when specifying URI vs local file in bundle file and other configuration files. But in this case, the logic to handle ECR regi vs local file import/export is quite different if I understand it correctly. Thus I wanna bring up this alternative to see if it worths us to have different arg names for different behaviors.

g-gaston:
> > we can validate the args and fail the cmd if `from-registry` and `from-disk` are passed in at the same time (same thing for `to`). Ambiguity is mainly that the args are less verbose since the contents can be different types (regi vs tarball). I understand we find similar use case when specifying URI vs local file in bundle file and other configuration files. But in this case, the logic to handle ECR regi vs local file import/export is quite different if I understand it correctly. Thus I wanna bring up this alternative to see if it worths us to have different arg names for different behaviors.
> 
> we can validate the args and fail the cmd if `from-registry` and `from-disk` are passed in at the same time (same thing for to). Ambiguity mainly that args are less verbose since the contents can be different types (regi vs tarball). I understand we find similar use case when specifying URI vs local file in bundle file and other configuration files. But in this case, the logic to handle ECR regi vs local file import/export is quite different if I understand it correctly. Thus I wanna bring up this alternative to see if it worths us to have different arg names for different behaviors.
> 
> another option is something you mentioned like `--to disk=mypath/to/image.tar`. so there is still some information about the type of the arg. but idk how frequent developers expect this kind of format.

Yeah, of course we can validate the multiple flags, but from an UI perspective, isn't that less clear than just having one flag for `to` and one flag for `from`, even if those flags can take different inputs?

Btw, this is similar to what `aws s3 cp` does, where you can pass an s3 url or a local path.

Another option is to have `--to disk --tar-path mypath/to/image.tar`. But I'm nor sure that's better.

jiayiwang7:
> > > we can validate the args and fail the cmd if `from-registry` and `from-disk` are passed in at the same time (same thing for `to`). Ambiguity is mainly that the args are less verbose since the contents can be different types (regi vs tarball). I understand we find similar use case when specifying URI vs local file in bundle file and other configuration files. But in this case, the logic to handle ECR regi vs local file import/export is quite different if I understand it correctly. Thus I wanna bring up this alternative to see if it worths us to have different arg names for different behaviors.
> > 
> > 
> > we can validate the args and fail the cmd if `from-registry` and `from-disk` are passed in at the same time (same thing for to). Ambiguity mainly that args are less verbose since the contents can be different types (regi vs tarball). I understand we find similar use case when specifying URI vs local file in bundle file and other configuration files. But in this case, the logic to handle ECR regi vs local file import/export is quite different if I understand it correctly. Thus I wanna bring up this alternative to see if it worths us to have different arg names for different behaviors.
> > another option is something you mentioned like `--to disk=mypath/to/image.tar`. so there is still some information about the type of the arg. but idk how frequent developers expect this kind of format.
> 
> Yeah, of course we can validate the multiple flags, but from an UI perspective, isn't that less clear than just having one flag for `to` and one flag for `from`, even if those flags can take different inputs?
> 
> Btw, this is similar to what `aws s3 cp` does, where you can pass an s3 url or a local path.
> 
> Another option is to have `--to disk --tar-path mypath/to/image.tar`. But I'm nor sure that's better.

yea make sense to me, go for it

vivek-koppuru:
I think regarding some of the conversion above, I think we should be more explicit in our commands if possible. Our target are either users who are used to kubectl or users who are less experienced than that. So using `aws s3 cp` as an example is not something I agree with following. That said, if we think that is the best UX, then that's fine.

I propose we have separate commands instead, like `download images` and `upload images`, and maybe having `import images that can do both of those commands. I think we can start off support just `import images` until users ask that they want it split up even, which should just be as easy as calling one less function to do the other action. I think of import as one action of importing the images to a desired location, and here the end goal is local registry from my understanding. But maybe for Snow that is not the case as you mentioned one step is not possible. If we want to go with this approach that you are suggesting instead of introducing new commands, maybe we can add a `download` and `upload` flag to signify what is happening, but at that point I think your approach is cleaner.

Also one other aspect to consider for the future is just thinking about only importing images that matter (which we talked offline and can be a separate task based on user feedback), but something to keep in mind when deciding when we decide on imperative commands and when we make use of the cluster spec. My vsphere template import design might have touched on that a little, so wondering if that is something where we want to document the description of this PR into a design doc to talk about this example where imperative commands actually make more sense than starting off supporting cluster spec.

Since these are basically utility commands, I am okay with whatever makes the most sense as this is not something we can easily get user feedback on and we should just make a decision. I would say that I would rather not introduce too many different UX experiences across the different utility commands if possible, wherever it makes sense.

mrajashree:
> Add a new command import
Add a sub-command import images
Deprecate the existing command import-images

👍 to these changes

for the new flags for import-images:
> eksctl anywhere import images --from origin --to disk
eksctl anywhere import images --from disk --to registry


based on above discussions, I would also vote for having either separate `upload` and `download` commands. Or if we're having separate flags for the upload/download flow in the import images command itself, can't it be this:
```
 eksctl anywhere import images --download-to
 eksctl anywhere import images --upload-from 
```
 
because regarding this part in the proposed changes:
>--from origin --to disk
--from disk --to registry

currently import-images command takes in the cluster spec as input, which contains the registry mirror endpoint. Is that going to change? If not, then we don't need `--from origin` or `--to registry` flags for the commands right? So both commands will only contain either download dir or upload dir. basically the cluster spec will contain registry endpoint instead of it being in the command.

g-gaston:
@jiayiwang7 @vivek-koppuru @mrajashree 
Thanks for the feedback!

Taking all your comments into account, I propose these changes to the proposal:
* `import images` will only import images into a registry, loading them from a file in disk
* `download images` will pull all images in the Bundle from their origin to a tarbal in disk

Example
```console
eksctl anywhere download images --output path/file_images.tar
eksctl anywhere import images --input path/file_images.tar --registry https://djfkdks.xy/whatever
```

The rest of the proposal, including most of the implementation skeleton stays the same.

chrisdoherty4:
The concept is awesome. I'm not 100% on the cmds though (spanner inbound).

- What other 'things' do we download and import and does it effect the user experience?
- Should we be taking a leaf out of Docker's book and for granular cmds specify in noun-verb order `eksctl anywhere images download`.

Rationale: I don't think its clear/intuitive using a verb-noun approach what verbs apply to my noun. If we organize _granular_ commands under their nouns it becomes very clear at a glance what I can do with images.

It reads better when typing `eksctl anywhere download images` but when forming relationships with other sub commands it makes it difficult to know whats available.

The cmd set could be:

```
// We may add options like --registry to some of these cmds. These cmds are aimed at
// illustrating a noun-verb approach rather than the specific options.

// In the original proposal there's a download an import; they aren't cohesive. Download and upload or import and export work well together. Hence, I've used save and load here.
eksctl anywhere images save <mandatory path>
eksctl anywhere images load <mandatory path>

// We still want an explicit cmd that synchronizes upstream with the target registry in 1
// cmd. 
eksctl anywhere images sync
```

Side note: while we aren't necessarily tied to Docker, these cmds have the benefit of compliment docker cmds.

g-gaston:
> The concept is awesome. I'm not 100% on the cmds though (spanner inbound).
> 
>     * What other 'things' do we download and import and does it effect the user experience?
> 
>     * Should we be taking a leaf out of Docker's book and for granular cmds specify in noun-verb order `eksctl anywhere images download`.
> 
> 
> Rationale: I don't think its clear/intuitive using a verb-noun approach what verbs apply to my noun. If we organize _granular_ commands under their nouns it becomes very clear at a glance what I can do with images. It might read better when typing `eksctl anywhere download images`, but when forming relationships with other sub commands it makes it difficult to know what applies hence I don't think the UX is good. This problem grows as more cmds are added.
> 
> The cmd set could be (take note of the commentary on cohesiveness):
> 
> ```
> // We may add options like --registry to some of these cmds. These cmds are aimed at
> // illustrating a noun-verb approach rather than the specific options.
> 
> // In the original proposal and cl there's a download an import; they aren't cohesive. 
> // Download and upload or import and export work well together. Hence, I've used 
> // 'save' and 'load' here to completely move away from the connotations the other verbs
> // have.
> $ eksctl anywhere images save <mandatory path>
> $ eksctl anywhere images load <mandatory path>
> 
> // We still want an explicit cmd that synchronizes upstream with the target registry in 1
> // cmd. 
> $ eksctl anywhere images {sync|import}
> ```
> 
> 'save' and 'load' can be debated given they don't necessarily imply network io which would happen. `download` and `upload` could work fine with `--to` and `--from` (or `--output` and `--input`) args respectively.

@chrisdoherty4 all fair points.

Regarding noun-verb vs verb-noun, I see your point but I don't think it's that black and white. I see it as a tradeoff between easiness to explore commands (noun + verb) and expressiveness/reads more natural (noun + verb).

That said, I think that ship has already sailed. All our commands use verb + noun and changing it now will completely destroy cohesiveness and I believe it was very much an intentional decision from the beginning (maybe trying to follow `kubectl` and other kubernetes CLIs, but I'm not sure of this). I'm always open to reconsider any passed decision but in this case, I don't feel comfortable making it in isolation. If we want to move to noun + verb it should be a team discussion and we should plan accordingly how to transition existing commands (migration, deprecation, etc.). If you think this a change worth pursuing, let's set up some time to talk to the team.

As for the verbs we use, I would like to keep `download` since we already have `download artifacts`, plus I think it represents slightly better the fact that we are saving to disk. As for making the two verbs semantically cohesive, I agree `download` and `import` are not "paired" in English, but I think the context where they are being used (container images) makes them work together. That said, if everyone prefers `upload` instead of `import`, I can make that change. @jiayiwang7 @vivek-koppuru @mrajashree opinions?

chrisdoherty4:
> maybe trying to follow kubectl and other kubernetes CLIs, but I'm not sure of this

I could see that being the case. I'm not convinced Kubernetes CLI is a good thing to follow though.

```
kubectl create deployment ... // works
kubectl create pod ... // bork :facepalm:
kubectl run pod ... // works
```

g-gaston:
PR updated with the final code for `import images` command

cmd/eksctl-anywhere/cmd/import_images.go
```diff
@@ -0,0 +1,125 @@
...
+		Reader:  deps.ManifestReader,
+		Bundles: bundle,
+		ImageMover: docker.NewImageMover(
+			docker.NewDiskSource(dockerClient, imagesFile),
+			docker.NewRegistryDestination(dockerClient, c.RegistryEndpoint),
+		),
+		ChartImporter: helm.NewChartRegistryImporter(
+			deps.Helm, artifactsFolder,
+			c.RegistryEndpoint,
+			os.Getenv("REGISTRY_USERNAME"),
```
jiayiwang7, line 118: should we validate env exists first?

cmd/eksctl-anywhere/cmd/internal/commands/artifacts/import_tools_image.go
```diff
@@ -0,0 +1,41 @@
...
+
+type ImportToolsImage struct {
+	Bundles            *releasev1.Bundles
+	ImageMover         ImageMover
+	UnPackager         UnPackager
+	InputFile          string
+	TmpArtifactsFolder string
+}
+
+type UnPackager interface {
```
jiayiwang7, line 20: should this be Packager?

cmd/eksctl-anywhere/cmd/import_images.go
```diff
@@ -0,0 +1,125 @@
...
+	dockerClient := executables.BuildDockerExecutable()
+	toolsImageFile := filepath.Join(artifactsFolder, eksaToolsImageTarFile)
+
+	// Import the eksa tools image into the registry first, so it can be used immediately
+	// after to build the helm executable
+	importToolsImage := artifacts.ImportToolsImage{
+		Bundles:            bundle,
+		InputFile:          c.InputFile,
+		TmpArtifactsFolder: artifactsFolder,
+		UnPackager:         tar.NewPackager(),
```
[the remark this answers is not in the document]
g-gaston: Packager is the struct that is able to package and unpackage
But this command receives a UnPackager entity, because it only needs to unpackage. Imo it's ok to have on struct that implements multiple interfaces, but I think entities should accept only the smaller interface they need (in general)

cmd/eksctl-anywhere/cmd/internal/commands/artifacts/import_tools_image.go
g-gaston: See https://github.com/aws/eks-anywhere/pull/1478#discussion_r845248872

cmd/eksctl-anywhere/cmd/import_images.go
g-gaston: Done

PR #1483, opened 2022-03-15
only validate flux after upgrade of cluster

danbudris:
the current validation occurs too soon before the cluster upgrade flow begins and can cause preflight validations to fail due to machine deployments still being in scaling up/down phase. Validating flux after the upgrade, as we had done prior to 11 days go, is sufficient for testing the feature.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/test/e2e/flux_test.go
+++ b/test/e2e/flux_test.go
@@ -20,7 +20,6 @@ const (
 func runUpgradeFlowWithFlux(test *framework.ClusterE2ETest, updateVersion v1alpha1.KubernetesVersion, clusterOpts ...framework.ClusterE2ETestOpt) {
 	test.GenerateClusterConfig()
 	test.CreateCluster()
-	test.ValidateFlux()
 	test.UpgradeCluster(clusterOpts)
 	test.ValidateCluster(updateVersion)
 	test.ValidateFlux()


test/e2e/flux_test.go
```diff
@@ -20,7 +20,6 @@ const (
 func runUpgradeFlowWithFlux(test *framework.ClusterE2ETest, updateVersion v1alpha1.KubernetesVersion, clusterOpts ...framework.ClusterE2ETestOpt) {
 	test.GenerateClusterConfig()
 	test.CreateCluster()
-	test.ValidateFlux()
```
jaxesn, line 23: Do we some kind of wait tho?  Wont upgradecluster end up failing if the state as expected?

vivek-koppuru: Yea I added this check thinking that we probably want to validate that flux was installed after create cluster itself

vivek-koppuru (approved):
/lgtm

test/e2e/flux_test.go
vivek-koppuru: Spoke with @danbudris and because we had this before, this is a quick addressing of the issue while we unblock the e2e tests and follow up with checking for the phase in the `ValidateFlux`

danbudris: once the issue is narrowed down we can follow up with https://github.com/aws/eks-anywhere/pull/1484 which adds the scaling checks to the flux validations

PR #1484, opened 2022-03-15
check for machine deployment scaling phase in flux validations

danbudris:
*Issue #, if available:*

*Description of changes:*
add a check for the Phase of the MachineDeployment to the worker node scaling validation check in the flux tests.

Previously, we were only checking for number of replicas; these replicas could be present but still in a ScalingUp or ScalingDown phase. We need to make sure we check the phase and wait until the MD stabalizes and is out of these phases before completing the validation. 

This will allow us to validate the state of the flux install on the cluster prior to an upgrade without the preflight validations of the upgrade catching the cluster in a ScalingUp or ScalingDown phase, which cause the validations to fail. 

*Testing (if applicable):*

### DIFF
--- a/test/framework/flux.go
+++ b/test/framework/flux.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"path"
 	"path/filepath"
+	"sigs.k8s.io/cluster-api/api/v1beta1"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


danbudris:
holding until we confirm that the removal of the `ValidateFlux` check prior to an upgrade does in fact resolve the failing tests; if it does, then we know that was in fact the issue (and not some other change), and get these scaling phase checks in the validation.

test/framework/flux.go
```diff
@@ -595,9 +597,9 @@ func (e *ClusterE2ETest) waitForWorkerScaling(targetvalue int) error {
 
 		for _, d := range md {
 			r := int(d.Status.Replicas)
-			if r != targetvalue {
-				e.T.Logf("Waiting for worker node MachineDeployment %s replicas to scale; target: %d, actual: %d", d.Name, targetvalue, r)
-				return fmt.Errorf(" MachineDeployment %s replicas are not at desired scale; target: %d, actual: %d", d.Name, targetvalue, r)
+			if r != targetvalue || d.Status.Phase == v1beta1.ScalingUpReason || d.Status.Phase == v1beta1.ScalingDownReason {
```
vivek-koppuru, line 600: Shouldn't we check for the inequality of whatever the ready phase is here?

danbudris:
this is stale; closing

PR #1485, opened 2022-03-16
Fix upgrade tests to use WithClusterUpgrade

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
The upgrade from latest tests were incorrectly using the `ClusterFiller` function instead of the `ClusterUpgrade` function, which was probably resulting in updating the wrong file. 

*Testing (if applicable):*
Ran the `TestDockerKubernetes121to122UpgradeFromLatestMinorRelease` test locally


jaxesn (approved):
/lgtm

PR #1490, opened 2022-03-16
Ensure e2e test hardware is powered off before tinkerbell tests

gwesterfieldjr:
- Add methods to pbnj client to power on, power off, and get machine power state.

- Update e2e tests to ensure all machines used for the test run are powered off before cluster create command.


pkg/providers/tinkerbell/pbnj/client.go
```diff
@@ -37,16 +47,67 @@ func NewPBNJClient(pbnjGrpcAuth string) (*Pbnj, error) {
...
+}
+
+func (p *Pbnj) PowerOff(ctx context.Context, bmcInfo BmcSecretConfig) error {
+	powerState, err := p.GetPowerState(ctx, bmcInfo)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get power state when trying to power off machine: %s %v", bmcInfo.Host, err)
+	}
+
+	// if power state is already off pbnj throws an error so we only issue the call if not already in off state
```
chrisdoherty4, line 73: Document the race condition. If a machine, for whatever reason, powers off _after_ the `GetPowerState()` but before `ExecutePowerRequest()` then we'll receive an error. It would be useful to maintainers to explicitly point that out in a short comment.

chrisdoherty4: Alternatively, given this is a gRPC call, you could do what we talked about and check the contents of the error. With gRPC errors are free-form text so its brittle but unless the upstream instruments an error payload with more data there's not much we can do.

chrisdoherty4: Yea I would do that. Combined with a clear error message it'll be clear pretty quickly that the code is looking for the wrong error message because it'll be erroring all the time. 

We probably should log when we swallow the error though. You'll want to inject a logger and log at a reasonably high verbosity level - I don't know what the teams current approach is to defining levels.

gwesterfieldjr: Ok I pushed the update. Also after furthre testing I found error only occurs if powered off. On powered on if you issue a power on command ipmi does not throw an error.

chrisdoherty4 (approved):

PR #1496, opened 2022-03-16
Upgrade go version to 1.17

a-cool-train:
*Description of changes:*
Upgrading go version to 1.17 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aws/eks-anywhere
 
-go 1.16
+go 1.17
 
 require (
 	github.com/aws/aws-sdk-go v1.38.40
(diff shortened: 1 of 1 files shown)


a-cool-train:
> We need to make some build tooling changes I think to support the new version, so let's get that done too. Any other motivation to upgrade apart from staying up to date?

Yes, we are currently planning on integrating eks-a-packages and the package is currently using go 1.17 and this is causing some versioning concerns.

abhay-krishna:
Was `go mod tidy` run after updating the go.mod file?

jaxesn:
Youll need to bump [this](https://github.com/aws/eks-anywhere/blob/main/Makefile#L23).

chrisdoherty4:
Why 1.17 when 1.18 released?

Issue #1436, opened 2022-03-10
Decouple eks-d crd & release manifest to maintain consistency with core component upgrades


taneyland:
The CRDs for eks-a are updated in the core component upgrades, and we later apply the changes to the eks-a objects.

We will follow the same pattern for eks-d: upgrade the CRDs in core component upgrades and then after the cluster is upgraded, create the new eks-d release object (if there is a new one).

PR discussion: https://github.com/aws/eks-anywhere/pull/1391#discussion_r823964727

PR #1497, opened 2022-03-16
Add EKS-D to core components upgrade

taneyland:
*Issue #, if available:*
#1436 

*Description of changes:*
This PR updates the EKS-D upgrade behavior to follow the same pattern as EKS-A components upgrade --> upgrade the CRDs in core component upgrades and then after the cluster is upgraded, create the new eks-d release object (if there is a new one).

Renamed the test components file name to allow us to use that file in different use cases

*Testing (if applicable):*
Created & upgraded a cluster from this branch
Created with 0.7.0, upgraded from this branch
Viewed controller logs with Tilt


pkg/cluster/fetch.go
```diff
@@ -71,12 +68,8 @@ func GetEksdReleaseForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bu
...
 	eksd, err := fetch(ctx, versionsBundle.EksD.Name, constants.EksaSystemNamespace)
 	if err != nil {
-		if strings.Contains(err.Error(), eksdReleaseObjDoesNotExist) {
-			logger.V(4).Info("EKS-D release objects are not present on the cluster. EKS-D release manifest will be retrieved from the URL in the bundle")
-			return nil, nil
-		} else {
-			return nil, fmt.Errorf("failed fetching EKS-D release for cluster: %v", err)
-		}
+		logger.V(4).Info("EKS-D release objects could not be fetched from the cluster. EKS-D release manifest will be retrieved from the URL in the bundle")
+		return nil, nil
```
taneyland, line 72: Instead of trying to string match all of the potential errors from kubectl when the release objs aren't there, we will just default to the normal behavior of getting it from the URL from the versions bundle if the EKS-D release objects cannot be fetched from the cluster

pkg/clustermanager/cluster_manager.go
```diff
@@ -999,6 +1004,17 @@ func (c *ClusterManager) applyResource(ctx context.Context, cluster *types.Clust
 	return nil
 }
 
+func (c *ClusterManager) applyEksdManifest(ctx context.Context, cluster *types.Cluster, eksdComponents *cluster.EksdManifests) error {
+	if err := c.Retrier.Retry(
+		func() error {
+			return c.clusterClient.ApplyKubeSpecFromBytesWithNamespace(ctx, cluster, eksdComponents.ReleaseManifestContent, constants.EksaSystemNamespace)
+		},
+	); err != nil {
+		return fmt.Errorf("error applying eksd release manifest: %v", err)
```
mitalipaygude, line 1013: adding the keyword `error` adds a new redundant keyword and can make an error message look confusing. can we remove the `error` word?

PR #1498, opened 2022-03-16
Fixed indents and graphics in vSphere prep doc

chrisnegus:
A few indents and image links needed to be fixed in the vsphere-prereq documentation.


TerryHowe (approved):
/lgtm

PR #1507, opened 2022-03-17
Instrument tinkerbell hardware generation

chrisdoherty4:
Switch the `generate hardware` cmd to use the reworked implementation of the `hardware` package. 

The new implementation facilitates validation on user input data and a separation of parsing and registration ensuring we don't register any hardware if some of the hardware is erroneous. This is useful when we need to inspect the entire data set to establish all hardware have unique host names.

The explicit validation performs a series of checks that provide faster feedback for end users. The validations performed include:

- [All static validation on a single machine instance](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/hardware/machine.go#L35).
- Unique hardware IDs in the set.
- Unique IP addresses in the set.
- Unique mac addresses in the set
- Unique hostnames in the set.
- Unique BMC IP addresses.

BMC validation is skipped if BMC configuration isn't supplied (see [`HasBmc()`](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/hardware/machine.go#L30) for semantics). This allows users to not specify a BMC and consequently not have BMC configuration generated complimenting the skip power actions flag when running `create cluster`.

Manual tests were performed to compare existing output with new output and everything maps correctly.


cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -62,105 +54,90 @@ func init() {
...
 
-	csv, err := hardware.NewCsvParser(hOpts.csvPath)
+	csvFile, err := os.Open(hOpts.csvPath)
 	if err != nil {
-		return err
+		return fmt.Errorf("csv: %v", err)
 	}
 
-	defer csv.Close()
+	outputDir, err := hardware.CreateDefaultDir()
```
gwesterfieldjr, line 67: Did you get rid of the outputPath cli option to override output of the harware-manifest directory to a different location besides always going into working dir?

pkg/providers/tinkerbell/hardware/json.go
```diff
@@ -14,17 +16,15 @@ import (
 // ErrTinkebellHardwareJsonRepeatWrites occurs when a TinkerbellHardwareJson receives multiple calls to Write().
 var ErrTinkebellHardwareJsonRepeatWrites = errors.New("TinkerbellHardwareJson can only be written to once")
 
-// TinkerbellHardwareJsonFactory creates new TinkerbellHardwareJson instances. The path used by the
-// TinkerbellHardwareJson is available as the second return parameter.
+// TinkerbellHardwareJsonFactory creates new TinkerbellHardwareJson instances.
 type TinkerbellHardwareJsonFactory interface {
-	Create(name string) (*TinkerbellHardwareJson, string, error)
```
chrisdoherty4, line 20: I changed the approach to retrieving all the data we want to register by instrumenting a journal when we write the output. See [`RecordingTinkerbellHardwareJsonFactory()`](https://github.com/aws/eks-anywhere/pull/1507/files#diff-17e36707f7afa6b3bdcdd809870cb9d2456e9b703b4cfc26a73fa9b90279d9ccR154) for details.

cmd/eksctl-anywhere/cmd/generatehardware.go
gwesterfieldjr: Please pass it to CreateDefaultDir so we can still support this functionality.

cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -62,105 +54,90 @@ func init() {
...
-	for {
-		items, err := csv.Read()
-		if err == io.EOF {
-			break
-		}
-		if err != nil {
-			return fmt.Errorf("failed reading csv: %v", err)
-		}
-
-		var id string
```
gwesterfieldjr, line 135: What happens if user provides guids in csv as we do now during e2e? Are you still accounting for this? The guids provided should be used rather than generating them

gwesterfieldjr (changes requested):

cmd/eksctl-anywhere/cmd/generatehardware.go
chrisdoherty4: Wowzers, I hadn't spotted that, thanks.

cmd/eksctl-anywhere/cmd/generatehardware.go
chrisdoherty4: This is fixed in a non-intutive manner. TL;DR the business logic is, appropriately, wrapped up in the `CsvReader` where it allows the Csv ID column to be optional and applies a generated UUID if not present.

gwesterfieldjr (approved):

PR #1513, opened 2022-03-17
Snow context unware defaults and validations

jiayiwang7:
*Issue #, if available:*

#1105 

*Description of changes:*

Add context unaware set defaults and validations for snow provider with the new config manager.

*Testing (if applicable):*


pkg/providers/snow/apibuilder_test.go
```diff
@@ -283,6 +283,7 @@ func TestSnowMachineTemplate(t *testing.T) {
 	got := SnowMachineTemplate(tt.machineConfigs["test-wn"])
 	wantAMIID := "eks-d-v1-21-5-ubuntu-ami-02833ca9a8f29c2ea"
 	wantSSHKey := "default"
+	wantPhysicalNetworkConnector := "" // TODO: update this value once configManager is hooked
```
g-gaston, line 286: curious: why does this need to wait?

pkg/cluster/snow.go
```diff
@@ -0,0 +1,81 @@
+package cluster
+
+import anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+
+func snowEntry() *ConfigManagerEntry {
```
g-gaston, line 5: We should probably add a test parsing one snow cluster
We don't need to reuse the same test as the other providers
It probably make sense to have one specific for snow that only checks snow datacenter and snow machines

g-gaston (commented):
Two minor comments, rest looks good

pkg/api/v1alpha1/snowmachineconfig.go
```diff
@@ -69,3 +72,35 @@ func GetSnowMachineConfigs(fileName string) (map[string]*SnowMachineConfig, erro
 	}
 	return configs, nil
 }
+
+func validateSnowMachineConfig(config *SnowMachineConfig) error {
```
g-gaston, line 76: We should test this, probably through `SnowMachineConfig.Validate()`

pkg/api/v1alpha1/snowmachineconfig.go
```diff
@@ -69,3 +72,35 @@ func GetSnowMachineConfigs(fileName string) (map[string]*SnowMachineConfig, erro
...
+		return fmt.Errorf("SnowMachineConfig SshKeyName is a required field. Create one in Snow Console if not any")
+	}
+
+	if config.Spec.InstanceType != SbeCLarge && config.Spec.InstanceType != SbeCXLarge && config.Spec.InstanceType != SbeC2XLarge && config.Spec.InstanceType != SbeC4XLarge {
+		return fmt.Errorf("SnowMachineConfig InstanceType %s is not supported, please use one of the following: %s, %s, %s, %s ", config.Spec.InstanceType, SbeCLarge, SbeCXLarge, SbeC2XLarge, SbeC4XLarge)
+	}
+	return nil
+}
+
+func setSnowMachineConfigDefaults(config *SnowMachineConfig) {
```
g-gaston, line 91: We should test this, probably through `SnowMachineConfig.SetDefaults()`

pkg/cluster/snow.go
jiayiwang7: i will add one in existing parsing test including cluster that covers the Ref to snow objects

g-gaston (approved):
/lgtm

PR #1515, opened 2022-03-17
Embed `cluster.Config` in `cluster.Spec`

g-gaston:
Continues the work started by #1343 

*Description of changes:*
This makes `cluster.Config` available almost everywhere in the CLI, since
we pass cluster.Spec as argument to most calls.

Removed a few parsed api structs from `cluster.Spec` that are not used
anymore.

Removed the GitOps default method from `cluster.Spec` and use the one for
`cluster.Config` when necessary (just in tests).

*Testing (if applicable):*
I ran all unit test and a couple e2e for vsphere and docker, all good.
However, I want to keep this on hold until we release, to avoid extra noise if any of the e2e tests fail.

### DIFF
--- a/pkg/cluster/spec.go
+++ b/pkg/cluster/spec.go
@@ -4,12 +4,10 @@ import (
 	"embed"
 	"fmt"
 	"net/url"
-	"path"
 	"path/filepath"
 	"strings"
 
 	eksdv1alpha1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	eksav1alpha1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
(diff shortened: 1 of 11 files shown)


pkg/cluster/spec.go
```diff
@@ -305,7 +255,8 @@ func BuildSpecFromBundles(cluster *eksav1alpha1.Cluster, bundles *v1alpha1.Bundl
 	}
 
 	s.Bundles = bundles
-	s.Cluster = cluster
+	s.Config.Cluster = cluster
+	s.Cluster = s.Config.Cluster
```
jiayiwang7, line 259: why do we still need this? is it going to be deprecated in a following PR? If so, can we add //TODO comment saying it's going to be deprecated so folks are aware not using it anymore.

g-gaston: We don't need it, I missed this, removing

jiayiwang7 (approved):

PR #1518, opened 2022-03-17
Docker image mover

g-gaston:

*Description of changes:*

Allows to set a source and destination for the images and orchestrates
the whole flow. For now, it admits disk and source and destination the
original registry as source and an external registry as destination.

Ideally we would clean the images loaded by the source after loading
them in the destination if they weren't there in the first place. We
will implement this as an improvement on top of this first iteration.
The drawback of not doing that cleanup is that we leave images in the
local docker storage, taking space.

*Testing (if applicable):*


pkg/docker/mover.go
```diff
@@ -0,0 +1,50 @@
...
+
+type DockerClient interface {
+	PullImage(ctx context.Context, image string) error
+	PushImage(ctx context.Context, image string, endpoint string) error
+	LoadFromFile(ctx context.Context, filepath string) error
+	SaveToFile(ctx context.Context, filepath string, images ...string) error
+}
+
+type (
+	ImageSourceLoader      func(ctx context.Context, client DockerClient, images ...string) error
```
chrisdoherty4, line 16: Should these just be modeled as an interface?

chrisdoherty4: Looking a little further: the `ImageMover` depends on a client simply to pass it forward. Why does the client need to be passed forward as opposed to created with a particular client?

pkg/docker/mover.go
```diff
@@ -0,0 +1,50 @@
...
+	ImageDestinationLoader func(ctx context.Context, client DockerClient, images ...string) error
+)
+
+type ImageMover struct {
+	client             DockerClient
+	loadSource         ImageSourceLoader
+	writeToDestination ImageDestinationLoader
+}
+
+func NewImageMover(client DockerClient) *ImageMover {
```
chrisdoherty4, line 26: I think there's an invariant here that's not being kept. For `ImageMover` to be useful, it _requires_ a source and destination. Therefore, any constructing initializers should accept a source and destination _or_ the `Move` call should accept them. The latter is useful if you need to delay constructing the source an destination because it separates `ImageMover` construction from the source and destination. 

I'd need the context under which this operates to help more. I suspect it can just accept them in a constructor though.

PR #1521, opened 2022-03-18
E2e test

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds E2E testing and minor touch-ups that were uncovered as part of said testing

*Testing (if applicable):*
Unit tests pass, and validating e2e tests pass


test/e2e/upgrade_test.go
```diff
@@ -12,16 +13,17 @@ import (
...
+	vsphereCpVmMemoryUpdate            = 16384
+	vsphereCpDiskGiBUpdateVar          = 40
+	vsphereWlVmNumCpuUpdateVar         = 4
+	vsphereWlVmMemoryUpdate            = 16384
+	vsphereWlDiskGiBUpdate             = 40
+	vsphereFolderUpdateVar             = "/SDDC-Datacenter/vm/capv/e2eUpdate"
+	vsphereNetwork2UpdateVar           = "/SDDC-Datacenter/network/sddc-cgw-network-2"
+	vsphereNetwork3UpdateVar           = "/SDDC-Datacenter/network/sddc-cgw-network-3"
+	clusterNamespace                   = "test-namespace"
+	cloudstackComputeOfferingUpdateVar = "Larger Instance"
```
maxdrib, line 26: todo: get this from environment variable instead of hardcoding

test/framework/flux.go
```diff
@@ -588,6 +711,7 @@ func (e *ClusterE2ETest) waitForWorkerScaling(targetvalue int) error {
 	return retrier.Retry(120, time.Second*10, func() error {
 		md, err := e.KubectlClient.GetMachineDeployments(ctx,
 			executables.WithKubeconfig(e.managementKubeconfigFilePath()),
+			executables.WithNamespace("eksa-system"),
```
maxdrib, line 714: todo: understand why this wasn't included previously

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -530,7 +531,7 @@ func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterCon
 	if oldCsmc.Spec.ComputeOffering != newCsmc.Spec.ComputeOffering {
 		return true
 	}
-	return false
+	return !reflect.DeepEqual(oldCsmc.Spec.UserCustomDetails, newCsmc.Spec.UserCustomDetails)
```
wanyufe, line 534: Comments in my previous PR mentioned that deepequal is not recommended in non-test code.
May write a customer equal method for usercustomerdetails or loop through its keys to compare. Currently, we may not need to sort the keys before camparing.

scripts/eksa_components_override.sh
```diff
@@ -56,13 +56,13 @@ else
...
-    make release-manifests RELEASE_DIR=$REPO_ROOT/bin RELEASE_MANIFEST_TARGET=local-eksa-components.yaml
-    curl $BUNDLE_MANIFEST_URL -o $REPO_ROOT/bin/local-bundle-release.yaml
-    CLUSTER_CONTROLLER_OVERRIDE_IMAGE=$(yq e ".spec.versionsBundles[0].eksa.clusterController.uri" $REPO_ROOT/bin/local-bundle-release.yaml)
-    KUBE_RBAC_PROXY_OVERRIDE_IMAGE=$(yq e ".spec.versionsBundles[0].clusterAPI.kubeProxy.uri" $REPO_ROOT/bin/local-bundle-release.yaml)
-    $SED -i "s,public.ecr.aws/.*/eks-anywhere-cluster-controller:.*,${CLUSTER_CONTROLLER_OVERRIDE_IMAGE}," $REPO_ROOT/bin/local-eksa-components.yaml
-    $SED -i "s,public.ecr.aws/.*/brancz/kube-rbac-proxy:.*,${KUBE_RBAC_PROXY_OVERRIDE_IMAGE}," $REPO_ROOT/bin/local-eksa-components.yaml
-    yq e -i '.spec.versionsBundles[].eksa.components.uri |= "bin/local-eksa-components.yaml"' $REPO_ROOT/bin/local-bundle-release.yaml
-fi
\ No newline at end of file
+#if [[ "$CONFIG_FILES_CHANGED" != "" ]]; then
```
wanyufe: why these lines is comment out?

test/e2e/vsphere/vsphere_autoimport_test.go
```diff
@@ -0,0 +1,122 @@
+//go:build e2e
```
wanyufe: Are we adding e2e test for vsphere in this PR?

test/framework/cloudstack.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+)
+
+const (
+	cloudstackDomainVar                = "T_CLOUDSTACK_DOMAIN"
+	cloudstackZoneVar                  = "T_CLOUDSTACK_ZONE"
```
wanyufe, line 12: Should we add zone1, zone2 to support multiple-zone? Similar for network. 

not a block for this PR, we can create a issue/feature so that we remember to do it.

maxdrib: Agreed, will create an issue

pkg/providers/cloudstack/cloudstack.go
maxdrib: I can replace with
```	
        if len(oldCsmc.Spec.UserCustomDetails) != len(newCsmc.Spec.UserCustomDetails) {
		return true
	}
	for key, value := range oldCsmc.Spec.UserCustomDetails {
		if value != newCsmc.Spec.UserCustomDetails[key] {
			return true
		}
	}
	return false
```

Makefile
```diff
@@ -122,6 +122,7 @@ EKS_A_RELEASE_CROSS_PLATFORMS := $(foreach platform,$(EKS_A_PLATFORMS),eks-a-rel
 
 DOCKER_E2E_TEST := TestDockerKubernetes121SimpleFlow
 LOCAL_E2E_TESTS ?= $(DOCKER_E2E_TEST)
+CLOUDSTACK_E2E_TEST := TestCloudStackKubernetes121SimpleFlow
```
vivek-koppuru, line 125: We don't really need these Makefile changes

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
```diff
@@ -52,6 +52,10 @@ spec:
                   each other and a set of delegated administrators with some authority
                   over the domain and its subdomains
                 type: string
+              managementApiEndpoint:
```
vivek-koppuru, line 55: These changes are being introduced as part of e2e tests?

test/e2e/awsiam_test.go
```diff
@@ -109,6 +110,28 @@ func TestVSphereKubernetes121BottleRocketAWSIamAuth(t *testing.T) {
 	runAWSIamAuthFlow(test)
 }
 
+func TestCloudStackKubernetes120AWSIamAuth(t *testing.T) {
```
vivek-koppuru, line 113: Let's not include these special tests just yet, unless we anticipate these features being used right away. This may cause a lot of load in the short term until we consolidate some of these tests.

test/framework/cloudstack.go
```diff
@@ -0,0 +1,145 @@
...
+type CloudStack struct {
+	t           *testing.T
+	fillers     []api.CloudStackFiller
+	podCidr     string
+	serviceCidr string
+}
+
+type CloudStackOpt func(*CloudStack)
+
+func UpdateRedhatTemplate120Var() api.CloudStackFiller {
```
vivek-koppuru, line 51: We only support redhat for cloudstack for now?

test/framework/flux.go
vivek-koppuru: It was a recent change that we realized was a bug

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
maxdrib: These were missed in the previous PR, seems like make release-manifests was not run

test/framework/cloudstack.go
maxdrib: I believe those are the only images we have built. I know CAPC itself works with some Ubuntu images as well but I'm not sure if we've tested those at the EKS-A level. Maybe @vignesh-goutham has some more thoughts here

test/framework/flux.go
maxdrib: Should I remove this then? I thought these objects are in the eksa-system namespace

Makefile
maxdrib: Are you suggesting to remove all the new targets and just run the test tool manually?

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -110,6 +110,13 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 		)
 	}
 
+	if old.Spec.ManagementApiEndpoint != new.Spec.ManagementApiEndpoint {
```
jiayiwang7, line 113: we will need unit test for this

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1106,3 +1122,11 @@ func (p *cloudstackProvider) validateMachineConfigsNameUniqueness(ctx context.Co
 func machineDeploymentName(clusterName, nodeGroupName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, nodeGroupName)
 }
+
+func getHostnameFromUrl(rawurl string) (string, error) {
+	url, err := url.Parse(rawurl)
+	if err != nil {
+		return "", fmt.Errorf("#{rawurl} is not a valid url")
```
jiayiwang7, line 1129:
```suggestion
		return "", fmt.Errorf("%s is not a valid url", url)
```

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -741,6 +753,10 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
 
 		// Add no-proxy defaults
 		noProxyList = append(noProxyList, common.NoProxyDefaults...)
+		cloudStackManagementApiEndpointHostname, err := getHostnameFromUrl(datacenterConfigSpec.ManagementApiEndpoint)
+		if err == nil {
```
jiayiwang7, line 757: don't we need to handle the error case?

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
jiayiwang7: i think im ok to have a follow up tt

pkg/providers/cloudstack/cloudstack.go
maxdrib: done

pkg/providers/cloudstack/cloudstack.go
jiayiwang7: right, this is more context aware defaults/validations.. if it's too much work, im ok with creating an issue to follow up

jiayiwang7 (approved):

PR #1522, opened 2022-03-18
Simplify machine tee writer

chrisdoherty4:
The MultiMachineWriter better compliments the behavior we're modelling given Tee only outputs to 2 things and is a CLI tool. Similalry, the stdlib has a MultiWriter function so we better align.


pkg/providers/tinkerbell/hardware/multi.go
```diff
@@ -0,0 +1,25 @@
+package hardware
```
chrisdoherty4, line 1: File move, github hasn't picked it up because of the rename from `TeeWriter` to `multiWriter`.

PR #1526, opened 2022-03-18
Cilium policy mode

mrajashree:
*Issue #, if available:*

*Description of changes:*
Design doc for reference: https://github.com/aws/eks-anywhere/blob/main/designs/cilium-configuration-policy.md

This commit configures cilium to use the user supplied value for policy
enforcement mode, by passing setting it as a value for the helm chart.
The accepted values for this field by cilium are "default", "always" and "never"
If no value is passed, cilium uses "default" as the default value.
When "always" mode is used, cilium blocks all communication.
This can prevent block cluster creation. To prevent this, eks anywhere
will create networkPolicy objects to allow communication between pods crucial
to eks-a. For a management cluster, network policies will be created for all
core kubernetes pods, CAPI, eks-a and flux controllers. For workload cluster,
the network policy will only be created for the kube-system pods.

*Testing (if applicable):*
```
TestVSphereKubernetes121CiliumAlwaysPolicyEnforcementModeSimpleFlow
TestVSphereKubernetes120UbuntuTo121Upgrade
```

### DIFF
--- a/controllers/controllers/clusters/vsphere.go
+++ b/controllers/controllers/clusters/vsphere.go
@@ -271,8 +271,8 @@ func (v *VSphereClusterReconciler) reconcileCNI(ctx context.Context, cluster *an
 		if err != nil {
 			return reconciler.Result{}, err
 		}
-
-		ciliumSpec, err := cilium.GenerateManifest(ctx, specWithBundles)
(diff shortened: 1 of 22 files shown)


mrajashree:
opened smaller PRs

PR #1531, opened 2022-03-18
Don't set defaults when generating clusterconfig

mrajashree:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/api/v1alpha1/cluster.go
```diff
@@ -197,9 +197,6 @@ func GetClusterConfigFromContent(content []byte) (*Cluster, error) {
 	if err != nil {
 		return clusterConfig, err
 	}
-	if err := setClusterDefaults(clusterConfig); err != nil {
```
g-gaston, line 200: Can we update the doc comment for this method?

g-gaston (approved):

PR #1535, opened 2022-03-19
Modify worker node checks for e2e

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
This should fix the following test: TestVSphereUpgradeMulticlusterWorkloadClusterWithFlux

This will query for the right MachineDeployment instead of just wait on the first one the list returns, which is not the one we are changing as it would be for the mgt cluster

*Testing (if applicable):*
Ran test


jaxesn (approved):
/lgtm

PR #1536, opened 2022-03-19
Utilize clusterSpec.config to dynamically fetch snow objects

jiayiwang7:
*Issue #, if available:*

#1343 

*Description of changes:*

providerFactory was using the old`GetSnowMachineConfigs` to fetch the snow machine configs instead of from the new cluster.Config. Remove the datacenter and machineconfig from snow provider struct, and instead dynamically fetch them from cluster.Spec.

This pattern will be applied to all providers in the future.

*Testing (if applicable):*

- unit tests


mrajashree (approved):
/lgtm

PR #1539, opened 2022-03-20
Generate network policy based on cluster spec and provider

mrajashree:
*Issue #, if available:*

*Description of changes:*
If user selects "always" policy enforcement mode for cilium, all communication between pods including core k8s and CAPI pods will be disallowed, blocking cluster creation/upgrades. To avoid this, EKS-A will create NetworkPolicy objects to allow communication between pods crucial to EKS-A. Refer this design doc to see which network policies will be created based on the cluster spec: https://github.com/aws/eks-anywhere/blob/main/designs/cilium-configuration-policy.md
This new templater method will be called while generating manifest for cilium during create and upgrade if the policy enforcement mode is set to "always". The part of calling this during cilium installation/upgrade will follow in a separate PR along with an e2e test

*Testing (if applicable):*
unit tests 

### DIFF
--- a/pkg/networking/cilium/network_policy.yaml
+++ b/pkg/networking/cilium/network_policy.yaml
@@ -1,0 +1,242 @@
+{{- if .managementCluster }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-all-kube-system
+  namespace: kube-system
+spec:
+  podSelector: {}
+  ingress:
+  - {}
+  egress:
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: eksa-system
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-all-eksa-system
+  namespace: eksa-system
+spec:
+  podSelector: {}
+  ingress:
+  - {}
+  egress:
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+{{- if .gitopsEnabled }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .fluxNamespace }}
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-all-flux-system
+  namespace: {{ .fluxNamespace }}
+spec:
+  podSelector: {}
+  ingress:
+  - {}
+  egress:
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+{{- end }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: capi-system
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cert-manager
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: capi-kubeadm-bootstrap-system
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: capi-kubeadm-control-plane-system
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: etcdadm-bootstrap-provider-system
+---
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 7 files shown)


pkg/networking/cilium/network_policy.yaml
```diff
@@ -0,0 +1,242 @@
...
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+{{- end }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: capi-system
```
mrajashree, line 62: generating the namespaces here because they haven't been created yet. Cilium gets installed first on the workload cluster, followed by CAPI and then eks-a/flux controllers on the mgmt cluster.
The other option would have been to modify clusterctl InitInfrastructure, in every component attach a NetworkPolicy, because otherwise the pods won't come up when using "always" mode. This seems cleaner, and does not affect the cluster

pkg/networking/cilium/templater.go
```diff
@@ -73,6 +78,30 @@ func (c *Templater) GenerateManifest(ctx context.Context, spec *cluster.Spec) ([
...
+
+	/* k8s versions 1.21 and higher label each namespace with key `kubernetes.io/metadata.name:` and value is the namespace's name.
+	This can be used to create a networkPolicy that allows traffic only between pods within kube-system ns, which is ideal for workload clusters. (not needed
+	for mgmt clusters).
+	So we will create networkPolicy using this default label as namespaceSelector for all versions 1.21 and higher
+	For 1.20 we will create a networkPolicy that allows allow traffic to/from kube-system pods, and document this. Users can still modify it and add new policies
+	as needed*/
+	k8sVersion, err := semver.New(spec.VersionsBundle.KubeDistro.Kubernetes.Tag)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing kubernetes version %v: %v", spec.Cluster.Spec.KubernetesVersion, err)
```
mitalipaygude, line 96: nit: we can remove the keyword `error` while returning an Errorf

mrajashree: I think since fmt.Errorf doesn't include `error` keyword, it's fine to include it in the msg

mrajashree: I think since the error returned by fmt.Errorf doesn't include `error` keyword, it's fine to include it in the msg

pkg/networking/cilium/network_policy.yaml
```diff
@@ -0,0 +1,242 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: flux-system
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-all-flux-system
+  namespace: flux-system
```
jiayiwang7, line 47: same thing, make flux system a var

pkg/networking/cilium/network_policy.yaml
```diff
@@ -0,0 +1,242 @@
...
+  podSelector: {}
+  ingress:
+  - {}
+  egress:
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+{{- range $providerNamespace := .providerNamespaces }}
```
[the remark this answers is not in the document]
mrajashree: discussed offline that we'll let this be for now, once all deployments become part of cluster spec we will change this

pkg/networking/cilium/templater.go
```diff
@@ -73,6 +78,36 @@ func (c *Templater) GenerateManifest(ctx context.Context, spec *cluster.Spec) ([
...
 
+func (c *Templater) GenerateNetworkPolicyManifest(spec *cluster.Spec, namespaces []string) ([]byte, error) {
+	values := map[string]interface{}{
+		"managementCluster":  spec.Cluster.IsSelfManaged(),
+		"providerNamespaces": namespaces,
+	}
+
+	if spec.Cluster.Spec.GitOpsRef != nil {
+		values["gitopsEnabled"] = true
+		if spec.Config.GitOpsConfig != nil {
```
jiayiwang7, line 89: u can directly use `spec.GitOpsConfig`

jiayiwang7 (approved):

PR #1541, opened 2022-03-21
Always validate on tinkerbell hardware generate

chrisdoherty4:
When generating Tinkerbell manifests the `--dry-run` option should follow as closely as possible the normal path of operation. This includes performing the validation we would expect if the tool were to run normally. The change ensures validation does run even with `--dry-run`.

Additionally, change the `--dry-run` flag to `--skip-registration` so its clear what behavior to expect. Typical dry run behavior would still perform validations, for example, where as we don't want validation on the Tinkerbell stack config if we don't want to register the hardware.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generatehardware.go
+++ b/cmd/eksctl-anywhere/cmd/generatehardware.go
@@ -15,12 +15,12 @@ import (
 )
 
 type hardwareOptions struct {
-	csvPath      string
-	outputPath   string
-	tinkerbellIp string
-	grpcPort     string
-	certPort     string
-	dryRun       bool
+	csvPath          string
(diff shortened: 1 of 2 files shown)


chrisdoherty4:
> Looks good, but please update the generateHardwareConfig method in test/framework/cluster.go to pass the tinkerbell ip option from the env var as below. Otherwise, e2e will break due to validation error.
> 
> `generateHardwareConfigArgs := []string{"generate", "hardware", "--tinkerbell-ip", os.Getenv(tinkerbellServerEnvVar), "-f", e.HardwareCsvLocation, "-o", e.ClusterConfigFolder}`

As discussed offline, changed the flag to `skip-registration` so its more representative.

PR #1543, opened 2022-03-21
Rework ip uniqueness checks

chrisdoherty4:
The vSphere and Tinkerbell providers execute some logic to try and validate that an IP isn't in use. This required depending on an `IPGenerator` construct that supplied the 'is unique' behavior creating an unnecessary dependency. 

This change reworks the uniqueness check so its available independently of `IPGenerator`. Additionally, it removes a superfluous interface representing `IPGenerator` behaviors leaving it to consumers to define the behavior they require.


pkg/networkutils/networkutils.go
```diff
@@ -22,3 +23,22 @@ func ValidateIP(ip string) error {
 
 	return nil
 }
+
+// IsIPInUse performs a soft check to see if there are any services listening on a selection of common ports at
+// ip by trying to establish a TCP connection. Ports checked include: 22, 23, 80, 443 and 6443 (Kubernetes API Server).
+// Each connection attempt allows up-to 500ms for a response.
+//
+// todo(chrisdoherty) change to an icmp approach to eliminate the need for ports.
```
ptrivedi, line 31: Yeah, not a big fan of this

PR #1544, opened 2022-03-21
Refactoring cloudstack resource schema to use name/id instead of type/value

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR primarily refactors the cloudstack API to use the name/id interface instead of value/type to represent ambiguous entries for CloudStack resources, which may be either UUID's or simple names. 

It also introduces the managementApiEndpoint entry into the cluster spec, which will be used for setting no-proxy entries. It also introduces some extra arguments into the templates for ensuring the proper tls cipher is used.

*Testing (if applicable):*
Unit tests, e2e tests in cloudstack branch, testing in customer environment


pkg/api/v1alpha1/cloudstackdatacenterconfig.go
```diff
@@ -18,14 +18,11 @@ func NewCloudStackDatacenterConfigGenerate(clusterName string) *CloudStackDatace
...
-			Zones: []CloudStackZoneRef{
+			Zones: []CloudStackZone{
 				{
-					Zone: CloudStackResourceRef{
-						Value: "",
-						Type:  Name,
-					}, Network: CloudStackResourceRef{
-						Value: "",
-						Type:  Name,
+					Name: "",
```
chrisdoherty4, line 23: This is redundant as the zero value will be assumed when not specified.

pkg/api/v1alpha1/cloudstackdatacenterconfig.go
```diff
@@ -18,14 +18,11 @@ func NewCloudStackDatacenterConfigGenerate(clusterName string) *CloudStackDatace
...
+					Network: CloudStackResourceIdentifier{
+						Name: "",
```
chrisdoherty4, line 25: [Ditto.](https://github.com/aws/eks-anywhere/pull/1544/files?diff=split&w=0#r831414049)

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -29,35 +29,28 @@ type CloudStackDatacenterConfigSpec struct {
...
 	// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain name must also be provided.
 	Account string `json:"account,omitempty"`
+	// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+	ManagementApiEndpoint string `json:"managementApiEndpoint"`
 }
 
 type CloudStackResourceIdentifier struct {
-	Id   string `json:"id"`
-	Name string `json:"name"`
+	// +optional
```
chrisdoherty4, line 40: Is there some other documentation we could put here?

pkg/executables/cmk.go
```diff
@@ -42,13 +42,13 @@ func (c *Cmk) Close(ctx context.Context) error {
 	return nil
 }
 
-func (c *Cmk) ValidateTemplatePresent(ctx context.Context, domainId string, zoneId string, account string, template v1alpha1.CloudStackResourceRef) error {
+func (c *Cmk) ValidateTemplatePresent(ctx context.Context, domainId string, zoneId string, account string, template v1alpha1.CloudStackResourceIdentifier) error {
```
chrisdoherty4, line 45: Is there an open ticket to refactor `cmk.go`? Executables shouldn't be performing validations, they should just provide programmatic access to atomic APIs for the technology they represent.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -29,35 +29,28 @@ type CloudStackDatacenterConfigSpec struct {
...
+	ManagementApiEndpoint string `json:"managementApiEndpoint"`
 }
 
 type CloudStackResourceIdentifier struct {
-	Id   string `json:"id"`
-	Name string `json:"name"`
+	// +optional Id Unique CloudStack ID of the resource in the CLoudStack enironment. Mutually exclusive with Name
+	Id string `json:"id,omitempty"`
+	// +optional Name CloudStack Name of the resource in the CloudStack environment. Mutually exclusive with Id
+	Name string `json:"name,omitempty"`
```
chrisdoherty4, line 43: I think the comments are typically done above the `+` lines.

```suggestion
	// Id of a resource in the CloudStack environment. Mutually exclusive with Name
	// +optional
	Id string `json:"id,omitempty"`
	// Name of a resource in the CloudStack environment. Mutually exclusive with Id
	// +optional
	Name string `json:"name,omitempty"`
```

pkg/executables/cmk.go
maxdrib: Thanks for the comment. My reasoning for this verbiage was that the binary `cmk` doesn't expose a "get" operation - only list, with filters, which is what we are using. I couldn't call it "GetTemplate" since that's not really what cmk is doing. The only "validation" taking place here is ensuring there is exactly 1 result from the list operation, which doesn't feel like application level business logic to me. I'm inclined to keep the nomenclature/implementation as is

PR #1559, opened 2022-03-21
Create NetworkPolicy if required during cilium upgrade

mrajashree:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/726

*Description of changes:*
Refer design doc: https://github.com/aws/eks-anywhere/blob/main/designs/cilium-configuration-policy.md
[Previously merged PR](https://github.com/aws/eks-anywhere/pull/1552) creates NetworkPolicy objects during cluster creation if policy enforcement mode is set to `always`
This PR does the same during cluster upgrade, if the policy enforcement mode value has changed.
Along with this, it also upgrades cilium if the values passed to cilium chart have changed. Right now we only accept and configure the value for the policy enforcement mode so only that is checked

*Testing (if applicable):*
- `TestVSphereKubernetes121UbuntuTo122UpgradeCiliumPolicyEnforcementMode`
-  Created a cluster with no policy enforcement mode specified. The `cilium-config` cm had `enable-policy: default`. Changed cluster spec to use `never` as the value for policy enforcement mode. Configmap got updated. When value was changed to `always`, required network policies also got created
- `TestVSphereKubernetes121UbuntuUpgradeFromLatestMinorReleaseAlwaysNetworkPolicy`

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -94,7 +94,7 @@ type ClusterClient interface {
 
 type Networking interface {
 	GenerateManifest(ctx context.Context, clusterSpec *cluster.Spec, namespaces []string) ([]byte, error)
-	Upgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error)
+	Upgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec, namespaces []string) (*types.ChangeDiff, error)
 }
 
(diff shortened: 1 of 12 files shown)


vignesh-goutham (approved):
/lgtm

Question - Do we support/want to support upgrade from `always` to `default`, that would involve removing network policies, which we might want to think twice.

mrajashree:
@vignesh-goutham good question, if switching from `always` to `default`, user would have to remove all NetworkPolicy objects they created before upgrading. Let's document the NetworkPolicy objects we create, so they can delete those too

PR #1564, opened 2022-03-22
Modify policy enforcement changed check to include clusters created from 0.7

mrajashree:
*Issue #, if available:*

*Description of changes:*
design doc for reference: https://github.com/aws/eks-anywhere/blob/main/designs/cilium-configuration-policy.md
This [previous PR](https://github.com/aws/eks-anywhere/pull/1559) added checks to see if policy enforcement mode was modified. But the existing check works only for cases where the existing cluster has the cniConfig case and does not cover clusters that don't have cniconfig set (clusters on versions release-0.7 and lower). This PR checks if the existing spec does not have cniConfig initialized (only possible for clusters created with older version), and if the current policy enforcement mode is the non-default option

*Testing (if applicable):*
- Created cluster using release-0.7, upgraded with cluster.spec containing non-default policy enforcement mode
- `TestVSphereKubernetes121UbuntuUpgradeFromLatestMinorReleaseAlwaysNetworkPolicy`

### DIFF
--- a/pkg/networking/cilium/upgrader.go
+++ b/pkg/networking/cilium/upgrader.go
@@ -72,6 +72,7 @@ func (u *Upgrader) Upgrade(ctx context.Context, cluster *types.Cluster, currentS
 
 	if chartValuesChanged {
 		if newSpec.Cluster.Spec.ClusterNetwork.CNIConfig.Cilium.PolicyEnforcementMode == v1alpha1.CiliumPolicyModeAlways {
+			logger.V(3).Info("Installing NetworkPolicy resources for policy enforcement mode 'always'")
 			networkPolicyManifest, err := u.templater.GenerateNetworkPolicyManifest(newSpec, namespaces)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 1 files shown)


pkg/networking/cilium/upgrader.go
```diff
@@ -138,11 +139,15 @@ func ChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 }
 
 func ciliumHelmChartValuesChanged(currentSpec, newSpec *cluster.Spec) bool {
-	if currentSpec.Cluster.Spec.ClusterNetwork.CNIConfig != nil {
-		if currentSpec.Cluster.Spec.ClusterNetwork.CNIConfig.Cilium != nil {
```
vivek-koppuru, line 142: Are you missing this check now?

mrajashree: not checking on purpose because it won't be needed. For clusters created with release <= 0.7, CNIConfig != nil check will be enough. For clusters with newer versions, by this point in the cilium upgrader logic, CiliumConfig will be initialized so we don't need to check

vivek-koppuru (approved):
/lgtm

PR #1567, opened 2022-03-22
Fix ip generator rand source

chrisdoherty4:
Stop seeding the default rand source midway through program execution.

The default rand source should be seeded on program startup as a bootstrapping task, not midway through program execution. This change alters algorithms seeding the default rand by switching to a separate `Rand` instance. In doing so, we remove possible side effects/introducing unintended dependencies in logic executed at later stages.

### DIFF
--- a/pkg/networkutils/ipgenerator.go
+++ b/pkg/networkutils/ipgenerator.go
@@ -9,10 +9,14 @@ import (
 
 type IPGenerator struct {
 	netClient NetClient
+	rand      *rand.Rand
 }
 
 func NewIPGenerator(netClient NetClient) IPGenerator {
-	return IPGenerator{netClient: netClient}
+	return IPGenerator{
+		netClient: netClient,
+		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
 }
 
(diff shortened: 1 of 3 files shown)


danbudris (approved):

pkg/networkutils/ipgenerator.go
```diff
@@ -35,10 +39,9 @@ func (ipgen IPGenerator) GenerateUniqueIP(cidrBlock string) (string, error) {
 
 // generates a random ip within the specified cidr block
 func (ipgen IPGenerator) randIp(cidr *net.IPNet) (net.IP, error) {
-	rand.Seed(time.Now().UnixNano())
 	newIp := *new(net.IP)
 	for i := 0; i < 4; i++ {
-		newIp = append(newIp, byte(rand.Intn(256))&^cidr.Mask[i]|cidr.IP[i])
+		newIp = append(newIp, uint8(ipgen.rand.Intn(256))&^cidr.Mask[i]|cidr.IP[i])
```
ptrivedi, line 44: What's the reason for this? byte should be type-defined to uint8

chrisdoherty4: Left over from a small experiment, sorry. Thanks!

PR #1577, opened 2022-03-22
Adding upgrade components for cloudstack

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds the controller level changes for cloudstack to continue supporting the upgrade flow, with feature gating via environment variable on the deployment.

*Testing (if applicable):*
Unit tests, main business logic was tested in separate branch. E2e testing in main branch is yet to come


vivek-koppuru (commented):
Will review the rest after

controllers/controllers/resource/template.go
```diff
@@ -159,7 +166,146 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
...
+
+func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster *anywherev1.Cluster, clusterSpec *cluster.Spec, csdc anywherev1.CloudStackDatacenterConfig, cpCsmc, etcdCsmc anywherev1.CloudStackMachineConfig, workerCsmcs map[string]anywherev1.CloudStackMachineConfig) ([]*unstructured.Unstructured, error) {
+	if len(workerCsmcs) > 1 {
+		return nil, fmt.Errorf("multiple worker node group configuration is not supported in CloudStack provider")
+	}
+	workerNodeGroupMachineSpecs := make(map[string]anywherev1.CloudStackMachineConfigSpec, len(workerCsmcs))
+	for _, wnConfig := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		workerNodeGroupMachineSpecs[wnConfig.MachineGroupRef.Name] = workerCsmcs[wnConfig.MachineGroupRef.Name].Spec
+	}
+	// control plane and etcd updates are prohibited in controller so those specs should not change
```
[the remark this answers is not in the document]
maxdrib: I think that's something we want to support. What part of this PR doesn't support GitOps?

controllers/controllers/resource/reconciler.go
```diff
@@ -96,6 +103,42 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 			return err
 		}
 		resources = append(resources, r...)
+	case anywherev1.CloudStackDatacenterKind:
+		if !features.IsActive(features.CloudStackProvider()) {
```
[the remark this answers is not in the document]
maxdrib: Which webhooks? Do you mean cloudstackdatacenterconfig_webhooks and cloudstackmachineconfig_webhooks?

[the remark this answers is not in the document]
vivek-koppuru: Yes exactly, we can add checks for this feature there too.

controllers/controllers/resource/template.go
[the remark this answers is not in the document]
vivek-koppuru: Managing a workload cluster with GitOps, not just management cluster. With workload cluster, we can modify any part of the cluster spec like the CLI.

controllers/controllers/resource/reconciler.go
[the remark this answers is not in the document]
maxdrib: Added checks in webhooks

controllers/controllers/resource/template.go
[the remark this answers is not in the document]
maxdrib: I'm not sure I fully understand the problem. Do you think you could create a GH issue so that we can track it? Or is there an example in vsphere, for example, that we could reuse?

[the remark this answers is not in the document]
vivek-koppuru: We have an annotation that we set on workload clusters to signify what to do in the reconciliation. So we have these used in the webhooks and controllers. Some parts might be reusable (especially if it is related to the cluster). Then it's just a matter of handling changing the template names, which I think is already handled by this code.

machine config webhook: https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheremachineconfig_webhook.go#L80

template code: https://github.com/aws/eks-anywhere/blob/918bbd2e19c662ce70e13695027a5fb2d6ed8b73/controllers/controllers/resource/template.go#L69

pkg/clustermanager/cluster_manager.go
```diff
@@ -904,6 +906,11 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
 	if err = c.InstallEksdComponents(ctx, clusterSpec, cluster); err != nil {
 		return err
 	}
+	if features.IsActive(features.CloudStackProvider()) {
+		if err = c.clusterClient.SetControllerCloudStackProviderEnv(ctx, cluster.KubeconfigFile); err != nil {
```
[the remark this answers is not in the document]
maxdrib: Sounds good, will abstract out the specific environment variable being set and pass it in here as an argument

pkg/executables/kubectl.go
```diff
@@ -677,6 +678,15 @@ func (k *Kubectl) ValidateEKSAClustersCRD(ctx context.Context, cluster *types.Cl
 	return nil
 }
 
+func (k *Kubectl) SetControllerCloudStackProviderEnv(ctx context.Context, kubeconfig string) error {
+	params := []string{"set", "env", "deployment/eksa-controller-manager", fmt.Sprintf("%s=true", features.CloudStackProviderEnvVar), "--kubeconfig", kubeconfig}
```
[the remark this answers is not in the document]
maxdrib: will pass in cloudstack specific env var and value and make the kubectl function generic

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -588,11 +703,23 @@ func (p *cloudstackProvider) generateCAPISpecForCreate(ctx context.Context, clus
...
+		workloadTemplateNames[workerNodeGroupConfiguration.Name] = common.WorkerMachineTemplateName(clusterSpec.Cluster.Name, workerNodeGroupConfiguration.Name, p.templateBuilder.now)
+		kubeadmconfigTemplateNames[workerNodeGroupConfiguration.Name] = common.KubeadmConfigTemplateName(clusterSpec.Cluster.Name, workerNodeGroupConfiguration.Name, p.templateBuilder.now)
+		p.templateBuilder.WorkerNodeGroupMachineSpecs[workerNodeGroupConfiguration.MachineGroupRef.Name] = p.machineConfigs[workerNodeGroupConfiguration.MachineGroupRef.Name].Spec
+	}
+	workerOpt := func(values map[string]interface{}) {
+		values["workloadTemplateName"] = workloadTemplateNames[clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].Name]
+		values["workloadkubeadmconfigTemplateName"] = kubeadmconfigTemplateNames[clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].Name]
 	}
-	workersSpec, err = p.templateBuilder.GenerateCAPISpecWorkers(clusterSpec, workersOpt)
+	workersSpec, err = p.templateBuilder.GenerateCAPISpecWorkers(clusterSpec, workloadTemplateNames, kubeadmconfigTemplateNames, workerOpt)
```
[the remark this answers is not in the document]
maxdrib: I thought I remembered @jiayiwang7 and @wanyufe discussing this in depth as an issue for multiple worker node groups, but I just removed it and unit tests are passing so I'll revert that change for now

controllers/controllers/resource/fetcher.go
```diff
@@ -260,6 +269,54 @@ func (r *CapiResourceFetcher) VSphereEtcdMachineTemplate(ctx context.Context, cs
 	return vsphereMachineTemplate, nil
 }
 
+func (r *CapiResourceFetcher) CloudStackCluster(ctx context.Context, cs *anywherev1.Cluster, wnc anywherev1.WorkerNodeGroupConfiguration) (*cloudstackv1.CloudStackCluster, error) {
```
chrisdoherty4, line 272: Are there unit tests covering these?

controllers/controllers/resource/fetcher.go
```diff
@@ -430,6 +487,38 @@ func (r *CapiResourceFetcher) ExistingVSphereWorkerMachineConfig(ctx context.Con
 	return MapMachineTemplateToVSphereMachineConfigSpec(vsMachineTemplate, users)
 }
 
+func (r *CapiResourceFetcher) ExistingCloudStackDatacenterConfig(ctx context.Context, cs *anywherev1.Cluster, wnc anywherev1.WorkerNodeGroupConfiguration) (*anywherev1.CloudStackDatacenterConfig, error) {
```
chrisdoherty4, line 490: Ditto re unit tests?

controllers/controllers/resource/reconciler.go
```diff
@@ -96,6 +103,42 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
...
+		workerCsmcs := make(map[string]anywherev1.CloudStackMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations))
+		err := cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.DatacenterRef.Name}, csdc)
+		if err != nil {
+			return err
+		}
+		err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.ControlPlaneConfiguration.MachineGroupRef.Name}, cpCsmc)
+		if err != nil {
+			return err
+		}
+		if len(cs.Spec.WorkerNodeGroupConfigurations) != 1 {
```
chrisdoherty4, line 123: Is this not something that should also be a part of the validating webhook?

controllers/controllers/resource/reconciler.go
```diff
@@ -106,7 +149,7 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 		return fmt.Errorf("unsupport Provider %s", cs.Spec.DatacenterRef.Kind)
 	}
 
-	// Reconcling IdentityProviders
+	// Reconciling IdentityProviders
```
chrisdoherty4, line 152:
```suggestion
	// Reconcile IdentityProviders
```

controllers/controllers/resource/template.go
```diff
@@ -38,6 +39,12 @@ type VsphereTemplate struct {
 	now anywhereTypes.NowFunc
 }
 
+type CloudStackTemplate struct {
+	ResourceFetcher
```
chrisdoherty4, line 43: Why are these embedded? I'm not sure why a `CloudStackTemplate` would provide fetching behavior?

Jumping the gun: if its because the other constructs do this, ignore the other constructs.

controllers/controllers/resource/template.go
```diff
@@ -159,7 +166,146 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
...
+	oldCsdc, err := r.ExistingCloudStackDatacenterConfig(ctx, eksaCluster, clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0])
+	if err != nil {
+		return nil, err
+	}
+	oldCpCsmc, err := r.ExistingCloudStackControlPlaneMachineConfig(ctx, eksaCluster)
+	if err != nil {
+		return nil, err
+	}
+
+	controlPlaneTemplateName, err := r.getControlPlaneTemplateName(ctx, eksaCluster, oldCsdc, csdc, oldCpCsmc, cpCsmc, clusterName)
```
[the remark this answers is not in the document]
chrisdoherty4: Levels of abstraction apply to methods and functions also. Functions don't _just_ serve a reusability use-case. By extracting the detail to other lower level of abstraction methods we can focus this higher level on orchestration of those lower levels making it extremely clear, assuming good method/function names, what we're doing.

Our code base doesn't do enough of this and I've noticed in many places we _mix_ levels of abstraction which is generally bad as its harder to understand.

pkg/executables/kubectl.go
```diff
@@ -677,6 +677,15 @@ func (k *Kubectl) ValidateEKSAClustersCRD(ctx context.Context, cluster *types.Cl
 	return nil
 }
 
+func (k *Kubectl) SetControllerEnvVar(ctx context.Context, envVar, envVarVal, kubeconfig string) error {
```
chrisdoherty4, line 680: Should we be providing a mechanism for toggling stuff on post deployment? That seems dangerous because what you deploy originally might not be what's running in the end due to arbitrary logic altering the environment.

controllers/controllers/resource/fetcher.go
maxdrib: Looks like they're missing for this method, good catch. However, they seem to be missing across the board for all providers, so I'm inclined to create an issue about it for follow-up.

controllers/controllers/resource/reconciler.go
maxdrib: Do you mean on create webhook? Upgrade?

controllers/controllers/resource/fetcher.go
chrisdoherty4: I don't think existing code not having unit tests is justification for not writing unit tests. We'd end up merging a lot of untested code if we took that as precedent. 

An issue is required either way, no need to write missing unit tests.

controllers/controllers/resource/template.go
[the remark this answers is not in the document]
maxdrib: Looking into this, it seems like a decent number of changes are required, so I'm inclined to hold off on implementation and create an issue instead. The current code supports multi-cluster creation/management which should suffice for the time being

controllers/controllers/resource/reconciler.go
chrisdoherty4: Whatever it applies to? Forgive me, I'm not 100% how we're using webhooks at present but based on other comments we already have a validating webhook so I'm not quite tracking why this wouldn't be there?

pkg/executables/kubectl.go
chrisdoherty4: I expect a secret to be changed because they rotate. Unlike secrets that are typically data, we're toggling behavior with this configuration which seems awfully dangerous to me. @vivek-koppuru do you have any suggestions on how else this could be done? 'easiest' isn't necessarily 'right'.

controllers/controllers/resource/fetcher.go
maxdrib: I agree. Created https://github.com/aws/eks-anywhere/issues/1644

controllers/controllers/resource/template.go
maxdrib: Created https://github.com/aws/eks-anywhere/issues/1645

controllers/controllers/resource/fetcher.go
maxdrib: One contributing factor to that is it's difficult to mock the client.Reader dependency. It populates values by reference and doesn't actually return any values. This might require more effort

controllers/controllers/resource/template.go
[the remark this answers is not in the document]
vivek-koppuru: I think this shouldn't actually require a crazy amount of changes? It should just be checking for those annotations in those resources like in the machine config webhook above right?

controllers/controllers/resource/reconciler.go
vivek-koppuru: We can probably read the DatacenterRef to see if CloudStack is specified and then check based on that and the feature flag here.

pkg/executables/kubectl.go
vivek-koppuru: Yea I thought about this but refrained from commenting on it because I couldn't think of a different way without affecting the manifest itself. I think the right way would be to add this condition in our manifest code through our bundle, but that requires more refactoring efforts in terms of scope. I think we should be able to provide a way to edit our deployment env vars in the future, so if we have this as an action item we can address it then. Since this is behind a feature flag and we know that the blast radius is minimized, I am okay with this current approach right now.

chrisdoherty4: Lets commit to this. @vivek-koppuru would you be able to raise an issue with some of your ideas?

controllers/controllers/resource/fetcher.go
maxdrib: Added unit tests here

controllers/controllers/resource/reconciler.go
maxdrib: Done

controllers/controllers/resource/fetcher.go
maxdrib: Added

controllers/controllers/resource/template.go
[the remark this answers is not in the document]
maxdrib: Will add a check for the annotations in the webhook

pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
```diff
@@ -53,6 +53,11 @@ func (r *CloudStackMachineConfig) ValidateUpdate(old runtime.Object) error {
 		return nil
 	}
 
+	if oldCloudStackMachineConfig.IsManagement() {
```
vivek-koppuru, line 56: This check alone is actually redundant. I was referring to having a list of fields that we don't want immutable, but we can have it as a follow up before we remove the feature flag if you can log it as an issue.

https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheremachineconfig_webhook.go#L58

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -588,11 +705,19 @@ func (p *cloudstackProvider) generateCAPISpecForCreate(ctx context.Context, clus
 	if err != nil {
 		return nil, nil, err
 	}
-	workersOpt := func(values map[string]interface{}) {
-		values["workloadTemplateName"] = common.WorkerMachineTemplateName(clusterName, clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].Name, p.templateBuilder.now)
-		values["cloudstackWorkerSshAuthorizedKey"] = p.workerSshAuthKey
+
+	if len(clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations) > 1 {
```
vivek-koppuru, line 709: This check should probably happen earlier.

pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
vivek-koppuru: https://github.com/aws/eks-anywhere/issues/1669

vivek-koppuru (approved):

PR #1581, opened 2022-03-22
Print command output before returning error

abhay-krishna:
In case of failures, these print statements could give us useful info.


danbudris (approved):
/lgtm

PR #1588, opened 2022-03-23
Add support to upgrade Kubernetes version through GitOps for Docker

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Updating KubernetesVersion through GitOps for Docker is currently not supported, so this will update the machine templates appropriately with the kind node image. It checks to see if there is a difference in version compared to what is specified on KubeadmControlPlane. For worker nodes because we can roll them out on a self managing cluster as well, we will check if there is a newer kind image in the bundle as well in order to decide whether we need a new template or not.

Followup: Need to enable support for changes to the KubeadmConfigTemplate spec as well.

*Testing (if applicable):*
Local testing with tilt against the TestDockerUpgradeWorkloadClusterWithFlux

### DIFF
--- a/controllers/controllers/resource/fetcher.go
+++ b/controllers/controllers/resource/fetcher.go
@@ -19,6 +19,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	kubeadmv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	dockerv1 "sigs.k8s.io/cluster-api/test/infrastructure/docker/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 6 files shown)


jaxesn (commented):
I ran into an issue one time trying to upgrade a capd cluster where the kube version was the same but the kind image, which is somewhere in the spec as custom image or something, was different. It ended up erroring out with something about that custom image field being immutable, I think that was an indication that we need to make a new md or whatever. Not sure if this solves that and/or if we need, but felt related so calling it out.

vivek-koppuru:
> I ran into an issue one time trying to upgrade a capd cluster where the kube version was the same but the kind image, which is somewhere in the spec as custom image or something, was different. It ended up erroring out with something about that custom image field being immutable, I think that was an indication that we need to make a new md or whatever. Not sure if this solves that and/or if we need, but felt related so calling it out. 

This may not resolve it but I can modify this to resolve that case too. I decided not to address it in the PR initially but might be worth it.

danbudris (approved):
/lgtm

PR #1592, opened 2022-03-23
Upgrade go version to 1.17

a-cool-train:
*Description of changes:*
Upgrading go to a version that is compatible with curated packages.


test/e2e/tools/eks-anywhere-test-tool/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/eks-anywhere-test-tool
 
-go 1.16
+go 1.17
```
chrisdoherty4, line 3: Did you try 1.18?

a-cool-train: @chrisdoherty4, this is not the permanent upgrade to the latest version. The destination branch is a custom branch. I have created this branch to get eks-anywhere to be compatible with eks-curated-packages which currently uses go version 1.17.

go.mod
```diff
@@ -49,5 +136,5 @@ replace (
 	github.com/tinkerbell/cluster-api-provider-tinkerbell => github.com/pokearu/cluster-api-provider-tinkerbell v0.0.0-20220128001529-79d851d0861f
 
 	// need the modifications eksa made to the capi api structs
-	sigs.k8s.io/cluster-api => github.com/mrajashree/cluster-api v1.0.3-0.20220301005127-382d70d4a76f
+	sigs.k8s.io/cluster-api => github.com/a-cool-train/cluster-api v1.0.3-custom.0.20220321135508-ebcbebe2dbfe
```
lewisdiamond, line 139: Is this necessary?

a-cool-train: Yes momentarily. Cluster-api had an incompatible logr version, I fixed it in my fork and it might take sometime to get the necessary approvals. Once I get the approvals, I'll be reverting this change.

vivek-koppuru: Wait so did we fix this? If not, we should actually change it in Rajashree's fork itself because we use that as the one that includes all of our patches from main.

Makefile
```diff
@@ -20,10 +20,11 @@ SHELL := /bin/bash
 ARTIFACTS_BUCKET?=my-s3-bucket
 GIT_VERSION?=$(shell git describe --tag)
 GIT_TAG?=$(shell git tag -l --sort -v:refname | head -1)
-GOLANG_VERSION?="1.16"
-GO ?= $(shell source ./scripts/common.sh && build::common::get_go_path $(GOLANG_VERSION))/go
+GOLANG_VERSION?="1.17"
+GO_VERSION ?= $(shell source ./scripts/common.sh && build::common::get_go_path $(GOLANG_VERSION))
```
ewollesen, line 24: The name of this variable is a little misleading (it's not a version, it's a path) and a little confusing (we also have a GOLANG_VERSION, but that's something completely different).

A better name for this would be GOBIN.

release/Makefile
```diff
@@ -41,6 +41,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Setup Go
+GOLANG_VERSION?="1.17"
+GO_VERSION ?= $(shell source ../scripts/common.sh && build::common::get_go_path $(GOLANG_VERSION))
```
ewollesen, line 46: Like my other comment, this name is misleading and/or confusing.

scripts/make_attribution.sh
```diff
@@ -44,12 +44,13 @@ source "$REPO_ROOT/scripts/attribution_helpers.sh"
 function build::attribution::generate(){
     cd $REPO_ROOT
     build::common::use_go_version "$GOLANG_VERSION"
-    go mod vendor
+    local -r GO=$(build::common::get_go_path "$GOLANG_VERSION")/go
```
ewollesen, line 47: Was this necessary? In theory, `use_go_version` should take care of the PATH env var, so you can simply call `go`. That said, this doesn't hurt I guess.

Issue #1430, opened 2022-03-10
Refactor `BuildSpecForCluster` to better account for the growing number of args


taneyland:
Refactor `BuildSpecForCluster` to take an interface since the number of arguments is increasing. This will make it easier to add more options here as it continues to grow,

Original post: https://github.com/aws/eks-anywhere/pull/1391#discussion_r821188968

PR #1593, opened 2022-03-23
Refactor BuildSpecForCluster to take in an interface

taneyland:
*Issue #, if available:*
#1430 

*Description of changes:*
Refactor this method to take in an interface instead of adding more parameters as we introduce new fetcher options. This will make it easier to add more fetcher options as we continue to add more features

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -23,20 +23,27 @@ type EksdReleaseFetch func(ctx context.Context, name, namespace string) (*eksdv1
 
 type OIDCFetch func(ctx context.Context, name, namespace string) (*v1alpha1.OIDCConfig, error)
 
-func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, oidcFetch OIDCFetch) (*Spec, error) {
-	bundles, err := GetBundlesForCluster(ctx, cluster, bundlesFetch)
+type Fetcher interface {
+	BundlesFetch(ctx context.Context, name, namespace string) (*v1alpha1release.Bundles, error)
+	GitOpsFetch(ctx context.Context, name, namespace string) (*v1alpha1.GitOpsConfig, error)
+	EksdReleaseFetch(ctx context.Context, name, namespace string) (*eksdv1alpha1.Release, error)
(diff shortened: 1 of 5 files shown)


pkg/cluster/fetch.go
```diff
@@ -23,20 +23,27 @@ type EksdReleaseFetch func(ctx context.Context, name, namespace string) (*eksdv1
 
 type OIDCFetch func(ctx context.Context, name, namespace string) (*v1alpha1.OIDCConfig, error)
 
-func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, oidcFetch OIDCFetch) (*Spec, error) {
-	bundles, err := GetBundlesForCluster(ctx, cluster, bundlesFetch)
+type Fetcher interface {
```
chrisdoherty4, line 26: Why force an implementation to know about all these things? When we want another 'thing' we have to extend the interface which is bad. Perhaps you're looking for a config object?

```go
type BuildSpecForClusterConfig struct {
  BundlesFetcher BundlesFetcher
  ...
}

func BuildSpecForClsuter(context.Context, BuildSpecForClusterConfig) ...
```

pkg/cluster/fetch.go
```diff
@@ -23,20 +23,27 @@ type EksdReleaseFetch func(ctx context.Context, name, namespace string) (*eksdv1
...
+	OIDCFetch(ctx context.Context, name, namespace string) (*v1alpha1.OIDCConfig, error)
+}
+
+func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, fetchers Fetcher) (*Spec, error) {
+	bundles, err := GetBundlesForCluster(ctx, cluster, fetchers.BundlesFetch)
 	if err != nil {
 		return nil, err
 	}
-	gitOpsConfig, err := GetGitOpsForCluster(ctx, cluster, gitOpsFetch)
+	gitOpsConfig, err := GetGitOpsForCluster(ctx, cluster, fetchers.GitOpsFetch)
```
chrisdoherty4, line 38: FWIW, this contributed to my [other comment](https://github.com/aws/eks-anywhere/pull/1593/files#r834413447).

The fact none of the behaviors on the new interface are used in conjunction with each other makes the statement that the behaviors are not cohesive and shouldn't be on the same interface, i.e. this is a code smell.

PR #1594, opened 2022-03-23
Strip superfluous 'error' prepended to error message context

chrisdoherty4:
Remove superfluous 'error' prefix from error messages when wrapping with context.

When constructing error messages and passing them up the call stack its superfluous to prepend the context with the word 'error' and unnecessarily elongates the message. In general, only constructs that _present_ the error to the end user should prepend 'error' (for example Cobra when we return from the `RunE` funcs).


danbudris (approved):
/lgtm

PR #1597, opened 2022-03-23
Fix Audit log constant error on api-server

maxdrib:
*Issue #, if available:*

*Description of changes:*
The current CAPC template has a regression in that the audit log is being managed by eks-a instead of the api-server controller, so it doesn't get rotated.

*Testing (if applicable):*
Unit tests pass, cluster created in customer environment


vignesh-goutham (approved):
/lgtm

PR #1598, opened 2022-03-23
Adding validation on managementApiEndpoints to be consistent

maxdrib:
*Issue #, if available:*

*Description of changes:*
The compromise discussed exposes the managementApiEndpoint for the cloudstack provider in two ways: environment variable, and cluster spec. This decision was made to be future-proof. This PR adds validation on cluster creation to ensure that the two values are the same.

*Testing (if applicable):*
Unit tests pass

### DIFF
--- a/pkg/providers/cloudstack/validator.go
+++ b/pkg/providers/cloudstack/validator.go
@@ -58,6 +58,9 @@ func (v *Validator) ValidateCloudStackDatacenterConfig(ctx context.Context, data
 	if len(datacenterConfig.Spec.Domain) <= 0 {
 		return fmt.Errorf("CloudStackDatacenterConfig domain is not set or is empty")
 	}
+	if datacenterConfig.Spec.ManagementApiEndpoint == "" {
+		return fmt.Errorf("CloudStackDatacenterConfig managementApiEndpoint is not set or is empty")
+	}
 	_, err := getHostnameFromUrl(datacenterConfig.Spec.ManagementApiEndpoint)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/providers/cloudstack/validator.go
```diff
@@ -61,6 +62,14 @@ func (v *Validator) ValidateCloudStackDatacenterConfig(ctx context.Context, data
 	if err != nil {
 		return fmt.Errorf("error while checking management api endpoint: %v", err)
 	}
+	execConfig, err := decoder.ParseCloudStackSecret()
+	if err != nil {
+		return fmt.Errorf("parsing cloudstack secret: %v", err)
+	}
+	if execConfig.ManagementUrl != datacenterConfig.Spec.ManagementApiEndpoint {
+		return fmt.Errorf("cloudstack secret management url (%s) differs from cluster spec management url (%s)",
```
vignesh-goutham, line 70: This is probably a great error message. Im wondering if it makes sense to also have a validation if the ManagementApiEndpoint provided on cluster spec to be non-empty, and throw an appropriate message at that point?

maxdrib: Sure, I can add another check on the provided spec's managementApiEndpoint to make sure it's nonempty

maxdrib: actually the managementApiEndpoint is a required field per https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go#L36. Does that error on empty string inputs?

vignesh-goutham (approved):
/lgtm

Issue #1602, opened 2022-03-23
A suffix ':443' to the registry mirror URL doesn't work in a certain environment. 


wongni:
**What happened**:
The `import-images` command was not working with the certain docker registry. We noticed that there is a suffix `:443` that’s added in the local image tag: 

`
Pushing
{"image": "[docker.apple.com:443/l0g8r8j6/kubernetes-sigs/cluster-api/kubeadm-bootstrap-controller:v1.0.2-eks-a-v0.0.0-dev-cloudstack-build.0](http://docker.apple.com:443/l0g8r8j6/kubernetes-sigs/cluster-api/kubeadm-bootstrap-controller:v1.0.2-eks-a-v0.0.0-dev-cloudstack-build.0)”}
`

**What you expected to happen**:
When the mirror registry endpoint port is not specified or the same as the default HTTPS port (443) the generated image URL should not have ':443'. 

**How to reproduce it (as minimally and precisely as possible)**:
Creating a cluster with registry mirror to a registry does not work with ':443' in the image URLs. 

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: 0.7.2
- EKS Distro Release: 0.88.0

ptrivedi:
@wongni do you mind posting the error you are seeing with this?

wongni:
Looks like image mirroring with the combination of `download images` and `import images` commands does not have this issue. Check the following logs. 

```
$ eksctl anywhere download images --output artifacts.tar 
<completed>

$ REGISTRY_USERNAME=yyy REGISTRY_PASSWORD=xxx eksctl anywhere import images --input artifacts.tar --registry docker.company.com -v 9 --bundles manifest.yaml

...
2022-08-04T19:30:28.033Z        V0      Writing images to registry
2022-08-04T19:30:28.033Z        V3      Starting registry write {"numberOfImages": 1}
2022-08-04T19:30:28.033Z        V0      Tagging image   {"image": "public.ecr.aws/eks-anywhere/cli-tools:v0.7.2-eks-a-12", "local image": "docker.company.com/eks-anywhere/cli-tools:v0.7.2-eks-a-12"}
2022-08-04T19:30:28.034Z        V6      Executing command       {"cmd": "/usr/bin/docker tag public.ecr.aws/eks-anywhere/cli-tools:v0.7.2-eks-a-12 docker.company.com/eks-anywhere/cli-tools:v0.7.2-eks-a-12"}
[...]

PR #1603, opened 2022-03-23
Skip appending ':443: suffix to the image URLs 

wongni:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/1602

*Description of changes:* Skip appending ':443: suffix to the image URLs when the registry mirror endpoint port is not specified or the same as the default HTTPS port. 

*Testing (if applicable):*


cmd/eksctl-anywhere/cmd/importimages.go
```diff
@@ -83,13 +83,16 @@ func importImages(ctx context.Context, spec string) error {
 	if err != nil {
 		return err
 	}
+	endpoint := net.JoinHostPort(host, port)
+	if port == constants.DefaultHttpsPort {
+		endpoint = host
+	}
 	for _, image := range images {
-		if err := importImage(ctx, de, image.URI, net.JoinHostPort(host, port)); err != nil {
+		if err := importImage(ctx, de, image.URI, endpoint); err != nil {
```
maxdrib, line 91: how was this tested? It seems like it may have impact across many providers, no?

abhinavmpandey08:
Does appending the ":443" port cause any issues with pushing the images?

cmd/eksctl-anywhere/cmd/importimages.go
jiayiwang7: this won't work. 443 is required for some registry (harbor) or it will fail pulling images

jiayiwang7: > > Does appending the ":443" port cause any issues with pushing the images?
> 
> A certain customer registry rejects the push request. They manually tested the push after removing the suffix ":443", and it worked in their environment.

Does the customer set up a proxy in front? Can it be the reason?

wongni: > > > Does appending the ":443" port cause any issues with pushing the images?
> > 
> > 
> > A certain customer registry rejects the push request. They manually tested the push after removing the suffix ":443", and it worked in their environment.
> 
> Does the customer set up a proxy in front? Can it be the reason?

I don't think they have a proxy in front of the registry. 
They are using 
```
registryMirroConfiguration:
  endpoint: docker.[customer name].com
```

vignesh-goutham:
Different registries might have different behaviour when it comes to pushing images. I suggest we /hold this until we research a bit more on the topic.

cmd/eksctl-anywhere/cmd/importimages.go
jiayiwang7: yea understand. the conditional check works for me. just it changes the existing eksa behavior a bit (since it proposes to remove the default 443 port). we need to update the doc if we introduce the change.

abhinavmpandey08:
> > Does appending the ":443" port cause any issues with pushing the images?
> 
> A certain customer registry rejects the push request. They manually tested the push after removing the suffix ":443", and it worked in their environment.

Are you running the registry on port 443? Or is it running on a different port? Also, can you share the exact error you see when you try to run `docker push <some-registry>:443/<some-image>`

robbo10:
We are also running into the exact same issue, has there been any progress made with the research?

Issue #1604, opened 2022-03-23
Make cloudstack preflight check timeout configurable


wongni:
**What would you like to be added**:
The timeout value in cmk.ini is a hard-coded 30 (seconds). In a certain environment, this is not long enough for cmk's "sync" command.

**Why is this needed**:
Let's make the timeout value configurable with the default value of 30.

PR #1605, opened 2022-03-23
Feature/cmk timeout configurable

wongni:
*Issue #, if available:* [1604](https://github.com/aws/eks-anywhere/issues/1604)

*Description of changes:*
Make the cloudstack cloud config's timeout value configurable with the default value of 30.

*Testing (if applicable):*
Unit testing. 


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -445,6 +445,7 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
 		"cloudstackManagementApiEndpoint":            managementApiEndpoint,
 		"managerImage":                               bundle.CloudStack.ClusterAPIController.VersionedImage(),
 		"verifySsl":                                  verifySsl,
+		"timeout":                                    timeout,
```
maxdrib, line 448: these fields only need to be set if we're passing the value to CAPC, which we're not in this case. Same thing for TemplateMapMD

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -52,11 +52,20 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
 			return nil, fmt.Errorf("'verify-ssl' has invalid boolean string %s: %v", verifySslValue, err)
 		}
 	}
+	timeout, err := section.GetKey("timeout")
```
maxdrib, line 55: I don't think this is the right place to put this field. It seems like this should be an independent environment variable, not encoded in the b64 exec config string. This encoded string will be passed to CAPC directly and it doesn't make sense to include this variable there in my opinion.

maxdrib (changes requested):

pkg/providers/cloudstack/decoder/decoder.go
maxdrib: The only inputs for using cloudstack with eks-a are this b64-encoded ini file, which contains the cloudstack management url, api key, secret key, and verify-ssl boolean, and then the eks-a cluster spec. I was proposing we introduce a new environment variable for overriding the timeout since that only applies to the preflight check, whereas Wonkun's initial implementation had it as part of the b64-encoded input ini file

maxdrib (approved):

chrisdoherty4 (approved):

PR #1606, opened 2022-03-23
Remove CAPC asset packaging in prod releases

abhay-krishna:
The CAPC assets are added to bundle for only dev releases (L281). This is a duplication that causes it go get applied for prod releases too.


abhay-krishna:
Force merging since it's a oneline change that doesnt affect any functionality.

PR #1587, opened 2022-03-23
WIP: Check if chart needs auth

mrajashree:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1609, opened 2022-03-24
Remove need for auth with helm

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
As talked about in https://github.com/aws/eks-anywhere/pull/1587, we actually don't need the auth here as initially expected like docker. We thought we needed it because the repo we were testing against was private, so after I made it public, it works now.

*Testing (if applicable):*
Ran both `import-images` and `create cluster`

### DIFF
--- a/cmd/eksctl-anywhere/cmd/importimages.go
+++ b/cmd/eksctl-anywhere/cmd/importimages.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


cmd/eksctl-anywhere/cmd/importimages.go
```diff
@@ -52,11 +51,6 @@ var importImagesCmd = &cobra.Command{
 }
 
 func importImages(ctx context.Context, spec string) error {
-	registryUsername := os.Getenv("REGISTRY_USERNAME")
-	registryPassword := os.Getenv("REGISTRY_PASSWORD")
-	if registryUsername == "" || registryPassword == "" {
-		return fmt.Errorf("username or password not set. Provide REGISTRY_USERNAME and REGISTRY_PASSWORD for importing helm charts (e.g. cilium)")
-	}
```
TerryHowe, line 59: If you do need auth, should probably use `HELM_REGISTRY_CONFIG=~/.docker/config.json` to keep things simple

vivek-koppuru: That's a great suggestion! We actually don't need the auth however

PR #1622, opened 2022-03-24
Feature list packages

a-cool-train:
*Description of changes:*
Package Discovery


TerryHowe:
The 0.8.0 branch is cut, so this could probably merge to main, right? Not sure what is wrong with the attributions presubmit

a-cool-train:
> The 0.8.0 branch is cut, so this could probably merge to main, right? Not sure what is wrong with the attributions presubmit

Yes it is but currently there are a few pending discussions on curated packages as a dependency. I didn't want those discussions to hold up the cli from bugbash delivery. Once those discussions are completed, I will be raising the PRs as cloudstack team have done.

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,61 @@
...
+type listPackagesOption struct {
+	kubeVersion string
+	source      string
+}
+
+var lpo = &listPackagesOption{}
+
+func init() {
+	listCmd.AddCommand(listPackagesCommand)
+	listPackagesCommand.Flags().StringVar(&lpo.source, "source", "", "Discovery Location. Options (cluster, registry)")
```
ewollesen, line 22: You could use a custom flag here to handle validation for you in a consistent way.

https://cs.opensource.google/go/go/+/go1.18:src/flag/flag.go;l=893

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,61 @@
...
+	listPackagesCommand.Flags().StringVar(&lpo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	err := listPackagesCommand.MarkFlagRequired("source")
+	if err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var listPackagesCommand = &cobra.Command{
+	Use:          "packages",
+	Short:        "Generate a list of curated packages available to install",
```
ewollesen, line 32:
```suggestion
	Short:        "Lists curated packages available to install",
```

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,61 @@
...
+	Long:         "This command is used to generate a list of curated packages available to install",
```
ewollesen, line 33:
```suggestion
	Long:         "This command is used to generate a list of curated packages available to install",
```
Probably no need for this, since it doesn't add anything over the short value.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+		packageBundle, err = getLatestBundleFromRegistry(ctx, kubeVersion)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return packageBundle, nil
+}
+
+func createBundleManager(kubeVersion string) (manager bundle.Manager) {
```
ewollesen, line 44: Idiomatic Go would say to name this function `newBundleManager`.

cmd/eksctl-anywhere/cmd/validations.go
```diff
@@ -33,3 +39,24 @@ func commonValidation(ctx context.Context, clusterConfigFile string) (*v1alpha1.
 	}
 	return clusterConfig, nil
 }
+
+func sourceValidation(source string) error {
```
ewollesen, line 43: Also, it's more idiomatic to name functions that start with a verb[^1], ie validateSource or checkSource.

1: the exception is getter methods, they don't need get https://go.dev/doc/effective_go#Getters

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+
+func getPackageBundle(ctx context.Context, kubectl *executables.Kubectl, opts ...executables.KubectlOpt) (*api.PackageBundle, error) {
+	stdOut, err := kubectl.GetResources(ctx, "packageBundle", opts...)
+	if err != nil {
+		return nil, err
+	}
+	obj := &api.PackageBundle{}
+	if err = json.Unmarshal(stdOut.Bytes(), obj); err != nil {
+		return nil, fmt.Errorf("error parsing packageBundle response: %v", err)
+	}
```
ewollesen, line 90:
```suggestion
	obj := &api.PackageBundle{}
	if err = json.NewDecoder(stdOut).Decode(obj); err != nil {
		return nil, fmt.Errorf("unmarshaling package bundle: %w", err)
	}
```

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+
+func getActiveController(ctx context.Context, kubectl *executables.Kubectl, opts ...executables.KubectlOpt) (*api.PackageBundleController, error) {
+	stdOut, err := kubectl.GetResources(ctx, "packageBundleController", opts...)
+	if err != nil {
+		return nil, err
+	}
+	obj := &api.PackageBundleControllerList{}
+	if err = json.Unmarshal(stdOut.Bytes(), obj); err != nil {
+		return nil, fmt.Errorf("error parsing packageBundleController response: %v", err)
+	}
```
ewollesen, line 102:
```suggestion
	if err = json.NewDecoder(stdOut).Decode(obj); err != nil {
		return nil, fmt.Errorf("unmarshaling active package bundle controller: %w", err)
	}
```

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+	obj := &api.PackageBundle{}
+	if err = json.Unmarshal(stdOut.Bytes(), obj); err != nil {
+		return nil, fmt.Errorf("error parsing packageBundle response: %v", err)
+	}
+	return obj, nil
+}
+
+func getActiveController(ctx context.Context, kubectl *executables.Kubectl, opts ...executables.KubectlOpt) (*api.PackageBundleController, error) {
+	stdOut, err := kubectl.GetResources(ctx, "packageBundleController", opts...)
```
ewollesen, line 95: You can get the active controller by namespace/name:

https://github.com/aws/eks-anywhere-packages/blob/c0ebc872b77728fac33600d59e82703d206e2953/pkg/bundle/manager.go#L25

pkg/curatedpackages/common.go
```diff
@@ -0,0 +1,16 @@
+package curatedpackages
+
+import (
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/executables"
+)
+
+func createKubectl(ctx context.Context) (*dependencies.Dependencies, error) {
```
ewollesen, line 10: Idiomatic Go says name this "newDependencies".

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,55 @@
...
+
+	w.Init(os.Stdout, MinWidth, TabWidth, Padding, PadChar, flags)
+	defer w.Flush()
+
+	fmt.Fprintf(w, "\n %s\t%s\t", "Package", "Version(s)")
+	fmt.Fprintf(w, "\n %s\t%s\t", "----", "----")
+	for key, values := range m {
+		fmt.Fprintf(w, "\n %s\t%s\t", key, strings.Join(values, ","))
+	}
+}
```
ewollesen, line 34:
```suggestion
func DisplayPackages(packages []api.BundlePackage) {
	w := new(tabwriter.Writer)
	w.Init(os.Stdout, MinWidth, TabWidth, Padding, PadChar, flags)
	defer w.Flush()

	fmt.Fprintf(w, "\n %s\t%s\t", "Package", "Version(s)")
	fmt.Fprintf(w, "\n %s\t%s\t", "----", "----")
	for _, pkg := range packages {
		versions := convertBundleVersionToPackageVersion(pkg.Source.Versions)
		fmt.Fprintf(w, "\n %s\t%s\t", pkg.Name, strings.Join(versions, ","))
	}
}
```

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,55 @@
...
+	return packagesInBundle
+}
+
+func packagesToString(packages []api.BundlePackage) map[string][]string {
+	ptopv := make(map[string][]string)
+	for _, p := range packages {
+		ptopv[p.Name] = append(ptopv[p.Name], convertBundleVersionToPackageVersion(p.Source.Versions)...)
+	}
+	return ptopv
+}
```
ewollesen, line 47: Assuming the above, then this goes away:
```suggestion

```

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+		packageBundle *api.PackageBundle
+		err           error
+	)
+
+	switch strings.ToLower(source) {
+	case Cluster:
+		packageBundle, err = getActiveBundleFromCluster(ctx, kubeConfig)
+	case Registry:
+		packageBundle, err = getLatestBundleFromRegistry(ctx, kubeVersion)
+	}
```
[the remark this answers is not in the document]
chrisdoherty4: I wouldn't panic. We don't know at this point what supplied the incorrect source so we should still return an error. 

To appreciate why we can look at what actually happens: the user supplies the source. That's a user input error so I'd expect the error to bubble up and exit the program gracefully. The sticking point though is that this function has insufficient context to justify a panic. 

```go
return nil, errors.New("unknown source")
```

pkg/curatedpackages/bundle.go
chrisdoherty4: Drop the return variable also, it can have unintended side effects.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func createBundleManager(kubeVersion string) (manager bundle.Manager) {
+	versionSplit := strings.Split(kubeVersion, ".")
+	major, minor := versionSplit[0], versionSplit[1]
+	log := logr.Discard()
+	discovery := testutil.NewFakeDiscovery(major, minor)
+	puller := artifacts.NewRegistryPuller()
+	bm := bundle.NewBundleManager(log, discovery, puller)
+	return bm
```
chrisdoherty4, line 51:
```suggestion
	return bundle.NewBundleManager(log, discovery, puller)
```

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func getLatestBundleFromRegistry(ctx context.Context, kubeVersion string) (*api.PackageBundle, error) {
+	bm := createBundleManager(kubeVersion)
+	bundle, err := bm.LatestBundle(ctx, RegistryBaseRef)
+	return bundle, err
```
chrisdoherty4, line 57:
```suggestion
	return bm.LatestBundle(ctx, RegistryBaseRef)
```

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,55 @@
...
+	defer w.Flush()
+
+	fmt.Fprintf(w, "\n %s\t%s\t", "Package", "Version(s)")
+	fmt.Fprintf(w, "\n %s\t%s\t", "----", "----")
+	for key, values := range m {
+		fmt.Fprintf(w, "\n %s\t%s\t", key, strings.Join(values, ","))
+	}
+}
+
+func GetPackages(bundle *api.PackageBundle) []api.BundlePackage {
```
chrisdoherty4, line 36: This function is redundant, just have the consumer reach to `bundle.Spec.Packages`.

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,55 @@
...
+	TabWidth = 8
+	Padding  = 0
+	PadChar  = '\t'
+	flags    = 0
+)
+
+func DisplayPackages(packages []api.BundlePackage) {
+	m := packagesToString(packages)
+
+	// initialize tabwriter
```
chrisdoherty4, line 23:
```suggestion
```

pkg/curatedpackages/package.go
chrisdoherty4: How extensively has this been tested? Trying to create alignment can be a tricky problem particularly when you have variable length strings. I.e. are we confident `pkg.Name` is always going to be short enough to fit under `Package\t`? There are packages dedicated to this kind of formatting problem.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+		packageBundle, err = getActiveBundleFromCluster(ctx, kubeConfig)
+	case Registry:
+		packageBundle, err = getLatestBundleFromRegistry(ctx, kubeVersion)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return packageBundle, nil
+}
```
chrisdoherty4, line 42:
```suggestion
func GetLatestBundle(ctx context.Context, kubeConfig string, source string, kubeVersion string) (*api.PackageBundle, error) {
	switch strings.ToLower(source) {
	case Cluster:
		return getActiveBundleFromCluster(ctx, kubeConfig)
	case Registry:
		return getLatestBundleFromRegistry(ctx, kubeVersion)
	default:
		return nil, errors.New("unknown source")
	}
}
```

Try not to over complicate things.

pkg/curatedpackages/common.go
chrisdoherty4: This is also only used in `bundle.go` so move it there and drop this source file.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+
+func createBundleManager(kubeVersion string) (manager bundle.Manager) {
+	versionSplit := strings.Split(kubeVersion, ".")
+	major, minor := versionSplit[0], versionSplit[1]
+	log := logr.Discard()
```
chrisdoherty4, line 47: Whats the rationale for the noop logger?

pkg/curatedpackages/bundle.go
chrisdoherty4: Looking at the function further `new` isn't necessarily appropriate here. Its not wrong, but neither is `create`. `new` is generally used in initializing constructors and this isn't strictly an initializing constructor. It _uses_ an initializing constructor, `NewBundleManager`, but doesn't perform the initialization itself.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+	return activeController, nil
+}
+
+func getActiveBundleController(bc *api.PackageBundleControllerList) (*api.PackageBundleController, error) {
+	for _, v := range bc.Items {
+		if v.Status.State == api.BundleControllerStateActive {
+			return &v, nil
+		}
+	}
+	return nil, fmt.Errorf("no Active Bundle Controller Found")
```
chrisdoherty4, line 116:
```suggestion
	return nil, errors.New("no active bundle controller found")
```

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	kubectl := deps.Kubectl
+
+	// Active Bundle is set at the bundle Controller
+	bundleController, err := getActiveController(ctx, kubectl, params...)
+	if err != nil {
+		return nil, err
+	}
+	activeBundle := bundleController.Spec.ActiveBundle
+	params = append(params, executables.WithArg(activeBundle))
```
chrisdoherty4, line 74:
```suggestion
	params = append(params, executables.WithArg(bundleController.Spec.ActiveBundle))
```

pkg/curatedpackages/bundle.go
a-cool-train: There is a need for a logger in the implementation of the NewBundleManager and since I don't need the logger for my specification, there was the need for the noop logger.

pkg/curatedpackages/package.go
a-cool-train: By any chance, could you recommend the package that solves this kind of formatting problem?

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -51,7 +51,8 @@ func getResources(ctx context.Context, resourceType string, output string, args
 	if output != "" {
 		params = append(params, executables.WithOutput(output))
 	}
-	packages, err := kubectl.GetResources(ctx, resourceType, params...)
+	stdOut, err := kubectl.GetResources(ctx, resourceType, params...)
+	packages := stdOut.String()
 	if err != nil {
 		fmt.Print(packages)
```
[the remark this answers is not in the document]
a-cool-train: It is actually not a debug item. In case some packages are successful, and other packages fail, kubectl prints the successful packages then prints the packages that failed as an error.

cmd/eksctl-anywhere/cmd/listpackages.go
a-cool-train: @ewollesen, I am not sure I understand this comment. I am currently using StringVar to handle the validation. Are you suggesting to use Var instead of StringVar for the validation?

pkg/curatedpackages/bundle.go
a-cool-train: Done.

cmd/eksctl-anywhere/cmd/validations.go
```diff
@@ -4,12 +4,18 @@ import (
...
 	"fmt"
 	"runtime"
+	"strings"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
+const (
```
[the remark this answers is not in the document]
a-cool-train: Just to make sure I understand, is the idea to use a consistent type in both validations file and bundle.go?

pkg/curatedpackages/bundle.go
chrisdoherty4: Generally it should be the main func or some derivative of the main func that stipulates what logger to use. In constructing the logger in-line like this we deprive ourselves of the log output from the bundle manager when we really need it (aka there's a problem). 

I think you'll want to pass the logger as an argument and use it in `NewBundleManaer()`.

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,38 @@
...
+	"text/tabwriter"
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+const (
+	MinWidth = 16
+	TabWidth = 8
+	Padding  = 0
+	PadChar  = '\t'
```
chrisdoherty4, line 16:
```suggestion
	minWidth = 16
	tabWidth = 8
	padding  = 0
	padChar  = '\t'
```

pkg/curatedpackages/package.go
chrisdoherty4: I don't recall what I've used before but a quick search shows https://github.com/buger/goterm which seems well supported.

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -51,7 +51,8 @@ func getResources(ctx context.Context, resourceType string, output string, args
 	if output != "" {
 		params = append(params, executables.WithOutput(output))
 	}
-	packages, err := kubectl.GetResources(ctx, resourceType, params...)
```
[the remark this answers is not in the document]
a-cool-train: Quick context: Initial implementation of GetResources was only needed for a single purpose. In order to refactor and reuse GetResources, it made sense to return the output of GetResources directly so that the user of the method is able to decide what to do with the result.

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func getPackageBundle(ctx context.Context, kubectl *executables.Kubectl, kubeConfig string, activeBundle string) (*api.PackageBundle, error) {
+	params := []executables.KubectlOpt{executables.WithOutput("json"), executables.WithKubeconfig(kubeConfig), executables.WithNamespace(constants.EksaPackagesName), executables.WithArg(activeBundle)}
+	stdOut, err := kubectl.GetResources(ctx, "packageBundle", params...)
+	if err != nil {
+		return nil, err
+	}
+	obj := &api.PackageBundle{}
+	if err = json.NewDecoder(&stdOut).Decode(obj); err != nil {
```
chrisdoherty4, line 84:
```suggestion
	if err := json.Unmarshal(stdOut.Bytes(), obj); err != nil {
```

a-cool-train: @chrisdoherty4 any reason for using json.Unmarshal vs json.NewDecoder? @ewollesen suggested using the NewDecoder but I was curious whether one was preferred over the other?

pkg/curatedpackages/bundle.go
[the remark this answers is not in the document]
ewollesen: Chris is totally right, if this were to stay a string variable, then that would be the case. If source becomes a user type, as I think it should, then I'd still panic because that validation of user input should have already happened, and there's really no reason there should ever be an unexpected value here, and if there is, something is very wrong. I'm a fan of fail early and loudly. Some folks aren't. 

I'm 100% fine with returning an error though. So long as it's identified and not swallowed, I'm happy.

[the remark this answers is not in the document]
ewollesen: Hmm... If you move to a user type, then you could get rid of this entirely...

(Pseudo-code follows)

```go
type bundleGetter interface {
    getLatestBundle(ctx context.Context, kubeConfig, kubeVersion string) (*Bundle, err)
}
func (c Cluster) getLatestBundle(ctx context.Context, kubeConfig, kubeVersion string) (*Bundle, err) {
    // implement me
}
func (r Registry) getLatestBundle(ctx context.Context, kubeConfig, kubeVersion string) (*Bundle, err) {
    // implement me
}
```

With the above, then this switch goes away and becomes:

```go
bundle, err := source.getLatestBundle(ctx, kubeConfig, kubeVersion)
```

https://lostechies.com/derickbailey/2008/11/26/ptom-descriptive-state-enumeration/

chrisdoherty4 (approved):

cmd/eksctl-anywhere/cmd/validations.go
```diff
@@ -33,3 +35,14 @@ func commonValidation(ctx context.Context, clusterConfigFile string) (*v1alpha1.
 	}
 	return clusterConfig, nil
 }
+
+func validateKubeVersion(kubeVersion string, bs curatedpackages.BundleSource) error {
+	if bs.String() != curatedpackages.Registry {
+		return nil
+	}
```
ewollesen, line 42:
```suggestion
	if bs != curatedpackages.Registry {
		return nil
	}
```

ewollesen (approved):
/lgtm

PR #1624, opened 2022-03-24
Package-Controller Helm Chart to Release Bundle 

jonahjon:
Adds the Helm chart to dev account release bundle.


release/pkg/assets_package_controller.go
```diff
@@ -16,20 +16,22 @@ package pkg
...
 	"github.com/pkg/errors"
 
 	anywherev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
 const (
-	packagesRootPath = "projects/aws/eks-anywhere-packages"
+	packagesRootPath  = "projects/aws/eks-anywhere-packages"
+	packagesImageName = "eks-anywhere-packages"
+	packagesHelmChart = "eks-anywhere-packages-helm"
```
TerryHowe, line 29: Is `eks-anywhere-packages-helm` the repository for the helm chart? I thought we were going to do that, but not sure if that was done yet

release/pkg/assets_package_controller.go
```diff
@@ -64,27 +66,49 @@ func (r *ReleaseConfig) GetPackagesAssets() ([]Artifact, error) {
...
+	// Remove the OS, and Arch for the Helm chart, and modify the tag based off build-tooling helm/push.sh script.
+	// Currently we must Trim leading v of the tag after the :, and add "-helm to the end of tag.
+	// 123456.dkr.ecr.us-west-2.amazonaws.com/eks-anywhere-packages:v0.1.2-2e9994fd1afb2216a51fa474ac5d7dc6a772bb62
+	// >>>
+	// 123456.dkr.ecr.us-west-2.amazonaws.com/eks-anywhere-packages:0.1.2-2e9994fd1afb2216a51fa474ac5d7dc6a772bb62-helm
 	artifacts := []Artifact{{Image: imageArtifact}}
+	helmImageArtifact := &ImageArtifact{
+		AssetName:         packagesHelmChart, // This needs to differ from the image name for later steps
+		SourceImageURI:    sourceImageUri,
+		ReleaseImageURI:   strings.ReplaceAll(fmt.Sprintf("%s-helm", releaseImageUri), "packages:v", "packages:"),
```
TerryHowe, line 78: Wouldn't we use the shasum for the tag?

jonahjon: This is looking up the dev private ECR tag. I looked in the account, and the image tags are getting `v${GIT_TAG}-{GIT_HASH)`  and the helm had `${GIT_TAG}-{GIT_HASH)-helm` we just needed to trim the image tag based off the `helm/push.sh` script tagging.

release/pkg/assets_package_controller.go
jonahjon: Made some changes, good catch. LMK what you think about this format

abhay-krishna: honestly would be nice if we had a repo like `eks-anywhere-packages/charts` and use the same versioning.

jonahjon: > honestly would be nice if we had a repo like `eks-anywhere-packages/charts` and use the same versioning.

For the immediate term (bugbash) being this is probably not going to happen, but I'll bring it up during standup to discuss the pros/cons on this one.

release/pkg/assets_package_controller.go
```diff
@@ -64,27 +66,49 @@ func (r *ReleaseConfig) GetPackagesAssets() ([]Artifact, error) {
...
+	// Remove the OS, and Arch for the Helm chart, and modify the tag based off build-tooling helm/push.sh script.
+	// Currently we must Trim leading v of the tag after the :, and add "-helm to the end of tag for source URI.
+	// 123456.dkr.ecr.us-west-2.amazonaws.com/eks-anywhere-packages:v0.1.2-2e9994fd1afb2216a51fa474ac5d7dc6a772bb62
+	// >>>
+	// 123456.dkr.ecr.us-west-2.amazonaws.com/eks-anywhere-packages:0.1.2-2e9994fd1afb2216a51fa474ac5d7dc6a772bb62-helm
 	artifacts := []Artifact{{Image: imageArtifact}}
+	helmImageArtifact := &ImageArtifact{
+		AssetName:         packagesHelmChart, // This needs to differ from the image name for later steps
+		SourceImageURI:    strings.ReplaceAll(fmt.Sprintf("%s-helm", sourceImageUri), "packages:v", "packages:"),
+		ReleaseImageURI:   strings.ReplaceAll(releaseImageUri, "packages:v", "packages:"),
```
abhay-krishna, line 78: The release URI won't have the helm suffix, instead they will be the exact same as for the package controller image, minus the `v` at the start of the tag?

jonahjon: Yea I think that's what we're going for. With helm 3.8.1 we need to need the preceding  `v`, and the trailing `-helm`

PR #1632, opened 2022-03-25
Add Tinkerbell to new cluster parsing code

chrisdoherty4:
The Tinkerbell provider isn't represented in the cluster parsing code
resulting in warnings that could mislead the customer in logging. This
cl adds Tinkerbell type so the parsing logic is at least aware of them
but won't do anything with them.


abhinavmpandey08 (approved):
/lgtm

PR #1608, opened 2022-03-24
Add registry mirror feature to snow

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1574, opened 2022-03-22
Snow cluster features support


jiayiwang7:
- [x] Kubernetes 1.21
- [x] Kubernetes 1.22
- [x] Kubernetes 1.23
- [x] Cilium as CNI
- [x] Registry Mirror
- [x] Multiple devices setting
- [x] Management and Workload clusters
- [x] Multiple worker node groups
- [x] AWS IAM Authenticator
- [x] OIDC
- [x] Pod IAM (IRSA) (code ready, e2e not tested)
- [x] Taints (code ready, e2e not tested)
- [x] Labels (code ready, e2e not tested)
- [x] Proxy configuration
- [x] GitOps (controller)
- [x] #3283

PR #1637, opened 2022-03-25
Add insecure skip verify option for registry mirror

jiayiwang7:
*Issue #, if available:*

Parent task #1574 

Follow up PR for #1608 

*Description of changes:*

`InsecureSkipVerify` skips the registry certificate verification. Only use this solution for isolated testing or in a tightly controlled, air-gapped environment.

Currently only supported for Snow provider. Setting this field to `true` in other providers will fail the preflight validations.

Reason to include this option in Snow is that passing the TLS certificate for local registry verification exceeds the EC2 user data limit (16K). It's also a valid use case to provision and manage Snow cluster in air-gapped environment.

API change: add `InsecureSkipVerify` field in `Cluster.Spec.RegistryMirrorConfiguration`. e.g.

```yaml
kind: Cluster
metadata:
  name: test-snow
spec:
  ...
  registryMirrorConfiguration:
    endpoint: 10.111.62.105
    insecureSkipVerify: true
    ...
```

*Testing (if applicable):*

- unit tests
- local e2e on Snow devices

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -15,4 +15,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.42.1
+          version: latest
--- a/config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
+++ b/config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
@@ -81,8 +81,12 @@ spec:
                         situation, only 'Shared' network is supported.
                       properties:
                         id:
+                          description: Id of a resource in the CloudStack environment.
+                            Mutually exclusive with Name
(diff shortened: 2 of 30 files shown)


g-gaston (commented):
Is the idea to support this for all providers eventually or to remove it once we are able to pass the cert for snow?

I'm fine either way, but I would suggest creating a ticket for whatever the solution is, just so we don't forget. This change introduces a gap in functionality between providers, which is no ideal, so it would be great if we could track it.

jiayiwang7:
> Is the idea to support this for all providers eventually or to remove it once we are able to pass the cert for snow?
> 
> I'm fine either way, but I would suggest creating a ticket for whatever the solution is, just so we don't forget. This change introduces a gap in functionality between providers, which is no ideal, so it would be great if we could track it.

make sense, https://github.com/aws/eks-anywhere/issues/1647

g-gaston (approved):
/lgtm

PR #1638, opened 2022-03-26
Post cluster delete powerOff wait

pokearu:
*Description of changes:*
The changes involve adding to the delete cluster workflow. This resolves the issue of kind cluster getting deleted before CAPT can initiate powerOff call to the nodes.

1. Added logic to fetch hardware corresponding to a tinkerbell cluster during delete cluster `SetupAndValidateDeleteCluster` step. 
2. Added validation to ensure we have at least one hardware CRD mapping to the cluster we are trying to delete.
3. Added a `PostClusterDeleteValidate` call to the delete cluster workflow.
4. On the tinkerbell provider, we get the status of the BMC and retry till the status.powerState is the desired value.

*Testing:*
Tested by running create and delete cluster from eksctl-anywhere.

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -1412,9 +1412,8 @@ func (k *Kubectl) GetResources(ctx context.Context, resourceType string, opts ..
 	return stdOut.String(), err
 }
 
-// GetHardwareForCluster gets the hardwares with ownerName label.
-// Then iterates to filter the hardwares that correspond to the given clusterName.
-func (k *Kubectl) GetHardwareForCluster(ctx context.Context, clusterName, kubeconfigFile, namespace string) ([]tinkv1alpha1.Hardware, error) {
+// GetHardwareWithOwnerName gets the hardwares with ownerName label.
+func (k *Kubectl) GetHardwareWithOwnerName(ctx context.Context, kubeconfigFile, namespace string) ([]tinkv1alpha1.Hardware, error) {
 	params := []string{
 		"get", captHardwareResourceType, "-o", "json", "--kubeconfig",
 		kubeconfigFile, "--namespace", namespace, "--selector=v1alpha1.tinkerbell.org/ownerName",
@@ -1430,34 +1429,19 @@ func (k *Kubectl) GetHardwareForCluster(ctx context.Context, clusterName, kubeco
 		return nil, fmt.Errorf("parsing get hardware response: %v", err)
 	}
 
-	// Filtering the hardware with ownerName mapping to cluster name.
-	var filteredHardwareList []tinkv1alpha1.Hardware
-	for _, hw := range response.Items {
-		if strings.Contains(hw.Labels["v1alpha1.tinkerbell.org/ownerName"], clusterName) {
-			filteredHardwareList = append(filteredHardwareList, hw)
-		}
-	}
-
-	return filteredHardwareList, nil
+	return response.Items, nil
 }
 
-// ValidateBmcsPowerState validates if the given BMC CRs all have the desired powerState.
-func (k *Kubectl) ValidateBmcsPowerState(ctx context.Context, bmcNames []string, powerState, kubeconfigFile, namespace string) error {
+// GetBmcsPowerState gets the .status.powerState for the Bmcs.
+func (k *Kubectl) GetBmcsPowerState(ctx context.Context, bmcNames []string, kubeconfigFile, namespace string) ([]string, error) {
 	params := []string{"get", captBmcResourceType}
 	params = append(params, bmcNames...)
 	params = append(params, "-o", "jsonpath={.items[*].status.powerState}", "--kubeconfig", kubeconfigFile, "-n", namespace)
 
 	buffer, err := k.Execute(ctx, params...)
 	if err != nil {
-		return fmt.Errorf("executing get: %v", err)
+		return nil, fmt.Errorf("executing get: %v", err)
 	}
 
-	powerStates := strings.Fields(buffer.String())
-	for _, state := range powerStates {
-		if !strings.Contains(state, powerState) {
-			return fmt.Errorf("bmc current power state: %s expected power state: %s", state, powerState)
-		}
-	}
-
-	return nil
+	return strings.Fields(buffer.String()), nil
 }
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -12,6 +12,8 @@ import (
 	cloudstackv1 "github.com/aws/cluster-api-provider-cloudstack/api/v1beta1"
 	eksdv1alpha1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
 	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
+	pbnjv1alpha1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/pbnj/api/v1alpha1"
+	tinkv1alpha1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/tink/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/version"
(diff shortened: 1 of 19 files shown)


pkg/executables/kubectl.go
```diff
@@ -1404,3 +1408,53 @@ func (k *Kubectl) GetResources(ctx context.Context, resourceType string, opts ..
...
+		if strings.Contains(hw.Labels["v1alpha1.tinkerbell.org/ownerName"], clusterName) {
+			filteredHardwareList = append(filteredHardwareList, hw)
+		}
+	}
+
+	return filteredHardwareList, nil
+}
+
+// ValidateBmcsPowerState validates if the given BMC CRs all have the desired powerState.
+func (k *Kubectl) ValidateBmcsPowerState(ctx context.Context, bmcNames []string, powerState, kubeconfigFile, namespace string) error {
```
chrisdoherty4, line 1442: This is business logic - it shouldn't exist in the `Kubctl` executable. Why do this here instead of on a dedicated post delete validation construct within the tinkerbell provider package?

pkg/executables/kubectl.go
```diff
@@ -1404,3 +1408,53 @@ func (k *Kubectl) GetResources(ctx context.Context, resourceType string, opts ..
 	stdOut, err := k.Execute(ctx, params...)
 	return stdOut.String(), err
 }
+
+// GetHardwareForCluster gets the hardwares with ownerName label.
+// Then iterates to filter the hardwares that correspond to the given clusterName.
+func (k *Kubectl) GetHardwareForCluster(ctx context.Context, clusterName, kubeconfigFile, namespace string) ([]tinkv1alpha1.Hardware, error) {
```
chrisdoherty4, line 1414: This is also business logic. I would expect this method to exist on some other object within the tinkerbell provider and to simply call out to `Kubectl` to retrieve the Kubernetes objects.

pkg/providers/tinkerbell/validator.go
```diff
@@ -227,6 +228,15 @@ func (v *Validator) ValidateMinimumRequiredTinkerbellHardwareAvailable(spec v1al
 	return nil
 }
 
+// ValidateHardwaresAreFound ensures that there are one or more hardware CRDs presnt in the hardware list for a cluster.
+func (v *Validator) ValidateHardwaresAreFound(hardwares []tinkv1alpha1.Hardware) error {
```
chrisdoherty4, line 232: Why does a consumer need to call this vs `if len(hardwares.Hardware) != 0`?

pkg/executables/kubectl.go
pokearu: Sure! I can remove the logic and make this method solely `kubectl get ...`.
And add a validation method that performs the checks.

pkg/executables/kubectl.go
pokearu: Sure. I can make this method into just `GetHardwareWithOwnerLabel` and add the filtering logic into a local method in the tinkerbell provider package.

pkg/providers/tinkerbell/validator.go
pokearu: I initially just had that `if` inside `SetupAndValidateDeleteCluster`. But then thought to maybe follow the practice of having a separate validation method inside `validator`. 
I do agree its a bit much, I could change it back and later if we have more things to validate we can move it out 👍

pokearu: Another reason I thought about was for padding the `error` message. 
`failed setup and validations: no hardware found for cluster`

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -410,14 +448,40 @@ func (p *tinkerbellProvider) setMachinesToPXEBoot(ctx context.Context) error {
...
+	p.hardwares = filteredHws
+
 	return nil
 }
 
+// filterHardwareForCluster filters hardware with ownerName label that contains cluster name.
+func filterHardwareForCluster(hardwares []tinkv1alpha1.Hardware, clusterName string) ([]tinkv1alpha1.Hardware, error) {
+	var filteredHardwareList []tinkv1alpha1.Hardware
+	for _, hw := range hardwares {
+		if strings.Contains(hw.Labels["v1alpha1.tinkerbell.org/ownerName"], clusterName) {
```
ptrivedi, line 474: would be nice to have this label name as a const

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -248,6 +255,37 @@ func (p *tinkerbellProvider) DeleteResources(ctx context.Context, clusterSpec *c
...
+	// In the current version of kubectl in EKSA --for does not support jsonpath.
+	err := retrier.Retry(10, 10*time.Second, func() error {
+		powerStates, err := p.providerKubectlClient.GetBmcsPowerState(ctx, bmcRefs, managementCluster.KubeconfigFile, constants.EksaSystemNamespace)
+		if err != nil {
+			return err
+		}
+
+		for _, state := range powerStates {
+			if !strings.Contains(state, bmcStatePowerActionHardoff) {
+				return fmt.Errorf("bmc current power state: %s expected power state: %s", state, bmcStatePowerActionHardoff)
```
chrisdoherty4, line 276: The context separation here (addition of `:`) is misleading, I think you want it to read as a single sentence. To do that, use `;`.

```suggestion
				return fmt.Errorf("bmc current power state %s; expected power state %s", state, bmcStatePowerActionHardoff)
```

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -248,6 +255,37 @@ func (p *tinkerbellProvider) DeleteResources(ctx context.Context, clusterSpec *c
...
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("cluster nodes not in power off state: %v", err)
```
chrisdoherty4, line 283: This doesn't provide any additional context given the original error already stipulates the power state is incorrect.

```suggestion
		return err
```

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -248,6 +255,37 @@ func (p *tinkerbellProvider) DeleteResources(ctx context.Context, clusterSpec *c
...
+	for _, hw := range p.hardwares {
+		bmcRefs = append(bmcRefs, hw.Spec.BmcRef)
+	}
+
+	// TODO (pokearu): The retry logic can be substituted by changing GetBmcsPowerState to use kubectl wait --for
+	// In the current version of kubectl in EKSA --for does not support jsonpath.
+	err := retrier.Retry(10, 10*time.Second, func() error {
+		powerStates, err := p.providerKubectlClient.GetBmcsPowerState(ctx, bmcRefs, managementCluster.KubeconfigFile, constants.EksaSystemNamespace)
+		if err != nil {
+			return err
```
chrisdoherty4, line 271: We can indicate the action we took that caused the error.

```suggestion
			return fmt.Errorf("retrieving bmc power state: %w", err)
```

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -86,6 +91,8 @@ type ProviderKubectlClient interface {
 	DeleteEksaDatacenterConfig(ctx context.Context, eksaTinkerbellDatacenterResourceType string, tinkerbellDatacenterConfigName string, kubeconfigFile string, namespace string) error
 	DeleteEksaMachineConfig(ctx context.Context, eksaTinkerbellMachineResourceType string, tinkerbellMachineConfigName string, kubeconfigFile string, namespace string) error
 	GetMachineDeployment(ctx context.Context, machineDeploymentName string, opts ...executables.KubectlOpt) (*clusterv1.MachineDeployment, error)
+	GetHardwareWithOwnerName(ctx context.Context, kubeconfigFile, namespace string) ([]tinkv1alpha1.Hardware, error)
```
chrisdoherty4, line 94: What is the `WithOwnerName` referencing? The method signature has nothing that would suggest there's anything special beyond retrieving the `tinkv1alpha1.Hardware` from the cluster.

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: Its used in 2 places so it really should be a constant somewhere.

chrisdoherty4 (approved):

PR #1640, opened 2022-03-28
test cluster name wont start with -

danbudris:
*Issue #, if available:*

*Description of changes:*
We ran into an issue where when the BRANCH_NAME env var was not set in a pipeline the test cluster names generated started with `-`, which cobra interpreted as a flag (e.g. `-i-asdflksj`) and caused test runs to fail. If the BRANCH_NAME passed to the cluster name func is "", we now just go ahead and use the instance id.

*Testing (if applicable):*


mitalipaygude (approved):

abhay-krishna (approved):
/lgtm

PR #1661, opened 2022-03-28
Pass force flag to Tinkerbell provider

chrisdoherty4:
We want the force flag to create behavior that ensures machines are
usable when creating a new cluster even if they appear to have been
provisioned previously.

We can't model the behavior using dynamic construction logic so we're
passing the force flag all the way into the Tinkerbell provider where
we'll implement the behavior.


abhinavmpandey08 (approved):
/lgtm

PR #1665, opened 2022-03-28
Remove kube-rbac-proxy from EKS-A codebase

abhay-krishna:
CAPI removed `kube-rbac-proxy` from its `Deployment`s during the `v1alpha3` to `v1alpha4` migration, and the various infrastructure providers also followed suit. This PR removes kube-rbac-proxy from the release bundle and also references in the code for `clusterctl` overrides.


release/api/v1alpha1/bundle_types.go
```diff
@@ -147,31 +147,27 @@ type CertManagerBundle struct {
 type CoreClusterAPI struct {
 	Version    string   `json:"version"`
 	Controller Image    `json:"controller"`
-	KubeProxy  Image    `json:"kubeProxy"`
```
g-gaston, line 150: This is a breaking change. I would just leave all the fields as they are, add a `ommitempty` tag and mark as deprecated with a doc comment

abhay-krishna: Makes sense. Can we just retain the API struct for backward compatibility but remove its references from everywhere else in the codebase?

PR #1667, opened 2022-03-28
Tinkerbell implement force create cluster

chrisdoherty4:
Builds on #1661

Implement the force node reset to PXE behavior desired when the --force-cleanup flag is specified. Node should power off, be set to PXE boot and power on again where they'll sit waiting for the Tinkerbell stack to respond.

Manually tested by powering machines on and observed the logging to power off during create cluster cmd.

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -10,6 +10,8 @@ import (
 
 	tinkhardware "github.com/tinkerbell/tink/protos/hardware"
 	tinkworkflow "github.com/tinkerbell/tink/protos/workflow"
+	corev1 "k8s.io/api/core/v1"
+	errorutil "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -353,6 +367,47 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context,
...
+			errs = append(errs, err)
+		}
+
+		if err := p.pbnj.SetBootDevice(ctx, conf, pbnj.BootDevicePXE); err != nil {
+			errs = append(errs, err)
+		}
+
+		if err := p.pbnj.PowerOn(ctx, conf); err != nil {
+			errs = append(errs, err)
+		}
```
abhinavmpandey08, line 403: why are we powering the machines on here? 
CAPT's pbnj should take care of this and that also won't power on all the machines, only the ones that are needed for the cluster

chrisdoherty4: Yea I realized that. Will remove once I can actually get a test functional.

PR #1673, opened 2022-03-29
Bring back auth for helm with import-images

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
We reverted the commit in #1609 after doing some testing that didn't require us to be logged in to the registry, but when testing in our ci environment it seemed like it didn't work as expected. The creds might have been cached or the set up for dev is different. Now, we will be asking users to specify the username and password when running import images so helm that is running inside our docker container will be allowed to push. We could have used the `--registry-config` if we had access to the `~/.docker/config.json` to avoid the user from having to login to the registry mirror for docker and also provide these env vars for helm, but I am not sure if there is a quick way for us to do that apart from mounting the `config.json` file after checking for the existence of it in the default path. Ideally if there was a way to get these credential information from the docker service that was accessible from within the docker container, then we can avoid having to provide these env vars.

This change would need to be cherry-picked to the `release-0.8` branch as well

/cherry-pick release-0.8

*Testing (if applicable):*
Tested e2e against ci env

### DIFF
--- a/cmd/eksctl-anywhere/cmd/importimages.go
+++ b/cmd/eksctl-anywhere/cmd/importimages.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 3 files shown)


danbudris (approved):

mrajashree (approved):
/lgtm

Issue #1669, opened 2022-03-29
Move provider specific logic out of cluster manager and into provider interface


maxdrib:
**What would you like to be added**:
As discussed in https://github.com/aws/eks-anywhere/pull/1675, VSphere and CloudStack have introduced provider-specific logic into the EksaClusterComponentsChanged method. This logic should instead be outsourced to the provider class as a new method on the provider interface to keep the cluster manager generic, and improve testability

**Why is this needed**:

maxdrib:
Successfully refactored the cluster manager to remove provider specific logic

PR #1675, opened 2022-03-29
Final CloudStack Provider to support Upgrade

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR continues the merge from the cloudstack branch and includes all the necessary changes under pkg/ to support the cloudstack upgrade workflow.

*Testing (if applicable):*
Unit tests, testing in client environment on cloudstack branch. A separate PR is in the works covering the corresponding e2e tests which have been run intermittently against this code


maxdrib:
Currently seeing the following when a workload cluster is being created after `Applying eksa yaml resources to cluster`:

```
2022-03-29T10:45:19.908-0400	V9	docker	{"stderr": "Error from server (BadRequest): error when creating \"STDIN\": admission webhook \"validation.cloudstackdatacenterconfig.anywhere.amazonaws.com\" denied the request: CloudStackProvider feature is not active, preventing CloudStackDataCenterConfig resource creation\nError from server (BadRequest): error when creating \"STDIN\": admission webhook \"validation.cloudstackmachineconfig.anywhere.amazonaws.com\" denied the request: CloudStackProvider feature is not active, preventing CloudStackMachineConfig resource creation\nError from server (BadRequest): error when creating \"STDIN\": admission webhook \"validation.cloudstackmachineconfig.anywhere.amazonaws.com\" denied the request: CloudStackProvider feature is not active, preventing CloudStackMachineConfig resource creation\nError from server (BadRequest): error when creating \"STDIN\": admission webhook \"validation.cloudstackmachineconfig.anywhere.amazonaws.com\" denied the request: CloudStackProvider feature is not active, preventing CloudStackMachineConfig resource creation\n"}
```

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -717,15 +801,138 @@ func (p *cloudstackProvider) generateCAPISpecForCreate(ctx context.Context, clus
...
+			if err != nil {
+				return nil, nil, err
+			}
+			workloadTemplateName = md.Spec.Template.Spec.InfrastructureRef.Name
+			workloadTemplateNames[workerNodeGroupConfiguration.Name] = workloadTemplateName
+		} else {
+			workloadTemplateName = common.WorkerMachineTemplateName(clusterName, workerNodeGroupConfiguration.Name, p.templateBuilder.now)
+			workloadTemplateNames[workerNodeGroupConfiguration.Name] = workloadTemplateName
+		}
+		p.templateBuilder.WorkerNodeGroupMachineSpecs[workerNodeGroupConfiguration.MachineGroupRef.Name] = p.machineConfigs[workerNodeGroupConfiguration.MachineGroupRef.Name].Spec
```
g-gaston, line 880: Why are we modifying the template builder? Doesn't it receive all the info it needs when we call `GenerateCAPISpecControlPlane` and `GenerateCAPISpecWorkers`

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -845,6 +1048,75 @@ func (p *cloudstackProvider) GenerateStorageClass() []byte {
...
+	cpMachineConfigName := clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name
+	if prevSpec.Spec.ControlPlaneConfiguration.MachineGroupRef.Name != cpMachineConfigName {
+		em, err := p.providerKubectlClient.SearchCloudStackMachineConfig(ctx, cpMachineConfigName, cluster.KubeconfigFile, clusterSpec.Cluster.GetNamespace())
+		if err != nil {
+			return err
+		}
+		if len(em) > 0 {
+			return fmt.Errorf("control plane VSphereMachineConfig %s already exists", cpMachineConfigName)
+		}
+	}
```
g-gaston, line 1102: Can we dry out this method? These two blocks of code seem to be almost the same

pkg/providers/cloudstack/config/secret.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: eksa-system
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: eksa-license
```
g-gaston, line 9: I'm missing something here. Why does the eksa license need to be in the cloudstack specific code?

controllers/controllers/resource/testdata/etcdadmcluster.yaml
```diff
@@ -1,28 +0,0 @@
-kind: EtcdadmCluster
```
vivek-koppuru, line 1: A lot of test files are getting deleted, is that intended?

pkg/clustermanager/cluster_manager.go
```diff
@@ -523,6 +525,52 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
...
+		existingCpCsmc, err := c.clusterClient.GetEksaCloudStackMachineConfig(ctx, cc.Spec.ControlPlaneConfiguration.MachineGroupRef.Name, cluster.KubeconfigFile, newClusterSpec.Cluster.Namespace)
+		if err != nil {
+			return false, err
+		}
+		cpCsmc := machineConfigMap[newClusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name]
+		if !reflect.DeepEqual(existingCpCsmc.Spec, cpCsmc.Spec) {
+			logger.V(3).Info("New control plane machine config spec is different from the existing spec")
+			return true, nil
+		}
+		existingWnCsmc, err := c.clusterClient.GetEksaCloudStackMachineConfig(ctx, cc.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef.Name, cluster.KubeconfigFile, newClusterSpec.Cluster.Namespace)
```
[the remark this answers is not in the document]
vivek-koppuru: I think we have already added validations for multiple worker node groups so I think this is still covered by a follow up issue?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -441,10 +460,71 @@ func (p *cloudstackProvider) SetupAndValidateDeleteCluster(ctx context.Context)
...
+	if oldSpec.Cluster.Spec.KubernetesVersion != newSpec.Cluster.Spec.KubernetesVersion {
+		return true
+	}
+	if oldSpec.Bundles.Spec.Number != newSpec.Bundles.Spec.Number {
+		return true
+	}
+	if !v1alpha1.WorkerNodeGroupConfigurationSliceTaintsEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
+		!v1alpha1.WorkerNodeGroupConfigurationsLabelsMapEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) {
+		return true
+	}
```
vivek-koppuru, line 489: This logic looks exactly the same as vsphere and looks like it can be shared.

release/pkg/assets_capc.go
```diff
@@ -130,7 +130,6 @@ func (r *ReleaseConfig) GetCapcAssets() ([]Artifact, error) {
 func (r *ReleaseConfig) GetCloudStackBundle(imageDigests map[string]string) (anywherev1alpha1.CloudStackBundle, error) {
 	cloudstackBundleArtifacts := map[string][]Artifact{
 		"cluster-api-provider-cloudstack": r.BundleArtifactsTable["cluster-api-provider-cloudstack"],
-		"kube-rbac-proxy":                 r.BundleArtifactsTable["kube-rbac-proxy"],
```
vivek-koppuru, line 133: We have a follow up item regarding the kube rbac proxy that will add this back in?

controllers/controllers/resource/testdata/etcdadmcluster.yaml
maxdrib: I removed these files with safe delete in GoLand. Unit tests are still passing. I also did a search for the filenames in our code base and nothing turned up. I thought it would be nice to clean up some unused test data files

pkg/executables/kubectl.go
```diff
@@ -682,10 +682,13 @@ func (k *Kubectl) ValidateEKSAClustersCRD(ctx context.Context, cluster *types.Cl
 }
 
 func (k *Kubectl) SetControllerEnvVar(ctx context.Context, envVar, envVarVal, kubeconfig string) error {
```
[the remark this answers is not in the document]
maxdrib: Renamed to SetEksaControllerEnvVar for consistency and clarity

pkg/providers/cloudstack/cloudstack.go
maxdrib: Would you suggest putting the first 3 if statements into the common package? As I recall, Chris had some objections to the dumping ground

release/pkg/assets_capc.go
maxdrib: I don't think we have any need for kube-rbac-proxy inside the cloudstack bundle since we've moved to v1beta1. We do plan to add kube-vip however

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -441,10 +460,71 @@ func (p *cloudstackProvider) SetupAndValidateDeleteCluster(ctx context.Context)
 	return nil
 }
 
+func NeedsNewControlPlaneTemplate(oldSpec, newSpec *cluster.Spec, oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig) bool {
```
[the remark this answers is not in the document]
maxdrib: Yes, it looks like some unit tests were missed, good catch. I'll see what's available to be reused from the vsphere provider code

pkg/clustermanager/cluster_manager.go
```diff
@@ -523,6 +525,52 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
...
 		}
+	case v1alpha1.CloudStackDatacenterKind:
+		machineConfigMap := make(map[string]*v1alpha1.CloudStackMachineConfig)
+
+		existingCsdc, err := c.clusterClient.GetEksaCloudStackDatacenterConfig(ctx, cc.Spec.DatacenterRef.Name, cluster.KubeconfigFile, newClusterSpec.Cluster.Namespace)
+		if err != nil {
+			return false, err
+		}
+		csDc := datacenterConfig.(*v1alpha1.CloudStackDatacenterConfig)
+		if !reflect.DeepEqual(existingCsdc.Spec, csDc.Spec) {
```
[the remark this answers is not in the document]
maxdrib: I have seen this requested elsewhere, so I'll have a look at what that might entail. Although I don't fully understand the caution of using deepEqual, it seems easier to maintain while still including all the functionality. Could you explain your concern? Is there a possibility of some transient metadata in the objects making them appear to be non-equal, when functionally they are the same?

pkg/providers/cloudstack/cloudstack.go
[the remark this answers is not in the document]
maxdrib: It looks like vsphere does not have unit tests for these functions directly, only testing the caller which is GenerateCAPISpecForUpgrade. I'm actually not sure why these functions need to be public, I'll try making them private

pkg/clustermanager/cluster_manager.go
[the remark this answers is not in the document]
chrisdoherty4: `DeepEqual` uses reflection and consequently is remarkably slow. On simple strings, for example, we can see a 32x slow down.

```go
var inputA, inputB string = "hello world", "world hello"

func BenchmarkDeepEqual(b *testing.B) {
	for i := 0; i <= b.N; i++ {
		reflect.DeepEqual(inputA, inputB)
	}
}

func BenchmarkAEqualsB(b *testing.B) {
	for i := 0; i <= b.N; i++ {
		if inputA == inputB {
		}
	}
}
```

```
BenchmarkDeepEqual-12           12468793                97.20 ns/op
BenchmarkAEqualsB-12            394434675                3.097 ns/op
```

pkg/providers/cloudstack/cloudstack.go
maxdrib: @vivek-koppuru do you have any thoughts on why this is being set? It's taken from vsphere

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -845,6 +1048,75 @@ func (p *cloudstackProvider) GenerateStorageClass() []byte {
...
+	}
+
+	cpMachineConfigName := clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name
+	if prevSpec.Spec.ControlPlaneConfiguration.MachineGroupRef.Name != cpMachineConfigName {
+		em, err := p.providerKubectlClient.SearchCloudStackMachineConfig(ctx, cpMachineConfigName, cluster.KubeconfigFile, clusterSpec.Cluster.GetNamespace())
+		if err != nil {
+			return err
+		}
+		if len(em) > 0 {
+			return fmt.Errorf("control plane VSphereMachineConfig %s already exists", cpMachineConfigName)
```
maxdrib, line 1100: fix error message

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -250,6 +251,32 @@ func TestProviderGenerateDeploymentFileWithMirrorAndCertConfig(t *testing.T) {
...
+	kubectl := mocks.NewMockProviderKubectlClient(mockCtrl)
+	cluster := &types.Cluster{Name: "test"}
+	clusterSpec := givenClusterSpec(t, clusterSpecManifest)
+	datacenterConfig := givenDatacenterConfig(t, clusterSpecManifest)
+	machineConfigs := givenMachineConfigs(t, clusterSpecManifest)
+	ctx := context.Background()
+	cmk := givenWildcardCmk(mockCtrl)
+	provider := newProviderWithKubectl(t, datacenterConfig, machineConfigs, clusterSpec.Cluster, kubectl, cmk)
+
+	if err := provider.SetupAndValidateCreateCluster(ctx, clusterSpec); err != nil {
```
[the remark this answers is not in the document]
maxdrib: Offline discussion - better to test GenerateCapiSpecForCreate independently, not rely on the logic in SetupAndValidateCreateCluster

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -45,6 +45,19 @@ type CloudStackResourceIdentifier struct {
 	Name string `json:"name,omitempty"`
 }
 
+func (r *CloudStackResourceIdentifier) Equals(o *CloudStackResourceIdentifier) bool {
```
g-gaston, line 48: nit: I think we have been using `Equal` as the standard

g-gaston, line 48: Tests for all these `Equal` methods?

pkg/providers/cloudstack/config/template-md.yaml
```diff
@@ -73,7 +85,7 @@ kind: MachineDeployment
 metadata:
   labels:
     cluster.x-k8s.io/cluster-name: {{.clusterName}}
-  name: {{.clusterName}}-md-0
```
jiayiwang7, line 76: md name should be "workerNodeGroupName" instead of `workloadkubeadmconfigTemplateName`

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
maxdrib: Added tests

maxdrib: renamed

controllers/controllers/resource/testdata/etcdadmcluster.yaml
vivek-koppuru: I see it was because we split it up between vsphere and cloudstack.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -717,15 +801,138 @@ func (p *cloudstackProvider) generateCAPISpecForCreate(ctx context.Context, clus
...
 	}
+
 	workersSpec, err = p.templateBuilder.GenerateCAPISpecWorkers(clusterSpec, workloadTemplateNames, kubeadmconfigTemplateNames)
 	if err != nil {
 		return nil, nil, err
 	}
 	return controlPlaneSpec, workersSpec, nil
 }
 
+func (p *cloudstackProvider) generateCAPISpecForUpgrade(ctx context.Context, bootstrapCluster, workloadCluster *types.Cluster, currentSpec, newClusterSpec *cluster.Spec) (controlPlaneSpec, workersSpec []byte, err error) {
```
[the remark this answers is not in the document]
maxdrib: done

pkg/providers/cloudstack/cloudstack.go
maxdrib: done

pkg/clustermanager/cluster_manager.go
```diff
@@ -523,6 +525,52 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
 				return true, nil
 			}
 		}
+	case v1alpha1.CloudStackDatacenterKind:
```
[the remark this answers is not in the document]
maxdrib: looping over datacenter configs made it much shorter - does that look better?

pkg/providers/cloudstack/cloudstack.go
maxdrib: dried out :)

pkg/providers/cloudstack/cloudstack_test.go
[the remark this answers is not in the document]
maxdrib: We need to call both because we're asserting on a legitimate and valid output file. `SetupAndValidateCreateCluster` sets some internal provider attributes which are then used by `GenerateCAPISpecForCreate`. If we want to isolate a test to only `GenerateCAPISpecForCreate`, this requires either manually modifying the internal components of the provider as part of the test setup, or asserting against an illogical/invalid output file.

I am inclined to keep the tests as they are for now, since this appears to be a deeper issue than malformed tests. There is an implicit dependency between these two methods and I don't feel like this is the PR to unravel it. If you'd like, I'd be happy to create a GitHub issue to track it

pkg/clustermanager/cluster_manager.go
```diff
@@ -527,6 +528,30 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
...
+		if err != nil {
+			return false, err
+		}
+		csDc := datacenterConfig.(*v1alpha1.CloudStackDatacenterConfig)
+		if !existingCsdc.Spec.Equal(&csDc.Spec) {
+			logger.V(3).Info("New provider spec is different from the new spec")
+			return true, nil
+		}
+
+		for _, config := range machineConfigs {
```
vivek-koppuru, line 542: This loop may not work because it is possible that someone added a new machine config object, so if it looks for it below it will throw an error. I think that is why we had it the way we did before where we check based on the references and not just looping through all the machine configs.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -248,11 +250,7 @@ func (p *cloudstackProvider) UpdateKubeConfig(_ *[]byte, _ string) error {
 }
 
 func (p *cloudstackProvider) BootstrapClusterOpts() ([]bootstrapper.BootstrapClusterOption, error) {
-	execConfig, err := decoder.ParseCloudStackSecret()
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse environment variable exec config: %v", err)
-	}
-	return common.BootstrapClusterOpts(execConfig.ManagementUrl, p.clusterConfig)
+	return common.BootstrapClusterOpts(p.datacenterConfig.Spec.ManagementApiEndpoint, p.clusterConfig)
```
vivek-koppuru, line 253: This is a new change right? Is this a change where we are actually reading the value from our spec now instead of the secret? Does this warrant a separate PR or is it just this simple change?

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: That's fair, but I also think this is a bigger change that should be more of a follow up now that I think about it with the validations we will need to do.

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: It seems like this is the only place we are telling the templateBuilder what the machine config map is. It seems like it would be better if we pass it into the function instead of set it on the object, but that might be better as a follow up.

pkg/clustermanager/cluster_manager.go
maxdrib: Refactored again. Could you take another look?

pkg/clustermanager/cluster_manager.go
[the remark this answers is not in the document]
maxdrib: Split out machine configs after Vivek's feedback that looping over machine configs is bad :)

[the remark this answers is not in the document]
maxdrib: Created https://github.com/aws/eks-anywhere/issues/1669 to cover the refactoring

pkg/providers/cloudstack/cloudstack.go
maxdrib: This is a simple change in my opinion. We already have validation in validator.go to ensure the values are the same

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -117,6 +117,51 @@ func (c *CloudStackMachineConfig) OSFamily() OSFamily {
...
+	if c.Affinity != o.Affinity {
+		return false
+	}
+	if len(c.AffinityGroupIds) != len(o.AffinityGroupIds) ||
+		len(c.Users) != len(o.Users) ||
+		len(c.UserCustomDetails) != len(o.UserCustomDetails) {
+		return false
+	}
+	for i, groupId := range c.AffinityGroupIds {
+		if groupId != o.AffinityGroupIds[i] {
```
vivek-koppuru, line 140: This implies that they have to be the same order. Is that an okay assumption? We have implementations elsewhere where we use a map to compare the actual values. 

https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L450

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -117,6 +117,51 @@ func (c *CloudStackMachineConfig) OSFamily() OSFamily {
...
+	}
+	for i, user := range c.Users {
+		if user.Name != o.Users[i].Name {
+			return false
+		}
+		if len(user.SshAuthorizedKeys) != len(o.Users[i].SshAuthorizedKeys) {
+			return false
+		}
+		for j, key := range user.SshAuthorizedKeys {
+			if key != o.Users[i].SshAuthorizedKeys[j] {
```
vivek-koppuru, line 152: Same with checking for slices here as above:

https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L450

vivek-koppuru (commented):
Just have some last comments regarding slice comparisons

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: neat, will refactor to use this, thank you

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
maxdrib: I found UsersSliceEqual and will use that instead

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -441,13 +462,74 @@ func (p *cloudstackProvider) SetupAndValidateDeleteCluster(ctx context.Context,
...
+	if oldSpec.Cluster.Spec.KubernetesVersion != newSpec.Cluster.Spec.KubernetesVersion {
+		return true
+	}
+	if oldSpec.Bundles.Spec.Number != newSpec.Bundles.Spec.Number {
+		return true
+	}
+	return AnyImmutableFieldChanged(oldCsdc, newCsdc, oldCsmc, newCsmc)
+}
+
+func (p *cloudstackProvider) needsNewMachineTemplate(ctx context.Context, workloadCluster *types.Cluster, currentSpec, newClusterSpec *cluster.Spec, workerNodeGroupConfiguration v1alpha1.WorkerNodeGroupConfiguration, csdc *v1alpha1.CloudStackDatacenterConfig, prevWorkerNodeGroupConfigs map[string]v1alpha1.WorkerNodeGroupConfiguration) (bool, error) {
```
g-gaston, line 509: nit: the number of method arguments here is kind of getting out of hand. I'm ok addressing this separately

pkg/providers/cloudstack/cloudstack.go
g-gaston: It seems like we are building `templateBuilder.WorkerNodeGroupMachineSpecs` in the provider constructor
I could be missing something but we either don't need this line or we need to improve the construction logic in the controller (and then skip this line).
It seems like we only need `machineConfigs` to "update" `templateBuilder.WorkerNodeGroupMachineSpecs` and that is already available in the constructor.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -410,6 +410,25 @@ func (n *KindnetdConfig) Equal(o *KindnetdConfig) bool {
 	return true
 }
 
+func UsersSliceEqual(a, b []UserConfiguration) bool {
```
vivek-koppuru, line 413: Not sure if this makes sense under cluster. Would think under `machineconfig_types.go`, but am okay with this.

vivek-koppuru (approved):

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -853,6 +1075,84 @@ func (p *cloudstackProvider) GenerateStorageClass() []byte {
...
+		etcdUser := p.machineConfigs[p.clusterConfig.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name].Spec.Users[0]
+		p.etcdSshAuthKey = etcdUser.SshAuthorizedKeys[0]
+		if len(p.etcdSshAuthKey) > 0 {
+			p.etcdSshAuthKey, err = common.StripSshAuthorizedKeyComment(p.etcdSshAuthKey)
+			if err != nil {
+				return err
+			}
+		}
+		etcdUser.SshAuthorizedKeys[0] = p.etcdSshAuthKey
+	}
```
g-gaston, line 1110: could we do all of this by iterating over `p.clusterConfig.MachineConfigRefs()`?

pkg/providers/cloudstack/cloudstack_test.go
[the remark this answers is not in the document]
g-gaston: Got it. Yes, please create a ticket. Having an implicit dependency between two exported methods is quite bad.

pkg/api/v1alpha1/cloudstackdatacenterconfig_test.go
```diff
@@ -152,3 +153,36 @@ func TestGetCloudStackDatacenterConfig(t *testing.T) {
...
+	cloudStackDatacenterConfigSpec1.Domain = "newDomain"
+	assert.False(t, cloudStackDatacenterConfigSpec1.Equal(cloudStackDatacenterConfigSpec2), "Domain comparison in CloudStackDatacenterConfigSpec not detected")
+	cloudStackDatacenterConfigSpec2 = cloudStackDatacenterConfigSpec1.DeepCopy()
+
+	cloudStackDatacenterConfigSpec1.Account = "newAccount"
+	assert.False(t, cloudStackDatacenterConfigSpec1.Equal(cloudStackDatacenterConfigSpec2), "Account comparison in CloudStackDatacenterConfigSpec not detected")
+	cloudStackDatacenterConfigSpec2 = cloudStackDatacenterConfigSpec1.DeepCopy()
+
+	cloudStackDatacenterConfigSpec1.Zones = nil
+	assert.False(t, cloudStackDatacenterConfigSpec1.Equal(cloudStackDatacenterConfigSpec2), "Zones comparison in CloudStackDatacenterConfigSpec not detected")
```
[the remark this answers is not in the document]
maxdrib: sure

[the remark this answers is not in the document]
maxdrib: done

pkg/api/v1alpha1/cloudstackmachineconfig_test.go
```diff
@@ -206,3 +207,62 @@ func TestGetCloudStackMachineConfigs(t *testing.T) {
...
+	cloudStackMachineConfigSpec2.Users = append(cloudStackMachineConfigSpec2.Users, UserConfiguration{Name: "newUser", SshAuthorizedKeys: []string{"newKey"}})
+	assert.False(t, cloudStackMachineConfigSpec1.Equal(cloudStackMachineConfigSpec2), "Account comparison in CloudStackMachineConfigSpec not detected")
+	cloudStackMachineConfigSpec2 = cloudStackMachineConfigSpec1.DeepCopy()
+
+	cloudStackMachineConfigSpec2.UserCustomDetails = nil
+	assert.False(t, cloudStackMachineConfigSpec1.Equal(cloudStackMachineConfigSpec2), "UserCustomDetails comparison in CloudStackMachineConfigSpec not detected")
+	cloudStackMachineConfigSpec2 = cloudStackMachineConfigSpec1.DeepCopy()
+
+	cloudStackMachineConfigSpec2.UserCustomDetails["i"] = "j"
+	assert.False(t, cloudStackMachineConfigSpec1.Equal(cloudStackMachineConfigSpec2), "UserCustomDetails comparison in CloudStackMachineConfigSpec not detected")
```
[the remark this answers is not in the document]
maxdrib: done

pkg/api/v1alpha1/cloudstackmachineconfig_test.go
```diff
@@ -206,3 +207,62 @@ func TestGetCloudStackMachineConfigs(t *testing.T) {
 		})
 	}
 }
+
+func TestEqualCloudStackMachineConfigSpec(t *testing.T) {
```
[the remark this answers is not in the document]
maxdrib: done

pkg/providers/cloudstack/cloudstack.go
maxdrib: I'd be a bit nervous to do this. It seems like we're at least setting the internal attributes `p.controlPlaneSshAuthKey`, `p.etcdSshAuthKey`, and `p.workerSshAuthKey` explicitly which we won't be able to do if we're just iterating, which are being used elsewhere. Otherwise we'd have to track which machine config we're looking at on each iteration which seems to defeat the purpose

I guess we could iterate over the machineconfigrefs and build a map between machineconfigref name -> auth key but then it's still not clear how to go back from that map to assign back to p.etcdSshAuthKey, for example. I'd be inclined to wait on this work

chrisdoherty4: Where are the ssh key fields used in the provider? For Tinkerbell the idea of populating ssh key state has been stripped out because it made no sense. We actually iterate over machine configs (albeit using some other state) as suggested by @g-gaston because the ssh key generation logic was remarkably convoluted.

Specifically, we found object state being used to communicate between method calls on a single call path - i.e. they should have been parameters. Tinkerbell still isn't right but it simplifies the ssh key generation problem. We hope to make that logic re-usable soon.

maxdrib: If it's going to be made reusable soon, would it make sense to leave the current logic as is (since it works and has been tested) and wait for that functionality to be made reusable? Especially since it's the same across providers

The ssh key fields are used in capi spec generation

pkg/providers/cloudstack/cloudstack.go
maxdrib: I think we're building those based on the old machineconfigs fetched from kubectl, right? So if some values have changed and we need to re-build the capi spec, shouldn't we persist the newer input machine configs instead of the ones we've fetched?

pkg/providers/cloudstack/cloudstack_test.go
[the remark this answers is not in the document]
maxdrib: Created https://github.com/aws/eks-anywhere/issues/1759. Please feel free to add additional details

pkg/providers/cloudstack/cloudstack.go
g-gaston: Can we have a ticket?

pkg/providers/cloudstack/cloudstack.go
g-gaston: What is the need for `p.controlPlaneSshAuthKey`, `p.etcdSshAuthKey`, and `p.workerSshAuthKey`?
They all seem to be extracted from `p.machineConfigs`, which is also a provider member, hence available everywhere. So first sight, they seem redundant. I could be missing something here.

pkg/providers/cloudstack/cloudstack.go
g-gaston: Are you saying `p.machineConfigs` is built from the old version of machine configs? I was under the impression `p.machineConfigs` are the new ones: they are injected in the provider constructor and they don't seem to be modified (except when setting default values).

As far if this is blocking, if we don't fully understand what this code is doing and why is here, then it is blocking. Once we understand it, we can look at the possible solutions and evaluate if they can be addressed separately.

pkg/providers/cloudstack/cloudstack.go
maxdrib: Removing the attributes in favor of using p.machineConfigs directly

pkg/providers/cloudstack/cloudstack.go
maxdrib: Created https://github.com/aws/eks-anywhere/issues/1772 to address the issues caused by this implementation

PR #1676, opened 2022-03-29
CloudStack e2e tests

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR covers the e2e tests for CloudStack

*Testing (if applicable):*
Running these tests is in progress against the provider/controller code in https://github.com/aws/eks-anywhere/pull/1675

### DIFF
--- a/internal/pkg/api/cloudstack.go
+++ b/internal/pkg/api/cloudstack.go
@@ -1,0 +1,242 @@
+package api
+
+import (
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/templater"
+)
+
+type CloudStackConfig struct {
+	datacenterConfig    *v1alpha1.CloudStackDatacenterConfig
+	cpMachineConfig     *v1alpha1.CloudStackMachineConfig
+	workerMachineConfig *v1alpha1.CloudStackMachineConfig
+	etcdMachineConfig   *v1alpha1.CloudStackMachineConfig
+}
+
+type CloudStackFiller func(config CloudStackConfig)
+
+func AutoFillCloudStackProvider(filename string, fillers ...CloudStackFiller) ([]byte, error) {
+	var etcdMachineConfig *v1alpha1.CloudStackMachineConfig
+	// only to get name of control plane and worker node machine configs
+	clusterConfig, err := v1alpha1.GetClusterConfig(filename)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cluster config from file: %v", err)
+	}
+	if clusterConfig.Spec.ControlPlaneConfiguration.MachineGroupRef == nil {
+		return nil, fmt.Errorf("no machineGroupRef defined for control plane")
+	}
+	if len(clusterConfig.Spec.WorkerNodeGroupConfigurations) == 0 {
+		return nil, fmt.Errorf("no worker nodes defined")
+	}
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef == nil {
+		return nil, fmt.Errorf("no machineGroupRef defined for worker nodes")
+	}
+	if clusterConfig.Spec.ExternalEtcdConfiguration != nil {
+		if clusterConfig.Spec.ExternalEtcdConfiguration.MachineGroupRef == nil {
+			return nil, fmt.Errorf("no machineGroupRef defined for etcd machines")
+		}
+	}
+	cpName := clusterConfig.Spec.ControlPlaneConfiguration.MachineGroupRef.Name
+	workerName := clusterConfig.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef.Name
+	cloudstackDatacenterConfig, err := v1alpha1.GetCloudStackDatacenterConfig(filename)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cloudstack datacenter config from file: %v", err)
+	}
+
+	cloudstackMachineConfigs, err := v1alpha1.GetCloudStackMachineConfigs(filename)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cloudstack machine config from file: %v", err)
+	}
+	cpMachineConfig, ok := cloudstackMachineConfigs[cpName]
+	if !ok {
+		return nil, fmt.Errorf("unable to find cloudstack control plane machine config %v", cpName)
+	}
+	workerMachineConfig, ok := cloudstackMachineConfigs[workerName]
+	if !ok {
+		return nil, fmt.Errorf("unable to find cloudstack worker node machine config %v", workerName)
+	}
+
+	config := CloudStackConfig{
+		datacenterConfig:    cloudstackDatacenterConfig,
+		cpMachineConfig:     cpMachineConfig,
+		workerMachineConfig: workerMachineConfig,
+	}
+
+	if clusterConfig.Spec.ExternalEtcdConfiguration != nil {
+		etcdMachineConfig, ok = cloudstackMachineConfigs[clusterConfig.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name]
+		if !ok {
+			return nil, fmt.Errorf("unable to find cloudstack etcd machine config %s", clusterConfig.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name)
+		}
+		config.etcdMachineConfig = etcdMachineConfig
+	}
+	for _, f := range fillers {
+		f(config)
+	}
+
+	cloudstackDatacenterConfigOutput, err := yaml.Marshal(cloudstackDatacenterConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling cloudstack datacenter config: %v", err)
+	}
+	cpMachineConfigOutput, err := yaml.Marshal(cpMachineConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling cloudstack control plane machine config: %v", err)
+	}
+	workerMachineConfigOutput, err := yaml.Marshal(workerMachineConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling cloudstack worker node machine config: %v", err)
+	}
+	cloudstackConfigOutput := templater.AppendYamlResources(cloudstackDatacenterConfigOutput, cpMachineConfigOutput, workerMachineConfigOutput)
+	if clusterConfig.Spec.ExternalEtcdConfiguration != nil {
+		etcdMachineConfigOutput, err := yaml.Marshal(etcdMachineConfig)
+		if err != nil {
+			return nil, fmt.Errorf("error marshalling cloudstack etcd machine config: %v", err)
+		}
+		cloudstackConfigOutput = templater.AppendYamlResources(cloudstackConfigOutput, etcdMachineConfigOutput)
+	}
+	return cloudstackConfigOutput, nil
+}
+
+func WithCloudStackTemplate(value string) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.cpMachineConfig.Spec.Template.Name = value
+		config.workerMachineConfig.Spec.Template.Name = value
+		if config.etcdMachineConfig != nil {
+			config.etcdMachineConfig.Spec.Template.Name = value
+		}
+	}
+}
+
+func WithCloudStackComputeOffering(value string) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.cpMachineConfig.Spec.ComputeOffering.Name = value
+		config.workerMachineConfig.Spec.ComputeOffering.Name = value
+		if config.etcdMachineConfig != nil {
+			config.etcdMachineConfig.Spec.ComputeOffering.Name = value
+		}
+	}
+}
+
+func WithCloudStackManagementServer(value string) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.datacenterConfig.Spec.ManagementApiEndpoint = value
+	}
+}
+
+func WithCloudStackAffinityGroupIds(value []string) CloudStackFiller {
+	return func(config CloudStackConfig) {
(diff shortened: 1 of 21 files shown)


test/e2e/README.md
```diff
@@ -60,6 +64,28 @@ T_TINKERBELL_INVENTORY_CSV # path to hardware-inventory.csv file
...
+T_CLOUDSTACK_DOMAIN
+T_CLOUDSTACK_ZONE
+T_CLOUDSTACK_ACCOUNT
+T_CLOUDSTACK_NETWORK
+T_CLOUDSTACK_MANAGEMENT_SERVER
+T_CLOUDSTACK_SSH_AUTHORIZED_KEY
+T_CLOUDSTACK_TEMPLATE_REDHAT_1_20
+T_CLOUDSTACK_TEMPLATE_REDHAT_1_21
+T_CLOUDSTACK_COMPUTE_OFFERING_LARGE
+T_CLOUDSTACK_COMPUTE_OFFERING_LARGER
```
vivek-koppuru, line 80: Why do we have these two env vars?

test/e2e/README.md
```diff
@@ -60,6 +64,28 @@ T_TINKERBELL_INVENTORY_CSV # path to hardware-inventory.csv file
...
+T_CLOUDSTACK_CONTROL_PLANE_HOST
+T_CLOUDSTACK_POD_CIDR
+T_CLOUDSTACK_SERVICE_CIDR
+
+EKSA_CLOUDSTACK_B64ENCODED_SECRET (when running against local eks-a)
```
vivek-koppuru, line 85: Why is this documented as such? We don't need this set for CI environment?

test/e2e/conformance_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build conformance_e2e
```
vivek-koppuru, line 1: Is this something your editor is replacing?

test/e2e/flux_test.go
```diff
@@ -164,6 +165,57 @@ func TestVSphereKubernetes122GitopsOptionsFlux(t *testing.T) {
 			api.WithFluxNamespace(fluxUserProvidedNamespace),
 			api.WithFluxConfigurationPath(fluxUserProvidedPath),
 		),
+	)
+	runFluxFlow(test)
+}
+
+func TestCloudStackKubernetes120Flux(t *testing.T) {
```
vivek-koppuru, line 172: We probably don't need all the CloudStack tests here, we only need the ones that would specifically test Cloudstack functionality compared to our tests, so based on this it seems like we only really need one of the Flux tests.

test/e2e/workload_clusters_test.go
```diff
@@ -140,3 +141,32 @@ func TestDockerUpgradeWorkloadClusterWithFlux(t *testing.T) {
 		provider.WithProviderUpgradeGit(),
 	)
 }
+
+func TestCloudStackKubernetes121WorkloadClusterDemo(t *testing.T) {
+	t.Skip("Skipping CloudStack in CI/CD")
```
vivek-koppuru, line 146: Any reason we are skipping this specifically and not the others?

internal/pkg/api/vsphere.go
```diff
@@ -176,15 +176,15 @@ func WithDatacenter(value string) VSphereFiller {
 	}
 }
 
-func WithStringFromEnvVar(envVar string, opt func(string) VSphereFiller) VSphereFiller {
+func WithVSphereStringFromEnvVar(envVar string, opt func(string) VSphereFiller) VSphereFiller {
```
vivek-koppuru, line 179: There is no need for this to be VSphere specific right? We have have it generic with an interface I believe, so that we don't unnecessarily maintain these functions in two places.

test/e2e/workload_clusters_test.go
maxdrib: This would require kube-vip or some load balancer to be natively supported but that's not currently an option with CAPC v0.4.1. It will be unblocked with CAPC v0.4.2

internal/pkg/api/vsphere.go
maxdrib: Is it already implemented? From what I can tell, we're returning a VSphereFiller type. What's your suggestion here?

test/e2e/conformance_test.go
maxdrib: I'm not sure. I guess so? Is it an issue?

test/e2e/README.md
maxdrib: Yes, will change documentation. This is a required var for running eksa with cloudstack

test/e2e/README.md
maxdrib: That would be for supporting upgrade from the smaller compute offering to the larger one as a test case

test/e2e/flux_test.go
maxdrib: Which one of these do you feel is best to leave in that case?

test/e2e/README.md
vivek-koppuru: Can we just make these a part of the test value then instead of as an env var? It doesn't seem necessary to have these as env vars to me, but maybe we wanted to abstract it?

test/e2e/workload_clusters_test.go
vivek-koppuru: Can we add a comment mentioning this, or not have this test until we check that code in?

internal/pkg/api/vsphere.go
vivek-koppuru: Now that I look at the code more, we may need to make an interface that can handle `VSphereConfig` and `CloudStackConfig` and make `VSphereFiller` more generic like `ProviderFiller`, but that would probably be better for a follow up. I think this is fine for now.

maxdrib: Created https://github.com/aws/eks-anywhere/issues/1758, feel free to add additional details

internal/pkg/api/cloudstack.go
```diff
@@ -0,0 +1,242 @@
...
+	workerMachineConfig *v1alpha1.CloudStackMachineConfig
+	etcdMachineConfig   *v1alpha1.CloudStackMachineConfig
+}
+
+type CloudStackFiller func(config CloudStackConfig)
+
+func AutoFillCloudStackProvider(filename string, fillers ...CloudStackFiller) ([]byte, error) {
+	var etcdMachineConfig *v1alpha1.CloudStackMachineConfig
+	// only to get name of control plane and worker node machine configs
+	clusterConfig, err := v1alpha1.GetClusterConfig(filename)
```
jiayiwang7, line 25: you can use `v1alpha1.GetAndValidateClusterConfig` to avoid the validations below.

jiayiwang7: actually you might just use `cluster.ParseConfigFromFile` to get the `Config` object that has everything you need in the `CloudStackConfig`

test/framework/flux.go
```diff
@@ -280,6 +281,57 @@ func (e *ClusterE2ETest) validateWorkerNodeMultiConfigUpdates(ctx context.Contex
 			return fmt.Errorf("machine template name should change on machine resource updates, old %s and new %s", machineTemplateName, newMachineTemplateName)
 		}
 		return nil
+	case v1alpha1.CloudStackDatacenterKind:
```
jiayiwang7, line 284: I suggest not using this method anymore for flux upgrade test. This method is messy and not flexible. Instead, use `UpgradeWithGitOps` method when creating the tests.

reference: https://github.com/aws/eks-anywhere/blob/main/test/e2e/workload_clusters_test.go#L66

test/framework/flux.go
```diff
@@ -474,6 +540,25 @@ func (e *ClusterE2ETest) providerConfig(clusterConfGitPath string) (*providerCon
 			e.ClusterConfig.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef.Name,
 			etcdName,
 			machineConfigs)
+	case v1alpha1.CloudStackDatacenterKind:
```
jiayiwang7, line 543: All the extra code here can be avoided with the new Flux framework I mentioned above, where you just need to pass the ClusterFillers to `UpgradeWithGitOps` and control what fields to update there.

maxdrib:
Also need to modify this file to skip all cloudstack tests initially, until test infrastructure is ready https://github.com/aws/eks-anywhere/blob/main/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml

internal/pkg/api/cloudstack.go
maxdrib: done

test/framework/flux.go
wanyufe: Updated to use new method

test/framework/flux.go
wanyufe: Does this mean those test cases in flux.go for cloudstack can be removed? since the workload_clusters_test.go have the flux clodstack test case (upgrade and horizonal and scale combined)

internal/pkg/api/cloudstack.go
wanyufe: updated to use v1alpha1.GetAndValidateClusterConfig.
confg returned from cluster.ParseConfigFromFile does not have cloudstack info. 
```
type Config struct {
	Cluster               *anywherev1.Cluster
	VSphereDatacenter     *anywherev1.VSphereDatacenterConfig
	DockerDatacenter      *anywherev1.DockerDatacenterConfig
	SnowDatacenter        *anywherev1.SnowDatacenterConfig
	VSphereMachineConfigs map[string]*anywherev1.VSphereMachineConfig
	SnowMachineConfigs    map[string]*anywherev1.SnowMachineConfig
	OIDCConfigs           map[string]*anywherev1.OIDCConfig
	AWSIAMConfigs         map[string]*anywherev1.AWSIamConfig
	GitOpsConfig          *anywherev1.GitOpsConfig
	FluxConfig            *anywherev1.FluxConfig
}
```

Does Cloudstack miss a field under Config?

jiayiwang7: yes it is new. So you will need to add cloudstack field in this Config struct

test/e2e/workload_clusters_test.go
```diff
@@ -140,3 +141,71 @@ func TestDockerUpgradeWorkloadClusterWithFlux(t *testing.T) {
 		provider.WithProviderUpgradeGit(),
 	)
 }
+
+func TestCloudStackKubernetes121WorkloadClusterDemo(t *testing.T) {
```
jiayiwang7, line 145: nit: 
```suggestion
func TestCloudStackKubernetes121WorkloadCluster(t *testing.T) {
```

test/framework/cluster.go
```diff
@@ -593,6 +601,30 @@ func (e *ClusterE2ETest) GetEksaVSphereMachineConfigs() []v1alpha1.VSphereMachin
 	return machineConfigs
 }
 
+func (e *ClusterE2ETest) GetEksaCloudStackMachineConfigs() []v1alpha1.CloudStackMachineConfig {
```
jiayiwang7, line 604: is this func still needed?

test/framework/flux.go
jiayiwang7: workload_clusters_test only covers workload cluster flux upgrade. we still need management cluster's flux upgrade test (with the new flux framework). And this `case` block can be removed.

wanyufe: when you say `management cluster`, you mean the mangement cluster in a multiple cluster `MulticlusterE2ETest` ? if so, where is the simliar code for vsphere ( I assume it already used the new flux framework)

jiayiwang7: 1. I mean the management cluster test in the simple/NOT MulticlusterE2ETest case
2. upgrade I mean the upgrade command -- anything that's testable for upgrade
3. no. upgrade K8s version only applies to workload cluster

wanyufe: Thanks for confirm my understand. 
Is the test case in file flux_test.go you are looking for? 
```
func TestCloudStackKubernetes121GitopsOptionsFlux(t *testing.T) {
	provider := framework.NewCloudStack(t, framework.WithRedhat120())
	test := framework.NewClusterE2ETest(
			t,
			provider,
			framework.WithFlux(),
			framework.WithClusterFiller(
				api.WithKubernetesVersion(v1alpha1.Kube120),
				api.WithControlPlaneCount(1),
				api.WithWorkerNodeCount(1),
				api.WithStackedEtcdTopology(),
			),
	)

	test.RunClusterFlowWithGitOps(
		framework.WithClusterUpgradeGit(
			api.WithWorkerNodeCount(3),
		),
		provider.WithProviderUpgradeGit(
			framework.UpdateRedhatTemplate121Var(),
		),
	)
}
```

test/framework/cluster.go
wanyufe: Not used.

test/framework/flux.go
wanyufe: case block removed since not being used after using newer flux test methods.

test/framework/flux.go
```diff
@@ -327,6 +328,20 @@ func (e *ClusterE2ETest) convertVSphereMachineConfigs(cpName, workerName, etcdNa
 	return configs
 }
 
+func (e *ClusterE2ETest) convertCloudStackMachineConfigs(cpName, workerName, etcdName string, cloudstackMachineConfigs map[string]*v1alpha1.CloudStackMachineConfig) []providers.MachineConfig {
```
jiayiwang7, line 331: unused method

test/framework/flux.go
```diff
@@ -577,6 +592,44 @@ func (e *ClusterE2ETest) validateWorkerNodeMachineSpec(ctx context.Context, clus
 			e.T.Logf("Worker MachineTemplate values have matched expected values")
 			return nil
 		})
+	case v1alpha1.CloudStackDatacenterKind:
```
jiayiwang7, line 595: can we pls remove this? we don't need it anymore right?

wanyufe: The new flux test for cloudstack still uses it:
TestCloudStackUpgradeMulticlusterWorkloadClusterWithFlux -> runWorkloadClusterFlowWithGitOps -> UpgradeWithGitOps -> validateWorkerNodeUpdates -> validateWorkerNodeMachineSpec

jiayiwang7 (approved):

PR #1681, opened 2022-03-29
Tag before pushing in docker registry destination

g-gaston:
Builds on top of https://github.com/aws/eks-anywhere/pull/1518

## Description of changes

In order to be able to push to an external registry, the image needs to
have a proper tag pointing to that registry url. This adds the tag just
before pushing.

## Testing (if applicable)
Added unit tests


pkg/docker/mover.go
```diff
@@ -17,6 +17,11 @@ type ImagePusher interface {
 	PushImage(ctx context.Context, image string, endpoint string) error
 }
 
+type ImageTaggerPusher interface {
+	ImagePusher
+	TagImage(ctx context.Context, image string, endpoint string) error
```
jiayiwang7, line 22: why dont we just add this in `ImagePusher`?

g-gaston: Because `ImagePusher` is supposed to only push (hence the name)
I can delete `ImagePusher` and have both methods defined in `ImageTaggerPusher` if you prefer

jiayiwang7: yea lets remove `imagePusher` then. the level in interface is not that necessary imo

PR #1685, opened 2022-03-29
Added tinkerbellHegelURL field in the Tinkerbell clusterconfig file

panktishah26:
*Description of changes:*
Added tinkerbellHegelURL field in the clusterconfig file to replace metadata_url in the template field. I haven't added e2e test changes in this PR. 

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/testdata/cluster_tinkerbell_external_etcd.yaml
+++ b/pkg/providers/tinkerbell/testdata/cluster_tinkerbell_external_etcd.yaml
@@ -42,7 +42,7 @@ kind: TinkerbellDatacenterConfig
 metadata:
   name: test
 spec:
-  tinkerbellCertURL: "http://1.2.3.4/cert"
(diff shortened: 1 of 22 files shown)


pkg/providers/tinkerbell/testdata/cluster_tinkerbell_external_etcd.yaml
```diff
@@ -147,18 +148,18 @@ spec:
       - environment:
           CONTENTS: |
             datasource:
-            Ec2:
-              metadata_urls: ["http://1.2.3.4:50061"]
-              strict_id: false
+              Ec2:
```
ptrivedi, line 151: looks like a space got introduced here. That indentation will make Ec2 a child of datasource

panktishah26: Ec2 is a child of datasource so I had to make that change for the test files.

PR #1688, opened 2022-03-29
Process images concurrently in docker image mover

g-gaston:
Builds on top of https://github.com/aws/eks-anywhere/pull/1518

*Description of changes:*
Little performance improvement for the docker image mover.

Now it runs image push and pulls concurrently. Before this is was doing it sequentially, one after the other. It will use GOMAXPROCS/2 routines, which should be equivalent to half the number of processor cores (in most platforms).

Introduced a new matcher for contexts. It only checks that the argument can be assigned to a `context.Context` variable.

*Testing (if applicable):*
Added unit tests

### DIFF
--- a/pkg/docker/concurrent.go
+++ b/pkg/docker/concurrent.go
@@ -1,0 +1,137 @@
+package docker
+
+import (
+	"context"
+	"fmt"
+	"sync"
+)
+
+type ConcurrentImageProcessor struct {
+	maxRoutines int
+}
+
+func NewConcurrentImageProcessor(maxRoutines int) *ConcurrentImageProcessor {
+	return &ConcurrentImageProcessor{maxRoutines: maxRoutines}
+}
+
+type ImageProcessor func(ctx context.Context, image string) error
+
+func (c *ConcurrentImageProcessor) Process(ctx context.Context, images []string, process ImageProcessor) error {
+	contextWithCancel, abortRemainingJobs := context.WithCancel(ctx)
+	workers := min(len(images), c.maxRoutines)
+	workers = max(workers, 1)
+	jobsChan := make(chan job)
+	wg := &sync.WaitGroup{}
+	errors := make(chan error)
+	doneChan := make(chan struct{})
+
+	for i := 0; i < workers; i++ {
+		w := &worker{
+			jobs:        jobsChan,
+			process:     process,
+			waitGroup:   wg,
+			errorReturn: errors,
+		}
+		go w.start()
+	}
+
+	f := &feeder{
+		jobs:            jobsChan,
+		images:          images,
+		workersWaiGroup: wg,
+		done:            doneChan,
+	}
+	go f.feed(contextWithCancel)
+
+	var firstError error
+
+loop:
+	for {
+		select {
+		case <-doneChan:
+			break loop
+		case err := <-errors:
+			if firstError == nil {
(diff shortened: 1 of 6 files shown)


pkg/docker/concurrent.go
```diff
@@ -0,0 +1,137 @@
...
+func NewConcurrentImageProcessor(maxRoutines int) *ConcurrentImageProcessor {
+	return &ConcurrentImageProcessor{maxRoutines: maxRoutines}
+}
+
+type ImageProcessor func(ctx context.Context, image string) error
+
+func (c *ConcurrentImageProcessor) Process(ctx context.Context, images []string, process ImageProcessor) error {
+	contextWithCancel, abortRemainingJobs := context.WithCancel(ctx)
+	workers := min(len(images), c.maxRoutines)
+	workers = max(workers, 1)
```
jiayiwang7, line 22: isn't # of workers max always 1 here?

jiayiwang7, line 22: isn't this always return 0 or 1?

g-gaston: it will never return 0, that was the goal.
`workers` has to be at least one, never more than `maxRoutines` and never more than `len(images)`

pkg/docker/concurrent.go
```diff
@@ -0,0 +1,137 @@
...
+
+	var firstError error
+
+loop:
+	for {
+		select {
+		case <-doneChan:
+			break loop
+		case err := <-errors:
+			if firstError == nil {
```
jiayiwang7, line 54: curious why this is needed? can't we run `abortRemainingJobs()` whenever there's an err?

pkg/docker/concurrent.go
```diff
@@ -0,0 +1,137 @@
...
+	return &ConcurrentImageProcessor{maxRoutines: maxRoutines}
+}
+
+type ImageProcessor func(ctx context.Context, image string) error
+
+func (c *ConcurrentImageProcessor) Process(ctx context.Context, images []string, process ImageProcessor) error {
+	contextWithCancel, abortRemainingJobs := context.WithCancel(ctx)
+	workers := min(len(images), c.maxRoutines)
+	workers = max(workers, 1)
+	jobsChan := make(chan job)
```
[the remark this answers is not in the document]
g-gaston: Umm interesting
What would we gain from that? The only thing that would do is to not block the feeder while the workers are busy. But the feeder runs on its own subroutine, so it's not blocking the main execution.

pkg/docker/concurrent.go
g-gaston: Great question
There is a race condition: 
* One worker fails
* The error is sent through the errors channel
* The main routine reads it and cancels the context
* The feeder stops feeding tasks, closes the channel and waits for all workers to finish the current tasks.
* Since the tasks also receive the context, depending on the implementation, they could be stopped when the context is cancelled. This will return an error through the errors channel.
* If this error is returned before the `doneChan` is closed, the main routine will receive it as well.
* If we don't preserve the first error, it will be overwritten by this second error and we lose the original cause of the problem

So answering you question, yes, we can run `abortRemainingJobs()` whenever. But we don't want to overwrite the first error.

jiayiwang7: i see, make sense

PR #1690, opened 2022-03-29
Fix fetch artifacts command for extra flux and oidc files

taneyland:
*Issue #, if available:*

*Description of changes:*
Tests with more than one directory for its artifacts, like Flux and OIDC, had some misplaced files when using the `fetch artifacts` command due to recent changes. This PR fixes those issues to place those files in their correct directories.

*Testing (if applicable):*
`fetch artifacts` command


danbudris (approved):
/lgtm

PR #1706, opened 2022-03-30
Download images command

g-gaston:

*Testing (if applicable):*
It has unit tests


cmd/eksctl-anywhere/cmd/downloadimages.go
```diff
@@ -0,0 +1,71 @@
...
+		ctx := context.Background()
+
+		return downloadImagesRunner.Run(ctx)
+	},
+}
+
+func init() {
+	downloadCmd.AddCommand(downloadImagesCmd)
+
+	downloadImagesCmd.Flags().StringVarP(&downloadImagesRunner.outputFile, "output", "o", "", "Output tarball containing all downloaded images")
```
jiayiwang7, line 38: can we include a bundles-override or similar option to download specific images? this is mostly for dev testing

jiayiwang7: for snow case, the images downloaded and built in the admin AMI can be different from `import images` user runs in admin AMI

cmd/eksctl-anywhere/cmd/internal/commands/move_images.go
```diff
@@ -0,0 +1,36 @@
...
+	Version version.Info
+	Mover   Mover
+}
+
+func (m MoveImages) Run(ctx context.Context) error {
+	images, err := m.Reader.ReadImages(m.Version.GitVersion)
+	if err != nil {
+		return fmt.Errorf("moving images: %v", err)
+	}
+	imageTags := make([]string, 0, len(images))
```
jiayiwang7, line 30: nit: can we rename it to `taggedImages`

pkg/dependencies/factory.go
```diff
@@ -54,6 +57,8 @@ type Dependencies struct {
 	DignosticCollectorFactory diagnostics.DiagnosticBundleFactory
 	CAPIManager               *clusterapi.Manager
 	ResourceSetManager        *clusterapi.ResourceSetManager
+	FileReader                *files.Reader
```
jiayiwang7, line 60: why do we need FileReader as a dependency?

g-gaston: `FileReader` is a dependency for `ManifestReader`
It also gives the flexibility to only build by itself if needed

g-gaston: We are using it to build `ManifestReader`

cmd/eksctl-anywhere/cmd/downloadimages.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := downloadImagesCmd.MarkFlagRequired("output"); err != nil {
+		log.Fatalf("Failed making output flag as required: %v", err)
```
chrisdoherty4, line 40:
```suggestion
		log.Fatalf("Failed marking output flag as required: %v", err)
```

cmd/eksctl-anywhere/cmd/downloadimages.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/aws/eks-anywhere/cmd/eksctl-anywhere/cmd/internal/commands"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/docker"
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
+// imagesCmd represents the images command
+var downloadImagesCmd = &cobra.Command{
```
chrisdoherty4, line 21: If we're going to start a new cmd pattern (i.e. using `internal/commands`) can we begin using initializing constructors to create cmds in an effort to remove package level state that has promoted a noteworthy number of copy-n-paste mistakes.

```go
func NewDownloadImagesCmd() &cobra.Command { ... }
```

This approach much better compliments the ability to bind arguments and error out naturally too as there's no need for an `init` func that prevents anything other than an immediate exit.

chrisdoherty4: Ah sorry, was this a move of existing logic? I missed that if so! If we're just moving stuff around then its not worth changing.

With respect to Cobra cmds and 'compatibility' I wasn't suggesting we couldn't move this a second time, it was more should we begin a precedent in this PR.

chrisdoherty4 (approved):

cmd/eksctl-anywhere/cmd/downloadimages.go
chrisdoherty4: Spoke offline.

Will focus any attempts to adjust how we write commands in a dedicated refactor with design.

Issue #1643, opened 2022-03-28
Test Tool: don't pull artifacts of passed tests


danbudris:
when we pull artifacts for tests using `e2e fetch artifacts`, we end up pulling all the artifacts for all the tests in a given execution.

this is because we used to only publish artifacts for failed tests, so we relied on the presence of artifacts to indicate that they were meaningful. Now, however, we publish junit reports for EACH test, regardless of wether it failed or not, so when we do `fetch artifacts` we end up pulling artifacts for all the tests both passed and failed.

we should add some logic to the tool that allows you to by default only pull artifacts for failed tests, and have a flag to pull all artifacts.

PR #1711, opened 2022-03-30
Default to only fetch artifacts for test failures in eks-a-test-tool

taneyland:
*Issue #, if available:*
#1643 

*Description of changes:*
We originally fetched artifacts for all e2e tests when using the `fetch artifacts` command in the eks-a-test-tool. This PR changes that to default to only fetching artifacts for failed e2e tests. We are also introducing an `--all` flag, so if the user wants to fetch artifacts for all tests, they can do so.

*Testing (if applicable):*


danbudris (commented):
for the methods which we are duplicating from the logFetcher pkg, we should consider just exporting those or locating them in a third package to export them from rather than duplicate them here. it seems that things like getting all of the failed tests, or the `testResult` type, are going to be more universally useful

test/e2e/tools/eks-anywhere-test-tool/pkg/artifacts/artifactFetcher.go
```diff
@@ -71,13 +94,36 @@ func (l *testArtifactFetcher) FetchArtifacts(opts ...FetchArtifactsOpt) error {
...
+	logs, err := l.buildAccountCwClient.GetLogs(*g, *s)
+	if err != nil {
+		return fmt.Errorf("fetching cloudwatch logs: %v", err)
+	}
+
+	failedTests, err := filterFailedTests(logs)
+	if err != nil {
+		return err
 	}
+	failedTestsMap := testResultMap(failedTests)
```
danbudris, line 126: I think it would be useful to have a single public method, such as `GetFailedTests`, which we could use across the tool to do this exact routinue

danbudris (approved):

PR #1717, opened 2022-03-30
Impl tinkerbell workflow removal on force.

chrisdoherty4:
When force creating a cluster the program should remove all workflows associated with the hardware that's been submitted. This is to ensure we pass validation and can correctly provision the hardware with a new workflow.

#### Todo
- Make MAC address checks case insensitive.
- Whatever unit tests we can write.
- Manual e2e test.

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -3,6 +3,7 @@ package tinkerbell
 import (
 	"context"
 	_ "embed"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -98,6 +99,7 @@ type ProviderKubectlClient interface {
 type ProviderTinkClient interface {
 	GetHardware(ctx context.Context) ([]*tinkhardware.Hardware, error)
 	GetWorkflow(ctx context.Context) ([]*tinkworkflow.Workflow, error)
+	DeleteWorkflow(ctx context.Context, workflowIDs ...string) error
 }
 
 type ProviderPbnjClient interface {
@@ -351,6 +353,12 @@ func (p *tinkerbellProvider) configureSshKeys() error {
 func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpec *cluster.Spec) error {
 	logger.Info("Warning: The tinkerbell infrastructure provider is still in development and should not be used in production")
 
+	hardware, err := p.providerTinkClient.GetHardware(ctx)
+	if err != nil {
+		return fmt.Errorf("retrieving tinkerbell hardware: %v", err)
+	}
+	logger.MarkPass("Connected to tinkerbell stack")
+
 	if err := setupEnvVars(p.datacenterConfig); err != nil {
 		return fmt.Errorf("failed setup and validations: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -326,13 +335,18 @@ func (p *tinkerbellProvider) SetupAndValidateCreateCluster(ctx context.Context,
...
+	if err := p.validator.ValidateHardwareConfig(ctx, p.hardwareConfigFile, hardware, p.skipPowerActions); err != nil {
 		return err
 	}
 
-	// If the force flag was set we want to force machines off and set them to boot from PXE.
-	if p.force && !p.skipPowerActions {
-		if err := p.setMachinesToPXEBoot(ctx); err != nil {
+	if p.force {
+		if !p.skipPowerActions {
+			if err := p.setMachinesToPXEBoot(ctx); err != nil {
```
gwesterfieldjr, line 344: does this also power the machines off?

chrisdoherty4: Discussed offline.

abhinavmpandey08 (approved):
/lgtm

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -448,6 +461,102 @@ func (p *tinkerbellProvider) setMachinesToPXEBoot(ctx context.Context) error {
...
+	return nil
+}
+
+func createHardwareIDToMACMapping(hardware []*tinkhardware.Hardware) (map[string]string, error) {
+	hardwareMACLookup := make(map[string]string)
+	for _, h := range hardware {
+		if len(h.Network.Interfaces) == 0 {
+			return nil, fmt.Errorf("hardware manifest without interface: hardware ID = '%v'", h.Id)
+		}
+		hardwareMACLookup[h.Id] = h.Network.Interfaces[0].Dhcp.Mac
```
ptrivedi, line 501: what is the point of creating this map? why can't we just create a set of mac addresses here?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -448,6 +461,102 @@ func (p *tinkerbellProvider) setMachinesToPXEBoot(ctx context.Context) error {
...
+
+	return manifestHardwareMACs, nil
+}
+
+func getWorkflowsIDsFromMACs(hardwareMACs macAddressSet, workflows []*tinkworkflow.Workflow) ([]string, error) {
+	var workflowIDs []string
+	for _, w := range workflows {
+		mac, err := macFromWorkflow(w)
+		if err != nil {
+			return nil, err
```
ptrivedi, line 526: this would just return the error message that unmarshal returns. it would be good to add more specific things to this error here. we might also discuss later if we should print an error and continue with the other workflows here instead of completely bailing

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -448,6 +461,102 @@ func (p *tinkerbellProvider) setMachinesToPXEBoot(ctx context.Context) error {
...
+
+func createMACSetFromHardwareManifests(hardwareMACLookup map[string]string, hardware []tinkv1alpha1.Hardware) (macAddressSet, error) {
+	manifestHardwareMACs := make(macAddressSet)
+	for _, h := range hardware {
+		mac, found := hardwareMACLookup[h.Spec.ID]
+		if !found {
+			return nil, fmt.Errorf("couldn't find mac address for hardware manifest: manifest hardware ID = '%v'", h.Spec.ID)
+		}
+
+		manifestHardwareMACs.Insert(mac)
```
ptrivedi, line 515: Hmmm, looking more, why create this set at all? Why can't the map's key just be the mac? You are going through each hardware, creating a map of id:mac, then going through the hardware again, looking it up in the map to get the mac, then creating a set of the mac addresses, then looking up the mac of the workflow in the set using Contains... phew!

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: See https://github.com/aws/eks-anywhere/pull/1717/files/e81272a79ff4255ae6df850f58a3ab83cad3058b#r839844571

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: Do you mean adding some context?

PR #1718, opened 2022-03-30
Add cilium-chart as a project to be created for registry mirror

mrajashree:
*Description of changes:*
We switched to installing cilium through helm chart, so for a disconnected setup, this chart will be used from the registry mirror. Import-images command pushes the chart under a project `cilium-chart`. Users will have to create this project under their registry


chrisnegus (commented):
/lgtm

PR #1720, opened 2022-03-30
Helm registry mirror

g-gaston:
Related to https://github.com/aws/eks-anywhere/issues/1166

## Description of changes
Instead of modifying the downloaded bundle manifests to point the helm
charts to the registry mirror, we leave them untouched, and we make helm
patch the requests urls on the fly.

The main advantage is that downloaded artifacts remain agnostic from the
registry. This allows to download them before having access to the
registry or even used the same artifacts for different registries.

## Testing (if applicable):
* Added unit tests
* ~I'll test the whole flow e2e before merging~ Tested e2e manually

### DIFF
--- a/cmd/eksctl-anywhere/cmd/importimages.go
+++ b/cmd/eksctl-anywhere/cmd/importimages.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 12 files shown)


cmd/eksctl-anywhere/cmd/importimages.go
```diff
@@ -146,3 +148,7 @@ func getChartUriAndVersion(chart v1alpha1.Image) (uri, version string) {
 	version = chart.Tag()
 	return uri, version
 }
+
+func pushChartURI(chart v1alpha1.Image, registryEndpoint string) string {
+	return urls.ReplaceHost(ociPrefix+filepath.Dir(chart.Image()), registryEndpoint)
```
mrajashree, line 153: nit: Sprintf instead of `+`
```suggestion
	orgURL := fmt.Sprintf("%s%s", ociPrefix, filepath.Dir(chart.Image()))
	return urls.ReplaceHost(orgURL, registryEndpoint)
```

mrajashree (approved):
looks good

PR #1723, opened 2022-03-31
Fix source S3 path for imagebuilder assets

abhay-krishna:
This fixes the Prod release looking at the wrong place for the S3 artifact.


abhay-krishna:
Force merging to expedite release.
cc @danbudris

PR #1727, opened 2022-03-31
Fix delete workflows in Tinkerbell provider

chrisdoherty4:
Builds on #1727

We don't separate static and interactive validation in Tinkerbell. Consequently, the static yaml validation and interactive workflow validation happen in the same function. This means we don't have an opportunity post yaml validation to clean up workflows before we validate none exist.
    
This change passes the force flag all the way down to validation logic so it can warn in stead of error when it sees workflows already in the Tinkerbell stack. In doing so, we gain yaml validation but have to assume something will clean up the workflows (the scrub function) later in the call path.

### DIFF
--- a/pkg/providers/tinkerbell/hardware/config.go
+++ b/pkg/providers/tinkerbell/hardware/config.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 	"github.com/aws/eks-anywhere/pkg/templater"
 )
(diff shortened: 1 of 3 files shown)


pkg/providers/tinkerbell/hardware/config.go
```diff
@@ -109,7 +110,17 @@ func (hc *HardwareConfig) ValidateHardware(skipPowerActions bool, tinkHardwareMa
...
 			if _, ok := tinkWorkflowMap[mac.Mac]; ok {
-				return fmt.Errorf("workflow %s already exixts for the hardware id %s", tinkWorkflowMap[mac.Mac].Id, hw.Spec.ID)
+				message := fmt.Sprintf("workflow %s already exixts for the hardware id %s", tinkWorkflowMap[mac.Mac].Id, hw.Spec.ID)
+
+				// If the --force-cleanup flag was set we have to warn. This is beacuse we haven't separated static
+				// and interactive validations so there's no opportunity, after performing static yaml validation, to
+				// delete any workflows before we check if workflows alredy exist. To avoid erroring out before getting
+				// the chance to delete workflows this code assumes workflows will be deleted at a later stage,
+				// therefore warns only.
+				if !force {
```
ptrivedi, line 120: As discussed, this would really need revisiting. I understand static and interactive validations are not separate, but this can be definitely made cleaner. This feels like an after-thought and a band-aid.

PR #1730, opened 2022-03-31
 Allow users to configure cloud provider and CSI Driver with different credentials 

mdsgabriel:
*Issue #, if available:*

*Description of changes:*
This will allow users to configure cloud provider and CSI Driver with different credentials using new environment variables.
The new environment variables are optional, and if not set, the required variables (EKSA_VSPHERE_USERNAME and EKSA_VSPHERE_PASSWORD) will be used.

Currently the same credentials are used when configuring CAPV, cloud provider and CSI driver. This change will allow users to set new environment variables to be used when configuring cloud provider and CSI credentials:

```
        // cloud provider credentials
	"EKSA_VSPHERE_CP_USERNAME"
	"EKSA_VSPHERE_CP_PASSWORD"
	// CSI driver credentials 
	"EKSA_VSPHERE_CSI_USERNAME"
	"EKSA_VSPHERE_CSI_PASSWORD"
```

The new credentials are optional, if not set then the following credentials will be used
```
     "EKSA_VSPHERE_USERNAME"
     "EKSA_VSPHERE_PASSWORD"
```

*Testing (if applicable):*
Added unit tests

### DIFF
--- a/pkg/providers/vsphere/testdata/expected_results_main_cp_cloud_provder_and_csi_driver_credentials.yaml
+++ b/pkg/providers/vsphere/testdata/expected_results_main_cp_cloud_provder_and_csi_driver_credentials.yaml
@@ -2,134 +2,91 @@ apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
   labels:
-    cluster.x-k8s.io/cluster-name: {{.clusterName}}
-  name: {{.clusterName}}
-  namespace: {{.eksaSystemNamespace}}
+    cluster.x-k8s.io/cluster-name: test
+  name: test
+  namespace: eksa-system
 spec:
   clusterNetwork:
     pods:
-      cidrBlocks: {{.podCidrs}}
+      cidrBlocks: [192.168.0.0/16]
     services:
-      cidrBlocks: {{.serviceCidrs}}
+      cidrBlocks: [10.96.0.0/12]
   controlPlaneRef:
     apiVersion: controlplane.cluster.x-k8s.io/v1beta1
     kind: KubeadmControlPlane
-    name: {{.clusterName}}
+    name: test
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
     kind: VSphereCluster
-    name: {{.clusterName}}
-{{- if .externalEtcd }}
+    name: test
   managedExternalEtcdRef:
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/config/template-cp.yaml
```diff
@@ -1043,8 +1043,8 @@ stringData:
       name: cloud-provider-vsphere-credentials
       namespace: kube-system
     stringData:
-      {{.vsphereServer}}.password: "{{.eksaVspherePassword}}"
-      {{.vsphereServer}}.username: "{{.eksaVsphereUsername}}"
+      {{.vsphereServer}}.password: "{{(or .eksaControlPlanePassword .eksaVspherePassword)}}"
```
chrisdoherty4, line 1046: Is there a plan to migrate to 1 way of doing this? In general, 1 way is better than 2 but with env variables I feel like its more of a necessity as it can be confusing whats configuring what.

chrisdoherty4 (changes requested):
.

pkg/providers/vsphere/config/template-cp.yaml
```diff
@@ -577,8 +577,8 @@ stringData:
         thumbprint = "{{.thumbprint}}"
 
         [VirtualCenter "{{.vsphereServer}}"]
-        user = "{{.eksaVsphereUsername}}"
-        password = "{{.eksaVspherePassword}}"
+        user = "{{( or .eksaCSIUsername .eksaVsphereUsername)}}"
```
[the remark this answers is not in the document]
mdsgabriel: done

mdsgabriel:
> Could we also change the username and password to be wrapped in single quotes making them literals so escape characters don't cause issues?

done

pkg/providers/vsphere/config/template-cp.yaml
mdsgabriel: At some point this should be modified to use something other than environment variables.

jiayiwang7 (approved):
/lgtm

mdsgabriel:
> > Could we also change the username and password to be wrapped in single quotes making them literals so escape characters don't cause issues?
> 
> done

decided not to use single quotes for now

chrisdoherty4 (approved):

jiayiwang7 (approved):
/lgtm

PR #1732, opened 2022-03-31
pin linter execution in github actions to specific go version

danbudris:
*Issue #, if available:*

*Description of changes:*
Checking if we can set goversion this way for the linter

*Testing (if applicable):*

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -8,7 +8,9 @@ on:
   pull_request:
 jobs:
   golangci:
-    go-version: 1.16.x
+    strategy:
+      matrix:
+        go-version: 1.16.x
     name: lint
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -11,8 +11,12 @@ jobs:
...
     steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.15
       - uses: actions/checkout@v2
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
           version: latest
+          skip-go-installation: true
```
chrisdoherty4, line 22: I think this is redundant. See https://github.com/golangci/golangci-lint-action#compatibility

chrisdoherty4 (approved):

.github/workflows/golangci-lint.yml
danbudris: ah I see

danbudris: turns out it looks like we need this

abhinavmpandey08 (approved):
/lgtm

PR #1743, opened 2022-04-01
Add unit test to verify generated bundle against golden file

abhay-krishna:
Add unit test to verify generated bundle against golden file
Following changes were done to make the bundle file reproducible in dry-run
* Set predictable fake SHA for image and archives
* Set release time to Unix epoch start time 

The test will run the dry-run of dev-release on both `main` branch and release branches.


release/pkg/assets_capa.go
```diff
@@ -193,7 +193,7 @@ func (r *ReleaseConfig) GetAwsBundle(imageDigests map[string]string) (anywherev1
 		}
 	}
 
-	componentChecksum := generateComponentHash(artifactHashes)
+	componentChecksum := generateComponentHash(artifactHashes, r.DryRun)
```
g-gaston, line 196: nit: we should maybe make `generateComponentHash` a `ReleaseConfig` receiver to avoid passing flags
Or maybe inject a `hashGenerator` on construction that would be different for dry run than for normal execution
We can do this separately

release/pkg/file_reader.go
```diff
@@ -349,3 +327,29 @@ func ReadHttpFile(uri string) ([]byte, error) {
 
 	return data, nil
 }
+
+func CheckFilesEquals(t *testing.T, actualPath, expectedPath string) {
```
g-gaston, line 331: should this method be in a test package?

release/pkg/generate_spec_test.go
```diff
@@ -0,0 +1,25 @@
...
+		{
+			testName:           "Default dev-release dry-run",
+			actualBundleFile:   "../downloaded-artifacts/dry-run-bundle-release.yaml",
+			expectedBundleFile: "testdata/bundle-release.yaml",
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.testName, func(t *testing.T) {
+			CheckFilesEquals(t, tt.actualBundleFile, tt.expectedBundleFile)
```
g-gaston, line 22: I'm missing something, how is `../downloaded-artifacts/dry-run-bundle-release.yaml` being generated?
I was expecting it to be generated from the test

release/pkg/file_reader.go
abhay-krishna: Done

release/pkg/assets_capa.go
abhay-krishna: Will make a note and implement this in a follow up

release/pkg/generate_spec_test.go
```diff
@@ -0,0 +1,169 @@
...
+				t.Fatalf("Error generating bundles manifest: %+v\n", err)
+			}
+
+			bundleManifest, err := yaml.Marshal(bundle)
+			if err != nil {
+				t.Fatalf("Error marshaling bundles manifest: %+v\n", err)
+			}
+
+			expectedBundleManifestFile := filepath.Join(gitRoot, releaseFolder, testdataFolder, fmt.Sprintf("%s-bundle-release.yaml", tt.buildRepoBranchName))
+			generatedBundleManifestFile := filepath.Join(releaseConfig.ArtifactDir, fmt.Sprintf("%s-dry-run-bundle-release.yaml", tt.buildRepoBranchName))
```
g-gaston, line 160: nit: I would pass this file names from the testcase

g-gaston (approved):
/hold
/lgtm

Just one nit

Leaving the last approval for @chrisdoherty4

release/pkg/versions.go
```diff
@@ -98,7 +100,10 @@ func (v *cliVersioner) patchVersion() (string, error) {
 	return v.cliVersion, nil
 }
 
-func generateComponentHash(hashes []string) string {
+func generateComponentHash(hashes []string, dryRun bool) string {
```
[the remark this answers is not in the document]
g-gaston: Related to this: https://github.com/aws/eks-anywhere/pull/1743#discussion_r840728120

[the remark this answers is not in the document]
chrisdoherty4: @g-gaston @abhay-krishna I'm not thrilled about leaving this as a follow up. It feels like we're just churning out tech debt. This seems like something we should address now.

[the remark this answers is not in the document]
abhay-krishna: So the only reason we need the bool in the function is to return a constant fake hash value in case we're doing a dry run and not an actual release. Do you think we can have a getter method for the constant and do the if check bifurcation in these assets files instead of in the method to compute manifests hash?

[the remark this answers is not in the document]
abhay-krishna: I have addressed this using the method described above.

release/pkg/utils/utils.go
```diff
@@ -83,3 +86,17 @@ func GetManifestFilepaths(devRelease bool, bundleNumber int, kind, branch string
 	}
 	return manifestFilepath
 }
+
+func GetFakeSHA(hashType int) (string, error) {
+	if (hashType != 256) && (hashType != 512) {
+		return "", fmt.Errorf("Unsupported hash algorithm: %d", hashType)
```
chrisdoherty4, line 92:
```suggestion
		return "", fmt.Errorf("unsupported hash algorithm: %d", hashType)
```

abhay-krishna: Done

release/pkg/generate_spec_test.go
abhay-krishna: I initially had it that way, but parametrized it since I wanted these filenames to be constructed using branch information from the test case.

PR #1750, opened 2022-04-04
Update domain and account resolve

wongni:
*Issue #, if available:*

*Description of changes:*
Currently domain: property is treated as a domain name, however, the domain name is not something unique. Therefore, when there are ROOT/domain1 and ROOT/foo/domain1 for "domain1" domain name validation fails because there are two results. To be more robust, domain: property now can have domain path as well as domain name. In the above example, a user can pass "foo/domain1" to specify the domain with "ROOT/foo/domain1" instead of "ROOT/domain1". If "domain1" name is unique enough then the user still can pass "domain1" into domain: property.

*Testing (if applicable):*
Unit testing.


pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -55,9 +55,12 @@ type CloudStackZone struct {
 	Network CloudStackResourceIdentifier `json:"network"`
 }
 
+// Important: Run "make" to regenerate code after modifying this file
+
 // CloudStackDatacenterConfigStatus defines the observed state of CloudStackDatacenterConfig
 type CloudStackDatacenterConfigStatus struct { // INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// UUID of the CloudStack domain
+	DomainId string `json:"domainId"`
```
maxdrib, line 63: why is this in status? Shouldn't it be in spec, in line with the existing "domain" attribute, or expanded to be a CloudStackResourceIdentifier? (line 30). At the very least, we should include a better description of the behavior for the customer for input domain

pkg/providers/cloudstack/validator.go
```diff
@@ -105,6 +97,28 @@ func (v *Validator) ValidateCloudStackDatacenterConfig(ctx context.Context, data
 	return nil
 }
 
+func (v *Validator) validateDomainAndAccount(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
+	if (datacenterConfig.Spec.Domain != "" && datacenterConfig.Spec.Account == "") ||
+		(datacenterConfig.Spec.Domain == "" && datacenterConfig.Spec.Account != "") {
+		return fmt.Errorf("both domain and account must be specified or none of them must be specified")
```
maxdrib, line 103: i think you can specify domain without account

pkg/executables/cmk.go
```diff
@@ -192,7 +195,15 @@ func (c *Cmk) ValidateZonesPresent(ctx context.Context, zones []v1alpha1.CloudSt
 func (c *Cmk) ValidateDomainPresent(ctx context.Context, domain string) (v1alpha1.CloudStackResourceIdentifier, error) {
 	domainIdentifier := v1alpha1.CloudStackResourceIdentifier{Name: domain, Id: ""}
 	command := newCmkCommand("list domains")
-	applyCmkArgs(&command, withCloudStackName(domain), appendArgs("listall=true"))
+
+	if domain == rootDomain {
```
chrisdoherty4, line 199: Is `ROOT` a fixed constant in CloudStack or is it configurable?

pkg/executables/cmk.go
```diff
@@ -192,7 +195,15 @@ func (c *Cmk) ValidateZonesPresent(ctx context.Context, zones []v1alpha1.CloudSt
...
+		applyCmkArgs(&command, withCloudStackName(domain))
+	} else {
+		tokens := strings.Split(domain, domainDelimiter)
+		domainName := tokens[len(tokens)-1]
```
chrisdoherty4, line 203: This seems convoluted. If a user supplies a domain `foo/bar` can we not search `foo/bar` instead of `foo`? 

If `foo/bar` is an absolute reference to an existing domain, great, we have a match. 

In the event the user only supplies `bar` and we get back `ROOT/bar` and `ROOT/foo/bar` like the PR description suggests is possible then we error because the domain they've specified is ambiguous.

chrisdoherty4: The link below also suggests there are some other options that may alter the response structure such as `level`. Have we experimented with those as they may yield something thats more easily inspectable for ambiguous domains?

pkg/executables/cmk.go
```diff
@@ -208,15 +219,26 @@ func (c *Cmk) ValidateDomainPresent(ctx context.Context, domain string) (v1alpha
...
 	}
 	domains := response.CmkDomains
-	if len(domains) > 1 {
-		return domainIdentifier, fmt.Errorf("duplicate domain %s found", domain)
-	} else if len(domains) == 0 {
+	if len(domains) == 1 {
+		domainIdentifier.Id = domains[0].Id
+		domainIdentifier.Name = domains[0].Name
+	} else if len(domains) > 1 {
+		domainPath := rootDomain + domainDelimiter + domain
```
chrisdoherty4, line 226: Use `strings.Join([]string{rootDomain, domain}, domainDelimeter)` to minimize allocations.

pkg/executables/cmk.go
wongni: > This seems a little convoluted. If a user supplies a domain `foo/bar` can we not search `foo/bar` instead of extracting and searching just `bar`?

Unfortunately, ACS 'list domains' API does not support querying by 'path'. That's why I have to parse the 'name' part to list all domains with it. Then I find a single domain that has the domain 'path'

> If `foo/bar` is an absolute reference to an existing domain, great, we have a match.
> 
> In the event the user only supplies `bar` and we get back `ROOT/bar` and `ROOT/foo/bar` like the PR description suggests is possible then we error because the domain they've specified is ambiguous.

Yes, in this case the pre-flight check returns an error that it discovered multiple domains based on the provided information.

pkg/providers/cloudstack/validator.go
wongni: > i think you can specify domain without account

When deploy a VM in a sub-domain if a matching account is not passed along with the domain ID the VM is created under the parent domain. This is why we enforce both account and domain.

pkg/executables/cmk.go
chrisdoherty4: I see. Can someone search for a parent domain? 

E.g.

```
# Domains
# ROOT/foo/bar
# ROOT/foo/qux

$ cmk list domains foo listall=true # results in both ROOT/foo/bar and ROOT/foo/qux
```

chrisdoherty4: Spoke with Wonkun. 

We noted a few examples where modeling the code to make assumptions about what the user wants to do could be quite undesirable. 

Specifically, if the user has a complex path structure that exhibits the same domain name we're assuming we know which one they want which is wrong. If there's ambiguity the user needs to clear that up for us. 

Example
```
ROOT/foo
ROOT/qux/foo
ROOT/bar/foo
ROOT/aux/bar/foo
```
We would need to choose 1 of 4 options and we're essentially guessing what the user wants.

I proposed introducing a requirement to specify _either_ a path or a domain name using the same parameter and they must be distinguishable so handling code knows what to do. This could simply mean paths must be specified with a preceeding slash.

Using the example above, if a user supplies `/foo` its clear they want `ROOT/foo`. If they specify `foo` its a domain and its ambiguous so we would error. 

I feel being able to distinguish the users intent is imperative to modeling the correct behavior.

wongni: > I see. Can someone search for any portion of a domain?
> 
> E.g.
> 
> ```
> # Domains
> # ROOT/foo/bar
> # ROOT/foo/qux
> 
> $ cmk list domains foo listall=true # results in both ROOT/foo/bar and ROOT/foo/qux
> ```

In this case, only "ROOT/foo" is returned.

rejoshed: @chrisdoherty4 the CAPC team discussed this issue, and we believe that FQDN is the correct way to go. We can update the CAPC API description of the field to reflect that.

In this manner we won't have ambiguous logic regarding the resolution of Domain path.

`foo == ROOT/foo` at least for now

In the future we do mean to break out domain resolution into a Go type including Path and ID, and at that point we could even entertain the idea of a Name field which would fill the niche for usage of just `foo` as you've detailed above. Unfortunately, to change field names and or types, we'd need to introduce API version conversion webhooks.

These webhooks are on our roadmap, but not something we can implement in time for our customer's needs.

This PR will solve the use case of VM creation in a subdomain while not breaking backward compatibility.

pkg/executables/cmk.go
```diff
@@ -192,7 +195,10 @@ func (c *Cmk) ValidateZonesPresent(ctx context.Context, zones []v1alpha1.CloudSt
 func (c *Cmk) ValidateDomainPresent(ctx context.Context, domain string) (v1alpha1.CloudStackResourceIdentifier, error) {
 	domainIdentifier := v1alpha1.CloudStackResourceIdentifier{Name: domain, Id: ""}
 	command := newCmkCommand("list domains")
-	applyCmkArgs(&command, withCloudStackName(domain), appendArgs("listall=true"))
+	tokens := strings.Split(domain, domainDelimiter)
```
vivek-koppuru, line 198: Can we explain through a comment or some sort what the logic is here so we are clear that we are only looking for the last value of the domain?

pkg/executables/cmk.go
chrisdoherty4: Having chatted offline we discussed sticking with an FQDN. In the case a single response is returned we still need to validate the complete path matches. This means users currently using the product _may_ need to correct to use an FQDN.

With respect to paths that I mistakenly thought we were trying to accommodate here (I think), its deferred as a separate piece of work. 

This behavior will create consistency in all situations and provide a pretty clear expectation for customers.

pkg/executables/cmk.go
chrisdoherty4: Yea agree. I made a comment stipulating a full path search and turns out the cloud monkey interface doesn't support it - would be great to have that as a inline code comment.

wongni: Updated.

vivek-koppuru (approved):
/lgtm

PR #1751, opened 2022-04-04
Download helm charts when saving images to disk

g-gaston:

## Description of changes
This add support for helm charts to the download images command. It
creates a tarball with the tar that docker cli creates plus one tarball
per helm chart.

The command name 'download images' is not very accurate anymore and we
should rethink it. `download artifacts` would make a lot of sense but
since we already have an existing command using that pattern, I will
leave that refactor for a separate future PR.

## Testing
Unit tests and manually run the command and untared the result


pkg/dependencies/factory.go
```diff
@@ -130,17 +130,38 @@ func (f *Factory) WithRegistryMirror(mirror string) *Factory {
...
+		if f.executablesImage != "" {
+			return nil
+		}
+
+		bundles, err := f.dependencies.ManifestReader.ReadBundlesForVersion(version.Get().GitVersion)
+		if err != nil {
+			return fmt.Errorf("retrieving executable tools image from bundle in dependency factory: %v", err)
+		}
+
+		f.executablesImage = bundles.Spec.VersionsBundles[0].Eksa.CliTools.VersionedImage()
```
jiayiwang7, line 151: are the cli image all same across k8s versions?

jiayiwang7 (commented):
looks good, minor comments

pkg/dependencies/factory.go
g-gaston: Right now they all are. I don't see us building different versions of this image but there are no guarantees and the bundles does support that case.

This is just the "default" for commands than don't take a cluster config file (hence we don't have a kubernetes version)
All the ones that do take it (create, upgrade, delete...) will use the correct one for their kubernetes version

cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
```diff
@@ -0,0 +1,83 @@
...
+	Version          version.Info
+	ImageMover       ImageMover
+	ChartDownloader  ChartDownloader
+	Packager         Packager
+	TmpDowloadFolder string
+	DstFile          string
+}
+
+func (d Download) Run(ctx context.Context) error {
+	if err := os.MkdirAll(d.TmpDowloadFolder, os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
g-gaston: Interesting
So use create a filewriter so the folder gets initialized? Because I'm not writing any file directly (with `[]byte`).

Or do you want to integrate the FileWriter with the `Tar` method?

pkg/dependencies/factory.go
g-gaston: Offline discussion: write a comment

Issue #1679, opened 2022-03-29
Test Tool: gracefully handle non-existent CW log streams/groups


danbudris:
Sometimes, when tests fail to run, the SSM run command does not publish any log stream at all. The test tool, when pulling the logs, derrives the log stream names from the SSM command ID, and it the log stream does not exist it barfs. We should gracefully handle this case, print a warning if it does not exist, and continue pulling the logs for the rest of the commands/tests.

PR #1754, opened 2022-04-04
Gracefully handle non-existent CW log streams

taneyland:
*Issue #, if available:*
#1679 

*Description of changes:*
The `eks-a-test-tool e2e fetch logs` command would error out if the log stream for a test did not exist. This PR changes that behavior to send a warning that the log stream does not exist & continue fetching the logs for the rest of the tests.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
+++ b/test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
 
@@ -53,8 +54,12 @@ func (c *Cloudwatch) getLogs(logGroupName string, logStreamName string, startTim
 	for {
 		l, err := c.getLogSegment(logGroupName, logStreamName, startTime, endTime, nextToken)
 		if err != nil {
-			logger.Info("error fetching cloudwatch logs", "group", logGroupName, "stream", logStreamName, "err", err)
-			return nil, err
+			if isInvalidParameterError(err) {
+				logger.Info("log stream does not exist. Proceeding to fetch next log events", "logStream", logStreamName)
+			} else {
+				logger.Info("error fetching cloudwatch logs", "group", logGroupName, "stream", logStreamName, "err", err)
+				return nil, err
+			}
 		}
 		if l.NextForwardToken == nil || nextToken != nil && *nextToken == *l.NextForwardToken {
 			logger.Info("finished fetching logs", "logGroup", logGroupName, "logStream", logStreamName)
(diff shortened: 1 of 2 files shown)


test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
```diff
@@ -83,5 +84,20 @@ func (c Cloudwatch) getLogSegment(logGroupName string, logStreamName string, sta
 	if endTime != nil {
 		input.EndTime = endTime
 	}
-	return c.svc.GetLogEvents(input)
+	logEvents, err := c.svc.GetLogEvents(input)
+	if err != nil {
+		if isInvalidParameterError(err) {
+			logger.Info("log stream does not exist. Proceeding to fetch next log events", "logStream", logStreamName)
+		} else {
```
danbudris, line 91: can we remove this `else` block and just return `logEvents, err`

taneyland: that would still return the `InvalidParameterException` error. However, I did just update this to only check if the error is the `InvalidParameterException` error. If so, give the warning & return `logEvents, nil`. All other cases will return `logEvents, err`. That simplifies it a bit more.

test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
```diff
@@ -83,5 +84,17 @@ func (c Cloudwatch) getLogSegment(logGroupName string, logStreamName string, sta
 	if endTime != nil {
 		input.EndTime = endTime
 	}
-	return c.svc.GetLogEvents(input)
+	logEvents, err := c.svc.GetLogEvents(input)
+	if isInvalidParameterError(err) {
```
chrisdoherty4, line 88: I think this is should be 1 level of abstraction up (`getLogs()`). The `getLogs()` method (the consumer) should really receive the error and _decide to ignore it_. Its nuanced, I know, because there's only 1 consumer right now. To follow my rationale, consider a refactor where `getLogSegments` is called by 2 consumers where 1 wants to ignore the error but the other doesn't; as is, that wouldn't be possible.

test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
```diff
@@ -85,3 +90,10 @@ func (c Cloudwatch) getLogSegment(logGroupName string, logStreamName string, sta
 	}
 	return c.svc.GetLogEvents(input)
 }
+
+func isInvalidParameterError(err error) bool {
+	if awsErr, ok := err.(awserr.Error); ok {
+		return awsErr.Code() == constants.InvalidParameterError
```
chrisdoherty4, line 96: The `cloudwatch` pkg has an `ErrCodeInvalidParameterException` constant we should use for error detection.

chrisdoherty4 (approved):
Good stuff! Just need to update the constant so we don't experience any funky string change problems.

test/e2e/tools/eks-anywhere-test-tool/pkg/cloudwatch/cloudwatch.go
taneyland: Oh awesome. Updated!

chrisdoherty4 (approved):
/lgtm

PR #1755, opened 2022-04-04
Add snow preflight validation and set defaults

jiayiwang7:
*Issue #, if available:*

#1105 

This is a draft PR for initial reviews. will split into features with unit testing.

*Description of changes:*

- check control plane ip uniqueness
- set default ssh key pair if not specified
- validate key pair
- validate image existence on device

*Testing (if applicable):*

will add unit tests


pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,78 @@
...
+	configManager := cluster.NewConfigManager()
+	return &ConfigManagerEntry{
+		configManager: configManager,
+		validator:     NewValidator(aws),
+		defaulters:    NewDefaulters(aws, writer),
+	}
+}
+
+func (cm *ConfigManagerEntry) SetDefaultsAndValidate(ctx context.Context, config *cluster.Config) error {
+	if err := cm.configManager.Register(cm.snowEntry(ctx)); err != nil {
```
g-gaston, line 29: I would recommend creating the `configManager` here on the fly. Register will store the entry in the config manager, hence if you run this twice, it will run the old entry and the new entry

pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,78 @@
...
+	providerValidator "github.com/aws/eks-anywhere/pkg/providers/validator"
+)
+
+type ConfigManagerEntry struct {
+	configManager *cluster.ConfigManager
+	validator     *Validator
+	defaulters    *Defaulters
+}
+
+func NewConfigManagerEntry(aws aws.Clients, writer filewriter.FileWriter) *ConfigManagerEntry {
```
g-gaston, line 19: I would pass the validator and the defaulters as args here

pkg/providers/snow/snow.go
```diff
@@ -38,19 +37,21 @@ type snowProvider struct {
...
+	configManagerEntry    *ConfigManagerEntry
 }
 
 type ProviderKubectlClient interface {
 	DeleteEksaDatacenterConfig(ctx context.Context, snowDatacenterResourceType string, snowDatacenterConfigName string, kubeconfigFile string, namespace string) error
 	DeleteEksaMachineConfig(ctx context.Context, snowMachineResourceType string, snowMachineConfigName string, kubeconfigFile string, namespace string) error
 }
 
-func NewProvider(providerKubectlClient ProviderKubectlClient, writer filewriter.FileWriter, now types.NowFunc) *snowProvider {
+func NewProvider(providerKubectlClient ProviderKubectlClient, writer filewriter.FileWriter, aws aws.Clients, now types.NowFunc) *snowProvider {
```
g-gaston, line 48: I would pass the maybe the config manager entry as input to this constructor to keep it aware only of its direct dependencies

pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/aws"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/networkutils"
+	providerValidator "github.com/aws/eks-anywhere/pkg/providers/validator"
+)
+
+type ConfigManagerEntry struct {
```
g-gaston, line 13: Nit:
```suggestion
type ConfigManager struct {
```

Since from the outside this will read as `snow.ConfigManager`

pkg/providers/snow/snow.go
```diff
@@ -79,6 +80,10 @@ func (p *snowProvider) SetupAndValidateCreateCluster(ctx context.Context, cluste
 		return fmt.Errorf("failed setting up credentials: %v", err)
 	}
 	p.setupMachineConfigs(clusterSpec)
```
g-gaston, line 82: Do we still need this?

pkg/dependencies/factory.go
```diff
@@ -41,6 +43,8 @@ type Dependencies struct {
 	Cmk                       *executables.Cmk
 	Tink                      *executables.Tink
 	Pbnj                      *pbnj.Pbnj
+	Aws                       *aws.Client
+	AwsSnow                   aws.Clients
```
g-gaston, line 47:
```suggestion
	AwsClient                       *aws.Client
	AwsSnowClient                   aws.Clients
```

pkg/dependencies/factory.go
```diff
@@ -300,6 +307,64 @@ func (f *Factory) WithCmk() *Factory {
 	return f
 }
 
+func (f *Factory) WithAws() *Factory {
```
g-gaston, line 310: Where is this one being used?

pkg/aws/snow.go
```diff
@@ -0,0 +1,52 @@
...
+
+func snowEndpoint(deviceIP string) *ServiceEndpoint {
+	return &ServiceEndpoint{
+		ServiceID:     "EC2",
+		SigningRegion: "snow",
+		URL:           fmt.Sprintf("https://%s:%d", deviceIP, snowEC2Port),
+	}
+}
+
+func WithSnow(deviceIP string, certsFile string) awsConfigOpts {
```
g-gaston, line 24: nit:
```suggestion
func WithSnowEndpoint(deviceIP string, certsFile string) awsConfigOpts {
```

pkg/providers/snow/defaults.go
```diff
@@ -0,0 +1,57 @@
...
+func (d *Defaulters) setDefaultSshKey(ctx context.Context, m *v1alpha1.SnowMachineConfig) error {
+	if m.Spec.SshKeyName != "" {
+		return nil
+	}
+
+	keyName := fmt.Sprintf("%s-%s.pem", defaultAwsSshKeyNamePrefix, m.GetName())
+
+	logger.V(1).Info("SnowMachineConfig SshKey is empty. Creating default key pair", "default key name", keyName)
+
+	// Only need to fetch IP from single device to create ssh key pair. The created key will then be validated in validateSshKeyPair method
```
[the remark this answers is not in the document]
jiayiwang7: we only need to use one device's credentials to create a key pair. and later validate this new key works on all the devices. the client map is also used for validating all devices connection, ami image existence

pkg/providers/snow/entry.go
jiayiwang7: what's the reason for that?

pkg/aws/snow.go
jiayiwang7: it is not just endpoint, but with ca bundles and device creds

pkg/providers/snow/snow.go
jiayiwang7: yes

jiayiwang7: yes, to set the managed-by annotation. but i can move it to cluster validation.

go.mod
```diff
@@ -4,9 +4,13 @@ go 1.16
 
 require (
 	github.com/aws/aws-sdk-go v1.38.40
+	github.com/aws/aws-sdk-go-v2 v1.16.2
+	github.com/aws/aws-sdk-go-v2/config v1.15.3
+	github.com/aws/aws-sdk-go-v2/service/ec2 v1.34.0
```
chrisdoherty4, line 9: What are the compatibility rules between these modules? They seem related but are clearly versioned separately.

pkg/aws/client.go
```diff
@@ -0,0 +1,48 @@
...
+
+type ServiceEndpoint struct {
+	ServiceID     string
+	URL           string
+	SigningRegion string
+}
+
+type awsConfigOpts []func(*config.LoadOptions) error
+
+func LoadConfig(ctx context.Context, credsFilePath string, opts ...awsConfigOpts) (aws.Config, error) {
```
chrisdoherty4, line 28: This signature is a little odd, its forcing consumers to build option sets rather than simply passing the individual options. I think that complexity should lie in the option itself, not this signature. 

The only options we have at present is `WithSnow()` which is an option set where everything in the set is required. We can model that while still keeping the `LoadConfig` simple.

```go
type AwsConfigOpt func(*config.LoadOptions) error

func LoadConfig(..., opts AwsConfigOpt) (aws.Config, error) { ... }

func AwsConfigOptSet(opts ...AwsConfigOpt) AwsConfigOpt {
  return func(conf *config.LoadOptions) error {
    for _, opt := range opts {
      if err := opt(conf); err != nil {
        return err
      }
    }

    return nil
  }
}

func WithSnow(deviceIP string, certsFile string) AwsConfigOpt {
	return AwsConfigOptSet(
		withCustomCABundleFile(certsFile),
		config.WithSharedConfigProfile(deviceIP),
		config.WithEndpointResolverWithOptions(SnowEndpointResolver(deviceIP)),
	)
}


```

chrisdoherty4: A secondary note, if we expect consumers of the  package to interact with a type its typically best to export that type. It really helps with testability beacuse you can test options modify the `config.LoadOptions` as you expect. This is why, in the example above, I changed the config opt type to be exported.

pkg/aws/creds.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func AwsCABundlesFile() (filePath string, err error) {
+	return validateFileFromEnv(EksaAwsCABundlesFileKey)
+}
+
+func validateFileFromEnv(envKey string) (filePath string, err error) {
+	filePath, ok := os.LookupEnv(envKey)
+	if !ok || len(filePath) <= 0 {
+		return "", fmt.Errorf("env %s is not set or is empty", envKey)
```
chrisdoherty4, line 31:
```suggestion
		return "", fmt.Errorf("env '%s' is not set or is empty", envKey)
```

pkg/aws/creds.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	fileExists := validations.FileExists(filePath)
+	if !fileExists {
+		return "", fmt.Errorf("file %s does not exist", filePath)
```
chrisdoherty4, line 36:
```suggestion
		return "", fmt.Errorf("file '%s' does not exist", filePath)
```

pkg/aws/creds.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func validateFileFromEnv(envKey string) (filePath string, err error) {
+	filePath, ok := os.LookupEnv(envKey)
+	if !ok || len(filePath) <= 0 {
+		return "", fmt.Errorf("env %s is not set or is empty", envKey)
+	}
+
+	fileExists := validations.FileExists(filePath)
+	if !fileExists {
```
chrisdoherty4, line 35:
```suggestion
	if validations.FileExists(filePath) {
```

pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			func(c *cluster.Config) error {
+				for _, m := range c.SnowMachineConfigs {
+					if err := cm.validator.validateSshKeyPair(ctx, m); err != nil {
+						return err
+					}
+				}
+				return nil
+			},
+		},
```
chrisdoherty4, line 72: Could we break this into unexported named functions for readability?

pkg/providers/snow/snow.go
chrisdoherty4: 'Inject all the things' with a root initializing constructor

pkg/cluster/snow.go
```diff
@@ -80,3 +81,22 @@ func processSnowMachineConfig(c *Config, objects ObjectLookup, machineRef *anywh
 
 	c.SnowMachineConfigs[m.GetName()] = m.(*anywherev1.SnowMachineConfig)
 }
+
+func SetSnowMachineConfigsAnnotations(c *Config) error {
```
chrisdoherty4, line 85: Needs tests given the branch logic.

pkg/aws/client.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+)
+
+// Client provides the single API client to make operations call to aws services
+type Client struct {
+	ec2 *ec2.Client
```
chrisdoherty4, line 14: Can we define an interface with the calls we consume and test the `Client`?

pkg/aws/snow.go
```diff
@@ -0,0 +1,55 @@
...
+	return &ServiceEndpoint{
+		ServiceID:     "EC2",
+		SigningRegion: "snow",
+		URL:           fmt.Sprintf("https://%s:%d", deviceIP, snowEC2Port),
+	}
+}
+
+func withCustomCABundleFile(certsFile string) config.LoadOptionsFunc {
+	return func(o *config.LoadOptions) error {
+		caPEM, err := ioutil.ReadFile(certsFile)
```
chrisdoherty4, line 26: Can this be split into `WithCA` and `WithCAFromFile` where the latter calls the former?

pkg/aws/client.go
jiayiwang7: yes i did in an upcoming pr which breaks things down w unit tests

go.mod
jiayiwang7: each submodule has different version calculation. u can check the v1.16.2 release.

pkg/providers/snow/defaults.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return fmt.Errorf("creating ssh key pair in snow: %v", err)
+	}
+
+	path, err := d.writer.Write(keyName, []byte(keyVal), filewriter.PersistentFile, filewriter.Permission0600)
+	if err != nil {
+		return fmt.Errorf("writing private key: %v", err)
+	}
+
+	logger.Info("AWS key pem saved", "sshCommand", fmt.Sprintf("ssh -i %s <username>@<Node-IP-Address>", path))
```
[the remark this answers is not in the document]
jiayiwang7: im not following. yes user can specify multiple machines. each key generated in this function can access the node created from this machine template.

pkg/providers/snow/entry.go
jiayiwang7: im using the same format as all other entries defined in `pkg/cluster`. if u think this is not super readable, we can refactor them all together in a separate pr

pkg/providers/snow/defaults.go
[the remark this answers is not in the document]
chrisdoherty4: We can carry the discussion on in the other PR given its the more up-to-date one.

Issue #1753, opened 2022-04-04
eks anywhere 0.8.0 fail to download artifacts 


dborysenko:
**What happened**:
eksctl anywhere 0.8.0 fails to download artifacts with below error:
```2022-04-04T13:08:40.479-0500	V3	Downloading artifact: https://anywhere-assets.eks.amazonaws.com/releases/bundles/8/artifacts/cluster-api/manifests/infrastructure-docker/v1.0.2/metadata.yaml
2022-04-04T13:08:40.479-0500	V3	Creating local artifact file: eks-anywhere-downloads/1.20/cluster-api-provider-docker/metadata.yaml
2022-04-04T13:08:40.692-0500	V3	Successfully downloaded artifact https://anywhere-assets.eks.amazonaws.com/releases/bundles/8/artifacts/cluster-api/manifests/infrastructure-docker/v1.0.2/metadata.yaml to eks-anywhere-downloads/1.20/cluster-api-provider-docker/metadata.yaml
2022-04-04T13:08:40.692-0500	V3	Downloading artifact: 
2022-04-04T13:08:40.692-0500	V3	Creating local artifact file: eks-anywhere-downloads/1.20/cluster-api-provider-cloudstack
Error: error downloading artifact for component cluster-api-provider-cloudstack: failed reading local file [] for cluster spec: open : no such file or directory
```

**What you expected to happen**:
Expected successful artifacts download.

**How to reproduce it (as minimally and precisely as possible)**:
1. Install latest (0.8.0) version of eksctl anywhere tool.
2. Generate sample cluster config:
`eksctl anywhere generate clusterconfig mgmt --provider vsphere > eksa-mgmt-cluster.yaml`

3. Add registryMirrorConfiguration to generated config
```
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: Cluster
metadata:
  name: mgmt
spec:
  clusterNetwork:
    cniConfig:
      cilium: {}
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    services:
      cidrBlocks:
      - 10.96.0.0/12
  controlPlaneConfiguration:
    count: 2
    endpoint:
      host: ""
    machineGroupRef:
      kind: VSphereMachineConfig
      name: mgmt-cp
  datacenterRef:
    kind: VSphereDatacenterConfig
    name: mgmt
  externalEtcdConfiguration:
    count: 3
    machineGroupRef:
      kind: VSphereMachineConfig
[...]

PR #1763, opened 2022-04-05
Check if manifest exists before downloading

mrajashree:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/1753

*Description of changes:*
CAPC/CAPT providers are not GA and not a part of the bundle-release.yaml corresponding to releases 0.7.x and 0.8.x
But all providers are part of the dev bundle-release.yaml when they get added to the manifest list to download for `download artifacts`. This PR adds a check to see if the provider's artifacts exist in the bundle-release.yaml
This will ensure download artifacts runs against any bundle-release.yaml irrespective of which providers are added in a release.

*Testing (if applicable):*
Tested against release-0.8 branch


vivek-koppuru (approved):
/lgtm

PR #1766, opened 2022-04-05
Untar tools

g-gaston:
*Issue #, if available:*

## Description of changes
This adds tooling to untar files. It supports untaring from an `io.Reader`
which allows two chain this with other readers like gzip. It also
supports to route extracted files to different destinations, including
completely skipping their extraction.

Helpers are provided to extract a tarball into a folder.

Tar logic was updated to support nested folders.

`GzipFileDownloadExtract` was updated to use this new untar util.

## Testing
Unit tests

### DIFF
--- a/pkg/tar/router.go
+++ b/pkg/tar/router.go
@@ -1,0 +1,25 @@
+package tar
+
+import (
+	"archive/tar"
+	"path/filepath"
+)
+
+// Router instructs where to extract a file
+type Router interface {
+	// ExtractPath instructs the path where a file should be extracted.
+	// Empty strings instructs to omit the file extraction
+	ExtractPath(header *tar.Header) string
+}
+
+type FolderRouter struct {
+	folder string
+}
+
+func NewFolderRouter(folder string) FolderRouter {
+	return FolderRouter{folder: folder}
+}
+
+func (f FolderRouter) ExtractPath(header *tar.Header) string {
+	return filepath.Join(f.folder, header.Name)
+}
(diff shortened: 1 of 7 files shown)


pkg/tar/untar.go
```diff
@@ -0,0 +1,65 @@
+package tar
+
+import (
+	"archive/tar"
+	"io"
+	"os"
+)
+
+type UnPackager struct{}
```
jiayiwang7, line 9: why don't we do one Package type with `func (Packager) Package` and `func (Packager) UnPackage`?

pkg/tar/router.go
```diff
@@ -0,0 +1,25 @@
+package tar
+
+import (
+	"archive/tar"
+	"path/filepath"
+)
+
+// Router instructs where to extract a file
+type Router interface {
+	// ExtractPath instructs the path where a file should be extracted
```
jiayiwang7, line 10: nit: u can add a `.` in the end of the first sentense or it becomes 
```
ExtractPath instructs the path where a file should be extracted Empty strings instructs to omit the file extraction
```
in doc/IDE highlights

pkg/tar/tar.go
```diff
@@ -49,25 +40,16 @@ func Tar(source Walker, dst io.Writer) error {
 	return nil
 }
 
-func addToTar(tw *tar.Writer) filepath.WalkFunc {
-	return func(file string, info os.FileInfo, err error) error {
-		if err != nil {
+func addToTar(tw *tar.Writer) TarFunc {
```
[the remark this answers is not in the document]
g-gaston: I doesn't skip them, it just writes their header info
Since the folder has no "content", there is no need to "copy" it to the tar

Issue #1573, opened 2022-03-22
Snow preflight validations


jiayiwang7:
Connection, resource limit, field validation, etc. Ideally the validation logic should be modularized and shared between CLI and controller.

- control plane IP not occupied
- set default ssh key pairs if not specified
- validate ssh key pair
- validate AMI image exists on each device

PR #1767, opened 2022-04-06
Add aws client

jiayiwang7:
*Issue #, if available:*

#1573  

*Description of changes:*

AWS Client for snow preflight validation.

*Testing (if applicable):*

unit testing


pkg/aws/client.go
```diff
@@ -0,0 +1,57 @@
...
+func LoadConfig(ctx context.Context, opts ...AwsConfigOpt) (aws.Config, error) {
+	optFns := []func(*config.LoadOptions) error{}
+
+	for _, opt := range opts {
+		optFns = append(optFns, opt)
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx, optFns...)
+	if err != nil {
+		return aws.Config{}, fmt.Errorf("setting aws config: %v", err)
```
mitalipaygude, line 48: nit: can we just return `nil, fmt.Errorf("setting aws config: %v", err)`?

pkg/aws/creds.go
```diff
@@ -0,0 +1,91 @@
...
+
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := subtractProfileName(scanner.Text())
+		if net.ParseIP(line) != nil {
+			ips = append(ips, line)
+		}
+	}
+	if len(ips) == 0 {
+		return nil, errors.New("no ip address profile found in content")
```
mitalipaygude, line 77: nit: Is there a reason why we are using `errors.New()` as well as `fmt.Errorf` in the same pkg? Should we be consistent with creating those?

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+	_, err := a.ec2.DescribeImages(ctx, params)
+	if err == nil {
+		return true, nil
+	}
+
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "400" {
+		return false, nil
+	}
+	return false, fmt.Errorf("aws describe image [%s]: %v", imageID, err)
```
chrisdoherty4, line 36: This error results in an arbitrary hash that the user must assume represents an image ID. Easily cleared up with a key.

```suggestion
	return false, fmt.Errorf("aws describe image [imageID=%s]: %v", imageID, err)
```

We should also agree on whether errors use square or rounded parenthesis.

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (a *Client) KeyPairExists(ctx context.Context, keyName string) (bool, error) {
```
chrisdoherty4, line 39: I'm not sure a key name being on the target means the key pair exists/is valid. Could you elaborate on expectations here? My train of thought is that key names could be different between my machine and the host machine but the key pairs are still valid. 

Additionally, why does a client perform this check? I would think the client merely provides behavior for interacting with the target but the actual "does this exist on the target" logic would exist elsewhere. Concretely, a client could provide a `RetrieveKeysByName()` method and some other (probably a validating construct) uses that to retrieve the keys and then performs the necessary checks to ensure the key pair is valid (i.e. the business logic). Why is this important? Your validator is now isolatable because you can use an interface to mock up the `RetrieveKeysByName()` and test the business logic very easily.

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+		config.WithSharedCredentialsFiles([]string{credsFile}),
+		config.WithSharedConfigProfile(deviceIP),
+		config.WithEndpointResolverWithOptions(SnowEndpointResolver(deviceIP)),
+	)
+}
+
+func SnowEndpointResolver(deviceIP string) aws.EndpointResolverWithOptionsFunc {
+	endpoint := snowEndpoint(deviceIP)
+	return aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+		if service == endpoint.ServiceID {
```
chrisdoherty4, line 47:
```suggestion
	return aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
		endpoint := snowEndpoint(deviceIP)
		if service == endpoint.ServiceID {
```

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+	DescribeImages(ctx context.Context, params *ec2.DescribeImagesInput, optFns ...func(*ec2.Options)) (*ec2.DescribeImagesOutput, error)
+	DescribeKeyPairs(ctx context.Context, params *ec2.DescribeKeyPairsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeKeyPairsOutput, error)
+	CreateKeyPair(ctx context.Context, params *ec2.CreateKeyPairInput, optFns ...func(*ec2.Options)) (*ec2.CreateKeyPairOutput, error)
+}
+
+func NewEC2Client(config aws.Config) *ec2.Client {
+	return ec2.NewFromConfig(config)
+}
+
+func (a *Client) ImageExists(ctx context.Context, imageID string) (bool, error) {
```
chrisdoherty4, line 23: Using `a` as the receiver parameter which doesn't seem to be relatable to `*client` in any capacity.

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+
+func WithCustomCABundleFile(certsFile string) config.LoadOptionsFunc {
+	caPEM, err := ioutil.ReadFile(certsFile)
+	if err != nil {
+		return func(*config.LoadOptions) error {
+			return fmt.Errorf("reading aws certificates file: %v", err)
+		}
+	}
+
+	return config.WithCustomCABundle(bytes.NewReader(caPEM))
```
chrisdoherty4, line 32: This is OK but there is a nuance in behavior with respect to what a maintainer intuitively expects (probably).

Intuitively I feel a maintainer, given the `WithCustomCABundleFile` doesn't return an error, would assume the file is read _when the option is applied_. That gives cause for me to, hypothetically, write a file _after_ calling `WithCustomCABundleFile` but _before_ the option is applied. While that may seem silly, its quite legitimate when reflecting on the package API and would result in odd behavior because the option would always error.

Below is a suggestion to defer reading the file and still apply the option. It just constructs the actual option in-line when its needed. I also took the liberty to swap out reading the entire file into memory with a buffered io stream to be a little more efficient (probably negligable given the pem won't be huge but its good practice to prefer streams nonetheless; the buffered reader is likely unnecessary too but doesn't hurt). Finally, I changed the error to include the root error as `%w` meaning consumers can get at the root cause if they desire (its a standard library error so doesn't hurt making it part of the pkg api).

```suggestion
	return func(opts *config.LoadOptions) error {
		caPEM, err := os.Open(certsFile)
		if err != nil {
			return fmt.Errorf("reading aws certificates file: %w", err)
		}
		
		customBundleOpt := config.WithCustomCABundle(bufio.NewReader(caPEM))
		if err := customBundleOpt(opts); err != nil {
			return err
		}
		
		return 
	}
```

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+
+func snowEndpoint(deviceIP string) *ServiceEndpoint {
+	return &ServiceEndpoint{
+		ServiceID:     "EC2",
+		SigningRegion: "snow",
+		URL:           fmt.Sprintf("https://%s:%d", deviceIP, snowEC2Port),
+	}
+}
+
+func WithCustomCABundleFile(certsFile string) config.LoadOptionsFunc {
```
chrisdoherty4, line 24: For consistency on this package API, we probably want to return an `AwsConfigOpt`. The compiler shouldn't complain because its an alias.

chrisdoherty4:
All exported package names should have good documentation and documenting non trivial unexported names is strongly encouraged.

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return func(*config.LoadOptions) error {
+			return fmt.Errorf("reading aws certificates file: %v", err)
+		}
+	}
+
+	return config.WithCustomCABundle(bytes.NewReader(caPEM))
+}
+
+func WithSnow(deviceIP string, certsFile, credsFile string) AwsConfigOpt {
```
g-gaston, line 35: maybe `WithSnowEndpoint` or `WithSnowEndpointAccess`?

pkg/aws/snow_test.go
```diff
@@ -0,0 +1,43 @@
+package aws
```
g-gaston, line 1: can we test this from `aws_test`? or do the tests depend on something unexported?

pkg/dependencies/testdata/cluster_snow.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
```
g-gaston, line 1: Is this file being used?

pkg/providers/factory/providerfactory.go
```diff
@@ -24,6 +25,7 @@ type ProviderFactory struct {
 	TinkerbellKubectlClient   tinkerbell.ProviderKubectlClient
 	TinkerbellClients         tinkerbell.TinkerbellClients
 	SnowKubectlClient         snow.ProviderKubectlClient
+	SnowAwsClients            aws.Clients
```
g-gaston, line 28: Is this being used?

pkg/aws/client.go
jiayiwang7: no it's not a pointer

pkg/aws/ec2.go
g-gaston: If we go this route, I would suggest we make sure the client doesn't return any aws sdk specific type.

jiayiwang7: this is amazon ec2 key pairs: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html

pkg/aws/client.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+)
+
+// Client provides the single API client to make operations call to aws services
+type Client struct {
+	EC2 EC2Client
```
chrisdoherty4, line 13: Is there a need to export this?

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,122 @@
...
+func newEC2Test(t *testing.T) *ec2Test {
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	ec2 := mocks.NewMockEC2Client(ctrl)
+	return &ec2Test{
+		WithT: NewWithT(t),
+		ctx:   ctx,
+		ec2:   ec2,
+		client: &aws.Client{
+			EC2: ec2,
```
g-gaston, line 33: nit: we might want to create a constructor to avoid having to export this

g-gaston (approved):
/lgtm
/hold

pkg/aws/ec2.go
chrisdoherty4: #### EC2 key pairs

I read the function name as checking if a key pair exists hence felt the behavior was wrong. Looking at the replies and docs you linked, I think the method name is wrong. We're only trying to provide a guarentee that the key name they supplied exists on the EC2 instance so the method should reflect that: `KeyNameExists()`. Does that resonate?

Off the back of that, all the methods seem to be EC2 specific which makes me think the struct name isn't specific enough - could we rename it to `EC2Client`?

--- 

Re behavior splitting among constructs: I empathize with your sentiment - 👍🏻. I don't know if its the right thing to do because "[doesn't contain] much validation logic" > 0 which means this construct both interfaces with EC2 APIs and validates stuff hence could be violating the SRP. On the flip side, these could just be framed as convenience methods on the EC2 API so I commit!

pkg/aws/creds.go
chrisdoherty4: Take a look at the code for [`fmt.Errorf`](https://cs.opensource.google/go/go/+/go1.18:src/fmt/errors.go;l=17) - invoking that unnecessarily is, less than optimal relative to the simple [`errors.New`](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/errors/errors.go;l=58) 😆.

pkg/aws/ec2.go
jiayiwang7: 1. I feel we should stick to KeyPairExists. We check the aws ec2 key pair existence by the key name.
2. For now it is ec2 specific but I want this struct to be general enough to handle other aws services.

PR #1768, opened 2022-04-06
Add snow preflight validations and set defaults

jiayiwang7:
*Issue #, if available:*

#1573 
Follow up on #1767 

*Description of changes:*

Part of this PR uses aws clients in #1767. Once #1767 is merged, this PR should not be super long.

- check control plane ip uniqueness
- set default ssh key pair if not specified
- validate key pair
- validate image existence on device

*Testing (if applicable):*


pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+
+func snowEndpoint(deviceIP string) *ServiceEndpoint {
+	return &ServiceEndpoint{
+		ServiceID:     "EC2",
+		SigningRegion: "snow",
+		URL:           fmt.Sprintf("https://%s:%d", deviceIP, snowEC2Port),
+	}
+}
+
+func WithCustomCABundleFile(certsFile string) config.LoadOptionsFunc {
```
chrisdoherty4, line 24: For consistency on this package API, you probably want to return an `AwsConfigOpt`. The compiler shouldn't complain because its an alias but if it does you can wrap the return with `AwsConfigOpt()` to perform a type conversion.

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+	caPEM, err := ioutil.ReadFile(certsFile)
+	if err != nil {
+		return func(*config.LoadOptions) error {
+			return fmt.Errorf("reading aws certificates file: %v", err)
+		}
+	}
+
+	return config.WithCustomCABundle(bytes.NewReader(caPEM))
```
chrisdoherty4, line 32: This is OK but there is a nuance in behavior from what a maintainer likely intuitively expects.

Intuitively I feel a maintainer, given the `WithCustomCABundleFile` doesn't return an error, would assume the file is read _when the option is applied_. That gives cause for me to, hypothetically, write a file _after_ calling `WithCustomCABundleFile` but _before_ the option is applied. While that may seem silly, its quite legitimate when reflecting on the package API.

Consequently, is it possible for us to defer reading the file until we actually apply the option so we don't have any potential gotchas? See below.

```suggestion
	return func(opts *config.LoadOptions) error {
		caPEM, err := ioutil.ReadFile(certsFile)
		if err != nil {
			return fmt.Errorf("reading aws certificates file: %v", err)
		}
		
		customBundleOpt := config.WithCustomCABundle(bytes.NewReader(caPEM))
		
		if err := customBundleOpt(opts); err != nil {
			return err
		}
		
		return 
	}
```

pkg/aws/client.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+)
+
+// Client provides the single API client to make operations call to aws services
+type Client interface {
```
chrisdoherty4, line 12: From what I can tell, this interface is superfluous. It looks like we've defined an interface in a provider which is, in the majority of cases, unnecessary - let consumers define the interface with behaviors they expect.

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+	DescribeImages(ctx context.Context, params *ec2.DescribeImagesInput, optFns ...func(*ec2.Options)) (*ec2.DescribeImagesOutput, error)
+	DescribeKeyPairs(ctx context.Context, params *ec2.DescribeKeyPairsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeKeyPairsOutput, error)
+	CreateKeyPair(ctx context.Context, params *ec2.CreateKeyPairInput, optFns ...func(*ec2.Options)) (*ec2.CreateKeyPairOutput, error)
+}
+
+func NewEC2Client(config aws.Config) *ec2.Client {
+	return ec2.NewFromConfig(config)
+}
+
+func (a *client) ImageExists(ctx context.Context, imageID string) (bool, error) {
```
chrisdoherty4, line 23: Using `a` as the receiver parameter which doesn't seem to be related to `*client` in any capacity.

pkg/aws/snow.go
```diff
@@ -0,0 +1,56 @@
...
+		config.WithSharedCredentialsFiles([]string{credsFile}),
+		config.WithSharedConfigProfile(deviceIP),
+		config.WithEndpointResolverWithOptions(SnowEndpointResolver(deviceIP)),
+	)
+}
+
+func SnowEndpointResolver(deviceIP string) aws.EndpointResolverWithOptionsFunc {
+	endpoint := snowEndpoint(deviceIP)
+	return aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+		if service == endpoint.ServiceID {
```
chrisdoherty4, line 47:
```suggestion
	return aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
		endpoint := snowEndpoint(deviceIP)
		if service == endpoint.ServiceID {
```

pkg/providers/snow/defaults.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (d *Defaulters) setDefaultSshKey(ctx context.Context, m *v1alpha1.SnowMachineConfig) (keyName, keyVal string, err error) {
+	if m.Spec.SshKeyName != "" {
+		return "", "", nil
+	}
+
+	keyName = fmt.Sprintf("%s-%s.pem", defaultAwsSshKeyNamePrefix, m.GetName())
+
+	logger.V(1).Info("SnowMachineConfig SshKey is empty. Creating default key pair", "default key name", keyName)
```
chrisdoherty4, line 32: I'm still not sure this is adequate for associating a given key with a particular machine config.

If I have 3 different machine configurations - CP, worker and etcd - how do I know which key belongs to which machine config and therefore which machine?

Moreover, this doesn't compliment other provider behavior where it re-uses the same key across all machine configs missing ssh keys. I battled with how to perform this logic in Tinkerbell because it wasn't readily re-usable from the vSphere provider. In the end, I followed an approximate flow of: (1) find machine configs without keys; (2) if we have machine configs without keys generate a key and tell the user where it is in a machine config agnostic manner; (3) apply the key to machine configs. It was a substantial simplification on the vSphere approach which is extremely convoluted.

Amittedly this all needs reworking so we have reusable defaulting logic but I still think there's some changes required here to meet the same behavior and be easy for the end user.

pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (cm *ConfigManager) snowEntry(ctx context.Context) *cluster.ConfigManagerEntry {
+	return &cluster.ConfigManagerEntry{
+		Defaulters: []cluster.Defaulter{
+			func(c *cluster.Config) error {
+				for _, m := range c.SnowMachineConfigs {
+					if err := cm.defaulters.setDefaultSshKeyAndSaveToFile(ctx, m); err != nil {
+						return err
+					}
```
chrisdoherty4, line 47: I think this embeds too much knowledge for defaulting machine configs. If we macrofy this code to wanting to apply 10 different defaults to a machine config we'll observe 10 different if statements in this inline function, thats not good because its very difficult to test. 

Instead, would we be able to model a `SnowMachineConfigDefaulter` that accepts a `SnowMachineConfig` and knows how to apply all the necessary defaults? Or something to that effect?

pkg/providers/snow/entry.go
```diff
@@ -0,0 +1,74 @@
...
+				return nil
+			},
+			func(c *cluster.Config) error {
+				for _, m := range c.SnowMachineConfigs {
+					if err := cm.validator.validateSshKeyPair(ctx, m); err != nil {
+						return err
+					}
+				}
+				return nil
+			},
```
chrisdoherty4, line 71: [Ditto](https://github.com/aws/eks-anywhere/pull/1768/files#r843890658).

These inline functions shouldn't contain the knowledge of how to actually perform validation.

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "400" {
+		return false, nil
+	}
+	return false, fmt.Errorf("aws describe image [%s]: %v", imageID, err)
+}
+
+func (a *client) KeyPairExists(ctx context.Context, keyName string) (bool, error) {
```
chrisdoherty4, line 39: I'm not sure a key name being on the target means the key pair exists/is valid. Could you elaborate on expectations here? My train of thought is that key names could be different between my machine and the host machine but the key pairs are still valid.

pkg/aws/ec2.go
```diff
@@ -0,0 +1,62 @@
...
+	_, err := a.ec2.DescribeImages(ctx, params)
+	if err == nil {
+		return true, nil
+	}
+
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "400" {
+		return false, nil
+	}
+	return false, fmt.Errorf("aws describe image [%s]: %v", imageID, err)
```
chrisdoherty4, line 36: This error results in an arbitrary hash that the user must assume represents an image ID. Easily cleared up with a key.

```suggestion
	return false, fmt.Errorf("aws describe image [image_id=%s]: %v", imageID, err)
```

We should also agree on whether errors use square or rounded parenthesis.

pkg/providers/snow/entry.go
jiayiwang7: i'd prefer to have a separate convo w @g-gaston on the configManager entry. this is the format we are using in `pkg/cluster`. if we decide to refactor it we can do it in another pr. Not this one.

pkg/providers/snow/entry.go
jiayiwang7: same as above. separate conversation needed

pkg/providers/snow/defaults.go
jiayiwang7: this makes sense. i agree we should do the check and only generate one ssh key for all machines. I will have a separate PR for this and work on it next if you are ok. It is a feature improvement

pkg/providers/snow/validator.go
```diff
@@ -13,22 +13,28 @@ const (
...
 
 func NewValidator(aws aws.Clients) *Validator {
 	return &Validator{
-		awsClients: aws,
+		awsClientMap: NewAwsClientMap(aws),
 	}
 }
 
-func (v *Validator) validateSshKeyPair(ctx context.Context, m *v1alpha1.SnowMachineConfig) error {
+func NewValidatorFromAwsClientMap(awsClientMap AwsClientMap) *Validator {
```
g-gaston, line 25: curious about this, is there a different constructor and that's why we add the `FromAwsClientMap` suffix?

pkg/providers/snow/defaults.go
```diff
@@ -0,0 +1,73 @@
...
+	return &Defaulters{
+		awsClientMap: awsClientMap,
+	}
+}
+
+func (d *Defaulters) GenerateDefaultSshKey(ctx context.Context, m *v1alpha1.SnowMachineConfig) (keyName, keyVal string, err error) {
+	if m.Spec.SshKeyName != "" {
+		return "", "", nil
+	}
+
```
g-gaston, line 35: nit: we probably can remove this?
```suggestion
```

g-gaston (approved):
/lgtm
/hold in case @chrisdoherty4 is still reviewing

Btw, seems like you `cmd/eksctl-anywhere/cmd/internal/commands/artifacts/mocks/download_artifacts.go` got renamed? Maybe some bad rebase?

pkg/providers/snow/entry.go
g-gaston: My 2c: I would be ok with either: what we have here or what @chrisdoherty4 is proposing

I get the fear of this becoming a ball of mud if it keeps growing uncontrolled. However, we currently only have one defaulter of this type and we might not need to add more, so do should we add another entity now or can it wait? I just realized this comment is probably not very helpful, sorry folks 🙂 

As far the config manager is general, it's supposed to just be a "framework" that doesn't necessarily enforce any patterns and it leaves it up to the user to decide what's the best way to use it (multiple defaults in one _defaulter_ closure, multiple closures, nested defaults, etc.

pkg/providers/snow/validator.go
jiayiwang7: this is for unit testing so that we don't export `awsClientMap` field in the struct

pkg/providers/snow/defaults.go
jiayiwang7: i kept it because `GenerateDefaultSshKey` is a exported method and I wanna prevent nil pointer when misused

pkg/providers/snow/entry.go
chrisdoherty4: We talked about this offline and will address separately. 

To elaborate a little further, consider a desire to validate or default a machine config separate from any `ConfigManager` construct. We can't do that using this approach. We have to either use the `ConfigManager` (which is super odd given I just want to validate something) _or_ duplicate all the rules elsewhere hence we aren't embodying validation in a single re-usable construct. Consequently, I would go as far as to say this is the wrong way to model this stuff.

To your point, Guillermo, I totally agree the `ConfigManager` is a framework and I shared the same thing with Joey. How we construct the actual logic is orthogonal so any existing stuff isn't strictly a 'pattern', its just how its been done for that piece of code for now and if done as is here, should probably change also at some point.

chrisdoherty4 (approved):

Issue #1656, opened 2022-03-28
automate checksum generation


danbudris:

danbudris:
this is done

PR #1769, opened 2022-04-06
Automate generating checksum files for release

mitalipaygude:
*Issue #, if available:*
#1656 
*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/scripts/generate_checksum.sh
+++ b/scripts/generate_checksum.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+SHASUM_FILES="shasums-eksctl-anywhere"
+
+#Clean up any existing old folders for shasum files
+if [ -d "$SHASUM_FILES" ]; then rm -Rf $SHASUM_FILES; fi
+
+YQ_LATEST_RELEASE_URL="https://github.com/mikefarah/yq/releases/latest"
+
+if ! command -v yq &> /dev/null
+then
+    echo "Please install YQ from $YQ_LATEST_RELEASE_URL and rerun this script."
+    exit
+fi
+
+echo "YQ is installed, continuing.."
+
+FILE_PREFIX="eksctl-anywhere"
+
+TAR="amd64.tar.gz"
+
+
+CURRENT_YQ_VERSION=$(yq -V | awk '{print $NF}')
+
+versionSplit=($(echo $CURRENT_YQ_VERSION | tr "." "\n"))
+CURRENT_YQ_MAJOR_VERSION=${versionSplit[0]}
+CURRENT_YQ_MINOR_VERSION=${versionSplit[1]}
+EXPECTED_YQ_MAJOR_VERSION=4
+EXPECTED_YQ_MINOR_VERSION=18
+
+if [ $CURRENT_YQ_MAJOR_VERSION -lt $EXPECTED_YQ_MAJOR_VERSION ]; then
+    echo "ERROR: Current yq major version v$CURRENT_YQ_MAJOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.x"
+    echo "Please install the latest version of yq from $YQ_LATEST_RELEASE_URL"
+    exit 1
+fi
+
+if [ $CURRENT_YQ_MAJOR_VERSION == $EXPECTED_YQ_MAJOR_VERSION ] && [ $CURRENT_YQ_MINOR_VERSION -lt $EXPECTED_YQ_MINOR_VERSION ]; then
+    echo "ERROR: Current yq minor version v$CURRENT_YQ_MAJOR_VERSION.$CURRENT_YQ_MINOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.$EXPECTED_YQ_MINOR_VERSION.x"
+    echo "Please install the latest version of yq from $YQ_LATEST_RELEASE_URL"
+    exit 1
+fi
+
+mkdir -p $SHASUM_FILES
+cd $SHASUM_FILES
+curl --silent 'https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml' -o release.yaml
+
+latest_release_version=$(cat release.yaml | yq e '.spec.latestVersion')
+
+latest_release=$(cat release.yaml | yq e '.spec.releases[] | select(.version == "'$latest_release_version'")' >> latest_release.yaml)
+
+for sha in 256 512
+do
+  for os in linux darwin
+  do
+    SHA_VALUE=$(cat latest_release.yaml | yq e ".eksABinary.${os}.sha${sha}")
+    echo "${SHA_VALUE}  ${FILE_PREFIX}-${latest_release_version}-${os}-${TAR}" >> "${FILE_PREFIX}-${latest_release_version}-sha${sha}sums.txt"
+  done
+done
+
+echo "Generated the check sum files inside the ${SHASUM_FILES} folder at the directory where you ran this script from."


scripts/generate_checksum.sh
```diff
@@ -0,0 +1,47 @@
...
+SHASUM_FILES="SHASUMS"
+FILE_PREFIX="eksctl-anywhere"
+
+DARWIN_TAR="darwin-amd64.tar.gz"
+LINUX_TAR="linux-amd64.tar.gz"
+
+release_manifest=$(curl --silent 'https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml')
+
+latest_release_version=$(echo $release_manifest | yq e '.spec.latestVersion' - -v)
+echo $latest_release_version
```
chrisdoherty4, line 25: I don't know where the script is run but its typically good practice to give context to echo statements like this unless they're used in some bash pipe cmd (doesn't look like it judging by the rest of the file).

```suggestion
echo "Latest release version: $latest_release_version"
```

Same applies to some of the other naked echo statements.

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,47 @@
...
+cd $SHASUM_FILES
+
+SHA256_FILENAME="{FILE_PREFIX}-{latest_release_version}-sha256sums.txt"
+SHA512_FILENAME="{FILE_PREFIX}-{latest_release_version}-sha512sums.txt"
+
+echo "${linux_sha256} ${FILE_PREFIX}-${latest_release_version}-${LINUX_TAR}" >> SHA256_FILENAME
+echo "${darwin_sha256} ${FILE_PREFIX}-${latest_release_version}-${DARWIN_TAR}" >> SHA256_FILENAME
+
+echo "${linux_sha512} ${FILE_PREFIX}-${latest_release_version}-${LINUX_TAR}" >> SHA512_FILENAME
+echo "${darwin_sha512} ${FILE_PREFIX}-${latest_release_version}-${DARWIN_TAR}" >> SHA512_FILENAME
```
chrisdoherty4, line 47: Is everything after `>>` meant to be a variable? I know this is a draft so you're probably working out kinks.

scripts/generate_checksum.sh
mitalipaygude: Hey this was for my debugging, wont be a part of the final script.

scripts/generate_checksum.sh
mitalipaygude: yeah I missed the `$` sign here, its updated now

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,46 @@
...
+SHASUM_FILES="shasums-eksctl-anywhere"
+FILE_PREFIX="eksctl-anywhere"
+
+DARWIN_TAR="darwin-amd64.tar.gz"
+LINUX_TAR="linux-amd64.tar.gz"
+
+mkdir -p $SHASUM_FILES
+cd $SHASUM_FILES
+
+release_manifest=$(curl --silent 'https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml' -o release.yaml)
```
abhay-krishna, line 25: Why are we assigning this to a variable?

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,46 @@
...
+
+latest_release_version=$(cat release.yaml | yq e '.spec.latestVersion')
+
+latest_release=$(cat release.yaml | yq e '.spec.releases[] | select(.version == "'$latest_release_version'")' >> latest_release.yaml)
+
+linux_sha256=$(cat latest_release.yaml | yq e '.eksABinary.linux.sha256')
```
abhay-krishna, line 31: There must be a way to put this in a loop over [256, 512] since we're doing very similar operations.

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,46 @@
...
+
+DARWIN_TAR="darwin-amd64.tar.gz"
+LINUX_TAR="linux-amd64.tar.gz"
+
+mkdir -p $SHASUM_FILES
+cd $SHASUM_FILES
+
+release_manifest=$(curl --silent 'https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml' -o release.yaml)
+
+latest_release_version=$(cat release.yaml | yq e '.spec.latestVersion')
```
abhay-krishna, line 27: If piping yaml to `yq`, we need to put a dash after the yq expression to make it read inline string.
```suggestion
latest_release_version=$(cat release.yaml | yq e '.spec.latestVersion' -)
```

mitalipaygude: I think that changed since 4.18.1
https://mikefarah.gitbook.io/yq/

abhay-krishna: Ah I see, we need to make sure we have a check for that then or enforce 4.18+. similar to [this](https://github.com/aws/eks-anywhere-build-tooling/blob/main/build/lib/generate_projects_list.sh#L9-L18) but not quite the same.

Also not everyone might have yq locally, so testing that edge case would be good too.

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,74 @@
...
+EXPECTED_YQ_MINOR_VERSION=18
+
+if [ $(($CURRENT_YQ_MAJOR_VERSION)) -lt $(($EXPECTED_YQ_MAJOR_VERSION)) ]; then
+    echo "ERROR: Current yq major version v$CURRENT_YQ_MAJOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.x"
+    echo "Please install the latest version of yq from $YQ_LATEST_RELEASE_URL"
+    exit 1
+fi
+
+if [ $CURRENT_YQ_MAJOR_VERSION == $(($EXPECTED_YQ_MAJOR_VERSION)) ] && [ $CURRENT_YQ_MINOR_VERSION -lt $(($EXPECTED_YQ_MINOR_VERSION)) ]; then
+    echo "ERROR: Current yq minor version v$EXPECTED_YQ_MAJOR_VERSION.$CURRENT_YQ_MINOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.$EXPECTED_YQ_MINOR_VERSION.x"
```
mrajashree, line 50:
```suggestion
    echo "ERROR: Current yq minor version v$CURRENT_YQ_MAJOR_VERSION.$CURRENT_YQ_MINOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.$EXPECTED_YQ_MINOR_VERSION.x"
```

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,74 @@
...
+
+latest_release=$(cat release.yaml | yq e '.spec.releases[] | select(.version == "'$latest_release_version'")' >> latest_release.yaml)
+
+for sha in 256 512
+do
+
+  linux_sha=$(cat latest_release.yaml | yq e ".eksABinary.linux.sha${sha}")
+  darwin_sha=$(cat latest_release.yaml | yq e ".eksABinary.darwin.sha${sha}")
+
+  echo "${darwin_sha}  ${FILE_PREFIX}-${latest_release_version}-${DARWIN_TAR}" >> "${FILE_PREFIX}-${latest_release_version}-sha${sha}sums.txt"
```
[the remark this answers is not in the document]
mitalipaygude: we clean up these files at the very beginning of this script.

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,74 @@
...
+EXPECTED_YQ_MAJOR_VERSION=4
+EXPECTED_YQ_MINOR_VERSION=18
+
+if [ $(($CURRENT_YQ_MAJOR_VERSION)) -lt $(($EXPECTED_YQ_MAJOR_VERSION)) ]; then
+    echo "ERROR: Current yq major version v$CURRENT_YQ_MAJOR_VERSION.x is older than expected v$EXPECTED_YQ_MAJOR_VERSION.x"
+    echo "Please install the latest version of yq from $YQ_LATEST_RELEASE_URL"
+    exit 1
+fi
+
+if [ $CURRENT_YQ_MAJOR_VERSION == $(($EXPECTED_YQ_MAJOR_VERSION)) ] && [ $CURRENT_YQ_MINOR_VERSION -lt $(($EXPECTED_YQ_MINOR_VERSION)) ]; then
```
mrajashree, line 49: nit: since the check on line 43 already compares the major versions, I think we won't need to compare `if [ $CURRENT_YQ_MAJOR_VERSION == $(($EXPECTED_YQ_MAJOR_VERSION)) ]` again right?

scripts/generate_checksum.sh
```diff
@@ -0,0 +1,74 @@
...
+
+mkdir -p $SHASUM_FILES
+cd $SHASUM_FILES
+curl --silent 'https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml' -o release.yaml
+
+latest_release_version=$(cat release.yaml | yq e '.spec.latestVersion')
+
+latest_release=$(cat release.yaml | yq e '.spec.releases[] | select(.version == "'$latest_release_version'")' >> latest_release.yaml)
+
+for sha in 256 512
```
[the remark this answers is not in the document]
abhay-krishna: c12656e8736166eaead65e0579afd8e8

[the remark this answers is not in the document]
mitalipaygude: LOL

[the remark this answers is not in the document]
mitalipaygude: this bot is giving me the creeps

scripts/generate_checksum.sh
mitalipaygude: we are just checking for less than above for the major versions, but say we have a 5.1 comparing to 4.18, the minor versions should be checked only when the major version is 4, and nothing else.

PR #1771, opened 2022-04-06
Update IRSA doc to point to the working commit and image tag

mrajashree:
*Description of changes:*
The pod identity webhook repository has some recent changes due to which its installation doesn't work. This PR updates the link of the pod identity webhook repo to an older commit that does work, and updates the image to a previous tag instead of latest.

*Testing (if applicable):*
Installed pod identity webhook with the updated Readme and IRSA worked as expected

### DIFF
--- a/docs/content/en/docs/reference/clusterspec/irsa.md
+++ b/docs/content/en/docs/reference/clusterspec/irsa.md
@@ -96,7 +96,13 @@ Set the remaining fields in [cluster spec](https://anywhere.eks.amazonaws.com/do
 
 1. After hosting the service account public signing key and OIDC discovery documents, the applications running in pods can start accessing the desired AWS resources, as long as the pod is mounted with the right service account tokens. This part of configuring the pods with the right service account tokens and env vars is automated by the [amazon pod identity webhook](https://github.com/aws/amazon-eks-pod-identity-webhook). Once the webhook is deployed, it mutates any pods launched using service accounts annotated with `eks.amazonaws.com/role-arn`
 
-1. Follow the [in-cluster installation steps](https://github.com/aws/amazon-eks-pod-identity-webhook#in-cluster) for amazon-eks-pod-identity-webhook. While running the make target, make sure the $KUBECONFIG env var is set to the path of the EKS Anywhere cluster.
+1. Check out [this commit](https://github.com/aws/amazon-eks-pod-identity-webhook/commit/a65cc3d9c61cf6fc43f0f985818c474e0867d786) of the amazon-eks-pod-identity-webhook.
+
+1. Set the $KUBECONFIG env var to the path of the EKS Anywhere cluster. Run the following command:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/reference/clusterspec/irsa.md
```diff
@@ -96,7 +96,13 @@ Set the remaining fields in [cluster spec](https://anywhere.eks.amazonaws.com/do
 
 1. After hosting the service account public signing key and OIDC discovery documents, the applications running in pods can start accessing the desired AWS resources, as long as the pod is mounted with the right service account tokens. This part of configuring the pods with the right service account tokens and env vars is automated by the [amazon pod identity webhook](https://github.com/aws/amazon-eks-pod-identity-webhook). Once the webhook is deployed, it mutates any pods launched using service accounts annotated with `eks.amazonaws.com/role-arn`
 
-1. Follow the [in-cluster installation steps](https://github.com/aws/amazon-eks-pod-identity-webhook#in-cluster) for amazon-eks-pod-identity-webhook. While running the make target, make sure the $KUBECONFIG env var is set to the path of the EKS Anywhere cluster.
+1. Check out [this commit](https://github.com/aws/amazon-eks-pod-identity-webhook/commit/a65cc3d9c61cf6fc43f0f985818c474e0867d786) of the amazon-eks-pod-identity-webhook.
```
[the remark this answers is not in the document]
mrajashree: There is an open issue in that repo, I've also opened up [this PR](https://github.com/aws/amazon-eks-pod-identity-webhook/pull/150) to address it. But till it gets reviewed and merged we will have to point our docs to a commit that works.

docs/content/en/docs/reference/clusterspec/irsa.md
```diff
@@ -96,7 +96,13 @@ Set the remaining fields in [cluster spec](https://anywhere.eks.amazonaws.com/do
...
+
+1. Set the $KUBECONFIG env var to the path of the EKS Anywhere cluster. Run the following command:
```
chrisnegus, line 101: Is this two separate steps? If so, you could write it something like this:
```suggestion
1. Set the $KUBECONFIG env var to the path of the EKS Anywhere cluster.
1. Run the following command:
```

chrisnegus: Also, I wondered if there is other information that a user would need from the [installation instructions](https://github.com/aws/amazon-eks-pod-identity-webhook#in-cluster) that you are missing now. For example, the need to install cert-manager as a pre-requisite and where to get the `Makefile`.

chrisnegus (commented):
I have just a couple of comments, then it's okay from my side.

docs/content/en/docs/reference/clusterspec/irsa.md
mrajashree: @chrisnegus EKS-A already installs cert-manager so we don't need to add this instruction

chrisnegus (approved):
/lgtm

PR #1774, opened 2022-04-06
Make artifacts download tests more robust

g-gaston:
## Description of changes
Before this change, the code was depending on bundles.Charts, which calls
Bundles.Charts, which return a map. This is made the order non
predictable, hence making any tests depending on it, flaky.

This change has the added benefit that it makes the tests independent
from the number of charts stored in the bundles manifest. Hence we can
keep adding more info in the VersionsBundle without having to update
unrelated unit tests every time.


jiayiwang7 (approved):
/lgtm

PR #1782, opened 2022-04-07
Avoid race concition in docker concurrent tests

g-gaston:
*Description of changes:*
Before this change, the processing of the second image could be
triggered before the error from the second one was returned. This would
trigger a panic (since there was no expectation set for such calls in
the mock). The panic would kill the routine without reducing the counter
in the wait group, effectively creating a deadlock.


jiayiwang7 (approved):
good catch!

PR #1789, opened 2022-04-08
Generate single default SSH key and import to all snow devices

jiayiwang7:
*Issue #, if available:*

#1573 

Currently we pick an arbitrary snow device credential to create a key pair in the device. This key is not accessible from other devices unless we call aws ec2 import-key-pair api for each device. 

Refactoring this default key generation behavior so that we use the existing `GenerateSSHAuthKey` method to generate the key pair and import it to all the snow devices. 

*Testing (if applicable):*

- Unit testing
- Manual e2e with default key generated


pkg/providers/snow/defaults.go
```diff
@@ -25,49 +26,35 @@ func NewDefaulters(aws aws.Clients, writer filewriter.FileWriter) *Defaulters {
...
-		return "", "", fmt.Errorf("creating ssh key pair in snow: %v", err)
+		return nil
 	}
 
-	return keyName, keyVal, nil
-}
-
-func (d *Defaulters) GenerateDefaultSshKeyAndSaveToFile(ctx context.Context, m *v1alpha1.SnowMachineConfig) error {
-	if m.Spec.SshKeyName != "" {
+	if common.ValidateSSHAuthKeyExists(d.writer) {
```
jiayiwang7, line 38: another way is to add a field in defaulters struct indicating if the default key is already be generated once and skip the rest.

pkg/providers/common/common.go
```diff
@@ -67,17 +69,18 @@ func GenerateSSHAuthKey(writer filewriter.FileWriter) (string, error) {
 		return "", fmt.Errorf("generating ssh key pair: %v", err)
 	}
 
-	logger.Info(fmt.Sprintf(
-		"Private key saved to %[1]s. Use 'ssh -i %[1]s <username>@<Node-IP-Address>' to login to your cluster node",
-		privateKeyPath,
-	))
+	logger.Info("Private key saved", "sshCommand", fmt.Sprintf("ssh -i %s <username>@<Node-IP-Address>", privateKeyPath))
```
g-gaston, line 72: nit: wondering if we should just print the path to the ssh key as opposed to the command, since this is now generic and not for an specific provider.

pkg/providers/common/common.go
```diff
@@ -67,17 +69,18 @@ func GenerateSSHAuthKey(writer filewriter.FileWriter) (string, error) {
...
 
 	key := string(sshAuthorizedKeyBytes)
 	key = strings.TrimRight(key, "\n")
 
 	return key, nil
 }
 
+func ValidateSSHAuthKeyExists(writer filewriter.FileWriter) bool {
+	return validations.FileExists(filepath.Join(writer.Dir(), publicKeyFileName)) && validations.FileExists(filepath.Join(writer.Dir(), privateKeyFileName))
```
g-gaston, line 81: nit: maybe we should look into moving this `FileExists` method into a `pkg/files` package. I'm not sure if the validations package was designed to be imported from other places, dues to the amount of dependencies it has.

pkg/providers/common/common.go
```diff
@@ -67,17 +69,18 @@ func GenerateSSHAuthKey(writer filewriter.FileWriter) (string, error) {
...
-	))
+	logger.Info("Private key saved", "sshCommand", fmt.Sprintf("ssh -i %s <username>@<Node-IP-Address>", privateKeyPath))
 
 	key := string(sshAuthorizedKeyBytes)
 	key = strings.TrimRight(key, "\n")
 
 	return key, nil
 }
 
+func ValidateSSHAuthKeyExists(writer filewriter.FileWriter) bool {
```
g-gaston, line 80: I would pass the directory path here instead of the writer. That makes the method more reusable since it doesn't depend on having a file writer

pkg/providers/snow/defaults_test.go
```diff
@@ -4,35 +4,28 @@ import (
...
-	g.Expect(k).To(Equal(""))
-	g.Expect(v).To(Equal(""))
 }
 
 func TestSetDefaultSshKeyError(t *testing.T) {
 	g := newConfigManagerTest(t)
 	g.machineConfig.Spec.SshKeyName = ""
-	g.aws.EXPECT().EC2CreateKeyPair(g.ctx, "eksa-default-cp-machine.pem").Return("v", errors.New("error"))
-	k, v, err := g.defaulters.GenerateDefaultSshKey(g.ctx, g.machineConfig)
+	g.aws.EXPECT().EC2ImportKeyPair(g.ctx, "eksa-default", gomock.Any()).Return(errors.New("error"))
```
g-gaston, line 28: I wonder if we should inject a ssh key generator in the defaulter, so we don't need to use `Any()` (eg. we don't generate the real ssh keys in these unit tests). That would make this assertion a bit more robust and the test faster

pkg/providers/common/common.go
chrisdoherty4: Logging this way also creates a less than desirable user experience because it dumps a json string. Ideally users aren't looking at JSON strings in the console (I know we have a bunch of that today).

pkg/providers/snow/defaults_test.go
chrisdoherty4: I did something similar in Tinkerbell. The driver was deterministic behavior for unit testing.

pkg/providers/common/common.go
jiayiwang7: but this cmd is also generic. it applies to all providers

jiayiwang7: 1. this cmd is generic across all providers
2. I don't have strong preference on whether using json or not. I will just remove this change in this PR

pkg/providers/snow/defaults.go
```diff
@@ -25,49 +26,49 @@ func NewDefaulters(aws aws.Clients, writer filewriter.FileWriter) *Defaulters {
...
+			return err
+		}
 	}
 
-	return keyName, keyVal, nil
+	return nil
 }
 
-func (d *Defaulters) GenerateDefaultSshKeyAndSaveToFile(ctx context.Context, m *v1alpha1.SnowMachineConfig) error {
+func (d *Defaulters) SetupDefaultSshKey(ctx context.Context, m *v1alpha1.SnowMachineConfig, keyGenerated *bool) error {
```
g-gaston, line 45: I would make this method part of a different type
Like `MachineConfigsDefaulter`. And that type can contain the boolean
Just to avoid passing a flag as an argument :)

pkg/providers/snow/defaults_test.go
jiayiwang7: updated

pkg/providers/snow/defaults_test.go
```diff
@@ -4,35 +4,34 @@ import (
...
-	g.Expect(k).To(Equal(""))
-	g.Expect(v).To(Equal(""))
 }
 
 func TestSetDefaultSshKeyError(t *testing.T) {
 	g := newConfigManagerTest(t)
 	g.machineConfig.Spec.SshKeyName = ""
-	g.aws.EXPECT().EC2CreateKeyPair(g.ctx, "eksa-default-cp-machine.pem").Return("v", errors.New("error"))
-	k, v, err := g.defaulters.GenerateDefaultSshKey(g.ctx, g.machineConfig)
+	g.keyGenerator.EXPECT().GenerateSSHAuthKey(gomock.Any()).Return(sshKey, nil)
```
g-gaston, line 33: don't we have the writer available so we can set the right expect here?

g-gaston (approved):
/lgtm
/hold

g-gaston (commented):
/lgtm
/hold

PR #1793, opened 2022-04-08
Add support for extraPortMappings in Kind

abhinavmpandey08:
*Description of changes:*
Adding support for extraPortMappings in Kind. This is needed for Tinkerbell provider

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/kind.go
+++ b/pkg/executables/kind.go
@@ -48,6 +48,7 @@ type kindExecConfig struct {
 	KubernetesVersion      string
 	RegistryMirrorEndpoint string
 	RegistryCACertPath     string
+	ExtraPortMappings      []int
 	DockerExtraMounts      bool
 	DisableDefaultCNI      bool
 }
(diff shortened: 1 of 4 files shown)


pkg/executables/kind.go
```diff
@@ -134,6 +135,22 @@ func (k *Kind) WithExtraDockerMounts() bootstrapper.BootstrapClusterClientOption
 	}
 }
 
+func (k *Kind) WithExtraPortMappings(ports []int) bootstrapper.BootstrapClusterClientOption {
+	return func() error {
+		if k.execConfig == nil {
+			return errors.New("kind exec config is not ready")
```
chrisdoherty4, line 141: Should this panic? Its not like a user can do anything about this and it feels very much like a developer has made a significant error that's rendered the program invalid.

pkg/executables/kind_test.go
```diff
@@ -89,6 +89,18 @@ func TestKindCreateBootstrapClusterSuccess(t *testing.T) {
 			wantKindConfig:     "testdata/kind_config_docker_mount_networking.yaml",
 			registryMirrorTest: false,
 		},
+		{
+			name:           "With extra port mappings option",
+			wantKubeconfig: kubeConfigFile,
+			options: []testKindOption{
+				func(k *executables.Kind) bootstrapper.BootstrapClusterClientOption {
```
chrisdoherty4, line 96: How are we checking these ports have actually been mapped?

pkg/executables/kind.go
abhinavmpandey08: I just followed what all the other kind options are doing. If we decide to panic instead, that would be a broader change and I would prefer to do that in a separate PR

pkg/executables/kind_test.go
abhinavmpandey08: In the unit test, I'm just checking if the generated kind config file has the expected port mapping configuration here https://github.com/aws/eks-anywhere/pull/1793/files#diff-9531b36781b70449d56a847ba73cb20fea4b287f003a2e605797e9e518c3db10R1-R23

Since we don't actually create a Kind cluster in unit tests, idk if there's another way to check if the ports actually get mapped in the Kind cluster. We will most likely have to verify this at integration test stage

chrisdoherty4 (approved):

PR #1799, opened 2022-04-08
Use default ssh key if already exists in snow devices

jiayiwang7:
*Issue #, if available:*

Currently we create a default ssh key with name `eksa-default` and import it to snow devices. 
There is a race. If multiple users create clusters in the same device set at the same time, eks-a creates default key for first cluster. The second cluster validation will fail at ec2 import-key-pair step because the key already exists in devices.

*Description of changes:*

Before creating default ssh key, we check if the key exists on devices first. If the key exists on all devices, we skip the key generation and only assign the key name to machine config.

*Testing (if applicable):*

- unit tests

### DIFF
--- a/pkg/providers/snow/defaults.go
+++ b/pkg/providers/snow/defaults.go
@@ -60,12 +60,41 @@ func NewMachineConfigDefaulters(d *Defaulters) *MachineConfigDefaulters {
 	}
 }
 
+func (md *MachineConfigDefaulters) defaultKeyCount(ctx context.Context) (int, error) {
+	var count int
+	for ip, client := range md.defaulters.awsClientMap {
+		keyExists, err := client.EC2KeyNameExists(ctx, defaultAwsSshKeyName)
+		if err != nil {
+			return count, fmt.Errorf("describing key pair on snow device [deviceIP=%s]: %v", ip, err)
+		}
+		if keyExists {
+			count += 1
+		}
+	}
+	return count, nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/providers/snow/defaults.go
```diff
@@ -70,6 +84,21 @@ func (md *MachineConfigDefaulters) SetupDefaultSshKey(ctx context.Context, m *v1
 		return nil
 	}
 
+	keyCount, err := md.defaultKeyCount(ctx)
+	if err != nil {
+		return err
+	}
+
+	if keyCount > 0 && keyCount < len(md.defaulters.awsClientMap) {
+		return fmt.Errorf("default key [keyName=%s] only exists on some of the devices. Use 'aws ec2 import-key-pair' to import this key to all the devices", defaultAwsSshKeyName)
```
g-gaston, line 93: nit: shouldn't we ideally just import it wherever is needed? I'm ok if you want that as an improvement later

pkg/providers/snow/defaults.go
```diff
@@ -70,6 +84,21 @@ func (md *MachineConfigDefaulters) SetupDefaultSshKey(ctx context.Context, m *v1
...
+	}
+
+	if keyCount == len(md.defaulters.awsClientMap) {
```
g-gaston, line 96: do we need a test case for this scenario?

pkg/providers/snow/defaults.go
jiayiwang7: we don't have access to the public key tho. Thus we can't import for the user

jiayiwang7: will follow up on this

g-gaston (approved):
/lgtm

PR #1800, opened 2022-04-09
[WIP] Implement tinkerbell stack installation on Kind

abhinavmpandey08:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/tinkerbell/stack/manifests/pbnj.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
```
jaxesn, line 1: Might be worth considering treating these manifest similar to what we do for capi/capv/etc which are built/pushed to s3 in build tooling and then included in the bundle vs hardcoded into the cli.

PR #1808, opened 2022-04-11
Refactoring cluster manager to invoke provider's SpecChanged method

maxdrib:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/1669

*Description of changes:*
This PR removes the cloudstack-specific logic (in a backwards compatible way) from the cluster-manager and moves it to the provider for both CloudStack and VSphere

*Testing (if applicable):*
Unit tests pass

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -531,7 +531,6 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
 	case v1alpha1.CloudStackDatacenterKind:
 		// TODO: Move outside of just the CloudStack workflow
 		return provider.SpecChanged(ctx, cc, cluster, newClusterSpec, datacenterConfig, machineConfigs)
-		// TODO: Make sure happy paths at least are unit tested, especially if moving code to provider
 	default:
 		// Run upgrade flow
 		return true, nil
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -448,7 +448,7 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster, 
 	return nil
 }
 
-func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *types.Cluster, newClusterSpec *cluster.Spec, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig, provider providers.Provider) (bool, error) {
+func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *types.Cluster, newClusterSpec *cluster.Spec) (bool, error) {
 	cc, err := c.clusterClient.GetEksaCluster(ctx, cluster, newClusterSpec.Cluster.Name)
 	if err != nil {
 		return false, err
@@ -476,15 +476,8 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
 		}
 	}
 
-	logger.V(3).Info("Clusters are the same, checking provider spec")
-	// compare provider spec
-	switch cc.Spec.DatacenterRef.Kind {
-	case v1alpha1.VSphereDatacenterKind, v1alpha1.CloudStackDatacenterKind:
-		return provider.SpecChanged(ctx, cc, cluster, newClusterSpec, datacenterConfig, machineConfigs)
-	default:
-		// Run upgrade flow
-		return true, nil
-	}
+	logger.V(3).Info("Clusters are the same")
+	return false, nil
 }
 
 func (c *ClusterManager) InstallCAPI(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster, provider providers.Provider) error {
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -449,7 +449,7 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster, 
 	return nil
 }
 
-func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *types.Cluster, newClusterSpec *cluster.Spec, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig) (bool, error) {
+func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *types.Cluster, newClusterSpec *cluster.Spec, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig, provider providers.Provider) (bool, error) {
 	cc, err := c.clusterClient.GetEksaCluster(ctx, cluster, newClusterSpec.Cluster.Name)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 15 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -199,6 +200,50 @@ func (p *vsphereProvider) UpdateKubeConfig(_ *[]byte, _ string) error {
...
+		existingVmc, err := p.providerKubectlClient.GetEksaVSphereMachineConfig(ctx, oldMcRef.Name, cluster.KubeconfigFile, newClusterSpec.Cluster.Namespace)
+		if err != nil {
+			return false, err
+		}
+		csmc, ok := machineConfigMap[oldMcRef.Name]
+		if !ok {
+			logger.V(3).Info(fmt.Sprintf("Old machine config spec %s not found in the existing spec", oldMcRef.Name))
+			return true, nil
+		}
+		if !reflect.DeepEqual(existingVmc.Spec, csmc.Spec) {
```
maxdrib, line 238: I recognize that using reflect is non-ideal, but this is what the cluster manager was using previously and to keep the changes smaller, I'd like to continue using it for now

pkg/clustermanager/cluster_manager.go
```diff
@@ -480,105 +479,12 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, cluster *ty
...
-
-		machineConfigsSpecChanged, err := c.cloudstackMachineConfigsSpecChanged(ctx, cc, cluster, newClusterSpec, machineConfigs)
-		if err != nil {
-			return false, err
-		}
-		if machineConfigsSpecChanged {
-			return true, nil
-		}
-		// TODO: Make sure happy paths at least are unit tested, especially if moving code to provider
+	case v1alpha1.VSphereDatacenterKind, v1alpha1.CloudStackDatacenterKind:
```
g-gaston, line 482: Since you implemented it for the two providers (btw thanks!), I would vote to use this directly from the upgrade workflow and avoid injecting the provider here.

In facts providers already implement a method `UpgradeNeeded` that is being called from the workflow. I would suggest moving the implementation of `SpecChanged` to `UpgradeNeeded`

g-gaston (commented):
Looks good! just one suggestion that I think would make it a bit cleaner

pkg/providers/provider.go
```diff
@@ -33,7 +33,7 @@ type Provider interface {
 	GenerateMHC() ([]byte, error)
 	ChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ComponentChangeDiff
 	RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpec *cluster.Spec, clusterSpec *cluster.Spec, workloadCluster *types.Cluster, managementCluster *types.Cluster) error
-	UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec) (bool, error)
+	UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec, cluster *types.Cluster, datacenterConfig DatacenterConfig, machineConfigs []MachineConfig) (bool, error)
```
g-gaston, line 36: I don't think we need to add here `datacenterConfig DatacenterConfig, machineConfigs []MachineConfig`

The providers can retrieve them if necessary
In fact, if you look at how the upgrade workflow is getting them, they come from the provider

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1057,10 +1083,32 @@ func (p *cloudstackProvider) MachineConfigs(_ *cluster.Spec) []providers.Machine
...
-func (p *cloudstackProvider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec) (bool, error) {
+func (p *cloudstackProvider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec, cluster *types.Cluster, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig) (bool, error) {
 	newV, oldV := newSpec.VersionsBundle.CloudStack, currentSpec.VersionsBundle.CloudStack
 
-	return newV.ClusterAPIController.ImageDigest != oldV.ClusterAPIController.ImageDigest, nil
+	if newV.ClusterAPIController.ImageDigest != oldV.ClusterAPIController.ImageDigest {
+		logger.V(3).Info("New provider controller is different from the old spec")
+		return true, nil
+	}
+	cc, err := p.providerKubectlClient.GetEksaCluster(ctx, cluster, newSpec.Cluster.Name)
```
g-gaston, line 1093: I missed this the first time, sorry!
You can get this cluster from `currentSpec.Cluster`
Same for the other providers

maxdrib: using `currentSpec.Cluster` instead of querying with kubectl

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1057,10 +1083,28 @@ func (p *cloudstackProvider) MachineConfigs(_ *cluster.Spec) []providers.Machine
 	return configs
 }
 
-func (p *cloudstackProvider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec) (bool, error) {
+func (p *cloudstackProvider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec, cluster *types.Cluster) (bool, error) {
 	newV, oldV := newSpec.VersionsBundle.CloudStack, currentSpec.VersionsBundle.CloudStack
 
-	return newV.ClusterAPIController.ImageDigest != oldV.ClusterAPIController.ImageDigest, nil
+	if newV.ClusterAPIController.ImageDigest != oldV.ClusterAPIController.ImageDigest {
```
g-gaston, line 1089: I'm not sure this is needed
The capi components upgrade us separate from the cluster upgrade
Only components that need a full cluster upgrade should be considered in this method

pkg/providers/docker/docker.go
```diff
@@ -474,7 +474,7 @@ func (p *provider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpe
 	return nil
 }
 
-func (p *provider) UpgradeNeeded(_ context.Context, _, _ *cluster.Spec) (bool, error) {
+func (p *provider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec, cluster *types.Cluster) (bool, error) {
```
g-gaston, line 477: I would keep the `_`

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -599,6 +599,11 @@ func (m *macAddressSet) Insert(mac string) {
 	(*m)[strings.ToLower(mac)] = struct{}{}
 }
 
+func (p *tinkerbellProvider) SpecChanged(ctx context.Context, cc *v1alpha1.Cluster, cluster *types.Cluster, newClusterSpec *cluster.Spec, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig) (bool, error) {
```
g-gaston, line 602: Can we remove this?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -884,7 +889,7 @@ func (p *tinkerbellProvider) RunPostControlPlaneUpgrade(ctx context.Context, old
 	return nil
 }
 
-func (p *tinkerbellProvider) UpgradeNeeded(_ context.Context, _, _ *cluster.Spec) (bool, error) {
+func (p *tinkerbellProvider) UpgradeNeeded(ctx context.Context, newSpec, currentSpec *cluster.Spec, cluster *types.Cluster) (bool, error) {
```
g-gaston, line 892: Same here, would leave the `_`

g-gaston (commented):
Looks good, just minor comments

pkg/providers/cloudstack/cloudstack.go
maxdrib: I don't think these are capi components - this is the CAPC controller image. If we're replacing the CAPC controller image that should cause an upgrade right? This is the same behavior we had previously.

pkg/providers/tinkerbell/tinkerbell.go
maxdrib: done

pkg/providers/tinkerbell/tinkerbell.go
maxdrib: removed, good catch

pkg/providers/docker/docker.go
maxdrib: reverted

pkg/providers/cloudstack/cloudstack.go
g-gaston: Ahh good points.

I consider the CAPC controller image part of the capi components (or any other provider controller for that matter).

And we don't need a cluster upgrade just to upgrade the CAPC controller. We manage those component upgrades separately and they can be performed in place without updating the cluster definition.

I see now that code was already there. I don't think it's necessary and in fact it can trigger unnecessary upgrades. I would vote to remove it.

maxdrib: Discussed offline - vsphere's components make sense to validate here because they're using a Cluster Resource Set. However, since CloudStack doesn't, this logic is not necessary here

g-gaston (approved):

pkg/providers/vsphere/vsphere.go
```diff
@@ -199,6 +200,32 @@ func (p *vsphereProvider) UpdateKubeConfig(_ *[]byte, _ string) error {
 	return nil
 }
 
+func (p *vsphereProvider) machineConfigsSpecChanged(ctx context.Context, cc *v1alpha1.Cluster, cluster *types.Cluster, newClusterSpec *cluster.Spec) (bool, error) {
+	machineConfigMap := make(map[string]*v1alpha1.VSphereMachineConfig)
+	for _, config := range p.MachineConfigs(nil) {
```
vivek-koppuru, line 205: We need to clean this up, either remove the parameter from the function or actually use it instead of referencing the values from the struct.

vivek-koppuru (approved):
/lgtm

pkg/providers/vsphere/vsphere.go
maxdrib: The parameter is currently being used in the [snow provider](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/snow/snow.go#L209-L215), but nowhere else I believe. Is this the model all the providers should adopt? It may be worth creating an issue about it

vivek-koppuru: Since this is not an exported method, we have control on what parameters to pass in based on the provider right?

PR #1814, opened 2022-04-11
Fix upgrade workflow

mitalipaygude:
*Issue #, if available:*
Fix upgrade workflow
*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgradecluster.go
+++ b/cmd/eksctl-anywhere/cmd/upgradecluster.go
@@ -105,29 +105,23 @@ func (uc *upgradeClusterOptions) upgradeCluster(ctx context.Context) error {
 		KubeconfigFile: getKubeconfigPath(clusterSpec.Cluster.Name, uc.wConfig),
 	}
 
-	var cluster *types.Cluster
+	var managementCluster *types.Cluster
 	if clusterSpec.ManagementCluster == nil {
-		cluster = &types.Cluster{
-			Name:           clusterSpec.Cluster.Name,
-			KubeconfigFile: getKubeconfigPath(clusterSpec.Cluster.Name, uc.wConfig),
-		}
+		managementCluster = workloadCluster
 	} else {
-		cluster = &types.Cluster{
-			Name:           clusterSpec.ManagementCluster.Name,
-			KubeconfigFile: clusterSpec.ManagementCluster.KubeconfigFile,
-		}
+		managementCluster = clusterSpec.ManagementCluster
 	}
 
 	validationOpts := &validations.Opts{
 		Kubectl:           deps.Kubectl,
 		Spec:              clusterSpec,
 		WorkloadCluster:   workloadCluster,
-		ManagementCluster: cluster,
+		ManagementCluster: managementCluster,
 		Provider:          deps.Provider,
 	}
 	upgradeValidations := upgradevalidations.New(validationOpts)
 
-	err = upgradeCluster.Run(ctx, clusterSpec, cluster, upgradeValidations, uc.forceClean)
+	err = upgradeCluster.Run(ctx, clusterSpec, managementCluster, workloadCluster, upgradeValidations, uc.forceClean)
 	return err
 }
 
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -323,12 +323,17 @@ func (c *ClusterManager) DeleteCluster(ctx context.Context, managementCluster, c
 }
 
 func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster, workloadCluster *types.Cluster, newClusterSpec *cluster.Spec, provider providers.Provider) error {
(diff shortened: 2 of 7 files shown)


pkg/workflows/upgrade.go
```diff
@@ -280,7 +276,7 @@ func (s *pauseEksaAndFluxReconcile) Name() string {
 }
 
 func (s *createBootstrapClusterTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
-	if commandContext.BootstrapCluster != nil && commandContext.BootstrapCluster.ExistingManagement {
+	if commandContext.ManagementCluster != nil && commandContext.ManagementCluster.ExistingManagement {
```
g-gaston, line 279: it might be cleaner if we use `!commandContext.WorkloadCluster.IsSelfManaged()`?

pkg/clustermanager/cluster_manager.go
```diff
@@ -342,12 +342,17 @@ func (c *ClusterManager) DeleteCluster(ctx context.Context, managementCluster, c
 }
 
 func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster, workloadCluster *types.Cluster, newClusterSpec *cluster.Spec, provider providers.Provider) error {
-	currentSpec, err := c.GetCurrentClusterSpec(ctx, workloadCluster, newClusterSpec.Cluster.Name)
+	eksaMgmtCluster := workloadCluster
+	if managementCluster != nil && managementCluster.ExistingManagement {
```
g-gaston, line 346: Why do we need to do this here?
I'm probably missing something, but I would expect the workflow to pass the correct clusters, so ClusterManager doesn't need to to do any checking

mitalipaygude: here we send the mgmt and workload clusters, but what we need inside `UpgradeCluster` are the clusters where we have capi resources and eksa crds. That is how these clusters are used in here

pkg/workflows/upgrade.go
mitalipaygude: `IsSelfManaged` isn't a method we can call here because this Cluster object is different from the cluster type object from the v1alpha1 pkg.

cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -107,15 +107,9 @@ func (uc *upgradeClusterOptions) upgradeCluster(ctx context.Context) error {
 
 	var cluster *types.Cluster
```
g-gaston, line 108: nit: can we rename this to `managementCluster`?

pkg/workflows/create.go
```diff
@@ -349,8 +349,8 @@ func (s *DeleteBootstrapClusterTask) Name() string {
 
 func getManagementCluster(commandContext *task.CommandContext) *types.Cluster {
```
g-gaston, line 350: I there a way we can avoid having this method and just use grab `commandContext.ManagementCluster`?

pkg/clustermanager/cluster_manager.go
```diff
@@ -341,12 +341,17 @@ func (c *ClusterManager) DeleteCluster(ctx context.Context, managementCluster, c
...
```
g-gaston, line 345: Can we add a unit test for this scenario?

g-gaston (commented):
Looks great!

pkg/workflows/create.go
mitalipaygude: This is only used once where we need the eksa management cluster for the ApplyBundles method. Can just substitute the if else here though.

g-gaston: Oh nvm, I get it now, this is cool with me

pkg/clustermanager/cluster_manager.go
mitalipaygude: Added

pkg/clustermanager/cluster_manager_test.go
```diff
@@ -484,6 +484,41 @@ func TestClusterManagerUpgradeWorkloadClusterSuccess(t *testing.T) {
...
+		Name:               mgmtClusterName,
+		ExistingManagement: true,
+	}
+	wCluster := &types.Cluster{
+		Name: workClusterName,
+	}
+
+	tt := newSpecChangedTest(t)
+	tt.mocks.client.EXPECT().GetEksaCluster(tt.ctx, mCluster, mgmtClusterName).Return(tt.oldClusterConfig, nil)
+	tt.mocks.client.EXPECT().GetBundles(tt.ctx, tt.cluster.KubeconfigFile, tt.cluster.Name, "").Return(test.Bundles(t), nil)
```
g-gaston, line 501: For clarity, I think we either use `tt.cluster` or we use `cluster`, but not both
You can also assign `tt.cluster = wCluster` and the just reference `tt.cluster` everywhere

g-gaston (approved):
/approve
/lgtm

PR #1817, opened 2022-04-12
Upgrading eks-a to use capc v0.4.2

maxdrib:
*Issue #, if available:*
Upgrading eks-a to use latest CAPC in anticipation of next release

*Description of changes:*
CAPC version is already updated in build-tools, so we can update it here to use the latest type definitions for CAPC structs

*Testing (if applicable):*
Unit tests pass. CloudStack cluster successfully created without using bundle overrides. Upgrading now using shared network with kube-vip


wongni (approved):
/lgtm

Issue #1033, opened 2022-01-25
Misleading error message in preflight validation


jiayiwang7:
**What happened**:

Adding a non eksa object with custom namespace in cluster spec file when creating a cluster will fail preflight validation with misleading error message:
"VsphereMachineConfig and Cluster objects must have the same namespace specified"

**What you expected to happen**:

Preflight validation should fail w error message like "Unsupported Object Kind: xyzKind"

**How to reproduce it (as minimally and precisely as possible)**:

Take a valid eksa cluster spec, and add the following object to the top:

```yaml
apiVersion: x.y.z.com/v1alpha1
kind: NonEKSAKind
metadata:
  name: non-eksa-object
  namespace: custom-system
spec:
  x: y
```
then create a cluster with that file. 

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

PR #1822, opened 2022-04-12
Parse kind without unmarshalling a spec

wongni:
*Issue #, if available:* #1033

*Description of changes:*
There is a bug in ParseClusterConfigFromContent that the namespace of the first spec is copied over to the second spec when the second spec has the expected kind and its namespace is not defined. 

For example, if the expected kind is 'myKind'
```
---
apiVersion: v1
kind: otherKind
metadata:
  namespace: otherNamespace
  name: otherName
---
apiVersion: v1
kind: myKind
metadata:
   name: myName
---
```
The returned spec's namespace is 'otherNamespace' from the spec one above because the unmarshalled spec for the first yaml remains in the memory. Therefore, the namespace from the first yaml is used for the second spec unintentionally, which is a bug. 

*Testing (if applicable):*
Added and modified unit tests to reveal this bug.


maxdrib (approved):
I don't see any issues with this change, however I'd defer to a member of the eks-a team (@jiayiwang7) to identify if any undesirable side effects may occur. As I recall this change was not so trivial, and is related to https://github.com/aws/eks-anywhere/issues/1033

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -220,6 +220,66 @@ func TestProviderGenerateCAPISpecForCreate(t *testing.T) {
 	test.AssertContentToFile(t, string(md), "testdata/expected_results_main_md.yaml")
 }
 
+func TestNamespaceMismatchBetweenClusterAndDatacenter(t *testing.T) {
```
g-gaston, line 223: Why are we adding tests to the cloudstack provider? If these are supposed to cover the bug fix, can we add them in the v1alpha1 package?

pkg/api/v1alpha1/cluster.go
```diff
@@ -250,18 +250,28 @@ func ParseClusterConfig(fileName string, clusterConfig KindAccessor) error {
...
+			continue
 		}
+		return yaml.UnmarshalStrict([]byte(c), clusterConfig)
 	}
 
 	return fmt.Errorf("yamlop content is invalid or does not contain kind %s", clusterConfig.ExpectedKind())
 }
 
+func parseKind(spec string) string {
+	kindRegex := regexp.MustCompile(`(?m)^kind: (.*)`)
```
g-gaston, line 263: I would like to avoid a regex if possible

One possible solution would be to first unmarshall the object to a simple struct that just has a `Kind` string member. Example: https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config_manager.go#L126
You can reset that object on each iteration so it doesn't carry any information from the previous one

g-gaston (commented):
Thanks for finding this!

pkg/providers/cloudstack/cloudstack_test.go
wongni: Done

pkg/api/v1alpha1/cluster.go
wongni: Done

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -220,6 +220,32 @@ func TestProviderGenerateCAPISpecForCreate(t *testing.T) {
 	test.AssertContentToFile(t, string(md), "testdata/expected_results_main_md.yaml")
 }
 
+func TestProviderGenerateDeploymentFileWithNonEksaResources(t *testing.T) {
```
g-gaston, line 223: Curious, what is this test for? It doesn't seem like the cloudstack code has changes

g-gaston (approved):
/lgtm

pkg/providers/cloudstack/cloudstack_test.go
wongni: Pushed.

maxdrib (approved):

PR #1823, opened 2022-04-12
Air Gapped experience consolidation proposal

g-gaston:
*Description of changes:*
In the last few months, we have built multiple features aiming in the same direction but they organically grew in different directions. This design aims to combine them into a more cohesive solution.

### DIFF
--- a/designs/air-gapped.md
+++ b/designs/air-gapped.md
@@ -1,0 +1,97 @@
+# Air Gapped user experience consolidation
+
+## Introduction
+
+**Problem:** we offer a set features that allow users to run EKS-A cluster in disconnected environments, but they lack cohesion.
+
+Overtime, we have developed a particular set of features and tools that, even though they were aimed to solve independent problems, they all converge to offer to support for running EKS-A in environments without Internet connection.
+As a result of their independent development, their interfaces and behaviors are not in sync, which makes the user experience not ideal.
+
+This document aims to solve that, learning from the experience gained building such tools and offering a more cohesive and simple to use solution, without drastically changing the backbone of those features and how they interact with each other.
+
+### Tenets
+
+* ****Simple:**** minimal interface, minimum number of manual steps.
+
+### Goals and Objectives
+
+As a EKS-A cluster administrator I want to:
+
+* Manage clusters in an environment without Internet connection.
+* Populate my environment with the necessary dependencies without having to worry about what those dependencies are and with minimal effort.
+* Interact with EKS-A clusters in air-gapped environments in the same way I interact with any other EKS-A cluster.
+
+### Statement of Scope
+
+**In scope**
+* Download dependencies and package them.
+* Populate disconnected environments with dependencies without Internet connection.
+
+**Not in scope**
+* Setup additional infrastructure to support disconnected environments (eg. private registries).
+
+**Future scope**
+* Package dependencies selectively.
+* Populate disconnected environments with dependencies with Internet connection.
+ 
+## Current state
+In order to support disconnected environments, users currently have to:
+1. Provision an OCI registry.
+1. Run `eksctl anywhere download artifacts` to download manifests to disk.
+1. Run `eksctl anywhere download images` to download images and charts to disk.
+1. Run `eksctl anywhere import images` to push container images and charts to a registry.
+1. Configure their cluster spec file to point to the registry 
+1. Run all cluster commands (`create`, `upgrade`, etc.) with the `--bundles-override` flag pointing to the `Bundles` manifest downloaded in the first step.
+
+## Overview of Solution
+1. Users will provide an OCI registry.
+1. We will store all dependencies (container images, helm charts and yaml manifests) in that registry.
+1. If a cluster config with a [`registryMirrorConfiguration`](https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L51) is provided, the CLI and the rest of components will always pull them from it.
+No extra flags/configuration will be needed and the user will interact with this cluster as with any other one.
+
+## Solution Details
+
+### Populating the registry
+
+In order to populate the registry, we will offer two commands:
+```sh
+eksctl anywhere export artifacts --output artifacts.tar
+eksctl anywhere import artifacts --input artifacts.tar --registry myregistry.com
+```
+
+The first one downloads all dependencies (yaml manifests, images and helm charts). The only command argument here is the destination file. This command will create a tarball containing all 3 types of artifacts in that location.
+The second one unpackages the tarball created in the first command, reads the packaged `Bundles` manifest and imports the referenced dependencies to a registry. The input here will be the artifacts tarball, the registry endpoint and the registry credentials (provided through env vars).
+
+This two commands can be expanded in the future to add more capabilities like selectively exporting dependencies (eg. only for one kubernetes version or for only one provider).
+
+### Pre packaged dependencies
+
+A specific version of the CLI tied to one `Bundles` manifest will always produce the exact same artifacts tarball.
+This means that we can prepackage the dependencies, store them in a public bucket and reference the tarball in the `Release` manifest.
+This simplifies the experience for users who are interested in the default dependencies bundle.
+
+## Implementation
+
+The proposed design can be implemented in 4 incremental phases:
+1. Add the two new commands but only push images and charts to the registry. This will require to keep using the `--bundles-override` pointing to a `Bundles` manifest in disk. This is an incremental improvement over the current state of `download images` so work should be minimal.
+2. Push manifests to OCI registry and add the capability to the CLI to download them.
+3. Start storing our default manifests in public ECR as opposed to Cloudfront + S3. This unifies even more the behavior of connected and disconnected environments.
+4. Start packaging dependencies and serving them from the `Release` manifest.
+
+## User Experience
+
+### Deprecated commands
+This proposal deprecates several commands:
+* `download artifacts`
+* `download images`
+* `import-images`
+* `import images`
+
+We should inform users in the next release notes and keep the old commands in the codebase for at least one more release, printing a warning when executed.
+
+## Testing
+
+We should add at least one E2E test for the whole flow:
+1. Download artifacts to disk
+1. Import artifacts to a private registry
+1. Create a cluster with `registryMirrorConfiguration` (ideally in an environment without external Internet connection).


jaxesn (commented):
I like this!

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+
+An specific version of the CLI tied to one `Bundles` manifest will always produce the exact same artifacts tarball.
+This means that we can prepackage the dependencies, store them in a public bucket and reference the tarball in the `Release` manifest.
+This simplifies the experience for users who are interested in the default dependencies bundle.
+
+## Implementation
+
+The proposed design can be implemented in 4 incremental phases:
+1. Add the two new commands but only push images and charts to the registry. This will require to keep using the `--bundles-override` pointing to a `Bundles` manifest in disk. This is an incremental improvement over the current state of `download images` so work should be minimal.
+2. Push manifests to OCI registry and add the capability to the CLI to download them.
```
jaxesn, line 77: @abhay-krishna this is something interesting we should test out to see our options. https://oras.land/cli/1_pushing/ could be interesting for this.

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+In order to populate the registry, we will offer two commands:
+```sh
+eksctl anywhere export artifacts --output artifacts.tar
+eksctl anywhere import artifacts --input artifacts.tar --registry myregistry.com
+```
+
+The first one downloads all dependencies (yaml manifests, images and helm charts). The only command argument here is the destination file. This command will create a tarball containing all 3 types of artifacts in that location.
+The second one unpackages the tarball created in the first command, reads the packaged `Bundles` manifest and imports the referenced dependencies to a registry. The input here will be the artifacts tarball, the registry endpoint and the registry credentials (provided through env vars).
+
+This two commands can be expanded in the future to add more capabilities like selectively exporting dependencies (eg. only for one kubernetes version or for only one provider).
```
jaxesn, line 65: Def a future thing, but I see this being important before too long.

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+* Setup additional infrastructure to support disconnected environments (eg. private registries).
+
+**Future scope**
+* Package dependencies selectively.
+* Populate disconnected environments with dependencies with Internet connection.
+ 
+## Current state
+In order to support disconnected environments, users currently have to:
+1. Provision an OCI registry.
+1. Run `eksctl anywhere download artifacts` to download manifests to disk.
```
TerryHowe, line 40: Late to the party here, does artifacts include OVAs?

g-gaston: The current implementation of `download commands` doesn't download them, it only downloads yaml manifests

The new command I'm proposing, `export artifacts`, doesn't either, at least in this first proposal.

But it could definitely be expanded to include them. If we think we are gonna keep adding more types of artifacts, I was thinking that the tarball could contain a manifest with all the "exported" artifacts.

We would just need to define what's the behavior for `export artifacts`: do we upload them to a vcenter? or just untar them to disk?. But I would leave that for a different proposal when the times comes.

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+This means that we can prepackage the dependencies, store them in a public bucket and reference the tarball in the `Release` manifest.
+This simplifies the experience for users who are interested in the default dependencies bundle.
+
+## Implementation
+
+The proposed design can be implemented in 4 incremental phases:
+1. Add the two new commands but only push images and charts to the registry. This will require to keep using the `--bundles-override` pointing to a `Bundles` manifest in disk. This is an incremental improvement over the current state of `download images` so work should be minimal.
+2. Push manifests to OCI registry and add the capability to the CLI to download them.
+3. Start storing our default manifests in public ECR as opposed to Cloudfront + S3. This unifies even more the behavior of connected and disconnected environments.
+4. Start packaging dependencies and serving them from the `Release` manifest.
```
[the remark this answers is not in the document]
g-gaston: We will keep the flag but it won't be necessary for air gapped

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Populating the registry
+
+In order to populate the registry, we will offer two commands:
+```sh
+eksctl anywhere export artifacts --output artifacts.tar
+eksctl anywhere import artifacts --input artifacts.tar --registry myregistry.com
+```
+
+The first one downloads all dependencies (yaml manifests, images and helm charts). The only command argument here is the destination file. This command will create a tarball containing all 3 types of artifacts in that location.
```
mrajashree, line 62: will this download the artifacts for individual components (CAPI/CAPV etc) across all k8s versions that the cli release in use supports?

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+The second one unpackages the tarball created in the first command, reads the packaged `Bundles` manifest and imports the referenced dependencies to a registry. The input here will be the artifacts tarball, the registry endpoint and the registry credentials (provided through env vars).
+
+This two commands can be expanded in the future to add more capabilities like selectively exporting dependencies (eg. only for one kubernetes version or for only one provider).
+
+### Pre packaged dependencies
+
+An specific version of the CLI tied to one `Bundles` manifest will always produce the exact same artifacts tarball.
+This means that we can prepackage the dependencies, store them in a public bucket and reference the tarball in the `Release` manifest.
+This simplifies the experience for users who are interested in the default dependencies bundle.
```
mrajashree, line 71: not sure I follow this part...is this so that the new `export artifacts` command has a single location to download everything from?

designs/air-gapped.md
mrajashree, line 79: eks-a will use the registry for downloading images and charts based on cluster config spec. But for it to know where to pull the manifests of CAPI & providers from, will we have to update the `Release` manifest to point to the private oci registry?

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+1. If a cluster config with a [`registryMirrorConfiguration`](https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L51) is provided, the CLI and the rest of components will always pull them from it.
+No extra flags/configuration will be needed and the user will interact with this cluster as with any other one.
+
+## Solution Details
+
+### Populating the registry
+
+In order to populate the registry, we will offer two commands:
+```sh
+eksctl anywhere export artifacts --output artifacts.tar
```
mrajashree, line 58: how about `download artifacts` and `upload artifacts`?

mrajashree (commented):
looks great

designs/air-gapped.md
g-gaston: Yeah, exactly

designs/air-gapped.md
g-gaston: No. This for users to be able to skip running `export artifacts`. They can just download the provided tarball, ship it to their air gapped environment and run `import artifacts`.

We would generate such prepackage tarball using `export artifacts`

designs/air-gapped.md
g-gaston: No, we would do something similar to what we do today for helm. We would swap the host in runtime if the cluster config has a `registryMirrorConfig`

designs/air-gapped.md
g-gaston: I look at it from the other end, I think export/import is more generic.
Download/upload implies there is a network connection. Which for now, it is true for all 3 artifacts. But this might not be true for other types of artifacts we might add in the future (imagine something that we just store in disk in a particular location).

The idea behind this proposal is to be able to be flexible enough to accommodate future expansions. Hence, my preference for a more generic export/import.

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+eksctl anywhere import artifacts --input artifacts.tar --registry myregistry.com
+```
+
+The first one downloads all dependencies (yaml manifests, images and helm charts). The only command argument here is the destination file. This command will create a tarball containing all 3 types of artifacts in that location.
+The second one unpackages the tarball created in the first command, reads the packaged `Bundles` manifest and imports the referenced dependencies to a registry. The input here will be the artifacts tarball, the registry endpoint and the registry credentials (provided through env vars).
```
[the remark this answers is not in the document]
g-gaston: What do you mean by validating? Checking that all the artifacts referenced in the `Bundle` are present in the tarball? That might not be feasible for container images, since they are all packaged together. What we can validate is that all referenced images end up in the registry. But that validation is kind of implied since the `push` action would fail in case one image is not present.

What I was thinking is that I might backtrack on reading from the `Bundles` and just include an extra "artifacts list" in the tarball, referencing all included artifacts. That would allow us to add options to selectively include/exclude artifacts in the future.

designs/air-gapped.md
```diff
@@ -0,0 +1,97 @@
...
+An specific version of the CLI tied to one `Bundles` manifest will always produce the exact same artifacts tarball.
+This means that we can prepackage the dependencies, store them in a public bucket and reference the tarball in the `Release` manifest.
+This simplifies the experience for users who are interested in the default dependencies bundle.
+
+## Implementation
+
+The proposed design can be implemented in 4 incremental phases:
+1. Add the two new commands but only push images and charts to the registry. This will require to keep using the `--bundles-override` pointing to a `Bundles` manifest in disk. This is an incremental improvement over the current state of `download images` so work should be minimal.
+2. Push manifests to OCI registry and add the capability to the CLI to download them.
+3. Start storing our default manifests in public ECR as opposed to Cloudfront + S3. This unifies even more the behavior of connected and disconnected environments.
```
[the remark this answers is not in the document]
g-gaston: Using a "yaml" mediatype (either reusing `application/vnd.oci.yaml-sample.config.v3+yaml` or just creating our own). I haven't looked deeply into the technical details, but I know it's possible to use an OCI registry to store "random" files.

I don't think it's orthogonal though. At least not to simplifying the UX. By doing this, users don't need to worry anymore about having to pass the `--bundles-override` flag and about having to have all yaml manifests stored in local disk.

If what you are suggesting is that we could possibly use a different storage mechanism and achieve the same UI, yeah that's true. But that would involve having two dependencies instead of one. Unless you have a better idea which I would be happy to talk about.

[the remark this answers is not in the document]
chrisdoherty4: I've seen it done before. I'm more saying just because we can, doesn't mean we should. OCI is about image formats, storing something that doesn't contain an image feels misleading. 

From the users point of view, they interact with 2 endpoints with respect to an image and a bundle. Is that really a problem?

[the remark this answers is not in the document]
g-gaston: I don't agree with the fact that OCI is only about images formats, at least not anymore.
 
It might have started with that, but now a days we see other types of assets being stored in registries (mostly related with packaged software). Helm charts is one example.

There is even a [repo](https://github.com/opencontainers/artifacts) in the open containers org about this. You can even see SBOM mentioned there, which in a way, is _similar_ to our `Bundle`.

I could be in the wrong here but I don't think this is a "we do it because we can". At least in my opinion, it is the right solution.

As far as 2 endpoint vs 1 being a problem: that's tricky question. What is really a problem? That depends on the definition of problem, which is driven by previous experiences, expectations and a bit of personal preference.

Answering with mine, yes I think it is a problem. It's not about providing two flags, it's about managing two dependencies. That's double the work and double the things that can go wrong. We already got instances where managing a private registry was "too much effort".

Do you have any other concern (aside from being misleading) about using the registry? Technical or other.

Also, what alternative would you propose instead?

designs/air-gapped.md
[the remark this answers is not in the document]
jiayiwang7: wait how will this implementation prevents us from using `bundles-override` for dev when the bundle release keeps changing?

designs/air-gapped.md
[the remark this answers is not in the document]
chrisdoherty4: > I don't agree with the fact that OCI is only about images formats, at least not anymore.

You're right, its expanded way beyond what I remember having combed through a bunch of resources.

How wide-spread is the support for custom media types? Is there an existing media type we can leverage or would we need to define our own?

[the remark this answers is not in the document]
chrisdoherty4: Scratch that "existing types" question. I see in an earlier reply some commentary.

designs/air-gapped.md
vivek-koppuru: Are you suggesting that download/export is the difference between manifests/tarball? That sounds a bit confusing to me too, unless we are planning to deprecate?

designs/air-gapped.md
vivek-koppuru: If we have the ability to directly export to a registry in the future in one command, what would you suggest calling that command? Or is the plan to just have users run both?

designs/air-gapped.md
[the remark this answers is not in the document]
vivek-koppuru: Yea I was more referring to running import based on what is referenced in `Bundle` and ensuring that we can find it in the tarball. 

Another option is that we have the tarball created as part of our build infra and just download that? Just throwing ideas.

chrisdoherty4 (approved):

designs/air-gapped.md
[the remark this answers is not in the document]
g-gaston: Resolved offline

designs/air-gapped.md
g-gaston: Yeah, I'm proposing to deprecate the old ones

designs/air-gapped.md
g-gaston: That would depend on use feedback.

Two commands are necessary to support "truly disconnected" environments where there not one machine that can act as a bridge between the internet and the registry.

I'm not sure if it's worth the effort since they can just chain the commands with `&&`, but if there are enough folks who want one single command, we could build it. As far as the name, I think we can wait until we decide to build it. But maybe: `populate artifacts`, `export-import artifacts`...

chrisdoherty4:
Just a thought for the future - it would be nice to build a list of registry implementations that support custom OCI artifacts so we can quickly respond to customers in the event they have preferred tool sets that include a registry.

designs/air-gapped.md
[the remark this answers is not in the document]
g-gaston: Yeah that's what I was thinking. But we could leave this decision for the implementation. Another option more future proof is to include a "artifact manifest" in the tarball that references all included artifacts. That makes it more flexible if one day we want to offer "selectively including artifacts". Like opting out from some provider images or opt in for curated packages.

About prebuilding the tarball and offering it to the public, yeah agreed, that's what I proposed in the `Pre packaged dependencies` section.

vivek-koppuru (approved):

PR #1827, opened 2022-04-13
Use `FluxConfig` internally instead of `GitOpsConfig`

danbudris:
*Issue #, if available:*
[Project](https://github.com/aws/eks-anywhere/projects/9)
[Issue](https://github.com/aws/eks-anywhere/issues/1826)
[Design](https://github.com/aws/eks-anywhere/blob/ed59f2c747c3afe394c5115f0046d3c8f3a85e66/designs/generic-git-provider-gitops.md)

*Description of changes:*
This change lays the groundwork for further development of [Bring Your Own Git.](https://github.com/aws/eks-anywhere/blob/ed59f2c747c3afe394c5115f0046d3c8f3a85e66/designs/generic-git-provider-gitops.md)

The `GitOpsConfig` type is being deprecated in favor of the new `FluxConfig` type, which makes it easy to support multiple Providers (e.g. GitHub, BYO Git, etc). While the feature is in development, usage of a `FluxConfig` as a `GitOpsRef`[ will be feature gated](https://github.com/aws/eks-anywhere/pull/1827/commits/cb74219e736fe0adcad42223a7fee56936d14a60), even though we'll be using the `FluxConfig` internally for all GitOps related workflows end-to-end.

These changes set up the cluster to [automatically generate a `FluxConfig` from a user-provided `GitOpsConfig`](https://github.com/aws/eks-anywhere/pull/1827/commits/9380c11bea0b185401a00ea678efc88f177e394a#diff-c0ac5c9b3e6dfc86e94a13471d3207ef5d2a305a6705039c8951e1ca99d73f63R48), and replace the usage of the `GitOpsConfig` type with the `FluxConfig` type throughout EKS Anywhere; additionally, there are a number of supporting changes in order to allow this, such as making sure that [defaults are set for `FluxConfig`](https://github.com/aws/eks-anywhere/pull/1827/commits/133080581dc353246b97ea394704d6f1f0bb23bc), that [it's marshaled properly](https://github.com/aws/eks-anywhere/pull/1827/commits/2adaaf10a8aef17863d16d5904dffe92e6f6e212), and that [the cluster config deep copy methods support it](https://github.com/aws/eks-anywhere/pull/1827/commits/133080581dc353246b97ea394704d6f1f0bb23bc). 

We will continue to support the `GitOpsConfig` for the time being (deprecation schedule TBD...thinking +2 minor versions?), transparently converting it to a `FluxConfig` internally when provided by a user. Once deprecated the type and the conversion will be removed.

Additional refactoring is going to follow soon after this initial groundwork, including refactoring the E2E tests to use the `FluxConfig` while also testing cases with an existing, user-provided `GitOpsConfig`. 

**UX**
After this PR merges:
- A user can still provide a `GitOpsConfig` in their cluster configuration.
- The user-provided `GitOpsConfig` will be transparently converted to a `FluxConfig` for internal use by the system.
- When a cluster configuration is marshaled to file, the originally provided `GitOpsConfig` will be written, making this internal conversion transparent to the user.

After this feature is GA:
- A user can provide either a `GitOpsConfig` or a `FluxConfig` via their cluster configuration.
- A user-provided `GitOpsConfig` will be transparently convereted a a `FluxConfig` for internal use
- Deprecation warnings will be given to the user regarding the `GitOpsConfig`

Final UX:
- A user may ONLY provide a `FluxConfig` via their cluster configuration.
- Upgrade supports a path to automatically converting a a `GitOpsConfig` to a `FluxConfig` when upgrading between a minor version in which `GitOpsConfig` exists and a version in which it is fully deprecated

**Deprecation Path**
Deprecation path is open to conversation; I was thinking +2 minor versions we remove the `GitOpsConfig` and provide any tooling necessary for a user to easily convert. 

In the mean time, once this feature is ready to go GA and the feature gate is ready to be removed, we'll add deprecation warnings to the `GitOpsConfig` and make note of the deprecation path in the `v0.9.0` release notes.

*Testing (if applicable):*
`make`
`eksctl anywhere create cluster` with a `GitOpsConfig`; modify the cluster via flux.
`eksctl anywhere upgrade cluster` with a `GitOpsConfig`, scaling and upgrading the cluster.
run E2E flux tests

Currently, since we're transparently converting `GitOpsConfig` to `FluxConfig` internally, the current E2E tests work without modification for this change.  Once we introduce the `FluxConfig` externally, we'll add additional E2E tests which test it; then, once `GitOpsConfig` is deprecated fully, we'll remove all the tests that use `GitOpsConfig`.

 E2e tests are being worked on in a separate PR; this currently functions with our existing E2E test suite given the transparent conversion. Once we add the BYO Git functionality we will add e2e tests for the FluxConfig case with both GitHub and Git provider types.

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/yaml"
 
+	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 )
@@ -572,11 +573,17 @@ func validateGitOps(clusterConfig *Cluster) error {
 	if gitOpsRef == nil {
 		return nil
 	}
-	if gitOpsRef.Kind != GitOpsConfigKind {
+
+	if gitOpsRef.Kind == FluxConfigKind && !features.IsActive(features.GenericGitProviderSupport()) {
+		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported")
+	}
+
+	if gitOpsRef.Kind != GitOpsConfigKind && gitOpsRef.Kind != FluxConfigKind {
 		return errors.New("only GitOpsConfig Kind is supported at this time")
 	}
+
 	if gitOpsRef.Name == "" {
-		return errors.New("GitOpsConfig name can't be empty; specify a valid name for GitOpsConfig")
+		return errors.New("GitOpsRef name can't be empty; specify a valid GitOpsConfig name")
 	}
 	return nil
 }
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -574,12 +574,20 @@ func validateGitOps(clusterConfig *Cluster) error {
 		return nil
 	}
 
-	if gitOpsRef.Kind == FluxConfigKind && !features.IsActive(features.GenericGitProviderSupport()) {
-		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported")
+	gitOpsRefKind := gitOpsRef.Kind
+	fluxConfigActive := features.IsActive(features.GenericGitProviderSupport())
+
+	if gitOpsRefKind == FluxConfigKind && !fluxConfigActive {
+		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported; " +
+			"to use this experimental feature, please set the environment variable GENERIC_GIT_PROVIDER_SUPPORT to true")
 	}
 
-	if gitOpsRef.Kind != GitOpsConfigKind && gitOpsRef.Kind != FluxConfigKind {
+	if gitOpsRefKind != GitOpsConfigKind && !fluxConfigActive {
 		return errors.New("only GitOpsConfig Kind is supported at this time")
+	}
+
+	if gitOpsRefKind != GitOpsConfigKind && gitOpsRefKind != FluxConfigKind {
+		return errors.New("only GitOpsConfig or FluxConfig Kind are supported at this time")
 	}
 
 	if gitOpsRef.Name == "" {
--- a/pkg/api/v1alpha1/gitopsconfig_types.go
+++ b/pkg/api/v1alpha1/gitopsconfig_types.go
@@ -110,7 +110,6 @@ func (c *GitOpsConfig) ConvertToFluxConfig() *FluxConfig {
 			},
 		},
 	}
-	config.ObjectMeta.Name = c.Name
 	return config
 }
 
--- a/pkg/api/v1alpha1/gitopsconfig_types.go
+++ b/pkg/api/v1alpha1/gitopsconfig_types.go
@@ -86,6 +86,30 @@ func (c *GitOpsConfig) ExpectedKind() string {
 	return GitOpsConfigKind
 }
 
+func (c *GitOpsConfig) ConvertToFluxConfig() *FluxConfig {
+	if c == nil {
+		return nil
+	}
+	config := &FluxConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       FluxConfigKind,
+			APIVersion: c.APIVersion,
+		},
+		Spec: FluxConfigSpec{
+			SystemNamespace:   c.Spec.Flux.Github.FluxSystemNamespace,
+			Branch:            c.Spec.Flux.Github.Branch,
+			ClusterConfigPath: c.Spec.Flux.Github.ClusterConfigPath,
+			Github: &GithubProviderConfig{
+				Owner:      c.Spec.Flux.Github.Owner,
+				Repository: c.Spec.Flux.Github.Repository,
+				Personal:   c.Spec.Flux.Github.Personal,
+			},
+		},
+	}
+	config.ObjectMeta.Name = c.Name
+	return config
+}
+
 func (c *GitOpsConfig) ConvertConfigToConfigGenerateStruct() *GitOpsConfigGenerate {
 	namespace := defaultEksaNamespace
 	if c.Namespace != "" {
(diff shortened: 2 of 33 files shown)


pkg/cluster/gitops.go
```diff
@@ -15,29 +15,16 @@ func gitOpsEntry() *ConfigManagerEntry {
...
-			func(c *Config) error {
-				if c.GitOpsConfig != nil {
-					c.GitOpsConfig.SetDefaults()
-				}
-				return nil
-			},
+			setGitOpsDefaults,
 			SetDefaultFluxGitHubConfigPath,
+			setFluxDefaults,
+			SetDefaultFluxConfigPath,
```
g-gaston, line 21: nit: we don't need to register the defaults for flux here, those will be already added by the flux entry

pkg/cluster/gitops.go
```diff
@@ -15,29 +15,16 @@ func gitOpsEntry() *ConfigManagerEntry {
...
-					if err := validateSameNamespace(c, c.GitOpsConfig); err != nil {
-						return err
-					}
-				}
-				return nil
-			},
+			validateGitOps,
+			validateGitOpsNamespace,
+			validateFlux,
+			validateFluxNamespace,
```
g-gaston, line 27: Same here, these are already added by the flux entry

pkg/cluster/gitops_test.go
```diff
@@ -0,0 +1 @@
+package cluster
```
g-gaston, line 1: What's this for?

g-gaston (commented):
Looks good, minor comments

pkg/api/v1alpha1/cluster.go
```diff
@@ -572,11 +573,17 @@ func validateGitOps(clusterConfig *Cluster) error {
...
 		return nil
 	}
-	if gitOpsRef.Kind != GitOpsConfigKind {
+
+	if gitOpsRef.Kind == FluxConfigKind && !features.IsActive(features.GenericGitProviderSupport()) {
+		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported")
+	}
+
+	if gitOpsRef.Kind != GitOpsConfigKind && gitOpsRef.Kind != FluxConfigKind {
 		return errors.New("only GitOpsConfig Kind is supported at this time")
```
jiayiwang7, line 582: should we also mention `FluxConfig` is supported under feature gate?

pkg/api/v1alpha1/gitopsconfig_types.go
```diff
@@ -86,6 +86,34 @@ func (c *GitOpsConfig) ExpectedKind() string {
...
+			Branch:            c.Spec.Flux.Github.Branch,
+			ClusterConfigPath: c.Spec.Flux.Github.ClusterConfigPath,
+			Github: &GithubProviderConfig{
+				Owner:      c.Spec.Flux.Github.Owner,
+				Repository: c.Spec.Flux.Github.Repository,
+				Personal:   c.Spec.Flux.Github.Personal,
+			},
+		},
+	}
+	config.ObjectMeta.Name = c.Name
```
jiayiwang7, line 113: why do we need this?

pkg/api/v1alpha1/cluster.go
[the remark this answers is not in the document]
danbudris: I'm waiting until we remove the feature gate to update this error message

pkg/cluster/gitops_test.go
danbudris: oh yeah that was the start of some test I was adding that wasn't supposed to be part of this, will remove

pkg/api/v1alpha1/cluster.go
danbudris: yeah, I was going to wait to add it until we removed the feature gate to avoid potential confusion, but can update it to indicate that it's available under the feature flag

pkg/api/v1alpha1/gitopsconfig_types.go
danbudris: good catch, I can remove that -- in my first pass at making this work, I converted the type and then used the `FluxConfig` type on the cluster as well instead of marshalling the GitOpsConfig, this is left over from that. Will remove.

pkg/cluster/gitops.go
danbudris: ah ok I see, because the FluxConfig entry here https://github.com/aws/eks-anywhere/blob/c5e281d6a1a008fc4c0efb095d22a29e4e402a3d/pkg/cluster/default_manager.go#L12

is processed after the GitOps entry, we'll have the FluxConfig present in the spec and it'll get defaulted and vaildated

ok thanks!

pkg/api/v1alpha1/cluster.go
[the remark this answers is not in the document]
g-gaston: But my vote would be for the code to align with the error message: if the `if` clause accepts both kinds, then the error message should say so. If not, then I would change the `if` condition to only support `GitOpsConfig` for now.

pkg/cluster/gitops.go
danbudris: https://github.com/aws/eks-anywhere/pull/1827/commits/e0c1cea03b3576d9edf4a990155eec4410d6222d

pkg/cluster/gitops.go
danbudris: https://github.com/aws/eks-anywhere/pull/1827/commits/e0c1cea03b3576d9edf4a990155eec4410d6222d

pkg/api/v1alpha1/gitopsconfig_types.go
danbudris: https://github.com/aws/eks-anywhere/pull/1827/commits/ee9424dc17ca74abb8f530dc7c1fb36d1373f901

pkg/api/v1alpha1/cluster.go
danbudris: https://github.com/aws/eks-anywhere/pull/1827/commits/89624bcff876de204a0cfbb8433650548caa81b0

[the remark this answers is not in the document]
danbudris: https://github.com/aws/eks-anywhere/pull/1827/commits/89624bcff876de204a0cfbb8433650548caa81b0

PR #1828, opened 2022-04-13
Default to fetching eks-d release from url if it can't be fetched from the cluster

taneyland:
*Issue #, if available:*

*Description of changes:*
Fetching the EKS-D release manifest could return a number of different errors if those release objs aren't present on the cluster. We should default to fetching the EKS-D object from the URL & then applying it to the cluster after upgrade.

*Testing (if applicable):*
Tilt, cluster created with 0.7 and upgraded with 0.8.1

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mrajashree (approved):
/lgtm

PR #1836, opened 2022-04-14
Marshal FluxConfig if present

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1826

*Description of changes:*
Make sure that we marshal flux config if it is present. 

If both a FluxConfig and a GitOpsConfig are present, use the GitOpsConfig.
This will allow us to use the FluxConfig internally while ensuring we marshal the expected cluster configuration for the user.

*Testing (if applicable):*


pkg/clustermarshaller/clustermarshaller.go
```diff
@@ -28,6 +28,10 @@ func MarshalClusterSpec(clusterSpec *cluster.Spec, datacenterConfig providers.Da
 		marshallables = append(marshallables, clusterSpec.GitOpsConfig.ConvertConfigToConfigGenerateStruct())
 	}
 
+	if clusterSpec.FluxConfig != nil {
```
g-gaston, line 31: Should we have an if/else to only marshal either `GitOpsConfig` or `FluxConfig`?

g-gaston, line 31: Maybe a add a unit test fir this case?

danbudris: yeah, I added a check that throws an error if both are present, and filed at ticket to add a higher-level check for mutual exclusivity. Additionally, added a test which checks that if both are present we throw an error.

danbudris: add a unit test which tests that an error is thrown in both are present

danbudris: added a unit test for this specific case as well

g-gaston (approved):

PR #1841, opened 2022-04-14
Upgrade go version

a-cool-train:
*Issue #, if available:*

*Description of changes:*

Upgrade go version to 1.17


Makefile
```diff
@@ -320,7 +320,7 @@ cluster-controller-images: cluster-controller-binaries $(ORGANIZE_BINARIES_TARGE
 
 
 .PHONY: generate-attribution
-generate-attribution: GOLANG_VERSION ?= "1.16"
+generate-attribution: GOLANG_VERSION ?= "1.17"
```
jaxesn, line 323: I think you can just remove this line now that we have defined above.

Makefile
```diff
@@ -20,10 +20,11 @@ SHELL := /bin/bash
...
 GIT_VERSION?=$(shell git describe --tag)
 GIT_TAG?=$(shell git tag -l --sort -v:refname | head -1)
-GOLANG_VERSION?="1.16"
-GO ?= $(shell source ./scripts/common.sh && build::common::get_go_path $(GOLANG_VERSION))/go
+GOLANG_VERSION?="1.17"
+GO_VERSION ?= $(shell source ./scripts/common.sh && build::common::get_go_path $(GOLANG_VERSION))
+GO ?= $(GO_VERSION)/go
 GO_TEST ?= $(GO) test
-
+export PATH := $(GO_VERSION):$(PATH)
```
jaxesn, line 27: Where is go getting called that this is necessary? somewhere in one of the scripts this makefile is calling?

go.mod
```diff
@@ -28,22 +28,118 @@ require (
...
+	k8s.io/client-go v0.23.0
+	sigs.k8s.io/cluster-api v1.1.3
 	sigs.k8s.io/cluster-api-provider-vsphere v1.0.1
 	sigs.k8s.io/cluster-api/test v1.0.0
-	sigs.k8s.io/controller-runtime v0.10.3
+	sigs.k8s.io/controller-runtime v0.11.1
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
```
jaxesn, line 47: I am assuming this is a change in go 1.17 that they want to have the indirects in the go.mod?

scripts/make_attribution.sh
```diff
@@ -44,12 +44,13 @@ source "$REPO_ROOT/scripts/attribution_helpers.sh"
 function build::attribution::generate(){
     cd $REPO_ROOT
     build::common::use_go_version "$GOLANG_VERSION"
-    go mod vendor
+    local -r GO=$(build::common::get_go_path "$GOLANG_VERSION")/go
```
jaxesn, line 47: We may actually need to leave this one how it was and rely on the default go version installed.  The go-licenses call can get weird if the currently configured go is not the same version that go-licenses was installed with. On builder base we only install it with the default version of go, not all the other versions.

jaxesn (changes requested):
This looks pretty good. I *think* well need to leave the attribution/licenses code alone and let it keep using 1.16 for the time being.

go.mod
a-cool-train: Yes that's correct.

scripts/make_attribution.sh
a-cool-train: Gotcha. Leaving this as is.

Makefile
a-cool-train: One thing I observed was that the go1.17 was not in the path and because of this go1.17 was failing hence the need to add it to the path.

jaxesn: I removed this in my commit, we can put it back if we need it, but since we are calling go via the makefile var im not sure why would need this...

jaxesn:
I have made a commit to attempt to fix the attribution generation.  I have also included a few changes from the common scripts we have made in the other repos that clean the get_go_path func up.

jaxesn (approved):
This looks good to me from a build perspective. It would be best just to have someone validate the go files changes. I'll lgtm tomorrow if no one else reviews.

vivek-koppuru (approved):
This looks good to me too as long as it builds and passes e2e

g-gaston (commented):
/lgtm

PR #1855, opened 2022-04-15
Feature list packages (#1622)

a-cool-train:
*Description of changes:*
Implement curated packages discovery.


cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+var listPackagesCommand = &cobra.Command{
+	Use:          "packages",
+	Short:        "Lists curated packages available to install",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := validateKubeVersion(lpo.kubeVersion, lpo.source); err != nil {
```
g-gaston, line 36: Should we validate too that `source` has a valid value?

pkg/executables/kubectl.go
```diff
@@ -780,6 +781,10 @@ func WithArgs(args []string) KubectlOpt {
 	return appendOpt(args...)
 }
 
+func WithArg(arg string) KubectlOpt {
```
g-gaston, line 784: Why don't we define a new option/s for the arguments we need?

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,38 @@
+package curatedpackages
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
```
g-gaston, line 9: nit: I would name this `packagesv1`. It follows the pattern we have been trying to move towards when importing api struct packages.

pkg/curatedpackages/package.go
```diff
@@ -0,0 +1,38 @@
...
+	tabWidth = 8
+	padding  = 0
+	padChar  = '\t'
+	flags    = 0
+)
+
+func DisplayPackages(packages []api.BundlePackage) {
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, minWidth, tabWidth, padding, padChar, flags)
+	defer w.Flush()
```
g-gaston, line 23: nit: it's probably a bit cleaner to add this just after creating the writer in line 22

pkg/curatedpackages/bundlesource.go
```diff
@@ -0,0 +1,31 @@
...
+	switch strings.ToLower(s) {
+	case Cluster, Registry:
+		*b = BundleSource(s)
+	default:
+		return fmt.Errorf("unknown source: %q", s)
+	}
+	return nil
+}
+
+func (b BundleSource) Type() string {
```
g-gaston, line 29: What is this for?

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/executables"
+)
+
+const (
+	RegistryBaseRef = "public.ecr.aws/q0f6t3x4/eksa-package-bundles"
+)
+
+func GetLatestBundle(ctx context.Context, kubeConfig string, source BundleSource, kubeVersion string) (*api.PackageBundle, error) {
+	switch source {
```
g-gaston, line 25: Instead of having a switch, you could make `BundleSource` and interface and make it implement `GetLatestBundle`

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+	obj := &api.PackageBundleController{}
+	if err := json.Unmarshal(stdOut.Bytes(), obj); err != nil {
+		return nil, fmt.Errorf("unmarshaling active package bundle controller: %w", err)
+	}
+	return obj, nil
+}
+
+func newDependencies(ctx context.Context) (*dependencies.Dependencies, error) {
```
g-gaston, line 94: I would suggest building this at the `cmd` level an only injecting whatever dependencies you have

pkg/curatedpackages/bundle.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+	bundle, err := getPackageBundle(ctx, kubectl, kubeConfig, bundleController.Spec.ActiveBundle)
+	if err != nil {
+		return nil, err
+	}
+	return bundle, nil
+}
+
+func getPackageBundle(ctx context.Context, kubectl *executables.Kubectl, kubeConfig string, activeBundle string) (*api.PackageBundle, error) {
```
g-gaston, line 68: We might want to create a `CuratedPackagesKubeClient` for these two methods, just wrapping around the `kubectl` executable

g-gaston (changes requested):
Overall it looks good

However, we are missing unit tests for all this code, could we add some? Also most of this methods are private methods, so we might want to split the code into multiple entities (or public standalone methods if that's the pattern you want to use) to facilitate writing tests

cmd/eksctl-anywhere/cmd/listpackages.go
a-cool-train: The validation occurs in BundleSource.

pkg/executables/kubectl.go
a-cool-train: This provides ability to use kubectlOpt with something like: get package my-flux. So in this case my-flux would be the arg. I think this follows the approach we have been trying to follow for KubectlOpt.

pkg/curatedpackages/bundlesource.go
a-cool-train: This is to validate the bundlesource i.e. the user is only able to pass in cluster and registry for source flag.

cmd/eksctl-anywhere/cmd/validations.go
```diff
@@ -33,3 +35,14 @@ func commonValidation(ctx context.Context, clusterConfigFile string) (*v1alpha1.
 	}
 	return clusterConfig, nil
 }
+
+func validateKubeVersion(kubeVersion string, source curatedpackages.BundleSource) error {
+	if source != curatedpackages.Registry {
+		return nil
+	}
+	versionSplit := strings.Split(kubeVersion, ".")
+	if len(versionSplit) < 2 {
```
chrisdoherty4, line 44: This allows `1.2.3.4`. We have a `semver` package that _should_ parse this appropriately, it might be better to use that logic.

a-cool-train:
> Overall it looks good
> 
> However, we are missing unit tests for all this code, could we add some? Also most of this methods are private methods, so we might want to split the code into multiple entities (or public standalone methods if that's the pattern you want to use) to facilitate writing tests

@g-gaston, I haven't added unit tests in this PR because there are several changes that occur in this class in other implementations. I plan on adding unit tests once the other branches get merged.

g-gaston:
> > Overall it looks good
> > However, we are missing unit tests for all this code, could we add some? Also most of this methods are private methods, so we might want to split the code into multiple entities (or public standalone methods if that's the pattern you want to use) to facilitate writing tests
> 
> @g-gaston, I haven't added unit tests in this PR because there are several changes that occur in this class in other implementations. I plan on adding unit tests once the other branches get merged.

It's ok to have to change the tests when you change the code! If you already have the tests written then just copy-paste the relevant ones in this PR and we can check how much they cover. And based on that we can see if we need to add more or the rest can be added later.

cmd/eksctl-anywhere/cmd/validations.go
TerryHowe: probably only support `== 2`. Maybe an e.g. on the error message `e.g.: 1.22`

pkg/curatedpackages/bundle.go
TerryHowe: Up above there was talk about making BundleSource an interface and now moving some of this code to a new executable. I'd chose one or the other and I think the derived BundleSource idea was a bit better.

pkg/curatedpackages/bundle_test.go
```diff
@@ -0,0 +1,168 @@
+package curatedpackages
```
g-gaston, line 1: could we test this from `curatedpackages_test`? We always try to do blackbox testing when possible

pkg/curatedpackages/bundle_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+type mockBundleManager struct{}
+
+// generated via
+// impl -dir $GOPATH/pkg/mod/github.com/aws/eks-anywhere-packages@v0.1.1/pkg/bundle 'bm *mockBundleManager' bundle.Manager
+
+// IsActive returns true if the given namespace and name matche the active
+// bundle's.
+func (bm *mockBundleManager) IsActive(ctx context.Context, client client.Client, namespacedName types.NamespacedName) (bool, error) {
```
g-gaston, line 124: could we have a different package for the mocks?
Any reason why we can't use `gomock` for this one?

pkg/curatedpackages/bundlesource.go
```diff
@@ -0,0 +1,32 @@
...
+const (
+	Cluster  = "cluster"
+	Registry = "registry"
+)
+
+func (b BundleSource) String() string {
+	return string(b)
+}
+
+func (b *BundleSource) Set(s string) error {
```
g-gaston, line 19: Could we add a doc comment to these methods explaining that they implement an interface required by cobra for validation?

pkg/curatedpackages/customregistry.go
```diff
@@ -0,0 +1,27 @@
...
+func (cm *CustomRegistry) Login(ctx context.Context) error {
+	err := cm.RegistryLogin(ctx, cm.Registry, cm.Username, cm.Password)
+	if err != nil {
+		return fmt.Errorf("unable to login to registry %v", err)
+	}
+	return nil
+}
+
+func (cm *CustomRegistry) GetRegistryBaseRef(ctx context.Context) (string, error) {
+	return fmt.Sprintf("%s/%s", cm.Registry, RepositoryName), nil
```
g-gaston, line 26: nit: this could be a good addition to the `oci` package?

pkg/executables/kubectl.go
```diff
@@ -1453,3 +1457,10 @@ func (k *Kubectl) GetBmcsPowerState(ctx context.Context, bmcNames []string, kube
 
 	return strings.Fields(buffer.String()), nil
 }
+
+func (k *Kubectl) ApplyResources(ctx context.Context, opts ...KubectlOpt) (bytes.Buffer, error) {
```
g-gaston, line 1461: We already had a method for `kubectl apply`, can we reuse it?

pkg/curatedpackages/kubeclient.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"bytes"
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/executables"
+)
+
+type KubeClient interface {
+	ApplyResources(ctx context.Context, opts ...executables.KubectlOpt) (bytes.Buffer, error)
```
g-gaston, line 11: This interface is a bit problematic because it receives a concrete type from the `executables` package. That creates a dependency on `executables` and break the abstraction.

I'm ok addressing this in separate PR if we come up with a solution and create a ticket :)

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,130 @@
...
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+
+	bm := createBundleManager(lpo.kubeVersion)
+	registry, err := newRegistry(ctx, deps)
+	if err != nil {
+		return err
+	}
+
+	b := curatedpackages.Bundle{
```
TerryHowe, line 77: bundle is a painfully overloaded word here

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,130 @@
...
+		BundleManager: bm,
+		CliVersion:    version.Get(),
+		Registry:      registry,
+	}
+
+	bundle, err := b.GetLatestBundle(ctx)
+	if err != nil {
+		return err
+	}
+	packages := curatedpackages.Package{
```
TerryHowe, line 91: Same with this Package, some consistent pattern here that this is a client or context to get a Package. Also a bit odd it is called packages in the var name, but maybe if this becomes `PackageClient` the var could be `packageClient`

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+func createBundleManager(kubeVersion string) bundle.Manager {
+	versionSplit := strings.Split(kubeVersion, ".")
+	if len(versionSplit) != 2 {
+		return nil
+	}
+	major, minor := versionSplit[0], versionSplit[1]
+	log := logr.Discard()
```
TerryHowe, line 104: A comment about using this might clear things up. I assume this returns a client

pkg/curatedpackages/defaultregistry.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/manifests/bundles"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
```
TerryHowe, line 11: Why is this called DefaultRegistry?

cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -41,33 +40,25 @@ func preRunPackages(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
 	kubectl := deps.Kubectl
 
 	params := []executables.KubectlOpt{executables.WithKubeconfig(kubeConfig), executables.WithArgs(args), executables.WithNamespace(constants.EksaPackagesName)}
 	if output != "" {
-		params = append(params, executables.WithOutput(output))
+		params = append(params, executables.WithResource("get"), executables.WithResourceType(resourceType), executables.WithOutput(output))
 	}
```
TerryHowe, line 52: I totally don't get why getResources would call apply

cmd/eksctl-anywhere/cmd/listpackages.go
```diff
@@ -102,29 +95,34 @@ func createBundleManager(kubeVersion string) bundle.Manager {
...
+func newRegistry(ctx context.Context, deps *dependencies.Dependencies, lpo *listPackagesOption) (curatedpackages.BundleRegistry, error) {
 	if lpo.registry != "" {
-		registry := &curatedpackages.CustomRegistry{
-			Helm:     deps.Helm,
-			Registry: lpo.registry,
-			Username: lpo.username,
-			Password: lpo.password,
+		registryUsername := os.Getenv("REGISTRY_USERNAME")
+		registryPassword := os.Getenv("REGISTRY_PASSWORD")
+		if registryUsername == "" || registryPassword == "" {
```
g-gaston, line 107: nit: I think we have method already that does this `readRegistryCredentials`

g-gaston (approved):

pkg/curatedpackages/packageclient.go
```diff
@@ -0,0 +1,48 @@
...
+	return &PackageClient{
+		packages: packages,
+	}
+}
+
+func (p *PackageClient) DisplayPackages() {
+	w := new(tabwriter.Writer)
+	defer w.Flush()
+	w.Init(os.Stdout, minWidth, tabWidth, padding, padChar, flags)
+	fmt.Fprintf(w, "%s\t%s\t \n", "PackageClient", "Version(s)")
```
vivek-koppuru, line 34: How does kubectl do this? For example, the bars below would vary based on the length of the other strings right?

chrisdoherty4: I'd raised this previously. Have a read of the tabwriter pkg https://pkg.go.dev/text/tabwriter. 

> Package tabwriter implements a write filter (tabwriter.Writer) that translates tabbed columns in input into properly aligned text

PR #1865, opened 2022-04-18
Add EKS-D to core components upgrades

taneyland:
*Issue #, if available:*
#1436 

*Description of changes:*

*Testing (if applicable):*


pkg/clustermanager/upgrader.go
```diff
@@ -54,3 +54,39 @@ func EksaChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 	}
 	return nil
 }
+
+func (u *Upgrader) EksdUpgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error) {
```
g-gaston, line 58: can we move this to a new entity? like `eksd.Upgrader`
Ideally we should stop adding more things to `clustermanager`, it already has too many responsibilities.

pkg/clustermanager/upgrader.go
```diff
@@ -54,3 +54,39 @@ func EksaChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
...
+		return nil, nil
+	}
+	changeDiff := EksdChangeDiff(currentSpec, newSpec)
+	if changeDiff == nil {
+		logger.V(1).Info("Nothing to upgrade for EKS-D components")
+		return nil, nil
+	}
+	logger.V(1).Info("Starting EKS-D components upgrade")
+	oldVersion := currentSpec.VersionsBundle.EksD.Name
+	newVersion := newSpec.VersionsBundle.EksD.Name
```
g-gaston, line 71: Can you get these two from `changeDiff`?

pkg/clustermanager/upgrader.go
```diff
@@ -54,3 +54,39 @@ func EksaChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
...
+	if err := u.retrier.installEksdComponents(ctx, newSpec, cluster); err != nil {
+		return nil, fmt.Errorf("failed upgrading EKS-D components from version %v to version %v: %v", oldVersion, newVersion, err)
```
g-gaston, line 73:
```suggestion
		return nil, fmt.Errorf("upgrading EKS-D components from version %s to version %s: %v", oldVersion, newVersion, err)
```

pkg/workflows/create.go
```diff
@@ -265,6 +265,12 @@ func (s *InstallEksaComponentsTask) Run(ctx context.Context, commandContext *tas
 			commandContext.SetError(err)
 			return &CollectDiagnosticsTask{}
 		}
+		logger.Info("Installing EKS-D components on workload cluster")
+		err = commandContext.ClusterManager.InstallEksdComponents(ctx, commandContext.ClusterSpec, commandContext.WorkloadCluster)
```
g-gaston, line 269: If we move the upgrade method maybe we should move this method to something like `eksd.Installer`

pkg/eksd/installer.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/aws/eks-anywhere/pkg/retrier"
+	"github.com/aws/eks-anywhere/pkg/types"
+)
+
+const (
+	maxRetries    = 5
+	backOffPeriod = 5 * time.Second
+)
+
+type Eksd struct {
```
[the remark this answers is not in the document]
taneyland: Very true. Yeah, I think having both `Installer` and `Upgrader` will most accurately depict what's happening here

pkg/eksd/upgrader.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/types"
+)
+
+func (e *Installer) Upgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error) {
```
g-gaston, line 12: Was this supposed to be `(u *Upgrader)`?

pkg/workflows/upgrade.go
```diff
@@ -207,6 +210,17 @@ func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.Co
 	}
 	commandContext.UpgradeChangeDiff.Append(changeDiff)
 
+	target := commandContext.ManagementCluster
+	if commandContext.ManagementCluster.ExistingManagement {
```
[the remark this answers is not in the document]
taneyland: Good catch. It still worked since the management cluster already had the release CRD, but it shouldn't apply the release CRD to the workload cluster. Fixed it!

pkg/eksd/upgrader.go
taneyland: I will separate the two interfaces in my follow-up PR.

pkg/workflows/interfaces/interfaces.go
```diff
@@ -60,3 +59,9 @@ type CAPIManager interface {
 	Upgrade(ctx context.Context, managementCluster *types.Cluster, provider providers.Provider, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error)
 	EnsureEtcdProvidersInstallation(ctx context.Context, managementCluster *types.Cluster, provider providers.Provider, currSpec *cluster.Spec) error
 }
+
+type Eksd interface {
```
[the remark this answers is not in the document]
taneyland: I will separate the two interfaces in my follow-up PR.

pkg/eksd/installer.go
```diff
@@ -24,24 +26,26 @@ type EksdInstallerClient interface {
...
 	client  EksdInstallerClient
 	retrier *retrier.Retrier
+	reader  *filereader.Reader
 }
 
 func NewEksdInstaller(client EksdInstallerClient) *Installer {
 	return &Installer{
 		client:  client,
 		retrier: retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
+		reader:  filereader.NewReader(),
```
g-gaston, line 36: Can we inject this as a dependency?
If not, the `clusterSpec` has a method to read manifests

g-gaston (approved):

PR #1869, opened 2022-04-18
Log storage class install only for supported providers

g-gaston:
*Description of changes:*
It was confusing that we were logging this for all providers, even when
there wasn't any storage class being installed. Now it only logs when
the provider returns a storage class.


pokearu (approved):
/lgtm

PR #1879, opened 2022-04-19
Make insecure flag and thumbprint in datacenter config mutable

mitalipaygude:
*Issue #, if available:*
#1206 
*Description of changes:*

*Testing (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1880, opened 2022-04-19
Make insecure flag and thumbprint in datacenter config mutable

mitalipaygude:
*Issue #, if available:*
#1879 
*Description of changes:*

*Testing (if applicable):*


pkg/providers/vsphere/vsphere_test.go
```diff
@@ -2699,7 +2699,7 @@ func TestValidateNewSpecStoragePolicyNameImmutableWorker(t *testing.T) {
 	assert.Error(t, err, "StoragePolicyName should be immutable")
 }
 
-func TestValidateNewSpecTLSInsecureImmutable(t *testing.T) {
+func TestValidateNewSpecTLSInsecureMutable(t *testing.T) {
```
g-gaston, line 2702: I'm ok with this but I would also be ok just removing the test

Issue #1878, opened 2022-04-19
CloudStackMAchineConfig.spec.users takes an array but eksa only passes the first one to capc


wongni:
**What happened**:
When defining multiple users under CloudStackMAchineConfig.spec.users, the only first user is applied in the created cluster. 

**What you expected to happen**:
All users defined under CloudStackMAchineConfig.spec.users should be valid. 

**How to reproduce it (as minimally and precisely as possible)**:
Create a cloudstack cluster with multiple users defined. 
Check if you are able to ssh in with all users.

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: v0.8.3-custom
- EKS Distro Release:

PR #1883, opened 2022-04-19
Fix/1878 support multiple users

wongni:
*Issue #, if available:* #1878

*Description of changes:*
Even though `users:` field takes an array of users but the only first user is passed into the provider layer. Therefore, multiple ssh users are not supported.

*Testing (if applicable):*
Added a new unit test case that verifies multiple ssh users
Manual e2e test using a workload cluster with multiple SSH users.


pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -262,6 +262,32 @@ func TestProviderGenerateCAPISpecForCreateWithAffinity(t *testing.T) {
 	test.AssertContentToFile(t, string(md), "testdata/expected_results_affinity_md.yaml")
 }
 
+func TestProviderGenerateCAPISpecForCreateWithMultpleUsers(t *testing.T) {
```
maxdrib, line 265: it would be good to have a test for upgrade spec as well

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -253,69 +253,49 @@ func (p *cloudstackProvider) MachineResourceType() string {
...
-		} else {
-			if useKeyGeneratedForControlplane { // use the same key
-				workerUser.SshAuthorizedKeys[0] = controlPlaneUser.SshAuthorizedKeys[0]
-			} else {
-				logger.Info("Provided worker sshAuthorizedKey is not set or is empty, auto-generating new key pair...")
-				generatedKey, err := common.GenerateSSHAuthKey(p.writer)
+	return nil
+}
+
+func (p *cloudstackProvider) processUsers(users []v1alpha1.UserConfiguration, generatedKey string) (string, error) {
```
maxdrib, line 277: could we rename to something like `processSshKeysForUser` instead? First, because it only processes a single user, and second so that we can include how we're processing the user

maxdrib (commented):
Have you tested this e2e at all? I'd like to see us actually be able to ssh in with multiple users with cloud-init

pkg/providers/cloudstack/cloudstack.go
wongni: I just renamed it to processSshKeysForUsers because it processes multiple users. `users []v1alpha1.UserConfiguration`

internal/pkg/api/cloudstack.go
```diff
@@ -102,11 +102,40 @@ func WithCloudStackConfigNamespace(ns string) CloudStackFiller {
...
+		for _, user := range config.workerMachineConfig.Spec.Users {
+			if len(user.SshAuthorizedKeys) == 0 {
+				user.SshAuthorizedKeys = []string{""}
+			}
+			for i := range user.SshAuthorizedKeys {
+				user.SshAuthorizedKeys[i] = value
+			}
+		}
+		if config.etcdMachineConfig != nil {
+			if len(config.cpMachineConfig.Spec.Users) == 0 {
```
maxdrib, line 128: any idea what's going on with these errors?

pkg/providers/cloudstack/cloudstack_test.go
maxdrib: bump

vignesh-goutham:
Can we add this support for all the providers? One more thing to note, Bottlerocket does not support multiple users at this point, so lets add a check if osFamily == Bottlerocket, and if multiple users are provided, to provide a meaningful message and fail the workflow.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -261,69 +261,49 @@ func (p *cloudstackProvider) MachineResourceType() string {
...
-				etcdUser.SshAuthorizedKeys[0] = controlPlaneUser.SshAuthorizedKeys[0]
-			} else if useKeyGeneratedForWorker {
-				etcdUser.SshAuthorizedKeys[0] = workerUser.SshAuthorizedKeys[0] // if cp key was provided by user, check if worker key was generated by cli and use that
 			} else {
-				logger.Info("Provided etcd sshAuthorizedKey is not set or is empty, auto-generating new key pair...")
-				generatedKey, err := common.GenerateSSHAuthKey(p.writer)
-				if err != nil {
-					return err
+				if len(generatedKey) == 0 {
+					logger.Info("Provided control plane sshAuthorizedKey is not set or is empty, auto-generating new key pair...")
```
maxdrib, line 296: this log message shouldn't talk about control plane, should it? It could be any provided key, right?

pkg/providers/cloudstack/testdata/cluster_multiple_users.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - name: "cp-user2"
+    sshAuthorizedKeys: # The key below was manually generated and not used in any production systems
+    - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC1BK73XhIzjX+meUr7pIYh6RHbvI3tmHeQIXY5lv7aztN1UoX+bhPo3dwo2sfSQn5kuxgQdnxIZ/CTzy0p0GkEYVv3gwspCeurjmu0XmrdmaSGcGxCEWT/65NtvYrQtUE5ELxJ+N/aeZNlK2B7IWANnw/82913asXH4VksV1NYNduP0o1/G4XcwLLSyVFB078q/oEnmvdNIoS61j4/o36HVtENJgYr0idcBvwJdvcGxGnPaqOhx477t+kfJAa5n5dSA5wilIaoXH5i1Tf/HsTCM52L+iNCARvQzJYZhzbWI1MDQwzILtIBEQCJsl2XSqIupleY8CxqQ6jCXt2mhae+wPc3YmbO5rFvr2/EvC57kh3yDs1Nsuj8KOvD78KeeujbR8n8pScm3WDp62HFQ8lEKNdeRNj6kB8WnuaJvPnyZfvzOhwG65/9w13IBl7B1sWxbFnq2rMpm5uHVK7mAmjL0Tt8zoDhcE1YJEnp9xte3/pvmKPkST5Q/9ZtR9P5sI+02jY0fvPkPyC03j2gsPixG7rpOCwpOdbny4dcj0TDeeXJX8er+oVfJuLYz0pNWJcT2raDdFfcqvYA0B0IyNYlj5nWX4RuEcyT3qocLReWPnZojetvAG/H8XwOh7fEVGqHAKOVSnPXCSQJPl6s0H12jPJBDJMTydtYPEszl4/CeQ== testemail@test.com"
```
maxdrib, line 65: i would suggest having each user's key be slightly unique in this test for more comprehensive test coverage

maxdrib (commented):
nice work, I'm excited for this feature to be available

wongni:
> Awesome work to refactor and condense all the logic. Have you tested this e2e at all? I'd like to see us actually be able to ssh in with multiple users with cloud-init

I tested the multi ssh users in a workload cluster and it worked well.

pkg/providers/cloudstack/cloudstack.go
wongni: Good point!

pkg/providers/cloudstack/testdata/cluster_multiple_users.yaml
wongni: Good point.

vignesh-goutham:
I think we overall need to add some logic for bottlerocket. 

Bottlerocket only support "ec2-user" as the username and only accept one key for that user-name at this point. We can add multiple keys to the same username, but that's to be tested. Can we add input preflights validation to block cluster-spec from allowing multiple keys when the osFamily is Bottlerocket?

Issue #1886, opened 2022-04-19
Passing a bad file to --kubeconfig makes a panic occur


wongni:
**What happened**:
When a user passes a bad file (in this example, a cluster manifest is passed in), the eksa executable throws a panic instead of printing a proper error message. 

`eksctl anywhere create cluster -f cluster-config-workload1.yaml -v 9 --force-cleanup --kubeconfig=cluster-config-workload1.yaml
`

`
2022-04-19T22:52:53.771Z        V4      Logger init completed   {"vlevel": 9}
2022-04-19T22:52:53.773Z        V1      RegistryMirrorConfiguration.Port is not specified, default port will be used    {"Default Port": "443"}
2022-04-19T22:52:53.773Z        V1      First worker node group name not specified. Defaulting name to md-0.
2022-04-19T22:52:53.773Z        V6      Executing command       {"cmd": "/usr/bin/docker version --format {{.Client.Version}}"}
2022-04-19T22:52:53.834Z        V6      Executing command       {"cmd": "/usr/bin/docker info --format '{{json .MemTotal}}'"}
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/aws/eks-anywhere/pkg/cluster.LoadManagement(0x7ffea94551f6, 0x1d, 0xc00084e060, 0x2c, 0x0)
        /Users/vgg/Code/Github/eks-anywhere/pkg/cluster/load.go:35 +0x325
github.com/aws/eks-anywhere/cmd/eksctl-anywhere/cmd.newClusterSpec(0x7ffea94551b6, 0x1d, 0x0, 0x0, 0x7ffea94551f6, 0x1d, 0x4142db, 0xc00083fbd8, 0x21ea16b3)
        /Users/vgg/Code/Github/eks-anywhere/cmd/eksctl-anywhere/cmd/options.go:32 +0x18f
github.com/aws/eks-anywhere/cmd/eksctl-anywhere/cmd.(*createClusterOptions).createCluster(0x43057e0, 0x42e9740, 0xc00007fe60, 0x0, 0x6, 0x0, 0x0)
        /Users/vgg/Code/Github/eks-anywhere/cmd/eksctl-anywhere/cmd/createcluster.go:127 +0x3a9
github.com/spf13/cobra.(*Command).execute(0x42e9740, 0xc00007fe00, 0x6, 0x6, 0x42e9740, 0xc00007fe00)
[...]

PR #1887, opened 2022-04-19
Fix/1886 print error for bad kubeconfig

wongni:
*Issue #, if available:* #1886

*Description of changes:* 
When passing an invalid kubeconfig file into --kubeconfig of eksa executable a panic is thrown. We should print out a proper error message instead of throwing a panic.

*Testing (if applicable):*
Manual test. 


vignesh-goutham (approved):
/approve

Issue #1876, opened 2022-04-19
Flux executable supports generic `git bootstrap`


danbudris:
the `flux` executable will need to support [`bootstrap git`](https://fluxcd.io/docs/cmd/flux_bootstrap_git/) in addition to the existing [`bootstrap github`](https://fluxcd.io/docs/cmd/flux_bootstrap_github/).

currently the flux executable package only supports [`bootstrap github`](https://github.com/aws/eks-anywhere/blob/7f4d5882e89bcffec002ef60d5a00c90db0e1ba5/pkg/executables/flux.go#L32).

we need to add a layer under this method which looks at the FluxConfig, checks which provider is present (github vs git), and then executes a sub-command like `fluxBootstrapGit` or `fluxBootstrapGithub`, passing in the relevant configuration values. 

We should consider setting up the `options` pattern for `flux` so that we can pass options to a generic method like `fluxBootstrap(...opts FluxBootstrapOpts)` which sets the command opts for flux.

the ultimate goal of this ticket is to set us up so that if the `git` field of the `FluxConfig` is non-nil we properly process it and run the right bootstrap command.

PR #1900, opened 2022-04-20
Flux executable support for generic git provider

mitalipaygude:
*Issue #, if available:*
#1876 

*Description of changes:*
The flux executable needs to support [bootstrap git](https://fluxcd.io/docs/cmd/flux_bootstrap_git/) in addition to the existing `bootstrap github`.
This PR adds the method to bootstrap git.

*Testing (if applicable):*
- Unit tests
- Create cluster


pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"os"
+
+	"github.com/aws/eks-anywhere/pkg/logger"
+)
+
+const (
+	GitPasswordTokenEnv       = "GIT_PASSWORD"
+	EksaGitPasswordTokenEnv   = "EKSA_GIT_PASSWORD"
+	GitPrivateKeyTokenEnv     = "GIT_PRIVATE_KEY"
```
danbudris, line 13: do we need `GIT_PRIVATE_KEY`?

I know we need `GIT_PASSWORD` as the env var we are passing to `flux bootstrap git`, but I think we can just read `EKSA_GIT_PRIVATE_KEY` straight from env and not worry about having `GIT_PRIVATE_KEY` as an intermediate env var, since we're setting the file path as a string in the command execution

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,71 @@
...
+func getTokenFromEnv(EksaEnvVar, GitEnvVar string) (string, error) {
+	logger.V(4).Info("Checking for environment variable", "env var", EksaEnvVar)
+
+	env := make(map[string]string)
+	val, ok := os.LookupEnv(EksaEnvVar)
+	if !ok || len(val) == 0 {
+		logger.Info("%s environment variable is not set", EksaEnvVar)
+		return "", fmt.Errorf("environment variable %s is not set", EksaEnvVar)
+	}
+	if len(val) > 0 {
```
danbudris, line 54: I don't think we need to do any of this for the `EKSA_GIT_PRIVATE_KEY` env var, right? we just need to get the path value as a string.

pkg/executables/flux_test.go
```diff
@@ -433,3 +462,143 @@ func TestFluxReconcile(t *testing.T) {
...
+				},
+			},
+
+			wantExecArgs: []interface{}{
+				"bootstrap", gitProvider, "--url", repoUrl, "--username", username, "--path", path, "--ssh-key-algorithm", "ecdsa", "--branch", "main",
+				"--private-key-file", privateKeyFilePath,
+			},
+		},
+		//{
+		//	testName: "with namespace",
```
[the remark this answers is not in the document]
mitalipaygude: still working on the test

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,71 @@
...
+	GitPrivateKeyTokenEnv     = "GIT_PRIVATE_KEY"
+	EksaGitPrivateKeyTokenEnv = "EKSA_GIT_PRIVATE_KEY"
+)
+
+type Options struct {
+	RepositoryUrl string
+	User          string
+}
+
+func GetGitPasswordFromEnv() (string, error) {
```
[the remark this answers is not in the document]
mitalipaygude: in case this returns an error, we don't return an error from the method that calls this, we just log a message

pkg/git/providers/git/git.go
mitalipaygude: that's true, we don't need this var

pkg/git/providers/git/git.go
mitalipaygude: Added a validation on the path to see if it exists and the file is not empty.

pkg/api/v1alpha1/fluxconfig.go
```diff
@@ -71,10 +77,26 @@ func validateGitProviderConfig(config GitProviderConfig) error {
 	if len(config.RepositoryUrl) <= 0 {
 		return errors.New("'repositoryUrl' is not set or empty in gitProviderConfig; repositoryUrl is a required field")
 	}
+
 	err := validateRepositoryUrl(config.RepositoryUrl)
+	if err != nil {
+		return err
+	}
+
+	err = validateAuthentication()
```
danbudris, line 86: nit: just return this

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -129,11 +129,19 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
 
-	deps, err := dependencies.ForSpec(ctx, clusterSpec).WithExecutableMountDirs(cc.mountDirs()...).
+	var dirs []string
+	dirs = cc.mountDirs()
+	if clusterSpec.FluxConfig != nil && clusterSpec.FluxConfig.Spec.Git != nil {
+		privateKeyFilePath, fileExists := mountPrivateKeyFile()
+		if fileExists {
+			dirs = append(dirs, privateKeyFilePath)
+		}
+	}
```
g-gaston, line 139: We are probably going to need this again (upgrade?) so I would vote to make it a bit more reusable

Ideally we would have on single method that, upon receiving all the necessary info, it returns the directories that need to be mounted

we can have something like
`func directoriesToMount(clusterOptions, clusterConfig)`

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -23,6 +27,17 @@ func (c clusterOptions) mountDirs() []string {
 	return dirs
 }
 
+func mountPrivateKeyFile() (string, bool) {
+	privateKeyFilePath, err := git.GetTokenFromEnv(EksaGitPrivateKeyTokenEnv)
```
g-gaston, line 31: It's a bit weird to get a path from a method called GetToken
Can we just read the env var normally? Or just rename that method to ReadEnvVar or something like that
Which unless there is some special git logic, should be in a different package

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -23,6 +27,17 @@ func (c clusterOptions) mountDirs() []string {
...
+	if err != nil {
+		return "", false
+	}
+	if validations.FileExistsAndIsNotEmpty(privateKeyFilePath) {
```
g-gaston, line 35: Shouldn't this return an error if it doesn't exist?
The method signature is a bit weird. If the file not existing is not considered an error, then can we just signify that scenario with an empty string?

pkg/api/v1alpha1/fluxconfig_test.go
```diff
@@ -1,13 +1,24 @@
...
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	validPassword           = "testPassword"
+	eksaGitPasswordTokenEnv = "EKSA_GIT_PASSWORD"
+)
+
+func SaveContext() {
```
g-gaston, line 16: why is this method public?
Also we should cleanup the env var after the test. You can use t.Cleanup
There are exampled like this somewhere, maybe vsphere

pkg/api/v1alpha1/fluxconfig.go
```diff
@@ -71,10 +77,26 @@ func validateGitProviderConfig(config GitProviderConfig) error {
...
 	return err
 }
 
+func validateAuthentication() error {
+	password, _ := git.GetTokenFromEnv(EksaGitPasswordTokenEnv)
+	privateKeyFilePath, _ := git.GetTokenFromEnv(EksaGitPrivateKeyTokenEnv)
```
g-gaston, line 92: Would it possible to read this before hand and inject it as the rest of the dependencies in the constructor?

pkg/executables/flux.go
```diff
@@ -63,17 +67,61 @@ func (f *Flux) BootstrapToolkitsComponentsGithub(ctx context.Context, cluster *t
...
 
-// BootstrapToolkitsComponentsGit creates the Git repository if it does not exist, and commits the toolkit
-// components manifests to the main branch. Then it configures the target cluster to synchronize with the repository.
-// If the toolkit components are present on the cluster, the bootstrap command will perform an upgrade if needed.
+// BootstrapToolkitsComponentsGit commits the toolkit components manifests to the branch of a Git repository.
+// It then configures the target cluster to synchronize with the repository. If the toolkit components are present on the cluster, the
+// bootstrap command will perform an upgrade if needed.
 func (f *Flux) BootstrapToolkitsComponentsGit(ctx context.Context, cluster *types.Cluster, fluxConfig *v1alpha1.FluxConfig) error {
-	return nil
+	c := fluxConfig.Spec
```
g-gaston, line 80: This is almost like copy paste from `BootstrapToolkitsComponentsGithub`
Can we have a common private method that both of them use?

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,69 @@
...
+	password, err := GetTokenFromEnv(EksaGitPasswordTokenEnv)
+	if err != nil {
+		return "", err
+	}
+	if len(password) > 0 {
+		if err := os.Setenv(GitPasswordTokenEnv, password); err != nil {
```
g-gaston, line 27: Curious, why is this needed? Can't it be passed to the executable as an argument?

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,69 @@
...
+
+func GetGitPrivateKeyFileFromEnv() (string, error) {
+	privateKeyFilePath, err := GetTokenFromEnv(EksaGitPrivateKeyTokenEnv)
+	if err != nil {
+		return "", err
+	}
+	privateKeyfileExistsAndIsNotEmpty := validateFilePath(privateKeyFilePath)
+	if !privateKeyfileExistsAndIsNotEmpty {
+		logger.V(4).Info("Private key file does not exist or is empty", "file path", privateKeyFilePath)
+		return "", fmt.Errorf("private key file does not exist at %s or is empty", privateKeyFilePath)
```
g-gaston, line 43: This would fail during the flux execution
Should this be a pre flight validation of some sorts?

cmd/eksctl-anywhere/cmd/options.go
danbudris: yeah we should rename that method -- it was called 'GetTokenFromEnv' bc it was used for the Github Personal Access Token, but we should use some generic accessor

pkg/api/v1alpha1/fluxconfig_test.go
danbudris: we're now on 1.17 so you can use `t.Setenv` to set it for the duration of the test only too

pkg/git/providers/git/git.go
mitalipaygude: It can be passed as an argument as well as an env var. we can decide to always send it in as an argument.

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -188,3 +191,21 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 
 	return createCluster.Run(ctx, clusterSpec, createValidations, cc.forceClean)
 }
+
+func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec) []string {
```
g-gaston, line 195: Maybe I'm wrong but, don't we need this for upgrade as well?

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -23,6 +27,17 @@ func (c clusterOptions) mountDirs() []string {
 	return dirs
 }
 
+func mountPrivateKeyFile() string {
```
g-gaston, line 30: nit: 
```suggestion
func gitPrivateKeyFilePath() string {
```

cmd/eksctl-anywhere/cmd/options.go
g-gaston: I understand the env var might not be set, but if it is, and the file doesn't exist, shouldn't we return an error? Btw, that kind of validation should probably be somewhere else

If we want to see if the env var is set, we should probably use `os.LookupEnv`

Summary of what I'm proposing:

* If the env var is not set, ignore it and don't mount anything
* If it's set, mount it and validate it is a valid and existing path (but move the validation somewhere else)

pkg/executables/flux.go
```diff
@@ -63,17 +58,62 @@ func (f *Flux) BootstrapToolkitsComponentsGithub(ctx context.Context, cluster *t
...
+	params := []string{
+		"bootstrap",
+		gitProvider,
+		"--url", c.Git.RepositoryUrl,
+		"--username", c.Git.Username,
+		"--path", c.ClusterConfigPath,
+		"--ssh-key-algorithm", privateKeyAlgorithm,
+	}
```
g-gaston, line 79: Why not make these part of the common private method as well?
Maybe we could simplify all this code by having some kind of `bootstrap` command builder construct.

like
```go
newBoostrapCommand(gitProviderm,
  withUrl(c.Git.RepositoryUrl),
  withUsername(c.Git.Username),
  ...
  withPassword(password),
)
```

pkg/executables/flux_test.go
```diff
@@ -433,3 +441,109 @@ func TestFluxReconcile(t *testing.T) {
 		})
 	}
 }
+
+func TestFluxInstallGitToolkitsSuccess(t *testing.T) {
```
g-gaston, line 445: can we name this based on the tested method?
```suggestion
func TestFluxBootstrapToolkitsComponentsGitSuccess(t *testing.T) {
```

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,48 @@
...
+func GetGitPrivateKeyFile() (string, error) {
+	privateKeyFilePath, err := ReadEnvVar(EksaGitPrivateKeyTokenEnv)
+	if err != nil {
+		return "", err
+	}
+
+	return privateKeyFilePath, nil
+}
+
+func ReadEnvVar(EksaEnvVar string) (string, error) {
```
g-gaston, line 38: I think we should move this to a different package, it's not git specific

Also, what do you think about a different name that tells us that it will fail if the variable is not set or empty?
I'm out of good ideas but for example `ReadAndExpectEnvVar`

pkg/validations/createvalidations/gitops.go
```diff
@@ -28,6 +41,30 @@ func ValidateGitOps(ctx context.Context, k validations.KubectlClient, cluster *t
...
+	return err
+}
+
+func validateAuthenticationGitProvider(clusterSpec *cluster.Spec) error {
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig.Spec.Git == nil {
+		return nil
+	}
+
+	password, _ := git.ReadEnvVar(EksaGitPasswordTokenEnv)
```
g-gaston, line 54: why use `git.ReadEnvVar` and then ignore the error?
Returning an error is basically the only thing that `git.ReadEnvVar` adds on top of the `os` methods

pkg/validations/createvalidations/gitops.go
```diff
@@ -28,6 +41,30 @@ func ValidateGitOps(ctx context.Context, k validations.KubectlClient, cluster *t
...
+	privateKeyFilePath, _ := git.ReadEnvVar(EksaGitPrivateKeyTokenEnv)
+
+	if password == "" {
+		if privateKeyFilePath == "" {
+			return fmt.Errorf("for git provider in the Flux config either set a password using %s env var or path to private key file using %s", EksaGitPasswordTokenEnv, EksaGitPrivateKeyTokenEnv)
+		}
+		if privateKeyFilePath != "" {
```
g-gaston, line 61: you don't need is, if you get here, `privateKeyFilePath` is not empty

pkg/validations/createvalidations/gitops.go
```diff
@@ -28,6 +41,30 @@ func ValidateGitOps(ctx context.Context, k validations.KubectlClient, cluster *t
...
+			privateKeyfileExistsAndIsNotEmpty := validations.FileExistsAndIsNotEmpty(privateKeyFilePath)
+			if !privateKeyfileExistsAndIsNotEmpty {
```
g-gaston, line 63:
```suggestion
			if !validations.FileExistsAndIsNotEmpty(privateKeyFilePath) {
```

pkg/validations/createvalidations/gitops.go
```diff
@@ -28,6 +41,30 @@ func ValidateGitOps(ctx context.Context, k validations.KubectlClient, cluster *t
...
+func validateAuthenticationGitProvider(clusterSpec *cluster.Spec) error {
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig.Spec.Git == nil {
+		return nil
+	}
+
+	password, _ := git.ReadEnvVar(EksaGitPasswordTokenEnv)
+	privateKeyFilePath, _ := git.ReadEnvVar(EksaGitPrivateKeyTokenEnv)
+
+	if password == "" {
```
g-gaston, line 57: you could unnest this whole block by doing
```go
if password != "" {
  return nil
}
```

pkg/validations/createvalidations/gitops.go
```diff
@@ -5,12 +5,25 @@ import (
...
 	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/git/providers/git"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
+const (
+	EksaGitPasswordTokenEnv   = "EKSA_GIT_PASSWORD"
+	EksaGitPrivateKeyTokenEnv = "EKSA_GIT_PRIVATE_KEY"
```
g-gaston, line 16: We have this constant defined at least 3 times, could move it to just one place?

Also, this makes me realize we are reading this env var at least 3 times. Ideally we should read it once and pass it around.

pkg/validations/createvalidations/gitops_test.go
```diff
@@ -315,3 +330,172 @@ func TestValidateGitopsForSelfManagedCluster(t *testing.T) {
...
+
+			err := createvalidations.ValidateGitOps(ctx, k, cluster, clusterSpec)
+			if !reflect.DeepEqual(err, tc.wantErr) {
+				t.Errorf("%v got = %v, \nwant %v", tc.name, err, tc.wantErr)
+			}
+		})
+	}
+}
+
+func TestValidateNoGitProviderAuthForWorkloadCluster(t *testing.T) {
```
g-gaston, line 466: Can we rename these methods to follow the pattern `Test[entity-name][tested-method-name][scenario]`?

pkg/git/providers/git/git.go
```diff
@@ -0,0 +1,48 @@
...
+	User          string
+}
+
+func GetGitPasswordFromEnv() (string, error) {
+	password, err := ReadEnvVar(EksaGitPasswordTokenEnv)
+	if err != nil {
+		return "", err
+	}
+
+	return password, nil
```
g-gaston, line 26: If you are not going to add context to the error, you can just
```suggestion
	return ReadEnvVar(EksaGitPasswordTokenEnv)
```

pkg/api/v1alpha1/fluxconfig.go
g-gaston: @mitalipaygude what do you think about this?

cmd/eksctl-anywhere/cmd/createcluster.go
mitalipaygude: We won't because gitops is immutable at the moment, can't change it for upgrade.

pkg/executables/flux.go
mitalipaygude: The parameters for both bootstrap commands are a little different, so I pulled out the common params in a different method.

pkg/validations/createvalidations/gitops.go
mitalipaygude: The reason we don't want to return here is because the user can specify both password and private key, in which case we want to validate values on them both.

cmd/eksctl-anywhere/cmd/options.go
mitalipaygude: The validation currently does exist in the preflight validations, however we mount the dirs before getting to the preflight validations, which is why we validate the file here as well.

pkg/validations/createvalidations/gitops.go
mitalipaygude: So we read the private key file once when we mount the file on the container, we also read both those vars when we  do preflight validations. with the validations, we only return errors throughout which is why there isn't really a way currently to pass these values back.

pkg/executables/flux.go
mitalipaygude: To be fixed: https://github.com/aws/eks-anywhere/issues/1952

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -185,8 +193,45 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
+	if clusterSpec.FluxConfig == nil {
+		return nil
+	}
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig.Spec.Git == nil {
+		return nil
+	}
+	var cliConfig config.CliConfig
+
+	password, _ := config.ReadEnvVar(EksaGitPasswordTokenEnv)
```
g-gaston, line 230: nit: I would suggest just using `os.GetEnv`

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -185,8 +193,45 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
+	return dirs, nil
+}
+
+func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
+	if clusterSpec.FluxConfig == nil {
+		return nil
+	}
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig.Spec.Git == nil {
+		return nil
```
g-gaston, line 226: I would suggest just returning an empty struct instead of nil
The code that needs to read this can decide what to do with empty fields
If we return nil here and the rest of the code relies on that, all that code need to change if we add more fields to this config

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -185,8 +193,45 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
 		ManagementCluster: cluster,
 		Provider:          deps.Provider,
+		CliConfig:         cliConfig,
 	}
 	createValidations := createvalidations.New(validationOpts)
 
 	return createCluster.Run(ctx, clusterSpec, createValidations, cc.forceClean)
 }
+
+func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig *config.CliConfig) ([]string, error) {
```
g-gaston, line 203: seems like we never return error here, I think we can remove it and simplify the code :)

pkg/api/v1alpha1/fluxconfig_test.go
```diff
@@ -1,13 +1,28 @@
...
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	validPassword           = "testPassword"
+	eksaGitPasswordTokenEnv = "EKSA_GIT_PASSWORD"
+)
+
+func saveContext(t *testing.T) {
```
g-gaston, line 16: Do we need this anymore?

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -14,6 +14,11 @@ type clusterOptions struct {
 	managementKubeconfig string
 }
 
+const (
+	EksaGitPasswordTokenEnv   = "EKSA_GIT_PASSWORD"
+	EksaGitPrivateKeyTokenEnv = "EKSA_GIT_PRIVATE_KEY"
```
g-gaston, line 19: Can we delete this and use the env vars from the `config` package?

pkg/validations/createvalidations/gitops.go
```diff
@@ -28,6 +36,24 @@ func ValidateGitOps(ctx context.Context, k validations.KubectlClient, cluster *t
...
+func validateAuthenticationGitProvider(clusterSpec *cluster.Spec, cliConfig *config.CliConfig) error {
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig.Spec.Git == nil {
+		return nil
+	}
+	if cliConfig.GitPassword == "" && cliConfig.GitPrivateKeyFile == "" {
+		return fmt.Errorf("for git provider in the Flux config either set a password using %s env var or path to private key"+
+			" file using %s", config.EksaGitPasswordTokenEnv, config.EksaGitPrivateKeyTokenEnv)
+	}
+	if cliConfig.GitPrivateKeyFile != "" {
```
g-gaston, line 52: If both are set, what happens? This is mostly for my understanding, I'm not asking to change anything. Just wondering if we need a validation or if just one of them takes precedence (which we could print a warning or something like that)

pkg/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+	EksaGitPasswordTokenEnv   = "EKSA_GIT_PASSWORD"
+	EksaGitPrivateKeyTokenEnv = "EKSA_GIT_PRIVATE_KEY"
+)
+
+type CliConfig struct {
+	GitPassword       string
+	GitPrivateKeyFile string
+}
+
+func ReadEnvVar(EksaEnvVar string) (string, error) {
```
g-gaston, line 20: What do you think about 
```suggestion
func ReadExpectedEnvVar(EksaEnvVar string) (string, error) {
```

g-gaston, line 20: We should add a test for this :)

g-gaston (commented):
Good job, looking awesome

pkg/validations/createvalidations/gitops.go
mitalipaygude: If both are set, we pass them both as params without any one taking specific precendence
https://fluxcd.io/docs/cmd/flux_bootstrap_git/#examples

pkg/config/config.go
mitalipaygude: Ended up using os.getEnv instead of this, so removed the method.

g-gaston (approved):

PR #1901, opened 2022-04-20
Add .ignore directory to .gitignore

chrisdoherty4:
It can be useful to keep files around in a repository that you use
frequently but we don't have a dedicated directory for housing them.
Now we do, `.ignore` :)


abhinavmpandey08 (approved):
/lgtm

PR #1903, opened 2022-04-21
Gzip artifacts tarball

g-gaston:
*Description of changes:*
This decreases the tarball size from 9.5GB to 2.9 GB.
It increases the time `download images` takes from ~6:30 to ~12:30

*Testing (if applicable):*
Run download and import images manually


jiayiwang7 (commented):
/lgtm

PR #1905, opened 2022-04-21
Avoid packaging tools image twice in tarball

g-gaston:
*Description of changes:*
Since the eks-a tools image is already packaged individually, we don't
need to include it in the big tar with the rest of the images.

*Testing (if applicable):*
Ran download-import flow manually.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
+++ b/cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
@@ -58,7 +58,7 @@ func (d Download) Run(ctx context.Context) error {
 		return fmt.Errorf("downloading images: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
```diff
@@ -58,7 +58,7 @@ func (d Download) Run(ctx context.Context) error {
 		return fmt.Errorf("downloading images: %v", err)
 	}
 
-	if err = d.BundlesImagesDownloader.Move(ctx, artifactNames(images)...); err != nil {
+	if err = d.BundlesImagesDownloader.Move(ctx, removeFromSlice(artifactNames(images), toolsImage)...); err != nil {
```
jiayiwang7, line 61: can u remind me of why we need separate downloader for tool img and bundle images?

PR #1919, opened 2022-04-22
Add best practice guide

chrisdoherty4:
The basics for our best practice guide. Additional sections will be written in-time but this has sufficient material to merge and iterate on. 

Please review for inconsistency, grammar and spelling, and anything else you feel might be added.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,3 @@
 [Contributing Guidelines](docs/content/en/docs/community/contributing.md)
+
+[Development Practices](docs/developer/README.md)
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -1,0 +1,7 @@
+# Developer
+
+[Development Best Practice Guidlines](best-practice.md)
+
+[Pull Request Guidelines](pull-request.md)
+
+[Issue Management](issues.md)
--- a/docs/developer/best-practice.md
+++ b/docs/developer/best-practice.md
@@ -48,9 +48,15 @@ Avoid [relentlessly] `types`, `interfaces`, `common`, `util` or `base` packages.
 
 Functions and methods (functions with a receiver parameter) should adequetly describe their behavior. Generally, they should follow a verb-noun form. The types of parameters should help inform the function name.
 
-##### Variable
+Capitalize acronyms in functions. When creating an unexported function starting with an acronym, lower-case the whole acronym.
+
+##### Fields & variables
 
 Variable names should be concise and descriptive. Prefer single word names. The further away from the site of declaration a variable is used, the more descriptive it needs to be. For example, a variable named `n` used 30 lines after declaration makes it unnecessarily difficult to reason what it represents at the site of use.
+
+Avoid acronyms that aren't well known. For example, `kcp` should be `kubeadmControlPlane`. If in doubt, prefer clarity over brevity.
+
+Capitalize acronyms in fields. When creating an unexported field starting with an acronym, lower-case the whole acronym.
 
 ### <a name="dependencies"></a> Dependencies
 
@@ -247,7 +253,7 @@ When declaring and not initializing, prefer `var`. For example, `var vehicle Veh
 
 When delcaring and initializing, use `:=`. For example, `vehicle := NewVehicle()`
 
-Make intent clear for complicated variable initializations (contradictory to the 2 above rules). For example `var p uint32 = 0x80` instead of `p := uint32(0x80)`.
+Make intent clear for complicated variable initializations (contradictory to the 2 above rules). For example `var p uint32 = 0x80` makes a statement about the type as opposed to `p := uint32(0x80)` that focus' on the value, prefer the former.
 
 ### Line length
 
--- a/docs/developer/best-practice.md
+++ b/docs/developer/best-practice.md
@@ -36,13 +36,13 @@ In avoiding package level state we avoid [hidden dependencies and unintended sid
 
 ##### Packages
 
-Package names help form the first impression of what a package provides and is used to reference the public API of the package. A good package name brings about clarity and purpose for consumers. Package names should be concise and represent the behavior the package provides not the types it contains. 
+Package names help form the first impression of what a package provides and is used to reference the public API of the package. A good package name brings about clarity and purpose for consumers. Package names should be concise and represent the behavior provided by the package, not the types it contains. 
 
-Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Decorator`. You may find you have a type named after the package, soemtimes it's appropriate.
+Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Decorator`. You may find you have a type named after the package, sometimes it's appropriate and you should lean on the reviewer process to make a judgement call.
 
-If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name.
+Prefer using the singular form of nouns for packgae names. If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name. For example, `string` is a type that could be shadowed hence the stdlib has a `strings` package.
 
-Avoid [relentlessly] `types`, `interfaces`, `common`, `util` or `base` packages. These names offer no indication on what the package provides beyond 'a bundle of stuff'. With time, these types of package turn into dumping grounds containing unrelated logic that can be the root cause for cyclic dependencies.  If a utility style package is required, make it specific. For example a `/utils/cmdutil` package may provide command utility functions.
+Avoid [relentlessly] `types`, `interfaces`, `common`, `util` or `base` packages. These names offer no indication on what the package provides beyond 'a bundle of stuff'. With time, these types of package turn into dumping grounds containing unrelated logic that can be the root cause for cyclic dependencies.  If a utility style package is required, make it specific. For example a `/cmdutil` package may provide command utility functions.
 
 ##### Function & method
 
@@ -54,7 +54,7 @@ Variable names should be concise and descriptive. Prefer single word names. The 
 
 ### <a name="dependencies"></a> Dependencies
 
-Functions should explicitly declare their dependencies as parameters. In doing so, they honor the [dependency inversion principle](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle) and provide the necessary flexibility to write loosely coupled programs.
+Functions should explicitly declare their dependencies as parameters. In doing so they make clear to consumers what is required and increase their flexibility and configurability.
 
 ```go
 func FastenPlanks(tools Toolbox, surface Surface, planks Planks) error {
@@ -72,7 +72,7 @@ Avoid constructing dependencies in-line. If you do construct a dependency in-lin
 
 The Go community talks of the "accept interfaces, return structs" addage. The meaning of this idiom can be hard to pin down, herein lies some clarity.
 
-"Accept interfaces" speaks to the Dependency Inversion Principle (D in 'SOLID'). When a function declares dependent behavior as an interface parameter it gives control of _how_ that behavior is satisfied to the consuming logic. This is the foundation of flexible code.
+"Accept interfaces" promotes flexible and loosely coupled code. When a function declares dependent behavior as an interface parameter it gives control of _how_ that behavior is satisfied to the consuming logic.
 
 "return structs" is less obvious. It, generally, speaks to initializing constructors where authors should return the struct being constructed as opposed to interface it satisifes. This is largely due to duck typing (["If it walks like a duck and it quacks like a duck, then it must be a duck"](https://en.wikipedia.org/wiki/Duck_typing)).
 
--- a/docs/developer/best-practice.md
+++ b/docs/developer/best-practice.md
@@ -1,0 +1,308 @@
+# Development Best Practice
+
+Software development is like nature: it's in a constant state of flux and continuously changing. As software engineers, it's our responsibility to embrace the flux and write clean code that is _optimized for change_. In doing so, we can rapidly innovate and satisfy customer needs.
+
+## Guiding principles
+
+These principles provide a guide for designing software. At times they may seem contradictory, but they ultimately aim to produce the same kind of code: _simple_ and _maintainable_.
+
+> "Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it." - _Brian W. Kernighan_
+
+- [SOLID](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#single-responsibility-principle) - The Single-responsibility Principle, Open-closed Principle, Liskov Substitution Principle, Interface Segregation Principle and Depdendency Inversion Principle collectively define a guide for writing maintainable code. While often referenced in the context of an object-oriented language, they are applicable to all languages.
+- [LoD](https://en.wikipedia.org/wiki/Law_of_Demeter) - The Law of Demeter tells us a construct should talk to their direct dependencies and _only_ their direct dependencies (don't talk to strangers). Reaching to transitive dependencies creates complex layers of interaction that drive toward [spaghetti code](https://en.wikipedia.org/wiki/Spaghetti_code).
+- [KISS](https://people.apache.org/~fhanik/kiss.html) - "Keep it simple, stupid" was coined by the US Navy. Systems should be designed as simply as possible.
+- [YAGNI](https://martinfowler.com/bliki/Yagni.html) - "You aren't gonna need it" if you don't have a concrete use-case, don't write it.
+- [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - "Don't repeat yourself" suggests you should preference code re-use over duplication. However, you should [avoid hasty abstractions](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction) as the wrong abstraction can be extremely costly to correct. Instead, lean into small amounts of duplication to help identify the right abstractions through multiple use cases and define the pathway to DRY design.
+
+> "Leave code better than you found it"
+
+Generally, you want to fix problems as they're identified. This includes micro refactors to accommodate new code that causes existing abstractions to, possibly, no longer make sense. Resist bolting on new code that is uncomplimentary to existing code.
+
+Finally, we have delivery commitments. Larger more intentional refactors may be required to fix a problem you've identified and attempting to tackle it may delay your deliverable. In such a case, write down your ideas and start a conversation with the wider team. 
+
+## Defining _"clean code that is optimized for change"_
+
+What constitutes clean code? We want simple and maintainable, but how do we define that? The following best practices offer niche advice. Collectively, it represent our definition of clean code.
+
+### State
+
+> Shared mutable state is believed by many to be the “root of all evil”, or at least the cause of most of the accidental complexity in our code. And “Complexity is the root cause of the vast majority of problems with software today.” - [Mauro Bieg](https://mb21.github.io/blog/2021/01/23/pure-functional-programming-and-shared-mutable-state.html#the-root-of-all-evil)
+
+Package level state is rarely required. If it is, it should be composed of re-usable constructs defined and exposed from the package itself.
+
+In avoiding package level state we avoid [hidden dependencies and unintended side effects](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_package_level_state) stemming from global state mutation. Model loosely-coupled components by [declaring your dependencies](#dependencies).
+
+### Names
+
+##### Packages
+
+Package names help form the first impression of what a package provides and is used to reference the public API of the package. A good package name brings about clarity and purpose for consumers. Package names should be concise and represent the behavior the package provides not the types it contains. 
+
+Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Decorator`. You may find you have a type named after the package, soemtimes it's appropriate.
+
+If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name.
+
+Avoid [relentlessly] `types`, `interfaces`, `common`, `util` or `base` packages. These names offer no indication on what the package provides beyond 'a bundle of stuff'. With time, these types of package turn into dumping grounds containing unrelated logic that can be the root cause for cyclic dependencies.  If a utility style package is required, make it specific. For example a `/utils/cmdutil` package may provide command utility functions.
+
+##### Function & method
+
+Functions and methods (functions with a receiver parameter) should adequetly describe their behavior. Generally, they should follow a verb-noun form. The types of parameters should help inform the function name.
+
+##### Variable
+
+Variable names should be concise and descriptive. Prefer single word names. The further away from the site of declaration a variable is used, the more descriptive it needs to be. For example, a variable named `n` used 30 lines after declaration makes it unnecessarily difficult to reason what it represents at the site of use.
+
+### <a name="dependencies"></a> Dependencies
+
+Functions should explicitly declare their dependencies as parameters. In doing so, they honor the [dependency inversion principle](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle) and provide the necessary flexibility to write loosely coupled programs.
+
+```go
+func FastenPlanks(tools Toolbox, surface Surface, planks Planks) error {
+    for ; planks.Count() > 0;  {
+        if err := tools.Hammer(surface, planks.Next()); err != nil {
+            return err
+        }
+    }
+}
+```
+
+Avoid constructing dependencies in-line. If you do construct a dependency in-line because it compliments the most common usage, provide a way to override it such as with a setter.
+
+##### "Accept interfaces, return structs"
+
+The Go community talks of the "accept interfaces, return structs" addage. The meaning of this idiom can be hard to pin down, herein lies some clarity.
+
+"Accept interfaces" speaks to the Dependency Inversion Principle (D in 'SOLID'). When a function declares dependent behavior as an interface parameter it gives control of _how_ that behavior is satisfied to the consuming logic. This is the foundation of flexible code.
+
+"return structs" is less obvious. It, generally, speaks to initializing constructors where authors should return the struct being constructed as opposed to interface it satisifes. This is largely due to duck typing (["If it walks like a duck and it quacks like a duck, then it must be a duck"](https://en.wikipedia.org/wiki/Duck_typing)).
+
+> Interfaces in Go provide a way to specify the behavior of an object: if something can do this, then it can be used here. - [Effective Go](https://go.dev/doc/effective_go#interfaces_and_types)
+
+Factory functions are not to be confused with intializing constructors. Factory functions typically return abstractions (an interface) satisfied by 1 of several implementations. A factory could be as simple as:
+
+```go
+// Foo and Bar satisfy fmt.Stringer
+func NewStringer(key string) (fmt.Stringer, error) {
+    switch key {
+    case "foo":
+        return Foo{}, nil
+    case "bar":
+        return Bar{}, nil
+    default:
+        return nil, errors.New("unknown key")
+    }
+}
+```
+
+Factories generally shouldn't contain complex construction logic. That's the job of the initializing constructor for the concrete type.
+
+##### Loggers
+
+Loggers are sometimes treated as a special depency by making entire projects depend on a package level logger. This _masks_ a crucial dependency that is hard to change without impacting the rest of the code base. To understand why this is bad, consider extracting a package into its own repository for reusability. Depending on a package level logger, or any other package level dependencies, makes the extraction process unnecessarily difficult.
+
+Logger's are not special dependencies, inject them like any other dependency. If you want to minimize the surface area of the default initializing constructor (`New...` funcs) you may construct noop loggers to satisfy the logger dependency and provide a means to override the logger with a setter.
+
+### Testing
+
+Software development is costly and hazardous. Tests provide example usage, guard rails, and assurances. Without tests, you cannot have high quality code. For this reason, all code we spend time designing and writing is worth testing. Tests should:
+
+- focus on the public API of a packge.
+- use the `_test` idiom for testing files.
+- treat the package as a black box.
+- cover at least 80% of the public API.
+- named after the entity they're testing; for example, `TestFooBarDependencyErrors` tests that an error occurs when `Foo` is executed and its `Bar` dependency errors.
+
+Testable code should:
+
+- isolate or provide overrides such as setter funcs for hard to test concerns like io, time and concurrency.
+- expose a flexible public API.
+
+If you public API is hard to test consumers will find it hard to use - rethink your design.
+
+##### `_test` idiom
+
+The Go compiler allows test files, files suffixed with `_test.go`, to reside in an `_test` package. For example:
+
+```go
+// foo.go
+package foo
+
+// implementation
+```
+
+```go
+// foo_test.go
+package foo_test
+
+// test cases
+```
+
+The compiler will build files that declare a package with the suffix "_test" as a separate package, then link and run with the main test binary. This ensures tests cases only have access to the public API of a package.
+
+##### Necessraily complex and unexported functions
+
+If you have attempted to break a function down into singular responsibilities and found its best to maintain the necessary complexity as a single unexported function it may be appropriate to white box test.
+
+When testing a _necessraily complex_ function isolate the tests to a separate testing file. For example `necessarily_complex_func_wb_test.go` where `wb` stands for white box.
+
+### Errors
+
+When generating an error, concisely describe the problem that occurred. Do not prefix the message with the word 'error' instead leaving that to a presentation layer of code. Use lower case unless describing a type the _user_ is aware of (generally a user is type and parameter unaware). 
+
+When handling an error, don't log it only to pass it up the call stack. Logging and passing the received error up the call stack typically results in lots verbose and unhelpful logging about the same error.
+
+If appropriate, add context by wrapping the error using `fmt.Errorf` or equivilent. Only add context that is known by the scope of your function and do not _assume_ an error represents something unless you can make a positive assertion. This avoids embedding knowledge that makes re-use and refactors harder. For example:
+
+```go
+// foo knows bar can return errZero so can perform an explicit check.
+func foo() error {
+    err := bar(0)
+    if errors.Is(err, errZero) {
+        return fmt.Errorf("ambiguous: %v", err)
+    }
+
+    if err != nil {
+        return err
+    }
+}
+
+var errZero = errors.New("zero value")
+
+func bar(i int) error {
+    if i > 5 {
+        return errors.New("out of bounds")
+    }
+
+    if i == 0 {
+        return errZero
+    }
+}
+```
+
+### Boolean parameters
+
+Boolean parameters are widely considered an antipattern. Context is hugely important when assessing a boolean parameter but in general they represent an on/off switch for behavior. The behavioral on/off switch creates hard to understand and maintain logic that often has poor re-use and is an indicator that a piece of code violates the SRP. 
+
+Instead, consider structuring your code to have dynamically constructable behavior logic. This allows you to isolate booleans to construction logic, a distinct concern, and construct the behavior you require for your context. 
+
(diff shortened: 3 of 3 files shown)


DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+Variable names should be concise and descriptive. Prefer single word names. The further away from the site of declaration a variable is used, the more descriptive it needs to be. For example, a variable named `n` used 30 lines after declaration makes it unnecessarily difficult to reason what it represents at the site of use.
+
+### Abstractions
+
+- Functions & Methods
+- Constructs
+- Interfaces
+
+### <a nane="dependencies"></a> Depencencies
```
ewollesen, line 55:
```suggestion
### <a name="dependencies"></a> Dependencies
```

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+- Accept interfaces, return structs
+- Declare your dependencies
+
+### Package API
+
+- this might be wrapped up in naming and testing. Do we have more explicit points?
```
ewollesen, line 62: You could always just link https://go.dev/blog/package-names I guess. :D

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Testing
+
+- test using the `_test` idiom
```
ewollesen, line 66: A reference to https://pkg.go.dev/testing might be helpful.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+- Use test to help identify if the package is easy to use for consumers
+- isolating concerns like io
```
ewollesen, line 68: ...and time! Time is a big one in tests in my experience.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+- Should be small
+- Preferably single method
+
+### Errors
+
+- Describe the action that failed
+- Use lower case
+- Don't prepend error
+- Don't overshare information 
+- Add context if you can
```
ewollesen, line 85: - wrap an underlying error when appropriate (e.g. when they provide more information)

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+### Disambiguating context.Context
+
+- Represents the context of execution
+    - Useful for tear down
+- There should be 1 and only 1 context
+- Should be the first parameter on the primary call path
+
+### Loggers
+
+- They're just another dependency, inject them.
```
ewollesen, line 107: Just an aside I guess, but I see a lot of package-level loggers in k8s code. I sure don't care for that, so I'm glad to see this.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+- Packages _should_ contain package level documentation
+- What makes a good comment?
+
+### Types
+
+- I forgot what this was for...
+
+### Variable initialization
+
+- What forms to preference under what circumstances
```
ewollesen, line 122: The guideline I remember reading was to use `var x int` when you aren't sure if you'll ever assign something to the value. Use `x := 0` when you _know_ you will. Applies more to slices and structs than ints I think, but int was shorter/easier to type.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+- If your API leverages them, inject them
+- Document channel behavior
+
+### Returns
+
+- Return early
+- Naked returns should be avoided
+
+### Panicing
```
ewollesen, line 134:
```suggestion
### Panicking
```

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+- [SOLID](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#single-responsibility-principle) - The Single-responsibility Principle, Open-closed Principle, Liskov Substitution Principle, Interface Segregation Principle and Depdendency Inversion Principle collectively define a guide for writing maintainable code. While often referenced in the context of an object-oriented language, they are applicable to all languages.
+- [LoD](https://en.wikipedia.org/wiki/Law_of_Demeter) - The Law of Demeter tells us a construct should talk to their direct dependencies and _only_ their direct dependencies. Reaching to transitive dependencies creates complex layers of interaction that drive toward [spaghetti code](https://en.wikipedia.org/wiki/Spaghetti_code).
+- [KISS](https://people.apache.org/~fhanik/kiss.html) - "Keep it simple, stupid" was coined by the US Navy. Systems should be designed as simply as possible.
+- [YAGNI](https://martinfowler.com/bliki/Yagni.html) - "You aren't gonna need it" if you don't have a concrete use-case, so don't write it.
```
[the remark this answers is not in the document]
chrisdoherty4: Not sure I agree with this one. Reading it with the suggestion seems odd.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+Package level state is rarely required. If it is, it should be composed of re-usable constructs defined and exposed from the package itself.
+
+In avoiding package level state we avoid [hidden dependencies and unintended side effects](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_package_level_state) stemming from global state mutation. Model loosely coupled components by [declaring your dependencies](#dependencies).
+
+### Names
+
+##### Packages
+
+Package names help form the first impression of what a package provides and is used to reference the public API of the package. A good package name brings about clarity and purpose for consumers. Package names should be concise and represent the behavior the package provides not the types it contains. For example, `meals` provides `dr
```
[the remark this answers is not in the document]
chrisdoherty4: Yes

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Dectorator`.
+
+If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name.
+
+Avoid (like the plague) `types`, `interfaces`, `common`, `util` or `base` packages. They don't represent anything cohesive and tend to collect unrelated types and algorithms that often require teething apart as a project progresses due to cyclic dependencies. If a utility style package is required, make it specific. For example a `/utils/cmdutil` package may provide command utility functions.
```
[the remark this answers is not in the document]
chrisdoherty4: 🤦🏻 thanks

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Yea I don't personally like the `klog` thing (and some other bits) Kubernetes does. Its the easy way out of managing a significant dependency (the logger). I've worked on projects, and you see it frequently in Go projects, where you let the code grow organically and decide to extract some code to a re-usable repository only to find there are package level dependencies 🤦🏻 .

I'll write as much here as an elaboration to 'why'.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,145 @@
...
+
+## "clean code that is optimized for change"
+
+What constitutes clean code? We want simple and maintainable, but how do we define that? The following best practices offer niche advice. Combined, they represent our definition of clean code.
+
+### State
+
+> Shared mutable state is believed by many to be the “root of all evil”, or at least the cause of most of the accidental complexity in our code. And “Complexity is the root cause of the vast majority of problems with software today.” - [Mauro Bieg](https://mb21.github.io/blog/2021/01/23/pure-functional-programming-and-shared-mutable-state.html#the-root-of-all-evil)
+
+Package level state is rarely required. If it is, it should be composed of re-usable constructs defined and exposed from the package itself.
```
chrisdoherty4, line 25: And unexported with the 1 exception of not being able to specify a struct as a const

DEVELOPMENT_BEST_PRACTICE.md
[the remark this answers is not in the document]
chrisdoherty4: Could drop the 'so' entirely?

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+
+Avoid constructing dependencies in-line. If you do construct a dependency in-line because it compliments default usage, provide a way to override it such as with a setter.
+
+#### Loggers
+
+Loggers are sometimes treated as a special depency by making entire projects depend on a package level logger. This _masks_ a crucial dependency that is hard to change without impacting the rest of the code base. To understand why this is bad, consider extracting a package into its own repository  for reusability. Depending on a package level logger, or any other package level dependencies, makes the extraction process unnecessarily difficult as you first need to pull the hidden dependency apart.
+
+Logger's are not special dependencies, inject them like any other dependency. If you want to minimize the surface area of the default initializing constructor (`New...` funcs) you may construct noop loggers to satisfy the logger dependency and provide a means to override the logger with a setter.
```
jacobweinstock, line 80: "inject" makes me think about dependency injection. Youre not suggested that here are you? might think about a synonym?

chrisdoherty4: That's exactly what I'm suggesting :).

chrisdoherty4: See https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_package_level_state and

jacobweinstock: Ah yes. I'm very much on the same page here. I believe my previous definition of dependency injection was incorrect.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,194 @@
...
+
+Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Decorator`. You may find you have a type named after the package, soemtimes it's appropriate.
+
+If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name.
+
+Avoid (like the plague) `types`, `interfaces`, `common`, `util` or `base` packages. They don't represent anything cohesive and tend to collect unrelated types and algorithms that often require teething apart as a project progresses due to cyclic dependencies. If a utility style package is required, make it specific. For example a `/utils/cmdutil` package may provide command utility functions.
+
+##### Function & method
+
+Functions and methods (functions with a receiver parameter) should adequetly describe their behavior. Generally, they should follow a verb-noun form. The types of parameters should help inform the function name.
```
chrisdoherty4, line 43: Parameter types help inform on the functions scope/purpose, not necessarily its name.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,228 @@
...
+    case "foo":
+        return Foo{}, nil
+    case "bar":
+        return Bar{}, nil
+    default:
+        return nil, errors.New("unknown key")
+    }
+}
+
+type Foo struct {}
```
chrisdoherty4, line 95: Replace with a note on the func for example brevity.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,304 @@
...
+- [KISS](https://people.apache.org/~fhanik/kiss.html) - "Keep it simple, stupid" was coined by the US Navy. Systems should be designed as simply as possible.
+- [YAGNI](https://martinfowler.com/bliki/Yagni.html) - "You aren't gonna need it" if you don't have a concrete use-case, don't write it.
+- [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - "Don't repeat yourself" suggests you should preference code re-use over duplication. However, you should [avoid hasty abstractions](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction) as the wrong abstraction can be extremely costly to correct. Instead, lean into small amounts of duplication to help identify the right abstractions through multiple use cases and define the pathway to DRY design.
+
+> "Leave code better than you found it"
+
+Generally, you want to fix problems as they're identified. This includes micro refactors to accommodate new code that causes existing abstractions to, possibly, no longer make sense. Resist bolting on new code that is uncomplimentary to existing code.
```
ewollesen, line 19: This may warrant a comment on how it's great to make this change, but probably in a different PR. Like code, PRs should aim to follow the SRP. :)

chrisdoherty4: Yea good callout. Basically commentary on _how_ to do this.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,304 @@
...
+
+Factories generally shouldn't contain complex construction logic. That's the job of the initializing constructor for the concrete type.
+
+##### Loggers
+
+Loggers are sometimes treated as a special depency by making entire projects depend on a package level logger. This _masks_ a crucial dependency that is hard to change without impacting the rest of the code base. To understand why this is bad, consider extracting a package into its own repository for reusability. Depending on a package level logger, or any other package level dependencies, makes the extraction process unnecessarily difficult.
+
+Logger's are not special dependencies, inject them like any other dependency. If you want to minimize the surface area of the default initializing constructor (`New...` funcs) you may construct noop loggers to satisfy the logger dependency and provide a means to override the logger with a setter.
+
+### Testing
```
danbudris, line 105: could be worth noting when to implement functional/end-to-end tests in this section

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,304 @@
...
+## References
+
+These references are the basis for this document.
+
+- Practical Go - Dave Cheney - https://dave.cheney.net/practical-go/presentations/gophercon-israel.html
+- Best Practices - Peter Bourgon - https://peter.bourgon.org/go-best-practices-2016/
+- Effective Go - The Go Team - https://go.dev/doc/effective_go
+- Code Review Comments - The Go Community - https://github.com/golang/go/wiki/CodeReviewComments
+
+## To do
```
danbudris, line 276: would love to see more about logging -- when, what, writing concise/clear log messages; really important for building for operations, especially since we'll be debugging a lot of this stuff 'blind' without access to the cluster in most cases

danbudris (approved):

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Yea great point. I've written a 1 pager on logging previously because of how important it is. I'll add another section in the to do.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,304 @@
...
+- [YAGNI](https://martinfowler.com/bliki/Yagni.html) - "You aren't gonna need it" if you don't have a concrete use-case, don't write it.
+- [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - "Don't repeat yourself" suggests you should preference code re-use over duplication. However, you should [avoid hasty abstractions](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction) as the wrong abstraction can be extremely costly to correct. Instead, lean into small amounts of duplication to help identify the right abstractions through multiple use cases and define the pathway to DRY design.
+
+> "Leave code better than you found it"
+
+Generally, you want to fix problems as they're identified. This includes micro refactors to accommodate new code that causes existing abstractions to, possibly, no longer make sense. Resist bolting on new code that is uncomplimentary to existing code.
+
+Finally, we have delivery commitments. Larger more intentional refactors may be required to fix a problem you've identified and attempting to tackle it may delay your deliverable. In such a case, write down your ideas and start a conversation with the wider team.
```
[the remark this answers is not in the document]
g-gaston: I would vote against mentioning "creating tickets". Compromising on quality and postponing refactors is one answer to the deliverable problem, but in my opinion, it shouldn't be the default one. I'm not saying that's what you were proposing, but I think mentioning in this doc can give the wrong idea.

I think the key part of this paragraph is "start a conversation with the wider team" (maybe we could expand on it?). We should try to always talk when this kind of thing happen, both to be able to find the right compromise (usually quality, time or scope) for each situation (one the team feels comfortable with) and to be able to detect underlying issues.

Defaulting to just postponing the work can hide other issues:
* We are underestimating tasks. This is quite common and perfectly normal, I'm confident we all do it, at least I definitely do it. But the only way to get better at it is to recognize when it happens and learn from it. Plus, I think we have to get used to accommodate some "rework" in our estimations. The codebase has grown to a point where almost nothing is greenfield anymore, so improving it as we go should be an intrinsic part of our day to day work.
* We produce "unclean" code. When that happens, it's usually difficult to maintain and expand, hence requiring significant refactors on every new feature. I personally don't think this has been a major problem for us but we should stay on top of it.
* We are accumulating too much tech debt. This has similar symptoms as the previous one but it's caused by little inoffensive changes accumulated over time.

Moreover, always postponing the work is unsustainable. The more we do it, the more tech debt we accumulate. And the more tech debt we accumulate, the more refactor will be needed to add new features, which feeds back into the same loop. This feedback loops will continue to the point where we start spending more time on refactors than on improving the product.

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: My advice on this is usually: if the refactor idea can stand on itself, create a separate PR with just that.

But how to decide if the idea stands on its own? A somehow ok rule of thumb: if you would write a separate commit for the refactor, then it might be worth its own PR.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+Functions and methods (functions with a receiver parameter) should adequetly describe their behavior. Generally, they should follow a verb-noun form. The types of parameters should help inform the function name.
+
+##### Variable
+
+Variable names should be concise and descriptive. Prefer single word names. The further away from the site of declaration a variable is used, the more descriptive it needs to be. For example, a variable named `n` used 30 lines after declaration makes it unnecessarily difficult to reason what it represents at the site of use.
+
+### <a name="dependencies"></a> Dependencies
+
+Functions should explicitly declare their dependencies as parameters. In doing so, they honor the [dependency inversion principle](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle) and provide the necessary flexibility to write loosely coupled programs.
```
g-gaston, line 57: This is totally correct but, reading it, gives me the impression that defining dependencies as arguments is enough to follow dependency inversion (even though you are not saying that explicitly).

Maybe this is just me (in that case ignore me), but I would maybe specify that doing this is necessary when following Dep Inversion but it's not enough.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+}
+```
+
+Avoid constructing dependencies in-line. If you do construct a dependency in-line because it compliments the most common usage, provide a way to override it such as with a setter.
+
+##### "Accept interfaces, return structs"
+
+The Go community talks of the "accept interfaces, return structs" addage. The meaning of this idiom can be hard to pin down, herein lies some clarity.
+
+"Accept interfaces" speaks to the Dependency Inversion Principle (D in 'SOLID'). When a function declares dependent behavior as an interface parameter it gives control of _how_ that behavior is satisfied to the consuming logic. This is the foundation of flexible code.
```
g-gaston, line 75: Repeating myself here a bit but: using interfaces in go is key for the Dependency Inversion Principle, but not enough. What that interface represents and the signature of its methods play a key role as well.

Maybe this is something we can elaborate more on the interfaces section: using an interface is not enough to achieve loosely coupling.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+}
+```
+
+Factories generally shouldn't contain complex construction logic. That's the job of the initializing constructor for the concrete type.
+
+##### Loggers
+
+Loggers are sometimes treated as a special depency by making entire projects depend on a package level logger. This _masks_ a crucial dependency that is hard to change without impacting the rest of the code base. To understand why this is bad, consider extracting a package into its own repository for reusability. Depending on a package level logger, or any other package level dependencies, makes the extraction process unnecessarily difficult.
+
+Logger's are not special dependencies, inject them like any other dependency. If you want to minimize the surface area of the default initializing constructor (`New...` funcs) you may construct noop loggers to satisfy the logger dependency and provide a means to override the logger with a setter.
```
g-gaston, line 103: I'm not against this at all. But due to the fact that all the cli code violates this advice, if we want to push for it we should probably pair it with a proposal on how to implement this on new code (eg. inject from the dependency factory to the constructors...) and a plan to little by little refactor existing code.

We should definitely do this separately to this PR but maybe worth adding a note here.

I get this doc aims to help us write better code in general and not only for the CLI. But since that's the piece of software we spent most time with, I figured it might be a good idea to bridge that gap.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Testing
+
+Software development is costly and hazardous. Tests provide example usage, guard rails, and assurances. Without tests, you cannot have high quality code. For this reason, all code we spend time designing and writing is worth testing. Tests should:
+
+- focus on the public API of a packge.
+- use the `_test` idiom for testing files.
+- treat the package as a black box.
+- cover at least 80% of the public API.
```
g-gaston, line 112: clarification: testing 80% of the public functions/methods (eg. just calling those methods) or covering 80% of the code accessible through public APIs?

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Concurrency
+
+Concurrency introduces harder to rationalize execution relative to serialized programs and is easy to get wrong. Concurrency doesn't necessarily speed up a program either. With these points in mind, prefer synchronous APIs instead leaving concurrency to the caller. Not only does this eliminate a whole host of problems, it ensures your APIs are predictable and easier to use.
+
+Dave Cheney has made some excellent [comments](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#concurrency) on concurrency in Go.
+
+#### Channels
+
+If you find the need for channels prefer accepting them as parameters. Accepting channels as a parameter allows consumers to specify channel properties, such as buffered or unbuffered, creating a more flexible API.
```
g-gaston, line 198: Clarification: channels for input, channels for input or both?
I would agree with channels for input, I would disagree with channels for output. At least as general advice.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+Clearly document channel behavior. For example, what does your algorithm do if it can't send data on a channel? Will it drop data, block until the channel is writable, or something else?
+
+Avoid using channels for data mutation operations where a traditional mutex would suffice. 
+
+###  Comments
+
+Functions and variables should be named so that their purpose is clear. However, good naming and well structured code isn't justification to not document APIs. All public package APIs should be documented including those that may seem trivial. This, at minimum, helps with code documentation generation. Additionally, each package should have a clear purpose that can be documented with examples using the [package documentation feature of Go](https://go.dev/doc/effective_go#commentary).
+
+Through good documentation that we reflect on and update regulalry, we can clearly communicate with maintainers what the original intent is making decisions easier when new capabilities are required.
```
g-gaston, line 210:
```suggestion
Through good documentation that we reflect on and update regularly, we can clearly communicate with maintainers what the original intent is making decisions easier when new capabilities are required.
```

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+Clearly document channel ownership thereby documenting who is responsible for closing it. If there are multiple writers, no-one can close it therefore no-one owns it. Note channels don't _need_ to be closed, they will be garbage collected once the necessary scoping conditions are met.
+
+Clearly document channel behavior. For example, what does your algorithm do if it can't send data on a channel? Will it drop data, block until the channel is writable, or something else?
+
+Avoid using channels for data mutation operations where a traditional mutex would suffice. 
+
+###  Comments
+
+Functions and variables should be named so that their purpose is clear. However, good naming and well structured code isn't justification to not document APIs. All public package APIs should be documented including those that may seem trivial. This, at minimum, helps with code documentation generation. Additionally, each package should have a clear purpose that can be documented with examples using the [package documentation feature of Go](https://go.dev/doc/effective_go#commentary).
```
g-gaston, line 208: Note: If everyone agrees with this, we should probably look into adding some linter to prevent new public entities/functions from being added without doc comments.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+    return nil
+}
+```
+
+In returning early, we also create clearer code that has lower [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) that can be the cause for bugs.
+
+Never use naked returns. Go allows return types to be specified with an identifier. The identifier is a scoped variable in the context of the function and its value can be set during program execution. Doing so, however, often creates unnecessraily difficult to rationalize logic.
+
+Avoid named returns. If you don't specify names for your return values you can't accidentally use naked returns.
```
g-gaston, line 232: I disagree with this. I think named returns are very useful when returning things like `(string, string)` and I think we should encourage them.

I would agree with avoiding them when unnecessary (eg. `ReadFileName() (string, error)`). Also 100% agree with "Never use naked returns".

g-gaston (commented):
I love this

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: I'd be keen to call out a separate section for testing types. This topic can be quite large and often involves turmoil due to a lack of definition around what constitutes an integration test vs unit vs e2e etc.

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Names
+
+##### Packages
+
+Package names help form the first impression of what a package provides and is used to reference the public API of the package. A good package name brings about clarity and purpose for consumers. Package names should be concise and represent the behavior the package provides not the types it contains. 
+
+Avoid stutter in struct and interface names by treating the package name as a namespace and considering how the types are consumed. For example, `drink.DrinkDecorator` should be `drink.Decorator`. You may find you have a type named after the package, soemtimes it's appropriate.
+
+If you find yourself using an appropriate package name that's commonly used as a variable, consider pluralising the package name.
```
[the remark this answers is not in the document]
chrisdoherty4: Sure. I'll add examples from the stdlib also.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Its intentionally worded as 'honor'ing not satisfying for the reasons you purport. Perhaps there's a better word here that will clarify it, or perhaps the back half of the sentence is misleading.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: I'm not sure I agree. Dependency inversion can be achieved with an interface parameter. I don't think we should complicate it beyond that. Whether your _design_ is providing increased flexibility is related but different. Your interface designs speaks to the _degree_ of flexibility but the degree of flexibility goes beyond dependency inversion.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Definitely. I think the CLI has lots of flaws, actually, that need tidying up - this is 1 of those flaws.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Good point. 80% line coverage. Your code should be flexible enough to enable that. I've never had a problem and even strived for higher, 90%. With respect to the public API, we probably want something like 90%+ covered.

I think its worth calling out I chose this number based on personal experience. We don't have to stick to it but I'm definitely advocating for it.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Both. Even more so for output. As the doc notes, allowing consumers to pass the channel lets them control buffering which, as a consumer, is hugely important if you have an API experiencing bursts while you try to provide a more consistent throughput from your API. 

We should also add a note about conditionality of the channel.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Thats why this is `avoid`. Common sense needs to be used and if you have a function that returns the same type twice then clearly named returns help clarify whats what. Thats pretty much the only case I thought about that warrants named returns and caused my to choose the wording 'avoid'.

DEVELOPMENT_BEST_PRACTICE.md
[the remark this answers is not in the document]
danbudris: yeah no I get you, I'm mostly just saying that when we 'write it down somewhere' as mentioned explicitly above that place should often be a github issue

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: Totally, as I said, your wording is completely accurate. But I just wonder if I know this because I already knew the difference between one and the other.

One option is to add something like:

```suggestion
Functions should explicitly declare their dependencies as parameters. In doing so, they honor (but don't completely satisfy) the [dependency inversion principle](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle) and provide the necessary flexibility to write loosely coupled programs.
```

Maybe this is just me, if no one else comments about this, feel free to ignore me.

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: My bad, I think I didn't explain myself correctly. I might do better with an example:

```go
# mysql/auth.go

type Credentials struct {
   User, Password string
}

func (a *Authenticator) Authenticate(ctx context.Context, creds Credentials) error {...}


# user/registry.go

type DBConnector interface {
    Authenticate(ctx context.Context, creds Credentials) error
}

type Registry struct {
   dBConnector DBConnector
}

func NewRegistry(dBConnector DBConnector) *Registry {...}

func (r *Registry) Start(ctx context.Context) error {
   ...
   err := r.Authenticate(ctx, mysql.Credentials{User: user, Password: password})
   ...
}
```

Here, the registry is taking its dependency as an interface, however, it doesn't achieve dependency inversion because it still depends on a type from the concrete implementation. This is what I meant that just using an interface is not enough. You need to design the interface appropriately, for example:

```go
# mysql/auth.go
func (a *Authenticator) Authenticate(ctx context.Context, user, password string) error {...}


# user/registry.go
type DBConnector interface {
    Authenticate(ctx context.Context, user, password string) error
}

func (r *Registry) Start(ctx context.Context) error {
   ...
   err := r.Authenticate(ctx, user, password)
   ...
}
```

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: If this gets approved (which I do), let's talk about adding a rule to codecov.

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: If you feel strongly about this, go ahead, I'm ok with it 🙂 

However, I don't necessarily agree this should always be the answer. What you are proposing does in fact offer a more flexible API but in some cases the price to pay would be a more complicated implementation.

To you example, burst smoothing can be easily implemented (kind of as a decorator) even if the concurrent code method returns a non buffered channel. I'm not saying it has to be that way, just that imo this is more a tradeoff topic than a clear rule of thumb.

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: What would you think about replacing "avoid" with something like:
```suggestion
Use named returns only if strictly necessary for disambiguation (for example, a function that returns the same type twice). If you don't specify names for your return values you can't accidentally use naked returns. 
```

DEVELOPMENT_BEST_PRACTICE.md
```diff
@@ -0,0 +1,308 @@
...
+
+Avoid [relentlessly] `types`, `interfaces`, `common`, `util` or `base` packages. These names offer no indication on what the package provides beyond 'a bundle of stuff'. With time, these types of package turn into dumping grounds containing unrelated logic that can be the root cause for cyclic dependencies.  If a utility style package is required, make it specific. For example a `/utils/cmdutil` package may provide command utility functions.
+
+##### Function & method
+
+Functions and methods (functions with a receiver parameter) should adequetly describe their behavior. Generally, they should follow a verb-noun form. The types of parameters should help inform the function name.
+
+##### Variable
+
+Variable names should be concise and descriptive. Prefer single word names. The further away from the site of declaration a variable is used, the more descriptive it needs to be. For example, a variable named `n` used 30 lines after declaration makes it unnecessarily difficult to reason what it represents at the site of use.
```
jiayiwang7, line 53: Is this doc serving for general Go practice or EKS Anywhere best practice?

If latter, can we expand a bit for recommended eks-a naming convention? e.g. long vs short:`kubeadmControlPlane` vs `kcp`; capitalize acronym or not: `capi` vs `CAPI`, `ip` vs `IP`, etc.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: I think that example alludes to different degrees of coupling which, while related, isn't strictly required to satisfy the dependency inversion principle. At the most basic level, dependency inversion can still be achieved for the purpose of testability by accepting an interface parameter. That interface may provide behavior that accepts types associated with a specific package. Yes, its advantageous to only use standard library types on interfaces and when you model complete abstractions a set of shared types but that again starts to delve into software design.

Given the contention I think I'll alter the wording by replacing any reference to DIP with 'loose coupling' or similar. The DIP commentary isn't necessary to convey the point.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Its both but its scope only extends to EKS-A :). 

Re naming, we can expand the variable naming section to elaborate on recommendations. We should probably end with 'if in doubt, prefer clarity over brevity'.

DEVELOPMENT_BEST_PRACTICE.md
g-gaston: Sounds good, I'm ok with removing the DIP reference

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Added the todo.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: I'm making notes on what we need to follow up on and have included this. I'll create a well written issue describing the problem and referencing this guide once its merged.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Addressed.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Addressed.

DEVELOPMENT_BEST_PRACTICE.md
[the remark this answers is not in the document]
chrisdoherty4: Addressed.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Addressed.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Addressed.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Made a note.

DEVELOPMENT_BEST_PRACTICE.md
chrisdoherty4: Addressed.

g-gaston (approved):
Love the index!

PR #1922, opened 2022-04-22
tweak vsphere network preflight check

jaxesn:
*Issue #, if available:*

*Description of changes:*
In our dev environment there are now two networks with the same name.  The current network check just checks for the response to be non-nil, but in the case where there are two with the same name they are both returned.  This ends up causing an error later in the process when capv tries to create the vmwares with the following error:

```
    - lastTransitionTime: "2022-04-21T19:50:57Z"
      message: 'error getting network specs for "infrastructure.cluster.x-k8s.io/v1beta1,
        Kind=VSphereVM eksa-system/jgw-vsphere-etcd-p4v8d": unable to find network
        "/SDDC-Datacenter/network/sddc-cgw-network-1": path ''/SDDC-Datacenter/network/sddc-cgw-network-1''
        resolves to multiple networks'
```

This changes the call to govc to return the response in json and then parses it to ensure there is 1 and only 1 result.

*Testing (if applicable):*


pkg/executables/govc.go
```diff
@@ -618,12 +618,19 @@ func (g *Govc) NetworkExists(ctx context.Context, network string) (bool, error)
...
 
-		if networkResponse.String() == "" {
+		foundNetworks := make([]string, 0)
+		if err = json.Unmarshal(networkResponse.Bytes(), &foundNetworks); err != nil {
+			logger.V(2).Info(fmt.Sprintf("Failed unmarshalling govc response: %s, %v", networkResponse.String(), err))
+			exists = false
+			return nil
+		}
+
+		if len(foundNetworks) != 1 {
```
g-gaston, line 633: nit: maybe we should return an error if `len(foundNetworks) > 1`?
since returning false in that case is a bit misleading

jaxesn: yeah i would like to change the final user message in these cases from `network not found` to something like `multiple networks found`, but wasnt sure the best way to surface that difference.  Returning an error could serve that need?

g-gaston: Yeah, I think that works for now
We should eventually change this method to something like `SearchNetwork(...) []string` and make the caller decide what's a valid result and what's not. But we can create a ticket for that.

jaxesn: ok ill give that a shot

PR #1923, opened 2022-04-22
Remove internal use of `gitOpts` struct from FluxAddonClient

danbudris:
*Issue #, if available:*

*Description of changes:*
removing internal use of the unneeded `gitOpts` struct from the FluxAddonClient. The whole options argument will be removed in favor of directly injecting dependencies in an upcoming pr.

this is an intermediate step to factoring out the `git.Provider` interface into two parts, a GitClient (responsible for local repo actions) and a GitProvider (responsible for actions on a remote provider), which will be generated by the factory and injected directly into the FluxAddonClient (along with the writer) rather than using the intermediate `gitOpts` struct.

*Testing (if applicable):*
`make`, created a cluster with gitops, upgrade it, and deleted it.
created a cluster w/out gitops.

### DIFF
--- a/pkg/addonmanager/addonclients/fluxaddonclient.go
+++ b/pkg/addonmanager/addonclients/fluxaddonclient.go
@@ -54,7 +54,8 @@ const (
 
 type FluxAddonClient struct {
 	flux    Flux
-	gitOpts *GitOptions
+	writer  filewriter.FileWriter
+	git     git.Provider
 	retrier *retrier.Retrier
 }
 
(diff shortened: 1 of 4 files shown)


pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -95,9 +96,13 @@ func NewGitOptions(ctx context.Context, cluster *v1alpha1.Cluster, fluxConfig *v
 }
 
 func NewFluxAddonClient(flux Flux, gitOpts *GitOptions) *FluxAddonClient {
+	if gitOpts == nil {
```
danbudris, line 99: this is an intermediate step to replicate the existing behavior which will be removed once we completely remove the gitOpts and pass in individual dependencies; a different method for determining wether GitOps install should be skipped will be put in place.

danbudris: by 'existing behavior', which check if the 'gitOpts' is nil to determine wether ot install flux or not; we're now just checking wether FluxAddonClient itself is nil

jiayiwang7 (approved):
/lgtm

pkg/dependencies/factory_test.go
```diff
@@ -79,7 +79,7 @@ func TestFactoryBuildWithMultipleDependencies(t *testing.T) {
 	tt.Expect(deps.Bootstrapper).NotTo(BeNil())
 	tt.Expect(deps.ClusterManager).NotTo(BeNil())
 	tt.Expect(deps.Provider).NotTo(BeNil())
-	tt.Expect(deps.FluxAddonClient).NotTo(BeNil())
+	tt.Expect(deps.FluxAddonClient).To(BeNil())
```
g-gaston, line 82: why this change? the factory is not changing

PR #1927, opened 2022-04-22
Remove registry mirror from download artifacts e2e

mrajashree:
This test doesn't need registry mirror configuration. It's failing currently because it expects those vars to be set but the test name doesn't have the required regex. This PR removes the registry mirror config and so the need for those env vars


jaxesn (approved):
/approve

PR #1928, opened 2022-04-24
move interfaces to consumer site; split GitProvider and GitClient int…

danbudris:
…erface

*Issue #, if available:*

*Description of changes:*

A bit of refactoring to help support a generic git provider.
Here, we're moving the `git.Provider` interface definition that the `FluxAddonClient` uses to the consumer site, and splitting the responsibility into `GitClient` and `GitProviderClient` -- e.g. the local git calls and the interactions with the remote repository. Internally to the `FluxAddonClient` we're making this distinction as well, even if the same object is being used for both at the moment (the `git.Provider`). 

The next step is going to be splitting the `git.Provider` that we're passing in into two separate entities which satisfy the GitClient and GitProviderClient interfaces as defined by the `FluxAddonClient`.

### Interfaces
`GitClient` interface is to be used for provider-agnostic actions on a generic git repository, such as push/pull/init/clone/etc.
This interface will be responsible for any generic actions that you would perform with a `git` CLI executable.

`GitProviderClient` interface is to be used for any provider-dependant actions on a Git repository, such as create a repository in a Git hosting provider like Github. This interface is meant to wrap Git provider APIs (be it GitHub, gitlab, etc).

Once these two interfaces are broken up, the Generic Git Provider will use ONLY a `GitClient`, ignoring the `GitProviderClient` calls; this will be done using a switch on the provider type. Generic actions will be consolidated and called before/after provider-specific logic.

*Testing (if applicable):*


pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -52,11 +52,31 @@ const (
 	deleteClusterconfigCommitMessage  = "Delete commit of cluster configuration; generated by EKS-A CLI"
 )
 
+type GitClient interface {
+	Add(filename string) error
+	Remove(filename string) error
+	Clone(ctx context.Context) error
+	Commit(message string) error
+	Push(ctx context.Context) error
```
jiayiwang7, line 60: Thought you mentioned `GitClient` is for local git operations while `GitProviderClient` for remote ops.
Doesn't Push and Pull interact with remote repo? Or is `GitClient` interface for go-git client while `GitProviderClient` for operations go-git does not handle?

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -52,11 +52,31 @@ const (
...
+}
+
 type FluxAddonClient struct {
-	flux    Flux
-	writer  filewriter.FileWriter
-	git     git.Provider
-	retrier *retrier.Retrier
+	flux              Flux
+	writer            filewriter.FileWriter
+	gitClient         GitClient
```
jiayiwang7, line 77: Wonder what's the reason to split the git provider from the FluxClient level?

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: yes, that's right.

GitClient is for operations that are provider-agnostic (like push/pull/init/clone/etc) while GitProviderClient is for provider dependent operations (like create repo)

jiayiwang7: If it's the latter, wonder if we can name  `GitClient` differently like `GoGitClient`

danbudris: well I want it to be a generic client for any provider-agnostic Git actions, akin to the git CLI executable, so I think calling it just `GitClient` makes sense; the idea is to keep GoGit behind the scenes as an implementation detail

jiayiwang7: Yea make sense, do u mind adding doc explaining what each interface is responsible for?

pkg/addonmanager/addonclients/fluxaddonclient.go
chrisdoherty4: What code needs a `GitClient`? Does the `GitClient` represent a single repository (seems that way)? If so, what does a `FluxAddonClient` represent? Seems odd that a `FluxAddonClient` would have a reference to a specific repository but also be able to create other repositories in some capacity (SRP)?

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -52,11 +52,35 @@ const (
...
+	Commit(message string) error
+	Push(ctx context.Context) error
+	Pull(ctx context.Context, branch string) error
+	Init() error
+	Branch(name string) error
+}
+
+// GitProviderClient is responsible for provider-dependant actions on a hosted git provider API
+// Any actions you would take via the API of a git hosting provider should be included in the GitProviderClient
+type GitProviderClient interface {
```
chrisdoherty4, line 70: Based on the discussion above, could this have a better name? `GitRepositoryAdministrator` for example.

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -52,11 +52,35 @@ const (
 	deleteClusterconfigCommitMessage  = "Delete commit of cluster configuration; generated by EKS-A CLI"
 )
 
+// GitClient is responsible for provider-agnostic actions on a generic git repository.
```
chrisdoherty4, line 55: I think, with this comment, we could make it clearer that the backing construct represents a _single_ repository.

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: I can take a look at a more narrow name; this is going to be changing again today so will think about it

PR #1930, opened 2022-04-25
Add env var support for lifecycle api work with tilt

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
As of now we had these as local changes whenever we wanted to test the full lifecycle api work so checking some of the automation in. I added the ability to read and set the env var for Tilt, as well as set the env var when running create cluster. 

I debated using the bundle manifest for getting the controller image and copying the govc and helm binaries from that image using a multi-stage build, but for this use case it didn't feel necessary to have to write the logic to parse that file in the Tiltfile and use an image that is built from main instead of from local. I also looked into setting the env vars using kustomize substitutions like they do in cluster-api, but realized that that is more effort than needed for this use case due to the amount of tooling that is required to make that happen. 

I also added the ability to clear the cache for the features for unit testing. Because of how our unit tests are run, there is a chance for env vars to be leftover in the mutexMap, so clearing it for some of the test cases for now.

*Testing (if applicable):*
Ran unit tests and tested this functionality through tilt. Also testing create cluster with and without this env var.

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -5,6 +5,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -19,6 +20,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/diagnostics"
 	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers"
@@ -828,6 +830,13 @@ func (c *ClusterManager) removeOldWorkerNodeGroups(ctx context.Context, workload
 func (c *ClusterManager) InstallCustomComponents(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster, provider providers.Provider) error {
 	if err := c.clusterClient.installCustomComponents(ctx, clusterSpec, cluster); err != nil {
 		return err
+	}
+	fullLifecycleAPI := features.IsActive(features.FullLifecycleAPI())
+	if fullLifecycleAPI {
+		err := c.clusterClient.SetEksaControllerEnvVar(ctx, features.FullLifecycleAPIEnvVar, strconv.FormatBool(fullLifecycleAPI), cluster.KubeconfigFile)
+		if err != nil {
+			return err
+		}
 	}
 	return provider.InstallCustomProviderComponents(ctx, cluster.KubeconfigFile)
 }
--- a/pkg/clustermanager/cluster_manager_test.go
+++ b/pkg/clustermanager/cluster_manager_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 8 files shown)


pkg/clustermanager/cluster_manager.go
```diff
@@ -829,6 +831,13 @@ func (c *ClusterManager) InstallCustomComponents(ctx context.Context, clusterSpe
 	if err := c.clusterClient.installCustomComponents(ctx, clusterSpec, cluster); err != nil {
 		return err
 	}
+	fullLifecycleAPI := features.IsActive(features.FullLifecycleAPI())
+	if fullLifecycleAPI {
+		err := c.clusterClient.SetEksaControllerEnvVar(ctx, features.FullLifecycleAPIEnvVar, strconv.FormatBool(fullLifecycleAPI), cluster.KubeconfigFile)
```
g-gaston, line 836: I think automating this is awesome, thanks a lot!

However, I would vote to use flags for the controller instead of env vars
Main reason: env vars are mutable and flags are not. Changing the flags always gets reflected in the deployment spec. But env vars can be changed withing the pod.

I think env vars for the cli is not a bad idea, since it is not supposed to run for long. But the controllers is supposed to be a long running "process"

pkg/clustermanager/cluster_manager_test.go
```diff
@@ -905,12 +905,48 @@ func TestClusterManagerResumeEKSAControllerReconcileSuccessWithoutMachineConfig(
...
+
+func TestClusterManagerInstallCustomComponentsSuccessWithFullLifecycleAPI(t *testing.T) {
+	features.ClearCache()
+	oldFullLifecycleAPIFeatureValue := os.Getenv(features.FullLifecycleAPIEnvVar)
+	err := os.Setenv(features.FullLifecycleAPIEnvVar, "true")
+	if err != nil {
+		t.Errorf("error setting env var %v to true", features.FullLifecycleAPIEnvVar)
 	}
+	defer func() {
+		err := os.Setenv(features.FullLifecycleAPIEnvVar, oldFullLifecycleAPIFeatureValue)
```
g-gaston, line 945: something I learn from dan, you can now do `t.SetEnv` to set an env var only for a test duration

pkg/features/features.go
```diff
@@ -25,6 +25,11 @@ func IsActive(feature Feature) bool {
 	return feature.IsActive()
 }
 
+// ClearCache is mainly used for unit tests as of now
+func ClearCache() {
```
g-gaston, line 29: Can we add a test for this if possible?

g-gaston (commented):
love this

pkg/clustermanager/cluster_manager.go
vivek-koppuru: I agree, but the only issue with that is being able to set the flag when running the binary, where we don't support being able to edit the controller flags right now through kustomize or templating. We have some precedence in using flags so I would suggest going with this approach and creating an item to support using flags instead in the lifecycle story where we can modify it when deploying the controller. Do you have any other suggestions that we can implement immediately in this PR?

pkg/clustermanager/cluster_manager_test.go
vivek-koppuru: TIL

pkg/clustermanager/cluster_manager.go
g-gaston: We could do something similar to the envar and just update the deployment container.command on the fly
I would vote to give that a try. If it turns out to be more complicated that what it sounds, then I'm good with addressing separately in within this project

vivek-koppuru: I looked into this and it seems like it doesn't seem to be worth to use the patch commands or get the entire yaml and figure out where to inject the arg if we are anyways going to replace it with the approach of deploying the controller with the flag set through kustomize or templating. Going to go with the current approach.

PR #1934, opened 2022-04-25
Decouple `GitClient` and `GitProviderClient`; update GitFactory to support this

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/projects/9

*Description of changes:*

Last PR needed before we can start to incorporate the generic git provider flow into the `FluxAddonClient`. This decouples the responsibility for interacting with a hosted git provider API from the responsibility for interacting with a generic git repository on disk, and allows us to produce these clients based on the contents of the FluxConfig.

- divide the responsibility of the `GitProvider` interface into a `GitClient` and a `GitProviderClient` interfaces. [In the last pr](https://github.com/aws/eks-anywhere/pull/1928), we split these up on the flux side; now, we're formally producing two different objects to use.
- set up the `gitFactory` to return the three component parts needed by the fluxaddonclient: a `GitClient` a `GitProviderClient`, and a writer configured to write to the local git repository.
- set up the `gitFactory` to build a provider based on the input flux config. only produces github for now, will add the generic provider case next.

next steps:
- after https://github.com/aws/eks-anywhere/pull/1900 is merged, will update the FluxAddonClient to invoke `FluxBootstrapGit` if the `Git` provider is present
- add a case which builds a generic git provider to the `gitFactory`.

*Testing (if applicable):*

create cluster
upgrade cluster
delete cluster


pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -125,8 +115,8 @@ func NewFluxAddonClient(flux Flux, gitOpts *GitOptions) *FluxAddonClient {
 	}
 	return &FluxAddonClient{
 		flux:              flux,
-		gitProviderClient: gitOpts.Git,
-		gitClient:         gitOpts.Git,
+		gitProviderClient: gitOpts.GitProvider,
```
jiayiwang7, line 118: should we rename this to `gitProvider` too?

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,76 @@ package gitfactory
...
 
-type Options struct {
-	GithubGitClient github.GitClient
+func New() *gitProviderFactory {
+	return &gitProviderFactory{}
 }
 
-func New(opts Options) *gitProviderFactory {
-	return &gitProviderFactory{GitClient: opts.GithubGitClient}
+func (g *gitProviderFactory) Build(ctx context.Context, cluster *v1alpha1.Cluster, fluxConfig *v1alpha1.FluxConfig, writer filewriter.FileWriter) (git.ProviderClient, git.Client, filewriter.FileWriter, error) {
```
jiayiwang7, line 22: it's bit strange to return both gitProvider and gitClient from the `gitProviderFactory` method. Also what's the reason to keep this `gitProviderFactory` struct for if non of its field is used in any method?

pkg/addonmanager/addonclients/fluxaddonclient.go
jiayiwang7: what do u think abt having just one `git` as `FluxAddonClient` field?

```go
type git struct {
  provider
  client
  writer
}
```

pkg/git/factory/gitfactory.go
jiayiwang7: if using the `git` struct i mentioned above, this method can just return it? to avoid many return values

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,76 @@ package gitfactory
...
+		}
+	default:
+		return nil, nil, nil, fmt.Errorf("no valid git provider in FluxConfigSpec. Spec: %v", fluxConfig)
+	}
+	return providerClient, gitClient, repoWriter, nil
 }
 
-// BuildProvider will configure and return the proper Github based on the given GitOps configuration.
-func (g *gitProviderFactory) BuildProvider(ctx context.Context, fluxConfig *v1alpha1.FluxConfigSpec) (git.Provider, error) {
+func (g *gitProviderFactory) buildGithubProvider(ctx context.Context, clusterName string, owner string, repo string, personal bool, writer filewriter.FileWriter) (git.ProviderClient, git.Client, filewriter.FileWriter, error) {
```
jiayiwang7, line 40: what abt passing `github *GithubProviderConfig` as arg instead of `owner, repo, personal`?

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,76 @@ package gitfactory
...
-	provider, err := github.New(g.GitClient, githubProviderClient, githubProviderOpts, auth)
+	provider, err := github.New(githubProviderClient, githubProviderOpts, auth)
 	if err != nil {
-		return nil, err
+		return nil, nil, nil, err
 	}
 
-	return provider, nil
+	localGitRepoPath := filepath.Join(clusterName, "git", repo)
+	client := gitclient.New(auth, localGitRepoPath, github.RepoUrl(owner, repo))
```
jiayiwang7, line 59: shouldn't the gitclient agnostic among different providers? i would move it out of this `buildGithubProvider` method.

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,76 @@ package gitfactory
...
+
+	w, err := newRepositoryWriter(writer, repo)
```
jiayiwang7, line 61: should the writer also be provider agnostic?

pkg/git/gitclient/gitclient.go
```diff
@@ -28,38 +28,37 @@ const (
...
-	GoGitClient         GoGit
-}
-
-func New(opts Options) *GitClient {
-	return &GitClient{
-		Opts:    opts,
-		Client:  newClient(opts),
-		Retrier: retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
+func New(auth git.TokenAuth, repoDirectory string, repoUrl string) *GitClient {
+	c := &GitClient{
```
jiayiwang7, line 39: nit: 
```
return &GitClient{...}
```

pkg/git/gitclient/gitclient.go
```diff
@@ -308,11 +299,9 @@ type GoGit interface {
 	SetRepositoryReference(r *gogit.Repository, p *plumbing.Reference) error
 }
 
-type goGitClient struct {
-	goGit GoGit
-}
+type goGit struct{}
 
-func (ggc *goGitClient) Clone(ctx context.Context, dir string, repourl string, auth transport.AuthMethod) (*gogit.Repository, error) {
+func (ggc *goGit) Clone(ctx context.Context, dir string, repourl string, auth transport.AuthMethod) (*gogit.Repository, error) {
```
jiayiwang7, line 304: nit: `func (gg *goGit)`

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: yeah I like that I think it would make it much more clear

pkg/git/factory/gitfactory.go
danbudris: it's just the URL structure that's defendant on the provider in this case, with a github-specific URL scheme

pkg/git/factory/gitfactory.go
danbudris: yes, it is provider agnostic. the idea here was that we configure the writer to target the correct directory where we're initializing the git client for the user; will see if there's a better plac eto do this

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: this is now `GitTools`; it's returned from the factory, input to the `FluxAddonClient`, and used throughout.

pkg/git/factory/gitfactory.go
danbudris: now using `GitTools` for this

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,81 @@ package gitfactory
...
+		if err != nil {
+			return tools, fmt.Errorf("building github provider: %v", err)
+		}
+	default:
+		return tools, fmt.Errorf("no valid git provider in FluxConfigSpec. Spec: %v", fluxConfig)
+	}
+
+	localGitRepoPath := filepath.Join(cluster.Name, "git", repo)
+	clientOptions = append(clientOptions, gitclient.WithRepositoryDirectory(localGitRepoPath))
+	tools.Client = gitclient.New(clientOptions...)
```
jiayiwang7, line 47: should we maybe have a separate method that returns `client` that completely isolates git client config from provider?
```
func gitClient(...) Client {
  clientOptions := []gitclient.Opt{
    gitclient.WithRepositoryUrl(),
    gitclient.WithTokenAuth(),
    gitclient.WithRepositoryDirectory()),
  }
  
  return gitclient.New(clientOptions...)
}
```

pkg/git/factory/gitfactory.go
```diff
@@ -2,43 +2,81 @@ package gitfactory
...
-// BuildProvider will configure and return the proper Github based on the given GitOps configuration.
-func (g *gitProviderFactory) BuildProvider(ctx context.Context, fluxConfig *v1alpha1.FluxConfigSpec) (git.Provider, error) {
-	token, err := github.GetGithubAccessTokenFromEnv()
+	repoWriter, err := newRepositoryWriter(writer, repo)
 	if err != nil {
-		return nil, err
+		return tools, err
 	}
-	auth := git.TokenAuth{Token: token, Username: fluxConfig.Github.Owner}
+	tools.Writer = repoWriter
```
jiayiwang7, line 53: and in the end 
```go
tools := &addonclients.GitTools{
  provider: provider,
  client: client,
  writer: writer,
}

return tools, nil
```

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -76,59 +73,22 @@ type GitProviderClient interface {
...
-	retrier           *retrier.Retrier
+	flux     Flux
+	gitTools *GitTools
+	retrier  *retrier.Retrier
 }
 
-type GitOptions struct {
-	Git    git.Provider
-	Writer filewriter.FileWriter
+type GitTools struct {
```
jiayiwang7, line 81: does GitTools struct belong to addonclient? imo addonclient use GitTools as a dependency. 
git pkg should not import `addonclient.GitTools`. Might need to move this to git pkg?

pkg/git/factory/gitfactory.go
danbudris: we can move it to a separate method, but we can't completely isolate it -- there are provider-specific configurations that the client will receive, such as the repo URL and the kind of auth

danbudris: we can do something like 1e37fe5fc308bbcdaa8fa675a830cb9bed482244

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: yep make sense, it had originally been an import cylce but since moving the factory call to the dependency factory we can now do this

danbudris: cf051e7e207aa48f08b3b934a6d07e3f27438b4e

pkg/git/factory/gitfactory.go
danbudris: the issue is that the auth is going to be provider dependent; the repo URL is provider dependent; and we can either have the git client options configured here, as we processes the provider, or we can have it itself review the FluxConfig again and construct the values; either way, we need to be able to have some level of coordination between what the provider is and what the gitClient is

danbudris: ok further isolated this --9efcc6c0

danbudris: that method will change a bit once we introduce ssh auth in the next pr

jiayiwang7 (approved):

pkg/git/git.go
```diff
@@ -5,10 +5,7 @@ import (
 	"fmt"
 )
 
-// Provider acts as an interface for specific Git hosting providers -- e.g. GitHub, BitBucket, GitLab, etc.
-// It wraps a low-level git implementation (e.g. gogit) and abstracts auth and provider specific configurations
-// while providing a common interface for local git actions.
-type Provider interface {
+type Client interface {
```
chrisdoherty4, line 8: How many implementations of 'Client' are there?

pkg/git/factory/gitfactory.go
danbudris: donezo 26afe51e482afff7fd14a95c0a3411544dfcb319

pkg/git/git.go
danbudris: right now, there's a single implementation of Client -- using `go-git`. The only consumer at this point is the Flux system (misnamed `FluxAddonClient`) and the accompanying e2e tests. The thought behind having Git behind a  specific interface was that it would permit us to switch out the implementation underneath is more easily as needed -- while we're using the golang library [go-git](https://github.com/go-git/go-git) there was some discussion initially about wether or not to use use a git executable. If we ever wanted to swap this out we could do it pretty easily

the Provider interface (Client and Provider were one interface until just a few days ago) will potentially have more implementations as we support more git providers (gitlab, etc) as first-class citizens; right now, it's just Github.

jiayiwang7 (approved):

pkg/git/git.go
chrisdoherty4: >  it would permit us to switch out the implementation underneath it more easily 

This is a misconception. Duck typing renders this interface redundant so it can be removed entirely given your summary.

PR #1935, opened 2022-04-25
Remove provider factory

chrisdoherty4:
The provider factory created a level of indirection that forced dependencies of all providers to be constructed irrespective of what provider was being used just to create a valid provider factory construct. In the event we know, elsewhere in the code, what the provider factory is required the design promotes dispersed construction logic for providers.

This change removes the provider factory in favor of constructing it where we construct everything else: `/pkg/dependencies`. While `/pkg/dependencies` has its own problems, this alleviates the concerns above and paves a path for injecting dependencies into providers more easily.

I purposely chose not to break the switch statement up to avoid premature optimization as further refactoring is taking place in the Tinkerbell provider around dependency injection and it could help influence further modularization here.


pkg/dependencies/factory.go
```diff
@@ -219,29 +202,112 @@ func (f *Factory) WithProviderFactory(clusterConfigFile string, clusterConfig *v
...
-			CloudStackKubectlClient:   f.dependencies.Kubectl,
-			VSphereGovcClient:         f.dependencies.Govc,
-			VSphereKubectlClient:      f.dependencies.Kubectl,
-			SnowKubectlClient:         f.dependencies.Kubectl,
-			SnowConfigManager:         f.dependencies.SnowConfigManager,
-			TinkerbellKubectlClient:   f.dependencies.Kubectl,
-			TinkerbellClients:         tinkerbell.TinkerbellClients{ProviderTinkClient: f.dependencies.Tink, ProviderPbnjClient: f.dependencies.Pbnj},
-			Writer:                    f.dependencies.Writer,
-			ClusterResourceSetManager: f.dependencies.ResourceSetManager,
+		switch clusterConfig.Spec.DatacenterRef.Kind {
```
g-gaston, line 215: nit: we could avoid a second switch by encapsulating the building logic in a closure (for example) in the first switch

Issue #1109, opened 2022-01-31
CAPAS - EKS-A cluster controller v2 support


jiayiwang7:
Issue to track Snow support in the new EKS-A cluster controller (v2)

## User stories

As a user, when talking about the EKS-A API (accessible though kubectl or GitOps):

1. I would like to create a Snow cluster
1. I would like to scale up/down my worker nodes
1. I would like to scale up/down my control plane
1. I would like to upgrade the kubernetes version of my cluster
1. I would like to upgrade my cluster's dependencies to a new version bundled by AWS
1. I would like to delete my cluster
1. I would like to configure OIDC for my cluster
1. I would like to configure `iam-authenticator` for my cluster
1. I would like to configure an http proxy for my cluster
1. I would like to configure a registry mirror for my cluster
1. I would like to configure Cilium
1. I would like to configure my cluster network
1. I would like to configure taints and labels

Issue #1106, opened 2022-01-31
CAPAS - CLI upgrade workflow


jiayiwang7:
Issue to track Snow support for CLI upgrade workflow

- [x] components upgrade
- [x] cluster upgrade 
- [x] Set mutable/immutable fields in the spec for upgrade
- [x] #2497

PR #1941, opened 2022-04-26
Auth agnostic kube client

g-gaston:
Part of #1106 and #1109

## Description of changes
Wrapping around the the kubectl interface to make the client API
agnostic from the authentication mechanism. Also adding support for an
internal schema that avoids having to indicate the resource type because
it's inferred from the concrete CRD type.

This allows to provide a common interface that can be implemented by
both a kubectl binary wrapper and a controller runtime client. With
this, we will be be able to build logic that interacts with the api
server and that can be shared between the CLI and the controller.

## Testing
Added unit tests


pkg/executables/kubectl.go
```diff
@@ -1364,12 +1366,24 @@ func (k *Kubectl) GetResource(ctx context.Context, resourceType string, name str
 	return found, err
 }
 
-func (k *Kubectl) getObject(ctx context.Context, resourceType, name, namespace, kubeconfig string, obj client.Object) error {
-	stdOut, err := k.Execute(ctx, "get", "--namespace", namespace, resourceType, name, "-o", "json", "--kubeconfig", kubeconfig)
+// GetObject performas a GET call to the kube API server authenticating with a kubeconfig file
```
jiayiwang7, line 1369: you made me look up `performas` and see if it's a proper word. showing in multiple places

PR #1947, opened 2022-04-26
Feature Generate packages (#1622)

a-cool-train:
*Description of changes:*
Include Generation of curated packages

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -1,0 +1,100 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/validations"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
+type generatePackageOptions struct {
+	directory   string
+	source      curatedpackages.BundleSource
+	kubeVersion string
+	registry    string
+}
+
+var gepo = &generatePackageOptions{}
+
+func init() {
+	generateCmd.AddCommand(generatePackageCommand)
+	generatePackageCommand.Flags().StringVarP(&gepo.directory, "directory", "d", "", "Directory to save generated packages")
+	generatePackageCommand.Flags().Var(&gepo.source, "source", "Location to find curated packages: (cluster, registry)")
+	if err := generatePackageCommand.MarkFlagRequired("source"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	generatePackageCommand.Flags().StringVar(&gepo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	if err := generatePackageCommand.MarkFlagRequired("directory"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	generatePackageCommand.Flags().StringVar(&gepo.registry, "registry", "", "Used to specify an alternative registry for package generation")
+}
+
+var generatePackageCommand = &cobra.Command{
+	Use:          "packages [flags]",
+	Aliases:      []string{"package", "packages"},
+	Short:        "Generate package(s)",
+	Long:         "This command is used to generate curated packages",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE:         runGeneratePackages(),
+}
+
+func runGeneratePackages() func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := curatedpackages.ValidateKubeVersion(gepo.kubeVersion, gepo.source); err != nil {
+			return err
+		}
+		if !validations.FileExists(gepo.directory) {
+			return fmt.Errorf("directory %s does not exist", gepo.directory)
+		}
+		return generatePackages(cmd.Context(), args)
+	}
+}
+
+func generatePackages(ctx context.Context, args []string) error {
+	kubeConfig := kubeconfig.FromEnvironment()
+	deps, err := newDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+	bm := curatedpackages.CreateBundleManager(gepo.kubeVersion)
+	registry, err := curatedpackages.NewRegistry(ctx, deps, gepo.registry, gepo.kubeVersion)
+	if err != nil {
+		return err
+	}
+
+	b := curatedpackages.NewBundleReader(
+		kubeConfig,
+		gepo.kubeVersion,
+		gepo.source,
+		deps.Kubectl,
+		bm,
+		version.Get(),
+		registry,
+	)
+
+	bundle, err := b.GetLatestBundle(ctx)
+	if err != nil {
+		return err
+	}
+
+	packageClient := curatedpackages.NewPackageClient(
+		bundle,
+		args...,
+	)
+	packages, err := packageClient.GeneratePackages()
+	if err != nil {
+		return err
+	}
+	if err = packageClient.WritePackagesToFile(packages, gepo.directory); err != nil {
+		return err
+	}
+	return nil
+}
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -9,12 +9,10 @@ import (
 
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
-	"github.com/aws/eks-anywhere/pkg/validations"
 	"github.com/aws/eks-anywhere/pkg/version"
 )
 
 type generatePackageOptions struct {
-	directory   string
 	source      curatedpackages.BundleSource
 	kubeVersion string
 	registry    string
@@ -24,7 +22,6 @@ var gpOptions = &generatePackageOptions{}
 
 func init() {
 	generateCmd.AddCommand(generatePackageCommand)
-	generatePackageCommand.Flags().StringVarP(&gpOptions.directory, "directory", "d", ".", "Directory to save generated packages")
 	generatePackageCommand.Flags().Var(&gpOptions.source, "source", "Location to find curated packages: (cluster, registry)")
 	if err := generatePackageCommand.MarkFlagRequired("source"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
@@ -36,7 +33,7 @@ func init() {
 var generatePackageCommand = &cobra.Command{
 	Use:          "packages [flags]",
 	Aliases:      []string{"package", "packages"},
-	Short:        "Generate package(s)",
+	Short:        "Generate package(s) configuration",
 	Long:         "Generates Kubernetes configuration files for curated packages",
 	PreRunE:      preRunPackages,
 	SilenceUsage: true,
@@ -46,9 +43,6 @@ var generatePackageCommand = &cobra.Command{
 func runGeneratePackages(cmd *cobra.Command, args []string) error {
 	if err := curatedpackages.ValidateKubeVersion(gpOptions.kubeVersion, gpOptions.source); err != nil {
 		return err
-	}
-	if !validations.FileExists(gpOptions.directory) {
-		return fmt.Errorf("directory %s does not exist", gpOptions.directory)
 	}
 	return generatePackages(cmd.Context(), args)
 }
@@ -88,7 +82,7 @@ func generatePackages(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
-	if err = packageClient.WritePackagesToFile(packages, gpOptions.directory); err != nil {
+	if err = packageClient.WritePackagesToStdOut(packages); err != nil {
 		return err
 	}
 	return nil
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -3,11 +3,11 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"github.com/aws/eks-anywhere/pkg/config"
 	"log"
 
 	"github.com/spf13/cobra"
 
+	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 	"github.com/aws/eks-anywhere/pkg/version"
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"github.com/aws/eks-anywhere/pkg/config"
 	"log"
 
 	"github.com/spf13/cobra"
@@ -54,7 +55,11 @@ func generatePackages(ctx context.Context, args []string) error {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
 	bm := curatedpackages.CreateBundleManager(gpOptions.kubeVersion)
-	registry, err := curatedpackages.NewRegistry(ctx, deps, gpOptions.registry, gpOptions.kubeVersion)
+	username, password, err := config.ReadCredentials()
+	if err != nil && gpOptions.registry != "" {
+		return err
+	}
+	registry, err := curatedpackages.NewRegistry(deps, gpOptions.registry, gpOptions.kubeVersion, username, password)
 	if err != nil {
 		return err
 	}
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -20,20 +20,20 @@ type generatePackageOptions struct {
 	registry    string
 }
 
-var gepo = &generatePackageOptions{}
+var gpOptions = &generatePackageOptions{}
 
 func init() {
 	generateCmd.AddCommand(generatePackageCommand)
-	generatePackageCommand.Flags().StringVarP(&gepo.directory, "directory", "d", "", "Directory to save generated packages")
-	generatePackageCommand.Flags().Var(&gepo.source, "source", "Location to find curated packages: (cluster, registry)")
+	generatePackageCommand.Flags().StringVarP(&gpOptions.directory, "directory", "d", "", "Directory to save generated packages")
+	generatePackageCommand.Flags().Var(&gpOptions.source, "source", "Location to find curated packages: (cluster, registry)")
 	if err := generatePackageCommand.MarkFlagRequired("source"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
 	}
-	generatePackageCommand.Flags().StringVar(&gepo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	generatePackageCommand.Flags().StringVar(&gpOptions.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
 	if err := generatePackageCommand.MarkFlagRequired("directory"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
 	}
-	generatePackageCommand.Flags().StringVar(&gepo.registry, "registry", "", "Used to specify an alternative registry for package generation")
+	generatePackageCommand.Flags().StringVar(&gpOptions.registry, "registry", "", "Used to specify an alternative registry for package generation")
 }
 
 var generatePackageCommand = &cobra.Command{
@@ -48,11 +48,11 @@ var generatePackageCommand = &cobra.Command{
 
 func runGeneratePackages() func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		if err := curatedpackages.ValidateKubeVersion(gepo.kubeVersion, gepo.source); err != nil {
+		if err := curatedpackages.ValidateKubeVersion(gpOptions.kubeVersion, gpOptions.source); err != nil {
 			return err
 		}
-		if !validations.FileExists(gepo.directory) {
-			return fmt.Errorf("directory %s does not exist", gepo.directory)
+		if !validations.FileExists(gpOptions.directory) {
+			return fmt.Errorf("directory %s does not exist", gpOptions.directory)
 		}
 		return generatePackages(cmd.Context(), args)
 	}
@@ -64,16 +64,16 @@ func generatePackages(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
-	bm := curatedpackages.CreateBundleManager(gepo.kubeVersion)
-	registry, err := curatedpackages.NewRegistry(ctx, deps, gepo.registry, gepo.kubeVersion)
+	bm := curatedpackages.CreateBundleManager(gpOptions.kubeVersion)
+	registry, err := curatedpackages.NewRegistry(ctx, deps, gpOptions.registry, gpOptions.kubeVersion)
 	if err != nil {
 		return err
 	}
 
 	b := curatedpackages.NewBundleReader(
 		kubeConfig,
-		gepo.kubeVersion,
-		gepo.source,
+		gpOptions.kubeVersion,
+		gpOptions.source,
 		deps.Kubectl,
 		bm,
 		version.Get(),
@@ -93,7 +93,7 @@ func generatePackages(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
-	if err = packageClient.WritePackagesToFile(packages, gepo.directory); err != nil {
+	if err = packageClient.WritePackagesToFile(packages, gpOptions.directory); err != nil {
 		return err
 	}
 	return nil
--- a/cmd/eksctl-anywhere/cmd/generatepackage.go
+++ b/cmd/eksctl-anywhere/cmd/generatepackage.go
@@ -24,15 +24,12 @@ var gpOptions = &generatePackageOptions{}
 
 func init() {
 	generateCmd.AddCommand(generatePackageCommand)
-	generatePackageCommand.Flags().StringVarP(&gpOptions.directory, "directory", "d", "", "Directory to save generated packages")
+	generatePackageCommand.Flags().StringVarP(&gpOptions.directory, "directory", "d", ".", "Directory to save generated packages")
 	generatePackageCommand.Flags().Var(&gpOptions.source, "source", "Location to find curated packages: (cluster, registry)")
 	if err := generatePackageCommand.MarkFlagRequired("source"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
 	}
 	generatePackageCommand.Flags().StringVar(&gpOptions.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
-	if err := generatePackageCommand.MarkFlagRequired("directory"); err != nil {
-		log.Fatalf("Error marking flag as required: %v", err)
-	}
 	generatePackageCommand.Flags().StringVar(&gpOptions.registry, "registry", "", "Used to specify an alternative registry for package generation")
 }
 
@@ -40,22 +37,20 @@ var generatePackageCommand = &cobra.Command{
 	Use:          "packages [flags]",
 	Aliases:      []string{"package", "packages"},
 	Short:        "Generate package(s)",
-	Long:         "This command is used to generate curated packages",
+	Long:         "Generates Kubernetes configuration files for curated packages",
 	PreRunE:      preRunPackages,
 	SilenceUsage: true,
-	RunE:         runGeneratePackages(),
+	RunE:         runGeneratePackages,
 }
 
-func runGeneratePackages() func(cmd *cobra.Command, args []string) error {
-	return func(cmd *cobra.Command, args []string) error {
-		if err := curatedpackages.ValidateKubeVersion(gpOptions.kubeVersion, gpOptions.source); err != nil {
-			return err
-		}
-		if !validations.FileExists(gpOptions.directory) {
-			return fmt.Errorf("directory %s does not exist", gpOptions.directory)
-		}
-		return generatePackages(cmd.Context(), args)
+func runGeneratePackages(cmd *cobra.Command, args []string) error {
+	if err := curatedpackages.ValidateKubeVersion(gpOptions.kubeVersion, gpOptions.source); err != nil {
+		return err
 	}
+	if !validations.FileExists(gpOptions.directory) {
+		return fmt.Errorf("directory %s does not exist", gpOptions.directory)
+	}
+	return generatePackages(cmd.Context(), args)
 }
 
 func generatePackages(ctx context.Context, args []string) error {
--- a/pkg/curatedpackages/bundle.go
+++ b/pkg/curatedpackages/bundle.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	RepositoryName = "eksa-package-bundles"
+	RepositoryName = "eks-anywhere-packages-bundles"
 )
 
 type Reader interface {
--- a/pkg/curatedpackages/bundle.go
+++ b/pkg/curatedpackages/bundle.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	RepositoryName = "eks-anywhere-packages-bundles"
+	ImageRepositoryName = "eks-anywhere-packages-bundles"
 )
 
 type Reader interface {
--- a/pkg/curatedpackages/bundlesource_test.go
+++ b/pkg/curatedpackages/bundlesource_test.go
@@ -1,0 +1,91 @@
+package curatedpackages_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+)
+
+func TestBundleSourceSet(t *testing.T) {
+	h := newBSHelper(t)
+	h.Parallel()
+
+	type acceptCase struct {
+		// name of the test
+		name string
+		// input from the user
+		input string
+		// expected result
+		expected curatedpackages.BundleSource
+	}
+
+	accepts := []acceptCase{
+		{"golden path", "cluster", curatedpackages.Cluster},
+		{"case insensitivity", "cLuSTEr", curatedpackages.Cluster},
+		{"golden path", "registry", curatedpackages.Registry},
+		{"case insensitivity #2", "Registry", curatedpackages.Registry},
+		{"whitespace before", " registry", curatedpackages.Registry},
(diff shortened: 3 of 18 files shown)


pkg/curatedpackages/common.go
```diff
@@ -0,0 +1,54 @@
...
+	)
+	return defaultRegistry, nil
+}
+
+func CreateBundleManager(kubeVersion string) bundle.Manager {
+	versionSplit := strings.Split(kubeVersion, ".")
+	if len(versionSplit) != 2 {
+		return nil
+	}
+	major, minor := versionSplit[0], versionSplit[1]
```
vivek-koppuru, line 49: This looks like something that can go under a separate function that is more general. If there is nothing else that uses this logic, we can separate it out for now in its own function.

pkg/curatedpackages/common.go
```diff
@@ -0,0 +1,54 @@
...
+	log := logr.Discard()
+	discovery := NewDiscovery(major, minor)
```
vivek-koppuru, line 51: What's the point of having a Discovery struct that only has the major and minor version? It sounds like we are just creating a Version struct then right?

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+
+func (p *PackageClient) WritePackagesToFile(packages []packagesv1.Package, d string) error {
+	directory := filepath.Join(d, packageLocation)
+	if !validations.FileExists(directory) {
+		if err := os.Mkdir(directory, dirPermission); err != nil {
+			return fmt.Errorf("unable to create directory %s", directory)
+		}
+	}
+
+	for _, pack := range packages {
```
vivek-koppuru, line 86: Use `p` instead of `pack` imo, or just `package`

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
 	}
 	return versions
 }
+
+func (p *PackageClient) GeneratePackages() ([]packagesv1.Package, error) {
+	packageNameToPackage := p.getPackageNameToPackage()
+	var packages []packagesv1.Package
+	for _, v := range p.packages {
```
vivek-koppuru, line 66: Why is this a `v`?

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
 	}
 	return versions
 }
+
+func (p *PackageClient) GeneratePackages() ([]packagesv1.Package, error) {
+	packageNameToPackage := p.getPackageNameToPackage()
```
vivek-koppuru, line 64: Is this just a `packageMap`? I feel like this naming makes it confusing, even though you are literally describing the map's key/value.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+	var packages []packagesv1.Package
+	for _, v := range p.packages {
+		bundlePackage := packageNameToPackage[strings.ToLower(v)]
+		if bundlePackage.Name == "" {
+			fmt.Println(fmt.Errorf("unknown package %s", v).Error())
+			continue
+		}
+		name := customName + strings.ToLower(bundlePackage.Name)
```
vivek-koppuru, line 72: What is the reason for prepending `my-` here?

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+func writeToFile(dir string, packageName string, content []byte) error {
+	file := filepath.Join(dir, packageName) + ".yaml"
+	if err := os.WriteFile(file, content, filePermission); err != nil {
+		return fmt.Errorf("unable to write to the file: %s %v", file, err)
+	}
+	return nil
+}
+
+func (p *PackageClient) getPackageNameToPackage() map[string]packagesv1.BundlePackage {
+	pntop := make(map[string]packagesv1.BundlePackage)
```
vivek-koppuru, line 110: I would honestly just call this `pMap`

pkg/curatedpackages/packageclient_test.go
```diff
@@ -0,0 +1,59 @@
...
+					{
+						Name: "harbor-test",
+					},
+					{
+						Name: "redis-test",
+					},
+				},
+			},
+		},
+		customName: "my-",
```
vivek-koppuru, line 37: I would reuse the variable defined in the other package, even if you need to export it.

pkg/curatedpackages/packagetypes.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	api "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+// DisplayPackage wraps Package to omit undesired members (like Status).
+//
+// This is necessary in part because of https://github.com/golang/go/issues/11939
+// but also because we just don't want to generate a Status section when we're
+// emitting templates for a user to modify.
+type DisplayPackage struct {
```
vivek-koppuru, line 12: I kind of like `DisplayablePackage` as a name here instead to describe the type of package

pkg/curatedpackages/packagetypes.go
```diff
@@ -0,0 +1,19 @@
...
+	*api.Package
+	Status *interface{} `json:"status,omitempty"`
```
vivek-koppuru, line 14: I don't understand adding this if we are trying to omit the `Status` section. Maybe I misunderstand.

pkg/curatedpackages/common.go
a-cool-train: The Discovery struct implements an interface that is required by a bundleManager.

pkg/curatedpackages/bundle.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	RepositoryName = "eksa-package-bundles"
+	RepositoryName = "eks-anywhere-packages-bundles"
```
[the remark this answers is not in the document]
a-cool-train: Yes this is a repository name that has curated packages release bundle manifest.

pkg/curatedpackages/common.go
```diff
@@ -0,0 +1,54 @@
+package curatedpackages
```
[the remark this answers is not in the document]
a-cool-train: No reason. I renamed the file.

pkg/curatedpackages/packageclient.go
a-cool-train: No reason. Couldn't think of a better letter. Let me change it.

pkg/curatedpackages/packageclient.go
a-cool-train: That's the structure we are going with as of now. Whenever a user generates a package, the name of the package is called my-package. The idea is that they will change this if they need to.

pkg/curatedpackages/common.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/aws/eks-anywhere-packages/pkg/artifacts"
+	"github.com/aws/eks-anywhere-packages/pkg/bundle"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
+func NewRegistry(ctx context.Context, deps *dependencies.Dependencies, registry, kubeVersion string) (BundleRegistry, error) {
+	if registry != "" {
+		registryUsername := os.Getenv("REGISTRY_USERNAME")
```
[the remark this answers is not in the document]
a-cool-train: Ok merged both functions.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+
+func (pc *PackageClient) GeneratePackages() ([]packagesv1.Package, error) {
+	packageMap := pc.getPackageNameToPackage()
+	var packages []packagesv1.Package
+	for _, p := range pc.packages {
+		bundlePackage := packageMap[strings.ToLower(p)]
+		if bundlePackage.Name == "" {
+			fmt.Println(fmt.Errorf("unknown package %s", p).Error())
+			continue
+		}
```
ewollesen, line 71:
```suggestion
		bundlePackage, found := packageMap[strings.ToLower(p)]
		if !found {
			fmt.Fprintf(os.Stderr, "unknown package %q\n", p)
			continue
		}
```

This is a safer, and more idiomatic way to check for the presence of a value within a map. It's safer because if the element type of the map were to be a pointer, then the value would be nil, and you'd be dereferencing a nil pointer, which is of course a run-time error.

Also, `fmt.Println` doesn't take "%s" holders.

There's no need to generate an error, just to call its `Error()` function.

Lastly, error output should generally go to [`os.Stderr`](https://pkg.go.dev/os#pkg-variables).

PS- We should consider if we want to continue in the face of an unknown package. If said package were a dependency for something else, and the user didn't notice the error output, we could possibly do something destructive.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		name := CustomName + strings.ToLower(bundlePackage.Name)
+		packages = append(packages, convertBundlePackageToPackage(bundlePackage, name, pc.bundle.APIVersion))
+	}
+	return packages, nil
+}
+
+func (pc *PackageClient) WritePackagesToFile(packages []packagesv1.Package, d string) error {
+	directory := filepath.Join(d, packageLocation)
+	if !validations.FileExists(directory) {
```
ewollesen, line 80: The command already aborts if the directory doesn't exist (https://github.com/aws/eks-anywhere/pull/1947/files#diff-7f848af9ae27534974e24287c26b1a19c58373183398dc3672d1ddaa2fa1d58fR55). So this check is superfluous.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		if err := os.Mkdir(directory, dirPermission); err != nil {
```
ewollesen, line 81:
```suggestion
		if err := os.MkdirAll(directory, dirPermission); err != nil {
```
[`os.MkdirAll`](https://pkg.go.dev/os#MkdirAll)

pkg/curatedpackages/packageclient.go
```diff
@@ -3,37 +3,50 @@ package curatedpackages
...
-	flags    = 0
+	minWidth        = 16
+	tabWidth        = 8
+	padding         = 0
+	padChar         = '\t'
+	flags           = 0
+	CustomName      = "my-"
+	kind            = "Package"
+	filePermission  = 0o644
+	dirPermission   = 0o755
```
ewollesen, line 27:
```suggestion
	filePermission  = 0o600
	dirPermission   = 0o700
```
Better safe than sorry.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		packages = append(packages, convertBundlePackageToPackage(bundlePackage, name, pc.bundle.APIVersion))
+	}
+	return packages, nil
+}
+
+func (pc *PackageClient) WritePackagesToFile(packages []packagesv1.Package, d string) error {
+	directory := filepath.Join(d, packageLocation)
+	if !validations.FileExists(directory) {
+		if err := os.Mkdir(directory, dirPermission); err != nil {
+			return fmt.Errorf("unable to create directory %s", directory)
```
ewollesen, line 82:
```suggestion
			return fmt.Errorf("unable to create directory %q: %w", directory, err)
```
It's probably ok to let the user see the error. It could be a simple permissions issue that they could easily fix.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		}
+	}
+
+	for _, pack := range packages {
+		displayPackage := NewDisplayPackage(pack)
+		content, err := yaml.Marshal(displayPackage)
+		if err != nil {
+			fmt.Println(fmt.Errorf("unable to parse package %s %v", pack.Name, err).Error())
+			continue
```
ewollesen, line 91: Here again I think we should consider if we want to continue here. The lack of a package could cause more problems.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		}
+		err = writeToFile(directory, pack.Name, content)
+		if err != nil {
+			fmt.Println(err.Error())
```
ewollesen, line 95:
```suggestion
			fmt.Fprintln(os.Stderr, err.Error())
			continue // or return an error to abort.
```
This is just good hygiene in case someone adds code below later.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		if err := os.Mkdir(directory, dirPermission); err != nil {
+			return fmt.Errorf("unable to create directory %s", directory)
+		}
+	}
+
+	for _, pack := range packages {
+		displayPackage := NewDisplayPackage(pack)
+		content, err := yaml.Marshal(displayPackage)
+		if err != nil {
+			fmt.Println(fmt.Errorf("unable to parse package %s %v", pack.Name, err).Error())
```
ewollesen:
```suggestion
			fmt.Fprintf(os.Stderr, "unable to parse package %q: %s\n", pack.Name, err)
```
You get the idea on the fmt.Println now I hope and error stuff, I'll leave it to you to find/fix the rest.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+			continue
+		}
+		err = writeToFile(directory, pack.Name, content)
```
ewollesen: Since we're writing YAML files, maybe it makes more sense to just write a single YAML file, separated using the YAML document syntax? That way we don't need a directory, don't have to necessarily create directories, and can even output to stdout (in a nice UNIX-friendly way).

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		pMap[strings.ToLower(p.Name)] = p
+	}
+	return pMap
+}
+
+func convertBundlePackageToPackage(bp packagesv1.BundlePackage, name string, apiVersion string) packagesv1.Package {
+	p := packagesv1.Package{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: constants.EksaPackagesName,
```
ewollesen, line 121: I thought the user was allowed to select their own namespace?

pkg/curatedpackages/packagetypes.go
ewollesen: > I don't understand adding this if we are trying to omit the `Status` section. Maybe I misunderstand.
It's a pointer with omitempty, so it won't be output by the YAML library. Simpler might have been:
```suggestion
	Status interface{} `json:"-"`
```

ewollesen (changes requested):

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+	Use:          "packages [flags]",
+	Aliases:      []string{"package", "packages"},
+	Short:        "Generate package(s)",
+	Long:         "This command is used to generate curated packages",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE:         runGeneratePackages(),
+}
+
+func runGeneratePackages() func(cmd *cobra.Command, args []string) error {
```
g-gaston, line 49: curious why not just `func runGeneratePackages(cmd *cobra.Command, args []string) error` and reference with `runGeneratePackages`

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+		bundle,
+		args...,
+	)
+	packages, err := packageClient.GeneratePackages()
+	if err != nil {
+		return err
+	}
+	if err = packageClient.WritePackagesToFile(packages, gpOptions.directory); err != nil {
+		return err
+	}
```
g-gaston, line 98: There is quite a bit of orchestration in this method
I wonder if we should just keep the construction here and move the orchestration somewhere where we test it.

pkg/curatedpackages/bundlesource_test.go
```diff
@@ -0,0 +1,91 @@
...
+	type acceptCase struct {
+		// name of the test
+		name string
+		// input from the user
+		input string
+		// expected result
+		expected curatedpackages.BundleSource
+	}
+
+	accepts := []acceptCase{
```
g-gaston, line 24: can we move this to a table test with sub tests?

pkg/curatedpackages/bundlesource_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	type rejectCase struct {
+		// name of the test
+		name string
+		// input from the user
+		input string
+	}
+
+	rejects := []rejectCase{
```
g-gaston, line 43: Same here, move to a table test. And I would probably put it in a different test
So we have one test for the accepts and another for the rejects

pkg/curatedpackages/bundlesource_test.go
```diff
@@ -0,0 +1,91 @@
+package curatedpackages_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
g-gaston, line 6: Can you use `gomega` for this?

pkg/curatedpackages/bundlesource_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+)
+
+func TestBundleSourceSet(t *testing.T) {
+	h := newBSHelper(t)
+	h.Parallel()
```
g-gaston, line 13: Curious, any reason for this?

pkg/curatedpackages/packageclient.go
```diff
@@ -3,26 +3,39 @@ package curatedpackages
...
-	padChar  = '\t'
-	flags    = 0
+	minWidth        = 16
+	tabWidth        = 8
+	padding         = 0
+	padChar         = '\t'
+	flags           = 0
+	customName      = "my-"
+	kind            = "Package"
+	filePermission  = 0o644
```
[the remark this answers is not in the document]
a-cool-train: The o here was to resolve a linter issue.

pkg/curatedpackages/bundlesource_test.go
```diff
@@ -0,0 +1,91 @@
...
+	for _, testcase := range rejects {
+		h.runRejectCase(testcase.name, testcase.input)
+	}
+}
+
+//
+// Helpers
+//
+
+type bsHelper struct{ *testing.T }
```
g-gaston, line 61: I will leave it up to you but imo this code is overly _dried out_ to the point that it is difficult to follow

We basically have written our own testing framework for literally two tests. Also, it's not even making the code shorter, it's just abstracting the assertions. Putting this code directly in the table test loop might even make this smaller.

pkg/curatedpackages/packageclient.go
a-cool-train: There is no support for that as of now. There was a discussion that we would like to enforce eksa-packages for security reasons. I have added you to the discussion.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/aws/eks-anywhere-packages/pkg/artifacts"
+	"github.com/aws/eks-anywhere-packages/pkg/bundle"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/helm"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
+func NewRegistry(ctx context.Context, deps *dependencies.Dependencies, registry, kubeVersion string) (BundleRegistry, error) {
+	if registry != "" {
+		username, password, err := helm.ReadRegistryCredentials()
```
g-gaston, line 19: Imo we should either inject the configuration (user/password) or inject the configurator reader in the constructor.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/go-logr/logr"
+
+	"github.com/aws/eks-anywhere-packages/pkg/artifacts"
+	"github.com/aws/eks-anywhere-packages/pkg/bundle"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/helm"
+	"github.com/aws/eks-anywhere/pkg/version"
+)
+
+func NewRegistry(ctx context.Context, deps *dependencies.Dependencies, registry, kubeVersion string) (BundleRegistry, error) {
```
g-gaston, line 17: Maybe this logic would be better suited for the dependency factory

pkg/curatedpackages/curatedpackages.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			return nil, err
+		}
+		registry := NewCustomRegistry(
+			deps.Helm,
+			registry,
+			username,
+			password,
+		)
+		err = registry.Login(ctx)
```
g-gaston, line 29: Why don't we leave this until it is needed? Instead of doing it in the constructor (which is a bit unexpected) just call it from your "action" methods whenever is needed. 

Or you could add an `Init` method to `BundleRegistry` and make sure the callers invoke it before anything else.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		if err := os.Mkdir(directory, dirPermission); err != nil {
+			return fmt.Errorf("unable to create directory %s", directory)
+		}
+	}
+
+	for _, p := range packages {
+		displayPackage := NewDisplayablePackage(p)
+		content, err := yaml.Marshal(displayPackage)
+		if err != nil {
+			fmt.Println(fmt.Errorf("unable to parse package %s %v", p.Name, err).Error())
```
g-gaston, line 90: Can we just use `fmt.Printf`?

pkg/curatedpackages/packageclient_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func TestGeneratePackagesFail(t *testing.T) {
+	tt := newPackageTest(t)
+	packages := []string{"unknown-package"}
+	tt.command = curatedpackages.NewPackageClient(tt.bundle, packages...)
+
+	result, err := tt.command.GeneratePackages()
+	tt.Expect(err).To(BeNil())
+	tt.Expect(len(result)).To(BeEquivalentTo(0))
```
g-gaston, line 56: why not just `Equal(0)`?

pkg/curatedpackages/packagetypes.go
```diff
@@ -0,0 +1,19 @@
...
+// This is necessary in part because of https://github.com/golang/go/issues/11939
+// but also because we just don't want to generate a Status section when we're
+// emitting templates for a user to modify.
+type DisplayablePackage struct {
+	*api.Package
+	Status *interface{} `json:"status,omitempty"`
+}
+
+func NewDisplayablePackage(p api.Package) DisplayablePackage {
+	return DisplayablePackage{Package: &p}
```
g-gaston, line 18: why receive the package by value and then store a pointer?
Whatever is your decision, I would probably do the same thing for both.

Also, it's a bit misleading to return a struct by value that contains a pointer. Usually returning by value tells the caller that that object is _theirs_ and can be copied without side effects. But having a pointer inside breaks that assumption.

pkg/helm/helm.go
```diff
@@ -0,0 +1,20 @@
+package helm
```
g-gaston, line 1: I'm not completely sure this is the right package for this method. I would expect this to be in cmd/config or in other kind of "port" package. This `helm` package should be agnostic from how the user and password are read. In fact there could potentially be multiple ways to read it, depending on what's the interface/port for this codepath.

pkg/curatedpackages/packageclient.go
a-cool-train: @TerryHowe and I discussed this and we both agreed that the single YAML file is not an ideal solution here. The reasoning was that, with all the generated packages, the single file will get huge very quick and will be hard to read and update.

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+	source      curatedpackages.BundleSource
+	kubeVersion string
+	registry    string
+}
+
+var gpOptions = &generatePackageOptions{}
+
+func init() {
+	generateCmd.AddCommand(generatePackageCommand)
+	generatePackageCommand.Flags().StringVarP(&gpOptions.directory, "directory", "d", "", "Directory to save generated packages")
```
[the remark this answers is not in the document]
a-cool-train: No good reason. I'll update to that as the default.

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+			fmt.Println(fmt.Errorf("unknown package %s", p).Error())
+			continue
+		}
+		name := CustomName + strings.ToLower(bundlePackage.Name)
+		packages = append(packages, convertBundlePackageToPackage(bundlePackage, name, pc.bundle.APIVersion))
+	}
+	return packages, nil
+}
+
+func (pc *PackageClient) WritePackagesToFile(packages []packagesv1.Package, d string) error {
```
g-gaston, line 78: Could we add some tests for this method?

pkg/curatedpackages/packageclient.go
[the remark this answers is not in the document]
ewollesen: Heh, if you happen to remember the linter rule, I'd probably be for turning that one off. :D

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+		if !validations.FileExists(gepo.directory) {
+			return fmt.Errorf("directory %s does not exist", gepo.directory)
+		}
+		return generatePackages(cmd.Context(), args)
+	}
+}
+
+func generatePackages(ctx context.Context, args []string) error {
+	kubeConfig := kubeconfig.FromEnvironment()
+	deps, err := newDependenciesForPackages(ctx, kubeConfig)
```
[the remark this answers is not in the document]
a-cool-train: There are several things that are needed from the dependencies, a manifestreader, kubectl, and helm. All these three dependencies are needed by curated packages only and I believe that is the pattern that is followed in other parts of the codebase.

pkg/curatedpackages/packageclient.go
ewollesen: I don't see how one 500 line file is harder to deal with vs twenty 25 line files, but 🤷🏻

pkg/curatedpackages/packageclient.go
ewollesen: https://amzn-aws.slack.com/archives/C02EMRQAM3N/p1651087131545479?thread_ts=1648589664.078929&cid=C02EMRQAM3N

Thanks for the context.

pkg/curatedpackages/packageclient.go
a-cool-train: I did consider whether to continue or not, I decided to go with continuing and the reason is because the result here would be creating a package configuration and there is no expectation of dependency between the packages.

pkg/curatedpackages/bundlesource_test.go
ewollesen: Not saying to not use gomega, just throwing out there, that the packages team has blessed the use of testify for our code. Obviously, this is your sandbox and we should play by your rules. :D

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +59,74 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+	directory := filepath.Join(d, packageLocation)
```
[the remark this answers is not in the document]
a-cool-train: I see your viewpoint but also looking at it from another angle, if we create the files in the directory directly, there might be several files (package configurations in this case) that are created. This will create an explosion of files in the directory specified. Maybe I could add a message mentioning the location the files will be generated in.

pkg/curatedpackages/packageclient.go
a-cool-train: This appends curated-packages and checks the existence of this folder.

pkg/curatedpackages/packageclient.go
a-cool-train: Why is this necessary?

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+type generatePackageOptions struct {
+	directory   string
+	source      curatedpackages.BundleSource
+	kubeVersion string
+	registry    string
+}
+
+var gepo = &generatePackageOptions{}
```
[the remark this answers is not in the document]
a-cool-train: This is the current pattern followed in the current codebase. We can have another discussion on this separation if necessary.

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -0,0 +1,100 @@
...
+	SilenceUsage: true,
+	RunE:         runGeneratePackages(),
+}
+
+func runGeneratePackages() func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := curatedpackages.ValidateKubeVersion(gepo.kubeVersion, gepo.source); err != nil {
+			return err
+		}
+		if !validations.FileExists(gepo.directory) {
```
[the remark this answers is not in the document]
a-cool-train: The problem with that approach is that, if the user provides an existing directory with other files of different formats, the application of these files will fail. The design choice for this is to provide a single folder with only the curated-packages configurations that need to be applied.

pkg/curatedpackages/packageclient.go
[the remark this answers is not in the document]
a-cool-train: I don't know the linter rule but here is the error: ```pkg/curatedpackages/packageclient.go:26: File is not `gofumpt`-ed (gofumpt)
        filePermission  = 0600
```

File: pkg/curatedpackages/packageclient.go (hunk 19 above)
Comment (a-cool-train) [reply pkg/curatedpackages/packageclient.go:91]: Responded above

File: pkg/curatedpackages/packagetypes.go (hunk 10 above)
Comment (a-cool-train) [reply pkg/curatedpackages/packagetypes.go]: Looks like this doesn't omit the Status object.

File: pkg/curatedpackages/curatedpackages.go (hunk 33 above)
Comment (a-cool-train) [reply pkg/curatedpackages/curatedpackages.go:17]: I thought the factory was used for bootstrapping any type of dependency required from the release manifest?

File: pkg/curatedpackages/bundlesource_test.go (hunk 29 above)
Comment (a-cool-train) [reply pkg/curatedpackages/bundlesource_test.go:13]: No reason. Got rid of it.

File: cmd/eksctl-anywhere/cmd/generatepackage.go [hunk 45]
```diff
@@ -0,0 +1,100 @@
...
+	if err := generatePackageCommand.MarkFlagRequired("directory"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	generatePackageCommand.Flags().StringVar(&gepo.registry, "registry", "", "Used to specify an alternative registry for package generation")
+}
+
+var generatePackageCommand = &cobra.Command{
+	Use:          "packages [flags]",
+	Aliases:      []string{"package", "packages"},
+	Short:        "Generate package(s)",
```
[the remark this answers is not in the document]
a-cool-train: Please check whether the new wording helps.

cmd/eksctl-anywhere/cmd/generatepackage.go
a-cool-train: Added an action item to move it.

pkg/curatedpackages/packageclient.go
ewollesen: How can we know that?

pkg/curatedpackages/packageclient.go
ewollesen: [Mkdir](https://pkg.go.dev/os#Mkdir) versus [MkdirAll](https://pkg.go.dev/os#MkdirAll)

TL;DR: MkdirAll will create any missing directories along the way; so it's just safer. The only reason _not_ to use it, is if creating more than a single level of the directory hierarchy would be an error.

pkg/curatedpackages/packagetypes.go
ewollesen: Based on [the Marshal docs](https://pkg.go.dev/encoding/json#Marshal) it should have, but so long as you have a solution...

cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -9,53 +9,42 @@ import (
...
 
 type generatePackageOptions struct {
-	directory   string
 	source      curatedpackages.BundleSource
 	kubeVersion string
 	registry    string
 }
 
-var gepo = &generatePackageOptions{}
+var gpOptions = &generatePackageOptions{}
```
ewollesen, line 21: 👍🏻

ewollesen (commented):
/lgtm

pkg/curatedpackages/packageclient.go
```diff
@@ -46,3 +55,56 @@ func convertBundleVersionToPackageVersion(bundleVersions []packagesv1.SourceVers
...
+		if err != nil {
+			return fmt.Errorf("unable to parse package %s %v", p.Name, err)
+		}
+		output = append(output, content)
+	}
+	fmt.Println(string(templater.AppendYamlResources(output...)))
+	return nil
+}
+
+func (pc *PackageClient) getPackageNameToPackage() map[string]packagesv1.BundlePackage {
```
vivek-koppuru, line 87: Nit: I still think this should just be called getPackageMap() or packageMap()

pkg/curatedpackages/curatedpackages.go
```diff
@@ -0,0 +1,51 @@
...
+	return defaultRegistry, nil
+}
+
+func CreateBundleManager(kubeVersion string) bundle.Manager {
+	major, minor, err := parseKubeVersion(kubeVersion)
+	if err != nil {
+		return nil
+	}
+	log := logr.Discard()
+	discovery := NewDiscovery(major, minor)
```
vivek-koppuru, line 39: I thought something was going to change with the Discovery struct no?

vivek-koppuru (approved):

pkg/curatedpackages/curatedpackages.go
a-cool-train: Whoops don't know how I missed that. I'll include it in the next PR.

PR #1967, opened 2022-04-28
Add validation for byo git

vivek-koppuru:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1956

*Description of changes:*
Adding validation to see if the provider git repo is valid. Used https://github.com/go-git/go-git/blob/master/_examples/ls-remote/main.go as a reference.

Note that https://github.com/aws/eks-anywhere/pull/1934 needs to be merged and integrated with to properly test this feature.

/hold

*Testing (if applicable):*
e2e testing tbd

### DIFF
--- a/pkg/git/gitclient/gitclient.go
+++ b/pkg/git/gitclient/gitclient.go
@@ -266,7 +266,7 @@ func (g *GitClient) Branch(name string) error {
 	return nil
 }
 
-func (g *GitClient) Validate(ctx context.Context) error {
+func (g *GitClient) ValidateRemoteExists(ctx context.Context) error {
 	logger.V(3).Info("Validating git setup", "repoUrl", g.RepoUrl)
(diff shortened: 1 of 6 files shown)


pkg/git/gitclient/gitclient.go
```diff
@@ -250,6 +252,17 @@ func (g *GitClient) SetTokenAuth(token string, username string) {
 	g.Opts.Auth = &http.BasicAuth{Password: token, Username: username}
 }
 
+func (g *GitClient) Validate(ctx context.Context) error {
```
danbudris, line 255:
```suggestion
func (g *GitClient) ValidateRemoteExists(ctx context.Context) error {
```

danbudris: validate might be too broad, we only want to check if the remote exists in this case

vivek-koppuru:
I wonder why the project percentage is increasing but the patch percentage is lesser.

PR #1968, opened 2022-04-28
Initial implementation of generating snow objects for upgrade

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/common/common.go
```diff
@@ -97,3 +99,22 @@ func KubeadmConfigTemplateName(clusterName, workerNodeGroupName string, now type
...
+	if match == nil {
+		return "", fmt.Errorf(`invalid format of name [name=%s]. Name has to follow regex pattern "(-)(\d+)$", e.g. machinetemplate-cp-1`, name)
+	}
+
+	n, err := strconv.Atoi(match[2])
+	if err != nil {
+		return "", fmt.Errorf("converting object suffix to int: %v", err)
+	}
+
+	return re.ReplaceAllString(name, fmt.Sprintf("${1}%d", n+1)), nil
```
g-gaston, line 115: can we have another method like
`ObjectName(baseName string, version int) string`

And we can reuse that from `DefaultObjectName`. That way there is only one place where the actual name is built.

pkg/providers/common/common.go
```diff
@@ -97,3 +99,22 @@ func KubeadmConfigTemplateName(clusterName, workerNodeGroupName string, now type
 	t := now().UnixNano() / int64(time.Millisecond)
 	return fmt.Sprintf("%s-%s-template-%d", clusterName, workerNodeGroupName, t)
 }
+
+func IncrementName(name string) (string, error) {
+	re := regexp.MustCompile(`(-)(\d+)$`)
```
g-gaston, line 104: `MustCompile` will panic for an invalid regex. If we want to use that, we should probably move it to a global `var`. Otherwise we should just build it normally and return an error

pkg/providers/common/common.go
```diff
@@ -97,3 +99,22 @@ func KubeadmConfigTemplateName(clusterName, workerNodeGroupName string, now type
 	t := now().UnixNano() / int64(time.Millisecond)
 	return fmt.Sprintf("%s-%s-template-%d", clusterName, workerNodeGroupName, t)
 }
+
+func IncrementName(name string) (string, error) {
```
g-gaston, line 103: I think we should move this with the rest of the `clusterapi` generating code

pkg/providers/snow/fetch.go
```diff
@@ -0,0 +1,38 @@
...
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	snowv1 "github.com/aws/eks-anywhere/pkg/providers/snow/api/v1beta1"
+)
+
+func oldControlPlaneMachineTemplate(ctx context.Context, kubeClient KubeClient, oldSpec *cluster.Spec) (*snowv1.AWSSnowMachineTemplate, error) {
+	cp := &controlplanev1.KubeadmControlPlane{}
+	if err := kubeClient.Get(ctx, oldSpec.Cluster.GetName(), constants.EksaSystemNamespace, cp); err != nil {
```
g-gaston, line 16: could we have a single method that we use to derive the the `KubeadmControlPlane` from a `clusterSpec`? That way we don't need to know here that kubeadmCP are named after the eksa cluster

Or maybe we should do it with labels?

pkg/providers/snow/fetch.go
```diff
@@ -0,0 +1,38 @@
...
+	mt := &snowv1.AWSSnowMachineTemplate{}
+	if err := kubeClient.Get(ctx, cp.Spec.MachineTemplate.InfrastructureRef.Name, constants.EksaSystemNamespace, mt); err != nil {
+		return nil, err
+	}
+	return mt, nil
+}
+
+func oldWorkerMachineTemplate(ctx context.Context, kubeclient KubeClient, workerNodeGroupConfigName string) (*snowv1.AWSSnowMachineTemplate, error) {
+	md := &clusterv1.MachineDeployment{}
+	if err := kubeclient.Get(ctx, workerNodeGroupConfigName, constants.EksaSystemNamespace, md); err != nil {
```
g-gaston, line 29: same comment here about a single method for names
or labels

pkg/providers/snow/objects.go
```diff
@@ -0,0 +1,126 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/providers/common"
+	snowv1 "github.com/aws/eks-anywhere/pkg/providers/snow/api/v1beta1"
+)
+
+func ControlPlaneObjectsForCreate(clusterSpec *cluster.Spec) ([]runtime.Object, error) {
```
g-gaston, line 16: Could we have only one set of methods, without making the distinction between create an upgrade

pkg/providers/snow/objects.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func NewMachineTemplateName(new, old *snowv1.AWSSnowMachineTemplate) (string, error) {
+	if equality.Semantic.DeepDerivative(new.Spec, old.Spec) {
+		return old.GetName(), nil
+	}
+	return common.IncrementName(old.GetName())
+}
+
+func NewMachineTemplate(new, old *snowv1.AWSSnowMachineTemplate) error {
```
g-gaston, line 100: What about something like
```suggestion
func UpdateMachineTemplateName(new, old *snowv1.AWSSnowMachineTemplate) error {
```

pkg/providers/snow/objects.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, err
+	}
+	workerMachineTemplates := SnowMachineTemplates(clusterSpec)
+	machineDeployments := MachineDeployments(clusterSpec, kubeadmConfigTemplates, workerMachineTemplates)
+
+	return concatWorkersObjects(machineDeployments, kubeadmConfigTemplates, workerMachineTemplates), nil
+}
+
+func ControlPlaneObjectsForUpgrade(ctx context.Context, oldSpec, newSpec *cluster.Spec, kubeClient KubeClient) ([]runtime.Object, error) {
```
g-gaston, line 39: What do we need the `oldSpec` for? I thought comparing the output avoided needed the old Spec
Ideally we should avoid that, because it's no going to available in the controller

jiayiwang7: we need it to fetch the old cp/md in order to get old machinetemplates for comparision

PR #1974, opened 2022-04-28
Fix packages install test name

abhay-krishna:
The test name was not matching the regex required to set up vSphere environment.


TerryHowe (approved):
/lgtm

PR #1976, opened 2022-04-29
Feature install packages

a-cool-train:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/install.go
+++ b/cmd/eksctl-anywhere/cmd/install.go
@@ -1,0 +1,13 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var installCmd = &cobra.Command{
+	Use:   "install",
+	Short: "Install resources to the cluster",
+	Long:  "Use eksctl anywhere install to install artifacts into a cluster",
+}
+
+func init() {
+	rootCmd.AddCommand(installCmd)
+}
--- a/cmd/eksctl-anywhere/cmd/install.go
+++ b/cmd/eksctl-anywhere/cmd/install.go
@@ -5,7 +5,7 @@ import "github.com/spf13/cobra"
 var installCmd = &cobra.Command{
 	Use:   "install",
 	Short: "Install resources to the cluster",
-	Long:  "Use eksctl anywhere install to install artifacts into a cluster",
+	Long:  "Use eksctl anywhere install to install resources into a cluster",
 }
 
 func init() {
--- a/cmd/eksctl-anywhere/cmd/installpackages.go
+++ b/cmd/eksctl-anywhere/cmd/installpackages.go
@@ -36,9 +36,9 @@ func init() {
 }
 
 var installPackageCommand = &cobra.Command{
-	Use:          "package [package-name] [flags]",
+	Use:          "package [package] [flags]",
 	Aliases:      []string{"package"},
-	Short:        "Install package(s)",
+	Short:        "Install package",
 	Long:         "This command is used to Install a curated package. Use list to discover curated packages",
 	PreRunE:      preRunPackages,
 	SilenceUsage: true,
--- a/cmd/eksctl-anywhere/cmd/installpackages.go
+++ b/cmd/eksctl-anywhere/cmd/installpackages.go
@@ -47,7 +47,6 @@ var installPackageCommand = &cobra.Command{
 
 func runInstallPackages() func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-
 		if err := curatedpackages.ValidateKubeVersion(ipo.kubeVersion, ipo.source); err != nil {
 			return err
 		}
--- a/cmd/eksctl-anywhere/cmd/installpackages.go
+++ b/cmd/eksctl-anywhere/cmd/installpackages.go
@@ -14,10 +14,10 @@ import (
 )
 
 type installPackageOptions struct {
-	source      curatedpackages.BundleSource
-	kubeVersion string
-	packageName string
-	registry    string
+	source       curatedpackages.BundleSource
+	kubeVersion  string
+	resourceName string
+	registry     string
 }
 
 var ipo = &installPackageOptions{}
@@ -28,9 +28,9 @@ func init() {
 	if err := installPackageCommand.MarkFlagRequired("source"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
 	}
-	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
-	installPackageCommand.Flags().StringVarP(&ipo.packageName, "packagename", "p", "", "Custom name of the curated package to install")
-	if err := installPackageCommand.MarkFlagRequired("packagename"); err != nil {
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kube-version", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	installPackageCommand.Flags().StringVarP(&ipo.resourceName, "package-name", "n", "", "Custom name of the curated package to install")
+	if err := installPackageCommand.MarkFlagRequired("package-name"); err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
 	}
 }
@@ -94,7 +94,7 @@ func installPackages(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
-	err = packages.InstallPackage(ctx, p, ipo.packageName, kubeConfig)
+	err = packages.InstallPackage(ctx, p, ipo.resourceName, kubeConfig)
 	if err != nil {
 		return err
 	}
--- a/cmd/eksctl-anywhere/cmd/installpackages.go
+++ b/cmd/eksctl-anywhere/cmd/installpackages.go
@@ -1,0 +1,105 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/version"
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+)
+
+type installPackageOptions struct {
+	source      curatedpackages.BundleSource
+	kubeVersion string
+	name        string
+	registry    string
+}
+
+var ipo = &installPackageOptions{}
+
+func init() {
+	installCmd.AddCommand(installPackageCommand)
+	installPackageCommand.Flags().Var(&ipo.source, "source", "Location to find curated packages: (cluster, registry)")
+	if err := installPackageCommand.MarkFlagRequired("source"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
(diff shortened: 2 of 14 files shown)


cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,104 @@
...
+var ipo = &installPackageOptions{}
+
+func init() {
+	installCmd.AddCommand(installPackageCommand)
+	installPackageCommand.Flags().Var(&ipo.source, "source", "Location to find curated packages: (cluster, registry)")
+	if err := installPackageCommand.MarkFlagRequired("source"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	installPackageCommand.Flags().StringVar(&ipo.name, "name", "", "Custom name of the curated package to install")
```
vivek-koppuru, line 32: Why can't this be `package` instead? name makes me think of namespace, and I think it is okay to be like `eksctl anywhere install packages -p my-curated-package`. This is similar to how kubectl does `kubectl create deployment my-dep --image=busybox`

cmd/eksctl-anywhere/cmd/install.go
```diff
@@ -0,0 +1,13 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var installCmd = &cobra.Command{
+	Use:   "install",
+	Short: "Install resources to the cluster",
+	Long:  "Use eksctl anywhere install to install artifacts into a cluster",
```
g-gaston, line 8: nit:
```suggestion
	Long:  "Use eksctl anywhere install to install resources into a cluster",
```

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,104 @@
...
+	Use:          "packages [flags]",
+	Aliases:      []string{"package", "packages"},
+	Short:        "Install package(s)",
+	Long:         "This command is used to Install a curated package. Use list to discover curated packages",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE:         runInstallPackages(),
+}
+
+func runInstallPackages() func(cmd *cobra.Command, args []string) error {
```
g-gaston, line 48: Do we need to return a closure or can we just
```suggestion
func runInstallPackages(cmd *cobra.Command, args []string) error {
```

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,104 @@
...
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	installPackageCommand.Flags().StringVar(&ipo.name, "name", "", "Custom name of the curated package to install")
+	if err := installPackageCommand.MarkFlagRequired("name"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var installPackageCommand = &cobra.Command{
+	Use:          "packages [flags]",
+	Aliases:      []string{"package", "packages"},
```
g-gaston, line 40: curious: what's the reason to allow `install packages` if users can only install one package?

pkg/curatedpackages/packageclient.go
```diff
@@ -84,14 +87,39 @@ func (pc *PackageClient) WritePackagesToStdOut(packages []packagesv1.Package) er
...
 }
 
-func (pc *PackageClient) getPackageNameToPackage() map[string]packagesv1.BundlePackage {
+func (pc *PackageClient) GetPackageFromBundle(packageName string) (*packagesv1.BundlePackage, error) {
+	packageMap := pc.packageMap()
+	p := packageMap[strings.ToLower(packageName)]
+	if p.Name != "" {
+		return &p, nil
+	}
+	return nil, fmt.Errorf("package %s not found", packageName)
```
g-gaston, line 96:
```suggestion
	p, ok := packageMap[strings.ToLower(packageName)]
	if !ok {
		return nil, fmt.Errorf("package %s not found", packageName)
	}
	
	return &p, nil
```

pkg/curatedpackages/packageclient.go
```diff
@@ -84,14 +87,39 @@ func (pc *PackageClient) WritePackagesToStdOut(packages []packagesv1.Package) er
 	return nil
 }
 
-func (pc *PackageClient) getPackageNameToPackage() map[string]packagesv1.BundlePackage {
+func (pc *PackageClient) GetPackageFromBundle(packageName string) (*packagesv1.BundlePackage, error) {
```
g-gaston, line 90: Can we add tests for this method?

pkg/curatedpackages/packageclient.go
```diff
@@ -84,14 +87,39 @@ func (pc *PackageClient) WritePackagesToStdOut(packages []packagesv1.Package) er
...
+
+func (pc *PackageClient) packageMap() map[string]packagesv1.BundlePackage {
 	pMap := make(map[string]packagesv1.BundlePackage)
 	for _, p := range pc.bundle.Spec.Packages {
 		pMap[strings.ToLower(p.Name)] = p
 	}
 	return pMap
 }
 
+func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.BundlePackage, customName string, kubeConfig string) error {
```
g-gaston, line 107: tests?

pkg/curatedpackages/packageclient.go
```diff
@@ -84,14 +87,39 @@ func (pc *PackageClient) WritePackagesToStdOut(packages []packagesv1.Package) er
...
+	p := convertBundlePackageToPackage(*bp, customName, pc.bundle.APIVersion)
+	displayPackage := NewDisplayablePackage(&p)
+	params := []string{"create", "-f", "-", "--kubeconfig", kubeConfig}
+	packageYaml, err := yaml.Marshal(displayPackage)
+	if err != nil {
+		return err
+	}
+	stdOut, err := pc.kubectl.ExecuteCommandFromBytes(ctx, packageYaml, params...)
```
g-gaston, line 115: I would maybe use `CreateFromYaml` and pass packageYaml and kubeconfig as args

Also, this will fail if the package already exists, is that the intended behavior?

cmd/eksctl-anywhere/cmd/installpackages.go
a-cool-train: I didn't push the change but this has been fixed before. Sorry for the confusion.

cmd/eksctl-anywhere/cmd/installpackages.go
a-cool-train: That's a good point. Let me change that.

pkg/curatedpackages/packageclient.go
a-cool-train: Yes if a package already exists, it should fail.

cmd/eksctl-anywhere/cmd/installpackages.go
a-cool-train: A package has a conceptual meaning in curated packages. If we were to use a different name than just name it would be packageName and -p as the short form. Would that work better than name? Thoughts?

vivek-koppuru: I like that

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,102 @@
...
+var ipo = &installPackageOptions{}
+
+func init() {
+	installCmd.AddCommand(installPackageCommand)
+	installPackageCommand.Flags().Var(&ipo.source, "source", "Location to find curated packages: (cluster, registry)")
+	if err := installPackageCommand.MarkFlagRequired("source"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	installPackageCommand.Flags().StringVarP(&ipo.packageName, "packagename", "p", "", "Custom name of the curated package to install")
```
vivek-koppuru, line 32: We can camelcase packagename and kubeversion right?

a-cool-train: This is client's flag. I believe other commands use the same format as above. For example filename instead of fileName for the flag.

TerryHowe: I think packageName is better because it disambiguates with the name of the installation.

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := installPackageCommand.MarkFlagRequired("packagename"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var installPackageCommand = &cobra.Command{
+	Use:          "packages [flags]",
```
g-gaston, line 39: should this be
```suggestion
	Use:          "package [package-name] [flags]",
```
or something like that?

pkg/curatedpackages/packageclient.go
```diff
@@ -84,14 +87,39 @@ func (pc *PackageClient) WritePackagesToStdOut(packages []packagesv1.Package) er
...
 
+func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.BundlePackage, customName string, kubeConfig string) error {
+	p := convertBundlePackageToPackage(*bp, customName, pc.bundle.APIVersion)
+	displayPackage := NewDisplayablePackage(&p)
+	params := []string{"create", "-f", "-", "--kubeconfig", kubeConfig}
+	packageYaml, err := yaml.Marshal(displayPackage)
+	if err != nil {
+		return err
+	}
+	stdOut, err := pc.kubectl.CreateFromYaml(ctx, packageYaml, params...)
```
g-gaston, line 115: If the method is named `CreateFromYaml` I would expect the signature to be something like:
`CreateFromYaml(ctx context.Context, yaml []byte, kubeconfig string) (bytes.Buffer, error)`

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	installPackageCommand.Flags().StringVar(&ipo.kubeVersion, "kubeversion", "", "Kubernetes Version of the cluster to be used. Format <major>.<minor>")
+	installPackageCommand.Flags().StringVarP(&ipo.packageName, "packagename", "p", "", "Custom name of the curated package to install")
+	if err := installPackageCommand.MarkFlagRequired("packagename"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var installPackageCommand = &cobra.Command{
+	Use:          "package [package-name] [flags]",
```
vivek-koppuru, line 39: Why do we reference `package-name` here and have a flag for `packageName`? It should be one or the other I thought

cmd/eksctl-anywhere/cmd/installpackages.go
```diff
@@ -0,0 +1,102 @@
...
+	Aliases:      []string{"package"},
+	Short:        "Install package(s)",
```
vivek-koppuru, line 41: Don't need the s

cmd/eksctl-anywhere/cmd/installpackages.go
a-cool-train: There are two items. The first one is the package to install like harbor (in this case [package-name] and the flag packageName is to name the custom resource like my-harbor or something like that.

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #1981, opened 2022-04-29
Fix/missing certs mount

wongni:
*Issue #, if available:* #2054

*Description of changes:*
verify-ssl: true does not work for CloudStack provider because the host paths storing SSL cert files are not mounted to the eksa cli tools container. 

*Testing (if applicable):*


jaxesn:
What is the error you are seeing?  i think the cli-tools image should have the standard certs in it, not sure if its exactly those same paths you are mounting since it changes slightly between different distros.

wongni:
> What is the error you are seeing? i think the cli-tools image should have the standard certs in it, not sure if its exactly those same paths you are mounting since it changes slightly between different distros.

In the FS network, I got `x509: certificate signed by unknown authority` at the cmk (CloudStack CloudMonkey tool) command when the cmk tool connects to the customer's ACS via https.

jaxesn:
and this fixes it because the host machine has the custom root ca in that /etc folder?

wongni:
> and this fixes it because the host machine has the custom root ca in that /etc folder?

Yes.

jaxesn:
I think this has come up before and probably something we should solve for, but I'm not sure if this is the proper approach. Will the capc controller have the same issue once it's running in the cluster or does the node image have these root CAs built in? It seems like something we should allow the user to provide at creation time and we make sure and put them in the various key locations.

wongni:
> I think this has come up before and probably something we should solve for, but I'm not sure if this is the proper approach. Will the capc controller have the same issue once it's running in the cluster or does the node image have these root CAs built in? It seems like something we should allow the user to provide at creation time and we make sure and put them in the various key locations.

As far as I know, the node image already have these custom root CAs built in so capc doesn't have any problem with this. I agree on that we should take user inputs for these mounting host paths because they probably are different across the OSes and their distributions.

vignesh-goutham (approved):
I was talking with @jaxesn  about this, this looks good to go, but can you open an issue to get the paths from user, so that we can make it better user experience.

Looks like this needs more unit tests.

vivek-koppuru (commented):
This seems fine but we have a mix of using env vars like for CA Cert and now introducing a flag. Is it okay if we start off with an env var and have an effort to move all of our env vars to flags so that we don't have inconsistency there?

wongni:
> 



> This seems fine but we have a mix of using env vars like for CA Cert and now introducing a flag. Is it okay if we start off with an env var and have an effort to move all of our env vars to flags so that we don't have inconsistency there?

The flag I just added in this PR is a general purpose one, and one of the use cases of this flag is to mount host paths containing custom CA certs. However, it's the only use case of the flag for now. Do you still think adding this flag causes inconsistency?

PR #1985, opened 2022-04-29
Generic git flow

danbudris:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -140,7 +140,7 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		WithBootstrapper().
 		WithClusterManager(clusterSpec.Cluster).
 		WithProvider(cc.fileName, clusterSpec.Cluster, cc.skipIpCheck, cc.hardwareFileName, cc.skipPowerActions, cc.setupTinkerbell, cc.forceClean).
-		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig).
+		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig, cliConfig).
 		WithWriter().
 		WithEksdInstaller().
 		Build(ctx)
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -209,6 +209,7 @@ func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cl
 
 	if cliConfig.GitPrivateKeyFile != "" {
 		dirs = append(dirs, cliConfig.GitPrivateKeyFile)
+		dirs = append(dirs, "~/.ssh/known_hosts")
 	}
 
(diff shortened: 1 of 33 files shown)


cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -206,10 +207,8 @@ func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cl
 	if fluxConfig == nil || fluxConfig.Spec.Git == nil {
 		return dirs
 	}
-
-	if cliConfig.GitPrivateKeyFile != "" {
-		dirs = append(dirs, cliConfig.GitPrivateKeyFile)
-	}
+	dirs = append(dirs, filepath.Dir(cliConfig.GitPrivateKeyFile))
+	dirs = append(dirs, filepath.Dir(cliConfig.GitKnownHostsFile))
```
danbudris, line 211: we now need to mount a known hosts file into the cli tools container in order to securely bootstrap flux; the user will provide an env var, `EKSA_GIT_KNOWN_HOSTS`, which points to a known hosts file; we then mount that file into the tools container and tell the flux execution to look forthe known hosts at that file path -- for some background, https://security.stackexchange.com/a/20710 is useful.

mitalipaygude: why remove the git private key if condition? we are mounting an empty string?

pkg/dependencies/factory.go
```diff
@@ -772,7 +775,7 @@ func (f *Factory) WithGit(clusterConfig *v1alpha1.Cluster, fluxConfig *v1alpha1.
 	return f
 }
 
-func (f *Factory) WithFluxAddonClient(clusterConfig *v1alpha1.Cluster, fluxConfig *v1alpha1.FluxConfig) *Factory {
+func (f *Factory) WithFluxAddonClient(clusterConfig *v1alpha1.Cluster, fluxConfig *v1alpha1.FluxConfig, cliConfig *config.CliConfig) *Factory {
```
chrisdoherty4, line 778: Did you consider making this its own constructable thing in the factory instead of injecting it? I see for all 3 of our main operations - create, upgrade and delete - we inject nil. Passing nil as a parameter is an anti-pattern and considering those 3 pretty significant use-cases are passing nothing I feel like we've got this wrong.

chrisdoherty4: Scratch that, 2/3. The point still stands though.

cmd/eksctl-anywhere/cmd/createcluster.go
danbudris: we now have a validation which confirms that the private key file is non-empty at a higher level, since it's required, so we don't need to check that here

pkg/dependencies/factory.go
danbudris: yes, good call. I'm actually going to be passing that value into the create and delete flows as well, as we'll be needing them -- it wasn't present in this draft PR but your point is well taken

danbudris:
closing in favor of https://github.com/aws/eks-anywhere/pull/1992 which is rebased to exclude upgrade flow

Issue #1987, opened 2022-04-29
GitOpsConfigDeprecation


danbudris:
Tracking places where we need to clean up and deprecate the GitOpsConfig in favor of the FluxConfig.

PR #1988, opened 2022-04-29
support `upgrade cluster` with `FluxConfig` present

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1986

*Description of changes:*
Set up to support `eksctl anywhere upgrade cluster` with the `fluxConfig` present on the cluster.

Many of the methods in the fetcher and cluster spec builder shadow the existing GitOpsConfig methods. I've captured an issue to come back and remove these redundant methods once we deprecate the GitOpsConfig, will link this PR so that in +2 minor versions we can remove the redundant methods. https://github.com/aws/eks-anywhere/issues/1987 

*Testing (if applicable):*
create a cluster with the `fluxConfig`; upgrade the cluster.

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -16,19 +16,36 @@ type BundlesFetch func(ctx context.Context, name, namespace string) (*v1alpha1re
 
 type GitOpsFetch func(ctx context.Context, name, namespace string) (*v1alpha1.GitOpsConfig, error)
 
+type FluxConfigFetch func(ctx context.Context, name, namespace string) (*v1alpha1.FluxConfig, error)
+
 type EksdReleaseFetch func(ctx context.Context, name, namespace string) (*eksdv1alpha1.Release, error)
 
 type OIDCFetch func(ctx context.Context, name, namespace string) (*v1alpha1.OIDCConfig, error)
 
-func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, oidcFetch OIDCFetch) (*Spec, error) {
+func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, fluxConfigFetch FluxConfigFetch, oidcFetch OIDCFetch) (*Spec, error) {
 	bundles, err := GetBundlesForCluster(ctx, cluster, bundlesFetch)
 	if err != nil {
 		return nil, err
 	}
-	gitOpsConfig, err := GetGitOpsForCluster(ctx, cluster, gitOpsFetch)
-	if err != nil {
-		return nil, err
+
+	var fluxConfig *v1alpha1.FluxConfig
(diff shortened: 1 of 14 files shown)


pkg/cluster/fetch.go
```diff
@@ -49,6 +66,18 @@ func GetBundlesForCluster(ctx context.Context, cluster *v1alpha1.Cluster, fetch
 	return bundles, nil
 }
 
+func GetFluxConfigForCluster(ctx context.Context, cluster *v1alpha1.Cluster, fetch FluxConfigFetch) (*v1alpha1.FluxConfig, error) {
+	if fetch == nil || cluster.Spec.GitOpsRef == nil {
+		return nil, nil
+	}
+	fluxConfig, err := fetch(ctx, cluster.Spec.GitOpsRef.Name, cluster.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed fetching FluxCOnfig for cluster: %v", err)
```
vivek-koppuru, line 75:
```suggestion
		return nil, fmt.Errorf("failed fetching FluxConfig for cluster: %v", err)
```

vivek-koppuru: We had a recent initiative that removed words like fail and error in the error messages and make it something like `fetching FluxConfig for cluster`

pkg/cluster/fetch_test.go
```diff
@@ -32,3 +32,85 @@ func TestGetBundlesForCluster(t *testing.T) {
...
+			},
+		},
+	}
+	wantFlux := &v1alpha1.FluxConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "",
+			APIVersion: "",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:                       "",
```
vivek-koppuru, line 80: We don't list to list these fields then right? It would be cleaner to just specify an empty struct.

pkg/cluster/fetch.go
danbudris: yeah I just ran out of creativity at this point :)

pkg/cluster/fetch.go
```diff
@@ -16,19 +16,36 @@ type BundlesFetch func(ctx context.Context, name, namespace string) (*v1alpha1re
...
 	bundles, err := GetBundlesForCluster(ctx, cluster, bundlesFetch)
 	if err != nil {
 		return nil, err
 	}
-	gitOpsConfig, err := GetGitOpsForCluster(ctx, cluster, gitOpsFetch)
-	if err != nil {
-		return nil, err
+
+	var fluxConfig *v1alpha1.FluxConfig
+	var gitOpsConfig *v1alpha1.GitOpsConfig
```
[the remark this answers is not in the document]
danbudris: all support for the GitOpsConfig will be short lived, as we will remove it entirely in a near future minor version.

while we're converting the GitOpsConfig to a FluxConfig for internal use, when we marshall it and write it to the cluster, we still write it as a  GitOpsConfig crd, so that a user who provides a GitOpsConfig won't get a confusing result like us replacing their GitOpsConfig with a flux config. So when fetching objects from the cluster for inspection we still need to check for the existence of the GitOpsConfig

pkg/cluster/fetch_test.go
danbudris: fixed

pkg/cluster/fetch.go
danbudris: fixed

PR #1989, opened 2022-04-30
Fix flux tests to clone repository in the right path

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Recent changes in refactoring resulted in the clone path to follow the same path as where we edit the cluster config file for the flux tests, causing the tests to fail. Added the option to override the default path that is set so we don't affect the CLI code as well just for e2e test validations.

*Testing (if applicable):*
Ran e2e test

### DIFF
--- a/pkg/git/factory/gitfactory.go
+++ b/pkg/git/factory/gitfactory.go
@@ -14,17 +14,20 @@ import (
 )
 
 type GitTools struct {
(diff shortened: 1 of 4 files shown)


danbudris:
ah man good catch, thanks for finding and fixing that.

PR #1990, opened 2022-04-30
Add `Delete` method to kube client

g-gaston:
Part of https://github.com/aws/eks-anywhere/issues/1106, needed only in snow cli provider and not in the controller for now

*Testing (if applicable):*
Uni tests


g-gaston:
> should we have a method `func (c *KubeconfigClient) Delete(ctx context.Context, name, namespace string, obj runtime.Object) error` in kubeconfig.go?

Unless I'm missing something, the delete functionality is only needed at the provider level and not at the "templater" level. So we could add it, but I don't think we need it yet.

Issue #1043, opened 2022-01-26
Support a BYO git provider in additional to github.com


jaxesn:
**What would you like to be added**:

**Why is this needed**:

PR #1992, opened 2022-05-02
Git flux flow end to end

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1043

*Description of changes:*
Draft PR containing everything needed to get a generic git provider 'working' -- there's a number of sharp edges here, this will be actively rebased and split up. Opening as a draft to start getting initial feedback and start to identify areas for immediate improvement.

Things this contains:
- flow to execute `fluxBootstrapGit` in the `fluxAddonClient`
- add ssh known hosts to the `flux` executable containter
    - env var, `EKSA_GIT_KNOWN_HOSTS`, which points to a known hosts file containing the SSH known hosts associated with the given private key
    - mount the given known hosts file into the tools container
    - pass the env var `SSH_KNOWN_HOSTS` to the execution of the `flux` command, so that it looks at the mount known hosts file when executing git commands with SSH auth
- get `upgrade` working
    - account for the fact that we now have two kinds of gitops configs in the resource fetchers
- clean up and condition dependencies on GitHub spec 
- make sure we're building the FluxConfig crds

## Generic Git Provider Flow

user provides a path to a private key and known hosts file to the EKS Anywhere CLI, via the `EKSA_GIT_PRIVATE_KEY` and `EKSA_GIT_KNOWN_HOSTS` environment variables.

EKS Anywhere mounts the known hosts file and private key file into the EKS A tools container.

When we execute `flux bootstrap git`, we pass an argument telling flux to use the mounted private key file, and an env var (`SSH_KNOWN_HOSTS`) telling it where to find the known hosts file. 

flux then uses the given private key to clone the repo and initialize flux, and then uses the known hosts and private key to create the `flux-system/flux-system` secret on the cluster; this secret is used by the flux compoents (like source controller) for ongoing authentication to the repository. 

## The addition of SSH known hosts to EKSA Generic Git Provider

In order to securely authenticate to an arbitrary git repository via SSH we now need to mount a known hosts file into the cli tools container when running `flux bootstrap git`. The user will provide a path to a known hosts file via an env var, `EKSA_GIT_KNOWN_HOSTS`. We then mount that file into the tools container and tell the flux execution to look for the known hosts at that file path via the environment variable `SSH_KNOWN_HOSTS`. Some potentially useful background on known hosts: https://security.stackexchange.com/a/20710.

Documentation will be critical to ensure that the known hosts are set up correctly. Additional validations of the known hosts file should also be on the table here.

*Testing (if applicable):*
created cluster using a FluxConfig gitopsRef

upgraded the given cluster

deleted the given cluster


cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -139,7 +140,7 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		WithBootstrapper().
 		WithClusterManager(clusterSpec.Cluster).
 		WithProvider(cc.fileName, clusterSpec.Cluster, cc.skipIpCheck, cc.hardwareFileName, cc.skipPowerActions, cc.setupTinkerbell, cc.forceClean).
-		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig).
+		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig, cliConfig).
```
vivek-koppuru, line 143: I could see the cliConfig change in the future where if we need to pass in env vars from the controller and build this client, we would have a different abstraction here. But since the only uses of this method is from the CLI, I wouldn't worry about this now.

cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -98,7 +98,7 @@ func (dc *deleteClusterOptions) deleteCluster(ctx context.Context) error {
 		WithBootstrapper().
 		WithClusterManager(clusterSpec.Cluster).
 		WithProvider(dc.fileName, clusterSpec.Cluster, cc.skipIpCheck, dc.hardwareFileName, cc.skipPowerActions, cc.setupTinkerbell, false).
-		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig).
+		WithFluxAddonClient(clusterSpec.Cluster, clusterSpec.FluxConfig, nil).
```
vivek-koppuru, line 101: If this is nil everywhere else except for create, can we make this an option param? I did it recently here: https://github.com/aws/eks-anywhere/blob/main/pkg/git/factory/gitfactory.go#L23

vivek-koppuru: But thinking about, might be overkill if we are only setting it in the constructor and not doing anything else with it. I am not sold on this.

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -121,6 +129,38 @@ func (f *FluxAddonClient) InstallGitOps(ctx context.Context, cluster *types.Clus
...
 
+	if clusterSpec.FluxConfig.Spec.Github != nil {
+		err := f.installGitOpsGithub(ctx, cluster, fc, clusterSpec)
+		if err != nil {
+			return fmt.Errorf("installing GitHub gitops: %v", err)
+		}
+	}
+
+	if clusterSpec.FluxConfig.Spec.Git != nil {
+		err := f.installGitOpsGenericGit(ctx, cluster, fc, clusterSpec)
```
vivek-koppuru, line 140: Do we need to be explicit about Generic git? Git is not enough? I might like GithubProvider and GitProvider naming better if that is the issue, but that also shows that having a provider interface might be better.

pkg/addonmanager/addonclients/fluxaddonclient.go
```diff
@@ -530,9 +575,26 @@ func (fc *fluxForCluster) setupRepository(ctx context.Context) error {
...
+
+	logger.V(3).Info("Creating a new branch")
+	err = fc.gitTools.Client.Branch(fc.branch())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// cloneIfExists attempts to describe the provided remote repository via the Git Provider.
```
vivek-koppuru, line 595: Won't ls-remote check for this too? Or is this specifically calling GitHub api? If we are separating these functions based on provider then it shows a bigger selling point for having providers separate.

pkg/api/v1alpha1/fluxconfig.go
```diff
@@ -67,9 +67,6 @@ func validateFluxConfig(config *FluxConfig) error {
 }
 
 func validateGitProviderConfig(config GitProviderConfig) error {
-	if len(config.Username) <= 0 {
```
vivek-koppuru, line 70: username is not needed anymore?

pkg/git/factory/gitfactory.go
```diff
@@ -36,13 +46,23 @@ func Build(ctx context.Context, cluster *v1alpha1.Cluster, fluxConfig *v1alpha1.
...
+		if err != nil {
+			return nil, fmt.Errorf("building github provider: %v", err)
+		}
+
+		gitAuth = &http.BasicAuth{Password: githubToken, Username: fluxConfig.Spec.Github.Owner}
 		repo = fluxConfig.Spec.Github.Repository
 		repoUrl = github.RepoUrl(fluxConfig.Spec.Github.Owner, repo)
-		tokenAuth = &git.TokenAuth{Token: githubToken, Username: fluxConfig.Spec.Github.Owner}
-		tools.Provider, err = buildGithubProvider(ctx, *tokenAuth, fluxConfig.Spec.Github)
+	case fluxConfig.Spec.Git != nil:
```
vivek-koppuru, line 57: I don't like the idea of having a case here like this, all of this information should ideally be abstracted into methods for a specific provider.

cmd/eksctl-anywhere/cmd/deletecluster.go
danbudris: I'm going to pass a CLI config for these

pkg/api/v1alpha1/fluxconfig.go
danbudris: username is no longer needed -- any username information is contained in in the SSH URL provided by the user

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: to recap discussion with @vivek-koppuru -- 
we're going to take another look at the abstractions we're using here, and ideally implement some more narrow interfaces that we can use for Repository Management and other provider-specific actions, and get this logic out of the Flux workflow. That is going to be done as an ongoing effort after this pr is merged and we have end-to-end tests in place, ideally for the next+1 milestone, allowing us to both release this working feature for users and continuing to improve the overall maintainability. See: https://github.com/aws/eks-anywhere/issues/1993

pkg/addonmanager/addonclients/fluxaddonclient.go
danbudris: see above comment and https://github.com/aws/eks-anywhere/issues/1993

pkg/git/factory/gitfactory.go
danbudris: see above comment and https://github.com/aws/eks-anywhere/issues/1993

cmd/eksctl-anywhere/cmd/deletecluster.go
danbudris: updated to include the config here

vivek-koppuru (approved):

PR #1996, opened 2022-05-03
Add name util functions for capi apibuilder

jiayiwang7:
*Issue #, if available:*

#1106 

Break down the POC upgrade PR #1968 

*Description of changes:*

Part of the snow upgrade flow. 

*Testing (if applicable):*

unit tests

### DIFF
--- a/pkg/clusterapi/name.go
+++ b/pkg/clusterapi/name.go
@@ -1,0 +1,42 @@
+package clusterapi
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/clusterapi/name.go
```diff
@@ -0,0 +1,42 @@
...
+	if match == nil {
+		return "", fmt.Errorf(`invalid format of name [name=%s]. Name has to follow regex pattern "(-)(\d+)$", e.g. machinetemplate-cp-1`, name)
+	}
+
+	n, err := strconv.Atoi(match[2])
+	if err != nil {
+		return "", fmt.Errorf("converting object suffix to int: %v", err)
+	}
+
+	return nameRegex.ReplaceAllString(name, fmt.Sprintf("${1}%d", n+1)), nil
```
g-gaston, line 25: if we change the regex, we can extract the name as the group 1 and just do `ObjectName(match[1], n+1)`

g-gaston (approved):

PR #2002, opened 2022-05-04
Add Kubectl Wrappers for Curated Packages

a-cool-train:
*Issue #, if available:*

*Description of changes:*

This PR includes the following the functionalities:
### Deleting Curated Packages
 Using Kubectl, delete all curated packages provided by the user through eksctl anywhere delete packages.
### Describe Curated Packages
Using Kubectl, describe all curated packages requested by the user.
### Upgrading the Active Package Bundle in a bundle controller
Find the active bundle controller and update the active bundle used. 
### Creating Curated Packages
Using Kubectl, create curated packages from file(s) provided by the user.
### Applying Curated Packages
Using kubectl, apply curated packages from file(s) provided by the user.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/applypackages.go
+++ b/cmd/eksctl-anywhere/cmd/applypackages.go
@@ -1,0 +1,59 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+)
+
+type applyPackageOptions struct {
+	fileName string
+}
+
+var apo = &applyPackageOptions{}
+
+func init() {
+	applyCmd.AddCommand(applyPackagesCommand)
+	applyPackagesCommand.Flags().StringVarP(&apo.fileName, "filename", "f", "", "File with curated packages to apply")
+	err := applyPackagesCommand.MarkFlagRequired("filename")
+	if err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var applyPackagesCommand = &cobra.Command{
+	Use:          "packages",
+	Short:        "Apply curated packages",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := applyPackages(cmd.Context()); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+func applyPackages(ctx context.Context) error {
+	kubeConfig := kubeconfig.FromEnvironment()
+	deps, err := newDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+	packages := curatedpackages.NewPackageClient(
+		nil,
+		deps.Kubectl,
+	)
+
+	err = packages.ApplyPackages(ctx, apo.fileName, kubeConfig)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/cmd/eksctl-anywhere/cmd/applypackages.go
+++ b/cmd/eksctl-anywhere/cmd/applypackages.go
@@ -19,7 +19,7 @@ var apo = &applyPackageOptions{}
 
 func init() {
 	applyCmd.AddCommand(applyPackagesCommand)
-	applyPackagesCommand.Flags().StringVarP(&apo.fileName, "filename", "f", "", "File with curated packages to apply")
+	applyPackagesCommand.Flags().StringVarP(&apo.fileName, "filename", "f", "", "Filename that contains curated packages custom resources to apply")
 	err := applyPackagesCommand.MarkFlagRequired("filename")
 	if err != nil {
 		log.Fatalf("Error marking flag as required: %v", err)
@@ -27,8 +27,10 @@ func init() {
 }
 
 var applyPackagesCommand = &cobra.Command{
-	Use:          "packages",
+	Use:          "package(s) [flags]",
 	Short:        "Apply curated packages",
+	Long:         "Apply Curated Packages Custom Resources to the cluster",
+	Aliases:      []string{"package", "packages"},
 	PreRunE:      preRunPackages,
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/eksctl-anywhere/cmd/deletepackages.go
+++ b/cmd/eksctl-anywhere/cmd/deletepackages.go
@@ -1,0 +1,46 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+)
+
+func init() {
+	deleteCmd.AddCommand(deletePackageCommand)
+}
+
+var deletePackageCommand = &cobra.Command{
+	Use:          "package(s) [flags]",
+	Aliases:      []string{"package", "packages"},
+	Short:        "Delete package(s)",
+	Long:         "This command is used to delete the curated packages installed in the cluster",
+	PreRunE:      preRunPackages,
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return deleteResources(cmd.Context(), args)
+	},
+}
+
+func deleteResources(ctx context.Context, args []string) error {
+	kubeConfig := kubeconfig.FromEnvironment()
+	deps, err := newDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		return fmt.Errorf("unable to initialize executables: %v", err)
+	}
+	packages := curatedpackages.NewPackageClient(
+		nil,
+		deps.Kubectl,
+	)
+
+	err = packages.DeletePackages(ctx, args, kubeConfig)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/cmd/eksctl-anywhere/cmd/deletepackages.go
+++ b/cmd/eksctl-anywhere/cmd/deletepackages.go
@@ -18,7 +18,7 @@ var deletePackageCommand = &cobra.Command{
 	Use:          "package(s) [flags]",
 	Aliases:      []string{"package", "packages"},
 	Short:        "Delete package(s)",
-	Long:         "This command is used to delete the curated packages installed in the cluster",
+	Long:         "This command is used to delete the curated packages custom resources installed in the cluster",
 	PreRunE:      preRunPackages,
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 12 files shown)


pkg/curatedpackages/packageclient.go
```diff
@@ -116,7 +117,56 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
...
+	stdOut, err := pc.kubectl.ExecuteCommand(ctx, params...)
+	if err != nil {
+		fmt.Print(&stdOut)
+		return err
+	}
+	fmt.Print(&stdOut)
+	return nil
+}
+
+func (pc *PackageClient) DeletePackages(ctx context.Context, args []string, kubeConfig string) error {
```
g-gaston, line 146: can we make this method signature a bit more explicit? I would suggest splitting `args` into multiple method arguments with explicit names.
Right now, this is creating a coupling from the cmd command to the kubectl command passing through this method

pkg/curatedpackages/packageclient.go
```diff
@@ -116,7 +117,56 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
...
+	stdOut, err := pc.kubectl.ExecuteCommand(ctx, params...)
+	if err != nil {
+		fmt.Print(&stdOut)
+		return err
+	}
+	fmt.Print(&stdOut)
+	return nil
+}
+
+func (pc *PackageClient) DescribePackages(ctx context.Context, args []string, kubeConfig string) error {
```
g-gaston, line 158: Same comment here for args

pkg/curatedpackages/packageclient.go
```diff
@@ -116,7 +117,56 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
...
+	params := []string{"describe", "packages", "--kubeconfig", kubeConfig, "--namespace", constants.EksaPackagesName}
+	params = append(params, args...)
+	stdOut, err := pc.kubectl.ExecuteCommand(ctx, params...)
+	if err != nil {
+		fmt.Print(&stdOut)
+		return fmt.Errorf("kubectl execution failure: \n%v", err)
+	}
+	if len(stdOut.Bytes()) == 0 {
+		return errors.New("no resources found")
```
g-gaston, line 167: Test for this case?

pkg/curatedpackages/bundle.go
```diff
@@ -103,3 +105,17 @@ func (b *BundleReader) getActiveController(ctx context.Context) (*packagesv1.Pac
...
 }
+
+func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1.PackageBundleController, newBundle string) error {
+	controller.Spec.ActiveBundle = newBundle
+	controllerYaml, err := yaml.Marshal(controller)
+	if err != nil {
+		return err
+	}
+	params := []string{"create", "-f", "-", "--kubeconfig", b.kubeConfig}
+	_, err = b.kubectl.CreateFromYaml(ctx, controllerYaml, params...)
```
g-gaston, line 116: can we have something like `CreateFromYaml(ctx context.Context, yaml []byte, kubeconfig string) (bytes.Buffer, error)`?

pkg/curatedpackages/packageclient_test.go
```diff
@@ -103,3 +104,90 @@ func TestInstallPackagesFails(t *testing.T) {
 	err := tt.command.InstallPackage(tt.ctx, &tt.bundle.Spec.Packages[0], "my-harbor", "")
 	tt.Expect(err).To(MatchError(ContainSubstring("error installing package. Package exists")))
 }
+
+func TestApplyPackagesPass(t *testing.T) {
+	tt := newPackageTest(t)
+	tt.kubectl.EXPECT().ExecuteCommand(tt.ctx, gomock.Any()).Return(convertJsonToBytes(tt.bundle.Spec.Packages[0]), nil)
```
g-gaston, line 110: Can we avoid using `gomock.Any()`? This is making the test a bit less useful since we are not really testing that we are running the right command (which is the only thing this method does).

Same for the other tests.

cmd/eksctl-anywhere/cmd/applypackages.go
```diff
@@ -0,0 +1,61 @@
...
+func init() {
+	applyCmd.AddCommand(applyPackagesCommand)
+	applyPackagesCommand.Flags().StringVarP(&apo.fileName, "filename", "f", "", "Filename that contains curated packages custom resources to apply")
+	err := applyPackagesCommand.MarkFlagRequired("filename")
+	if err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var applyPackagesCommand = &cobra.Command{
```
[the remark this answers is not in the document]
TerryHowe: I thought the apply was a generate plus apply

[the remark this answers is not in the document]
a-cool-train: install package is an imperative command and apply is used for the declarative approach. To use apply, you will be doing running a generate command then run an apply command on the generated file.

go.mod
```diff
@@ -157,7 +157,7 @@ require (
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
 	oras.land/oras-go v1.1.0 // indirect
 	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
```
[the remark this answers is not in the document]
a-cool-train: I am not sure what happened here. Let me check rq.

pkg/curatedpackages/packageclient.go
a-cool-train: args is not a method argument but a customer input. I could name it in a better way but it is supposed to represent the packages the customer needs. I'll update the name.

pkg/curatedpackages/bundle.go
a-cool-train: Any reason for this? The current approach provides flexibility to pass in any arguments.

pkg/curatedpackages/packageclient_test.go
a-cool-train: Fixed it.

TerryHowe (approved):
/lgtm

if there is anything important that needs to be resolved here, can we get a follow up PR?

pkg/curatedpackages/packageclient.go
```diff
@@ -116,7 +117,56 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
 	if err != nil {
 		return err
 	}
-	fmt.Println(&stdOut)
+	fmt.Print(&stdOut)
+	return nil
+}
+
+func (pc *PackageClient) ApplyPackages(ctx context.Context, fileName string, kubeConfig string) error {
+	params := []string{"apply", "-f", fileName, "--kubeconfig", kubeConfig}
```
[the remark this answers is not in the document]
ewollesen: I might be completely wrong... but if you guys are referring to `ApplyKubeSpec` (pkg/executables/kubectl.go:189), then it seems like it could easily be modified to handle a nil cluster. 

I tend to agree that the ApplyPackages method feels out of place. It should be at least calling the existing Apply, and not sidestepping right to ExecuteCommand.

For that matter, it seems a little strange that in the exsting ApplyKubeSpec, the `Kubectl` object needs to know how to reach into a `Cluster` object, when it could just accept a `KubeconfigFile` (string), but that's a refactor for another day.

cmd/eksctl-anywhere/cmd/deletepackages.go
```diff
@@ -0,0 +1,46 @@
...
+		deps.Kubectl,
+	)
+
+	err = packages.DeletePackages(ctx, args, kubeConfig)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
ewollesen, line 46:
```suggestion
var deletePackageCommand = &cobra.Command{
	Use:          "package(s) [flags]",
	Aliases:      []string{"package", "packages"},
	Short:        "Delete package(s)",
	Long:         "This command is used to delete the curated packages custom resources installed in the cluster",
	PreRunE:      preRunPackages,
	SilenceUsage: true,
	RunE: func(cmd *cobra.Command, args []string) error {
		return deletePackages(cmd.Context(), args...)
	},
}

func deletePackages(ctx context.Context, packageNames string...) error {
	kubeConfig := kubeconfig.FromEnvironment()
	deps, err := newDependenciesForPackages(ctx, kubeConfig)
	if err != nil {
		return fmt.Errorf("unable to initialize executables: %v", err)
	}
	packages := curatedpackages.NewPackageClient(
		nil,
		deps.Kubectl,
	)

	err = packages.DeletePackages(ctx, kubeConfig, packageNames...)
	if err != nil {
		return err
	}

	return nil
}
```

pkg/curatedpackages/packageclient.go
```diff
@@ -116,7 +117,56 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
...
+	params := []string{"delete", "packages", "--kubeconfig", kubeConfig, "--namespace", constants.EksaPackagesName}
+	params = append(params, packages...)
+	stdOut, err := pc.kubectl.ExecuteCommand(ctx, params...)
+	if err != nil {
+		fmt.Print(&stdOut)
+		return err
+	}
+	fmt.Print(&stdOut)
+	return nil
+}
```
ewollesen, line 156:
```suggestion
func (pc *PackageClient) DeletePackages(ctx context.Context, kubeConfig string, packageNames string...) error {
	params := []string{"delete", "packages", "--kubeconfig", kubeConfig, "--namespace", constants.EksaPackagesName}
	params = append(params, packageNames...)
	stdOut, err := pc.kubectl.ExecuteCommand(ctx, params...)
	if err != nil {
		fmt.Print(&stdOut)
		return err
	}
	fmt.Print(&stdOut)
	return nil
}
```
These change are to match the changes suggested in a previous commit.

But using a splat at the end, you buy yourself extra flexibility in how you call the methods.

ewollesen (commented):
`<redacted>`

cmd/eksctl-anywhere/cmd/applypackages.go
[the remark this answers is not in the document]
g-gaston: Could `apply` then just be an option to `install`, like `install packages -f`?

pkg/curatedpackages/packageclient.go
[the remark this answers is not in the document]
g-gaston: Agreed with @ewollesen, `ApplyKubeSpec` should just take a `kubeconfig` string. To avoid adding that refactor to this task, you could just create a `types.Cluster` object with just the kubeconfig in it.

pkg/curatedpackages/packageclient.go
g-gaston: if they are package names wanted by the user, can it be just be called `packageNames`?

pkg/curatedpackages/bundle.go
g-gaston: Because the method is called `CreateFromYaml`. It doesn't make a lot of sense that we need to pass "create" as part of the params. The method already knows it needs to run a create command.

PR #2004, opened 2022-05-04
PR process and commit message guidelines

g-gaston:

### DIFF
--- a/docs/developer/pull-requests.md
+++ b/docs/developer/pull-requests.md
@@ -3,6 +3,142 @@ This document outlines best practices to follow when creating and reviewing PRs.
 
 ## As an author
 
+### Pull request process
+1. Create a GitHub PR from your fork
+1. Verify all checks are passing
+1. Get someone to review it
+1. Iterate your changes based on the feedback you receive. Don't rewrite the history (rebase or squash) with your new changes, keep them as individual commits. This facilitates the reviewers job.
+1. Once you haver al least one `lgtm` and are ready to merge, squash your commits into one.
+    * Even if you leave multiple commits, they will all be squashed by GitHub before merging. Prefer to do it manually to control the format.
+    * Avoid any message derived from the review process itself: "Fix typo", "Address comment", "Implement feedback", etc. This provides absolutely no context and clutters the commit history.
+    * Treat this commit as if you were writing the PR for the first time but make sure you include any new relevant information derived from the conversations you had during the review.
+    * Follow the [commit messages](#commit-messages) guidelines
+1. Get the final `/approve` to start the merge process.
+### Commit messages
+
+Follow these guidelines:
+
+* [Capitalize the subject line](#capitalize-the-subject-line)
+
+* [Use the imperative mood in the subject line](#use-the-imperative-mood-in-the-subject-line)
+
+* [Limit the subject line to 50 characters](#limit-the-subject-line-to-50-characters)
+
+* [Do not end the subject line with a period](#do-not-end-the-subject-line-with-a-period)
+
+* [Separate subject from body with a blank line](#separate-subject-from-body-with-a-blank-line)
+
+* [Use the body to explain what and why](#use-the-body-to-explain-what-and-why)
+
+* [Wrap the body at 72 characters](#wrap-the-body-at-72-characters)
+
+
+Glossary:
+* **Commit subject line**: first line of the commit message
+* **Commit body**: everything after the first blank line.
+
+This is a great example from [Chris Beams](https://cbea.ms/git-commit/):
+
+```
+Summarize changes in around 50 characters or less
+
+More detailed explanatory text, if necessary. Wrap it to about 72
+characters or so. In some contexts, the first line is treated as the
+subject of the commit and the rest of the text as the body. The
+blank line separating the summary from the body is critical (unless
+you omit the body entirely); various tools like `log`, `shortlog`
+and `rebase` can get confused if you run the two together.
+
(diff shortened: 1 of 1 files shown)


docs/developer/pull-requests.md
```diff
@@ -3,6 +3,142 @@ This document outlines best practices to follow when creating and reviewing PRs.
 
 ## As an author
 
+### Pull request process
+1. Create a GitHub PR from your fork
+1. Verify all checks are passing
+1. Get someone to review it
+1. Iterate your changes based on the feedback you receive. Don't rewrite the history (rebase or squash) with your new changes, keep them as individual commits. This facilitates the reviewers job.
+1. Once you haver al least one `lgtm` and are ready to merge, squash your commits into one.
```
chrisdoherty4, line 11: Imposing a 1 commit per PR requirement isn't good. I, for example, manually structure commits and put a lot of thought into them. That effort is wasted and even problematic when I want to go back and perhaps revert a particular change I made.

docs/developer/pull-requests.md
```diff
@@ -3,6 +3,142 @@ This document outlines best practices to follow when creating and reviewing PRs.
 
 ## As an author
 
+### Pull request process
+1. Create a GitHub PR from your fork
+1. Verify all checks are passing
+1. Get someone to review it
+1. Iterate your changes based on the feedback you receive. Don't rewrite the history (rebase or squash) with your new changes, keep them as individual commits. This facilitates the reviewers job.
```
chrisdoherty4, line 10: I'm OK with advocating not squashing updates until the end but is it really necessary to mandate no re-writing. I feel like that judgement should be left to developers.

To provide an example: if I re-write a function based on feedback why shouldn't I squash that, its a new version implemented differently.

This comes back to the auto removal of LGTM too. If we're auto removing LGTM we're forcing developers to do squashing via the UI else the LGTM is removed - that's not good.

docs/developer/pull-requests.md
g-gaston: These changes are not introducing that imposition though, we have prow a GitHub already configured that way. So unless I'm mistaken, even if you don't currently squash your commits manually, they will be automatically squashed before merging (I think I mention this a bit bellow). My main goal was to get the best commit history possible under the current constrains.

That said, if you want to reopen that debate (which includes merge commits, etc.), that's cool with me 🙂

chrisdoherty4: Resolved offline.

docs/developer/pull-requests.md
chrisdoherty4: Resolved offline.

chrisdoherty4 (commented):
/lgtm

docs/developer/pull-requests.md
TerryHowe, line 11:
```suggestion
1. Once you haver al least one `lgtm` and are ready to merge, squash your commits into one.
```
```suggestion
1. Once you have al least one `lgtm` and are ready to merge, squash your commits into one.
```
squash on merge no?

TerryHowe, line 11:
```suggestion
1. Once you have al least one `lgtm` and are ready to merge, squash your commits into one.
```

TerryHowe: auto squash on merge, no?

docs/developer/pull-requests.md
```diff
@@ -3,6 +3,142 @@ This document outlines best practices to follow when creating and reviewing PRs.
...
+
+#### Use the imperative mood in the subject line
+
+Example
+
+* ~~`Fixed error in code`~~
+* ~~`Some code fixes for errors`~~
+* `Fix error in code`
+
+A good rule of thumb from Chris Beams on writing a subject is it should complete this sentence:
```
TerryHowe, line 81:
```suggestion
A good rule is from Chris Beams on writing a subject is it should complete this sentence:
```
Disturbing connotations of that expression

docs/developer/pull-requests.md
TerryHowe: I think we can avoid the squash completely and just use auto squash if we have a good rule on the commit the makes the PR subject because I think the PR subject gets used for the squash, no? I'd have to try it out to be sure.

TerryHowe: Any other options to configure how auto squash works. Nobody has time to squash commits as the final step, I just don't think that is reasonable and while that commit history is not great I just don't think it is worth the extra manual effort to squash.

chrisdoherty4: I, strongly, disagree. I squash manually all the time. `git rebase -i HEAD~<commits to squash>` followed by some vim magic `%s/pick/f/gc` using `gc` options so I can confirm things. 

It literally takes <30s.

chrisdoherty4: That said, this workflow doesn't work with the existing setup because the minute you squash and force push `lgtm` is removed and you're left to bug people for lgtm again. Thats a bad flow.

g-gaston: If losing the `lgtm` label is the only concern, we could use [StoreTreeHash](https://pkg.go.dev/k8s.io/test-infra/prow/plugins#Lgtm) in the prow config. That should prevent the `lgtm` label from being removed if we are only squashing

chrisdoherty4: Lets do it and see where it goes.

chrisdoherty4 (approved):

PR #2005, opened 2022-05-04
Set up E2E tests for `FluxConfig` testing

danbudris:
*Issue #, if available:*

*Description of changes:*
re-configure GithubProviderConfig and GitProviderConfig options to take their own set of options -- so that we never end up with a nil provider in the FluxConfig and have the freedom to set it up however we like.

make sure to marshal the FluxConfig to the cluster spec we write to disk for executing the tests.

convert cluster `GitOpsConfig` to a `FluxConfig` during the Flux validation and initialization, and only validate against the `FluxConfig`.

set up the first working GitHub test using the new FluxConfig.

*Testing (if applicable):*

ran test locally using this pattern and confirmed the behavior


test/framework/flux.go
```diff
@@ -68,10 +69,6 @@ func WithFluxGit(opts ...api.FluxConfigOpt) ClusterE2ETestOpt {
 		for _, opt := range opts {
 			opt(e.FluxConfig)
 		}
-		// Adding Job ID suffix to repo name
-		// e2e test jobs have Job Id with a ":", replacing with "-"
-		jobId := strings.Replace(e.getJobIdFromEnv(), ":", "-", -1)
-		withFluxRepositorySuffix(jobId)(e.FluxConfig)
```
taneyland, line 74: No need for this anymore? I think we added this to avoid `repo already exists` errors in the remote repository

taneyland (approved):

test/framework/flux.go
danbudris: so this is just removing it from the FluxGit setup -- with the generic Git provider, we're requiring that a user provide an existing Git repository, and won't try to create or delete it, so no need to clean it up or make the repository name unique; making it unique in this case will actually cause the test to fail since the repo won't exist. We rely on the uniqueness of the cluster name here

PR #2011, opened 2022-05-04
Pass management cluster kubeconfig to GenerateCAPISpecForCreate

jiayiwang7:
*Issue #, if available:*

Fix the bug in `provider.GenerateCAPISpecForCreate` by passing in the management cluster. So that the kubeclient operations will be applied on management cluster instead of workload. 

*Description of changes:*

This change only affects snow provider since other providers do not use the arg `cluster *types.Cluster`

*Testing (if applicable):*

unit tests


g-gaston (commented):
Can we update the unit tests to prevent a regression on this? I suspect we are using the same name in the management cluster we pass as an arg and the new workload cluster.

PR #2018, opened 2022-05-05
Feature controller installation

a-cool-train:
*Description of changes:*

### Install Curated Packages During Cluster Creation
- During cluster creation, pass --install-packages to provide the location of curated packages configuration files to use. This typically would occur after eksctl anywhere generate command has been executed. There are two operations that occur as part of this: install controller, and install packages. Both operations occur after bootstrap cluster has been deleted. The failure of either of the above operation doesn't lead to a failure in the cluster creation hence there are some checks to make sure that doesn't occur. 

### Install Curated Packages Through CLI Command 
- In an existing cluster, run eksctl anywhere install packagecontroller --kube-version 1.21 to install curated packages controller. This can be followed with eksctl anywhere create command to create the packages generated. This command checks to make sure the curated packages controller hadn't already been created in the cluster. If it already exists, it fails with an error message.

*Testing (if applicable):*

Closes: https://github.com/aws/eks-anywhere-packages/issues/108


pkg/workflows/create.go
```diff
@@ -359,9 +371,49 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
+
+func (s *InstallPackageControllerTask) Name() string {
+	return "install-package-controller"
+}
+
+func installCuratedPackages(ctx context.Context, clusterName string, packagesLocation string) {
+	kubeConfig := kubeconfig.FromClusterName(clusterName)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		logger.MarkFail("Error when installing curated packages on workload cluster; please install through eksctl anywhere create command", "error", err)
```
chrisdoherty4, line 407: This is called during create so I'm not sure the log message is particularly helpful. Ditto for the one below.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -0,0 +1,80 @@
...
+	installCmd.AddCommand(installPackageControllerCommand)
+	installPackageControllerCommand.Flags().StringVar(&ico.kubeVersion, "kube-version", "", "Bundle version to use")
+	if err := installPackageControllerCommand.MarkFlagRequired("kube-version"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var installPackageControllerCommand = &cobra.Command{
+	Use:          "packagecontroller",
+	Aliases:      []string{"packagecontroller"},
```
chrisdoherty4, line 32: Isn't this superfluous?

pkg/workflows/create.go
```diff
@@ -359,9 +371,49 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
 func (s *DeleteBootstrapClusterTask) Name() string {
 	return "delete-kind-cluster"
 }
+
+func (s *InstallPackageControllerTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Installing curated packages controller on workload  cluster")
+	kubeConfig := kubeconfig.FromClusterName(commandContext.ClusterSpec.Cluster.Name)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		logger.MarkFail("Error when installing curated packages on workload cluster; please install through eksctl anywhere install command", "error", err)
```
chrisdoherty4, line 386: Same as the other comment on logging. Seems like this is called during `install packagecontroller` cmd so it seems unhelpful. Can you clarify?

pkg/workflows/create.go
```diff
@@ -60,7 +62,15 @@ func (c *Create) Run(ctx context.Context, clusterSpec *cluster.Spec, validator i
 		commandContext.BootstrapCluster = clusterSpec.ManagementCluster
 	}
 
-	return task.NewTaskRunner(&SetAndValidateTask{}).RunTask(ctx, commandContext)
+	err := task.NewTaskRunner(&SetAndValidateTask{}).RunTask(ctx, commandContext)
+	if err != nil {
+		return err
+	}
+
+	if packagesLocation != "" {
```
chrisdoherty4, line 70: This seems like it should be its _own_ task, not part of a create task.

pkg/workflows/create.go
a-cool-train: There are two different commands in play, maybe I should clarify that with the message but the idea is that if this fails during cluster creation (the current command in execution), you have the ability to create it through create -f command.

pkg/workflows/create.go
a-cool-train: Yes, this is to account for failure of curated packages controller during cluster creation. There is a separate command you could use to run this which is namely the install. I'll add better details.

pkg/workflows/create.go
```diff
@@ -359,9 +371,49 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
-	return nil
+	return &InstallPackageControllerTask{}
 }
 
 func (s *DeleteBootstrapClusterTask) Name() string {
 	return "delete-kind-cluster"
 }
+
+func (s *InstallPackageControllerTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Installing curated packages controller on workload  cluster")
```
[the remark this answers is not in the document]
a-cool-train: Where?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: What would be a more concise manner you would think?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -0,0 +1,80 @@
...
+
+type installControllerOptions struct {
+	kubeVersion string
+}
+
+var ico = &installControllerOptions{}
+
+func init() {
+	installCmd.AddCommand(installPackageControllerCommand)
+	installPackageControllerCommand.Flags().StringVar(&ico.kubeVersion, "kube-version", "", "Bundle version to use")
```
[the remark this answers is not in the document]
a-cool-train: Yes that's the intent. All curated packages commands will not be using the cluster.spec, they will either use the current kubecontext or pass in the kube-version.

pkg/workflows/create.go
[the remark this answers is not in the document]
a-cool-train: ignore this comment. Fixed it.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
chrisdoherty4: Its not about being concise, its the fact the alias matches its use. Alias' are for _alternate_ ways of calling the command.

From [the docs](https://pkg.go.dev/github.com/spf13/cobra#Command):

> Aliases is an array of aliases that can be used instead of the first word in Use.

pkg/workflows/create.go
chrisdoherty4: Thanks.

pkg/workflows/create.go
chrisdoherty4: OK, this stuff is going to be overhauled anyway so I'll make a note. Thanks.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: Gotcha. Thanks

pkg/workflows/create.go
chrisdoherty4: Sorry, I'm a little confused. The `create` cmd requires the `-f` flag so, if I'm not mistaken, they would've done that already and then received this log message that tells them to do the same thing. I feel like I'm missing something.

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -0,0 +1,71 @@
...
+		chartInstaller: ci,
+		command:        curatedpackages.NewPackageControllerClient(ci, k, kubeConfig, uri, chartName, chartVersion),
+		kubeConfig:     kubeConfig,
+		ociUri:         uri,
+		chartName:      chartName,
+		chartVersion:   chartVersion,
+	}
+}
+
+func TestInstallControllerSuccess(t *testing.T) {
```
chrisdoherty4, line 50: Good to see the happy path. Can we invoke a non-happy path too to model that behavior. I.e. what happens when the controller fails to install?

chrisdoherty4:
This looks great. I think there are 3 changes/additions:

- negative case unit testing.
- removal of the `Alias`.
- clarification in the logging.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := installPackageControllerCommand.MarkFlagRequired("kube-version"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
+}
+
+var installPackageControllerCommand = &cobra.Command{
+	Use:          "packagecontroller",
+	Aliases:      []string{"packagecontroller"},
+	Short:        "Install packagecontroller to the cluster",
```
TerryHowe, line 32:
```suggestion
	Short:        "Install packagecontroller on the cluster",
```

pkg/workflows/create.go
```diff
@@ -359,9 +371,49 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		logger.MarkFail("Error when installing curated packages on workload cluster; please install through eksctl anywhere install packagecontroller command", "error", err)
+		return nil
+	}
+	chart := commandContext.ClusterSpec.VersionsBundle.VersionsBundle.PackageController.HelmChart
+	pc := curatedpackages.NewPackageControllerClient(deps.Helm, deps.Kubectl, kubeConfig, chart.Image(), chart.Name, chart.Tag())
+	err = pc.InstallController(ctx)
+	if err != nil {
+		logger.MarkFail("Error when installing curated packages controller on workload cluster; please install through eksctl anywhere install packagecontroller command", "error", err)
```
TerryHowe: Is there going to be a command to uninstall the packagecontroller?

pkg/workflows/create.go
```diff
@@ -359,9 +371,49 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
+
+func (s *InstallPackageControllerTask) Name() string {
+	return "install-package-controller"
+}
+
+func installCuratedPackages(ctx context.Context, clusterName string, packagesLocation string) {
+	kubeConfig := kubeconfig.FromClusterName(clusterName)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		logger.MarkFail("Error when installing curated packages on workload cluster; please install through eksctl anywhere create packages command", "error", err)
```
TerryHowe: All these messages are exactly the same, not sure if there is enough context in the error, so it might be handy to have a little difference in each case

a-cool-train: Fixed the repetition issue.

pkg/workflows/create.go
a-cool-train: Currently that's not an expected behavior. We can discuss whether we want to add that or not.

pkg/workflows/create.go
a-cool-train: As discussed offline, these are two commands.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -0,0 +1,48 @@
...
+		kubectl:        kubectl,
+	}
+}
+
+func (pc *PackageControllerClient) InstallController(ctx context.Context) error {
+	uri := fmt.Sprintf("%s%s", "oci://", pc.ociUri)
+	return pc.chartInstaller.InstallChartFromName(ctx, uri, pc.kubeConfig, pc.chartName, pc.chartVersion)
+}
+
+func (pc *PackageControllerClient) ValidateControllerExists(ctx context.Context) error {
```
vivek-koppuru, line 41:
```suggestion
func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Context) error {
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -0,0 +1,48 @@
...
+	params := []string{"get", "packageBundleController", "--kubeconfig", pc.kubeConfig, "--namespace", constants.EksaPackagesName, bundle.PackageBundleControllerName}
+	stdOut, _ := pc.kubectl.ExecuteCommand(ctx, params...)
+	if len(stdOut.Bytes()) != 0 {
```
vivek-koppuru, line 44: What if `kubectl` returns a warning message or some other output that may not be the controller itself? Is this reliable?

pkg/workflows/create.go
```diff
@@ -365,3 +375,50 @@ func (s *DeleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
...
+	kubeConfig := kubeconfig.FromClusterName(clusterName)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
+	if err != nil {
+		return err
+	}
+	packageClient := curatedpackages.NewPackageClient(
+		nil,
+		deps.Kubectl,
+	)
+	err = packageClient.CreatePackages(ctx, packagesLocation, kubeConfig)
```
vivek-koppuru, line 419: We should already be validating whether the  `packagesLocation` is a dir or not by now. Apart from that, this method contains the logic to ensure the input is validated?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -0,0 +1,48 @@
...
+		ociUri:         ociUri,
+		chartName:      chartName,
+		chartVersion:   chartVersion,
+		chartInstaller: chartInstaller,
+		kubectl:        kubectl,
+	}
+}
+
+func (pc *PackageControllerClient) InstallController(ctx context.Context) error {
+	uri := fmt.Sprintf("%s%s", "oci://", pc.ociUri)
```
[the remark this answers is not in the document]
TerryHowe: could just call it registry instead I guess

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -0,0 +1,48 @@
...
+	return pc.chartInstaller.InstallChartFromName(ctx, uri, pc.kubeConfig, pc.chartName, pc.chartVersion)
+}
+
+func (pc *PackageControllerClient) ValidateControllerExists(ctx context.Context) error {
+	params := []string{"get", "packageBundleController", "--kubeconfig", pc.kubeConfig, "--namespace", constants.EksaPackagesName, bundle.PackageBundleControllerName}
+	stdOut, _ := pc.kubectl.ExecuteCommand(ctx, params...)
+	if len(stdOut.Bytes()) != 0 {
+		return errors.New("curated Packages Controller Exists in the current Cluster")
```
chrisdoherty4, line 45: Lower case errors.

pkg/curatedpackages/packagecontrollerclient.go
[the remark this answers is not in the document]
a-cool-train: Uri might make sense I think.

pkg/workflows/create.go
a-cool-train: This method forwards all that is needed to kubectl. Kubectl will fail during execution in case the path doesn't exist.

chrisdoherty4 (approved):

a-cool-train:
Bypassing codecov since create.go file is hard to test and will be refactored in upcoming PRs.

TerryHowe (approved):
/lgtm

PR #2019, opened 2022-05-05
Include Support Bundle

a-cool-train:
*Description of changes:*
### Include Support Bundle
- This PR adds eksa-packages namespace to the support-bundle.

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/diagnostic_bundle.go
+++ b/pkg/diagnostics/diagnostic_bundle.go
@@ -112,6 +112,7 @@ func newDiagnosticBundleFromSpec(af AnalyzerFactory, cf CollectorFactory, spec *
 		WithManagementCluster(spec.Cluster.IsSelfManaged()).
 		WithDefaultAnalyzers().
 		WithDefaultCollectors().
+		WithPackagesCollectors().
 		WithLogTextAnalyzers()
 
 	err := b.WriteBundleConfig()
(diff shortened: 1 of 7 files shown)


pkg/diagnostics/diagnostic_bundle.go
```diff
@@ -112,6 +112,7 @@ func newDiagnosticBundleFromSpec(af AnalyzerFactory, cf CollectorFactory, spec *
 		WithManagementCluster(spec.Cluster.IsSelfManaged()).
 		WithDefaultAnalyzers().
 		WithDefaultCollectors().
+		WithPackagesCollectors().
```
jaxesn, line 115: It probably doesnt matter, but just to confirm, is it an issue if we try to collect these and the controller was never installed? i assume troubleshoot will just skip over things that arent there?

a-cool-train: That's a very good point. Let me try that out.

TerryHowe (approved):

pkg/diagnostics/diagnostic_bundle.go
a-cool-train: Just verified. Nothing happens when the controller is not installed. The other support items are created.

a-cool-train:
Bypassing coverage since no unit tests currently exist for the analyzers.

PR #2021, opened 2022-05-05
Feature download images

a-cool-train:
*Description of changes:*
### Include Download Images for Curated packages components

Curated packages has the following artifacts to be extracted:
- Package Controller image
- Package Bundle Manifest
- Helm Charts for packages

In order to extract these details, these PR introduces two files: curatedpackages.reader which is responsible for extracting the package controller image and the helm charts for packages, and oras.download file which is responsible for the package bundle manifest and saving to disk. 


cmd/eksctl-anywhere/cmd/downloadimages.go
```diff
@@ -99,3 +106,17 @@ func packagerForFile(file string) packager {
...
+		return curatedpackages.NewPackageReader(reader)
+	}
+	return reader
+}
+
+func fetchManifestDownloader(downloadFolder string, includePackages bool) artifacts.ManifestDownloader {
+	if includePackages {
+		return oras.NewBundleDownloader(downloadFolder)
+	}
+	return &artifacts.Noop{}
```
mrajashree, line 121: can this return nil instead of defining Noop struct and the [Noop SaveManifests](https://github.com/aws/eks-anywhere/pull/2021/files#diff-54b0899b811165e615e0f50eb79e464c7848b322bf4f7b55af21b2c1004498dfR49) on it?

pkg/curatedpackages/bundle.go
```diff
@@ -119,3 +120,17 @@ func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1
...
 	return nil
 }
+
+func GetPackageBundleRef(vb releasev1.VersionsBundle) string {
+	packageController := vb.PackageController
+	// Use package controller registry to fetch packageBundles.
+	// Format of controller image is: <uri>/<env_type>/<repository_name>
+	controllerImage := strings.Split(packageController.Controller.Image(), "/")
+	major, minor, err := parseKubeVersion(vb.KubeVersion)
+	if err != nil {
```
mrajashree, line 130: can we log the error here if we're not returning it?

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
+package curatedpackages
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	eksdv1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
+	"sigs.k8s.io/yaml"
+
```
mrajashree, line 10: nit: the imports aren't ordered

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		images = append(images, packages...)
+	}
+	return images
+}
+
+func FetchPackages(versionsBundle releasev1.VersionsBundle, ctx context.Context, art string) ([]releasev1.Image, error) {
+	data, err := Pull(ctx, art)
+	ctrl := versionsBundle.PackageController.Controller
+	if err != nil {
```
mrajashree, line 70: might want to move this up before `ctrl := versionsBundle.PackageController.Controller`

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
...
+func (r *PackageReader) ReadBundlesForVersion(version string) (*releasev1.Bundles, error) {
+	return r.ManifestReader.ReadBundlesForVersion(version)
+}
+
+func (r *PackageReader) ReadEKSD(eksaVersion, kubeVersion string) (*eksdv1.Release, error) {
+	return r.ManifestReader.ReadEKSD(eksaVersion, kubeVersion)
+}
+
+func (r *PackageReader) ReadImages(eksaVersion string) ([]releasev1.Image, error) {
+	images, err := r.ManifestReader.ReadImages(eksaVersion)
```
mrajashree, line 36: nit
```suggestion
	return r.ManifestReader.ReadImages(eksaVersion)
```

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
...
+func (r *PackageReader) ReadImagesFromBundles(b *releasev1.Bundles) ([]releasev1.Image, error) {
+	images, err := r.ManifestReader.ReadImagesFromBundles(b)
+	for _, v := range b.Spec.VersionsBundles {
+		images = append(images, v.PackageControllerImage()...)
+	}
+	return images, err
+}
+
+func (r *PackageReader) ReadCharts(eksaVersion string) ([]releasev1.Image, error) {
+	images, err := r.ManifestReader.ReadCharts(eksaVersion)
```
mrajashree, line 49: nit
```suggestion
	return r.ManifestReader.ReadCharts(eksaVersion)
```

pkg/oras/download.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			fmt.Printf("unable to write to file %v", err)
+		}
+	}
+}
+
+func UniqueCharts(charts []string) []string {
+	c := types.SliceToLookup(charts).ToSlice()
+	// TODO: maybe optimize this, avoiding the sort and just following the same order as the original slice
+	sort.Strings(c)
```
mrajashree, line 45: why does this need to be sorted?

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
...
+	return images, err
+}
+
+func (r *PackageReader) ReadChartsFromBundles(ctx context.Context, b *releasev1.Bundles) []releasev1.Image {
+	images := r.ManifestReader.ReadChartsFromBundles(ctx, b)
+	for _, vb := range b.Spec.VersionsBundles {
+		artifact := GetPackageBundleRef(vb)
+		packages, err := FetchPackages(vb, ctx, artifact)
+		if err != nil {
+			fmt.Printf("error finding packages: %v", err)
```
TerryHowe, line 59: no logger here? also might be super handy to print the artifact

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,104 @@
...
+			continue
+		}
+		images = append(images, packages...)
+	}
+	return images
+}
+
```
TerryHowe, line 66: bit odd that ctx isn't first here

TerryHowe: I also wonder why FetchPackages and Pull are public

pkg/oras/download.go
```diff
@@ -0,0 +1,63 @@
...
+	return c
+}
+
+func writeToFile(dir string, packageName string, content []byte) error {
+	file := filepath.Join(dir, packageName) + ".yaml"
+	if err := os.WriteFile(file, content, 0o644); err != nil {
```
TerryHowe, line 51:
```suggestion
	if err := os.WriteFile(file, content, 0o640); err != nil {
```
Not sure what the "standard" is, but I see no need for global read

cmd/eksctl-anywhere/cmd/downloadimages.go
a-cool-train: It can. Any specific reason?

pkg/curatedpackages/reader.go
a-cool-train: Looks like the linter prefers this way.

pkg/curatedpackages/reader.go
a-cool-train: Fixed all three.

pkg/oras/download.go
a-cool-train: Doesn't need. Uniqueness is only required. Updated with unique fetch.

TerryHowe (approved):
/lgtm

cmd/eksctl-anywhere/cmd/downloadimages.go
g-gaston: I personally prefer the noop vs nil
That way `artifacts.Download` doesn't need to worry about the manifest downloader being conditional

A better pattern could be to have a slice of "downloaders" and just include the ones that are necessary. But that requires a bigger refactor so I would vote to leave it for later.

cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
```diff
@@ -24,6 +24,10 @@ type ChartDownloader interface {
 	Download(ctx context.Context, artifacts ...string) error
 }
 
+type ManifestDownloader interface {
+	SaveManifests(ctx context.Context, bundles *releasev1.Bundles)
```
g-gaston, line 28: nit:
```suggestion
	Download(ctx context.Context, bundles *releasev1.Bundles)
```

cmd/eksctl-anywhere/cmd/internal/commands/artifacts/download.go
```diff
@@ -37,8 +41,13 @@ type Download struct {
 	Packager                 Packager
 	TmpDowloadFolder         string
 	DstFile                  string
+	ManifestDownloader       ManifestDownloader
 }
 
+type Noop struct{}
```
g-gaston, line 47: nit: can we move this to a separate file?

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,87 @@
...
+	ManifestReader *manifests.Reader
+}
+
+func NewPackageReader(mr *manifests.Reader) *PackageReader {
+	return &PackageReader{
+		ManifestReader: mr,
+	}
+}
+
+func (r *PackageReader) ReadBundlesForVersion(version string) (*releasev1.Bundles, error) {
```
g-gaston, line 26: I would suggest defining PackageReader like

```go
type PackageReader struct {
	*manifests.Reader
}
```
That way, we don't need to define all these wrappers and `PackageReader` will get all those methods "for free"

In fact we should probably define an interface for the ManifestReader in this package, but we can leave that for when we add the unit tests.

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (r *PackageReader) ReadImages(eksaVersion string) ([]releasev1.Image, error) {
+	return r.ManifestReader.ReadImages(eksaVersion)
+}
+
+func (r *PackageReader) ReadImagesFromBundles(b *releasev1.Bundles) ([]releasev1.Image, error) {
+	images, err := r.ManifestReader.ReadImagesFromBundles(b)
+	for _, v := range b.Spec.VersionsBundles {
+		images = append(images, v.PackageControllerImage()...)
```
g-gaston, line 41: Is this the same image as https://github.com/aws/eks-anywhere/blob/main/release/api/v1alpha1/artifacts.go#L143

In that case I would probably remove it from `SharedImages()`

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ctrl := versionsBundle.PackageController.Controller
+	bundle := &packagesv1.PackageBundle{}
+	err = yaml.Unmarshal(data, bundle)
+	if err != nil {
+		return nil, err
+	}
+	var images []releasev1.Image
```
g-gaston, line 75: nit:
```suggestion
	images := make([]releasev1.Image, 0, len(bundle.Spec.Packages)
```

pkg/manifests/reader.go
```diff
@@ -92,6 +93,6 @@ func (r *Reader) ReadCharts(eksaVersion string) ([]releasev1.Image, error) {
 	return bundles.Charts(bundle), nil
 }
 
-func (r *Reader) ReadChartsFromBundles(b *releasev1.Bundles) []releasev1.Image {
+func (r *Reader) ReadChartsFromBundles(ctx context.Context, b *releasev1.Bundles) []releasev1.Image {
```
g-gaston, line 96: nit: let's create another method that doesn't need the context and leave this just to implement the interface from download/import

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			logger.MarkFail("error finding packages for artifact", artifact, "error", err)
+			continue
+		}
+		images = append(images, packages...)
+	}
+	return images
+}
+
+func fetchPackages(ctx context.Context, versionsBundle releasev1.VersionsBundle, art string) ([]releasev1.Image, error) {
```
g-gaston, line 64:
```suggestion
func fetchPackages(ctx context.Context, versionsBundle releasev1.VersionsBundle, artifact string) ([]releasev1.Image, error) {
```

pkg/curatedpackages/reader.go
```diff
@@ -0,0 +1,87 @@
...
+	return r.ManifestReader.ReadCharts(eksaVersion)
+}
+
+func (r *PackageReader) ReadChartsFromBundles(ctx context.Context, b *releasev1.Bundles) []releasev1.Image {
+	images := r.ManifestReader.ReadChartsFromBundles(ctx, b)
+	for _, vb := range b.Spec.VersionsBundles {
+		artifact := GetPackageBundleRef(vb)
+		packages, err := fetchPackages(ctx, vb, artifact)
+		if err != nil {
+			logger.MarkFail("error finding packages for artifact", artifact, "error", err)
```
g-gaston, line 56: shouldn't this return an error?
At the end if the user wants all the charts exported and we are not able to pull one, shouldn't we error out?

pkg/oras/download.go
```diff
@@ -0,0 +1,70 @@
+package oras
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
```
g-gaston, line 10: Is there a way we can avoid this dependency?
`oras` is a "generic" package (that can be used for storage operations in oci registries)

If this was intended to be curated packages exclusive, can we move it to `pkg/curatedpackages/oras`? Or even `pkg/curatedpackages/bundles` or something like that (since it's no actually doing anything oras specific

pkg/curatedpackages/curatedpackages.go
```diff
@@ -99,3 +100,17 @@ func PrintLicense() {
 	fmt.Println(LICENSE)
 	fmt.Println(strings.Repeat("-", WIDTH))
 }
+
+func Pull(ctx context.Context, art string) ([]byte, error) {
+	puller := artifacts.NewRegistryPuller()
```
g-gaston, line 105: this needs to be injected as a dependency

pkg/curatedpackages/bundle.go
```diff
@@ -119,3 +121,18 @@ func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1
...
+		logger.MarkFail("unable to parse kubeversion", "error", err)
```
g-gaston, line 132: curious, why not return an error?

pkg/curatedpackages/bundle.go
```diff
@@ -119,3 +121,18 @@ func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1
 	}
 	return nil
 }
+
+func GetPackageBundleRef(vb releasev1.VersionsBundle) string {
```
g-gaston, line 125: Just for my own understanding, what is this supposed to do?
Maybe an example of in->out will help me

a-cool-train: We currently don't include packagebundle manifests in eks-anywhere release to avoid tight coupling between eks-anywhere and packagebundles. What this is doing is it is getting the curated packages controller (which is in the release manifest) uri (also contains the repository), isolates the registry location and appends the eks-anywhere-package-bundles repository to it. Does this help?

pkg/curatedpackages/reader.go
a-cool-train: Ideally not. We might not have packagebundle manifest for every kubeversion that is currently supported. We only fetch those we can get.

pkg/curatedpackages/reader.go
a-cool-train: Discussed offline, second part will be taken care of in followup pr.

pkg/curatedpackages/reader.go
a-cool-train: Discussed offline, we will keep as is but revisit the error handling before GA.

pkg/manifests/reader.go
a-cool-train: Discussed offline, this will be included in a follow up PR.

g-gaston (approved):

PR #2023, opened 2022-05-05
Add the packages controller to the build artifacts.

ewollesen:
The intent here is to make sure that artifacts for curated packages (namely our controller image and our helm chart) get included in builds.

/cc @jonahjon  @ivyostosh


abhay-krishna (approved):
/lgtm

PR #2028, opened 2022-05-05
Add more validations for FluxConfig

taneyland:
*Issue #, if available:*

*Description of changes:*
We needed a way to access the FluxConfig spec for validations. We already have a `ParseClusterConfigFromContent` function that accepts any `KindAccessor` type (such as Cluster, FluxConfig, etc.) to parse those contents. So I renamed that function to `ParseConfigFromContent` since it isn't exclusive to the cluster config, and I am using it to pass in the FluxConfig type and parse those contents so that we can have access to that spec is the gitops validations function.

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/url"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/yaml"
 
+	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
@@ -168,7 +170,6 @@ var clusterConfigValidations = []func(*Cluster) error{
 	validateControlPlaneReplicas,
 	validateWorkerNodeGroups,
 	validateNetworking,
-	validateGitOps,
 	validateEtcdReplicas,
 	validateIdentityProviderRefs,
 	validateProxyConfig,
@@ -194,11 +195,23 @@ func GetClusterConfig(fileName string) (*Cluster, error) {
 // GetClusterConfigFromContent parses a Cluster object from a multiobject yaml content
 func GetClusterConfigFromContent(content []byte) (*Cluster, error) {
 	clusterConfig := &Cluster{}
-	err := ParseClusterConfigFromContent(content, clusterConfig)
+	err := ParseConfigFromContent(content, clusterConfig)
 	if err != nil {
 		return clusterConfig, err
 	}
 	return clusterConfig, nil
+}
+
+// GetFluxConfig parses a Cluster object from a multiobject yaml file in disk
+// and sets defaults if necessary
+func GetFluxConfig(fileName string) (*FluxConfig, error) {
+	fluxConfig := &FluxConfig{}
+	err := ParseClusterConfig(fileName, fluxConfig)
+	if err != nil {
+		return fluxConfig, err
+	}
+
+	return fluxConfig, nil
 }
 
 // GetClusterConfig parses a Cluster object from a multiobject yaml file in disk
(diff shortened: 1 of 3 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -213,6 +226,19 @@ func GetAndValidateClusterConfig(fileName string) (*Cluster, error) {
 		return nil, err
 	}
 
+	if clusterConfig.Spec.GitOpsRef != nil {
```
jiayiwang7, line 229: you dont validate the fluxConfig here, instead u can do it similar to https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/gitops.go

`GetAndValidateClusterConfig` only validate cluster object itself (including the refs)

pkg/api/v1alpha1/cluster.go
```diff
@@ -582,9 +608,22 @@ func validateGitOps(clusterConfig *Cluster) error {
 	gitOpsRefKind := gitOpsRef.Kind
 	fluxConfigActive := features.IsActive(features.GenericGitProviderSupport())
 
-	if gitOpsRefKind == FluxConfigKind && !fluxConfigActive {
-		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported; " +
-			"to use this experimental feature, please set the environment variable GENERIC_GIT_PROVIDER_SUPPORT to true")
+	if gitOpsRefKind == FluxConfigKind {
```
jiayiwang7, line 611: here too, validateGitOps is to validate cluster obj's gitops config, not the fluxconfig itself

jiayiwang7 (commented):
can we also add unit tests?

pkg/api/v1alpha1/cluster.go
```diff
@@ -194,13 +195,25 @@ func GetClusterConfig(fileName string) (*Cluster, error) {
...
 		return clusterConfig, err
 	}
 	return clusterConfig, nil
 }
 
+// GetFluxConfig parses a Cluster object from a multiobject yaml file in disk
+// and sets defaults if necessary
+func GetFluxConfig(fileName string) (*FluxConfig, error) {
+	fluxConfig := &FluxConfig{}
+	err := ParseClusterConfig(fileName, fluxConfig)
```
vivek-koppuru, line 209: Should we change this name too then?

pkg/api/v1alpha1/cluster.go
vivek-koppuru: Agreed, here we only validate if the ref has all the values or something from the perspective of the cluster object.

pkg/api/v1alpha1/cluster.go
```diff
@@ -582,9 +608,22 @@ func validateGitOps(clusterConfig *Cluster) error {
 	gitOpsRefKind := gitOpsRef.Kind
 	fluxConfigActive := features.IsActive(features.GenericGitProviderSupport())
 
-	if gitOpsRefKind == FluxConfigKind && !fluxConfigActive {
-		return fmt.Errorf("FluxConfig and the generic git provider are not currently supported; " +
-			"to use this experimental feature, please set the environment variable GENERIC_GIT_PROVIDER_SUPPORT to true")
+	if gitOpsRefKind == FluxConfigKind {
+		if !fluxConfigActive {
+			return fmt.Errorf("FluxConfig and the generic git provider are not currently supported; " +
```
vivek-koppuru, line 613: Are we sticking to the name being "generic git provider" in our documentation too? I like just calling this "git provider" tbh

pkg/api/v1alpha1/cluster.go
```diff
@@ -582,9 +608,22 @@ func validateGitOps(clusterConfig *Cluster) error {
...
+				"to use this experimental feature, please set the environment variable GENERIC_GIT_PROVIDER_SUPPORT to true")
+		}
+		if fluxConfig.Spec.Git != nil {
+			if fluxConfig.Spec.Github != nil {
+				return errors.New("only one provider can be specified in FluxConfig. Please only specify either git or github")
+			}
+			if privateKeyFile, ok := os.LookupEnv(config.EksaGitPrivateKeyTokenEnv); !ok || len(privateKeyFile) <= 0 {
```
vivek-koppuru, line 620: Do we need any other validations on the values of these env vars?

danbudris: ideally we should do a check on the permissions of the private key file, it should be set at 600

pkg/api/v1alpha1/cluster.go
taneyland: For sure. Once #2034 merges, I'll refactor this PR to add the validations in the proper place

danbudris (approved):

PR #2029, opened 2022-05-05
Run `make release-manifest` to generate fluxconfig CRD

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


abhay-krishna:
what happened to other presubmits?

abhay-krishna:
> what happened to other presubmits?

Never mind, my bad, I thought we were running those on all changes (`.*`). But I think we should run atleast the e2e one on config folder changes because it could catch unmarshalling issues when the eksa-components.yaml manifest is applied.

PR #2040, opened 2022-05-05
General curated package troubleshooting

TerryHowe:
Product has requested that we have more general troubleshooting and this pull request is to cover that.

I'll follow up with a PR that is more like our current troubleshooting that has log messages and possible solutions.

### DIFF
--- a/docs/content/en/docs/tasks/troubleshoot/_packages.md
+++ b/docs/content/en/docs/tasks/troubleshoot/_packages.md
@@ -1,0 +1,47 @@
+
+---
+title: "Curating Packages Troubleshooting"
+linkTitle: "Curating Packages Troubleshooting"
+weight: 50
+---
+
+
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/tasks/troubleshoot/_packages.md
```diff
@@ -0,0 +1,47 @@
+
+---
+title: "Curating Packages Troubleshooting"
+linkTitle: "Curating Packages Troubleshooting"
+weight: 50
+---
+
+
+The major component of Curated Packages is the package controller. If the container is not running or not running correctly, packages will not be installed. Generally it should be debugged like any other Kubernetes application. The first step is to check that then pod is running.
```
ewollesen, line 9:
```suggestion
The major component of Curated Packages is the package controller. If the container is not running or not running correctly, packages will not be installed. Generally it should be debugged like any other Kubernetes application. The first step is to check that the pod is running.
```

ewollesen (changes requested):

ewollesen (approved):

PR #2045, opened 2022-05-06
Curated packages bugfixes

a-cool-train:
*Description of changes:*

### Add License for Curated Packages
Provides the capability to print a license during the following scenarios:
- When installing curated packages as part of cluster creation
- When installing curated packages controller through install packagecontroller
- When creating curated packages through create, apply, and install commands

### Minor Bugfixes 
- Fixes bug with upgrade to use apply instead of create
- Fixes bug for get commands to print a new line carriage
- Fixes bug for kubeconfig outside of current directory

### Move Dependency creation to curatedpackages
- Unify curatedpackages dependencies creation. 

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createpackages.go
+++ b/cmd/eksctl-anywhere/cmd/createpackages.go
@@ -43,7 +43,7 @@ var createPackagesCommand = &cobra.Command{
 
 func createPackages(ctx context.Context) error {
 	kubeConfig := kubeconfig.FromEnvironment()
-	deps, err := newDependenciesForPackages(ctx, kubeConfig)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
 	if err != nil {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
--- a/cmd/eksctl-anywhere/cmd/get.go
+++ b/cmd/eksctl-anywhere/cmd/get.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 )
@@ -39,7 +40,7 @@ func preRunPackages(cmd *cobra.Command, args []string) error {
 func getResources(ctx context.Context, resourceType string, output string, args []string) error {
 	kubeConfig := kubeconfig.FromEnvironment()
 
-	deps, err := newDependenciesForPackages(ctx)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
 	if err != nil {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
@@ -56,7 +57,7 @@ func getResources(ctx context.Context, resourceType string, output string, args 
 		return fmt.Errorf("kubectl execution failure: \n%v", err)
 	}
 	if len(stdOut.Bytes()) == 0 {
-		fmt.Printf("No resources found in %v namespace", constants.EksaPackagesName)
+		fmt.Printf("No resources found in %v namespace\n", constants.EksaPackagesName)
 		return nil
 	}
 	fmt.Println(&stdOut)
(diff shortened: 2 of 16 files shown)


cmd/eksctl-anywhere/cmd/get.go
```diff
@@ -56,9 +57,9 @@ func getResources(ctx context.Context, resourceType string, output string, args
 		return fmt.Errorf("kubectl execution failure: \n%v", err)
 	}
 	if len(stdOut.Bytes()) == 0 {
-		fmt.Printf("No resources found in %v namespace", constants.EksaPackagesName)
+		fmt.Printf("No resources found in %v namespace\n", constants.EksaPackagesName)
```
vivek-koppuru, line 60: Shouldn't this make use of logger instead of a printf?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -66,6 +66,8 @@ func installPackageController(ctx context.Context) error {
 		helmChart.Tag(),
 	)
 
+	fmt.Println("tag: " + helmChart.Tag() + " Image: " + helmChart.Image())
```
vivek-koppuru, line 69: Don't want to use logger here? I understand if we are trying to relay the output from kubectl but that's not what we are doing here.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -75,3 +84,10 @@ func NewDependenciesForPackages(ctx context.Context, paths ...string) (*dependen
 		WithHelm().
 		Build(ctx)
 }
+
+func PrintLicense() {
+	// Currently, use the width of the longest line to repeat the dashes
+	fmt.Println(strings.Repeat("-", WIDTH))
```
vivek-koppuru, line 90: Can you comment here an example of the output from this?

pkg/workflows/create.go
```diff
@@ -68,6 +68,7 @@ func (c *Create) Run(ctx context.Context, clusterSpec *cluster.Spec, validator i
 	}
 
 	if packagesLocation != "" {
+		curatedpackages.PrintLicense()
```
vivek-koppuru, line 71: What is the purpose of printing out the license all the time?

cmd/eksctl-anywhere/cmd/get.go
a-cool-train: Actually this needs to be printed to the user.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: Whoops debug point lol. Taking it out.

pkg/workflows/create.go
a-cool-train: It is a business requirement to print out the license.

cmd/eksctl-anywhere/cmd/createpackages.go
```diff
@@ -43,7 +43,7 @@ var createPackagesCommand = &cobra.Command{
 
 func createPackages(ctx context.Context) error {
 	kubeConfig := kubeconfig.FromEnvironment()
-	deps, err := newDependenciesForPackages(ctx, kubeConfig)
+	deps, err := curatedpackages.NewDependenciesForPackages(ctx, kubeConfig)
```
g-gaston, line 46: If this method is using the factory, I would advice against putting it in the curatedpackages package

The dependency direction should be "ports" (like cmd or controllers) -> factory -> "packages with business logic"

If we break that, we will eventually find a cyclic dependency

pkg/curatedpackages/bundle.go
```diff
@@ -112,7 +112,7 @@ func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1
 	if err != nil {
 		return err
 	}
-	params := []string{"create", "-f", "-", "--kubeconfig", b.kubeConfig}
+	params := []string{"apply", "-f", "-", "--kubeconfig", b.kubeConfig}
 	_, err = b.kubectl.CreateFromYaml(ctx, controllerYaml, params...)
```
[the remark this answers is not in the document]
g-gaston: Addressed this offline: creating an issue to handle this separability

pkg/curatedpackages/packageclient.go
```diff
@@ -106,6 +106,7 @@ func (pc *PackageClient) packageMap() map[string]packagesv1.BundlePackage {
 }
 
 func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.BundlePackage, customName string, kubeConfig string) error {
+	PrintLicense()
```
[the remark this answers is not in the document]
g-gaston: Addressed this offline: creating an issue to handle this separability

cmd/eksctl-anywhere/cmd/createpackages.go
g-gaston: Addressed this offline: creating an issue to handle this separability

PR #2046, opened 2022-05-06
Remove packages image and chart from manifest list

g-gaston:
*Description of changes:*
We expect all the artifacts URIs to be http/s urls to publicly downloadable
files. These two are OCI artifacts.
*Testing (if applicable):*

### DIFF
--- a/release/api/v1alpha1/artifacts.go
+++ b/release/api/v1alpha1/artifacts.go
@@ -48,10 +48,6 @@ func (vb *VersionsBundle) Manifests() map[string][]*string {
 		"cilium": {
 			&vb.Cilium.Manifest.URI,
 		},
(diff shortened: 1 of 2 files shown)


release/api/v1alpha1/artifacts.go
```diff
@@ -48,10 +48,6 @@ func (vb *VersionsBundle) Manifests() map[string][]*string {
 		"cilium": {
 			&vb.Cilium.Manifest.URI,
 		},
-		"packages": {
```
a-cool-train, line 51: @ewollesen might have better context for adding this.

TerryHowe: It could be the build is broken, but we do want to do this

Issue #1888, opened 2022-04-20
CloudStack upgrade e2e test is failing with timeout on move


maxdrib:
**What happened**:
Error message displayed below:

```
Error: failed to upgrade cluster: moving CAPI management from source to target: failed moving management cluster: Performing move...
Discovering Cluster API objects
Moving Cluster API objects Clusters=1
Moving Cluster API objects ClusterClasses=0
Creating objects in the target cluster
Error: action failed after 10 attempts: error creating "cluster.x-k8s.io/v1beta1, Kind=Cluster" eksa-system/eksa-test-2ffeb29: Internal error occurred: failed calling webhook "default.cluster.cluster.x-k8s.io": Post "[https://capi-webhook-service.capi-system.svc:443/mutate-cluster-x-k8s-io-v1beta1-cluster?timeout=10s](https://capi-webhook-service.capi-system.svc/mutate-cluster-x-k8s-io-v1beta1-cluster?timeout=10s)": context deadline exceeded
```

**What you expected to happen**:
Upgrade should succeed

**How to reproduce it (as minimally and precisely as possible)**:
Execute the TestCloudStackKubernetes120RedhatTo121Upgrade e2e test again the code in main

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

[cluster.yaml.zip](https://github.com/aws/eks-anywhere/files/8523795/cluster.yaml.zip)

maxdrib:
I was able to successfully perform an upgrade when using an older cluster spec. Comparing the cluster specs, I noticed a difference in the cluster.spec.networkConfig attributes where the working spec had:

```
  clusterNetwork:
    cni: cilium
    pods: ...
    services: ...
```

and the broken spec had

```
  clusterNetwork:
    cniConfig:
      cilium: {}
    dns: {}
    pods: ...
    services: ...
```

I continued to attempt to isolate the problem and was able to reproduce the issue by adding/removing the `dns: {}` attribute in the cluster spec. When present, it caused the cluster to hang on upgrade. Unfortunately, each test run takes ~45 minutes so it is tedious to test changes.

[...]

PR #2057, opened 2022-05-06
Restarting cilium daemonset on upgrade workflow to address upgrade issue

maxdrib:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1888

*Description of changes:*
This PR restarts the cilium pods in a CloudStack upgrade workflow in order to prevent the issues observed in https://github.com/aws/eks-anywhere/issues/1888. We are still waiting to hear what the root cause may be from Isovalent but expect this change to resolve the majority of the issues.

Will create issue to make the logic non CloudStack specific

*Testing (if applicable):*
Unit tests. Executed the TestCloudStackKubernetes120RedhatTo121Upgrade locally several times and observed the networking issues recover during the upgrade

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -87,6 +87,7 @@ type ClusterClient interface {
 	GetEksaVSphereMachineConfig(ctx context.Context, VSphereDatacenterName string, kubeconfigFile string, namespace string) (*v1alpha1.VSphereMachineConfig, error)
 	GetEksaCloudStackMachineConfig(ctx context.Context, cloudstackMachineConfigName string, kubeconfigFile string, namespace string) (*v1alpha1.CloudStackMachineConfig, error)
 	SetEksaControllerEnvVar(ctx context.Context, envVar, envVarVal, kubeconfig string) error
+	RestartCiliumDaemonset(ctx context.Context, kubeconfig string) error
 	CreateNamespace(ctx context.Context, kubeconfig string, namespace string) error
 	GetNamespace(ctx context.Context, kubeconfig string, namespace string) error
 	ValidateControlPlaneNodes(ctx context.Context, cluster *types.Cluster, clusterName string) error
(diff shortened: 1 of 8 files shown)


pkg/executables/kubectl.go
```diff
@@ -690,6 +690,18 @@ func (k *Kubectl) ValidateEKSAClustersCRD(ctx context.Context, cluster *types.Cl
 	return nil
 }
 
+func (k *Kubectl) RestartCiliumDaemonset(ctx context.Context, kubeconfig string) error {
```
vivek-koppuru, line 693: This should be a `DaemonSetRolloutRestart call` instead of a reference to Cilium in kubectl

pkg/clustermanager/cluster_manager.go
```diff
@@ -395,6 +396,14 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
 		return fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
 	}
 
+	if provider.Name() == constants.CloudStackProviderName {
```
vivek-koppuru, line 399: Can we introduce a new method in the provider or add it to an existing one, or is that not possible?

vivek-koppuru: Can we not add this logic to here? https://github.com/aws/eks-anywhere/blob/main/pkg/clustermanager/cluster_manager.go#L376

maxdrib: See https://github.com/aws/eks-anywhere/issues/2061 for context about the long-term home for this operation and why we can't use the existing RunPostControlPlaneUpgrade method

pkg/executables/kubectl.go
maxdrib: Agreed, will pass in the daemonset name instead of hardcoding it here

pkg/executables/kubectl_test.go
```diff
@@ -1017,7 +1017,39 @@ func TestKubectlSetControllerEnvVarSuccess(t *testing.T) {
 
 	err := k.SetEksaControllerEnvVar(ctx, envVar, envVarValue, cluster.KubeconfigFile)
 	if err != nil {
-		t.Fatalf("Kubectl.GetApiServerUrl() error = %v, want nil", err)
+		t.Fatalf("Kubectl.DaemonSetRolloutRestart() error = %v, want nil", err)
+	}
+}
+
+func TestKubectlRetartCiliumDaemonsetSuccess(t *testing.T) {
```
vivek-koppuru, line 1024: We should change the unit tests to not include cilium also right?

vivek-koppuru (approved):

PR #2116, opened 2022-05-11
Adding production if/else check to use ECR public from staging for packages helm chart.

jonahjon:

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


release/pkg/generate_spec.go
```diff
@@ -492,6 +492,23 @@ func (r *ReleaseConfig) GetSourceImageURI(name, repoName string, tagOptions map[
 
 func (r *ReleaseConfig) GetSourceHelmURI(repoName string) (string, error) {
 	var sourceImageUri string
+	// If it's running a production release we only lookup from ECR public of the staging account.
+	if r.ReleaseEnvironment == "production" {
+		ecrpublicClient, err := NewECRPublicClient()
```
abhay-krishna, line 497: We probably don't need to create new clients. The release config struct has source clients and release clients configured in it, for ecr, ecr public and s3

jonahjon: This is using the SDK v2 so we might need to do so.

release/pkg/assets_package_controller.go
```diff
@@ -44,7 +44,14 @@ func (r *ReleaseConfig) GetPackagesAssets() ([]Artifact, error) {
 	if err != nil {
 		return nil, errors.Cause(err)
 	}
-	sourceHelmURI, err := r.GetSourceHelmURI(repoName)
+	var sourceHelmURI string
```
ewollesen, line 47: Is the value returned by `GetSourceHelmURI` incorrect for a production case? Is this the correct place to handle this branching? Or should the environment be passed to `GetShourceHelmURI`?

jonahjon: The production role doesn't have access to call `GetSourceHelmURI`, and it's not needed so we want to skip that all together in that scenario.

jonahjon: sounds good, fixed this to the way you mentioned.

release/pkg/generate_spec.go
```diff
@@ -490,22 +490,29 @@ func (r *ReleaseConfig) GetSourceImageURI(name, repoName string, tagOptions map[
...
-func (r *ReleaseConfig) GetSourceHelmURI(repoName string) (string, error) {
-	var sourceImageUri string
-	ecrClient, err := NewECRClient()
-	if err != nil {
-		return "", err
-	}
-	latestTag, err := ecrClient.GetLatestUploadHelmSha(repoName)
-	if err != nil {
-		return "", err
+//GetSourceHelmURI is a function which finds the latest helm chart tag for a given repository, and returns the full URI including that tag.
```
ewollesen, line 493:
```suggestion
// GetSourceHelmURI returns the URI for the latest tagged helm chart in the given repository.
```

ewollesen (approved):
/lgtm

PR #2149, opened 2022-05-12
[Harbor] add storageClass config options for PVCs

vincentni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


docs/content/en/docs/reference/packagespec/harbor/v2.5.0.md
```diff
@@ -42,6 +42,10 @@ The following table lists the configurable parameters of the Harbor package spec
...
 | `persistence.persistentVolumeClaim.registry.size`                    | The size of the volume                                                                                                                                                                                                                                                                                                                                                                                                                                            | `5Gi`                                 |
+| `persistence.persistentVolumeClaim.registry.storageClass`            | Specify the `storageClass` used to provision the volume. Or the default StorageClass will be used (the default). Set it to `-` to disable dynamic provisioning                                                                                                                                                                                                                                                                                                    |                                       |
```
ewollesen, line 45: Did you mean to actually fill in the defaults here (and several spots below)?

ewollesen, line 45:
```suggestion
| `persistence.persistentVolumeClaim.registry.storageClass`            | Specify the `storageClass` used to provision the volume. Defaults to X. Dynamic provision can be disabled by setting it to `-`.                                                                                                                                                                                                |                                       |
```

ewollesen (commented):
Really just wondering if you meant to fill in default values.

It would be worth taking a spin through and making sure you're using complete sentences (that end with periods).

ewollesen (approved):
Vincent tells me that these are just copied from the upstream. I guess it's an issue for another day what we do when the upstream docs aren't as detailed as we might like or have stylistic differences.

/lgtm

PR #2174, opened 2022-05-14
Deprecate Cilium field in bundle manifest

abhay-krishna:
Deprecate Cilium fields from bundle. The field will still be a part of the Bundle struct for backward compatibility, but the field will be empty in the generated bundle manifest for releases.


release/api/v1alpha1/artifacts.go
```diff
@@ -163,7 +159,6 @@ func (vb *VersionsBundle) Images() []Image {
 
 func (vb *VersionsBundle) Charts() map[string]*Image {
 	return map[string]*Image{
-		"cilium":                &vb.Cilium.HelmChart,
```
jaxesn, line 166: We still need the helm chart and image tag and what not, but not the manifests

Issue #2061, opened 2022-05-06
Rehome the Cilium Restart operation in cluster manager


maxdrib:
**What would you like to be added**:
The cilium restart operation introduced in https://github.com/aws/eks-anywhere/pull/2057 is currently only invoked for the CloudStack provider. There may be value in using it for Snow as well, since network connectivity has been reported there as well. It cannot go in the existing provider method RunPostControlPlaneUpgrade since that runs before the `waitForControlPlaneReady` method, which is where the network connectivity actually breaks.

Something to consider is we don't want to couple ourselves too closely with cilium so customers can choose their own CNI, so it may be best to add this as a new method in the provider interface instead of calling it for all providers

**Why is this needed**:
Cluster manager should not have provider specific logic. We should either restart cilium for all providers, or factor it out to a new provider interface method.

PR #2182, opened 2022-05-16
Rollout restart cilium during upgrade

jiayiwang7:
*Issue #, if available:*

#2061 

*Description of changes:*

Follow up to #2057. We run the rollout restart cilium daemonset operation on the target cluster, across all providers during cluster upgrade if using cilium as CNI.

I add a `RunPostControlPlaneUpgradeSetup(ctx context.Context, cluster *types.Cluster) error` method in the networking interface in `cluster_manager` pkg. So that we can define CNI provider specific tasks that need to run after control plane vm is updated.

*Testing (if applicable):*

- unit tests


maxdrib (approved):
TestCloudStackKubernetes120RedhatTo121Upgrade successfully recovered from the network issue using this code

PR #2201, opened 2022-05-18
Adding provider specific collectors/analyzers to management cluster support bundle

maxdrib:
*Issue #, if available:*

*Description of changes:*
I was trying to debug a failure in the eks-a cloudstack e2e tests when I noticed that the bootstrap support bundle doesn't actually have logs for the capc controller, so I am blocked to continue debugging the e2e test failures.

This PR plumbs the spec through to the bundle generation logic so we can grab the provider specific logs and crds even on the bootstrap/management cluster

*Testing (if applicable):*
Tested by setting control plane ready timeout to 1 minute and observed capc logs were captured.

### DIFF
--- a/pkg/diagnostics/analyzers.go
+++ b/pkg/diagnostics/analyzers.go
@@ -82,6 +82,10 @@ func (a *analyzerFactory) managementClusterDeploymentAnalyzers() []*Analyze {
 			Namespace:        constants.CapvSystemNamespace,
 			ExpectedReplicas: 1,
 		}, {
+			Name:             "capc-controller-manager",
+			Namespace:        constants.CapcSystemNamespace,
+			ExpectedReplicas: 1,
+		}, {
 			Name:             "cert-manager-webhook",
 			Namespace:        constants.CertManagerNamespace,
 			ExpectedReplicas: 1,
--- a/pkg/diagnostics/diagnostic_bundle.go
+++ b/pkg/diagnostics/diagnostic_bundle.go
@@ -46,7 +46,7 @@ type EksaDiagnosticBundle struct {
 	analysis         []*executables.SupportBundleAnalysis
 }
 
-func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactory, client BundleClient,
+func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactory, spec *cluster.Spec, client BundleClient,
 	kubectl *executables.Kubectl, kubeconfig string, writer filewriter.FileWriter,
 ) (*EksaDiagnosticBundle, error) {
 	b := &EksaDiagnosticBundle{
@@ -69,7 +69,7 @@ func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactor
 		writer:           writer,
 	}
 
-	b.WithDefaultCollectors().WithDefaultAnalyzers().WithManagementCluster(true)
+	b.WithDefaultCollectors().WithDefaultAnalyzers().WithManagementCluster(true).WithDatacenterConfig(spec.Cluster.Spec.DatacenterRef)
 
 	err := b.WriteBundleConfig()
 	if err != nil {
--- a/pkg/diagnostics/factory.go
+++ b/pkg/diagnostics/factory.go
@@ -37,17 +37,17 @@ func NewFactory(opts EksaDiagnosticBundleFactoryOpts) *eksaDiagnosticBundleFacto
 
 func (f *eksaDiagnosticBundleFactory) DiagnosticBundle(spec *cluster.Spec, provider providers.Provider, kubeconfig string, bundlePath string) (DiagnosticBundle, error) {
 	if bundlePath == "" && spec != nil {
-		b, err := f.DiagnosticBundleFromSpec(spec, provider, kubeconfig)
+		b, err := f.DiagnosticBundleWorkloadCluster(spec, provider, kubeconfig)
 		return b, err
 	}
 	return f.DiagnosticBundleCustom(kubeconfig, bundlePath), nil
 }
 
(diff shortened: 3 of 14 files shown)


pkg/diagnostics/analyzers.go
```diff
@@ -81,6 +81,10 @@ func (a *analyzerFactory) managementClusterDeploymentAnalyzers() []*Analyze {
 			Name:             "capv-controller-manager",
 			Namespace:        constants.CapvSystemNamespace,
 			ExpectedReplicas: 1,
+		}, {
```
jaxesn, line 84: I dont think this will add the logs, this i believe, is just to validate the pod is running if the user runs the analyze subcommand

danbudris: no, but https://github.com/aws/eks-anywhere/pull/2201/files#diff-d32ef215cc246d7530a6ff1a2722c31aef67351081e996a8ab33545deded959cR72 should add the logs; when the cloudstack datacenter kind is passed in, it'll return the relevant collectors, which include collectors for the whole namespace. 

to extend the collectors associated with a cloudstack data center kind, check out: https://github.com/aws/eks-anywhere/blob/2b8e7f172428db6a328066d5ea07f5e544d2d6ad/pkg/diagnostics/collectors.go#L87

pkg/diagnostics/diagnostic_bundle.go
```diff
@@ -69,7 +69,7 @@ func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactor
 		writer:           writer,
 	}
 
-	b.WithDefaultCollectors().WithDefaultAnalyzers().WithManagementCluster(true)
+	b.WithDefaultCollectors().WithDefaultAnalyzers().WithManagementCluster(true).WithDatacenterConfig(spec.Cluster.Spec.DatacenterRef)
```
danbudris, line 72: the one issue I foresee here (an edge case) is that if we collect and analyze from a bootstrap cluster which the CAPI components have been moved off of, the `WithDataCenterConfig` analyzers will still expect them to be present and alert on it. don't have a solution for that at the moment as the diagnostic bundle has no awareness of the 'phase' a cluster is in, but wanted to call it out for further thought

danbudris (approved):
thanks, we've needed this for a while 👍

pkg/diagnostics/diagnostic_bundle.go
maxdrib: Agreed. It seems like to have a comprehensive set of artifacts for troubleshooting, we need to either be surgical and phase-aware when we're collecting them, or just do a full sweet of all the possible artifacts and expect some to be missing. I feel like the latter is more likely to get us the artifacts we need without coupling this code with the main eks-a logic

pkg/diagnostics/factory.go
```diff
@@ -37,17 +37,17 @@ func NewFactory(opts EksaDiagnosticBundleFactoryOpts) *eksaDiagnosticBundleFacto
...
-		b, err := f.DiagnosticBundleFromSpec(spec, provider, kubeconfig)
+		b, err := f.DiagnosticBundleWorkloadCluster(spec, provider, kubeconfig)
 		return b, err
 	}
 	return f.DiagnosticBundleCustom(kubeconfig, bundlePath), nil
 }
 
-func (f *eksaDiagnosticBundleFactory) DiagnosticBundleManagementCluster(kubeconfig string) (DiagnosticBundle, error) {
-	return newDiagnosticBundleManagementCluster(f.analyzerFactory, f.collectorFactory, f.client, f.kubectl, kubeconfig, f.writer)
+func (f *eksaDiagnosticBundleFactory) DiagnosticBundleManagementCluster(spec *cluster.Spec, kubeconfig string) (DiagnosticBundle, error) {
```
jiayiwang7, line 46: If I understand this correctly, we might want to rename it to `DiagnosticBundleBootstrapCluster`. Looks like this method fetches kind/bootstrap cluster logs before the CAPI move.

pkg/diagnostics/diagnostic_bundle.go
jiayiwang7: is this method supposed to handle both bootstrap cluster before/after `clusterapi move`, eks-a management cluster in the management and workload setting? If so, why below are not part of the analyzers?
```
		WithGitOpsConfig(spec.GitOpsConfig).
		WithOidcConfig(spec.OIDCConfig).
		WithExternalEtcd(spec.Cluster.Spec.ExternalEtcdConfiguration).
		WithMachineConfigs(provider.MachineConfigs(spec)).
```

pkg/diagnostics/factory.go
maxdrib: This method gets invoked every time the `CollectDiagnosticsTask` task is run in workflows/diagnostics.go. If we're performing an operation from a management cluster to a workload cluster and something goes wrong and we start collecting diagnostics, it'll collect them from the management cluster, since there's no bootstrap cluster. So it seems like the terminology still doesn't make sense in that scenario. I can't think of the right thing to call this method since the definition of "management cluster" seems to be fluid. If a bootstrap cluster is in use, then that's the current management cluster. 

Is there a word for the cluster that's orchestrating the CLI operation?

pkg/diagnostics/diagnostic_bundle.go
maxdrib: It should probably handle the cases before/after `move`. I would imagine the reason is "because we've never needed it before"? But I'd prefer to redirect this conversation to a new issue, since I don't think it's related to my specific change to unblock the e2e tests.

jiayiwang7 (approved):

pkg/diagnostics/factory.go
jiayiwang7: I dont know that.. I've been confused by those terms in our codebase. I think we have a project to clean it up bit. https://github.com/aws/eks-anywhere/projects/10

maxdrib: I'm looking forward to re-learning the eks-a code flow through this lens

PR #2097, opened 2022-05-10
Support host paths to mount for cloudstack

wongni:
*Issue #, if available:* #2097

*Description of changes:*
Support CloudStack specific environment variable for the host paths to mount when running CLI executable to specify directories containing custom root CA certificates. 

*Testing (if applicable):*
Manual testing on the customer environment. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2219, opened 2022-05-20
Add cloudstack extra mounts for bootstrap

wongni:
*Issue #, if available:*

*Description of changes:*
I added extra host paths mounted to CLI container: https://github.com/aws/eks-anywhere/pull/2097 However, these paths are not mounted to the bootstrap kind cluster so the capc controller container cannot mount the host paths related to the custom root CA certificates. 

*Testing (if applicable):*
Manually tested


pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -61,6 +62,21 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
 	}, nil
 }
 
+func GetCloudStackExtraMounts() ([]string, error) {
+	var extraMounts []string
+	env, found := os.LookupEnv(EksaCloudStackHostPathToMount)
```
maxdrib, line 67: it seems like this should be a list object on the cloudstackdatacenterconfig, no? Could you remind me once more why we decided to go with the env var route?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -242,7 +242,16 @@ func (p *cloudstackProvider) UpdateKubeConfig(_ *[]byte, _ string) error {
 }
 
 func (p *cloudstackProvider) BootstrapClusterOpts() ([]bootstrapper.BootstrapClusterOption, error) {
-	return common.BootstrapClusterOpts(p.datacenterConfig.Spec.ManagementApiEndpoint, p.clusterConfig)
+	extraMounts, err := decoder.GetCloudStackExtraMounts()
```
maxdrib, line 245: could we add some unit tests here?

pkg/executables/kind.go
```diff
@@ -135,6 +136,22 @@ func (k *Kind) WithExtraDockerMounts() bootstrapper.BootstrapClusterClientOption
 	}
 }
 
+func (k *Kind) WithExtraCloudStackMounts(extraMounts []string) bootstrapper.BootstrapClusterClientOption {
```
maxdrib, line 139: could we add some unit tests here?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -221,16 +220,11 @@ func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cl
...
-			mountDirs := strings.Split(env, ",")
-			for _, dir := range mountDirs {
-				if _, err := os.Stat(dir); err != nil {
-					return nil, fmt.Errorf("invalid host path to mount: %v", err)
-				}
-				dirs = append(dirs, dir)
-			}
+		extraMounts, err := decoder.GetCloudStackExtraMounts()
+		if err != nil {
+			return nil, fmt.Errorf("invalid host path to mount: %v", err)
```
maxdrib, line 225: this error message may be confusing. We're not doing any validation on the host path itself, right? I feel like this more represents an issue with retrieving the extra mounts

pkg/bootstrapper/bootstrapper.go
```diff
@@ -134,6 +135,12 @@ func WithExtraPortMappings(ports []int) BootstrapClusterOption {
 	}
 }
 
+func WithExtraCloudStackMounts(extraMounts []string) BootstrapClusterOption {
```
maxdrib, line 138: it seems like a unit test here should be straightforward, no?

pkg/providers/cloudstack/decoder/decoder.go
wongni: Check this @vivek-koppuru 's comment about the env var approach: https://github.com/aws/eks-anywhere/pull/1981#issuecomment-1122626761

cmd/eksctl-anywhere/cmd/createcluster.go
wongni: GetCloudStackExtraMounts function checks if the host paths are existing. If one of the host paths doesn't exist the function returns an error.

pkg/executables/kind.go
```diff
@@ -50,6 +50,7 @@ type kindExecConfig struct {
 	RegistryCACertPath     string
 	ExtraPortMappings      []int
 	DockerExtraMounts      bool
+	CloudStackExtraMounts  []string
```
chrisdoherty4, line 53: Why does Kind have something Cloud Stack specific? Kind is just a development cluster management tool with some quirks around how it works. It has no reason to know about Cloud Stack.

pkg/executables/kind.go
```diff
@@ -135,6 +136,22 @@ func (k *Kind) WithExtraDockerMounts() bootstrapper.BootstrapClusterClientOption
...
 }
 
+func (k *Kind) WithExtraCloudStackMounts(extraMounts []string) bootstrapper.BootstrapClusterClientOption {
+	return func() error {
+		if k.execConfig == nil {
+			return errors.New("kind exec config is not ready")
+		}
+
+		if len(extraMounts) == 0 {
+			return errors.New("no extraMounts found in the list")
```
chrisdoherty4, line 146: `extraMounts` does mean anything to someone reading the error. Rule of thumb, don't include variable names in errors.

With respect to this behavior, why does it matter that no mounts were specified - wouldn't that just be a noop?

PR #2223, opened 2022-05-23
Remove obsolete flags from generatehardware cmd

chrisdoherty4:
With the kubeification of the Tinkerbell stack we no longer need flags pecific to registering with a pre-deployed Tinkerbell stack.

This doesn't update anything to generate new `Hardware` types. That work will be on a separate PR.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generatehardware.go
+++ b/cmd/eksctl-anywhere/cmd/generatehardware.go
@@ -1,38 +1,22 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 
 	"github.com/spf13/cobra"
 
-	"github.com/aws/eks-anywhere/pkg/executables"
-	"github.com/aws/eks-anywhere/pkg/networkutils"
 	"github.com/aws/eks-anywhere/pkg/providers/tinkerbell/hardware"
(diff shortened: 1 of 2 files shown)


cmd/eksctl-anywhere/cmd/generatehardware.go
```diff
@@ -91,75 +53,22 @@ func (hOpts *hardwareOptions) generateHardware(cmd *cobra.Command, args []string
...
-	jsonDir, err := hardware.CreateDefaultJSONDir(outputDir)
-	if err != nil {
-		return err
-	}
-
 	hardwareYAML, err := os.Create(filepath.Join(outputDir, hardware.DefaultHardwareManifestYAMLFilename))
 	if err != nil {
 		return fmt.Errorf("tinkerbell manifest yaml: %v", err)
 	}
 	yamlWriter := hardware.NewTinkerbellManifestYAML(hardwareYAML)
```
ptrivedi, line 60: I thought the intermediate hardware.yaml file was going away. Why are we still writing to DefaultHardwareManifestYAMLFilename file?

chrisdoherty4: Its not going away. We said we'd maintain it so users can generate hardware and apply it independently.

PR #2228, opened 2022-05-23
Remove old flag fields from Tinkerbell provider

chrisdoherty4:
Remove now redundant fields from the Tinkerbell provider. These fields are redundant because of an overhaul in validation logic and a change in design.

### DIFF
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -297,9 +297,7 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 				time.Now,
 				skipIpCheck,
 				hardwareConfigFile,
-				skipPowerActions,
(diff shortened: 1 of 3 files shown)


pkg/dependencies/factory.go
```diff
@@ -297,9 +297,7 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 				time.Now,
 				skipIpCheck,
 				hardwareConfigFile,
-				skipPowerActions,
```
ptrivedi, line 300: Why is skipPowerActions being removed? I understand that force flag might need cleanup and refactoring. Assuming there is refactoring being done around skipPowerActions as well and that a follow-up PR is imminent

pokearu (approved):
/lgtm

PR #2231, opened 2022-05-23
Rename providors validator pkg to validation

chrisdoherty4:
Rename the provider specific `validator` package, which is a noun, to `validation` to align with coding guidelines.

### DIFF
--- a/pkg/providers/snow/snow.go
+++ b/pkg/providers/snow/snow.go
@@ -16,7 +16,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 	"github.com/aws/eks-anywhere/pkg/providers"
-	providerValidator "github.com/aws/eks-anywhere/pkg/providers/validator"
+	"github.com/aws/eks-anywhere/pkg/providers/validation"
 	"github.com/aws/eks-anywhere/pkg/retrier"
 	"github.com/aws/eks-anywhere/pkg/templater"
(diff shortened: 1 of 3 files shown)


pkg/providers/snow/snow.go
```diff
@@ -72,7 +72,7 @@ func (p *SnowProvider) SetupAndValidateCreateCluster(ctx context.Context, cluste
 		return fmt.Errorf("setting defaults and validate snow config: %v", err)
 	}
 	if !p.skipIpCheck {
-		if err := providerValidator.ValidateControlPlaneIpUniqueness(clusterSpec.Cluster, &networkutils.DefaultNetClient{}); err != nil {
+		if err := validation.ValidateControlPlaneIpUniqueness(clusterSpec.Cluster, &networkutils.DefaultNetClient{}); err != nil {
```
ptrivedi, line 75: validator sounded okay, what's the reasoning behind changing it to 'validation'? Everything else is named like the 'doer' of the action e.g. retrier, templater, logger etc.

chrisdoherty4: `validator` is a noun. It implies there's some sort of state here. The majority of packages in the standard library don't use `er` names. They aren't 'wrong' but given this package describes a collection of related validation capabilities its peculiar to call it 'validator'.

chrisdoherty4: I should point out this isn't absolutely necessary. There's no hard rule here - `validation` seems more appropriate. We can decline the PR 😉

chrisdoherty4:
Closing as we might merge with the existing validations package at the top level.

PR #2234, opened 2022-05-24
Remove unused flux validations

taneyland:
*Issue #, if available:*

*Description of changes:*
We now use the config manager to validate configs for our clusters. We had some legacy validation functions in the codebase that led to engineers implementing validations twice. We also only had unit tests for those old validation functions so this PR adds unit tests using the new `Validate()` method.

This PR covers `FluxConfig` and `GitOpsConfig`. Will do other configs in separate PRs

*Testing (if applicable):*


pkg/cluster/flux_test.go
```diff
@@ -0,0 +1,176 @@
...
+	privateKeyFile         string
+	isPrivateKeyFileSet    bool
+	gitKnownHostsFile      string
+	isGitKnownHostsFileSet bool
+}
+
+func (tctx *testContext) SaveContext() {
+	tctx.privateKeyFile, tctx.isPrivateKeyFileSet = os.LookupEnv(EksaGitPrivateKeyTokenEnv)
+	tctx.gitKnownHostsFile, tctx.isGitKnownHostsFileSet = os.LookupEnv(EksaGitKnownHostsFileEnv)
+	os.Setenv(EksaGitPrivateKeyTokenEnv, "my/private/key")
```
g-gaston, line 31: Curious, why do we need to set these env vars? All the validations in the package `cluster` are supposed to be context agnostic and reading env vars is definitely context dependent.

pkg/cluster/flux_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+			if tt.gitProvider {
+				setupContext(t)
+			}
+
+			config, err := cluster.ParseConfigFromFile(tt.fileName)
+			g.Expect(c.Validate(config)).To(Succeed())
+			if err != nil {
+				t.Fatalf("Validate() error = %v, wantErr %v", err, nil)
+			}
```
g-gaston, line 99:
```suggestion
			g.Expect(err).NotTo(HaveOcurred())
```

pkg/cluster/flux_test.go
```diff
@@ -0,0 +1,176 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			g := NewWithT(t)
+			c := cluster.NewConfigManager()
+			c.RegisterValidations(
+				func(c *cluster.Config) error {
+					if c.FluxConfig != nil {
+						return c.FluxConfig.Validate()
```
g-gaston, line 154: Same here. It seems like these tests belong in the v1alpha1 package

g-gaston (approved):

PR #2235, opened 2022-05-24
Remove bogus entry from changelog

TerryHowe:
These items add no value that I know of


chrisnegus (approved):
lgtm

vivek-koppuru:
Actually this is used as our staging commit before we cut the release branch. However it's not being used as intended, I am adding it back here with the intention of this entry to change to `v0.10.0`: https://github.com/aws/eks-anywhere/pull/2414

PR #2243, opened 2022-05-24
Hide/default TinkerbellTemplateConfig

panktishah26:
*Description of changes:*
We are hiding TinkerbellTemplateConfig from the tinkerbell cluster config file and defaulting all the values of the field to avoid human error. However, a user can pass their own TinkerbellTemplateConfig to override default template.

### DIFF
--- a/pkg/api/v1alpha1/tinkerbelltemplateconfig.go
+++ b/pkg/api/v1alpha1/tinkerbelltemplateconfig.go
@@ -17,14 +17,14 @@ const TinkerbellTemplateConfigKind = "TinkerbellTemplateConfig"
 // +kubebuilder:object:generate=false
 type ActionOpt func(action *[]tinkerbell.Action)
 
-// NewDefaultTinkerbellTemplateConfigGenerate returns a default TinkerbellTemplateConfig with the required Tasks and Actions
-func NewDefaultTinkerbellTemplateConfigGenerate(name string, versionBundle v1alpha1.VersionsBundle) *TinkerbellTemplateConfigGenerate {
-	config := &TinkerbellTemplateConfigGenerate{
+// NewDefaultTinkerbellTemplateConfigCreate returns a default TinkerbellTemplateConfig with the required Tasks and Actions
+func NewDefaultTinkerbellTemplateConfigCreate(name string, versionBundle v1alpha1.VersionsBundle) *TinkerbellTemplateConfig {
+	config := &TinkerbellTemplateConfig{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       TinkerbellTemplateConfigKind,
 			APIVersion: SchemeBuilder.GroupVersion.String(),
 		},
-		ObjectMeta: ObjectMeta{
+		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
 		Spec: TinkerbellTemplateConfigSpec{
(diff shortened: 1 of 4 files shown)


pkg/providers/tinkerbell/template.go
```diff
@@ -47,6 +48,14 @@ func NewTinkerbellTemplateBuilder(datacenterSpec *v1alpha1.TinkerbellDatacenterC
 
 func (tb *TinkerbellTemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *cluster.Spec, buildOptions ...providers.BuildMapOption) (content []byte, err error) {
 	cpTemplateConfig := clusterSpec.TinkerbellTemplateConfigs[tb.controlPlaneMachineSpec.TemplateRef.Name]
```
chrisdoherty4, line 50: When would we ever expect this to not be nil?

pkg/api/v1alpha1/tinkerbelltemplateconfig.go
```diff
@@ -81,11 +81,10 @@ func GetTinkerbellTemplateConfig(fileName string) (map[string]*TinkerbellTemplat
 			if err = yaml.UnmarshalStrict([]byte(c), &template); err == nil {
 				templates[template.Name] = &template
 				continue
+			} else {
+				return nil, fmt.Errorf("invalid template config content: %v", err)
 			}
```
chrisdoherty4, line 86: Prefer returning early over nesting.

```suggestion
			if err := yaml.UnmarshalStrict([]byte(c), &template); err != nil {
				return nil, fmt.Errorf("invalid template config content: %v", err)
			}
			
			templates[template.Name] = &template
```

chrisdoherty4 (changes requested):
Seeking some clarifications and simplifications.

pkg/providers/tinkerbell/template.go
panktishah26: A user can define their own `TinkerbellTemplate` if they want to override the default one. I agree, we can have an optional action overrides compared to a complete custom resource but it needs some discussion and we can do that in the separate PR.

pkg/providers/tinkerbell/template.go
```diff
@@ -78,6 +87,13 @@ func (tb *TemplateBuilder) GenerateCAPISpecWorkers(clusterSpec *cluster.Spec, wo
 	workerSpecs := make([][]byte, 0, len(clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations))
 	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		wTemplateConfig := clusterSpec.TinkerbellTemplateConfigs[tb.WorkerNodeGroupMachineSpecs[workerNodeGroupConfiguration.MachineGroupRef.Name].TemplateRef.Name]
```
chrisdoherty4, line 89: What happens if I don't specify a template reference in the YAML? Won't this panic?

panktishah26: templateRef field is an optional field so it won't panic.

chrisdoherty4: Sorry, I thought `TemplateRef` was a pointer but its not, its a literal. All good!

chrisdoherty4 (approved):

cprivitere:
Just an FYI that this causes some issues when a user needs to customize the TinkerbellTemplateConfig. 
When it's not present in the cluster config file, there is literally no place the user can obtain the current   TinkerbellTemplateConfig that eksctl anywhere will use. The documentation has one that is old and is actually missing required fields.

PR #2249, opened 2022-05-24
Validate configs exist when calling GetFromRef

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/awsiam.go
+++ b/pkg/cluster/awsiam.go
@@ -1,6 +1,10 @@
 package cluster
 
-import anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+import (
+	"fmt"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
 
 func awsIamEntry() *ConfigManagerEntry {
 	return &ConfigManagerEntry{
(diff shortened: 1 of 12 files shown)


pkg/cluster/awsiam.go
```diff
@@ -54,3 +59,12 @@ func processAWSIam(c *Config, objects ObjectLookup) {
 		}
 	}
 }
+
+func validateAWSIamConfigName(c *Config) error {
+	for _, idr := range c.Cluster.Spec.IdentityProviderRefs {
+		if idr.Kind == anywherev1.AWSIamConfigKind && c.AWSIAMConfigs == nil {
+			return fmt.Errorf("no %s named %s", anywherev1.AWSIamConfigKind, idr.Name)
```
g-gaston, line 66: I would vote to have an error message a bit more descriptive, like:
```suggestion
			return fmt.Errorf("%s/%s referenced in Cluster but not present in Config", anywherev1.AWSIamConfigKind, idr.Name)
```

Or something else that gives us a bit more context on where the error is happening

g-gaston (commented):
Thanks for adding this! Could we add some tests?

pkg/cluster/awsiam_test.go
```diff
@@ -0,0 +1,181 @@
...
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			c := cluster.NewConfigManager()
+			c.RegisterValidations(
+				func(c *cluster.Config) error {
```
g-gaston, line 157: This is not really testing the code, it's just duplicating it in the tests. In fact you can see in the codecov annotations that your new code is not being covered by tests.

Curious, what was the reason to do this?

If you want to use the config manager, I would suggest just using the `cluster.Validate(config)` method. That will use the default config manager that already has your validation registered.

The other way, is to make your new validation methods public and just call them directly from here, without the config manager.

PR #2254, opened 2022-05-24
Validate Tinkerbell datacenter config on create

chrisdoherty4:
As part of the Tinkerbell validation refactoring we isolated datacenter config validation. It was left out of the core set of assertions we run. This adds it ensuring we validate everything in there.


pokearu (approved):
/approve
/lgtm

PR #2261, opened 2022-05-25
Add disk to Tinkerbell CSV input

chrisdoherty4:
Builds on #2268 

Add `Disk` to the Tinkerbell CSV so it can be consumed for `TinkerbellMachineTemplate` generation. The `Disk` will be validated in a subsequent PR where it will need to be consistent among hardware expected to be provisioned for control plane, external etcd or individual node groups.

The PR only adds CSV input data - it doesn't introduce any functional changes beyond that.

### DIFF
--- a/pkg/providers/tinkerbell/hardware/validator.go
+++ b/pkg/providers/tinkerbell/hardware/validator.go
@@ -3,6 +3,7 @@ package hardware
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"strings"
 
 	apimachineryvalidation "k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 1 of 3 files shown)


pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -96,6 +102,13 @@ func StaticMachineAssertions() MachineAssertion {
 			return fmt.Errorf("invalid hostname: %v: %v", m.Hostname, errs)
 		}
 
+		if !linuxPathValidation.MatchString(m.Disk) {
+			return fmt.Errorf(
+				"disk should be a linux paths (\"%v\")",
```
ptrivedi, line 107: nit: grammar

pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -43,6 +44,11 @@ func (mv *DefaultMachineValidator) Register(v ...MachineAssertion) {
 	mv.assertions = append(mv.assertions, v...)
 }
 
+var (
+	linuxPathRegex      = `^/|(/[\w-]+)+$`
```
ptrivedi, line 48: wonder if we should force either have the path start with '/dev' or if it doesn't prepend ourselves

pkg/providers/tinkerbell/hardware/validator.go
chrisdoherty4: Updated

pkg/providers/tinkerbell/hardware/validator.go
chrisdoherty4: Made the check stricter with a `/dev/` prefix requirement.

PR #2262, opened 2022-05-25
Snow proxy

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/clusterapi/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/templater"
+)
+
+//go:embed config/http-proxy.conf
+var proxyConfig string
+
+var NoProxyDefaults = []string{
```
g-gaston, line 17: Does this need to be public? If it is, I would vote to expose it through a method. This makes the variable mutable from the outside of the package

pkg/clusterapi/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func proxyConfigFile(cluster v1alpha1.ClusterSpec) (bootstrapv1.File, error) {
+	proxyConfig, err := proxyConfigContent(cluster)
+	if err != nil {
+		return bootstrapv1.File{}, err
+	}
+
+	return bootstrapv1.File{
+		Path:    "/etc/systemd/system/containerd.service.d/http-proxy.conf",
```
g-gaston, line 57: Similar comment here about different OS
I believe this works differently for Bottlerocket (for example)

jiayiwang7: proxy config setup in provider now

pkg/clusterapi/kubeadm_commands.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+var restartContainerdCommands = []string{
+	"sudo systemctl daemon-reload",
+	"sudo systemctl restart containerd",
```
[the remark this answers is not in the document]
jiayiwang7: called in provider

g-gaston (approved):

PR #2268, opened 2022-05-26
Update tinkerbell kubernetes types

chrisdoherty4:
The CSV to Kubernetes object conversions used old Hardware and baseboard management computer types. This updates them to use new types in preparation for introducing conversion code in the create cmd.

A subsequent PR will introduce logic for writing CSV entries to the catalogue so they can be marshalled and submitted to the cluster using existing mechanisms.

_This PR has no functional changes_


pkg/providers/tinkerbell/hardware/fmt.go
```diff
@@ -0,0 +1,11 @@
+package hardware
+
+import "fmt"
+
+func formatBMCRef(m Machine) string {
```
chrisdoherty4, line 5: Ignore these - they're used in a subsequent PR. They're dead because of some deletions.

pkg/providers/tinkerbell/hardware/fmt.go
```diff
@@ -0,0 +1,9 @@
+package hardware
+
+// func formatBMCRef(m Machine) string {
```
chrisdoherty4, line 3: These are used in a subsequent PR but the linter flags them as deadcode due to deletions hence they're commented out.

pkg/providers/tinkerbell/hardware/yaml.go
```diff
@@ -84,81 +70,21 @@ func (yw *TinkerbellManifestYAML) write(data []byte) error {
...
-				BmcRef: formatBMCRef(m),
-			},
-		},
-	)
+	return yaml.Marshal(tinkv1alpha1.Hardware{})
 }
 
 func marshalTinkerbellBMCYAML(m Machine) ([]byte, error) {
 	return yaml.Marshal(
-		pbnjv1alpha1.BMC{
```
chrisdoherty4, line 111: These will be updated. They currently only feature in the `generate` command so will circle round when we have a functional `create` cmd.

PR #2269, opened 2022-05-26
Set TINKERBELL_IP when retrieving Tinkerbell provider environment

chrisdoherty4:
Define the `TINKERBELL_IP` that gets propogated to CAPT installation code where its used as a value replacement so CAPT knows what address to specify in `Userdata` for contacting Hegel. 

This, in our context, is totally unnecessary. The only code that uses it is https://github.com/chrisdoherty4/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L192 which renders a default template. Given we override the template the configuration is meaningless for us and only necessary to satisfy CAPT's [`Deployment` spec](https://github.com/chrisdoherty4/cluster-api-provider-tinkerbell/blob/main/config/manager/manager.yaml#L23).

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -208,8 +208,22 @@ func (p *Provider) Version(clusterSpec *cluster.Spec) string {
 	return clusterSpec.VersionsBundle.Tinkerbell.Version
 }
 
-func (p *Provider) EnvMap(_ *cluster.Spec) (map[string]string, error) {
-	return map[string]string{}, nil
+func (p *Provider) EnvMap(spec *cluster.Spec) (map[string]string, error) {
+	return map[string]string{
+		// The TINKERBELL_IP is input for the CAPT deployment and used as part of default template
+		// generation. However, we use custom templates and leverage the template override
+		// functionality of CAPT hence this never gets used.
+		//
+		// Deployment manifest requiring the env var for replacement.
+		// https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/config/manager/manager.yaml#L23
+		//
+		// Template override
+		// https://github.com/chrisdoherty4/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L182
+		//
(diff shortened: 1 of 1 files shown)


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -208,8 +208,14 @@ func (p *Provider) Version(clusterSpec *cluster.Spec) string {
 	return clusterSpec.VersionsBundle.Tinkerbell.Version
 }
 
-func (p *Provider) EnvMap(_ *cluster.Spec) (map[string]string, error) {
-	return map[string]string{}, nil
+func (p *Provider) EnvMap(spec *cluster.Spec) (map[string]string, error) {
+	environment := map[string]string{}
+	if p.datacenterConfig == nil {
```
abhinavmpandey08, line 213: should this be 
```suggestion
	if p.datacenterConfig != nil {
```

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -208,8 +208,14 @@ func (p *Provider) Version(clusterSpec *cluster.Spec) string {
...
+		environment["TINKERBELL_IP"] = p.datacenterConfig.Spec.TinkerbellIP
+	} else {
+		logger.V(2).Info("Tinkerbell datacenter config not specified in provider; cannot define TINKERBELL_IP environment variable")
```
abhinavmpandey08, line 216: should this be an error since if it's not provided, the cluster creation would fail? Or can we just give it a dummy value temporarily?

chrisdoherty4: I've just changed to using dummy values for everything because I've realized this configuration is totally superfluous in our context.

See the in-line code commentary on why.

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: [See my other comment.](https://github.com/aws/eks-anywhere/pull/2269#discussion_r882690241)

abhinavmpandey08 (approved):

PR #2270, opened 2022-05-26
Skipping cloudstack tests to remove noise while setting up CI test env

maxdrib:
*Issue #, if available:*

*Description of changes:*
While we're in the process of setting up the test environment for Cloudstack, we should scale back how many tests are running. Once we get a single test running well then we can start adding more. For now, the extra tests are slowing down the cloudstack environment and causing timeouts left and right, and also adding noise to the test results.

This change leaves two Cloudstack tests in place which include TestCloudStackKubernetes121SimpleFlow and TestCloudStackKubernetes121WorkloadCluster.

*Testing (if applicable):*


jaxesn (approved):
/approve

mrog (approved):

PR #2274, opened 2022-05-26
Use a helm chart to install tinkerbell stack

abhinavmpandey08:
*Description of changes:*
This PR introduces helm chart for installing tinkerbell stack as it provides more configurability then static manifests

*Testing (if applicable):*
Created a cluster using the changes and verified everything worked as expected


pkg/providers/tinkerbell/create.go
```diff
@@ -42,50 +43,84 @@ func (p *Provider) BootstrapClusterOpts() ([]bootstrapper.BootstrapClusterOption
...
+		withHegel().
+		withTinkController().
+		withTinkServer()
+
+	if err := stack.install(ctx, cluster.KubeconfigFile); err != nil {
+		return fmt.Errorf("install Tinkerbell stack on bootstrap cluster: %v", err)
+	}
+
+	if err := stack.installBootsOnDocker(ctx, cluster.KubeconfigFile); err != nil {
+		return err
```
chrisdoherty4, line 69: I think there's too much implementation detail here. Why does the provider care about the fact boots is installed on docker? Can we model a default installer's behavior, which includes all components of the stack, and offer options where control is needed (probably only boots?)? 

For example:

```go
installer := NewStackInstaller(..., InstallBootsOnDocker())

if err := installer.Install(ctx, kubeconfig); err != nil {
  return ...
}
```

Here, the provider only cares about 2 things: (1) create an installer; (2) run installation. This paves the way to extract the installer and inject it having the provider leverage an interface scoped to `Install` only (_we don't need to do this now_).

```go
type StackInstaller interface {
  InstallStack(context.Context, string) error
}
```

I mention this because its also a little odd that the provider consumes a `Helm` related object. The provider should care what mechanism is used to install anything - only that it controls _when_ the install cmd happens.

pkg/providers/tinkerbell/stack.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (s *stack) getBootsEnv() map[string]string {
+	return map[string]string{
+		"DATA_MODEL_VERSION":        "kubernetes",
+		"TINKERBELL_TLS":            "false",
+		"TINKERBELL_GRPC_AUTHORITY": fmt.Sprintf("%s:%s", s.ip, grpcPort),
+		"BOOTS_EXTRA_KERNEL_ARGS":   fmt.Sprintf("tink_worker_image=%s", s.bundle.Tink.TinkWorker.URI),
+		// TODO: Pull this from bundle instead
+		"MIRROR_BASE_URL": "https://tinkerbell-storage-for-eksa.s3.us-west-2.amazonaws.com",
```
jaxesn, line 167: Is this the actual chart?  We should publish this via build-tooling into ecr like we have the packages charts and the cilium chart.

pkg/networkutils/networkutils.go
```diff
@@ -42,3 +42,13 @@ func IsIPInUse(client NetClient, ip string) bool {
 
 	return false
 }
+
+func GetLocalIP(client NetClient) (net.IP, error) {
```
chrisdoherty4, line 46: I don't think there's a need to inject the `NetClient`. I appreciate the testability concern but this isn't the API consumers really want. They really want a `GetLocalIP()` and thats all. We should probably embrace that - as should the IP checking function at some point - and do the best we can with testing. 

This also means we can get rid of the timeout and simplify to just `net.Dial(...)`.

pkg/providers/tinkerbell/stack.go
abhinavmpandey08: yep I will add that in the build-tooling

pkg/networkutils/networkutils.go
```diff
@@ -42,3 +42,13 @@ func IsIPInUse(client NetClient, ip string) bool {
 
 	return false
 }
+
+func GetLocalIP() (net.IP, error) {
+	conn, err := net.Dial("udp", "1.2.3.4:80")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve local IP: %v", err)
```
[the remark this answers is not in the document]
abhinavmpandey08: shouldn't we use UPPER CASE for acronyms?

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -0,0 +1,212 @@
+package stack
```
chrisdoherty4, line 1: Do we need a new package for this? This just seems like a component of the `tinkerbell` package.

pkg/networkutils/networkutils_test.go
```diff
@@ -96,9 +104,12 @@ func (m *MockConn) Close() error {
 	return ret0
 }
 
+func (m *MockConn) LocalAddr() net.Addr {
```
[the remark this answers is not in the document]
abhinavmpandey08: no longer needed, will remove

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -0,0 +1,212 @@
...
+
+type StackInstaller interface {
+	// WithNamespace(ns string, create bool) *Installer
+	// WithBootsOnDocker() *Installer
+	// WithBootsOnKubernetes() *Installer
+	Install(ctx context.Context, bundle releasev1alpha1.TinkerbellStackBundle, tinkServerIP, kubeconfig string, opts ...InstallerOption) error
+	UninstallLocal(ctx context.Context) error
+}
+
+func WithNamespace(ns string, create bool) InstallerOption {
```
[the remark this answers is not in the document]
abhinavmpandey08: this is needed because the cli performs different actions on bootstrap vs workload cluster currently. On bootstrap cluster, it creates the namespace initially but on workload it doesn't create the namespace until much later.
Helm errors out if we try to create a namespace that already exists. So for bootstrap we need to skip ns creation since the cli already creates it. For workload, we do need to create the namespace.

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (s *Installer) getBootsEnv(bundle releasev1alpha1.TinkerbellStackBundle, tinkServerIP string) map[string]string {
+	return map[string]string{
+		"DATA_MODEL_VERSION":        "kubernetes",
+		"TINKERBELL_TLS":            "false",
+		"TINKERBELL_GRPC_AUTHORITY": fmt.Sprintf("%s:%s", tinkServerIP, grpcPort),
+		"BOOTS_EXTRA_KERNEL_ARGS":   fmt.Sprintf("tink_worker_image=%s", bundle.Tink.TinkWorker.URI),
+		// TODO: Pull this from bundle instead
+		"MIRROR_BASE_URL": "https://tinkerbell-storage-for-eksa.s3.us-west-2.amazonaws.com",
```
ptrivedi, line 197: this would go away in favor of osiePathOverride

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -0,0 +1,212 @@
...
+	flags := []string{
+		"-v", fmt.Sprintf("%s:/kubeconfig", kubeconfig),
+		"--network", "host",
+	}
+
+	for name, value := range s.getBootsEnv(bundle, tinkServerIP) {
+		flags = append(flags, "-e", fmt.Sprintf("%s=%s", name, value))
+	}
+
+	cmd := []string{"-kubeconfig", "/kubeconfig", "-dhcp-addr", "0.0.0.0:67"}
```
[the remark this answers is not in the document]
abhinavmpandey08: We have to update the build tooling to track the latest commit of boots for it to work.
Do you mind if I do this in a follow-up PR after the build-tooling is updated? Would like to have one working commit in the mainline

pkg/providers/tinkerbell/stack/stack.go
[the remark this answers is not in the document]
chrisdoherty4: Can we make it a separate option given there's only ever 1 namespace at play. This allows consumers to specify the desired behavior independently of specifying the namespace and rely on default behavior around namespace creation without supplying any options.

pkg/providers/tinkerbell/stack/stack.go
chrisdoherty4: Packages can typically be argued any which way. I'm asking myself what is `stack` providing that I wouldn't expect from `tinkerbell`. With that in mind, I noticed `tinkerbell` imports `stack` to use some interfaces from it. At that point it started to feel like we've created a separation between 2 things that are working toward providing the same capability, hence should be in the same package (at least until we get all pieces in place and figure out a logical break of the `tinkerbell` package). 

Its trivial, for now, to change so you decide and we can rework as necessary after mulling it over for a few days.

pkg/providers/tinkerbell/stack/stack.go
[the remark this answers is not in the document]
ptrivedi: yep, makes sense to do it in a separate PR

PR #2275, opened 2022-05-26
Add adaptive wait duration between retries when waiting for machine replicas are ready

wongni:
*Issue #, if available:*

*Description of changes:*
Currently it waits 0 second between retries when waiting for machine deployment's replicas are ready. This could overload kube api server. So I introduce non zero wait time 1% proportional to the total timeout between retries. 

*Testing (if applicable):*
Unit testing.


pkg/retrier/retrier.go
```diff
@@ -40,6 +40,11 @@ func NewWithMaxRetries(maxRetries int, backOffPeriod time.Duration) *Retrier {
 	return New(time.Duration(math.MaxInt64), WithMaxRetries(maxRetries, backOffPeriod))
 }
 
+// NewWithNSecondsWaitRetries creates a new retrier with a global timeout and N-seconds wait retries policy
+func NewWithNSecondsWaitRetries(timeout time.Duration, waitSeconds int) *Retrier {
```
g-gaston, line 44:
```suggestion
func NewWithBackOff(timeout time.Duration, waitSeconds int) *Retrier {
```

pkg/retrier/retrier.go
```diff
@@ -48,6 +53,14 @@ func WithMaxRetries(maxRetries int, backOffPeriod time.Duration) RetrierOpt {
 	}
 }
 
+// WithNSecondsWaitRetries sets a retry policy that will retry forever with N-seconds wait time
+// between retries
+func WithNSecondsWaitRetries(waitSeconds int) RetrierOpt {
```
g-gaston, line 58: Can we take here a `time.Duration` instead of int?
```suggestion
func WithBackOff(backOffPeriod time.Duration) RetrierOpt {
```

pkg/retrier/retrier.go
```diff
@@ -102,6 +115,12 @@ func zeroWaitPolicy(_ int, _ error) (retry bool, wait time.Duration) {
 	return true, 0
 }
 
+func nSecondWaitPolicy(waitSeconds int) RetryPolicy {
```
g-gaston, line 118:
```suggestion
func backOffPolicy(backOffPeriod time.Duration) RetryPolicy {
```

pkg/retrier/retrier_test.go
```diff
@@ -171,3 +171,35 @@ func TestNewWithRetryPolicyFinishByPolicy(t *testing.T) {
...
+	if err := r.Retry(fn); err == nil {
+		t.Fatal("Retrier.Retry() error = nil, want not nil")
+	}
+
+	if gotRetries != wantRetries/waitSeconds {
+		t.Fatalf("Wrong number of retries, got %d, want %d", gotRetries, wantRetries/waitSeconds)
+	}
+
+	waitSeconds = 3
+	r = retrier.NewWithNSecondsWaitRetries(time.Duration(wantRetries/waitSeconds)*time.Second, waitSeconds)
```
g-gaston, line 195: can we separate these two cases and make a table test?

g-gaston (commented):
Looks good!
A few naming suggestions for consistency 🙂

pkg/clustermanager/cluster_manager.go
```diff
@@ -727,7 +728,7 @@ func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Conte
 		timeout = c.machinesMinWait
 	}
 
-	r := retrier.New(timeout)
+	r := retrier.NewWithProportionalBackOff(timeout, waitProportion)
```
g-gaston, line 731: This can lead to very long wait times if there is a large number of machines.
Imagine 20 machines. Total wait time will be 200 minutes. So the backoff time will be 2 minutes.

What do you think about implementing something similar to what we have for `waitForNodesReady`?
The backoff time is dynamic, so the more machines unready, the more we wait in between retries. They more machines ready, the less we wait. This same file, line 740

pkg/clustermanager/cluster_manager.go
```diff
@@ -91,7 +91,7 @@ type ClusterClient interface {
 	CreateNamespace(ctx context.Context, kubeconfig string, namespace string) error
 	GetNamespace(ctx context.Context, kubeconfig string, namespace string) error
 	ValidateControlPlaneNodes(ctx context.Context, cluster *types.Cluster, clusterName string) error
-	ValidateWorkerNodes(ctx context.Context, clusterName string, kubeconfigFile string) error
+	ValidateWorkerNodes(ctx context.Context, clusterName string, kubeconfigFile string) (int, int, error)
```
g-gaston, line 93: can you name these returns? :)
Specially with this method name, it's quite difficult to know what those ints are

pkg/clustermanager/cluster_manager.go
```diff
@@ -709,16 +709,23 @@ func (c *ClusterManager) waitForControlPlaneReplicasReady(ctx context.Context, m
...
+	}
+
 	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		machineDeploymentReplicasCount += workerNodeGroupConfiguration.Count
 	}
 
 	isMdReady := func() error {
-		return c.clusterClient.ValidateWorkerNodes(ctx, clusterSpec.Cluster.Name, managementCluster.KubeconfigFile)
+		readyReplicas, totalReplicas, err = c.clusterClient.ValidateWorkerNodes(ctx, clusterSpec.Cluster.Name, managementCluster.KubeconfigFile)
+		return err
```
g-gaston, line 724: shouldn't we compare here `readyReplicas == totalReplicas`?

pkg/executables/kubectl.go
```diff
@@ -480,26 +480,36 @@ func (k *Kubectl) ValidateControlPlaneNodes(ctx context.Context, cluster *types.
...
 		if machineDeployment.Status.ReadyReplicas != machineDeployment.Status.Replicas {
-			return fmt.Errorf("%v machine deployment replicas are not ready", machineDeployment.Status.Replicas-machineDeployment.Status.ReadyReplicas)
+			return readyReplicas, totalReplicas, fmt.Errorf("%v replicas of machine deployment %s replicas are not ready", machineDeployment.Status.Replicas-machineDeployment.Status.ReadyReplicas, machineDeployment.Name)
 		}
 	}
-	return nil
+	return readyReplicas, totalReplicas, nil
+}
+
+func CountReplicasByStatus(deployments []clusterv1.MachineDeployment) (int, int) {
```
g-gaston, line 506: this doesn't seem to take consideration the same factors that the original method did
for example, it doesn't check `machineDeployment.Status.Phase != "Running"`

pkg/executables/kubectl.go
```diff
@@ -480,26 +480,36 @@ func (k *Kubectl) ValidateControlPlaneNodes(ctx context.Context, cluster *types.
...
 		}
 
 		if machineDeployment.Status.UnavailableReplicas != 0 {
-			return fmt.Errorf("%v machine deployment replicas are unavailable", machineDeployment.Status.UnavailableReplicas)
+			return readyReplicas, totalReplicas, fmt.Errorf("%v replicas of machine deployment %s are unavailable", machineDeployment.Status.UnavailableReplicas, machineDeployment.Name)
 		}
 
 		if machineDeployment.Status.ReadyReplicas != machineDeployment.Status.Replicas {
-			return fmt.Errorf("%v machine deployment replicas are not ready", machineDeployment.Status.Replicas-machineDeployment.Status.ReadyReplicas)
+			return readyReplicas, totalReplicas, fmt.Errorf("%v replicas of machine deployment %s replicas are not ready", machineDeployment.Status.Replicas-machineDeployment.Status.ReadyReplicas, machineDeployment.Name)
```
g-gaston, line 500: I don't know if returning an error makes sense anymore when we are returning the number or ready and not ready replicas
I think when providing that info, it should up to the caller to determine if that's an error or not

pkg/clustermanager/cluster_manager.go
```diff
@@ -727,7 +734,7 @@ func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Conte
 		timeout = c.machinesMinWait
 	}
 
-	r := retrier.New(timeout)
+	r := retrier.New(timeout, retrier.WithRetryPolicy(policy))
```
g-gaston, line 737: Codecov is reporting this is not covered, we might need to add a test here 🙂

pkg/executables/kubectl.go
```diff
@@ -480,26 +480,36 @@ func (k *Kubectl) ValidateControlPlaneNodes(ctx context.Context, cluster *types.
 	return nil
 }
 
-func (k *Kubectl) ValidateWorkerNodes(ctx context.Context, clusterName string, kubeconfig string) error {
+func (k *Kubectl) ValidateWorkerNodes(ctx context.Context, clusterName string, kubeconfig string) (int, int, error) {
```
g-gaston, line 483: This method seems to not be covered by tests either 🙂

pkg/retrier/retrier.go
```diff
@@ -48,6 +48,13 @@ func WithMaxRetries(maxRetries int, backOffPeriod time.Duration) RetrierOpt {
 	}
 }
 
+// WithBackOff sets a retry policy with back-off between retries
+func WithBackOff(wait time.Duration) RetrierOpt {
```
g-gaston, line 52: If we are not going to use this, I would vote to remove it :)

pkg/clustermanager/cluster_manager.go
chrisdoherty4: Better yet - don't return those 2 things. Of the 3 places I found this function being used, _none of them_ used either of the returned integers. The 1 place I found they _might_ be used but its unclear is l.715 of those file. I'm not sure how that `policy` inline funciton definition is being used but its awfully convoluted. 

At a high level, looking at these changes, there's a mixture of 'Validate' and 'Retrieve' behavior munged into single functions that really needs to be pulled apart to honor the SRP.

pkg/clustermanager/cluster_manager.go
```diff
@@ -708,16 +708,23 @@ func (c *ClusterManager) waitForControlPlaneReplicasReady(ctx context.Context, m
 }
 
 func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec) error {
+	var err error
 	var machineDeploymentReplicasCount int
+	readyReplicas, totalReplicas := 0, 0
+	policy := func(_ int, _ error) (bool, time.Duration) {
```
chrisdoherty4, line 714: Why define an inline function with parameters and not use them?

chrisdoherty4: I see why.

Can we move this down and define it in the `retrier.WithRetryPolicy()`?

pkg/clustermanager/cluster_manager.go
```diff
@@ -708,16 +708,23 @@ func (c *ClusterManager) waitForControlPlaneReplicasReady(ctx context.Context, m
 }
 
 func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec) error {
+	var err error
```
chrisdoherty4, line 711: You don't really need to declare the error up here - its rather abnormal for Go.

chrisdoherty4 (changes requested):
See inline comments.

pkg/retrier/retrier.go
```diff
@@ -102,6 +102,12 @@ func zeroWaitPolicy(_ int, _ error) (retry bool, wait time.Duration) {
 	return true, 0
 }
 
+func DurationWaitPolicy(waitDuration time.Duration) RetryPolicy {
```
g-gaston, line 105: could we add a test for this one? :)

g-gaston (commented):
The patch coverage is a bit low, could we add some tests?
Also, before merging, would you make sure we squash all those commits? Thanks!

pkg/clustermanager/cluster_manager.go
```diff
@@ -709,11 +710,13 @@ func (c *ClusterManager) waitForControlPlaneReplicasReady(ctx context.Context, m
...
 	var machineDeploymentReplicasCount int
+	readyReplicas, totalReplicas := 0, 0
 	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		machineDeploymentReplicasCount += workerNodeGroupConfiguration.Count
 	}
 
 	isMdReady := func() error {
+		readyReplicas, totalReplicas, _ = c.countReplicasReady(ctx)
 		return c.clusterClient.ValidateWorkerNodes(ctx, clusterSpec.Cluster.Name, managementCluster.KubeconfigFile)
 	}
```
chrisdoherty4, line 722: Would you mind deleting l.723-726. The retry loop will execute N retries and if it succeeds on the first attempt the function returns as expected. As is, we have N+1 attempts which feels a little unnecessary.

pkg/clustermanager/cluster_manager.go
```diff
@@ -727,13 +730,30 @@ func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Conte
 		timeout = c.machinesMinWait
 	}
 
-	r := retrier.New(timeout)
+	retryPolicy := func(_ int, _ error) (bool, time.Duration) {
+		return true, c.machineBackoff * time.Duration(totalReplicas-readyReplicas)
```
chrisdoherty4, line 734: This wait time is a little odd. If I have 10 machines and 1 ready it waits 9 seconds. On a second attempt if I have 9 machines ready we still wait 9 seconds? 

There's lots of in-between obviously but this is peculiar, what's your intent?

pkg/clustermanager/cluster_manager.go
```diff
@@ -709,11 +710,13 @@ func (c *ClusterManager) waitForControlPlaneReplicasReady(ctx context.Context, m
...
 func (c *ClusterManager) waitForMachineDeploymentReplicasReady(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec) error {
 	var machineDeploymentReplicasCount int
+	readyReplicas, totalReplicas := 0, 0
 	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		machineDeploymentReplicasCount += workerNodeGroupConfiguration.Count
 	}
 
 	isMdReady := func() error {
+		readyReplicas, totalReplicas, _ = c.countReplicasReady(ctx)
 		return c.clusterClient.ValidateWorkerNodes(ctx, clusterSpec.Cluster.Name, managementCluster.KubeconfigFile)
```
g-gaston, line 720: I'm not sure I follow this
Shouldn't `readyReplicas == totalReplicas` be enough here?
Maybe I don't understand what this is doing

PR #2277, opened 2022-05-26
Configuring CloudStack e2e tests to run successfully

maxdrib:
*Issue #, if available:*

*Description of changes:*
The VM cleanup script at the end of the e2e tests for removing dangling CloudStack VM's is failing because cloudmonkey is timing out. This PR sets the [env variable to configure that timeout](https://github.com/aws/eks-anywhere/blob/main/pkg/executables/cmk.go#L458) to be 180s, up from the [default 30 seconds](https://github.com/aws/eks-anywhere/blob/main/pkg/executables/cmk.go#L28).

Also, even though the test framework does check if an IP is in use, we should still reserve a CIDR specifically for kube-vip control planes to prevent accidental conflicts. This change modifies the CIDR from being the entire network's subnet CIDR to being a reserved CIDR specifically for kube-vip IP addresses

*Testing (if applicable):*
N/A


maxdrib:
Timeout bump is not needed, and CIDR changes is covered in https://github.com/aws/eks-anywhere/pull/2394

PR #2280, opened 2022-05-26
Configurable replica ready timeout per machine

wongni:
*Issue #, if available:*

*Description of changes:*
The default 10 minutes per machine isn't enough for a certain customer when they upgrade a cluster. Instead of increasing the default value to 15m or 20m I made it configurable via an environment variable (EKSA_REPLICAS_READY_TIMEOUT)

*Testing (if applicable):*
Unit testing.


pkg/clustermanager/cluster_manager.go
```diff
@@ -137,6 +139,18 @@ func New(clusterClient ClusterClient, networking Networking, writer filewriter.F
 	return c
 }
 
+func GetMaxWaitPerMachine() time.Duration {
```
g-gaston, line 142: This is a good idea!

However I would like to see this being read at the beginning of the program (from the cmd package) and then being injected in the cluster manager constrcutor

We can probably use the `CliConfig`, pass it to the dependency factory and then inject this value from there into `ClusterManager`


Happy to to move this conversation offline if that's easier :)

pkg/dependencies/factory.go
```diff
@@ -653,6 +653,19 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster) *Factory {
 	return f
 }
 
+func (f *Factory) WithMaxWaitPerMachine(maxWaitPerMachine time.Duration) *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.dependencies.ClusterManager != nil {
+			return nil
+		}
+
+		f.dependencies.ClusterManager.SetMaxWaitPerMachine(maxWaitPerMachine)
```
g-gaston, line 663: Instead of doing this, which relays on order (if cluster manager is nil, this will fail), I would just store `CliConfig` in the factory

Then, when you build ClusterManager, pass the `MaxWaitPerMachine` to its constructor

pkg/dependencies/factory.go
```diff
@@ -647,20 +651,17 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster) *Factory {
...
 		)
 		return nil
 	})
 
 	return f
 }
 
-func (f *Factory) WithMaxWaitPerMachine(maxWaitPerMachine time.Duration) *Factory {
+func (f *Factory) WithCliConfig(cliConfig *config.CliConfig) *Factory {
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
```
g-gaston, line 663: We can get rid of the the build step completely :)
just 
```go
f.dependencies.CliConfig = cliConfig
return nil
```

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -243,6 +244,7 @@ func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
 		cliConfig.GitPrivateKeyFile = os.Getenv(config.EksaGitPrivateKeyTokenEnv)
 		cliConfig.GitKnownHostsFile = os.Getenv(config.EksaGitKnownHostsFileEnv)
 	}
+	cliConfig.MaxWaitPerMachine = config.GetMaxWaitPerMachine()
```
g-gaston, line 247: nit: can we do this around line 140 where we create the config object?

pkg/clustermanager/cluster_manager.go
```diff
@@ -152,6 +151,10 @@ func WithRetrier(retrier *retrier.Retrier) ClusterManagerOpt {
 	}
 }
 
+func (c *ClusterManager) SetMaxWaitPerMachine(maxWaitPerMachine time.Duration) {
```
g-gaston, line 154: Can we remove this? :)

pkg/dependencies/factory.go
```diff
@@ -637,6 +638,9 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster) *Factory {
 		if f.dependencies.ClusterManager != nil {
 			return nil
 		}
+		if f.dependencies.CliConfig == nil {
+			return nil
```
g-gaston, line 642: can we return an error here? or just use `0`

g-gaston (approved):

PR #2282, opened 2022-05-27
Fix create cluster hardware csv flag.

chrisdoherty4:
The hardware csv flag required for tinkerbell provider provisioning has some additional checks that used the old flag.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -27,6 +27,8 @@ import (
 	"github.com/aws/eks-anywhere/pkg/workflows"
 )
 
+const tinkerbellHardwareCSVFlag = "hardware-csv"
+
 type createClusterOptions struct {
 	clusterOptions
 	forceClean       bool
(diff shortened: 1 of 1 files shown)


cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -52,7 +54,7 @@ func init() {
 	createCmd.AddCommand(createClusterCmd)
 	createClusterCmd.Flags().StringVarP(&cc.fileName, "filename", "f", "", "Filename that contains EKS-A cluster configuration")
 	if features.IsActive(features.TinkerbellProvider()) {
-		createClusterCmd.Flags().StringVar(&cc.hardwareCSVPath, "hardware-csv", "", "A file path to a CSV file containing hardware data to be submitted to the cluster for provisioning")
+		createClusterCmd.Flags().StringVar(&cc.hardwareCSVPath, tinkerbellHardwareCSVFlag, "", "A file path to a CSV file containing hardware data to be submitted to the cluster for provisioning")
```
panktishah26, line 57: Should we give an alias like `w` or any other letter and also keep the name little smaller like hardwareCSV or something?

PR #2286, opened 2022-05-27
Extend testing section in best practice doc

g-gaston:

### DIFF
--- a/docs/developer/best-practice.md
+++ b/docs/developer/best-practice.md
@@ -147,6 +147,12 @@ package foo_test
 
 The compiler will build files that declare a package with the suffix "_test" as a separate package, then link and run with the main test binary. This ensures tests cases only have access to the public API of a package.
 
+Test one scenario per test. This makes reading and debugging tests easier plus it keeps the test name simple (remember that the last part of the test function name describes the scenario you are testing).
+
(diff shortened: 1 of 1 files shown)


docs/developer/best-practice.md
```diff
@@ -147,6 +147,12 @@ package foo_test
 
 The compiler will build files that declare a package with the suffix "_test" as a separate package, then link and run with the main test binary. This ensures tests cases only have access to the public API of a package.
 
+Test one scenario per test. This makes reading and debugging tests easier plus it keeps the test name simple (remember that the last part of the test function name describes the scenario you are testing).
+
+Avoid having logic in tests. If you need to make your assertions conditional, that's a good indicator that the test should be split.
```
drewvanstone, line 152: +1 

Plus, logic in the tests means that that functionality is not available to end users. If you need some logic for validation, or to do some set up, we should consider whether we can add that logic to the product itself.

drewvanstone, line 152: +1 

Plus, logic in the tests means that that functionality is not available to end users. If you need some logic for validation, or to do some set up, we should consider whether we can add that logic to the product itself.

vivek-koppuru (approved):

PR #2297, opened 2022-05-31
Hardware CSV file reading is only needed for create flow

ptrivedi:
Delete flow does not require or take a hardware csv file. So the reading of CSV file should only be done for the create flow.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


abhinavmpandey08 (approved):
/approve

PR #2299, opened 2022-06-01
Add assertion for hardware CSV disk consistency.

chrisdoherty4:
The new assertion ensures all Disk values within a set of CSV entries match. A 'set' of entries is defined as all entries matching a given selector where selectors are specified on TinkerbellMachineConfig's.

This is necessary because there is a constraint in Tinkerbell's current design where 1 TinkerbellTemplate corresponds to a control plane, etcd or worker node group. Any Hardware used within those groupings uses the same Template so must have the same disk for provisioning.


pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -231,3 +232,67 @@ func validateLabelValue(v string) error {
...
+
+// MachineSelector is a map of key-value pairs used to select Machine instances by label definition.
+type MachineSelector map[string]string
+
+func (m MachineSelector) String() string {
+	s, err := json.Marshal(m)
+	// We want to support fmt.Stringer so we have no choice but to panic here. In practice, this
+	// should never happen because we're merely stringifying a map.
+	if err != nil {
+		panic(err)
```
ptrivedi, line 244: shouldn't panic here, use Errorf

pkg/providers/tinkerbell/create.go
```diff
@@ -133,6 +133,10 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 	writer := hardware.NewMachineCatalogueWriter(p.catalogue)
 	machineValidator := hardware.NewDefaultMachineValidator()
 
+	// TODO(chrisdoherty4) Build the selectors slice using the selectors from TinkerbellMachineConfig's
+	var selectors []hardware.MachineSelector
+	machineValidator.Register(hardware.MatchingDisksForSelectors(selectors))
```
ptrivedi, line 138: check err return here

pkg/providers/tinkerbell/hardware/validator.go
chrisdoherty4: What do we do with the error returned from `fmt.Errorf()`?

pkg/providers/tinkerbell/create.go
chrisdoherty4: Sorry, I'm not tracking. What error needs checking?

ptrivedi: oh, i had thought machine validator's Register does have an error return, but i see that it doesn't

pkg/providers/tinkerbell/hardware/validator.go
ptrivedi: We have generally tried to avoid using panic. Not sure if it is spelled out in the code guidelines but panic should be used sparingly and reserved for programmer errors. I agree that it could be a gray area here, but bubbling an error back up has been the general pattern and would be good for consistency and to avoid blurring boundaries and abusing it by more casual use

PR #2300, opened 2022-06-01
Add Tinkerbell label selector and configure disk type in the template…

panktishah26:
*Description of changes:*
In this PR, I have added Tinkerbell label selector in the cluster config file and configured disk type in the template config field.


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -613,6 +613,10 @@ type Ref struct {
 	Name string `json:"name,omitempty"`
 }
 
+type HardwareSelector struct {
```
chrisdoherty4, line 616: Selectors need to be `map[string]string` to allow arbitrary key-value pairs.

chrisdoherty4 (changes requested):
Principally, this is on the right track - just needs a few adjustments.

We need selectors to be completely definable by the user. The user should control, entirely, what key-value pairs are used in Hardware CSV labels (mainline already has this), _and_ what key-value pairs are used in selectors to match those labels.

Having done that, for each selector we cache the first observed Disk from a matching Machine instance. An example of a selector-label matching algorithm can be found in the [PR I just raised](https://github.com/aws/eks-anywhere/pull/2299/files#diff-05d6aded64a4a29b5fd8c21346546229256a812aa21b9656d1ef238c7ad2865dR290) and we should probably use the same function.

Having cached the disks we can pass a construct with access to that cache to the template builder and _provide lookup behavior_. It is extremely important that we provide _behavior_ (methods) and don't wangle inner fields because we start to violate the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -613,6 +613,8 @@ type Ref struct {
 	Name string `json:"name,omitempty"`
 }
 
+type HardwareSelector map[string]string
+
```
abhinavmpandey08, line 617: Can you define this type in `tinkerbellmachineconfig_types.go` instead?

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -29,39 +31,39 @@ warnings:
...
-func withNetplanAction(b v1alpha1.VersionsBundle) ActionOpt {
+func withNetplanAction(b v1alpha1.VersionsBundle, diskType string) ActionOpt {
 	return func(a *[]tinkerbell.Action) {
 		*a = append(*a, tinkerbell.Action{
 			Name:    "write-netplan",
 			Image:   b.Tinkerbell.TinkerbellStack.Actions.WriteFile.URI,
 			Timeout: 90,
 			Environment: map[string]string{
-				"DEST_DISK": "/dev/sda2",
+				"DEST_DISK": fmt.Sprintf("%s2", diskType),
```
abhinavmpandey08, line 66: We are just appending `2` to `diskType`. Do we account for whether `diskType` is `sda` or `nvme`?
Because for `sda` you can just append 2 but with nvme you need `p2`

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 func TestWithDefaultActionsFromBundle(t *testing.T) {
 	vBundle := givenVersionBundle()
+	diskType := "/dev/sda"
```
abhinavmpandey08, line 13: Should add a test case with nvme as well

pkg/providers/tinkerbell/config/template-cp.yaml
```diff
@@ -178,6 +178,12 @@ spec:
   template:
     {{- if and .etcdTemplateOverride (ne .format "")}}
     spec:
+      hardwareAffinity:
+        required:
+        - labelSelector:
+            matchLabels: {{ range $key, $value := .etcdHardwareSelector}}
+              {{ $key }}: {{ $value}}
+            {{- end }}
```
abhinavmpandey08, line 186: Have you checked what happens here if the `etcdHardwareSelector` is empty?

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
chrisdoherty4: We talked about dealing with NVMe on a separate PR so we can unblock mainline.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults_test.go
chrisdoherty4: https://github.com/aws/eks-anywhere/pull/2300#discussion_r889709080

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -87,3 +87,9 @@ func WithTemplateRef(ref ProviderRefAccessor) TinkerbellMachineConfigGenerateOpt
 		}
 	}
 }
+
+func WithHardwareLabel() TinkerbellMachineConfigGenerateOpt {
```
chrisdoherty4, line 91: Map initialization shoudln't be optional. Could we move this to the `New*`.

pkg/api/v1alpha1/tinkerbelltemplateconfig.go
```diff
@@ -18,7 +18,7 @@ const TinkerbellTemplateConfigKind = "TinkerbellTemplateConfig"
 type ActionOpt func(action *[]tinkerbell.Action)
 
 // NewDefaultTinkerbellTemplateConfigCreate returns a default TinkerbellTemplateConfig with the required Tasks and Actions
-func NewDefaultTinkerbellTemplateConfigCreate(name string, versionBundle v1alpha1.VersionsBundle) *TinkerbellTemplateConfig {
+func NewDefaultTinkerbellTemplateConfigCreate(name string, versionBundle v1alpha1.VersionsBundle, diskType string) *TinkerbellTemplateConfig {
```
chrisdoherty4, line 21: Could we change to using `disk` or `diskPath`. We aren't really describing a type of disk. Also a code comment everywhere this is accepted, given its just a string, would go a long way so we establish expectations.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -74,14 +76,14 @@ func withNetplanAction(b v1alpha1.VersionsBundle) ActionOpt {
...
-func withTinkCloudInitAction(b v1alpha1.VersionsBundle) ActionOpt {
+func withTinkCloudInitAction(b v1alpha1.VersionsBundle, diskType string) ActionOpt {
 	return func(a *[]tinkerbell.Action) {
 		*a = append(*a, tinkerbell.Action{
 			Name:    "add-tink-cloud-init-config",
 			Image:   b.Tinkerbell.TinkerbellStack.Actions.WriteFile.URI,
 			Timeout: 90,
 			Environment: map[string]string{
-				"DEST_DISK": "/dev/sda2",
+				"DEST_DISK": fmt.Sprintf("%s2", diskType),
```
chrisdoherty4, line 86: The partition number shouldn't be the responsibility of individual actions to define. They should just accept a destination disk, in full. 

The partition number, for now, could be defined by the templating construct perhaps? We could modify this in a subsequent PR as we explore supporting multiple disk types (NVMe etc).

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -1,21 +1,67 @@
...
+
+	tinkv1alpha1 "github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	eksav1alpha1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 )
 
+type DiskExtractor struct {
```
chrisdoherty4, line 14: Can we get code documentation on _all_ exported types at a minimum. The docs should explain what the scope of responsibility is and any behavior expectations.

pkg/providers/tinkerbell/create.go
```diff
@@ -118,7 +118,14 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		validator.Register(NewIPNotInUseAssertion(p.netClient))
 	}
 
-	writer := hardware.NewMachineCatalogueWriter(p.catalogue)
+	catalogueWriter := hardware.NewMachineCatalogueWriter(p.catalogue)
+
+	for _, mc := range p.machineConfigs {
+		p.templateBuilder.diskExtractor.RegisterHardwareSelector(mc.Spec.HardwareSelector)
```
chrisdoherty4, line 124: We shouldn't be reaching into other constructs and modifying their members (Law of Demeter). 

Its not a great solution, but for now can we expose a `RegisterHardwareSelector` on the `TemplateBuilder` so we can call `p.templateBuilder.RegisterHardwareSelector()`. This, at least, maintains a degree of separation that will compliment any `TemplateBuilder` dependency injection we refactor for in the future.

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
chrisdoherty4, line 14: The `DiskExtractor` needs a suite of unit tests, positive and negative.

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -1,21 +1,67 @@
...
+	for key, val := range hardwareSelector {
+		selectorKey[key] = val
+	}
+	return selectorKey.String()
+}
+
+func (d *DiskExtractor) GetDisk(hardwareSelector eksav1alpha1.HardwareSelector) (string, error) {
+	key := serializedHardwareSelector(hardwareSelector)
+	if _, ok := d.Disks[key]; !ok {
+		return "", fmt.Errorf("hardware selector does not match with any given hardware machines %v", hardwareSelector)
```
chrisdoherty4, line 52: Use colon's in errors to separate added context from original error.

```suggestion
		return "", fmt.Errorf("hardware selector does not match with any given hardware machines: %v", hardwareSelector)
```

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -1,21 +1,67 @@
...
+	}
+	return d.Disks[key], nil
```
chrisdoherty4, line 54: Use colon's to separate context from original error. The disk variable just avoids unnecessary hashing in the map which, while negligible, has a cost and is easily avoided.

```suggestion
    disk, ok := d.Disks[key]
	if !ok {
		return "", fmt.Errorf("hardware selector does not match with any given hardware machines: %v", hardwareSelector)
	}
	return disk, nil
```

chrisdoherty4 (changes requested):
Great work - just some loose ends to tidy up.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
panktishah26: Yes, I am going to create a separate PR for this.

pkg/providers/tinkerbell/config/template-cp.yaml
panktishah26: I have added a basic validation where HardwareSelector is mandatory for each machine config for Tinkerbell.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -29,39 +31,39 @@ warnings:
...
-func withNetplanAction(b v1alpha1.VersionsBundle) ActionOpt {
+func withNetplanAction(b v1alpha1.VersionsBundle, disk string) ActionOpt {
 	return func(a *[]tinkerbell.Action) {
 		*a = append(*a, tinkerbell.Action{
 			Name:    "write-netplan",
 			Image:   b.Tinkerbell.TinkerbellStack.Actions.WriteFile.URI,
 			Timeout: 90,
 			Environment: map[string]string{
-				"DEST_DISK": "/dev/sda2",
+				"DEST_DISK": fmt.Sprintf("%s2", disk),
```
micahhausler, line 66: We don't know this will be the second disk 100% of the time

abhinavmpandey08: This is the 2nd disk partition not the 2nd disk. The ubuntu image we build has UEFI is the primary partition and the actual os on the 2nd partition. That's why we are appending 2 to the disk name

pkg/providers/tinkerbell/create.go
```diff
@@ -117,7 +117,10 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		validator.Register(NewIPNotInUseAssertion(p.netClient))
 	}
 
-	writer := hardware.NewMachineCatalogueWriter(p.catalogue)
+	catalogueWriter := hardware.NewMachineCatalogueWriter(p.catalogue)
+
+	writer := hardware.MultiMachineWriter(catalogueWriter, p.templateBuilder.diskExtractor)
```
chrisdoherty4, line 122: This should reference the extractor from `p` not the template builder.

chrisdoherty4 (approved):

abhinavmpandey08 (approved):
/lgtm

PR #2307, opened 2022-06-01
Make all e2e tests default to 1:1:1 topology

g-gaston:
This is a temporary fix to avoid overloading the CI system.
It should be removed eventually or added explicitly to all tests.


jaxesn (approved):
/approve

PR #2308, opened 2022-06-01
Software development lifecycle doc

g-gaston:
Small doc to codify some of our current practices/processes from a more e2e point of view. Ideally this should help new folks get on-boarded. As any other dev doc, this one is also supposed to be a living doc, evolving as we learn and progress.

### DIFF
--- a/docs/developer/development-lifecycle.md
+++ b/docs/developer/development-lifecycle.md
@@ -1,0 +1,33 @@
+# Development lifecycle
+
+This documents aims to codify our development process for the most common scenarios from the moment you are assigned to a task (generally tied to a GitHub issue, whether it's a new feature, improvement or bugfix) to the moment when you can call the task *done*.
+
+## Process
+
+1. Read the GitHub issue and try to answer the following questions:
+   1. What problem are we solving? This should be clear most of the time but if it's not, you might need to talk to the author of the issue and/or the stakeholders.
+   1. Do we need to solve this problem and if so, why? If a feature advanced all the way to a GitHub issue, the answer most of the time will be yes. But don't fear to question the requirements. It's important to understand not only what we are building but why, or we might end up implementing the wrong thing. If you don't think this problem needs to be solved, it's already been solved or there a better way, this is the moment to voice it out.
+   2. What are we building? Sometimes the issue will be prescriptive about the solution to the problem, sometimes it will just give you functional requirements and other times you will only have a description of the problem. Make sure you understand all of this and collect all the requirements before moving forward with the design. This will sometimes involve talking to other engineers, product or even users. If you disagree with any given requirement, challenge them.
+1. If the task needs a design, work on it before starting the implementation.
+   * Not all features need a design doc, but most will need a design. Study the problem space and existing code and come up with a solution before implementing it.
+   * When possible, favor getting eyes on your ideas/design/doc as soon as possible. Early feedback avoids rework and will set you up for a speedy review once you write the code. Not everything needs a formal design review on a PR. Just ask folks for a chat to bounce ideas, especially folks with special experience or interest in the are you are working on. This is important when introducing new ideas or patterns.
+   * If you are basing your design on one or more ideas you are not sure will work, feel free to play with a POC. The goal of designing the solution before implementing is to avoid rework and get early feedback, code is as valid as any other tool to come up with that design. That said, most features won't need a POC.
+   * Timebox this, specially for big features. If it's taking too long (relative to the size of the task) you might be trying to go too deep. In that case, ask someone to pair on the design, it's usually easier to stay focused on what is needed when you can keep each other accountable.
+   * Favor results over perfection, done is better than perfect. Look for two way doors and if a decision can be postponed without adding risk, leave it for later. At some point, you need to say "this is enough" and move forward with the implementation, even if that means making mistakes and having to solve them along the way.
+   * Make sure you include testing in your design. When a feature can benefit from E2E tests, it can be useful to enumerate the scenarios that need to be tested.
+1. If needed, split the task into manageable chunks. Feel free to create new GitHub issues. You can track them in the original one with a checklist or you can even create a GitHub project (specially useful for big features, even more so when they span across multiple releases). This helps to keep track of the progress and facilitates adding more people to the feature if needed.
+   * When the task becomes a project, try to think into parallelizable work streams, specially when you know you'll be working with more people.
+   * Make sure you account for documentation when needed.
+   * Follow [this](issues.md) to categorize your issues.
+1. Come up with an estimate of how long the task will take. You won't always need to commit to a date or even communicate it externally, but it helps keeping track of the progress and keeps us accountable.
+1. Implement your design. Reading our [development](best-practice.md) and [PR best practices](pull-request.md) is a must.
+   * Keep the team informed of your progress. If you get stuck, ask for help.
+   * If you think the delivery on time of the task is in risk, bring it up to the team. In this kind of situation, you can generally compromise on time, quality, scope or resources, even though the latter tends to work very rarely. Make sure the team feels comfortable with the compromise (or combination of them) and plan to account for the consequences (communicating a delay to external stakeholders, addressing tech debt, communicating a scope reduction, etc.).
+
+## When is a task *done*?
+
+* The solution has been implemented, it solves the original issue and follows all requirements.
+* All PRs have been merged.
+* All new code in properly tested.
+* If E2E tests were necessary, they are merged and running in CI (ideally without flakes).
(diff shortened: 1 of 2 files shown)


docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   * Timebox this, specially for big features. If it's taking too long (relative to the size of the task) you might be trying to go too deep. In that case, ask someone to pair on the design, it's usually easier to stay focused on what is needed when you can keep each other accountable.
+   * Favor speed over perfection. Look for two way doors and if a decision can be postponed without adding risk, leave it for later.
+   * Make sure you include testing in your design. When a feature can benefit from E2E tests, it can be useful to enumerate the scenarios that need to be tested.
+1. If needed, split the task into manageable chunks. Feel free to create new GitHub issues. You can track them in the original one with a checklist or you can even create a GitHub project (specially useful for big features, even more so when they span across multiple releases). This helps to keep track of the progress and facilitates adding more people to the feature if needed.
```
markapruett, line 18: What's our preferred method for delivering incrementally? Might be worth a call out to prefer smaller scoped PRs delivered over time with feature flags versus giant ones that implement everything in one go and are hard to rollback.

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
+# Development lifecycle
+
+This documents aims to codify our development process for the most common scenarios from the moment you are assigned to a task (generally tied to a GitHub issue, wether it's a new feature, improvement or bugfix) to the moment when you can call the task *done*.
```
vivek-koppuru, line 3: Nit
```suggestion
This documents aims to codify our development process for the most common scenarios from the moment you are assigned to a task (generally tied to a GitHub issue, whether it's a new feature, improvement or bugfix) to the moment when you can call the task *done*.
```

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   1. Do we need to solve this problem and if so, why? If a feature advanced all the way to a GitHub issue, the answer most of the time will be yes. But don't fear to question the requirements. It's important to understand not only what we are building but why, or we might end up implementing the wrong thing. If you don't think this problem needs to be solved, it's already been solved or there a better way, this is the moment to voice it out.
+   2. What are we building? Sometimes the issue will be prescriptive about the solution to the problem, sometimes it will just give you functional requirements and other times you will only have a description of the problem. Make sure you understand all of this and collect all the requirements before moving forward with the design. This will sometimes involve talking to other engineers, product or even users. If you disagree with any given requirement, challenge them.
+1. If the task needs a design, work on it before starting the implementation.
+   * Not all features need a design doc, but most will need a design. Study problem space and the existing code and come up with a solution before implementing it.
```
vivek-koppuru, line 12: Nit
```suggestion
   * Not all features need a design doc, but most will need a design. Study the problem space and existing code and come up with a solution before implementing it.
```

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   * When possible, favor getting eyes on your ideas/design/doc as soon as possible. Early feedback avoids rework and will set you up for a speedy review once you write the code. Not everything needs a formal design review on a PR, just ask folks for a chat to bounce ideas, specially folks with special experience or interest in the are you are working on. This is specially important when introducing new ideas or patterns.
```
vivek-koppuru, line 13:
```suggestion
   * When possible, favor getting eyes on your ideas/design/doc as soon as possible. Early feedback avoids rework and will set you up for a speedy review once you write the code. Not everything needs a formal design review on a PR. Just ask folks for a chat to bounce ideas, especially folks with special experience or interest in the are you are working on. This is important when introducing new ideas or patterns.
```

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   * Paper sustains everything so if you think you need to prove an idea, feel free to play with a POC. The goal of designing the solution before implementing is to avoid rework and get early feedback, code is as valid as any other tool to come up with that design. That said, most features won't need a POC.
```
vivek-koppuru, line 14: What do you mean here by "paper sustains everything"? Maybe I am unfamiliar with that phrase, but I like how in the previous bullets you started off direct and then elaborated, so maybe something like "Extend your design with a POC if needed" or something

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   * Timebox this, specially for big features. If it's taking too long (relative to the size of the task) you might be trying to go too deep. In that case, ask someone to pair on the design, it's usually easier to stay focused on what is needed when you can keep each other accountable.
+   * Favor speed over perfection. Look for two way doors and if a decision can be postponed without adding risk, leave it for later.
```
vivek-koppuru, line 16: I would say `efficiency` instead of `speed`

docs/developer/development-lifecycle.md
```diff
@@ -0,0 +1,33 @@
...
+   * Make sure you include testing in your design. When a feature can benefit from E2E tests, it can be useful to enumerate the scenarios that need to be tested.
+1. If needed, split the task into manageable chunks. Feel free to create new GitHub issues. You can track them in the original one with a checklist or you can even create a GitHub project (specially useful for big features, even more so when they span across multiple releases). This helps to keep track of the progress and facilitates adding more people to the feature if needed.
+   * When the task becomes a project, try to think into parallelizable work streams, specially when you know you'll be working with more people.
+   * Make sure you account for documentation when needed.
+   * Follow [this](issues.md) to categorize your issues.
+1. Come up with an length estimation. You won't always need to commit to it or even communicate it externally, but it helps keeping track of the progress and keeps us accountable.
```
vivek-koppuru, line 22: Nit
```suggestion
1. Come up with a duration estimate. You won't always need to commit to it or even communicate it externally, but it helps keeping track of the progress and keeps us accountable.
```

g-gaston: Yeah, I struggled phrasing this. I wanted to make the distinction between estimating how long something is gonna take a giving a date. I think we should always estimate the duration but not necessarily have date, unless we need one.

If you think "duration estimate" complies with that, I'm good with it 🙂

docs/developer/development-lifecycle.md
g-gaston: Umm yeah probably not a very straight forward sentence.
I was trying to say that a theoretical design can hold almost any idea but, most of the time it's only when you get to the real world implementation, when you find the limitations and constraints of what works and what doesn't.

What about something like this:
```suggestion
   * If you are basing your design on one or more ideas you are not sure will work, feel free to play with a POC. The goal of designing the solution before implementing is to avoid rework and get early feedback, code is as valid as any other tool to come up with that design. That said, most features won't need a POC.
```

docs/developer/development-lifecycle.md
g-gaston: I think I see what you are saying. However, I was trying to use two opposites in the spectrum so when in doubt, folks have a rule of thumb to decide what to do.

Maybe I'm getting too much into semantics, but I don't think efficiency and perfection are opposites, in fact, you can be very efficient while looking for perfection. If you are, it will take you the least amount of time to get to it (since you are being efficient), but it will still take you a long time compared to a not perfect solution.

This bullet point was about making sure we don't get stuck on the design phase, trying to come up with something completely round and perfect. That's not sustainable and, at some point, you need to say "this is enough" and move forward with the implementation, even if that means making mistakes and having to solve them along the way.

Maybe there is a way we can rephrase this to make it more clear?

docs/developer/development-lifecycle.md
vivek-koppuru: A lot better!

docs/developer/development-lifecycle.md
vivek-koppuru: The issue I have with this phrase is `speed` generally to me feels like we are okay with being careless, which maybe is because I am associating it a negative way myself. Would `progress` or `progression` be better, to show that you want to favor moving forward and being comfortable with the consequences rather than being stuck?

docs/developer/development-lifecycle.md
vivek-koppuru: I am not super happy with my suggestion either, I am wondering if "an estimate of how long the task will take" is more clear.

g-gaston:
```suggestion
1. Come up with an estimate of how long the task will take. You won't always need to commit to a date or even communicate it externally, but it helps keeping track of the progress and keeps us accountable.
```

vivek-koppuru (approved):

PR #2309, opened 2022-06-01
Update etcdadm controller references to use AWS org

abhay-krishna:
After aws/eks-anywhere-build-tooling#860 merged, the artifacts for the etcdadm repos will reference the AWS org in the URI instead of `mrajashree`. This PR updates the release tooling to fetch the appropriate artifacts from the new S3/ECR locations.


release/pkg/utils/utils.go
```diff
@@ -27,6 +27,7 @@ import (
 const hexAlphabet = "0123456789abcdef"
 
 func ExecCommand(cmd *exec.Cmd) (string, error) {
+	fmt.Println(cmd.String())
```
ewollesen, line 30: Should this be written via a log statement or behind a debug flag or something? Or did you mean to leave this in just like this?

ewollesen (commented):
Just had the one comment/question. I can approve as-is if desired.

release/pkg/utils/utils.go
abhay-krishna: Oh my bad this was used for testing. I'll remove it.

ewollesen (approved):
/lgtm

Issue #2212, opened 2022-05-19
Create CAPI machines under the EKSA's Cluster's namesapce


wongni:
**What would you like to be added**:
Currently, CAPI machines (k get machines -A) are created under `eksa-system` namespace. I want them to be under the namespace specified in the EKSA Cluster.

For example, 

```
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: Cluster
metadata:
  name: a-cluster
  namespace: cluster-ns
```

In this case, CAPI machines are created uder `cluster-ns` instead of `eksa-system`

**Why is this needed**:

I want to partition the clusters in the long-lived management cluster.

maxdrib:
Adding to this - in a scenario where a management cluster has hundreds of thousands of workload clusters, we may run into scaling issues if all these capi/capc resources are in the eksa-system namespace. However, it sounds like breaking cluster resources into separate namespaces may have its own tradeoffs as well, since many operations in eks-a rely on the hardcoded eksa-system namespace and can add complexity, like `clusterctl move`

PR #2311, opened 2022-06-01
Design for creating/managing underlying CAPI resources in custom namespace

maxdrib:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/2212

*Description of changes:*
Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace 

This document aims to solve that, by allowing customers to specify a new flag on the cluster spec. If set, the underlying CAPI/CAPX/etcd resources will be created under the Cluster's namespace, rather than all dumped into `eksa-system`. 
This change would make it so customers would be able to put these resources virtually in any namespace, despite the fact that they may be considered eks-a internal components.

*Testing (if applicable):*


designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
...
+
+**Future scope**
+* N/A
+ 
+## Current state
+1. Top level eks-a resources (e.g. Cluster, *DatacenterConfig, *MachineConfig) all go into the "default" namespace by default but can be customized to any namespace. Ideally, customers should only ever interact with these top-level eks-a resources
+2. Underlying CAPI/CAPX and etcd resources are created in the eksa-system, to be hidden from customers as implementation details
+
+## Overview of Solution
+1. Introduce a new boolean flag in the EKS-A Cluster CRD spec which allows users to indicate whether they would like the underlying resources to be created in the same namespace as the eks-a Cluster resource
```
vivek-koppuru, line 34: Instead of a boolean flag, what about introducing a new field in the cluster spec, like `systemNamespace` or something? Is the purpose of this that we just want some override functionality? Our controller would also need to know where to pull the resources, hence having it in the cluster spec would be easier.

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Implementation
+
+Currently, the `eksa-system` namespace is hardcoded throughout the entire codebase for underlying. It's used in
+1. providers' template generation logic, both by the CLI and the EKS-A controller
+2. kubectl and clusterctl executables for interacting with the cluster and moving resources between clusters
+3. ???
```
vivek-koppuru, line 41: The controller uses it to retrieve the capi resources as well.

vivek-koppuru (commented):
The idea is that we don't want users to be managing these resources, so what is the motivation of not having it under `eksa-system`? Do we anticipate them wanting to make changes to that somehow?

chrisdoherty4 (commented):
Our software was never designed, from what I've seen, to work across namespaces resulting in many reference types not including namespace information. I'd be interested in an analysis of inter resource referencing to see if/how any of that is impacted.

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
+# CAPI Resources in Custom Namespace
+
+## Introduction
+
+**Problem:** Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace
```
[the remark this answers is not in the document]
maxdrib: Good question - will follow up with the customer to go more in depth

designs/capi-resources-custom-namespace.md
maxdrib: Thank you

g-gaston (commented):
Awesome! Thanks a lot for writing this, love seeing design docs ❤️

designs/capi-resources-custom-namespace.md
danbudris, line 5: curious if we've had a specific user with an experience that indicates this feature is needed and what their experience has been

danbudris, line 5: also, I think that there's an important distinction between technical limitations and facilitating an organization approach by customers; I really like this idea from the organizational approach but don't see technical limitations.

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
+# CAPI Resources in Custom Namespace
+
+## Introduction
+
+**Problem:** Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace 
+
+This document aims to solve that, by allowing customers to specify a new flag on the cluster spec. If set, the underlying CAPI/CAPX/etcd resources will be created under the Cluster's namespace, rather than all dumped into `eksa-system`.
```
danbudris, line 7: so, in this proposal, the user would set a bool flag and EKS A would then create the objects in a namespace named for the cluster, is that right? what if we have multiple workload cluster with the same name (not sure if that's supported anyway but just wondering)?

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
+# CAPI Resources in Custom Namespace
+
+## Introduction
+
+**Problem:** Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace 
+
+This document aims to solve that, by allowing customers to specify a new flag on the cluster spec. If set, the underlying CAPI/CAPX/etcd resources will be created under the Cluster's namespace, rather than all dumped into `eksa-system`. 
+This change would make it so customers would be able to put these resources virtually in any namespace, despite the fact that they may be considered eks-a internal components.
```
danbudris, line 8: should we prevent customers from using certain namespaces? For example, `kube-system` and `default` come to mind. `default` can be subject to accidental exposure and `kube-system` exposes a path from our components to critical system components that would be better avoided.

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Tenets
+
+****Simple:**** make everything simple and secure by default, but give options to change those defaults if there is a need
```
danbudris, line 12: maybe "secure by default" is a separate tenant

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Goals and Objectives
+
+As a EKS-A cluster administrator I want to organize/partition each cluster into a dedicated namespace so that I can more easily manage my workload clusters
```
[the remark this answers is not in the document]
danbudris: yeah I would love to hear from experienced cluster operators or folks using CAPI about the strategies they currently use for organizing their systems

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,54 @@
...
+
+**Not in scope**
+* Support for customizing which resources are going into which namespace at different layers (e.g. CAPC resources into ns1, CAPI resources into ns2, eks-a resources into ns3, etc.)
+
+**Future scope**
+* N/A
+ 
+## Current state
+1. Top level eks-a resources (e.g. Cluster, *DatacenterConfig, *MachineConfig) all go into the "default" namespace by default but can be customized to any namespace. Ideally, customers should only ever interact with these top-level eks-a resources
+2. Underlying CAPI/CAPX and etcd resources are created in the eksa-system, to be hidden from customers as implementation details
```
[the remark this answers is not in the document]
danbudris: no that's the explanation I would give. 

partitioning the system by namespace facilitates the hiding of implementation details, organization of the components, and providing control over who/what can access the sensitive components of the system, especially as @g-gaston said, the secrets which contain infrastructure credentials and tokens etc used by the CAPX components.

designs/capi-resources-custom-namespace.md
danbudris: as a cluster administrator I'd like to be able to specify exactly which namespace the components go in (and have you stop me if I try to put it in default or kube-system)

danbudris (commented):
I love this idea, seems like it'd make administering a management cluster with a large number of workload clusters much easier and intuitive; and it'd allow a top-level admin to divide the responsibility for sub-cluster mgmt across teams via rbac rules on the individual namespaces without exposing other clusters.

Would love to see a 'security' section with anything you see as relevent and ideally discusses a couple things:
- access control for the secrets in the namespaces
- safeguards on namespace selection (no kube-system, no default namespaces if we provide option to pick specific namespace)

designs/capi-resources-custom-namespace.md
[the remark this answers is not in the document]
maxdrib: @g-gaston I updated the doc to reflect the two specific problems we have uncovered, please take another look.

maxdrib: The main technical limitation I could anticipate is in the clusterctl move command during a cluster upgrade. I ran an experiment with 3 workload clusters (1 etcd, 1 CP, 1 MD) on a management cluster (3 etcd, 4 CP, 3 MD) and observed that 188 resources were moved. Each resource is probably only a few KB, but if we increase that by one or two orders of magnitude, I am concerned about the API load on a bootstrap cluster running in a smaller machine somewhere. Does that seem valid to you? @danbudris

designs/capi-resources-custom-namespace.md
maxdrib: I think we'd run into a problem generating the filesystem folder to store the artifacts for that, right? It seems like we should prevent users from having clusters of the same name, even if they are in different namespaces. It would be an easy preflight check, if that makes sense to you? @danbudris

designs/capi-resources-custom-namespace.md
maxdrib: Good call! We can definitely add a denylist for these namespaces and validate that on cluster creation.

designs/capi-resources-custom-namespace.md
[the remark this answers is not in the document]
maxdrib: They want to have visibility into the underlying resources for troubleshooting purposes

designs/capi-resources-custom-namespace.md
maxdrib: Will add this user story, thanks folks.

designs/capi-resources-custom-namespace.md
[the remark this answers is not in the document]
maxdrib: Rephrased this note, thanks for the explanation.

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,63 @@
+# CAPI Resources in Custom Namespace
+
+## Introduction
+
+**Problem:** Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace. Specifically, they expect to encounter two problems:
+
+1. Difficulty troubleshooting issues for a single cluster when there are so many resources to sift through. 
+2. High levels of load on a bootstrap cluster when executing an upgrade, since the clusterctl move operation will take all the CAPI components inside a namespace and instatiate them on the bootstrap cluster. This can result in heavy and unnecessary load on the bootstrap cluster
```
chrisdoherty4, line 8: @maxdrib If I'm not mistaken, this problem is restricted to upgrades of the management cluster only?

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,63 @@
...
+
+This document aims to solve that, by allowing customers to instantiate the underlying CAPI/CAPX/etcd resources in a newly defined namespace, rather than all being dumped into `eksa-system`. 
+This change would make it so customers would be able to put these resources virtually in any namespace (except a few denylisted ones, discussed below), despite the fact that they may be considered eks-a internal components.
```
chrisdoherty4, line 11: @maxdrib I appreciate the PR title (+ parts of the discussion) suggests CAPX only but isn't this something that should be consistent among providers?

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Tenets
+
+* ****Secure By Default:**** make everything simple and secure by default, but give options to change those defaults if there is a need
+
+### Goals and Objectives
+
+* As a EKS-A cluster administrator I want to quickly and easily query resources for a given cluster so that I can more easily troubleshooting cluster issues with many workload clusters present
+* As an EKS-A cluster administrator, I want to be able to perform cluster upgrades without having to move all the cluster resources onto the bootstrap cluster
```
maxdrib, line 20: It sounds like this user story is controversial - according to Joey, we should continue to move workload cluster capi resources to the bootstrap cluster. Thats the safety mechanism — in case the mgmt cluster itself goes into unrecoverable state, at least we have all the capi resources in the bootstrap cluster

designs/capi-resources-custom-namespace.md
maxdrib: I was using the term CAPX to capture all the different CAPI providers (CAPC, CAPV, etc.) - please let me know if there's a better one I could use instead

designs/capi-resources-custom-namespace.md
maxdrib: Yep! That's the only scenario. where we would be moving all the workload+management clusters' resources onto a bootstrap cluster

designs/capi-resources-custom-namespace.md
chrisdoherty4: I see. Perhaps drop CAPX (I keep thinking its a specific provider of some sort).

designs/capi-resources-custom-namespace.md
chrisdoherty4: How does namespaces help with this? Does CAPI already provide the ability to move only resources for a specific namespace?

designs/capi-resources-custom-namespace.md
```diff
@@ -0,0 +1,63 @@
+# CAPI Resources in Custom Namespace
+
+## Introduction
+
+**Problem:** Customers managing large scale EKS-A clusters (hundreds or thousands of workload clusters for a given management cluster) expect to have difficulty managing all these resources if they are all in the same namespace. Specifically, they expect to encounter two problems:
+
+1. Difficulty troubleshooting issues for a single cluster when there are so many resources to sift through.
```
chrisdoherty4, line 7: This problem could also be resolved with labeling and selectors.

designs/capi-resources-custom-namespace.md
chrisdoherty4: This relates to another comment I made: I'm not sure how namespaces solves the problem of upgrading a management cluster.

maxdrib: See response above - clusterctl move only moves resources in a single namespace, currently hardcoded to eksa-system.

maxdrib:
Closing again because of the risk that CAPI resources will be left on the management cluster while it upgrades, which could make them irrecoverable.

PR #2312, opened 2022-06-01
Git Flux tests create their own repository for use during test

danbudris:
*Issue #, if available:*

*Description of changes:*
The Git Flux tests have been using the same git repository, pushing and pulling their flux configs and cluster configs at the same time. These concurrent git actions have caused frequent failures in the validations. Specifically, multiple simultaneous 'local' commits by validation code led to divergent histories among the clones on the test instances, causing the validation code to fail when attempting to push and pull from the remote and encountering non-fast-forward updates. This was not resolved by setting the default behavior to `rebase`, and it seems that the `pull` behavior of `go-git`, the library we are using for git actions, can't handle non-fast-forward updates even when the given branch is configured to rebase instead of merge. 

So, now, instead of using the same repository for each of the tests, we're instead going to imitate what the existing Github tests do anyway: create a github repository for use by each test.

So, here, when running through the framework, the tests for FluxGit now accept the github token and username, and use that to create a github repo (and add a generated deploy key to it) during test setup. This repo is then used by the test itself.

When running outside of the framework you will still provide an existing repository. This makes it easy for folks to run the tests individually without additional setup, and without requiring that the setup logic for the repository be included in the tests themselves -- instead, it's confined to the setup code in the framework.

Additionally, I've adjusted the behavior of the branches we use to default to `rebase` instead of `merge`, and added a few `pulls` before `push` to the test code in order to try and eliminate other cases where simultaneous modification of the repository could interfere and make the behavior more predictable.

Follow-on is going to be clean up for the git repos in the test framework; we're currently leaving them behaind and need to add cleanup code. This is in progress but wanted to get this PR open and reviewed. 

*Testing (if applicable):*

run the tests via the framework from my local machine; confirmed that the correct repository with the right deploy key was created and that the tests ran successfully.

### DIFF
--- a/internal/test/e2e/crypto.go
+++ b/internal/test/e2e/crypto.go
@@ -34,8 +34,8 @@ func pemFromPrivateKeyEcdsa(k *ecdsa.PrivateKey) ([]byte, error) {
 	}
 
 	p := pem.Block{
-		Type:    "PRIVATE KEY",
-		Bytes:   pk,
+		Type:  "PRIVATE KEY",
+		Bytes: pk,
 	}
 
 	return pem.EncodeToMemory(&p), nil
--- a/internal/test/e2e/crypto.go
+++ b/internal/test/e2e/crypto.go
@@ -1,0 +1,46 @@
+package e2e
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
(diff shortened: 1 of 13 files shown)


jaxesn (approved):
Lets give it a test!

/approve

PR #2315, opened 2022-06-02
Iso

wanyufe:
*Issue #, if available:*
cloudstack - add ISO attachment feature to cloudstack machine template configuration.
*Description of changes:*
A specified ISO cdrom will be attached to VM and mounted.
Scripts preKubeadmCommand.sh and postKubeadmCommand.sh in mount path root can be enabled to run with specified parameters.
*Testing (if applicable):*
Unit testing.


go.mod
```diff
@@ -160,6 +160,7 @@ require (
 
 // TODO: Once the repo is public, remove this so we use a versioned module
 replace (
+	github.com/aws/cluster-api-provider-cloudstack => github.com/wanyufe/cluster-api-provider-cloudstack v0.4.5-rc2.0.20220601153610-4f3e7128e44b
```
maxdrib, line 163: Can you explain this approach? Is this a temporary measure? I'd rather use the actual repo, and not use your fork in prod. It'll also be confusing when you try to bump the capc version only to realize that nothing is happening because of this replace

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -58,50 +74,59 @@ kind: CloudStackMachineTemplate
...
+{{- if $annotationsCP}}
   annotations:
 {{- end }}
-{{- if .cloudstackControlPlaneDiskOfferingProvided}}
-    mountpath.diskoffering.{{.cloudstackAnnotationSuffix}}: {{.cloudstackControlPlaneDiskOfferingPath}}
-    device.diskoffering.{{.cloudstackAnnotationSuffix}}: {{.cloudstackControlPlaneDiskOfferingDevice}}
-    filesystem.diskoffering.{{.cloudstackAnnotationSuffix}}: {{.cloudstackControlPlaneDiskOfferingFilesystem}}
-    label.diskoffering.{{.cloudstackAnnotationSuffix}}: {{.cloudstackControlPlaneDiskOfferingLabel}}
+{{- if $diskConfiguredCP}}
+    diskoffering.{{$suffix}}: "{{.cloudstackControlPlaneDiskOfferingJsonStr | js}}"
```
maxdrib, line 81: what does `| js` do? Can we use this tool to avoid passing in the redundant values to the templater?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -753,24 +747,22 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
...
-		"cloudstackDiskOfferingName":       workerNodeGroupMachineSpec.DiskOffering.Name,
-		"cloudstackDiskOfferingCustomSize": workerNodeGroupMachineSpec.DiskOffering.CustomSize,
-		"cloudstackDiskOfferingPath":       workerNodeGroupMachineSpec.DiskOffering.MountPath,
-		"cloudstackDiskOfferingDevice":     workerNodeGroupMachineSpec.DiskOffering.Device,
-		"cloudstackDiskOfferingFilesystem": workerNodeGroupMachineSpec.DiskOffering.Filesystem,
-		"cloudstackDiskOfferingLabel":      workerNodeGroupMachineSpec.DiskOffering.Label,
+		"cloudstackTemplate":               workerNodeGroupMachineSpec.Template,
+		"cloudstackOffering":               workerNodeGroupMachineSpec.ComputeOffering,
+		"cloudstackDiskOffering":           workerNodeGroupMachineSpec.DiskOffering,
+		"cloudstackDiskOfferingJsonStr":    string(diskOfferingJsonStr),
```
maxdrib, line 760: why do we need both? It seems redundant

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -1,3 +1,19 @@
+{{- $suffix := .cloudstackAnnotationSuffix -}}
+{{- $templateCP := .cloudstackControlPlaneTemplate -}}
+{{- $offeringCP := .cloudstackControlPlaneOffering -}}
+{{- $diskCP := .cloudstackControlPlaneDiskOffering -}}
+{{- $diskJsonStrCP := .cloudstackControlPlaneDiskOfferingJsonStr -}}
+{{- $isoCP := .cloudstackControlPlaneISOAttachment -}}
```
maxdrib, line 6: are these assignments doing anything? What's the purpose of assigning a variable to another variable?

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -58,50 +74,59 @@ kind: CloudStackMachineTemplate
...
 {{- end}}
 spec:
   template:
     spec:
       offering:
-{{- if .cloudstackControlPlaneComputeOfferingId}}
-        id: {{.cloudstackControlPlaneComputeOfferingId}}
+{{- if $offeringCP.Id}}
+        id: {{$offeringCP.Id}}
 {{- end}}
```
maxdrib, line 95: do we need these `if`'s? Or can we just always include this line and have an empty value? Same for the other conditional lines

```suggestion
        id: {{$offeringCP.Id}}
```

maxdrib (changes requested):
Let's also add some more unit tests to increase coverage. Overall this feels like a big change but a positive one in my mind to make the resource usage more efficient and tidy. Nice work.

go.mod
wanyufe: this is a temp solution when capc repo is not ready (in the middle of transfer). 
we need to change from github.com to kubernetes.sigs before release, I could not create a branch in CAPC's offical repo, so this may still be needed during dev period.

controllers/controllers/resource/fetcher.go
```diff
@@ -638,22 +637,50 @@ func MapMachineTemplateToCloudStackMachineConfigSpec(csMachineTemplate *cloudsta
...
+func parseISOAttachment(csMachineTemplate *cloudstackv1.CloudStackMachineTemplate) (ISOAttachment anywherev1.CloudStackISOAttachment, err error) {
+	ISOAttachment = anywherev1.CloudStackISOAttachment{}
+	ISOAttachmentAnnotation, exists := csMachineTemplate.Annotations["ISOAttachment."+constants.CloudstackAnnotationSuffix]
+	if !exists || len(ISOAttachmentAnnotation) == 0 {
+		return ISOAttachment, nil
+	}
+	err = json.Unmarshal([]byte(ISOAttachmentAnnotation), &ISOAttachment)
+	if err == nil {
+		ISOAttachment.Id = csMachineTemplate.Spec.Spec.Spec.ISOAttachment.ID
+		ISOAttachment.Name = csMachineTemplate.Spec.Spec.Spec.ISOAttachment.Name
```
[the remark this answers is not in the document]
wanyufe: yes, the annotation should have the ID/Name, these two lines can be removed.

controllers/controllers/resource/fetcher.go
```diff
@@ -638,22 +637,50 @@ func MapMachineTemplateToCloudStackMachineConfigSpec(csMachineTemplate *cloudsta
...
+	}
+	return ISOAttachment, err
+}
+
+func parseSymlinks(csMachineTemplate *cloudstackv1.CloudStackMachineTemplate) (symlinks anywherev1.SymlinkMaps, err error) {
```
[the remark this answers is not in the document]
wanyufe: I believe it needs typed object. But these several methods can be merged into one and using `switch` inside.

pkg/providers/cloudstack/config/template-cp.yaml
wanyufe: technically, we don't need these if -- the generated yaml may looks tidier if we have if.
I agree if we can remove lots of if for id/name in diskoffering, compute offering, template, isoattachment.

pkg/providers/cloudstack/config/template-cp.yaml
wanyufe: the only benefits is shorter, but if we use shorter variable instead of .cloudstackControlPlane..., we may not need these assignments.

pkg/providers/cloudstack/config/template-cp.yaml
wanyufe: it transfer {"a" : "b"} into {\"a\": \"b\"}

Since annotation is map[string]string, and " is used as start/end of a string, so we have to | js to escape "

wanyufe:
I will address the code coverage once we come back from OOO on 6/8/2022.

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -1,3 +1,9 @@
+{{- $diskConfiguredCP := or (gt (len .cloudstackControlPlaneDiskOffering.Id) 0) (gt (len .cloudstackControlPlaneDiskOffering.Name) 0) -}}
```
jiayiwang7, line 1: can we move these logics to cloudstack.go?

go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.16.2
 	github.com/aws/aws-sdk-go-v2/config v1.15.3
 	github.com/aws/aws-sdk-go-v2/service/ec2 v1.34.0
-	github.com/aws/cluster-api-provider-cloudstack v0.4.4
```
maxdrib: Why are we removing the dependency on CAPC?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -810,7 +802,7 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
 		values["noProxy"] = noProxyList
 	}
 
-	return values
+	return values, nil
```
maxdrib, line 805: What's the point of adding a new return error arg for `buildTemplateMapCP` and `buildTemplateMapMD` if we're just going to return `nil` every time?

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackmachineconfigs.yaml
```diff
@@ -38,6 +38,48 @@ spec:
             description: CloudStackMachineConfigSpec defines the desired state of
               CloudStackMachineConfig
             properties:
+              ISOAttachment:
+                description: ISOAttachment attach an ISO rom which has been previously
```
jweite-amazon, line 42: Reads more clearly without "ISOAttachment" in the description string.   Also, "attaches".

TerryHowe:
Reopen if needed. Looks dated.

PR #2316, opened 2022-06-02
Make repo Autoinit optional

danbudris:
*Issue #, if available:*

*Description of changes:*
We already go through a specific initialization for the Github provider for flux; this conflicts with the way go-github does their initialization when auot-init is set to true. 

So, let's make auto-init optional so we can use it for the flux git tests and not conflict with our existing initialization.

*Testing (if applicable):*


ewollesen:
Is there some clever way that you could get the information from your PR description into a comment in the code in a place/way that makes sense? It feels like that's a good bit of "why" information that would make a good comment to alert future devs as to why this option exists, and how to set it appropriately.

PR #2321, opened 2022-06-02
Strip superfluous Tinkerbell configuration from DatacenterConfig

chrisdoherty4:
Remove superfluous Tinkerbell configuration.

The removed configuration items were only necessary when deploying the stack independently of cluster provisioning. Having Kubeified and implemented the `--setup-tinkerbell` functionality we no longer need the configuration as its configured in a different way without user input.


chrisdoherty4:
I just noticed a bunch of formatting changes - my fault! They aren't particularly extensive and those PR is entirely about removing the unused datacenterconfig fields so I'll leave as is.

abhinavmpandey08 (approved):
/approve

PR #2327, opened 2022-06-03
Enable taints support in snow

jiayiwang7:
*Issue #, if available:*

#1574 

*Description of changes:*

Add taints support for snow provider. Need to recreate `KubeadmConfigTemplate` and `SnowMachineTemplate` whenever worker node group's taints are updated.

*Testing (if applicable):*

unit tests


pkg/providers/snow/fetch.go
```diff
@@ -56,3 +61,40 @@ func oldWorkerMachineTemplate(ctx context.Context, kubeclient kubernetes.Client,
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	return cluster, nil
+}
+
+func oldWorkerNodeGroupsMap(ctx context.Context, kubeClient kubernetes.Client, clusterSpec *cluster.Spec) (map[string]v1alpha1.WorkerNodeGroupConfiguration, error) {
+	eksaCluster, err := oldEksaCluster(ctx, kubeClient, clusterSpec)
```
g-gaston, line 94: I'm a bit concerned about this
What's the purpose of checking the "old" cluster?

The concerns comes from the fact that in the controller, you won't be able to get the old version, since by the time we hit this code, it has already been updated 

We should try to no rely on the "past" state of any eksa object

pkg/providers/snow/objects.go
```diff
@@ -80,30 +132,38 @@ func NewMachineTemplateName(new, old *snowv1.AWSSnowMachineTemplate) (string, er
...
+	if recreateKubeadmConfigTemplateNeeded(new, old) {
+		return clusterapi.IncrementName(old.GetName())
 	}
-	return m, nil
+
+	return old.GetName(), nil
+}
+
+func recreateKubeadmConfigTemplateNeeded(new, old *bootstrapv1.KubeadmConfigTemplate) bool {
+	return !v1alpha1.TaintsSliceEqual(new.Spec.Template.Spec.JoinConfiguration.NodeRegistration.Taints, old.Spec.Template.Spec.JoinConfiguration.NodeRegistration.Taints)
```
g-gaston, line 168: can we just compare `new.Spec` and `old.Spec`? That would make this a bit more future proof

pkg/providers/snow/objects_test.go
```diff
@@ -266,3 +432,58 @@ func TestWorkersObjectsGetMachineTemplateError(t *testing.T) {
...
+				Get(
+					g.ctx,
+					clusterapi.MachineDeploymentName(g.clusterSpec, g.clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0]),
+					constants.EksaSystemNamespace,
+					&clusterv1.MachineDeployment{},
+				).
+				Return(apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: ""}, ""))
+
+			g.clusterSpec.Cluster.Spec.RegistryMirrorConfiguration = tt.registryMirrorConfig
+			_, got, err := snow.WorkersMachineAndConfigTemplate(g.ctx, g.kubeconfigClient, g.clusterSpec)
```
g-gaston, line 451: why not asserting the first value?

pkg/providers/snow/objects.go
jiayiwang7: spec field changes should not trigger machine recreate, meaning no new kubeadmconfigtemplate created

jiayiwang7: chatted w folks offline -- we are treating kubeadmconfigtemplate immutable. Refer: https://github.com/aws/eks-anywhere/issues/189#issuecomment-1030421998

pkg/clusterapi/fetch.go
```diff
@@ -0,0 +1,42 @@
...
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	return md, nil
+}
+
+func KubeadmConfigTemplateInCluster(ctx context.Context, kubeclient kubernetes.Client, clusterSpec *cluster.Spec, md *clusterv1.MachineDeployment) (*bootstrapv1.KubeadmConfigTemplate, error) {
```
g-gaston, line 28:
```suggestion
func KubeadmConfigTemplateInCluster(ctx context.Context, kubeclient kubernetes.Client, md *clusterv1.MachineDeployment) (*bootstrapv1.KubeadmConfigTemplate, error) {
```

g-gaston (approved):

PR #2329, opened 2022-06-03
Pull hook from the bundle

abhinavmpandey08:
*Description of changes:*
This PR updates the cli to pull the Hook images from the URIs listed in the bundle instead of being hardcoded.
It also bumps up the lease duration for IPs handed out by boots to 1 month

*Testing (if applicable):*
Created the cluster with the hook from bundle

### DIFF
--- a/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
@@ -152,7 +152,7 @@ func hardwareFromMachine(m Machine) *v1alpha1.Hardware {
 							Gateway: m.Gateway,
 							Family:  4,
 						},
-						LeaseTime:   2630000,
+						LeaseTime:   86400,
 						Hostname:    m.Hostname,
 						NameServers: m.Nameservers,
 						UEFI:        true,
--- a/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
@@ -152,7 +152,7 @@ func hardwareFromMachine(m Machine) *v1alpha1.Hardware {
 							Gateway: m.Gateway,
 							Family:  4,
 						},
-						LeaseTime:   86400,
+						LeaseTime:   2630000,
 						Hostname:    m.Hostname,
 						NameServers: m.Nameservers,
 						UEFI:        true,
--- a/pkg/providers/tinkerbell/stack/stack.go
+++ b/pkg/providers/tinkerbell/stack/stack.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"sigs.k8s.io/yaml"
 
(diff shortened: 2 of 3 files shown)


pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -152,7 +152,7 @@ func hardwareFromMachine(m Machine) *v1alpha1.Hardware {
 							Gateway: m.Gateway,
 							Family:  4,
 						},
-						LeaseTime:   86400,
+						LeaseTime:   2630000,
```
ptrivedi, line 155: I would do this in a separate lone PR

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -119,6 +101,11 @@ func (s *Installer) Install(ctx context.Context, bundle releasev1alpha1.Tinkerbe
 		})
 	}
 
+	osiePath, err := getURIDir(bundle.Hook.Initramfs.Amd.URI)
```
ptrivedi, line 104: why is this path named as Initramfs? what about vmlinuz? This path is for both initrd and the kernel

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -210,3 +208,11 @@ func (s *Installer) uninstallBootsFromDocker(ctx context.Context) error {
 
 	return nil
 }
+
+func getURIDir(uri string) (string, error) {
+	index := strings.LastIndex(uri, "/")
+	if index == -1 {
+		return "", fmt.Errorf("uri is invalid: %s", uri)
```
ptrivedi, line 215: which uri is invalid? little more descriptive error would be good

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -140,6 +127,10 @@ func (s *Installer) Install(ctx context.Context, bundle releasev1alpha1.Tinkerbe
 			deploy: !s.bootsOnDocker,
 			image:  bundle.Boots.Image.URI,
 			env:    bootEnv,
+			args: []string{
+				"-dhcp-addr=0.0.0.0:67",
```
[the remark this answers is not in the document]
abhinavmpandey08: it doesn't default to that. The reason why we didn't have it previously was because we weren't providing custom args for boots. If you look in the helm values for boots, the args defaults to `dhcp-addr=0.0.0.0:67` if not provided https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/tinkerbell/tinkerbell-chart/chart/values.yaml#L14

pkg/providers/tinkerbell/stack/stack.go
abhinavmpandey08: It does print out the URI. This is just a generic helper. I added more descriptive message from where this function gets called. Let me know what else I can add here

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
abhinavmpandey08, line 155:
```suggestion
						LeaseTime:   86400,
```

pkg/providers/tinkerbell/stack/stack.go
ptrivedi: correct, on tinkerbell side, yes. but i was wondering more about whether that's verified anywhere on eks-a cli side in the bundle and stuff

pkg/providers/tinkerbell/stack/stack.go
ptrivedi: oh missed that, ok

Issue #2185, opened 2022-05-17
Rerun commands after failure


g-gaston:
## What would you like to be added

The ability to rerun CLI commands after an error.

## Why is this needed

Most of the errors returned by the CLI `cluster` commands are either transient or fixable with manual intervention. However, the CLI commands are not idempotent. This makes impossible to rerun commands after they have already failed, even if the root cause of the issues has already been solved.

This makes the experience very disruptive, specially for the `upgrade cluster` command, which could leave the cluster in an irrecoverable state without the necessary manual cleanup. This "cleanup" is not documented and it heavily dependents on the internal CLI's implementation. For certain scenarios, it is not even possible, requiring users to destroy and recreate their clusters from scratch.

## Solution

The ideal solution for this problem is to move most of the CLI logic to a kubernetes controller and run it in the bootstrap (or management) cluster. The reconciliation loop is the perfect pattern for this scenario. The current ongoing work to offer full cluster lifecycle APIs will push us forward into this direction. However, it will require a significant amount of time to be completed and it will not even be enough (it's not scoped to include running it from a bootstrap cluster).

For a CLI centric implementation, making the commands idempotent will be the ideal solution. This presents multiple challenges, like the dependency on external binaries operations that might not be idempotent or the difficulty of moving objects back to a cluster with an unreachable control plane. Unfortunately, this solution will probably take even more time than the previous one.

A third option, less robust but significantly simpler, would be to implement a "checkpoint" capability. The CLI will keep a registry of all completed steps for a workflow and the necessary data to restore its state. When a command is rerun after an error, the CLI will skip the completed steps, restoring the state of the program. This requires for each command step to be idempotent withing itself: if they fail mid execution, we need the to be able to rerun them from the beginning. If one the steps breaks this rule, it will need to be slit into two or more steps.

### Proposal

I propose implementing the third option: using a checkpoint. The checkpoint data would be stored as a file in disk and the idempotent steps would be the `Tasks` in the command `workflows`.

A POC proving this idea can be found [here](https://github.com/g-gaston/eks-anywhere/tree/workflow-checkpoint). For a codified example of the proposed user flow, refer to [this E2E test](https://github.com/g-gaston/eks-anywhere/blob/workflow-checkpoint/test/e2e/upgrade_test.go#L318).

This an example of the checkpoint file:

```yaml
completedTasks:
  bootstrap-cluster-init:
    ExistingManagement: false
[...]

PR #2330, opened 2022-06-03
Rerun CLI commands design doc

taneyland:
*Issue #, if available:*
#2185 

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/designs/rerun-cli-commands.md
+++ b/designs/rerun-cli-commands.md
@@ -1,0 +1,114 @@
+# Rerun CLI commands after failure
+
+## Introduction
+**Problem:**
+Most of the errors returned by the CLI cluster commands are either transient or fixable with manual intervention.
+However, the CLI commands are not idempotent.
+This makes impossible to rerun commands after they have already failed, even if the root cause of the issues has already been solved.
+This makes the experience very disruptive, specially for the upgrade cluster command, which could leave the cluster in an irrecoverable state without the necessary manual cleanup.
+This "cleanup" is not documented and it heavily dependents on the internal CLI's implementation.
+For certain scenarios, it is not even possible, requiring users to destroy and recreate their clusters from scratch.
+
+## Solution
+The proposed solution is to implement a "checkpoint" capability.
+This will serve as the first step to a larger story of improving the customer experience with upgrading clusters and troubleshooting in general.
+The CLI will keep a registry of all completed steps for a workflow and the necessary data to restore its state.
+If the `upgrade` command fails and the error is manually fixable, the user can simply rerun the same command after implementing the fix.
+When a command is rerun after an error, the CLI will skip the completed steps, restoring the state of the program and proceeding with the upgrade from there.
+The checkpoint data would be stored in the `<clusterName>/generated` folder as a file named `<clusterName>-checkpoint.yaml`.
+
+Example checkpoint file:
+
+```yaml
+completedTasks:
+  bootstrap-cluster-init:
+    checkpoint: 
+      ExistingManagement: false
+      KubeconfigFile: test/generated/test.kind.kubeconfig
+      Name: test
+  ensure-etcd-capi-components-exist:
+    checkpoint: null
+  install-capi:
+    checkpoint: null
+  pause-controllers-reconcile:
+    checkpoint: null
+  setup-and-validate:
+    checkpoint: null
+  update-secrets:
+    checkpoint: null
+  upgrade-core-components:
+    checkpoint:
+      components: []
+  upgrade-needed:
+    checkpoint: null
+```
+
+---
+We will add two methods, `Restore()` and `Checkpoint()`.
+The logic in both methods will vary depending on what is needed from each task.
+
+```go
+type Task interface {
+  Run(ctx context.Context, commandContext *CommandContext) Task
+  Name() string
+  Checkpoint() TaskCheckpoint
+  Restore(ctx context.Context, commandContext *CommandContext, checkpoint TaskCheckpoint) (Task, error)
+}
+```
+
+The `Checkpoint()` method will return the information necessary to successfully restore the task if the operation fails.
+This information will then be retained in the checkpoint file.
+
+The `Restore()` method will use the information from the checkpoint file to restore the state.
+Again, this will vary depending on what's needed for that specific task.
+Once the restoration succeeds for a task, we will return the next task (essentially skipping the task that was restored).
+
+ ---
+When a task finishes with no errors, we will add it as a completed task in the checkpoint file with the necessary information (if any) to allow the command to rerun successfully.
+
+```go
+if commandContext.OriginalError == nil {
+    // add task to completedTasks in checkpoint file
+    checkpointInfo.taskCompleted(task.Name(), task.Checkpoint())
+}
+```
+
+When the user re-runs the command, we will check if the task exists as a completed task in the checkpoint file.
+If so, we will restore any necessary information from the checkpoint file and set the next task.
+
+```go
+for task != nil {
+    if completedTask, ok := checkpointInfo.CompletedTasks[task.Name()]; ok {
+        // Restore any necessary information from the checkpoint file & set task to the next task
+        nextTask, err := task.Restore(ctx, commandContext, completedTask)
+        ...
+        continue
+    }
+    logger.V(4).Info("Task start", "task_name", task.Name())
+    ...
+    ...
+}
+```
+
+### Other notable changes:
+* If the CLI fails to install CAPI objects to the bootstrap cluster, the CLI currently deletes the bootstrap cluster.
+  This will obviously cause problems if we were to rerun the command at that point.
+  Instead of deleting the bootstrap cluster if `InstallCAPITask` fails, it will proceed to `CollectDiagnosticsTask`.
+  From there, the user can re-run the command to attempt installing the CAPI components on the previously created bootstrap cluster.
+
+
+* If the CLI fails when moving CAPI management, things start to get a bit tricky.
+  We could move CAPI management back to the source cluster if it fails, but this operation can be risky & potentially leave the cluster in an undesirable state.
+  For now, we will retry moving the CAPI objects if the task fails the first time.
+  A deeper dive would be required to decide on a better way to roll back those changes if moving CAPI fails.
+  An idea for a future improvement is to backup the CAPI objects associated with the move using the `clusterctl backup` command. See [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/cmd/backup.go).
+
+
+* The only task that we should always run is the `setup-and-validate` task.
+  Skipping this task would leave a blind spot for misconfigurations from the user input.
+
+### Future Work
+
+* Rolling back changes if moving CAPI management fails
+* Add more validations to ensure each task's purpose was successful & left the cluster in a healthy state before skipping
+* Building on top of this framework to further improve user experience


designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+This makes the experience very disruptive, specially for the upgrade cluster command, which could leave the cluster in an irrecoverable state without the necessary manual cleanup.
+This "cleanup" is not documented and it heavily dependents on the internal CLI's implementation.
+For certain scenarios, it is not even possible, requiring users to destroy and recreate their clusters from scratch.
+
+## Solution
+The proposed solution is to implement a "checkpoint" capability.
+This will serve as the first step to a larger story of improving the customer experience with upgrading clusters and troubleshooting in general.
+The CLI will keep a registry of all completed steps for a workflow and the necessary data to restore its state.
+When a command is rerun after an error, the CLI will skip the completed steps, restoring the state of the program.
+The checkpoint data would be stored in the `<clusterName>/generated` folder as a file named `<clusterName>-checkpoint.yaml`.
```
vivek-koppuru, line 17: Did you research into other examples of this?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+
+Example checkpoint file:
```
jiayiwang7, line 19: how would conditional tasks be reflected in the checkpoint?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+
+```yaml
+completedTasks:
+  bootstrap-cluster-init:
+    ExistingManagement: false
```
vivek-koppuru, line 24: What is going to be stored in the other structs? I am assuming it's more of a `true/false` right?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### Other notable changes:
+* If the CLI fails to install CAPI objects to the bootstrap cluster, it deletes the bootstrap cluster.
+  This will obviously cause problems if we were to rerun the command at that point.
+  Therefore, in the `Restore()` method for the `CreateBootstrapClusterTask`, we will re-create the bootstrap cluster based on the cluster information in the `bootstrap-cluster-init` key from the checkpoint file if the bootstrap cluster does not exist.
+  An alternative to this would be to not delete the bootstrap cluster if `InstallCAPITask` fails & go straight to `CollectDiagnosticsTask`.
+
+
+* If the CLI fails when moving CAPI management, things start to get a bit tricky.
```
jiayiwang7, line 67: what is the rerun strategy for handling control plane timing out?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+  We could move CAPI management back to the source cluster if it fails, but this operation can be risky & potentially leave the cluster in an undesirable state.
+  Therefore, we will move this out of the scope of this initial MVP.
+  A deeper dive would be required to decide on a better way to roll back those changes if moving CAPI fails.
+  An idea for a future improvement is to backup the CAPI objects associated with the move using the `clusterctl backup` command. See [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/cmd/backup.go).
+  For now, if the cluster fails at this point (which does not happen terribly often), it will still fail, and we will handle that case in the future.
+
+
+* The only task that we should always run is the `setup-and-validate` task.
```
vivek-koppuru, line 75: We want to make sure that no other step sets up structs in a way that is used in the subsequent steps. I think that is what @chrisdoherty4 was referring to.

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+  Skipping this task would leave a blind spot for misconfigurations from the user input.
+  However, if the CLI fails while CAPI management is temporarily in the bootstrap cluster, the preflight validations would fail because those CAPI objects are no longer present on the workload cluster.
+  To account for this, we could add a boolean flag to the command context to check whether or not CAPI management has been moved.
+  This flag can be set in the `Restore()` methods for the tasks where we are moving CAPI management.
+  If CAPI management has been moved from the workload cluster, we should skip the validation that checks if those objects are present on the cluster.
+
+### Future Work
+
+* Rolling back changes if moving CAPI management fails
```
vivek-koppuru, line 84: The ideal scenario is not having to rely on a checkpoint file and reading the state of the objects in the clusters to determine what phase we are in, similar to a reconciler. That might be a future work item too if we can remove the reliance of a checkpoint file.

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+**Problem:**
+Most of the errors returned by the CLI cluster commands are either transient or fixable with manual intervention.
+However, the CLI commands are not idempotent.
+This makes impossible to rerun commands after they have already failed, even if the root cause of the issues has already been solved.
+This makes the experience very disruptive, specially for the upgrade cluster command, which could leave the cluster in an irrecoverable state without the necessary manual cleanup.
+This "cleanup" is not documented and it heavily dependents on the internal CLI's implementation.
+For certain scenarios, it is not even possible, requiring users to destroy and recreate their clusters from scratch.
+
+## Solution
+The proposed solution is to implement a "checkpoint" capability.
```
vivek-koppuru, line 13: How would a user enable this checkpoint capability? I am assuming that if they want to rerun it based on an unfinished execution they would specify using a flag or something.

designs/rerun_cli_commands.md
g-gaston, line 17: isn't it redundant if it's already in a folder with the cluster name?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+    ...
+    ...
+}
+```
+
+### Other notable changes:
+* If the CLI fails to install CAPI objects to the bootstrap cluster, it deletes the bootstrap cluster.
+  This will obviously cause problems if we were to rerun the command at that point.
+  Therefore, in the `Restore()` method for the `CreateBootstrapClusterTask`, we will re-create the bootstrap cluster based on the cluster information in the `bootstrap-cluster-init` key from the checkpoint file if the bootstrap cluster does not exist.
+  An alternative to this would be to not delete the bootstrap cluster if `InstallCAPITask` fails & go straight to `CollectDiagnosticsTask`.
```
g-gaston, line 64: This is my preferred option

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+* If the CLI fails when moving CAPI management, things start to get a bit tricky.
+  We could move CAPI management back to the source cluster if it fails, but this operation can be risky & potentially leave the cluster in an undesirable state.
+  Therefore, we will move this out of the scope of this initial MVP.
+  A deeper dive would be required to decide on a better way to roll back those changes if moving CAPI fails.
+  An idea for a future improvement is to backup the CAPI objects associated with the move using the `clusterctl backup` command. See [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/cmd/backup.go).
```
g-gaston, line 71: This is a good idea, but for now, can't we just re run the `move` command?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+  install-capi: null
+  pause-controllers-reconcile: null
+  setup-and-validate: null
+  update-secrets: null
+  upgrade-core-components:
+    components: []
+  upgrade-needed: null
+```
+
+When the task finishes with no errors, we will add it as a completed task in the checkpoint file with the necessary information (if any) to allow the command to rerun successfully.
```
jiayiwang7, line 37: I wonder if we should add more validations to each step before updating the checkpoint step to be "true/success". or maybe validate each succeeded step when we start the rerun

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+  Therefore, we will move this out of the scope of this initial MVP.
+  A deeper dive would be required to decide on a better way to roll back those changes if moving CAPI fails.
+  An idea for a future improvement is to backup the CAPI objects associated with the move using the `clusterctl backup` command. See [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/cmd/backup.go).
+  For now, if the cluster fails at this point (which does not happen terribly often), it will still fail, and we will handle that case in the future.
+
+
+* The only task that we should always run is the `setup-and-validate` task.
+  Skipping this task would leave a blind spot for misconfigurations from the user input.
+  However, if the CLI fails while CAPI management is temporarily in the bootstrap cluster, the preflight validations would fail because those CAPI objects are no longer present on the workload cluster.
+  To account for this, we could add a boolean flag to the command context to check whether or not CAPI management has been moved.
```
g-gaston, line 78: I don't follow
If validations are another "phase" and that one has been completed, why do we need to add special handling? Shouldn't the framework skip it already as any other completed task?

designs/rerun_cli_commands.md
```diff
@@ -0,0 +1,85 @@
...
+    KubeconfigFile: test/generated/test.kind.kubeconfig
+    Name: test
+  ensure-etcd-capi-components-exist: null
+  install-capi: null
+  pause-controllers-reconcile: null
+  setup-and-validate: null
+  update-secrets: null
+  upgrade-core-components:
+    components: []
+  upgrade-needed: null
```
vivek-koppuru, line 34: We also need to address if the CLI version changes between running it subsequent times, maybe a validation of the version used in the checkpoint file?

chrisdoherty4:
Love the idea. Here's a bit of a brain dump.

- Our software accepts a bunch of inputs. How would we handle those inputs (always running `SetupAndCreate` style functions implies we always accept the input)? What happens if the user changes them between invocations where checkpoints have been created operating on different data to what has now been supplied? Do we need checks for each task (perhaps checksums of data or diff checks)? Do we expect subsets of input data to be mutable between invocations? We should operate on 0 trust with ABCs: Assume nothing, Believe no-one, Check everything.
- Building on the previous point, are tasks sufficiently granular to support sane restoration and checks before proceeding with a task. The [`CreateWorkloadCluster`](https://github.com/chrisdoherty4/eks-anywhere/blob/feature/read-csv-on-upgrade/pkg/workflows/create.go#L198) task, for example, has at least 5 things going on: it installs networking, AWS IAM authentication, storage classes, CAPI, machine health checks, and triggers some secret update (the workflow re-design I hope to embark on post baremetal GA would seek to break a lot of this up). Some of these actions may need to be atomic operations but installing networking definitely feels like it could be done separate to installing CAPI which would benefit the UX. Do we know how difficult restoring that state might be?
- Much of our code isn't honoring the SRP; consequently, it has side effects. How do we build confidence that any state required by side effects is captured and restored?

These questions are largely geared towards questioning whether we're in a position to be able to do a continuation type thing in a sensible way given the software was never designed to accommodate that.

taneyland:
> Love the idea. Here's a bit of a brain dump.
> 
>     * Our software accepts a bunch of inputs. How would we handle those inputs (always running `SetupAndCreate` style functions implies we always accept the input)? What happens if the user changes them between invocations where checkpoints have been created operating on different data to what has now been supplied? Do we need checks for each task (perhaps checksums of data or diff checks)? Do we expect subsets of input data to be mutable between invocations? We should operate on 0 trust with ABCs: Assume nothing, Believe no-one, Check everything.
> 
>     * Building on the previous point, are tasks sufficiently granular to support sane restoration and checks before proceeding with a task. The [`CreateWorkloadCluster`](https://github.com/chrisdoherty4/eks-anywhere/blob/feature/read-csv-on-upgrade/pkg/workflows/create.go#L198) task, for example, has at least 5 things going on: it installs networking, AWS IAM authentication, storage classes, CAPI, machine health checks, and triggers some secret update (the workflow re-design I hope to embark on post baremetal GA would seek to break a lot of this up). Some of these actions may need to be atomic operations but installing networking definitely feels like it could be done separate to installing CAPI which could benefit the UX as you could "retry" those separately. Do we know how difficult restoring that state might be?
> 
>     * Much of our code isn't honoring the SRP; consequently, it has side effects. How do we build confidence that any state required by side effects is captured and restored?
> 
> 
> These questions are largely geared towards questioning whether we're in a position to be able to do a continuation type thing in a sensible way given the software was never designed to accommodate that.

Thanks for this. Very valid concerns. To address a few of your thoughts, this MVP implementation for this feature is currently scoped for only the `upgrade` command with plans to expand it in the near future. The tasks on the `upgrade` command are a lot more granular so the restoration process is fairly straight-forward for these tasks. We are also planning to add more validations for the user input between invocations but that's scoped as future work beyond this MVP. We do have validations for immutable fields during upgrades so it would be an idea to expand those to help validate input when rerunning commands. I agree with you that a re-design for the workflow would be very beneficial for this feature, especially when expanding it to the `create` command.

PR #2336, opened 2022-06-03
Removing old tinkerbell delete workflow code

pokearu:
*Description of changes:*
With the new Kubified Tinkerbell stack eks-a no longer uses the older CAPT CRDs. 
Removing old logic from the delete cluster workflow.


abhinavmpandey08 (approved):
/lgtm

PR #2337, opened 2022-06-04
Add tinkerbell-chart to the bundle

abhinavmpandey08:
*Description of changes:*
This PR adds the `tinkerbell-chart` to the bundle


ptrivedi:
Would be good to have “tinkerbell-chart” as a const declaration. But I guess that applies to all other projects.

onky other thing is the arch being limited to amd64, but I guess it’s okay for now

abhinavmpandey08:
> Would be good to have “tinkerbell-chart” as a const declaration. But I guess that applies to all other projects.
> 
> onky other thing is the arch being limited to amd64, but I guess it’s okay for now

We only build amd64 images. Also, the image arch doesn't matter too much since it's a helm chart and not a docker image

PR #2339, opened 2022-06-04
Use helm chart from bundle in Tinkerbell

abhinavmpandey08:
*Description of changes:*
Use tinkerbell helm chart from bundle instead of using a hardcoded one


pokearu (approved):
/lgtm
/approve

PR #2342, opened 2022-06-06
Quick sleep in git setup

danbudris:
*Issue #, if available:*

*Description of changes:*
getting 404s when attempting to add a deploy key to the newly created repositories in the Github tests; I ran into this on my local one time out of dozens of tests; I'm wondering if this is related to eventual consistency int he GitHub API, want to try a sleep to see if it influences the problem.

*Testing (if applicable):*

### DIFF
--- a/internal/test/e2e/fluxGit.go
+++ b/internal/test/e2e/fluxGit.go
@@ -141,6 +141,7 @@ func (e *E2ESession) setupGithubRepo(repo string, envVars map[string]string) (*g
 		return nil, fmt.Errorf("creating repository in Github for test: %v", err)
 	}
 
+	// Wait for eventual consistency in the GitHub API when creating a repository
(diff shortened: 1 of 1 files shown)


internal/test/e2e/fluxGit.go
```diff
@@ -140,6 +141,9 @@ func (e *E2ESession) setupGithubRepo(repo string, envVars map[string]string) (*g
 		return nil, fmt.Errorf("creating repository in Github for test: %v", err)
 	}
 
+	// Wait for eventual consistency in the GitHub API when creating a repository before adding a deploy key
+	time.Sleep(time.Second * 5)
```
ewollesen, line 145: I can ok this for now, but this is still going to be brittle. It would be better to perform some sort of polling with an eventual limit.

ewollesen (commented):
/lgtm

Issue #1090, opened 2022-01-31
Create CAPI objects from EKS Anywhere spec


mdsgabriel:

maxdrib:
I think cloudstack work should be added to this issue

PR #2353, opened 2022-06-06
Add design doc for reference Cluster to Bundles

g-gaston:
Derived from [#1090](https://github.com/aws/eks-anywhere/issues/1090)

### DIFF
--- a/designs/cluster-bundles-reference.md
+++ b/designs/cluster-bundles-reference.md
@@ -1,0 +1,41 @@
+# `Cluster` reference to `Bundles`
+
+## Introduction
+
+**Problem:** we rely on naming and a default namespace to find the `Bundles` object being used by an EKS-A `Cluster`.
+
+Currently, when we create a cluster, we rename the Bundles object we are using to have the same name as the Cluster object. Then, we use that same name and same namespace to retrieve it from the cluster and rebuild a cluster.Spec in the CLI/controller.
+
+This is not a very clean pattern since it forces us to duplicate `Bundles` for multiple clusters. Also, in order to point a cluster to a new `Bundles` we need to update the existing one (or delete and recreate). Ideally, `Bundles` should be immutable in the cluster in the same way they are in real life.
+
+### Goals and Objectives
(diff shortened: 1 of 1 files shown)


designs/cluster-bundles-reference.md
```diff
@@ -0,0 +1,41 @@
+# `Cluster` reference to `Bundles`
+
+## Introduction
+
+**Problem:** we rely on naming and a default namespace to find the `Bundles` object being used by an EKS-A `Cluster`.
+
+Currently, when we create a cluster, we rename the `Bundles` object we are using to have the same name as such `Cluster`. Then, when use that same name an same namespace to retrieve it from the cluster to rebuild a `cluster.Spec` in the CLI/controller.
```
jiayiwang7, line 7: is it my english or i can't understand this sentence

designs/cluster-bundles-reference.md
```diff
@@ -0,0 +1,41 @@
...
+
+This is not a very clean pattern since it forces us to duplicate `Bundles` for multiple clusters. Also, in order to point a cluster to a new `Bundles` we need to update the existing one (or delete and recreate). Ideally, `Bundles` should be immutable in the cluster in the same way they are in real life.
```
jiayiwang7, line 9: should `Bundles` always be immutable? can eksa reconcile on bundle resources one day? i understand `Bundles` are more stateless CRD that eksa only reads it. But can user update certain component versions in bundle and expect it to work?

g-gaston: I would prefer if bundles were immutable.

Aside from making more sense from a CRD perspective, it makes cluster reconciliation easier: if you want to change a dependency, you need a new `Bundles` and point your `Cluster` to it.

That way we don't need to watch `Bundles` objects and back reference them to multiple `Cluster` objects when they change.

designs/cluster-bundles-reference.md
g-gaston: Fixed :)

jiayiwang7 (approved):

PR #2354, opened 2022-06-07
Use force-cleanup to delete local boots container

abhinavmpandey08:
*Description of changes:*
Use force cleanup flag to delete the local boots container if it is running

### DIFF
--- a/pkg/providers/tinkerbell/stack/stack.go
+++ b/pkg/providers/tinkerbell/stack/stack.go
@@ -53,7 +53,7 @@ type Installer struct {
 type InstallOption func(s *Installer)
 
 type StackInstaller interface {
-	CheckLocalBootsExistence(ctx context.Context) error
+	CleanupLocalBoots(ctx context.Context, forceCleanup bool) error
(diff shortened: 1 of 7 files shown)


pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -234,15 +234,25 @@ func getURIDir(uri string) (string, error) {
 	return uri[:index], nil
 }
 
-// CheckLocalBootsExistence determines whether Boots is already running locally on a Docker container
-// Returns nil if the container is not found and an error otherwise
-func (s *Installer) CheckLocalBootsExistence(ctx context.Context) error {
+// CleanupLocalBoots determines whether Boots is already running locally
+// and either cleans it up or errors out depending on the `remove` flag
+func (s *Installer) CleanupLocalBoots(ctx context.Context, remove bool) error {
```
ptrivedi, line 239: This is okay for now. Just a nit, 'CleanupLocalBoots' should just do clean up. Coming inside a function called 'CleanupLocalBoots' and then cleaning up only if 'remove'(force) is set -- could be changed to read better. Thanks for taking care of this!

PR #2359, opened 2022-06-07
Build `cluster.Config` from a kubernetes client

g-gaston:
Part of https://github.com/aws/eks-anywhere/issues/1090
Necessary to rebuild a `cluster.Spec` in the controller

## Description of changes

This introduces the ConfigClientBuilder that allows to register
processors to build different parts of the cluster.Config by retrieving
objects from a kubernetes cluster through a client. This is a very
similar pattern as the one used by the config manager.

Adds a default ConfigClientBuilder that pre-registers default
processors. For now, only identity providers, gitops and the snow
providers are implemented.

## Testing (if applicable)

Unit tests


jiayiwang7 (approved):
super clear!

PR #2364, opened 2022-06-08
Tinkerbell metadata url

panktishah26:
*Issue #, if available:*
Populate metadata_url based on tinkerbellIP.

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -50,7 +50,7 @@ type Provider struct {
 	hardwareCSVFile string
 	catalogue       *hardware.Catalogue
 	diskExtractor   hardware.DiskExtractor
-
+	tinkerbellIp    string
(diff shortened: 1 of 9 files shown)


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -117,6 +117,8 @@ func NewProvider(
 		}
 	}
 
+	tinkerbellIp, _ := networkutils.GetLocalIP()
```
chrisdoherty4, line 120: Any reason we aren't checking the error? I know this won't be the final code but we should still check the error.

panktishah26: True, NewProvider construct does not return error so I had to ignore it. Should I put this somewhere else?

abhinavmpandey08: you can get it in the provider factory and inject it from there to the provider

abhinavmpandey08 (approved):
/approve

PR #2367, opened 2022-06-08
Refactor validation to account for selectors

chrisdoherty4:
Ensure we have sufficient hardware to provision counts specified in the cluster configuration. This alters the validation algorithm to sum hardware based on selectors specified in the associated `TinkerbelMachineConfig`s.


pkg/providers/tinkerbell/create.go
```diff
@@ -109,26 +109,39 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
...
 
+	// Validate must happen last beacuse we depend on the catalogue entries for some checks.
+	if err := clusterSpecValidator.Validate(spec); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Provider) readCSVToCatalogue() error {
```
chrisdoherty4, line 136: This looks a little odd. To summarize, there's now the `readCSVToCatalogue()` function called on l.112 instead of having all the `readCSVToCatalogue()` logic in place.

pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -289,8 +275,13 @@ func MatchingDisksForSelectors(selectors []MachineSelector) MachineAssertion {
...
 }
 
-// labelsMatchSelector ensures all selector key-value pairs can be found in labels.
-func labelsMatchSelector(selector MachineSelector, labels Labels) bool {
+// LabelsMatchSelector ensures all selector key-value pairs can be found in labels.
+// If selector is empty true is always returned.
+func LabelsMatchSelector(selector v1alpha1.HardwareSelector, labels Labels) bool {
+	if selector.IsEmpty() {
+		return true
+	}
```
abhinavmpandey08, line 283: does this mean we are supporting a case where the labels are not specified and we pick hardware at random?
Based on [the conversation](https://github.com/aws/eks-anywhere/pull/2300#discussion_r889534093) on the label selector PR, it seems like the labels are required

chrisdoherty4: This is removed, albeit it was never required. The behavior is the same without it 🤦🏻 .

abhinavmpandey08 (approved):
/approve

PR #2371, opened 2022-06-09
Implement Bundles referenced in Cluster

g-gaston:
Part of #1090 
Implements #2353 

*Description of changes:*

* Add new field
* Set BundlesRef to downloaded Bundles in the CLI command flows when
building the cluster.Spec.
* Use BundlesRef if available when rebuilding cluster.Spec from cluster.

*Testing (if applicable):*

* Unit tests
* Run several E2E tests for create and upgrade with docker and vsphere.


pkg/cluster/default_manager.go
```diff
@@ -3,8 +3,20 @@ package cluster
 var defaultManager *ConfigManager
 
 func init() {
-	defaultManager = NewConfigManager()
-	err := defaultManager.Register(
+	var err error
+	defaultManager, err = NewDefaultConfigManager()
```
jiayiwang7, line 7: why do we still need this in `init()` if it's being called in spec.go? just for checking the error?

jiayiwang7 (approved):

pkg/cluster/default_manager.go
g-gaston: Discussed offline

PR #2372, opened 2022-06-09
Set `KUBEADM_BOOTSTRAP_TOKEN_TTL` to 120m for Tinkerbell provider

abhinavmpandey08:
*Description of changes:*
Baremetal nodes generally take longer to provision than the VMs which sometimes cause the kubeadm join tokens to surpass their TTL and expire.

This PR sets `KUBEADM_BOOTSTRAP_TOKEN_TTL` to 120m for Tinkerbell provider to give nodes enough time to provision and join the cluster.

### DIFF
--- a/pkg/providers/tinkerbell/tinkerbell.go
+++ b/pkg/providers/tinkerbell/tinkerbell.go
@@ -194,7 +194,8 @@ func (p *Provider) EnvMap(spec *cluster.Spec) (map[string]string, error) {
 		//
 		// Env read having set TINKERBELL_IP in the deployment manifest.
 		// https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L192
-		"TINKERBELL_IP": "<set in eks-a tinkerbell provider>",
+		"TINKERBELL_IP":               "IGNORED",
+		"KUBEADM_BOOTSTRAP_TOKEN_TTL": "120m",
 	}, nil
 }
 


pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -194,7 +194,8 @@ func (p *Provider) EnvMap(spec *cluster.Spec) (map[string]string, error) {
 		//
 		// Env read having set TINKERBELL_IP in the deployment manifest.
 		// https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L192
-		"TINKERBELL_IP": "<set in eks-a tinkerbell provider>",
+		"KUBEADM_BOOTSTRAP_TOKEN_TTL": "120m",
+		"TINKERBELL_IP":               "IGNORED",
```
chrisdoherty4, line 198: Can we maintain the note about where this is set so we know where to look if its spotted in the environment.

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -194,7 +194,8 @@ func (p *Provider) EnvMap(spec *cluster.Spec) (map[string]string, error) {
 		//
 		// Env read having set TINKERBELL_IP in the deployment manifest.
 		// https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/controllers/machine.go#L192
-		"TINKERBELL_IP": "<set in eks-a tinkerbell provider>",
+		"KUBEADM_BOOTSTRAP_TOKEN_TTL": "120m",
```
chrisdoherty4, line 197: 2hs feels like an excessive jump from 15m. I also think users want to control this so we should expose it in the cluster configuration.

pkg/providers/tinkerbell/tinkerbell.go
chrisdoherty4: 🤦🏻 _TINKERBELL_\_IP ... Ignore me!

chrisdoherty4 (approved):

PR #2375, opened 2022-06-09
add option to override the local tinkerbell IP in the bootstrap cluster

gwesterfieldjr:
Add option to override the local tinkerbell bootstrap ip.

This is needed for cases where tink needs to run on a different interface than the default interface ip that we infer.

Equinix requested this for their workflow. Also, testing using dhcp relay over vpn required this.

### DIFF
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/files"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
 	gitfactory "github.com/aws/eks-anywhere/pkg/git/factory"
+	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/manifests"
 	"github.com/aws/eks-anywhere/pkg/networking/cilium"
 	"github.com/aws/eks-anywhere/pkg/networking/kindnetd"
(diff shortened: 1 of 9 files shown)


abhinavmpandey08 (approved):
/approve

pkg/dependencies/factory.go
```diff
@@ -290,10 +291,19 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 				return fmt.Errorf("unable to get machine config from file %s: %v", clusterConfigFile, err)
 			}
 
-			tinkerbellIp, err := networkutils.GetLocalIP()
+			logger.V(4).Info("Inferring local Tinkerbell Bootstrap IP from environment")
+			localIp, err := networkutils.GetLocalIP()
```
abhinavmpandey08, line 295: Can you move the infer logic to line 303 so the IP is inferred only if the `tinkerbellBootstrapIp` is empty?

gwesterfieldjr: That's what I had earlier but I felt Chris asked for the opposite. Lol @chrisdoherty4

chrisdoherty4: We need the IP for templating. To avoid lazy state initialization I asked for this to be handled much earlier else we risk having [more] method ordering requirements when using the provider.

abhinavmpandey08 (approved):

chrisdoherty4 (approved):

PR #2377, opened 2022-06-09
Implement feature to rerun upgrade command

taneyland:
*Issue #, if available:*
#2185 

*Description of changes:*
Design doc: #2330 

This PR completes the MVP implementation for rerunning the upgrade command upon failure

*Testing (if applicable):*
Tested with upgrade (base, gitops enabled, and with additional workload cluster)


pkg/task/task.go
```diff
@@ -146,3 +207,52 @@ type CheckpointInfo struct {
...
+	return CheckpointInfo{
+		CompletedTasks: make(map[string]*CompletedTask),
+	}
+}
+
+func (c CheckpointInfo) taskCompleted(name string, completedTask *CompletedTask) {
+	c.CompletedTasks[name] = completedTask
+}
+
+func GetCheckpointFile(file string) (*CheckpointInfo, error) {
```
g-gaston, line 221: do these method need to be public?

pkg/task/task.go
```diff
@@ -146,3 +207,52 @@ type CheckpointInfo struct {
...
+	checkpointYaml, err := yaml.Marshal(taskCheckpoint)
+	if err != nil {
+		return nil
+	}
+	return yaml.Unmarshal(checkpointYaml, config)
+}
+
+func setupCheckpointInfo(commandContext *CommandContext, checkpointFileName string) (CheckpointInfo, error) {
+	checkpointInfo := newCheckpointInfo()
+	checkpointFilePath := filepath.Join(commandContext.Writer.Dir(), "generated", checkpointFileName)
```
g-gaston, line 246: I don't like hardcoding here `generated`, knowing the dir is a writer's concern
Is there any way to get that info from the writer?

pkg/task/task.go
```diff
@@ -146,3 +207,52 @@ type CheckpointInfo struct {
...
+	err = yaml.Unmarshal(content, checkpointInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed unmarshalling checkpoint: %v\n", err)
+	}
+
+	return checkpointInfo, nil
+}
+
+func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config interface{}) error {
+	checkpointYaml, err := yaml.Marshal(taskCheckpoint)
```
g-gaston, line 237: Why do we need to marshal and unmarshal?

pkg/task/task.go
```diff
@@ -146,3 +207,52 @@ type CheckpointInfo struct {
...
+	checkpointInfo := &CheckpointInfo{}
+	err = yaml.Unmarshal(content, checkpointInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed unmarshalling checkpoint: %v\n", err)
+	}
+
+	return checkpointInfo, nil
+}
+
+func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config interface{}) error {
```
g-gaston, line 236: Can we take checkpoint type instead of `interface{}`?

pkg/task/task.go
```diff
@@ -146,3 +207,52 @@ type CheckpointInfo struct {
...
+func setupCheckpointInfo(commandContext *CommandContext, checkpointFileName string) (CheckpointInfo, error) {
+	checkpointInfo := newCheckpointInfo()
+	checkpointFilePath := filepath.Join(commandContext.Writer.Dir(), "generated", checkpointFileName)
+
+	if _, err := os.Stat(checkpointFilePath); err == nil {
+		checkpointFile, err := GetCheckpointFile(checkpointFilePath)
+		if err != nil {
+			return checkpointInfo, err
+		}
+		if checkpointFile.CompletedTasks != nil {
```
g-gaston, line 253: curious why is this check necessary, can't we just make the assignment?

g-gaston (commented):
A few questions:
* Can we add more tests to the task runner with checkpoint?
* Can we add a test for the upgrade workflow with the checkpoint feature enabled?
* Is this the complete implementation or there something else that will come later? Could we add that info to the PR description?

pkg/task/task.go
```diff
@@ -126,6 +131,49 @@ func (pr *taskRunner) RunTask(ctx context.Context, commandContext *CommandContex
...
+				return fmt.Errorf("restoring checkpoint info: %v", err)
+			}
+			task = nextTask
+			continue
+		}
+		logger.V(4).Info("Task start", "task_name", task.Name())
+		commandContext.Profiler.SetStartTask(task.Name())
+		nextTask := task.Run(ctx, commandContext)
+		commandContext.Profiler.MarkDoneTask(task.Name())
+		commandContext.Profiler.logProfileSummary(task.Name())
```
vivek-koppuru, line 163: Why do we have some methods in the Profiler as exported and some as unexported funcs?

pkg/workflows/upgrade.go
```diff
@@ -479,10 +526,15 @@ func (s *moveManagementToWorkloadTask) Name() string {
...
 
 func (s *moveManagementToWorkloadTask) Checkpoint() *task.CompletedTask {
-	return nil
+	return &task.CompletedTask{
+		Checkpoint: nil,
+	}
 }
 
 func (s *moveManagementToWorkloadTask) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
+	if !commandContext.ManagementCluster.ExistingManagement {
```
vivek-koppuru, line 535: When are we toggling this during the workflow if we change the management from the workload cluster? Or is this to mainly support the self-management cluster + workload cluster case?

pkg/task/task.go
taneyland: Added a new method in the writer to return the temp directory

pkg/task/task.go
taneyland: When reading from a yaml file, there isn't a direct way in Go to do a type conversion from `interface{}` to the desired type. We use `interface{}` because the TaskCheckpoint type will vary depending on what's needed for a specific task. So the known workaround for this is to marshal & unmarshal it into the checkpoint type

pkg/task/task.go
taneyland: We are passing in the checkpoint type as the second parameter when we actual call this method, so it unmarshals into that type. We just have it as `interface{} `type in the function definition so that we can accept whatever checkpoint type was needed for the task

pkg/workflows/upgrade.go
taneyland: Yes this is to support the self-management cluster + workload cluster case

pkg/task/task.go
g-gaston: I was suggesting this
```suggestion
func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config TaskCheckpoint) error {
```

pkg/task/task.go
g-gaston: Got it
Could we write a comment explaining that?

chrisdoherty4:
I think we're introducing too much additional responsibility to the `taskRunner` which already has too much responsibility. With this implementation, the `taskRunner` is tightly bound to the checkpoint capability and is aware of too much checkpoint implementation detail:

- how to load checkpoint information.
- how to save checkpoint information.
- how to format the checkpoint data for save.
- how to inspect/extract checkpoint data from its in-memory representation.
- restoration establishes what task to run next (the existing model where tasks determine what task comes next is bad, if we can avoid it we should).

Have we considered a design that introduces a `Checkpointer` that wraps up some of these concerns instead of extending `taskRunner`?

pkg/workflows/delete.go
```diff
@@ -21,13 +21,14 @@ type Delete struct {
 }
 
 func NewDelete(bootstrapper interfaces.Bootstrapper, provider providers.Provider,
-	clusterManager interfaces.ClusterManager, addonManager interfaces.AddonManager,
+	clusterManager interfaces.ClusterManager, addonManager interfaces.AddonManager, writer filewriter.FileWriter,
```
chrisdoherty4, line 24: What did the delete workflow need a `FileWriter` for?

pkg/workflows/upgrade.go
```diff
@@ -119,37 +120,50 @@ func (s *setupAndValidateTasks) Run(ctx context.Context, commandContext *task.Co
...
 
 func (s *setupAndValidateTasks) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
-	return &updateSecrets{}, nil
+	s.FirstRunComplete = true
+	return s.Run(ctx, commandContext), nil
 }
 
 func (s *setupAndValidateTasks) Checkpoint() *task.CompletedTask {
-	return nil
+	return &task.CompletedTask{
```
chrisdoherty4, line 164: I took a peek at the `CompletedTask` and what not and it feels a little redundant because its a struct with a single field of type `interface{}`. Can we not use `interface{}`?

pkg/workflows/upgrade.go
```diff
@@ -377,7 +409,7 @@ func (s *installCAPITask) Run(ctx context.Context, commandContext *task.CommandC
 	err := commandContext.ClusterManager.InstallCAPI(ctx, commandContext.ClusterSpec, commandContext.BootstrapCluster, commandContext.Provider)
 	if err != nil {
 		commandContext.SetError(err)
-		return &deleteBootstrapClusterTask{}
+		return &CollectMgmtClusterDiagnosticsTask{}
```
chrisdoherty4, line 412: Was this a bug fix or something?

taneyland: No, we did this because we no longer want to delete the bootstrap cluster if installing CAPI fails because we will now be able to resume the operation at this point

pkg/workflows/upgrade.go
taneyland: We could. However, having it as a struct will allow us to build on top of it later if we want to retain even more information about the completed task

pkg/workflows/upgrade.go
```diff
@@ -119,37 +120,50 @@ func (s *setupAndValidateTasks) Run(ctx context.Context, commandContext *task.Co
...
+	result := []validations.Validation{
 		func() *validations.ValidationResult {
 			return &validations.ValidationResult{
 				Name: fmt.Sprintf("%s Provider setup is valid", commandContext.Provider.Name()),
 				Err:  commandContext.Provider.SetupAndValidateUpgradeCluster(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec),
 			}
 		},
-		func() *validations.ValidationResult {
+	}
+	if !s.FirstRunComplete {
```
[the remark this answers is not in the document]
taneyland: This was put into place because some preflight validations could fail depending on where the CLI failed & where we are attempting to resume it. A future addition of this feature is to add more validations to catch instances where the user changed a field between runs

pkg/workflows/delete.go
taneyland: We will need it eventually if we decide to add a checkpoint  feature for that command but i can remove it for now

taneyland:
> I think we're introducing too much additional responsibility to the `taskRunner` which already has too much responsibility. With this implementation, the `taskRunner` is tightly bound to the checkpoint capability and is aware of too much checkpoint implementation detail:
> 
>     * how to load checkpoint information.
> 
>     * how to save checkpoint information.
> 
>     * how to format the checkpoint data for save.
> 
>     * how to inspect/extract checkpoint data from its in-memory representation.
> 
>     * restoration establishes what task to run next (the existing model where tasks determine what task comes next is bad, if we can avoid it we should).
> 
> 
> Have we considered a design that introduces a `Checkpointer` that wraps up some of these concerns instead of extending `taskRunner`? At a minimum, I think we can get rid of loading checkpoint files during task running. That really isn't something a task runner should care about. It should only care about loading and creating checkpoints.

It was mentioned before that we should consider refactoring the task runner. Since that effort is outside of the scope of this task, we decided that the most straight-forward way to go about this is by extending the task runner. I think that attempting wrap up some of these concerns would still require some amount of refactoring the task runner & I think that effort should be its own entity (with its own formal design doc & what not). I would also like to mention that this is considered to be a temporary solution for this feature. Ideally, we would like to use something like a controller to orchestrate this process with a reconciliation loop. Open to discuss more offline

pkg/task/task.go
```diff
@@ -102,39 +108,114 @@ func (pp *Profiler) logProfileSummary(taskName string) {
...
+				task = nextTask
+				continue
+			}
+		}
 		logger.V(4).Info("Task start", "task_name", task.Name())
 		commandContext.Profiler.SetStartTask(task.Name())
 		nextTask := task.Run(ctx, commandContext)
 		commandContext.Profiler.MarkDoneTask(task.Name())
 		commandContext.Profiler.logProfileSummary(task.Name())
+		if commandContext.OriginalError == nil && pr.withCheckpoint {
```
g-gaston, line 165: same here, can't we execute this even if checkpoint is disabled?
For me, checkpoint disabled just means we don't read the checkpoint file at the beginning.

pkg/task/task.go
```diff
@@ -102,39 +108,114 @@ func (pp *Profiler) logProfileSummary(taskName string) {
...
+			checkpointInfo.taskCompleted(task.Name(), task.Checkpoint())
+		}
 		task = nextTask
 	}
+	if commandContext.OriginalError != nil && pr.withCheckpoint {
```
g-gaston, line 170: same about `pr.withCheckpoint`

pkg/task/task.go
```diff
@@ -146,3 +227,39 @@ type CheckpointInfo struct {
...
+		return nil, fmt.Errorf("failed unmarshalling checkpoint: %v\n", err)
+	}
+
+	return checkpointInfo, nil
+}
+
+// When reading from a yaml file, there isn't a direct way in Go to do a type conversion from interface{} to the desired type.
+// We use interface{} because the TaskCheckpoint type will vary depending on what's needed for a specific task.
+// The known workaround for this is to marshal & unmarshal it into the checkpoint type.
+func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config TaskCheckpoint) error {
```
g-gaston, line 259: nit: can we add a doc comment line on top of this describing what the method does?

pkg/task/task_test.go
```diff
@@ -56,3 +67,68 @@ func TestTaskRunnerRunTask(t *testing.T) {
...
+		{
+			name: "Task runs and next Task is triggered and profiles are captured",
+			fields: fields{
+				tasks: []task.Task{cleanTaskA, cleanTaskB, cleanTaskC},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := os.Setenv(features.CheckpointEnabledEnvVar, "true"); err != nil {
```
g-gaston, line 117: can we use `t.SetEnv`?

pkg/workflows/upgrade.go
```diff
@@ -145,11 +150,24 @@ func (s *setupAndValidateTasks) Name() string {
 }
 
 func (s *setupAndValidateTasks) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
+	if err := commandContext.Provider.SetupAndValidateUpgradeCluster(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec); err != nil {
+		commandContext.SetError(err)
+		return nil, err
+	}
+	logger.Info(fmt.Sprintf("%s Provider setup is valid", commandContext.Provider.Name()))
+	currentSpec, err := commandContext.ClusterManager.GetCurrentClusterSpec(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec.Cluster.Name)
```
g-gaston, line 158: Why have we moved this here?

pkg/workflows/upgrade_test.go
```diff
@@ -114,9 +119,11 @@ func newUpgradeManagedClusterTest(t *testing.T) *upgradeTestSetup {
 	return tt
 }
 
-func (c *upgradeTestSetup) expectSetup() {
+func (c *upgradeTestSetup) expectSetup(expectedCluster *types.Cluster) {
```
g-gaston, line 122: why do we need to pass expected cluster? isn't it always the same one? `tt.managementCluster`

pkg/filewriter/writer.go
```diff
@@ -51,6 +51,10 @@ func (t *writer) Dir() string {
 	return t.dir
 }
 
+func (t *writer) TempDir() string {
+	return filepath.Join(t.dir, DefaultTmpFolder)
```
g-gaston, line 55: this seems to be duplicated a few times, can we just store it in the struct?

pkg/filewriter/writer.go
```diff
@@ -51,6 +51,10 @@ func (t *writer) Dir() string {
 	return t.dir
 }
 
+func (t *writer) TempDir() string {
```
g-gaston, line 54: Test?

pkg/workflows/upgrade_test.go
taneyland: No it changes for unit tests that upgrade additional workload clusters

pkg/task/task.go
taneyland: Keeping it here because we shouldn't attempt to save the checkpoint & write it to a file if that feature isn't enabled

pkg/task/task.go
g-gaston: Ahh interesting
IMO the task runner is already fully aware of the checkpoint logic. If this is a concern, we can address it and move it somewhere else.

Otherwise, I would vote to remove the check. I think the only thing we should avoid doing when `withCheckpoint` is false is to restore the checkpoint info from the file at the beginning. The rest can stay the same, even writing to file.

pkg/task/task.go
g-gaston: Ah I see. Why do you think that? Writing to a file doesn't have any consequences at all. It's only reading from file and restoring tasks what affects the behavior of the program (unless I'm missing something).

pkg/workflows/upgrade_test.go
g-gaston: Just for my understanding, could you describe how and when it changes?
Happy to take this offline if it's too long :)

pkg/task/task.go
```diff
@@ -147,3 +222,41 @@ type CheckpointInfo struct {
...
+	return CheckpointInfo{
+		CompletedTasks: make(map[string]*CompletedTask),
+	}
+}
+
+func (c CheckpointInfo) taskCompleted(name string, completedTask *CompletedTask) {
+	c.CompletedTasks[name] = completedTask
+}
+
+func getCheckpointFile(file string) (*CheckpointInfo, error) {
```
g-gaston, line 236: maybe something like:
```suggestion
func readCheckpointFile(file string) (*CheckpointInfo, error) {
```

pkg/task/task.go
```diff
@@ -147,3 +222,41 @@ type CheckpointInfo struct {
...
+
+	return checkpointInfo, nil
+}
+
+/* When reading from a yaml file, there isn't a direct way in Go to do a type conversion from interface{} to the desired type.
+We use interface{} because the TaskCheckpoint type will vary depending on what's needed for a specific task. The known workaround
+for this is to marshal & unmarshal it into the checkpoint type. This UnmarshalTaskCheckpoint method marshals the received task checkpoint (type interface{})
+then unmarshalls it into the desired type.
+*/
+func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config TaskCheckpoint) error {
```
g-gaston, line 256: I was referring to this:
```suggestion
/* UnmarshalTaskCheckpoint unmarshals bla bla bla bla
When reading from a yaml file, there isn't a direct way in Go to do a type conversion from interface{} to the desired type.
We use interface{} because the TaskCheckpoint type will vary depending on what's needed for a specific task. The known workaround
for this is to marshal & unmarshal it into the checkpoint type. This UnmarshalTaskCheckpoint method marshals the received task checkpoint (type interface{})
then unmarshalls it into the desired type.
*/
func UnmarshalTaskCheckpoint(taskCheckpoint TaskCheckpoint, config TaskCheckpoint) error {
```

pkg/task/task.go
```diff
@@ -103,39 +109,108 @@ func (pp *Profiler) logProfileSummary(taskName string) {
...
+	task           Task
+	writer         filewriter.FileWriter
+	withCheckpoint bool
+}
+
+type TaskRunnerOpt func(*taskRunner)
+
+func WithCheckpointFile() TaskRunnerOpt {
+	return func(t *taskRunner) {
+		logger.V(4).Info("Checkpoint feature enabled", features.CheckpointEnabledEnvVar, features.IsActive(features.CheckpointEnabled()))
```
g-gaston, line 121: I'm confused, why are we logging the value of the feature gate?

pkg/workflows/upgrade_test.go
```diff
@@ -518,3 +523,69 @@ func TestUpgradeWorkloadRunSuccess(t *testing.T) {
...
+	test.expectCreateBootstrap()
+	test.expectMoveManagementToBootstrap()
+	test.expectUpgradeWorkloadToReturn(test.bootstrapCluster, test.workloadCluster, errors.New("failed upgrading"))
+	test.expectSaveLogs(test.workloadCluster)
+
+	err := test.run()
+	if err == nil {
+		t.Fatal("Upgrade.Run() err = nil, want err not nil")
+	}
+	features.ClearCache()
```
g-gaston, line 547: why are we calling this here?

pkg/task/task_test.go
```diff
@@ -56,3 +67,66 @@ func TestTaskRunnerRunTask(t *testing.T) {
 		}
 	}
 }
+
+func TestTaskRunnerRunTaskWithCheckpoint(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
g-gaston, line 73: I don't think this is needed anymore

pkg/workflows/upgrade_test.go
```diff
@@ -518,3 +523,69 @@ func TestUpgradeWorkloadRunSuccess(t *testing.T) {
...
+}
+
+type testCheckpointContext struct {
+	checkpointEnabled          string
+	isCheckpointFeatureEnabled bool
+}
+
+func (tctx *testCheckpointContext) SaveContext() {
+	tctx.checkpointEnabled, tctx.isCheckpointFeatureEnabled = os.LookupEnv(features.CheckpointEnabledEnvVar)
+	os.Setenv(features.CheckpointEnabledEnvVar, "true")
```
g-gaston, line 582: Can we use `t.SetEnv` to simplify all of this?

pkg/workflows/upgrade_test.go
taneyland: Of course! Since I moved the code to get the current cluster spec to the `setup-and-validate` task, I had to move the `EXPECT()` for this method here in the unit tests. And when we are upgrading the workload clusters, we pass in the management cluster as the expected cluster here (for the other tests, we pass in the workload cluster). So it's the same logic that was in the test before, just moved to the setup task.

pkg/task/task.go
taneyland: True, there aren't any actual consequences. I didn't want to write to the file because I didn't want to have an unnecessary file in the temp directory if the operation failed, potentially adding confusion from the user's POV if they didn't enable the checkpoint feature. But once we remove the feature flag, this will be the default behavior so I'm ok with removing the bool check here too

pkg/workflows/upgrade_test.go
taneyland: I believe you told me about how we need to clear the cache sometimes for unit tests with env vars. I was running into odd behavior without clearing the cache for this unit test. After I added this, it began to work as expected. We do this in other unit tests with gated features as well.

pkg/task/task.go
```diff
@@ -137,6 +185,19 @@ func NewTaskRunner(task Task, writer filewriter.FileWriter) *taskRunner {
 	}
 }
 
+func (pr *taskRunner) saveCheckpoint(checkpointInfo CheckpointInfo, filename string) error {
```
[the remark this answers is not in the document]
g-gaston: Gentle bump :)

PR #2382, opened 2022-06-09
Enhance package controller

a-cool-train:
*Description of changes:*
- Update packagecontroller to update sourceRegistry field during installation
- Install packagecontroller takes cluster spec instead of kube-version
- Update curated-packages version.
- Fix an existing install packages bug
- Fix an existing custom registry bug

*Testing (if applicable):*


cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -7,21 +7,24 @@ import (
...
+	fileName string
 }
 
 var ico = &installControllerOptions{}
 
 func init() {
 	installCmd.AddCommand(installPackageControllerCommand)
-	installPackageControllerCommand.Flags().StringVar(&ico.kubeVersion, "kube-version", "", "Bundle version to use")
-	if err := installPackageControllerCommand.MarkFlagRequired("kube-version"); err != nil {
+	installPackageControllerCommand.Flags().StringVarP(&ico.fileName, "filename", "f", "", "Filename that contains EKS-A cluster configuration")
```
vivek-koppuru, line 26: This is a breaking change, so do we want to go through a deprecation path where we give a warning message and then slowly remove it after 2 releases? Or are we okay with the consequences?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -37,31 +40,43 @@ var installPackageControllerCommand = &cobra.Command{
...
 		return err
 	}
+
+	registryEndpoint := ""
+	if clusterSpec.Spec.RegistryMirrorConfiguration != nil {
+		registryEndpoint = clusterSpec.Spec.RegistryMirrorConfiguration.Endpoint
+	}
+
 	helmChart := versionBundle.PackageController.HelmChart
+	imageUrl := urls.ReplaceHost(helmChart.Image(), registryEndpoint)
```
vivek-koppuru, line 74: Does this work if the registryEndpoint is empty string?

pkg/curatedpackages/curatedpackages.go
```diff
@@ -123,7 +125,7 @@ func Push(ctx context.Context, ref, fileName string, fileContent []byte) error {
 		return fmt.Errorf("creating registry: %w", err)
 	}
 	memoryStore := content.NewMemory()
-	desc, err := memoryStore.Add(fileName, "", fileContent)
+	desc, err := memoryStore.Add("bundle.yaml", "", fileContent)
```
vivek-koppuru, line 128: This should be a constant

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: Yes, ReplaceHost returns the same url if empty string.

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: We still haven't gone GA, I think this should be fine but I'll check with the team.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.16.2
 	github.com/aws/aws-sdk-go-v2/config v1.15.3
 	github.com/aws/aws-sdk-go-v2/service/ec2 v1.34.0
-	github.com/aws/eks-anywhere-packages v0.1.1
+	github.com/aws/eks-anywhere-packages v0.1.10
```
TerryHowe, line 10: But our release goes to 11. It would be convenient to have this in another PR, but no matter

TerryHowe (approved):

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
a-cool-train: Checked with the team and the deprecation process is not necessary. I'll update the docs to reflect this change though.

PR #2384, opened 2022-06-09
Added Bare Metal content to EKS-A/EKS comparison

chrisnegus:
*Description of changes:*
Added content to the table on [Compare EKS Anywhere and EKS Features](https://anywhere.eks.amazonaws.com/docs/concepts/eksafeatures/) table to describe EKS Anywhere Bare Metal features.

### DIFF
--- a/docs/content/en/docs/concepts/eksafeatures.md
+++ b/docs/content/en/docs/concepts/eksafeatures.md
@@ -29,7 +29,7 @@ To learn more about Amazon EKS, see [Amazon Elastic Kubernetes Service](https://
 | Cluster updates        | Manual CLI updates for control plane. Flux supported rolling updates for data plane. | Manual CLI updates for control plane.                | Managed in-place updates for control plane and managed rolling updates for data plane.                       |
 ||||
 | **Compute** |||
-| Compute options | Bare metal machines | VMware vSphere | Amazon EC2, AWS Fargate | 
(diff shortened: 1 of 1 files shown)


vipin-mohan:
Notes:
- Let's have one consolidated column for EKS-A (i.e., combine vSphere and Bare metal). The title of the column can just be "Amazon EKS Anywhere"
- "Cluster updates" --> Let's go with the text in the current vSphere column
- Compute options --> VMware vSphere, Bare metal servers
- Supported node operating systems --> Bottlerocket, Ubuntu (We will have support for both operating systems for vSphere and bare metal). Also, please note that Bottlerocket does not have an upper case "R"
- GitOps - Flux controller (even at GA for bare metal, Flux is supported. It will be manual effort to install (roadmap)

chrisnegus:
> Notes:
> 
> * Let's have one consolidated column for EKS-A (i.e., combine vSphere and Bare metal). The title of the column can just be "Amazon EKS Anywhere"
Done

> * "Cluster updates" --> Let's go with the text in the current vSphere column
Okay, but I added "vSphere only" to the flux line.

> * Compute options --> VMware vSphere, Bare metal servers
Done

> * Supported node operating systems --> Bottlerocket, Ubuntu (We will have support for both operating systems for vSphere and bare metal). Also, please note that Bottlerocket does not have an upper case "R"
Done.

> * GitOps - Flux controller (even at GA for bare metal, Flux is supported. It will be manual effort to install (roadmap)
Done

PR #2389, opened 2022-06-10
Add normalizations and lower case MAC address

chrisdoherty4:
When reading hardware data for Tinkerbell we need to apply normalizations so we can perform comparison operations. MAC Addresses are a culprit for incorrect comparisons because comparators don't take into consideration case insensitivty of a MAC.

We're lower-casing MAC as it seems to be lower-case more often than not.

This won't protect us against all failures. A real fix would be to edit all locations within Tinkerbell that perform a MAC comparison to be case-insensitive.

### DIFF
--- a/pkg/providers/tinkerbell/hardware/csv.go
+++ b/pkg/providers/tinkerbell/hardware/csv.go
@@ -28,16 +28,6 @@ func NewCSVReader(r io.Reader) (CSVReader, error) {
 	return CSVReader{reader: reader}, nil
 }
 
-// NewCSVReaderFromFile creates a CSVReader instance that reads from path.
-func NewCSVReaderFromFile(path string) (CSVReader, error) {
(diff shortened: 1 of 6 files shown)


pkg/providers/tinkerbell/hardware/csv.go
```diff
@@ -46,3 +36,19 @@ func (cr CSVReader) Read() (Machine, error) {
 	}
 	return machine.(Machine), nil
 }
+
+// NewCSVReaderFromFile creates a MachineReader instance backed by a CSVReader reading from path
```
abhinavmpandey08, line 40:
```suggestion
// NewNormalizedCSVReaderFromFile creates a MachineReader instance backed by a CSVReader reading from path
```

abhinavmpandey08 (approved):
/approve

PR #2392, opened 2022-06-10
Add customized configurations doc

a-cool-train:
*Description of changes:*
- Include Customized configurations for curated packages.

### DIFF
--- a/designs/customized-configurations.md
+++ b/designs/customized-configurations.md
@@ -1,0 +1,92 @@
+# Curated Packages Customized Configuration
+
+## Problem
+
+Currently, the eks anywhere CLI provides generate and install commands. Both these commands’ aim is to provide the ability for users to install a curated package in the cluster. Currently, both commands do not support the ability to override any configurations that would enable the user to customize the curated package.
+
+## Tenets
+
+1. *Simple*
+2. *Consistent*
+
+## Goals and Objectives
+
+As a curated packages user, I want to:
+
+* Provide custom configurations to override default configurations from the CLI
+
+## Possible Solutions
+
+### Provide separate key value pairs from the CLI
+
+The first solution would be to provide separate key value pairs through the cli similar to helm (https://all.docs.genesys.com/PrivateEdition/Current/PEGuide/HelmOverrides)
+
+*Sample*
+
+```bash
+$ eksctl anywhere install package harbor --set key=key --set key2=key2
+```
+Pros:
+
+1. The user doesn’t need to make one bulk configurations
+
+Cons:
+
+1. It might get cumbersome as the configurations become a lot more.
+
+### Provide consolidated key value pairs from the CLI
+
+Similar to the first solution, the solution would be to consolidate all the key value pairs into one but separated by a delimiter, similar to cobra (https://github.com/spf13/pflag/pull/133)
+
+*Sample*
+
+```bash
+$ eksctl anywhere install package harbor --config=key1=key1, key2=key2
+```
+
+Pros:
+
+1. The configurations are consolidated into one.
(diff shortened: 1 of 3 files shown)


designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
+# Curated Packages Customized Configuration
+
+## Problem
+
+Currently, the eks anywhere CLI provides generate and install commands. Both these commands’ aim is to provide the ability for users to install a curated package in the cluster. Currently, both commands do not support the ability to override any configurations that would enable the user to customize the curated package.
```
drewvanstone, line 5:
```suggestion
Currently, the eks anywhere CLI provides `generate` and `install` commands. Both these commands aim is to provide the ability for users to install a curated package in the cluster. Currently, both commands do not support the ability to override any configurations that would enable the user to customize the curated package.
```

drewvanstone (approved):

designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
+# Curated Packages Customized Configuration
+
+## Problem
+
+Currently, the eks anywhere CLI provides `generate` and `install` commands. Both these commands aim is to provide the ability for users to install a curated package in the cluster. Currently, both commands do not support the ability to override any configurations that would enable the user to customize the curated package.
```
vivek-koppuru, line 5:
```suggestion
Currently, the eks anywhere CLI provides `generate` and `install` commands. Both these commands aim to provide the ability for users to install a curated package in the cluster. However, both commands do not support the ability to override any configurations that would enable the user to customize the curated package.
```

designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Tenets
+
+1. *Simple*
```
vivek-koppuru, line 9: What about these tenets does this feature follow?

designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
...
+$ eksctl anywhere install package harbor --config=key1=key1, key2=key2
+```
+
+Pros:
+
+1. The configurations are consolidated into one.
+
+Cons:
+
+1. It might also be cumbersome when providing more configurations.
```
vivek-koppuru, line 53: Both have the same cons

designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Provide key value pair through a file and pass the file to the CLI
+
+This solution is different than both the previous solution given that it takes a file for the configuration
+
+*Sample*
+
+```bash
+$ cat config.txt
+secretKey=use-a-secret-key
```
vivek-koppuru, line 63: We should create a custom resource for this and specify the file to the command instead of introduce a new format.

designs/customized-configurations.md
```diff
@@ -0,0 +1,92 @@
...
+$ eksctl anywhere install package harbor --config ./config.txt
+```
+
+Pros:
+
+1. It can handle a growing number of configurations.
+2. It lets the user reuse the same file.
+
+## Proposed Solution
+Since the amount of configurations required for curated packages is a limited set, we will be implementing [Provide separate key value pairs from the CLI](#Provide-separate-key-value-pairs-from-the-CLI) option. In the future, we will include the [Provide key value pair through a file and pass the file to the CLI](#Provide-key-value-pair-through-a-file-and-pass-the-file-to-the-CLI)
```
vivek-koppuru, line 75: I am okay with this approach as a first step if we are clear that it is for overriding default configuration, otherwise we should stick to passing in a file.

PR #2396, opened 2022-06-13
Improve error when hardware CSV has missing required columns

chrisdoherty4:
Previously users would receive an error about missing data when an entire column is missing. This improves the error to explicitly call out a missing column.


abhinavmpandey08 (approved):
/approve

PR #2397, opened 2022-06-13
Reject cluster operations with external etcd for bare metal

chrisdoherty4:
For GA, bare metal doesn't support external etcd. This change explicitly rejects invocations that use an external etcd config.

### DIFF
--- a/pkg/providers/tinkerbell/testdata/expected_results_cluster_tinkerbell_cp_external_etcd.yaml
+++ b/pkg/providers/tinkerbell/testdata/expected_results_cluster_tinkerbell_cp_external_etcd.yaml
@@ -22,10 +22,6 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
     kind: TinkerbellCluster
     name: test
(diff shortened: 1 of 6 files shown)


pkg/providers/tinkerbell/testdata/expected_results_cluster_tinkerbell_cp_external_etcd.yaml
```diff
@@ -22,10 +22,6 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
     kind: TinkerbellCluster
     name: test
-  managedExternalEtcdRef:
```
maxdrib, line 25: it would be less confusing to rename this file or remove the test case if external etcd doesn't apply

chrisdoherty4: It will apply eventually- this is temporary.

jacobweinstock (approved):

PR #2399, opened 2022-06-13
Multi-endpoint design for cloudstack provider

maxdrib:
*Issue #, if available:*

*Description of changes:*
This design encapsulates a proposed approach for refactoring the CloudStack provider to support creating eks-a clusters across multiple Cloudstack endpoints with different sets of credentials.

*Testing (if applicable):*


designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,178 @@
...
+Domain string `json:"domain"`
+// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+Zones []CloudStackZone `json:"zones"`
+// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
+Account string `json:"account,omitempty"`
+// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+ManagementApiEndpoint string `json:"managementApiEndpoint"`
+```
+
+We would instead propose to remove all the existing attributes and instead, simply include a list of FailureDomain objects, where each FailureDomain object looks like
```
jiayiwang7, line 64: would it make sense to treat each `FailureDomain` as a datacenter? If so, is the reason we don't have multiple cloudstackdatacenterconfig purely engineering efforts too high?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,178 @@
...
+Domain string `json:"domain"`
+// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+Zone CloudStackZone `json:"zone"`
+// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
+Account string `json:"account,omitempty"`
+// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+ManagementApiEndpoint string `json:"managementApiEndpoint"`
+```
+
+and we would parse these resources and pass them into CAPC by modifying the templates we have currently implemented. We can then use this new model to read in credentials, perform pre-flight checks, plumb data to CAPC, and support upgrades in the controller. The goal would be to make these new resources backwards compatible via code
```
jiayiwang7, line 79: curious how to map this failureDomain list to CAPC? Can we maybe have some examples here?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+In practice, the pseudocode would look like:
+
+for failureDomain in failureDomains:
+  for machineConfig in machineConfigs:
+    validate resource presence with the failureDomain's instance of the CloudMonkey executable
+
+### Cloudstack credentials
+
+
+In a multi-endpoint Cloudstack cluster, each endpoint may have its own credentials. We propose that Cloudstack credentials will be passed in via environment variable in the same way as they are currently,
```
jiayiwang7, line 109: what abt user specifies a file path to the .ini and pass it as ENV?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+Domain string `json:"domain"`
+// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+Zones []CloudStackZone `json:"zones"`
+// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
+Account string `json:"account,omitempty"`
+// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+ManagementApiEndpoint string `json:"managementApiEndpoint"`
+```
+
+We would instead propose to remove all the existing attributes and instead, simply include a list of FailureDomain objects, where each FailureDomain object looks like
```
vivek-koppuru, line 64: These types of changes are breaking changes. Especially if we don't have a feature flag for Cloudstack, what about introducing a new object for this use case?

vivek-koppuru: How would this list look like here?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+We are also exploring converting the ini file to a yaml input file which contains a list of credentials and their associated endpoints. Either way, this environment variable would
+be passed along to CAPC and used by the CAPC controller just like it is currently.
+
+### Backwards Compatibility
+
+Our customer currently has clusters running with the old resource definition. In order to support backwards compatibility in the CloudstackDatacenterConfig resource, we can
+1. Make all the fields optional and see if customers have the old fields set or the new ones
+2. Introduce an eks-a version bump with conversion webhooks
+
+Between these two approaches, I would take the first and then deprecate the legacy fields in a subsequent release to simplify the code paths.
```
vivek-koppuru, line 149: I think we can still throw a warning if new clusters are created with the old fields.

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+
+simple flow cluster creation/deletion across multiple Cloudstack API endpoints:
+
+* create a management+workload cluster spanning multiple Cloudstack API endpoints
+* delete cluster
+
+## Other approaches explored
+
+Another direction we can go to support this feature is to refactor the entire EKS-A codebase so that instead of all the failure domains existing inside the CloudstackDatacenterConfig
+object, each CloudstackDatacenterConfig itself corresponds with a single failure domain. Then, the top level EKS-A Cluster object could be refactored to have a list of DatacenterRefs instead
```
vivek-koppuru, line 176: It may actually provide value as we talked about wanting to support multiple environments in the future. However, I can see why it's a bigger step with less data for a specific provider.

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Introduction
+
+**Problem:** 
+
+The mangement API endpoint for Cloudstack is a potential points of failure. If one endpoint goes down, then control of all of everything goes down. So, we want to spread our services across many Cloudstack endpoints and hosts to protect against that.
+
+For scalability, multiple Cloudstack endpoints will likely be required for storage and API endpoint throughput for our customer. Just one cluster creation generates probably a thousand API calls to ACS. There are many ways to mitigate this, but adding more Cloudstack hosts and endpoints is a fairly foolproof way to do so. Then, there’s the size and performance of the underlying database that each Cloudstack instance runs on.
+
+In CAPC, we are considering addressing the problem by relying on Failure Domains and distributing a cluster across the given ones. In order to support this functionality in EKS-A, we need to have a similar breakdown where an EKS-A cluster can span across multiple endpoints and Failure Domains.
```
pmotyka, line 11: Is there a section of the CAPI book to link to for more context on Failure Domains?

designs/cloudstack-multiple-endpoints.md
maxdrib: I think we could do that as well if you prefer. Although it might require some changes in the e2e test framework. I think there we're using the encoded value instead of any specific file. I'm not sure how/if we could download the file locally and then use it in the e2e tests but it does seem easier for the customer.

@vivek-koppuru any thoughts on this? I know we've discussed how to have customers pass data in at length.

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
+# Supporting Cloudstack clusters across endpoints
+
+## Introduction
+
+**Problem:** 
+
+The mangement API endpoint for Cloudstack is a potential points of failure. If one endpoint goes down, then control of all of everything goes down. So, we want to spread our services across many Cloudstack endpoints and hosts to protect against that.
+
+For scalability, multiple Cloudstack endpoints will likely be required for storage and API endpoint throughput for our customer. Just one cluster creation as many as a thousand API calls to ACS (estimated). There are many ways to support this scale, but adding more Cloudstack hosts and endpoints is a fairly foolproof way to do so. Then, there’s the size and performance of the underlying database that each Cloudstack instance runs on.
```
g-gaston, line 9:
```suggestion
For scalability, multiple Cloudstack endpoints will likely be required for storage and API endpoint throughput. Just one cluster creation triggers as many as a thousand API calls to ACS (estimated). There are many ways to support this scale, but adding more Cloudstack hosts and endpoints is a fairly foolproof way to do so. Then, there’s the size and performance of the underlying database that each Cloudstack instance runs on.
```

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+### Tenets
+
+* ****Simple:**** simple to use, simple to understand, simple to maintain
+* ****Declarative:**** intent based system, as close to a Kubernetes native experience as possible
+
+### Goals and Objectives
+
+As a Kubernetes administrator I want to:
+
+* Perform preflight checks when creating/upgrading clusters which span across multiple failure domains
```
g-gaston, line 22: nit: isn't this an implementation detail?
The user want's to create and upgrade across multiple "zones"
And part of that process involves validating all those zones before continuing

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+* Create EKS Anywhere clusters which span across multiple failure domains 
+* Upgrade EKS Anywhere clusters which span across multiple failure domains
+* Delete EKS Anywhere clusters which span across multiple failure domains
+
+### Statement of Scope
+
+**In scope**
+
+* Add support for create/upgrade/delete of EKS-A clusters across multiple Cloudstack API endpoints
+* Add test environment for CI/CD e2e tests which can be used as a second Cloudstack API endpoint
```
g-gaston, line 32: this seems like it's in scope for the "project" but not for the design doc, right?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+* 
+
+**Future scope**
+
+* Multiple network support to handle IP address exhaustion within a zone
+
+## Overview of Solution
+
+We propose to take the least invasive solution of repurposing the CloudstackDataCenterConfig to point to multiple failure domains, each of which contains the necessary
+information for interacting with a Cloudstack failure domain. The assumption is that the necessary Cloudstack resources (i.e. image, computeOffering, ISOAttachment, etc.)
```
g-gaston, line 45: to clarify, what do we mean by assumption? That we won't validate that?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+Domain string `json:"domain"`
+// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+Zones []CloudStackZone `json:"zones"`
+// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
+Account string `json:"account,omitempty"`
+// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+ManagementApiEndpoint string `json:"managementApiEndpoint"`
+```
+
+We would instead propose to remove all the existing attributes and instead, simply include a list of FailureDomain objects, where each FailureDomain object looks like
```
g-gaston, line 64: It seems (correct me if I'm missing something) that there is a direct translation between the old format and the new one

I would vote to keep the old fields and write a transformer in the defaulter code that would populate the new fields and cleanup the old ones. Up to you if a warning is printed in the cli if the old format is used as input. This should be added as a mutation webhook as well.

That keeps the api backwards compatible and reduces the friction of introducing a new format. And at the same time, it simplifies the rest of the codebase so it only has to deal with the new format.

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,177 @@
...
+3. Cloudstack zone
+4. Cloudstack account
+
+You can find more information about these Cloudstack resources [here](http://docs.cloudstack.apache.org/en/latest/conceptsandterminology/concepts.html#cloudstack-terminology)
+
+### `CloudstackDatacenterConfig` Validation
+
+With the multi-endpoint system for the Cloudstack provider, customers reference a CloudstackMachineConfig and it's created across multiple failure domains. The implication
+is that all the Cloudstack resources such as image, ComputeOffering, ISOAttachment, etc. must be available in *all* the failure domains, or all the Cloudstack endpoints,
+and these resources must be referenced by name, not unique ID. This would mean that for each CloudstackMachineConfig, we have to make sure that all the prerequisite
```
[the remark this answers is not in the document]
maxdrib: I think we could include a preflight check which fails on the following conditions
1. >1 Cloudstack endpoint
2. Cloudstack datacenter config uses ID to reference some resource (zone, network, domain, account)
3. Cloudstack machineconfigs use ID to reference some resources (DiskOffering, ComputeOffering, template, affinitygroupids)

designs/cloudstack-multiple-endpoints.md
```diff
@@ -0,0 +1,175 @@
...
+
+We would instead propose to remove all the existing attributes and instead, simply include a list of FailureDomain objects, where each FailureDomain object looks like
+```
+// Domain contains a grouping of accounts. Domains usually contain multiple accounts that have some logical relationship to each other and a set of delegated administrators with some authority over the domain and its subdomains
+// This field is considered as a fully qualified domain name which is the same as the domain path without "ROOT/" prefix. For example, if "foo" is specified then a domain with "ROOT/foo" domain path is picked.
+// The value "ROOT" is a special case that points to "the" ROOT domain of the CloudStack. That is, a domain with a path "ROOT/ROOT" is not allowed.
+//
+Domain string `json:"domain"`
+// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
+Zone CloudStackZone `json:"zone"`
```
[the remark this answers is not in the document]
jweite-amazon: Yes, exactly.  The customer ask is to enhance the CAPC Failure Domain, which currently is equivalent to ACS Zone to be Endpoint+Zone+Domain+Account.

designs/cloudstack-multiple-endpoints.md
jweite-amazon: By way of comparison, CAPV supports CAPI Failure Domains via node labeling (see [Node Zones/Regions Topology](https://cloud-provider-vsphere.sigs.k8s.io/cloud_provider_interface.html)).  None the less, EKS-A expects those zones/regions to be managed through a single Datacenter.

designs/cloudstack-multiple-endpoints.md
jweite-amazon: https://cluster-api.sigs.k8s.io/developer/providers/v1alpha2-to-v1alpha3.html?highlight=failure%20domain#optional-support-failure-domains

vivek-koppuru (approved):

designs/cloudstack-multiple-endpoints.md
g-gaston: I get what you are saying, however I don't think this is a user story. Cluster admins can't "perform validations", at least for now. It's the cli the ones that does it to prevent starting a cluster creation on a misconfigured environment.

I agree that this should be part of the scope of this project. But that's not what this section is about.

If you are firm on including this as part of the user stories, we could maybe rephrase it:

```suggestion
* My cluster and environment across multiple failure domains to be validated before creating/upgrading clusters
```

designs/cloudstack-multiple-endpoints.md
g-gaston: What do you think about a subsection under testing?
If we already use it to define what needs to be tested, it seems fair to use it to define the special infrastructure needed for those tests to be valid as well.

designs/cloudstack-multiple-endpoints.md
g-gaston: Ahh ok, what about:

```suggestion
We propose to take the least invasive solution of repurposing the CloudstackDataCenterConfig to point to multiple Availability Zones, each of which contains the necessary
information for interacting with a Cloudstack failure domain. I order for this to work, all the necessary Cloudstack resources (i.e. image, computeOffering, diskOffering, network, etc.)
will need to be available on *all* the Cloudstack API endpoints. We will validate this prior to create/upgrade.
```

designs/cloudstack-multiple-endpoints.md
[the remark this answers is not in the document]
g-gaston: That would work, for sure.
However, is the benefit worth all the effort? I'm not necessarily concerned about writing that validation, but just about the burden of maintaining support for names in the codebase, but only partially.

That seems like extra unnecessary work.

Just for my understanding, why would specifying the name wouldn't work for multi availability zone? Shouldn't it work as long as the resource exists with the same name in all zones?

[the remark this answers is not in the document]
g-gaston: Oh I got it in reverse :(

Yeah, leaving the validation as is works for me, great idea. It still prevents users from submitting an invalid configuration and it doesn't add extra work neither now nor for maintenance.

Whenever we move to `v1alpha2`, we should probably consider just removing the support for ID.

g-gaston (approved):

designs/cloudstack-multiple-endpoints.md
jiayiwang7: for snow we pass in file path as ENV vars

maxdrib: From some discussions internally, it feels like storing credentials in a transient environment variable is more secure than in a file written to disk. That is why we prefer to go the env var route, although it is slightly more effort for the customer

PR #2404, opened 2022-06-14
Support multiple endpoints for cloudstack preflight check

wongni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -29,42 +31,65 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
...
 	}, nil
 }
 
 type CloudStackExecConfig struct {
+	Instances []CloudStackInstanceConfig
+	VerifySsl string
+	Timeout   string
+}
+
+type CloudStackInstanceConfig struct {
```
maxdrib, line 90: I think CloudStackProfileConfig makes more sense here

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -13,6 +13,8 @@ const (
 	EksacloudStackCloudConfigB64SecretKey = "EKSA_CLOUDSTACK_B64ENCODED_SECRET"
 	CloudStackCloudConfigB64SecretKey     = "CLOUDSTACK_B64ENCODED_SECRET"
 	EksaCloudStackHostPathToMount         = "EKSA_CLOUDSTACK_HOST_PATHS_TO_MOUNT"
+	globalSectionName                     = "Global"
+	defaultSectionName                    = "DEFAULT"
```
maxdrib, line 17: why do we need this? If every ini file must have a DEFAULT section, could we use an existing constant instead of defining our own?

pkg/providers/cloudstack/decoder/decoder_test.go
```diff
@@ -42,130 +28,126 @@ func (tctx *testContext) restoreContext() {
...
-	tctx.restoreContext()
+	for _, tc := range tests {
+		t.Run(tc.name, func(tt *testing.T) {
+			var tctx testContext
+			tctx.backupContext()
+
+			g := NewWithT(t)
+			configString := test.ReadFile(t, tc.configFile)
+			encodedConfig := base64.StdEncoding.EncodeToString([]byte(configString))
+			os.Setenv(decoder.EksacloudStackCloudConfigB64SecretKey, encodedConfig)
```
maxdrib, line 140: instead of os.Setenv, could you try using tt.Setenv, like we do in https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_webhook_test.go#L1225?

maxdrib (commented):
1. Thanks for putting this together. We'll be tracking the work in https://github.com/aws/eks-anywhere/issues/2406
2. Another implementation to consider is instead of putting all the credentials/endpoints in a single instance of the cmk executable, if we should instead have multiple instances of the cmk executables, and each one would have its own config. I'm not sure which path is better, so I'll make a note of it in the design doc in https://github.com/aws/eks-anywhere/pull/2399 and await the team's input. Please have a look at that design when you get a chance and provide any feedback you may have

pkg/providers/cloudstack/testdata/cloudstack_config_multiple_instances.ini
```diff
@@ -0,0 +1,11 @@
+verify-ssl = false
```
maxdrib, line 1: Looks like this change would break CAPC actually. They're expecting everything under [Global]. https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/main/pkg/cloud/client.go#L56-L62. For backwards compatibility, can we keep the option of the global tag, and then add additional instances below it?

maxdrib (changes requested):

pkg/providers/cloudstack/testdata/cloudstack_config_multiple_instances.ini
wongni: If we want to keep the backward compatibility we should keep [Global] section as the first and mandatory failure domain and any additional ones should go under the global section. However, the old versions of capc would use only the global section as a single and only ACS management endpoint. A newer version of capc will treat each section as a failure domain. In this case, `verify-ssl` property can be shown in all sections and the values can be different each other. If we take this route creating a separate cmk config file per failure domain should be taken because `verify-ssl` is a global flag in a cmk config file.

maxdrib: I don't see any issues with having multiple cmk config files. That feels like an implementation detail that shouldn't affect the customer. It also gives them the flexibility to determine if they want to verify ssl for different failure domains independently.

pkg/dependencies/factory.go
```diff
@@ -48,7 +48,7 @@ type Dependencies struct {
 	DockerClient              *executables.Docker
 	Kubectl                   *executables.Kubectl
 	Govc                      *executables.Govc
-	Cmk                       *executables.Cmk
+	Cmks                      map[string]cloudstack.ProviderCmkClient
```
maxdrib, line 51: why do we need to arrange this as a map instead of a list? What additional value do we get from the labeling?

pkg/executables/cmk.go
```diff
@@ -440,15 +432,18 @@ func (c *Cmk) CleanupVms(ctx context.Context, clusterName string, dryRun bool) e
...
 	configFile, err := c.buildCmkConfigFile()
 	if err != nil {
-		return bytes.Buffer{}, fmt.Errorf("failed cmk validations: %v", err)
+		return stdout, fmt.Errorf("failed cmk validations: %v", err)
 	}
+
 	argsWithConfigFile := append([]string{"-c", configFile}, args...)
+	stdout, err = c.executable.Execute(ctx, argsWithConfigFile...)
 
-	return c.executable.Execute(ctx, argsWithConfigFile...)
```
maxdrib, line 451: we can keep this return statement from before, right?

pkg/executables/cmk_test.go
```diff
@@ -19,23 +19,46 @@ import (
...
+}
+
+var execConfigWithMultipleProfiles = decoder.CloudStackExecConfig{
+	Profiles: []decoder.CloudStackProfileConfig{
+		{
+			Name:          "test_name",
+			ApiKey:        "test",
+			SecretKey:     "test",
+			ManagementUrl: "http://1.1.1.1:8080/client/api",
+		},
```
maxdrib, line 54:
```suggestion
		...execConfig.Profiles[0],
```

Would this be easier? Doesn't matter much

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -63,7 +63,7 @@ type cloudstackProvider struct {
 	selfSigned            bool
 	templateBuilder       *CloudStackTemplateBuilder
 	skipIpCheck           bool
-	validator             *Validator
+	validators            map[string]*Validator
```
maxdrib, line 66: here too - why not make it a list? It seems a bit simpler that way in my mind. Is there any specific usecase you have in mind for making it a map? It just feels like more work for the user to maintain this mapping in my mind but open to other opinions

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -375,22 +384,28 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
 	return nil
 }
 
+func (p *cloudstackProvider) validateClusterSpec(ctx context.Context, clusterSpec *cluster.Spec) (err error) {
+	for _, validator := range p.validators {
```
maxdrib, line 388: it feels counterintuitive to run multiple validators all doing the same checks. From an abstraction perspective, the validator should just be invoked once. I think the implementation detail of multiple cmk executables should not leak outside of the validator itself.

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -29,39 +29,58 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
...
-	}
-	verifySsl, err := section.GetKey("verify-ssl")
-	verifySslValue := "true"
-	if err == nil {
-		verifySslValue = verifySsl.Value()
-		if _, err := strconv.ParseBool(verifySslValue); err != nil {
-			return nil, fmt.Errorf("'verify-ssl' has invalid boolean string %s: %v", verifySslValue, err)
+
+	cloudstackProfiles := []CloudStackProfileConfig{}
+	sections := cfg.Sections()
```
maxdrib, line 34: for backwards compatibility, we should also check here to make sure there is a [Global] section present I think

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -29,39 +29,58 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
...
+			return nil, fmt.Errorf("failed to extract value of 'api-url' from %s: %v", EksacloudStackCloudConfigB64SecretKey, err)
+		}
+		verifySsl, err := section.GetKey("verify-ssl")
+		verifySslValue := "true"
+		if err == nil {
+			verifySslValue = verifySsl.Value()
+			if _, err := strconv.ParseBool(verifySslValue); err != nil {
+				return nil, fmt.Errorf("'verify-ssl' has invalid boolean string %s: %v", verifySslValue, err)
+			}
+		}
```
maxdrib, line 59:
```suggestion
		verifySslValue := "true"
		if verifySsl, err := section.GetKey("verify-ssl"); err == nil {
			verifySslValue = verifySsl.Value()
			if _, err := strconv.ParseBool(verifySslValue); err != nil {
				return nil, fmt.Errorf("'verify-ssl' has invalid boolean string %s: %v", verifySslValue, err)
			}
		}
```

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -29,39 +29,58 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
...
+		cloudstackProfiles = append(cloudstackProfiles, CloudStackProfileConfig{
+			Name:          section.Name(),
```
maxdrib, line 61: what is the purpose of recording this name?

maxdrib (commented):
Could you validate that this also works e2e with CAPC for a single endpoint?

pkg/dependencies/factory.go
maxdrib: Never mind, I think a map is the right move so that we can match these Profiles to an Availability Zone for preflight checking

pkg/providers/cloudstack/cloudstack.go
maxdrib: Never mind, I think a map is the right move so that we can match these Profiles to an Availability Zone for preflight checking

pkg/executables/cmk.go
```diff
@@ -440,15 +432,18 @@ func (c *Cmk) CleanupVms(ctx context.Context, clusterName string, dryRun bool) e
...
 	}
+
 	configFile, err := c.buildCmkConfigFile()
 	if err != nil {
-		return bytes.Buffer{}, fmt.Errorf("failed cmk validations: %v", err)
+		return stdout, fmt.Errorf("failed cmk validations: %v", err)
 	}
+
 	argsWithConfigFile := append([]string{"-c", configFile}, args...)
+	stdout, err = c.executable.Execute(ctx, argsWithConfigFile...)
```
maxdrib, line 444:
```suggestion
	return c.executable.Execute(ctx, argsWithConfigFile...)
```

pkg/dependencies/factory.go
maxdrib: why are we using ProviderCmkClient instead of the executables interface?

wongni: ProviderCmkClient and executables.Cmk are compliant, however, it seems that the maps of them are not compliant in Golang. When I changed Cmks' type to map[string]*executables.Cmk I got this compile error. 

```
pkg/dependencies/factory.go:270:19: cannot use f.dependencies.Cmks (type map[string]*executables.Cmk) as type map[string]cloudstack.ProviderCmkClient in argument to cloudstack.NewProvider
```

pkg/executables/cmk_test.go
wongni: Good idea!

pkg/providers/cloudstack/decoder/decoder.go
wongni: Good point

pkg/dependencies/factory.go
maxdrib: I guess this is an OOP issue: https://stackoverflow.com/questions/46433860/golang-using-a-defined-interface-in-a-map-value. Maybe it makes sense to contain the map entirely to be inside the Cmk struct as a map of executable objects

maxdrib: My concern with using ProviderCmkClient is that now we can introduce a circular dependency between executables -> cloudstack -> executables. The executables.Cmk interface served to decouple the two packages as I understand so I'm hesitant to break this pattern

internal/test/cleanup/cleanup.go
```diff
@@ -99,12 +99,17 @@ func CleanUpCloudstackTestResources(ctx context.Context, clusterName string, dry
...
-	defer cmk.Close(ctx)
+	for _, config := range execConfig.Profiles {
+		cmk := executableBuilder.BuildCmkExecutable(tmpWriter, config)
+		defer cmk.Close(ctx)
 
-	if err := cmk.ValidateCloudStackConnection(ctx); err != nil {
-		return fmt.Errorf("validating cloudstack connection with cloudmonkey: %v", err)
-	}
+		if err := cmk.ValidateCloudStackConnection(ctx); err != nil {
+			return fmt.Errorf("validating cloudstack connection with cloudmonkey: %v", err)
```
maxdrib, line 107: I think it may be better to capture these errors in a multierror instead of returning in the middle of this loop. If the cloudstack connection fails for a profile, I'd `continue` to the next one

maxdrib: and then we can close on the `continue` as well as at the end of each loop

wongni:
> Could you validate that this also works e2e with CAPC for a single endpoint?

I was able to create a management cluster for a single endpoint.

pkg/providers/cloudstack/decoder/decoder_test.go
wongni: Done

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
 	return nil
 }
 
-func (v *Validator) validateDomainAndAccount(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
-	if (datacenterConfig.Spec.Domain != "" && datacenterConfig.Spec.Account == "") ||
-		(datacenterConfig.Spec.Domain == "" && datacenterConfig.Spec.Account != "") {
-		return fmt.Errorf("both domain and account must be specified or none of them must be specified")
+func (v *Validator) generateLocalAvailabilityZones(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
+	if datacenterConfig == nil {
+		return nil
```
maxdrib, line 121: shouldn't this return error if datacenterConfig is nil?

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
+		cmk := v.cmks[decoder.CloudStackGlobalAZ]
+		domain, err := cmk.ValidateDomainPresent(ctx, datacenterConfig.Spec.Domain)
+		if err != nil {
+			return err
 		}
-
-		errAccount := v.cmk.ValidateAccountPresent(ctx, datacenterConfig.Spec.Account, domain.Id)
-		if errAccount != nil {
-			return fmt.Errorf("checking account: %v", errAccount)
+		if err := cmk.ValidateAccountPresent(ctx, datacenterConfig.Spec.Account, domain.Id); err != nil {
```
maxdrib, line 130: we don't always need to check that the account is present. The logic is: when account is set, domain must also be set. But domain can also be set without account. See https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/validator.go#L103-L123

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
 
 	logger.MarkPass("Datacenter validated")
 	return nil
 }
 
-func (v *Validator) validateDomainAndAccount(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
-	if (datacenterConfig.Spec.Domain != "" && datacenterConfig.Spec.Account == "") ||
-		(datacenterConfig.Spec.Domain == "" && datacenterConfig.Spec.Account != "") {
-		return fmt.Errorf("both domain and account must be specified or none of them must be specified")
+func (v *Validator) generateLocalAvailabilityZones(ctx context.Context, datacenterConfig *anywherev1.CloudStackDatacenterConfig) error {
```
maxdrib, line 119: If we convert to the new schema using the `SetDefaults` method, we won't need this function. Then we'll know that going into the validator, we only have to check the new fields, not the old ones. Although, once you change to using the new schema, you basically have to use it throughout the entire provider and controller or else everything will be broken. Maybe this is a good way to split up the code changes into more reasonable changes. If we go with that approach, the PR's will be:

1. Add new fields to the interface
2. This PR, which does preflight checking for both the old CloudstackDatacenterConfig schema and the new schema, but uses old CloudstackDatacenterConfig attributes when generating the CAPI/CAPC templates
3. Third PR, which will invoke the SetDefaults method any time we're deserializing a CloudstackDatacenterConfig and then exclusively use the new fields through the provider and controller, but generating CAPI/CAPC templates for v0.4.5
4. Fourth PR, which will support CAPC v0.5 for template generation

Does that sound reasonable?

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
 	}
 	return nil
 }
 
 // TODO: dry out machine configs validations
 func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, cloudStackClusterSpec *Spec) error {
-	var etcdMachineConfig *anywherev1.CloudStackMachineConfig
+	if err := v.ValidateCloudStackDatacenterConfig(ctx, cloudStackClusterSpec.CloudStackDatacenter); err != nil {
+		return err
+	}
```
maxdrib, line 173: why do we want to do this? It'll make testing harder, and we're already calling this method from cloudstack.go. This way it's harder to tell which method is misbehaving if there's a dependency between them.

pkg/providers/cloudstack/validator.go
```diff
@@ -234,29 +280,24 @@ func (v *Validator) validateMachineConfig(ctx context.Context, datacenterConfig
...
-	zones, err := v.cmk.ValidateZonesPresent(ctx, datacenterConfig.Spec.Zones)
-	if err != nil {
-		return fmt.Errorf("checking zones %v", err)
-	}
-	account := datacenterConfig.Spec.Account
 
-	for _, zone := range zones {
-		if err = v.cmk.ValidateTemplatePresent(ctx, domainId, zone.Id, account, machineConfig.Spec.Template); err != nil {
+	for _, az := range v.availabilityZones {
+		cmk := v.cmks[az.Name]
```
maxdrib, line 285: I think it would be good to check that the cmk exists in the map again

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
+			return fmt.Errorf("checking management api endpoint: %v", err)
+		}
 
-	for _, zone := range datacenterConfig.Spec.Zones {
-		if len(zone.Network.Id) == 0 && len(zone.Network.Name) == 0 {
-			return fmt.Errorf("zone network is not set or is empty")
+		cmk := v.cmks[az.Name]
+		if cmk.GetManagementApiEndpoint(ctx) != az.ManagementApiEndpoint {
+			return fmt.Errorf("cloudstack secret management url (%s) differs from cluster spec management url (%s)",
+				cmk.GetManagementApiEndpoint(ctx), az.ManagementApiEndpoint)
```
maxdrib, line 97: it feels weird to call a public function twice. Can we save this as a variable instead of calling it a second time?

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
 		}
-		err := v.cmk.ValidateNetworkPresent(ctx, domainId, zone, zones, datacenterConfig.Spec.Account, len(zones) > 1)
+
+		zones, err := cmk.ValidateZonesPresent(ctx, []anywherev1.CloudStackZone{az.CloudStackAvailabilityZone.Zone})
```
maxdrib, line 100: it seems like we should refactor this to only check if a single zone is present, not multiple. That'll require less error handling below as well

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -29,39 +30,65 @@ func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
...
-	verifySslValue := "true"
-	if err == nil {
-		verifySslValue = verifySsl.Value()
-		if _, err := strconv.ParseBool(verifySslValue); err != nil {
-			return nil, fmt.Errorf("'verify-ssl' has invalid boolean string %s: %v", verifySslValue, err)
-		}
+
+	if !foundGlobalSection {
+		return nil, fmt.Errorf("[Global] section not found from %s", EksacloudStackCloudConfigB64SecretKey)
 	}
```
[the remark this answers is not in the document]
wongni: Removed

pkg/providers/cloudstack/testdata/cloudstack_config_invalid_format.ini
```diff
@@ -0,0 +1,5 @@
+[Global]
+verify-ssl ; false
+api-key ; test-key1
+secret-key ; test-secret1
+api-url ; http://127.16.0.1:8080/client/api
```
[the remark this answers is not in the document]
wongni: Done

pkg/providers/cloudstack/validator.go
```diff
@@ -48,83 +55,122 @@ type ProviderCmkClient interface {
...
+		azNamesToCheck = append(azNamesToCheck, decoder.CloudStackGlobalAZ)
+	}
+	for _, az := range datacenterConfig.Spec.AvailabilityZones {
+		azNamesToCheck = append(azNamesToCheck, az.Name)
 	}
-	logger.MarkPass("Connected to server")
 
+	for _, azName := range azNamesToCheck {
+		cmk := v.cmks[azName]
+		if cmk == nil {
```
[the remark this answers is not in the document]
wongni: Done

pkg/providers/cloudstack/validator.go
wongni: Sounds good to me.

pkg/providers/cloudstack/validator.go
wongni: Done

pkg/providers/cloudstack/validator.go
wongni: Done

pkg/executables/cmk.go
```diff
@@ -273,10 +260,10 @@ func (c *Cmk) ValidateDomainPresent(ctx context.Context, domain string) (v1alpha
 	return domainIdentifier, nil
 }
 
-func (c *Cmk) ValidateNetworkPresent(ctx context.Context, domainId string, zone v1alpha1.CloudStackZone, zones []v1alpha1.CloudStackResourceIdentifier, account string, multipleZone bool) error {
+func (c *Cmk) ValidateNetworkPresent(ctx context.Context, domainId string, network v1alpha1.CloudStackResourceIdentifier, zoneId string, account string, multipleZone bool) error {
```
maxdrib, line 263: Why change the zone reference v1alpha1.CloudStackResourceIdentifier to simply the zoneId? Don't we still need a way to obtain that zoneId?

EDIT: nvm looks like we do that in validator.go

pkg/providers/cloudstack/validator.go
maxdrib: bump

pkg/providers/cloudstack/validator.go
maxdrib: bump

maxdrib (commented):
Looks like there are some conflicts to resolve. Overall, minor comments and the logic looks good to me. Nice job

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -13,6 +13,7 @@ const (
 	EksacloudStackCloudConfigB64SecretKey = "EKSA_CLOUDSTACK_B64ENCODED_SECRET"
 	CloudStackCloudConfigB64SecretKey     = "CLOUDSTACK_B64ENCODED_SECRET"
 	EksaCloudStackHostPathToMount         = "EKSA_CLOUDSTACK_HOST_PATHS_TO_MOUNT"
+	CloudStackGlobalAZ                    = "Global"
```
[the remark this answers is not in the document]
wongni: yes

internal/test/cleanup/cleanup.go
```diff
@@ -99,12 +100,23 @@ func CleanUpCloudstackTestResources(ctx context.Context, clusterName string, dry
 	if err != nil {
 		return fmt.Errorf("building cmk executable: %v", err)
 	}
-	cmk := executableBuilder.BuildCmkExecutable(tmpWriter, *execConfig)
-	defer cmk.Close(ctx)
+	for _, config := range execConfig.Profiles {
+		cmk := executableBuilder.BuildCmkExecutable(tmpWriter, config)
+		if err := cleanupCloudStackVms(ctx, cmk, clusterName, dryRun); err != nil {
+			retErr = multierror.Append(retErr, err)
```
g-gaston, line 106: What's the point of aggregating errors if we are only doing it once?

pkg/dependencies/factory.go
g-gaston: It seems like `executables.Cmk` is a struct, right?

Either way, in general, we want to define interfaces in the consumer, so in this case that would be in the cloudstack provide package.

What I would recommend in this case is store the executables as `map[string]*executables.Cmk` in the deps struct and then convert the map to ` map[string]cloudstack.ProviderCmkClient` before injecting it in the provider.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/google/go-github/v35 v35.3.0
 	github.com/google/uuid v1.3.0
+	github.com/hashicorp/go-multierror v1.1.1
```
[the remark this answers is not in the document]
wongni: Please check https://github.com/aws/eks-anywhere/pull/2404#discussion_r906149410

internal/test/cleanup/cleanup.go
```diff
@@ -99,12 +100,23 @@ func CleanUpCloudstackTestResources(ctx context.Context, clusterName string, dry
...
+		}
+		cmk.Close(ctx)
```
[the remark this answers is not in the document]
wongni: We have a `cmk` per endpoint. We want to close `cmk` after cleaning up resources from one endpoint.

pkg/executables/cmk.go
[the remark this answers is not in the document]
wongni: CloudStackZone is passed into ValidateZonePresent to validate zone by either ID or name. However, other resources are validated in the single resolved zone using zone ID string.

pkg/executables/cmk.go
```diff
@@ -273,10 +260,10 @@ func (c *Cmk) ValidateDomainPresent(ctx context.Context, domain string) (v1alpha
...
 	command := newCmkCommand("list networks")
-	if len(zone.Network.Id) > 0 {
-		applyCmkArgs(&command, withCloudStackId(zone.Network.Id))
+	if len(network.Id) > 0 {
```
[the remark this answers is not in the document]
wongni: When a network ID is defined we use it to list networks. In this case, we expected to have one network returned because the ID is unique. When a network ID is not defined we list all networks in the zone and filter with the network name passed in.

pkg/executables/cmk.go
```diff
@@ -380,14 +353,18 @@ func (c *Cmk) ValidateAccountPresent(ctx context.Context, account string, domain
...
-func NewCmk(executable Executable, writer filewriter.FileWriter, config decoder.CloudStackExecConfig) *Cmk {
+func NewCmk(executable Executable, writer filewriter.FileWriter, config decoder.CloudStackProfileConfig) *Cmk {
 	return &Cmk{
 		writer:     writer,
 		executable: executable,
 		config:     config,
 	}
 }
 
+func (c *Cmk) GetManagementApiEndpoint(ctx context.Context) string {
```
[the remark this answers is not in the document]
wongni: Good point

PR #2031, opened 2022-05-05
add tinkerbell ci test harness to e2e test framework

gwesterfieldjr:
Update e2e test framework for tinkerbell CI

- Enable ability to run e2e tests from a vSphere vm
- Route all tinkerbell tests to vSphere
- Split tinkerbell tests into test runner suites by hardware count (4). This is not most efficient way to split the hardware, but it will do for now. Post GA I will update this to a better reservation system.
- Provision test runner instance to run tinkerbell tests
- Update test framework to account for multiple tests running from single test runner instance
- Update code build job file with required tinkerbell env vars
- Update test power actions to bmc lib rather than pbnj
- Clean up tests no longer needed post beta

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2407, opened 2022-06-14
Specify the network when deploying the VM from content library

abhinavmpandey08:
*Description of changes:*
Provide a network when deploying VM from a template in the content library.
If the network is not provided, vSphere will pick some network and try to deploy the vm on that network which can have unintended consequences.

Tested and verified with both bottlerocket and ubuntu

### DIFF
--- a/pkg/executables/govc.go
+++ b/pkg/executables/govc.go
@@ -36,13 +36,20 @@ const (
 	vSpherePasswordKey   = "EKSA_VSPHERE_PASSWORD"
 	vSphereServerKey     = "VSPHERE_SERVER"
 	byteToGiB            = 1073741824.0
-	deployOptsFile       = "deploy-opts.json"
+	DeployOptsFile       = "deploy-opts.json"
 )
 
 var requiredEnvs = []string{govcUsernameKey, govcPasswordKey, govcURLKey, govcInsecure}
 
-//go:embed config/deploy-opts.json
-var deployOpts []byte
+type networkMapping struct {
+	Name    string `json:"Name,omitempty"`
+	Network string `json:"Network,omitempty"`
+}
+
+type deployOption struct {
+	DiskProvisioning string           `json:"DiskProvisioning,omitempty"`
+	NetworkMapping   []networkMapping `json:"NetworkMapping,omitempty"`
+}
 
 type FolderType string
 
@@ -269,9 +276,9 @@ func (g *Govc) CreateLibrary(ctx context.Context, datastore, library string) err
 	return nil
 }
 
-func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templateName, library, datacenter, datastore, resourcePool string, resizeBRDisk bool) error {
+func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templateName, library, datacenter, datastore, network, resourcePool string, resizeBRDisk bool) error {
 	logger.V(4).Info("Deploying template", "dir", templateDir, "templateName", templateName)
-	if err := g.deployTemplate(ctx, library, templateName, templateDir, datacenter, datastore, resourcePool); err != nil {
+	if err := g.deployTemplate(ctx, library, templateName, templateDir, datacenter, datastore, network, resourcePool); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/executables/govc.go
```diff
@@ -354,7 +354,7 @@ func (g *Govc) deployTemplate(ctx context.Context, library, templateName, deploy
 		templateInLibraryPath = fmt.Sprintf("/%s", templateInLibraryPath)
 	}
 
-	deployOptsPath, err := g.writer.Write(deployOptsFile, deployOpts, filewriter.PersistentFile)
+	deployOptsPath, err := g.writer.Write(deployOptsFile, []byte(fmt.Sprintf(deployOpts, network)), filewriter.PersistentFile)
```
g-gaston, line 357: I would vote to move `deployOpts` to be a go template or even just a go struct and then marshall to json

pkg/executables/govc_test.go
```diff
@@ -181,7 +183,7 @@ func (dt *deployTemplateTest) expectMarkAsTemplateToReturn(err error) {
 
 func (dt *deployTemplateTest) DeployTemplateFromLibrary() error {
 	gomock.InOrder(dt.expectations...)
-	return dt.govc.DeployTemplateFromLibrary(dt.ctx, dt.deployFolder, dt.templateName, templateLibrary, dt.datacenter, dt.datastore, dt.resourcePool, dt.resizeDisk2)
+	return dt.govc.DeployTemplateFromLibrary(dt.ctx, dt.deployFolder, dt.templateName, templateLibrary, dt.datacenter, dt.datastore, dt.network, dt.resourcePool, dt.resizeDisk2)
```
g-gaston, line 186: now that the file is dynamic, maybe we should add a expectation for the content of the file?

g-gaston (commented):
Looks good, minor suggestions :)

pkg/executables/govc.go
```diff
@@ -36,13 +36,20 @@ const (
...
-//go:embed config/deploy-opts.json
-var deployOpts []byte
+type NetworkMapping struct {
+	Name    string
+	Network string
+}
+
+type DeployOption struct {
+	DiskProvisioning string
+	NetworkMapping   []NetworkMapping
```
g-gaston, line 51: nit: add json tags

pkg/executables/govc_test.go
```diff
@@ -196,11 +202,25 @@ func (dt *deployTemplateTest) assertDeployTemplateError(t *testing.T) {
 	}
 }
 
+func (dt *deployTemplateTest) assertDeployOptsMatches(t *testing.T) {
+	expected, err := executables.GetDeployOptions(dt.network)
```
g-gaston, line 206: this is really a super valid code, since it's using the code we are suppose to be testing to generate the expected output

I would suggest just hardcoding the json body here in a string/[]byte (even the network value) and comparing the file against that

g-gaston (approved):

pkg/executables/govc.go
```diff
@@ -269,9 +276,9 @@ func (g *Govc) CreateLibrary(ctx context.Context, datastore, library string) err
 	return nil
 }
 
-func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templateName, library, datacenter, datastore, resourcePool string, resizeBRDisk bool) error {
+func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templateName, library, datacenter, datastore, network, resourcePool string, resizeBRDisk bool) error {
```
gwesterfieldjr, line 279: Idk, if you have looked at my tink e2e PR yet, but I had to do very similar thing as I wanted to use this code to deploy OVA's in the lab. I tried to change as little as possible as I only needed it for the tests, but there are a ton more options in deploy options. Rather, than passing them individually why don't we just take deploy options as a param to account for all possible deploy options?

You can see the options I am using for tink e2e here: https://github.com/aws/eks-anywhere/pull/2031/files#diff-5dfd6bc5a441b6e4027710c63250668f8e0af8e594dfcb6c3c11067cca3e5845

But there are many more options available that we could support.

gwesterfieldjr: Link above is code in test package, actual govc changes I made were minimal to not disrupt. You can see them here: https://github.com/aws/eks-anywhere/pull/2031/files#diff-1b2bffa403bbd82a1ba98650cb15df32447b074ad23e1cc1cf7df592eca4554d

Thoughts?

abhinavmpandey08: oh just saw your comment. For CAPV provider, I think we only needed the network option. The disk provisioning option is static. But if more options are needed, then yeah we can make it more generic and provide a slice of options

abhinavmpandey08: yeah seems like you already handled that in your PR. Will review it and try to get it merged tomorrow so you don't have to deal with a bunch of merge conflicts

gwesterfieldjr: Yea, I have to get code build in order before merging, but please do review as I want to merge before end of week.

PR #2409, opened 2022-06-14
Remove unnecessary fetchs from tinkerbell template

g-gaston:
*Issue #, if available:*

*Description of changes:*
The versions bundle is already included in the `cluster.Spec`

*Testing (if applicable):*


d8660091:
/lgtm
nit: cover the added line with unit test

pokearu (approved):
Changes look good.

The `GetVersionsBundleForVersion` was initially needed elsewhere in the generate cluster config for Tinkerbell but we don't need that now after this #2243.

Issue #488, opened 2021-10-25
Support node CIDR mask config


g-gaston:
Right now, the kube controller manager is using the default for `--node-cidr-mask-size` (`24` for ipv4 and `64` for ipv6)
Add the ability to configure this through the eks-a cluster config CRD

CharudathGopal:
We are trying to deploy a EKS Anywhere setup with more than 500 nodes are hitting this issue too.

Appreciate any workarounds or suggestions for us to move ahead.

jaxesn:
We are going to go ahead and look into this one and see if we can something quickly. Well plan on having it in our late June release (0.10.0) but could produce a dev build if interested in testing it earlier once we have something in place.

CharudathGopal:
@jaxesn Please let us know when you have the fix, happy to give a shot!

JFYI, Here is the command we used to set the CIDE on the nodes after which the Cilium pods came up as expected. Before this fix, we had only 254 cilium pods coming up as /24 mask was used.

`kubectl annotate node --all --overwrite io.cilium.network.ipv4-pod-cidr=192.168.0.0/16`

jaxesn:
When you initially created the cluster with 500 nodes, what were some of the values of the cilium annotation before you changed it?

jaxesn:
@CharudathGopal would you mind giving me a few more details on your network setup and what cidr range and masks you would like to set?

CharudathGopal:
Here is the snapshot of cilium config.
```
auto-direct-node-routes                        false
bpf-lb-map-max                                 65536
bpf-map-dynamic-size-ratio                     0.0025
bpf-policy-map-max                             16384
cgroup-root                                    /run/cilium/cgroupv2
cilium-endpoint-gc-interval                    5m0s
cluster-id
cluster-name                                   poc-22
cni-chaining-mode                              portmap
custom-cni-conf                                false
debug                                          false
disable-cnp-status-updates                     true
enable-auto-protect-node-port-range            true
enable-bandwidth-manager                       false
enable-bpf-clock-probe                         true
enable-bpf-masquerade                          true
enable-endpoint-health-checking                true
enable-health-check-nodeport                   true
enable-health-checking                         true
enable-hubble                                  true
enable-ipv4                                    true
enable-ipv6                                    false
enable-l7-proxy                                true
enable-local-redirect-policy                   false
enable-metrics                                 true
enable-policy                                  default
enable-remote-node-identity                    true
enable-session-affinity                        true
[...]

PR #2411, opened 2022-06-15
Support for node cidr mask size documentation

mitalipaygude:
*Issue #, if available:*
#488 
*Description of changes:*
Documentation for the above issue
*Testing (if applicable):*


docs/content/en/docs/reference/clusterspec/optional/cni.md
```diff
@@ -169,4 +169,40 @@ EKS Anywhere will create the required NetworkPolicy objects for its core compone
...
+      - 10.96.0.0/12
+    cniConfig:
+      cilium: {}
+    nodes:
+      cidrMaskSize: 24
+```
+
+If the user does not specify this field in the cluster yaml spec, the value for this flag defaults to 24 for ipv4. 
+Please note that this mask size needs to be greater than the pods cidr mask size. In the above spec, the pod cidr mask size is `16`
+and the node cidr mask size is `24`. This ensures the cluster 256 blocks of /24 networks. For example, node1 will get
```
chrisnegus, line 197:
```suggestion
and the node CIDR mask size is `24`. This ensures the cluster 256 blocks of /24 networks. For example, node1 will get 
```

chrisnegus (commented):
Looks really good. Most of my comments are just editing suggestions.

docs/content/en/docs/reference/clusterspec/optional/cni.md
```diff
@@ -169,4 +169,40 @@ EKS Anywhere will create the required NetworkPolicy objects for its core compone
...
+      cidrBlocks:
+      - 10.96.0.0/12
+    cniConfig:
+      cilium: {}
+    nodes:
+      cidrMaskSize: 24
+```
+
+If the user does not specify the `clusterNetwork.nodes` field in the cluster yaml spec, the value for this flag defaults to 24 for IPv4. 
+Please note that this mask size needs to be greater than the pods CFIDR mask size. In the above spec, the pod CIDR mask size is `16`
```
chrisnegus, line 196:
```suggestion
Please note that this mask size needs to be greater than the pods CIDR mask size. In the above spec, the pod CIDR mask size is `16`
```

jaxesn (approved):

Issue #2406, opened 2022-06-14
Add support for clusters spanning across multiple Cloudstack API Endpoints 


maxdrib:
**What would you like to be added**:
Currently, EKS-A accepts one set of credentials for creating a cluster on one Cloudstack environment. The scope of this task is to add support in EKS-A for clusters which can span across multiple Cloudstack endpoints in preflight checking, cluster creation, cluster upgrades, and cluster deletion.

**Why is this needed**:
The customer has expressed a need for creating clusters across multiple Cloudstack environments/endpoints. Their architecture only enables highly available multi-zone clusters when they are distributed across multiple Cloudstack endpoints, so this feature would enable them to have more resilient and scalable clusters.

maxdrib:
https://github.com/aws/eks-anywhere/pull/2404 modifies the input schema for credentials to be passed to cloudmonkey and CAPC

PR #2412, opened 2022-06-15
Defining new types for Cloudstack multiendpoint development

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR introduces the new types as described in the Design PR https://github.com/aws/eks-anywhere/pull/2399 for supporting running Cloudstack clusters across availability zones. We will preserve the old types as well and introduce a transformer to convert the old types to the new ones once the internal logic is migrated to use the new fields.

*Testing (if applicable):*
Unit tests all pass


wongni (approved):
Looks good to me.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -29,14 +29,15 @@ type CloudStackDatacenterConfigSpec struct {
...
+	Domain string `json:"domain,omitempty"`
 	// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
-	Zones []CloudStackZone `json:"zones"`
+	Zones []CloudStackZone `json:"zones,omitempty"`
 	// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
 	Account string `json:"account,omitempty"`
 	// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
-	ManagementApiEndpoint string `json:"managementApiEndpoint"`
+	ManagementApiEndpoint string `json:"managementApiEndpoint,omitempty"`
+	// List of AvailabilityZones to distribute VMs across - corresponds to a list of CAPI failure domains
```
g-gaston, line 39:
```suggestion
	// AvailabilityZones lists different partitions to distribute VMs across - corresponds to a list of CAPI failure domains
```

Just so we can follow the go pattern for field comments
But feel free to rephrase it

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -142,6 +159,26 @@ func (v *CloudStackDatacenterConfig) Validate() error {
 	return nil
 }
 
+func (v *CloudStackDatacenterConfig) SetDefaults() {
+	if v.Spec.AvailabilityZones == nil || len(v.Spec.AvailabilityZones) == 0 {
+		v.Spec.AvailabilityZones = []CloudStackAvailabilityZone{}
```
g-gaston, line 164:
```suggestion
		v.Spec.AvailabilityZones = make([]CloudStackAvailabilityZone, 0, len(v.Spec.Zones)
```

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -157,6 +194,14 @@ func (s *CloudStackDatacenterConfigSpec) Equal(o *CloudStackDatacenterConfigSpec
 			return false
 		}
 	}
+	if len(s.AvailabilityZones) != len(o.AvailabilityZones) {
+		return false
+	}
+	for i, az := range s.AvailabilityZones {
+		if !az.Equal(&o.AvailabilityZones[i]) {
```
g-gaston, line 201: This is relying on order
Since we have a name, I would vote to create a map, and then compare equality searching in that map by name.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -142,6 +159,26 @@ func (v *CloudStackDatacenterConfig) Validate() error {
 	return nil
 }
 
+func (v *CloudStackDatacenterConfig) SetDefaults() {
+	if v.Spec.AvailabilityZones == nil || len(v.Spec.AvailabilityZones) == 0 {
```
g-gaston, line 163: I think we need to add two extra validations:
* Make sure that v.Spec.Zones, v.Spec.Domain, v.Spec.Account,	v.Spec.ManagementApiEndpoint are all empty
* Make sure that `len(v.Spec.AvailabilityZones) > 0`

g-gaston (commented):
Looks good
Should we wait to merge this? I believe it will break main for cloudstack, right?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
maxdrib: Makes sense. I think I'll have to change the SetDefaults function then, since that currently defaults an AZ name to "Global" when converting from the old schema to the new one. Validating will become more tricky however, since we won't have a way to map from the old input ini file with the [Global] tag to the new AZ's which are named by their zone names. We can deal with that in the subsequent PR

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
maxdrib: I see. I think that makes sense to do in a follow-up PR once we migrate to actually using these types throughout. I can add a TODO here

designs/cloudstack-multiple-endpoints.md
```diff
@@ -49,50 +49,50 @@ will need to be available on *all* the Cloudstack API endpoints. We will validat
...
+    // Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
+    Account string `json:"account,omitempty"`
+    // CloudStack Management API endpoint's IP. It is added to VM's noproxy list
+    ManagementApiEndpoint string `json:"managementApiEndpoint"`
 }
 ```
 
 We would instead propose to gradually deprecate all the existing attributes and instead, simply include a list of AvailabilityZone objects like so
 
 ```
```
g-gaston, line 65:
```suggestion
```go
```

designs/cloudstack-multiple-endpoints.md
```diff
@@ -49,50 +49,50 @@ will need to be available on *all* the Cloudstack API endpoints. We will validat
...
+    // List of AvailabilityZones to distribute VMs across - corresponds to a list of CAPI failure domains
+    AvailabilityZones []CloudStackAvailabilityZone `json:"availabilityZones,omitempty"`
 }
 ```
 
 where each AvailabilityZone object looks like
 
 ```
 type CloudStackAvailabilityZone struct {
-	// Name would be used to match the availability zone defined in the datacenter config to the credentials passed in from the cloud-config ini file
```
g-gaston, line 84: I would suggest to keep this unless there is a mayor concern
We learnt with the worker node groups that it makes comparing arrays easier (which is helpful to detect updates, removal, etc.)

designs/cloudstack-multiple-endpoints.md
```diff
@@ -107,7 +107,13 @@ CAPC currently utilizes them to distribute machines across CloudStack Zones. How
 2. Cloudstack domain
 3. Cloudstack zone
 4. Cloudstack account
-5. A unique name
+5. A reference to the customer-provider credentials
+
+To add more detail to the credential ref, we will continue to accept a base64 encoded ini file as input from the customer when running eks-a. However,
+in order to avoid taking a dependency on a meaningless section of a transient file, we will modify EKS-A to also create the secrets to be used by CAPC.
+These secrets will be generated from the ini file provided by customer and applied to the cluster. CAPC will then proceed to take ownership of them as
```
g-gaston, line 114: Maybe worth to mention in which scenarios this file will be required? For example: when creating a management cluster, always. But what about upgrade? What about workload clusters?

Another interesting questions is, what happens if the user provides an ini file with a "secret" that already exists? (imagine for management cluster upgrades). Will the secret be updated or should they be immutable?

maxdrib: I think we would model all these decisions from what the vsphere provider currently has defined. As I understand, the secrets can be updated on every eks-a operation (create, upgrade) so I would expect the file will be required for every operation. VSphere seems to apply the generated secrets from the input credentials on every operation and ignores old ones. I suppose this isn't the cleanest approach but old secrets could be cleaned up manually in the worst case.

designs/cloudstack-multiple-endpoints.md
vivek-koppuru: We can maybe default the names? I would argue that if users name they would be able to properly associate which one is related to the appropriate credentialref and all other fields, especially because we are not going with the CRD approach, which would anyways require a name. So I would also vote for a name unless we have a strong reason not to.

maxdrib: Noted - will re-add and autogenerate. We can also add a validation step to ensure the names are unique

designs/cloudstack-multiple-endpoints.md
maxdrib: Updated doc to reflect our offline discussion - EKS-A will create secrets but not modify existing ones to prevent unintentional overwrites. If users want to modify credentials, they can do so manually, see below

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -139,9 +160,34 @@ func (v *CloudStackDatacenterConfig) Marshallable() Marshallable {
...
+	// * Make sure that all the AZ names are unique
 	return nil
 }
 
+func (v *CloudStackDatacenterConfig) SetDefaults() {
+	if v.Spec.AvailabilityZones == nil || len(v.Spec.AvailabilityZones) == 0 {
+		v.Spec.AvailabilityZones = make([]CloudStackAvailabilityZone, 0, len(v.Spec.Zones))
+		for index, csZone := range v.Spec.Zones {
+			az := CloudStackAvailabilityZone{
+				Name:                  fmt.Sprintf("availability-zone-%d", index),
```
maxdrib, line 175: this is subject to change pending alignment with CAPC for backwards compatibility

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -29,14 +31,15 @@ type CloudStackDatacenterConfigSpec struct {
 	// Domain contains a grouping of accounts. Domains usually contain multiple accounts that have some logical relationship to each other and a set of delegated administrators with some authority over the domain and its subdomains
 	// This field is considered as a fully qualified domain name which is the same as the domain path without "ROOT/" prefix. For example, if "foo" is specified then a domain with "ROOT/foo" domain path is picked.
 	// The value "ROOT" is a special case that points to "the" ROOT domain of the CloudStack. That is, a domain with a path "ROOT/ROOT" is not allowed.
-	//
-	Domain string `json:"domain"`
+	Domain string `json:"domain,omitempty"`
```
jiayiwang7, line 34: do we want to add the deprecated note to each field?

maxdrib: We will add that after we've migrated EKS-A internally to using the new fields. Prior to that, the linter throws a ton of errors for using deprecated fields which will slow down our reviews :(

jiayiwang7 (approved):

jiayiwang7 (approved):

wongni (approved):

designs/cloudstack-multiple-endpoints.md
```diff
@@ -47,52 +47,54 @@ will need to be available on *all* the Cloudstack API endpoints. We will validat
...
-	// Zones is a list of one or more zones that are managed by a single CloudStack management endpoint.
-	Zone CloudStackZone `json:"zone"`
-	// Account typically represents a customer of the service provider or a department in a large organization. Multiple users can exist in an account, and all CloudStack resources belong to an account. Accounts have users and users have credentials to operate on resources within that account. If an account name is provided, a domain must also be provided.
-	Account string `json:"account,omitempty"`
-	// CloudStack Management API endpoint's IP. It is added to VM's noproxy list
-	ManagementApiEndpoint string `json:"managementApiEndpoint"`
+    // Name would be used as a unique identifier for each availability zone
+    Name string `json:"name"`
+    // CredentialRef would be used to match a secret in the eksa-system namespace
+    CredentialRef string `json:"credentialRef"
```
g-gaston, line 87: nit: maybe?
```suggestion
    CredentialsRef string `json:"credentialsRef"
```

g-gaston (approved):

Issue #1857, opened 2022-04-15
Cluster creation failed with certificate validation error when using mirror registry with self-signed certificate


rimaulana:
**What happened**:

Failed in validation step during cluster creation due to invalid registry certificate when using self-signed certificate and provided the ```registryMirrorConfiguration.caCertContent``` field

```
Warning: The recommended number of control plane nodes is 3 or 5
Performing setup and validations
✅ Connected to server
✅ Authenticated to vSphere
✅ Datacenter validated
✅ Network validated
✅ Datastore validated
✅ Folder validated
✅ Resource pool validated
✅ Datastore validated
✅ Folder validated
✅ Resource pool validated
✅ Control plane and Workload templates validated
✅ Vsphere Provider setup is valid
❌ Validation failed     {"validation": "create preflight validations pass", "error": "validation failed with 1 errors: invalid registry certificate: failed to verify certificate: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0", "remediation": ""}
Error: validations failed
```

**What you expected to happen**:

CA certificate is valid and cluster creation should proceed without any error.

**How to reproduce it (as minimally and precisely as possible)**:
Use registry mirror configuration that has a self-signed certificate and provided the self-signed CA certificate as well in caCertContent field.

**Anything else we need to know?**:

The CA and server certificate for the registry mirror I was using on harbor was generated using AWS Private CA. When validating server certificate using openssl_client the certificate is valid.

```
$ openssl s_client -showcerts -connect core.harbor.g4k.io:443 -CAfile ca.crt 
CONNECTED(00000003)
depth=1 C = US, O = AWS, OU = Support, ST = Virginia, CN = core.harbor.g4k.io, L = Herndon
verify return:1
depth=0 CN = core.harbor.g4k.io
verify return:1
---
Certificate chain
 0 s:CN = core.harbor.g4k.io
   i:C = US, O = AWS, OU = Support, ST = Virginia, CN = core.harbor.g4k.io, L = Herndon
-----BEGIN CERTIFICATE-----
MIIDtDCCApygAwIBAgIRAMi3uW9C4cli2HqdkIMQ5h4wDQYJKoZIhvcNAQELBQAw
bzELMAkGA1UEBhMCVVMxDDAKBgNVBAoMA0FXUzEQMA4GA1UECwwHU3VwcG9ydDER
[...]

PR #2414, opened 2022-06-15
Validate server cert against ca cert for registry mirror

vivek-koppuru:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1857

*Description of changes:*
As described in the issue above, we weren't checking for the validity of the cert if the caCert passed in is not the self-signed server cert instead. This allows for checking the validity of the cert against the server itself instead of just checking for whether the cert is valid without actually making a call to the server. Also modified the unit tests to spin up a test http server to test the cert setup.

*Testing (if applicable):*
Ran against ci registry mirror configuration and another public one

### DIFF
--- a/pkg/crypto/validator.go
+++ b/pkg/crypto/validator.go
@@ -11,7 +11,7 @@ import (
 type DefaultTlsValidator struct{}
 
 type TlsValidator interface {
-	ValidateCert(host, port, cert string) error
+	ValidateCert(host, port, caCertContent string) error
 	HasSelfSignedCert(host, port string) (bool, error)
 }
 
@@ -37,9 +37,9 @@ func (tv *DefaultTlsValidator) HasSelfSignedCert(host, port string) (bool, error
 }
 
 // ValidateCert parses the cert, ensures that the cert format is valid and verifies that the cert is valid for the url
-func (tv *DefaultTlsValidator) ValidateCert(host, port, cert string) error {
+func (tv *DefaultTlsValidator) ValidateCert(host, port, caCertContent string) error {
 	// Validates that the cert format is valid
-	block, _ := pem.Decode([]byte(cert))
+	block, _ := pem.Decode([]byte(caCertContent))
 	if block == nil {
 		return fmt.Errorf("failed to parse certificate PEM")
 	}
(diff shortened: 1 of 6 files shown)


pkg/crypto/validator.go
```diff
@@ -50,15 +50,14 @@ func (tv *DefaultTlsValidator) ValidateCert(host, port, cert string) error {
...
+		RootCAs:            roots,
 	}
-
-	// Verifies that the cert is valid
-	_, err = providedCert.Verify(opts)
+	// Verifies that the cert is valid by making a connection to the endpoint
+	_, err = tls.Dial("tcp", net.JoinHostPort(host, port), conf)
 	if err != nil {
-		return fmt.Errorf("failed to verify certificate: %v", err)
+		return fmt.Errorf("verifying connection to host: %v", err)
```
g-gaston, line 60:
```suggestion
		return fmt.Errorf("verifying connection to host with custom CA: %v", err)
```

pkg/crypto/validator_test.go
```diff
@@ -44,30 +115,77 @@ zwaAnAW1ZiriAbeFx+xOaO1fETVSm+5Poyl97r5Mmu97+3IpoWHFPO2z4Os9vn3q
...
 }
 
-func TestValidateCertInvalidCert(t *testing.T) {
-	tv := crypto.NewTlsValidator()
-	if err := tv.ValidateCert(endpoint, port, invalid_cert); err == nil {
-		t.Fatalf("Certificate validation passed for invalid cert")
+func runTestServer() (*httptest.Server, error) {
+	mux := http.NewServeMux()
+	svr := httptest.NewUnstartedServer(mux)
+	certificate, err := tls.X509KeyPair([]byte(serverCert), []byte(serverKey))
```
g-gaston, line 182: can we take `serverCert` and `serverKey` as arguments to this method?

pkg/crypto/validator.go
```diff
@@ -50,15 +50,14 @@ func (tv *DefaultTlsValidator) ValidateCert(host, port, cert string) error {
...
-		Roots:   roots,
+	conf := &tls.Config{
+		InsecureSkipVerify: false,
+		RootCAs:            roots,
 	}
-
-	// Verifies that the cert is valid
-	_, err = providedCert.Verify(opts)
+	// Verifies that the cert is valid by making a connection to the endpoint
+	_, err = tls.Dial("tcp", net.JoinHostPort(host, port), conf)
```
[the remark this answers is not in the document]
vivek-koppuru: I didn't because the above method didn't either, but that's probably a good idea didn't realize.

g-gaston (approved):

PR #2417, opened 2022-06-15
Increase fs inotify resources for e2e tests

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
The docker e2e tests are failing because of kubelet errors regarding too many open files.
It is a known issue and based on kind documentation [here](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files), this pr fixes the issue.

*Testing (if applicable):*

### DIFF
--- a/internal/test/e2e/setup.go
+++ b/internal/test/e2e/setup.go
@@ -25,8 +25,6 @@ const (
 	bundlesReleaseManifestFile = "local-bundle-release.yaml"
 	eksAComponentsManifestFile = "local-eksa-components.yaml"
 	testNameFile               = "e2e-test-name"
-	maxUserWatches             = 524288
-	maxUserInstances           = 512
 )
 
 type E2ESession struct {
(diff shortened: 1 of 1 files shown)


vivek-koppuru (approved):

internal/test/e2e/setup.go
```diff
@@ -153,6 +159,17 @@ func (e *E2ESession) setup(regex string) error {
 	return nil
 }
 
+func (e *E2ESession) updateFSInotifyResources() error {
+	command := fmt.Sprintf("sudo sysctl fs.inotify.max_user_watches=%v && sudo sysctl fs.inotify.max_user_instances=%v", maxUserWatches, maxUserInstances)
```
jaxesn, line 163: Can we call sudo via ssm?  Not sure if we are already or not...

mitalipaygude: we did [here](https://github.com/aws/eks-anywhere/blob/main/internal/test/e2e/registryMirror.go#L37) so should be fine

jaxesn (approved):

PR #2420, opened 2022-06-15
Add kubernetes.Client controller implementation

g-gaston:

*Description of changes:*

The controller runtime client implementation takes client.Client as an
input, which is a superset of runtime.Object. To make things easier, we
just make that a requirement in our kube client interface. To avoid
polluting our interface with something defined in controller runtime,
we just create a our own kubernetes.Object interface.

*Testing (if applicable):*
Unit tests


drewvanstone:
LGTM

Issue #2265, opened 2022-05-25
Collect workload cluster support bundle on control plane wait failure


jaxesn:
**What happened**:
During cluster creation failure if a failure happens after between [here](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/clustermanager/cluster_manager.go#L255) and [here](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/clustermanager/cluster_manager.go#L266) when collectdiagtask is run [here](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/workflows/create.go#L201) the workloadcluster is not set on the context. This results in the call to collect being skipped due to [this](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/clustermanager/cluster_manager.go#L659)

**What you expected to happen**:

Any failure after [this](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/clustermanager/cluster_manager.go#L250) should generate a support bundle from the workload cluster.

Talking with @g-gaston the quickest, most reasonable, change would be to move the cp wait and move extra objects calls to a new method, returning from createworkloadcluster immediately after getting a kubeconfig.  [CreateWorkloadCluster](https://github.com/aws/eks-anywhere/blob/a908ea1269a594def5fc79be4cce5fccf20c60e5/pkg/workflows/create.go#L196) would then call the new method to wait for the cp and apply extraobjects.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

vivek-koppuru:
Need to either support the newer version of troubleshoot or refactor the code to not always default to running the crd collectors. Will implement the refactoring first and then merge #2431 , and then upgrade troubleshoot to support podspec in our collectors.

PR #2431, opened 2022-06-16
Refactor post workload creation operations to generate support bundle…

vivek-koppuru:
… for workload cluster

*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/2265

*Description of changes:*
Refactor out some of the post creation methods to a separate method to be able to properly set the workload cluster context in order to be able to pull the support bundle right after workload kubeconfig is generated instead of waiting for the machines to also be fully ready.

*Testing (if applicable):*
Unit testing and e2e

### DIFF
--- a/pkg/workflows/create_test.go
+++ b/pkg/workflows/create_test.go
@@ -2,6 +2,7 @@ package workflows_test
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/golang/mock/gomock"
@@ -13,6 +14,7 @@ import (
 	writermocks "github.com/aws/eks-anywhere/pkg/filewriter/mocks"
 	"github.com/aws/eks-anywhere/pkg/providers"
 	providermocks "github.com/aws/eks-anywhere/pkg/providers/mocks"
+	"github.com/aws/eks-anywhere/pkg/task"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/workflows"
 	"github.com/aws/eks-anywhere/pkg/workflows/interfaces/mocks"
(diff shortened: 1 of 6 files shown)


vivek-koppuru:
Currently facing a pod didn't tolerate issue because the node is tainted with not-ready, checking how to circumvent to get the logs still

/hold

vivek-koppuru:
This PR needs to be on hold until we upgrade our troubleshoot dependency to `v0.33.0`+ to leverage `podSpec` field here: https://troubleshoot.sh/docs/collect/run-pod/#podspec-required

pkg/workflows/create_test.go
```diff
@@ -310,3 +316,48 @@ func TestCreateWorkloadClusterRunSuccess(t *testing.T) {
...
+	}
+
+	gomock.InOrder(
+		test.clusterManager.EXPECT().CreateWorkloadCluster(
+			test.ctx, test.bootstrapCluster, test.clusterSpec, test.provider,
+		).Return(nil, errors.New("test")),
+		test.clusterManager.EXPECT().SaveLogsManagementCluster(
+			test.ctx, test.clusterSpec, test.bootstrapCluster,
+		),
+		test.clusterManager.EXPECT().SaveLogsWorkloadCluster(
```
g-gaston, line 336: This one should not be executed, right? It should only be executed if `CreateWorkloadCluster` succeeds (if not, there is not kubeconfig)

pkg/workflows/create_test.go
```diff
@@ -310,3 +316,48 @@ func TestCreateWorkloadClusterRunSuccess(t *testing.T) {
...
+			test.ctx, test.provider, test.clusterSpec, nil,
+		),
+	)
+	err := task.NewTaskRunner(&workflows.CreateWorkloadClusterTask{}, test.writer).RunTask(test.ctx, &commandContext)
+	if err == nil {
+		t.Fatalf("expected error from task")
+	}
+
+	gomock.InOrder(
```
g-gaston, line 345: Can we separate this in a new test?

pkg/workflows/create_test.go
vivek-koppuru: Currently how it is written, it get's called but then returns early. We can add a condition to not call this method instead but wanted to keep it as it is.

vivek-koppuru: That's why the last arg is nil

jiayiwang7 (approved):

PR #2439, opened 2022-06-17
Watch child objects with OwnerReferences

g-gaston:

*Description of changes:*

On every reconciliation loop, check if the cluster child objects contain
an owner reference linked to that cluster. If they don't, add it and
update them.

When watching child objects, use their OwnerReferences to enqueue
reconcile requests for Clusters.

*Testing (if applicable):*
Unit and integration tests

### DIFF
--- a/controllers/controllers/cluster_controller.go
+++ b/controllers/controllers/cluster_controller.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -15,10 +16,15 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/aws/eks-anywhere/controllers/controllers/clients"
 	"github.com/aws/eks-anywhere/controllers/controllers/clusters"
+	"github.com/aws/eks-anywhere/controllers/controllers/utils/handlerutil"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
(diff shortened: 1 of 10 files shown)


controllers/controllers/cluster_controller_test_test.go
```diff
@@ -0,0 +1,112 @@
+package controllers_test
```
g-gaston, line 1: I created this file to have tests in `controllers_test` package
The rest of the current tests in `controllers/controllers/cluster_controller_test.go` should we move to this package if possible.

controllers/controllers/cluster_controller.go
```diff
@@ -32,32 +38,70 @@ const (
...
+	validator               *vsphere.Validator
+	defaulter               *vsphere.Defaulter
+	tracker                 *remote.ClusterCacheTracker
+	buildProviderReconciler BuildProviderReconciler
 }
 
-func NewClusterReconciler(client client.Client, log logr.Logger, scheme *runtime.Scheme, govc *executables.Govc, tracker *remote.ClusterCacheTracker) *ClusterReconciler {
+// TODO: this is not ideal and will need a refactor. I will follow up but for now this
+// allows us to decouple the cluster reconciler main logic from provider specific logic
+type BuildProviderReconciler func(datacenterKind string, client client.Client, log logr.Logger, validator *vsphere.Validator, defaulter *vsphere.Defaulter, tracker *remote.ClusterCacheTracker) (clusters.ProviderClusterReconciler, error)
```
g-gaston, line 51: I will come up with a better way of doing this, it's just a minimal change to allow me to dummy/mock the providers specific reconciler so I can test the cluster reconciler logic alone

I was thinking about having a `ClusterProviderReconcilerRegistry` that returns the proper reconciler based on the `Cluster`. This would be built as any other dependency and injected in the constructor. But TBD until I get there

controllers/controllers/cluster_controller.go
```diff
@@ -32,32 +38,70 @@ const (
...
+
+func NewClusterReconciler(client client.Client, log logr.Logger, scheme *runtime.Scheme, govc *executables.Govc, tracker *remote.ClusterCacheTracker, buildProviderReconciler BuildProviderReconciler) *ClusterReconciler {
```
jiayiwang7, line 53: will arg govc be removed? will validator and defaulter be part of the `buildProviderReconciler`? I don't quite follow why this decouples the main logic from provider logic

controllers/controllers/cluster_controller.go
```diff
@@ -32,32 +38,70 @@ const (
...
+		validator:               validator,
+		defaulter:               defaulter,
+		tracker:                 tracker,
+		buildProviderReconciler: buildProviderReconciler,
 	}
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	childObjectHandler := handlerutil.ChildObjectToClusters(r.log)
```
jiayiwang7, line 69: What's the main reason to set up ownerref? Is it mainly for GC and object adoption?
Also reason to handle enqueue request func ourselves? So the child object changes can be sent to the corresponding cluster reconciler?

g-gaston: About why setting up owner ref, good question. Copy paste from the design ticket since I can't link it:


> Objects like provider machine configs, flux config, datacenter config or identity provider configs are linked in the cluster but there is no back reference.
>
> This is useful to trigger cluster reconciliation loops when this "child" objects change and to (potentially, tbd if we want this or not) delete them if the cluster is deleted and there are no more clusters referencing that object.
>
> We should explore owner references for this. It's probably the right k8s pattern, it has built it support for deletion (we can avoid an object being deleted until it's parent object is deleted) and supports many to one relationships.
>
> # Solution
>
>At the beginning of the reconciliation loop, the cluster controller will retrieve all linked objects to the `Cluster`. If it hasn't been added yet, it will add the `Cluster` as a `metadata.OwnerReferences`.
>
> ```go
>ref := metav1.OwnerReference{
>	APIVersion:         anywherev1.GroupVersion().String(),
>	Kind:               "Cluster",
>	Name:               cluster.GetName(),
>       UID:                cluster.GetUID(),
>       BlockOwnerDeletion: pointer.BoolPtr(true),
>       Controller:         pointer.BoolPtr(true),
>}
>
> owners := object.GetOwnerReferences()
> owners = append(owners, ref)
> object.SetOwnerReferences(owners)
> ```
>
> Note: cross namespace owner references are no allowed, so all eksa objects linked to a `Cluster` will need to be in the same namespace as that `Cluster`.
This is already a validation we currently run and the present code assumes this is always true. But it's worth highlighting this since we won't be able to use owner references if we decide to support cross namespace references from the `Cluster`.

What do you mean exactly by `Also reason to handle enqueue request func ourselves?`?
We are just determining which clusters need to be reconciled when a child object changes. Were you thinking about another alternative?

jiayiwang7 (approved):

PR #2441, opened 2022-06-17
Proposed design - CAPC manager --metrics-bind-addr setting.

jweite-amazon:
*Issue #, if available:*
None

*Description of changes:*
Proposed design for EKS-A/CAPC change to support override of CAPC manager metrics-bind-addr by provisioner to enable metrics scraping, as requested by a stakeholder.

*Testing (if applicable):*
N/A

### DIFF
--- a/designs/cloudstack-metrics-bind-addr.md
+++ b/designs/cloudstack-metrics-bind-addr.md
@@ -1,0 +1,176 @@
+# Supporting Provisioner-set metrics-bind-addr for CAPC (and possibly other infrastructure provider) Managers
+
+## Introduction
+
+**Problem:**
+
+A need has been expressed for the scraping of CAPC Manager metrics, primarily to
+facilitate alerting on reconciliation failures. As do most KubeBuilder-based
+controllers, the CAPC manager offers metrics on an interface/port defined by its
+*--metrics-bind-addr* command-line parameter.  For security purposes This
+parameter defaults to *localhost:8080*, making the port only visible to other
+containers in the manager's pod (sidecars).  To expose this port from the pod
+one needs to bind it to the pod's NIC interface, typically by setting it to
+*0.0.0.0:8080* or just *:8080* (since the actual IP address that will be assigned to the pod is not easily known).
+
+CAPC users can easily add this argument to an overridden infrastructure-components.yaml.
+EKS-A users, however, have no available mechanism for doing so to the
+infrastructure-components.yaml that EKS-A will use.
+
+This design proposes a simple mechanism for supporting this.  The scope of the proposed change is CAPC,
+though CAPV, CAPD and CAPT are known to support the same parameter (and other providers likely do too.
+
+### Tenets
+
+* ****Simple:**** simple to use, simple to understand, simple to maintain
+* ****Minimally Impactful:**** to the EKS-A code base.
+* ****Secure by Default:**** if unused defaults to the most secure option
+
+### Goals and Objectives
+
+As a Kubernetes administrator I want to be able to expose the CAPC Manager's
+metrics port from the pod running it so I can scrape metrics from it.
+
+### Statement of Scope
+
+**In scope**
+
+* Add support for overriding the default CAPC manager's metrics-bind-addr at
+cluster provisioning time.**
+
+**Not in scope**
+
+* Doing this for all EKS-A providers.
+
+**Future scope**
+
+* Implement this for the other EKS-A providers.
+
+
+## Overview of Solution
+
+We propose to
+* Modify the CAPC infrastructure-components.yaml to define argument *--metrics-bind-addr* for the CAPC Manager deployment, from an optional environment variable which defaults to the same setting as the CAPC manager uses for this parameter.
+
+* Modify function EnvMap() in cloudstack.go, which sets up the environment that *clusterctl init* will run under, to read the value of a new environment
+variable (if defined by the provisioner in their *eksctl anywhere create cluster* session) into the map that it creates.
+
+## Solution Details
+
+### CAPC infrastructure-components.yaml change
+
+The deployment spec for the CAPC manager will be modified to define a *--metrics-bind-address*
+for the deployment's pod template's manager container, resolved with an environment
+variable substitution defaulting to the original secure default used by the manager:
+
+```
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+...
+  name: capc-controller-manager
+  namespace: capc-system
+spec:
+...
+    spec:
+      containers:
+      - name: manager
+        args:
+        - --leader-elect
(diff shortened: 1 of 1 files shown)


designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+containers in the manager's pod (sidecars).  To expose this port from the pod
+one needs to bind it to the pod's NIC interface, typically by setting it to
+*0.0.0.0:8080* or just *:8080* (since the actual IP address that will be assigned to the pod is not easily known).
+
+CAPC users can easily add this argument to an overridden infrastructure-components.yaml.
+EKS-A users, however, have no available mechanism for doing so to the
+infrastructure-components.yaml that EKS-A will use.
+
+This design proposes a simple mechanism for supporting this.  The scope of the proposed change is CAPC,
+though CAPV, CAPD and CAPT are known to support the same parameter (and other providers likely do too.
```
maxdrib, line 21: nit: missing `)`

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+
+	envMap := make(map[string]string)
+	for _, key := range requiredEnvs {
+		if env, ok := os.LookupEnv(key); ok && len(env) > 0 {
+			envMap[key] = env
+		} else {
+			return envMap, fmt.Errorf("warning required env not set %s", key)
+		}
+	}
+    for _, key := range optionalEnvs {
```
maxdrib, line 130: nit: spacing is off

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+	return envMap, nil
+}
+```
+
+#### Alternative 1:
+
+As an alternative to the provisioner passing the desired value for the metrics-bind-addr
+through an environment variable, a new parameter for this could be added to the CloudStackDatacenter struct,
```
maxdrib, line 142: nit: CloudStackDatacenterConfig*** struct

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+security-impacting decisions about their clusters already.
+
+## Testing
+
+The new code will be covered by unit and e2e tests, and the e2e framework will be extended to support cluster creation across multiple Cloudstack API endpoints.
+
+The following e2e test will be added:
+
+* provision a Cloudstack cluster with metrics-bind-addr overridden.  Confirm that the resulting cluster's
+CAPC manager pod is configured with this address.
```
maxdrib, line 176: what does this confirmation look like? Currently we're able to use the kubectl client in eks-a to query eks-a resources. I'm not sure how much we'd like to tweak that to fetch and assert on provider-specific resources, like the capc manager pod. Shouldn't a CAPC e2e test, combined with EKS-A unit tests be sufficient? Otherwise it feels like we're testing a little too deeply from eks-a's perspective.

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Backwards Compatibility
+
+The proposed solution is completely backward compatible, as it makes no
+non-defaulting interface changes.  When not specified EKS-A/CAPC provision
+a cluster with the same metric-bind-addr as currently used.
+
+The Alternative Solution 1 proposes a non-breaking interface change.
+
+## User Experience
```
jiayiwang7, line 160: can we remove this if no content?

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+        args:
+        - --leader-elect
+        - --metrics-bind-addr=${CAPC_MANAGER_METRICS_BIND_ADDR:-localhost:8080}
+        command:
+        - /manager
+        image: localhost:5000/cluster-api-provider-cloudstack:v0.4.5
+...
+```
+
+### EKS-A change to expose this environment variable, if defined in the provisioner's environment at the time of *eksctl anywhere cluster create*:
```
jiayiwang7, line 87: can we talk a bit about upgrade case? For example, if user specify this env during create, and not specify it during upgrade, will the capas pod be recreated with default metrics address? Also if `CloudStackMetricsBindAddr` is the only thing changed when running upgrade cmd, how do we trigger the components upgrade run.

jweite-amazon: Agreed Joey, great point.  I think that makes the case for formally including this parameter in one of the EKS-A CRDs so it gets preserved at creation time for reuse in subsequent activities.  Probably CloudStackDataCenter unless there was interest in adopting this across all the providers.

jweite-amazon: After reading your further comments @g-gaston, I think I'm understanding your point being that use of CAPI providers is an implementation detail, and you want to leave the door open to EKS-A using other kinds of providers in the future.

If so, fair enough.  Still, there is a need demonstrated by at least one key stakeholder for configuration of CAPC, so I think it serves the product to allow for this for this specific provider (and perhaps the other EKS-A CAPI providers individually as required).  Therefore, I'll re-propose that this new attribute go in the CloudStackDataCenter CRD, as that's the current home for all CAPC-specific configuration.

designs/cloudstack-metrics-bind-addr.md
```diff
@@ -0,0 +1,176 @@
...
+    for _, key := range optionalEnvs {
+		if env, ok := os.LookupEnv(key); ok && len(env) > 0 {
+			envMap[key] = env
+		}
+	}
+	return envMap, nil
+}
+```
+
+#### Alternative 1:
```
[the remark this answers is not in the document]
g-gaston: I see the benefit for us of having a common implementation for the different providers. However, if we always follow that approach, we will never completely move to a "future forward design", there will always be a piece of software that hasn't been updated.

I would argue we could start using `ControllerManagerConfig` where we can today, and contribute upstream to make it possible where it's not yet. Also, that's assuming we will ever want this for other providers.

Btw, even if we use command flags, it won't be consistent across providers: capv uses an old name for the [metrics flag](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/main.go#L73).

jweite-amazon:
After further deliberation, we've reached agreement that EKS-A should not become inextricably bound to CAPI by virtue of exposing CAPI-provider-specific parameters through EKS-A's API.  We will instead submit a PR to modify EKS-A's CAPC infrastructure-components.yaml to incorporate kube-rbac-proxy, securely exposing the metrics port from the CAPC controller pod in all EKS-A CloudStack clusters.

PR #2442, opened 2022-06-17
fix vsphere env var race conditions during e2e tests

gwesterfieldjr:
This is to account for govc envvars being needed by test runners and vsphere tests. To account for this we name them differently on code build box and override to govc vars at time of vm creation


pokearu (approved):
/approve
/lgtm

PR #2443, opened 2022-06-17
CloudStack Multi-Endpoint Design Proposal -- Multiple K8s Secrets via Yaml

rejoshed:
*Issue #, if available:* N/A

*Description of changes:*
The changes mostly describe themselves. Basically I'm making the case that now that we need multiple endpoints we should have multiple secrets.

Now that we should have multiple secrets, we should move to a modern format that easily supports that and can be generalized across providers.

*Testing (if applicable):* N/A

### DIFF
--- a/designs/cloudstack-multiple-endpoints.md
+++ b/designs/cloudstack-multiple-endpoints.md
@@ -47,7 +47,7 @@ will need to be available on *all* the Cloudstack API endpoints. We will validat
 
 ### Interface changes
 Currently, the CloudstackDataCenterConfig spec contains:
-```
+``` Go
 type CloudStackDatacenterConfigSpec struct {
 	// Domain contains a grouping of accounts. Domains usually contain multiple accounts that have some logical relationship to each other and a set of delegated administrators with some authority over the domain and its subdomains
 	Domain string `json:"domain"`
@@ -62,7 +62,7 @@ type CloudStackDatacenterConfigSpec struct {
 
 We would instead propose to gradually deprecate all the existing attributes and instead, simply include a list of AvailabilityZone objects like so
 
-```
+``` Go
 type CloudStackDatacenterConfigSpec struct {
 	// Deprecated
 	Domain string `json:"domain,omitempty"`
@@ -79,7 +79,7 @@ type CloudStackDatacenterConfigSpec struct {
 
 where each AvailabilityZone object looks like
 
-```
+``` Go
 type CloudStackAvailabilityZone struct {
 	// Name would be used to match the availability zone defined in the datacenter config to the credentials passed in from the cloud-config ini file
 	Name string `json:"name"`
--- a/designs/cloudstack-multiple-endpoints.md
+++ b/designs/cloudstack-multiple-endpoints.md
@@ -138,7 +138,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: "someendpoint:8080/client/api"
-  namespace: system
+  namespace: eksa-system
 type: Opaque
 data:
   apikey: "redacted1"
@@ -148,7 +148,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: "someotherendpoint:8080/client/api"
-  namespace: system
+  namespace: eksa-system
 type: Opaque
 data:
   apikey: "redacted2"
--- a/designs/cloudstack-multiple-endpoints.md
+++ b/designs/cloudstack-multiple-endpoints.md
@@ -195,7 +195,7 @@ api-url    = http://172.16.0.3:8080/client/api
     provide details. We should standardize on a general way to pass multiple secrets per provider. The data present in
     these secrets should be entirely provider specific. In this manner any arbitrary data form that a specific provider
     needs can be dictated by the provider yet the underlying secret passing mechanism can remain the same.
-8. ini what!? The ini format came out in the same decade as Bact to the Future. Great movie, but... : )
+8. ini what!? The ini format came out in the same decade as Back to the Future. Great movie, but... : )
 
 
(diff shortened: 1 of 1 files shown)


rejoshed:
``` ini
[section]
asdf = blah
```

designs/cloudstack-multiple-endpoints.md
```diff
@@ -131,20 +129,36 @@ for availabilityZone in availabilityZones:
...
+In a multi-endpoint Cloudstack cluster, each endpoint may have its own credentials.
+
+We propose that we make the leap to yaml k8s secrets for endpoint credentials of the form:
+
+``` yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "someendpoint:8080/client/api"
+  namespace: system
```
maxdrib, line 141: eksa-system?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -163,6 +177,27 @@ api-url    = http://172.16.0.3:8080/client/api
 ...
 ```
 
+1. `Global`, `AvailabilityZone2`, and `AvailabilityZone3` are entirely arbitrary names that are unnecessary.
+2. If these tags mismatch between the name in the AvailabilityZone and here there will be an error.
+3. If we use validation, then we can prevent mismatched names, but what about modifying this secret in an existing cluster?
+    - Validation doesn't operate on an existing cluster unless the secret is updated through EKS-A.
+    - If an existing cluster wants to update a secret, the entire secret must be modified together. If this results in
```
maxdrib, line 184: Is there a mechanism you'd propose to support updating secrets?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -163,6 +177,27 @@ api-url    = http://172.16.0.3:8080/client/api
...
+    an error, the entire set of AvalailabilityZones are at risk.
+4. To do validation of the secret ini with multiple tags the DataCenter needs to match the tags to the names of its
+   AvailabilityZones. This is fine, but with yaml the secrets can be iterated on and checked individually without
+   concern for the tags.
```
maxdrib, line 188: we still need a way to map from the secrets in the yaml file to the AvailabilityZones defined in the datacenterconfig to support failing fast in case a user has made a mistake. We can do that via the ACS url or a more friendly name. It looks like you're proposing using the ACS url, is that correct?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -131,20 +129,36 @@ for availabilityZone in availabilityZones:
...
-only as a list corresponding to AvailabilityZones. Currently, these credentials are passed in via environment variable, which contains a base64 encoded .ini file that looks like
-
-```
-[Global]
-api-key    = redacted
-secret-key = redacted
-api-url    = http://172.16.0.1:8080/client/api
+In a multi-endpoint Cloudstack cluster, each endpoint may have its own credentials.
+
+We propose that we make the leap to yaml k8s secrets for endpoint credentials of the form:
```
maxdrib, line 134: I'm not clear on what you're asking for here. Just that the input file format to EKS-A change? Or should EKS-A also apply these Secrets to the cluster at some point? 

If you're asking for the latter:

1. how would CAPC know which secrets to use for which CloudstackFailureDomain?
2. would CAPC support a gradual migration to using these secrets by supporting *both* an environment variable input, as well as looking for secrets on the cluster?

designs/cloudstack-multiple-endpoints.md
```diff
@@ -163,6 +177,27 @@ api-url    = http://172.16.0.3:8080/client/api
...
+   AvailabilityZones. This is fine, but with yaml the secrets can be iterated on and checked individually without
+   concern for the tags.
+5. CAPC wants a k8s secret anyway.
+6. Currently we create a k8s secret in a highly convoluted maner that requires b64 encoding an environment
+   variable that is expanded during clusterctl expansion but decoded in the CAPC controllers. The b64
+   encoding provides no extra security. It is used purely as a way to avoid potential command line special characters 
+   present in the cloud-config ini file.
+7. Other providers _will_ need multiple secrets soon enough. One in the pipeline already requires a CSV as a format to
+    provide details. We should standardize on a general way to pass multiple secrets per provider. The data present in
+    these secrets should be entirely provider specific. In this manner any arbitrary data form that a specific provider
```
maxdrib, line 196: Could you elaborate on what other providers you're referring to here? Additionally, which pipeline are you referring to?

rejoshed: I'll check with Vignesh. Pipeline meaning other providers that are not yet GA.

One of them requires a CSV per Node.

designs/cloudstack-multiple-endpoints.md
rejoshed: The particular point you've commented on is more about having multiple secrets to isolate them from each other.

Otherwise, I think an update of a secret could be done directly on a cluster, or through EKS-A.

As for mechanisms, if EKS-A was applying a secret and verifying it, then maybe it could also accept updates thereof. How would a user update a secret today? I think currently they'd have to manually apply it.

designs/cloudstack-multiple-endpoints.md
rejoshed: > Just that the input file format to EKS-A change? 
This and

> should EKS-A also apply these Secrets to the cluster at some point?
this.

Though I'm not sure EKS-A would need to explicitly apply the secrets file or if it could be inserted via kustomize.

Currently the secret is applied via clusterctl via EKS-A from an expanded environment variable against the CAPC manifest. (whew!)

While I would like it if the end user could run kustomize on the CAPC config directory to generate their own CAPC manifest, that probably gives the users a bit too much of a foot gun for EKS-A's liking. Barring that, EKS-A would probably need to apply the secrets.

PR #2446, opened 2022-06-20
Expose tinkerbell services on workload cluster using an lb

abhinavmpandey08:
*Description of changes:*
Expose tinkerbell services, hegel and tink-server, using a LoadBalancer to allow the clusters to scale up. 
A LoadBalancer is required because when the new nodes come up, they need to contact tink-server and hegel to provision, which requires these 2 services to be externally accessible.
This also required a change to the tinkerbell helm chart https://github.com/aws/eks-anywhere-build-tooling/pull/957
which will need to merge before this PR

*Testing (if applicable):*
Created baremetal clusters and verified that the tinkerbell services were exposed properly

### DIFF
--- a/pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
+++ b/pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
@@ -48,7 +48,7 @@ func getDiskPart(disk string) string {
 	}
 }
 
-func GetDefaultActionsFromBundle(b v1alpha1.VersionsBundle, disk, osImageOverride string, tinkerbellIp string, osFamily OSFamily) []ActionOpt {
+func GetDefaultActionsFromBundle(b v1alpha1.VersionsBundle, disk, osImageOverride, tinkerbellLocalIp, tinkerbellLBIp string, osFamily OSFamily) []ActionOpt {
 	var diskPart string
 
 	defaultActions := []ActionOpt{
@@ -60,7 +60,7 @@ func GetDefaultActionsFromBundle(b v1alpha1.VersionsBundle, disk, osImageOverrid
 		defaultActions = append(defaultActions,
 			withNetplanAction(b, diskPart, osFamily),
 			withBottlerocketBootconfigAction(b, diskPart),
-			withBottlerocketUserDataAction(b, diskPart, tinkerbellIp),
+			withBottlerocketUserDataAction(b, diskPart, tinkerbellLocalIp),
 			withRebootAction(b),
 		)
 	} else {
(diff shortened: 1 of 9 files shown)


abhinavmpandey08:
/hold
for https://github.com/aws/eks-anywhere-build-tooling/pull/957

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -156,8 +156,8 @@ func withDisableCloudInitNetworkCapabilities(b v1alpha1.VersionsBundle, disk str
 	}
 }
 
-func withTinkCloudInitAction(b v1alpha1.VersionsBundle, disk string, tinkerbellIp string) ActionOpt {
-	metadataString := fmt.Sprintf("\"http://%s:50061\"", tinkerbellIp)
+func withTinkCloudInitAction(b v1alpha1.VersionsBundle, disk string, tinkerbellLocalIp, tinkerbellLBIp string) ActionOpt {
+	metadataString := fmt.Sprintf("\"http://%s:50061\", \"http://%s:50061\"", tinkerbellLocalIp, tinkerbellLBIp)
```
ptrivedi, line 160: A brief comment here would be nice explaining how multiple urls work in the metadata_urls and how local IP vs. LB IP are relevant in bootstrap cluster vs. target cluster cases respectively

pkg/providers/tinkerbell/create.go
```diff
@@ -78,13 +78,14 @@ func (p *Provider) PostWorkloadInit(ctx context.Context, cluster *types.Cluster,
 
 	err := p.stackInstaller.Install(
 		ctx,
-		clusterSpec.VersionsBundle.Tinkerbell.TinkerbellStack,
+		clusterSpec.VersionsBundle.Tinkerbell,
```
ptrivedi, line 81: Looking at Install() below, wonder why this change was really needed. But no biggie either way.

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -158,7 +167,14 @@ func (s *Installer) Install(ctx context.Context, bundle releasev1alpha1.Tinkerbe
 		},
 		rufio: map[string]interface{}{
 			deploy: true,
-			image:  bundle.Rufio.Image.URI,
+			image:  bundle.TinkerbellStack.Rufio.Image.URI,
+		},
+		loadBalancer: map[string]interface{}{
+			"enabled": s.loadBalancer,
+			"ip":      tinkerbellIP,
```
ptrivedi, line 174: nit: It would be nice to call this tinkerbellVIP or tinkerbellServiceVIP, just so it is clear that it is a VIP for target cluster tinkerbell stack.

pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
abhinavmpandey08: Will add that 👍

pkg/providers/tinkerbell/stack/stack.go
abhinavmpandey08: This function is generic to both installing tinkerbell on bootstrap and on workload.
Calling this `tinkerbellVIP` would make it specific to installing this on workload since on bootstrap, it won't be a vip

PR #2451, opened 2022-06-20
Implement backreferences from CAPI to EKSA cluster

g-gaston:

*Description of changes:*

Using labels in CAPI objects to reference the parent eksa Cluster by its
name and namespace.

*Testing (if applicable):*

Unit tests

### DIFF
--- a/controllers/controllers/utils/handlerutil/capiobjects.go
+++ b/controllers/controllers/utils/handlerutil/capiobjects.go
@@ -1,0 +1,41 @@
+package handlerutil
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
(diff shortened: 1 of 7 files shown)


jiayiwang7 (approved):

controllers/controllers/utils/handlerutil/capiobjects.go
```diff
@@ -0,0 +1,41 @@
...
+		labels := o.GetLabels()
+		clusterName, ok := labels[clusterapi.EKSAClusterLabelName]
+		if !ok {
+			// Object not managed by a eks-a Cluster, don't enqueue
+			log.V(6).Info("Object not managed by an eks-a Cluster, ignoring", "type", fmt.Sprintf("%T", o), "name", o.GetName())
+			return nil
+		}
+
+		clusterNamespace := labels[clusterapi.EKSAClusterLabelNamespace]
+		if clusterNamespace == "" {
```
jiayiwang7, line 28: when will this happen? should we fill the namespace and enque?

g-gaston: It should never happen under normal circumstances
But I think the code should be able to handle that scenario since labels are a 'free form" field and they can be updated by other controllers, humans, etc.

I don't think we can fill the namespace since we don't know were the eksa cluster is.

jiayiwang7 (approved):

PR #2463, opened 2022-06-21
clean up vsphere test runner instances after every build

gwesterfieldjr:


internal/test/e2e/run.go
```diff
@@ -140,17 +140,12 @@ func RunTests(conf instanceRunConf) (testInstanceID string, testCommandResult *t
...
-	defer func(successRef *bool) {
-		if *successRef {
-			err := testRunner.decommInstance(conf)
-			if err != nil {
-				logger.V(1).Info("WARN: Failed to decomm e2e test runner instance", "error", err)
-			}
-		} else {
-			logger.V(1).Info("Persist instance on failure", "instance", instanceId)
+	defer func() {
+		err := testRunner.decommInstance(conf)
```
maxdrib, line 144: it seems like this should be configurable. We may not want to do this necessarily for non-baremetal providers, right?

abhinavmpandey08 (approved):

internal/test/e2e/run.go
gwesterfieldjr: Or are you saying add back ability to persist on failure? and make that configurable?

gwesterfieldjr: I can see need to allow persisting on failure which is why i put it originally, but for baremetal its an issue since we run pxe server on the test instance and it causes future test runs to fail. So we have to delete for baremetal or future tests will conflict and get routed to wrong pxe server

maxdrib: I was referring more to the latter. I'm not intimately familiar with the test running infrastructure. If some tests are failing and we can't investigate because the ec2 instances the tests are running on keep getting deleted, that seems like it could be a problem. My question is - would this affect our ability to investigate test failures? Or if something's wrong, would we have to make a code change to prevent this cleanup and unblock our investigation?

gwesterfieldjr: No since for other providers they run in ec2 and it just returns nil. So functionality is as it has always been.

maxdrib (approved):

PR #2481, opened 2022-06-21
Use proper artifact folder for tarball upload

abhay-krishna:
Use proper artifact folder for tarball upload.


pokearu (approved):
/approve
/lgtm

PR #2529, opened 2022-06-23
Fixing backwards compatibility bug and adding new e2e test

maxdrib:
*Issue #, if available:*

*Description of changes:*
In order to ensure our users are able to smoothly upgrade between eks-a versions, we need a test running automatically in our pipeline to make sure the changes we're making preserve backwards compatibility. Identified one bug as a result of this issue where some optional field was not parseable.

*Testing (if applicable):*
Testing locally against CI environment

### DIFF
--- a/test/e2e/upgrade_from_latest_test.go
+++ b/test/e2e/upgrade_from_latest_test.go
@@ -50,7 +50,7 @@ func TestVSphereKubernetes120BottlerocketUpgradeFromLatestMinorRelease(t *testin
 	)
 }
 
-func TestCloudstackKubernetes120UpgradeFromLatestMinorRelease(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/e2e/upgrade_from_latest_test.go
```diff
@@ -50,6 +50,23 @@ func TestVSphereKubernetes120BottlerocketUpgradeFromLatestMinorRelease(t *testin
 	)
 }
 
+func TestCloudstackKubernetes120UpgradeFromLatestMinorRelease(t *testing.T) {
```
taneyland, line 53: nit: to maintain consistency with the naming. If we do a regex comparison for CloudStack tests, this would slip through since it is case sensitive
 ```suggestion
func TestCloudStackKubernetes120UpgradeFromLatestMinorRelease(t *testing.T) {
```

wanyufe (approved):

PR #2540, opened 2022-06-23
Remove unused tinkerbell components from the bundle

abhinavmpandey08:
*Description of changes:*
Remove pbnj, tink-cli and the tinkerbell manifests from the bundle as they are no longer needed


abhinavmpandey08:
/woof
take that codecov

PR #2555, opened 2022-06-27
cleanup tinkerbell test environment between test runs

gwesterfieldjr:
During CI multiple Tinkerbell tests run from the same test runner. If one of the tests fail, the subsequent tests also fail since kind cluster and boots container from previous test are still running and cause issue. This cleans up kind and docker environment after each tinkerbell test that runs during CI build.

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -108,6 +108,11 @@ func NewClusterE2ETest(t *testing.T, provider Provider, opts ...ClusterE2ETestOp
 
 	e.T.Cleanup(func() {
 		e.CleanupVms()
+
+		tinkerbellCIEnvironment := os.Getenv(TinkerbellCIEnvironment)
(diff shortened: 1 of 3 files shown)


test/framework/cluster.go
```diff
@@ -612,6 +617,13 @@ func (e *ClusterE2ETest) CleanupVms() {
 	}
 }
 
+func (e *ClusterE2ETest) CleanupDockerEnvironment() {
+	e.T.Logf("cleanup kind enviornment...")
+	e.Run("kind", "get", "clusters", "|", "xargs", "-t", "-n1", "kind", "delete", "cluster", "--name")
```
abhinavmpandey08, line 622: you can ignore this, but this can also be achieved using 
```
kind delete clusters --all
```

abhinavmpandey08 (approved):
/approve

PR #2567, opened 2022-06-27
Added notes to docs about supported cluster tasks for Bare Metal

chrisnegus:
*Description of changes:*
Added docs notes to cluster management tasks to say when a feature is or is not supported for each provider (vSphere and Bare Metal). Making this a WIP while I get more information on scaling and cluster integrations.

### DIFF
--- a/docs/content/en/docs/tasks/cluster/cluster-delete.md
+++ b/docs/content/en/docs/tasks/cluster/cluster-delete.md
@@ -78,4 +78,5 @@ GitOps field not specified, clean up git repo skipped
 ```
 
 For vSphere, this will delete all of the VMs that were created in your provider.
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/tasks/cluster/cluster-delete.md
```diff
@@ -74,5 +77,5 @@ GitOps field not specified, clean up git repo skipped
 🎉 Cluster deleted!
 ```
 
-This will delete all of the VMs that were created in your provider.
+For vSphere, this will delete all of the VMs that were created in your provider.
```
ptrivedi, line 80: We should add a note saying for bare metal the servers will be powered off if the BMC information had been provided.

PR #2574, opened 2022-06-27
Generate failureDomains from availabilityZones for CAPC

wongni:
*Issue #, if available:*

*Description of changes:*
This PR uses CAPC v1beta2 and generate failureDomains instead of Zones to support mutiple ACS endpoints. 

*Testing (if applicable):*


pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -177,7 +177,7 @@ func (v *CloudStackDatacenterConfig) SetDefaults() {
 				Account:               v.Spec.Account,
 				Domain:                v.Spec.Domain,
 				ManagementApiEndpoint: v.Spec.ManagementApiEndpoint,
-				CredentialsRef:        "Global",
+				CredentialsRef:        "global",
```
maxdrib, line 180: let's also resolve the TODO above in the Validate() method. I think we can call Validate() from cloudstack.go or validator.go

controllers/resource/fetcher.go
```diff
@@ -589,20 +589,22 @@ func MapMachineTemplateToVSphereMachineConfigSpec(vsMachineTemplate *vspherev1.V
...
-	var zones []anywherev1.CloudStackZone
-	for _, csZone := range csCluster.Spec.Zones {
-		zones = append(zones, anywherev1.CloudStackZone{
-			Id:   csZone.ID,
-			Name: csZone.Name,
-			Network: anywherev1.CloudStackResourceIdentifier{
-				Id:   csZone.Network.ID,
-				Name: csZone.Network.Name,
+	var azs []anywherev1.CloudStackAvailabilityZone
+	for _, fd := range csCluster.Spec.FailureDomains {
```
maxdrib, line 593: I think we've answered this before - but will there be any point at which this updated eks-a controller is reading in v1beta1 CAPC resources, and has to map them back to AZ's? This code assumes we always have FailureDomains present

go.mod
```diff
@@ -185,4 +185,5 @@ replace (
 
 	// need the modifications eksa made to the capi api structs
 	sigs.k8s.io/cluster-api => github.com/mrajashree/cluster-api v1.1.3-custom
+	sigs.k8s.io/cluster-api-provider-cloudstack => github.com/rejoshed/cluster-api-provider-cloudstack v0.4.5-rc3.0.20220714221707-f31eecbdff3a
```
maxdrib, line 188: nice :) maybe add a comment explaining why this is needed

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -160,10 +161,29 @@ func (v *CloudStackDatacenterConfig) Marshallable() Marshallable {
...
+	if len(v.Spec.Zones) > 0 {
+		return errors.New("zones must be empty")
+	}
+	if len(v.Spec.AvailabilityZones) == 0 {
+		return errors.New("availabilityZones must not be empty")
+	}
+	azSet := map[string]CloudStackAvailabilityZone{}
+	for _, az := range v.Spec.AvailabilityZones {
+		if _, exists := azSet[az.Name]; exists {
+			return errors.New(fmt.Sprintf("All availabilityZone names must not be unique; found multiple %s", az.Name))
```
maxdrib, line 182:
```suggestion
			return errors.New(fmt.Sprintf("All availabilityZone names must be unique; found multiple %s", az.Name))
```

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -35,25 +35,19 @@ spec:
...
+  failureDomains:{{ range $az := .cloudstackAvailabilityZones }}
+  - name: {{ $az.Name }}
+    zone:
+      name: {{ $az.Zone.Name }}
+      network:
+        name: {{ $az.Zone.Network.Name }}
+    domain: {{ $az.Domain }}
+    account: {{ $az.Account }}
+    acsendpoint:
+      name: {{ $az.CredentialsRef }}
```
maxdrib, line 47: do we need a namespace here in addition to the credentialRef name?

maxdrib (commented):
I'd be curious to see how this was tested e2e. Does it work e2e? Is the branch of CAPC ready to roll? Would you be able to run the TestCloudStackKubernetes120UpgradeFromLatestMinorRelease test with this configuration?

Great work so far on all this, by the way. It's all so complicated but you've managed to split it up simply and effectively.

pkg/providers/cloudstack/config/template-cp.yaml
wongni: Yes, capc needs both name and namespace to get a secret. If we don't pass the namespace capc has to have a hardcoded namespace `eksa-system`.

pkg/providers/cloudstack/testdata/cluster_availability_zones.yaml
```diff
@@ -0,0 +1,110 @@
...
+spec:
+  availabilityZones:
+  - name: az-1
+    domain: domain-1
+    account: account-1
+    zone:
+      name: zone-1
+      network:
+        name: network-1
+    credentialsRef: global
```
maxdrib, line 51: could you use different credentials and configurations for each az?

maxdrib (commented):
Code LGTM, just the ask for the test case. Could you also make a note about how/when this PR could be merged, and the CAPC dependency we are waiting on?

wongni:
Splited into https://github.com/aws/eks-anywhere/pull/2766, https://github.com/wongni/eks-anywhere/pull/7, and https://github.com/wongni/eks-anywhere/pull/8

PR #2598, opened 2022-06-29
Adding e2e test support for TestCloudStackKubernetes121RedhatProxyConfig

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR refactors the e2e tests for proxy servers to have independent configuration for each provider. 

STILL TODO: Add the cloudstack proxy server configuration into Secrets Manager

*Testing (if applicable):*
Ran TestCloudStackKubernetes121RedhatProxyConfig locally against CI Cloudstack environment with proxy server's hostname instead of static IP


cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
```diff
@@ -49,6 +49,9 @@ env:
     T_HTTP_PROXY: "proxy-config-data:httpProxy"
     T_HTTPS_PROXY: "proxy-config-data:httpsProxy"
     T_NO_PROXY: "proxy-config-data:noProxy"
+    T_HTTP_PROXY_CLOUDSTACK: "proxy-config-data:httpProxyCloudStack"
+    T_HTTPS_PROXY_CLOUDSTACK: "proxy-config-data:httpsProxyCloudStack"
+    T_NO_PROXY_CLOUDSTACK: "proxy-config-data:noProxyCloudStack"
```
maxdrib, line 54: These fields should be added to secrets manager but I'm not sure how that's done. I could use some guidance for achieving this

jiayiwang7: do you think we could reuse T_HTTP_PROXY, T_HTTPS_PROXY and T_NO_PROXY?

pkg/api/v1alpha1/cluster.go
```diff
@@ -591,12 +591,14 @@ func validateProxyData(proxy string) error {
...
 	}
-	ip, port, err := net.SplitHostPort(proxyHost)
+	host, port, err := net.SplitHostPort(proxyHost)
 	if err != nil {
 		return fmt.Errorf("proxy %s is invalid, please provide a valid proxy in the format proxy_ip:port", proxy)
 	}
-	if net.ParseIP(ip) == nil {
-		return fmt.Errorf("proxy ip %s is invalid, please provide a valid proxy ip", ip)
+	if net.ParseIP(host) == nil {
+		if _, err := net.LookupIP(host); err != nil {
```
jiayiwang7, line 599: can we possibly add unit test for it?

abhinavmpandey08, line 599: maybe consolidate the checks in a single condition instead of nesting them using something like
```suggestion
        _, err := net.LookupIP(host)
	if err != nil && net.ParseIP(host) == nil {
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -591,12 +591,14 @@ func validateProxyData(proxy string) error {
...
+			return fmt.Errorf("proxy ip %s is invalid, please provide a valid proxy ip", host)
```
abhinavmpandey08, line 600: how about something like?
```suggestion
			return fmt.Errorf("proxy endpoint %s is invalid, please provide a valid proxy domain name or ip", host)
```

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
mrog: Using the same proxy would be ideal, if that's possible.

internal/test/e2e/proxy.go
```diff
@@ -14,13 +15,27 @@ func (e *E2ESession) setupProxyEnv(testRegex string) error {
...
 		}
 	}
 
 	return nil
 }
+
+func isTestProvider(testRegex string, providerName string) bool {
+	reProvider := regexp.MustCompile(fmt.Sprintf(`^.*%s.*$`, providerName))
+	return reProvider.MatchString(testRegex)
+}
```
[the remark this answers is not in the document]
maxdrib: After further conversations, there are not currently sufficient resources to spin up a new proxy server so we're inclined to proceed with the split-provider approach instead

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
maxdrib: Values added to secrets manager by @abhay-krishna

vivek-koppuru: I am assuming we don't have proxy tests for docker then, because right now those env vars assume proxy is for vsphere, so maybe need to go with this approach

test/e2e/proxy_test.go
```diff
@@ -68,7 +68,8 @@ func TestSnowKubernetes121UbuntuProxyConfig(t *testing.T) {
 		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube121)),
 		framework.WithClusterFiller(api.WithControlPlaneCount(1)),
 		framework.WithClusterFiller(api.WithWorkerNodeCount(1)),
-		framework.WithProxy(),
+		// TODO: provide separate Proxy Env Vars for Snow provider. Leaving VSphere for backwards compatibility
+		framework.WithProxy(framework.VsphereProxyRequiredEnvVars),
```
vivek-koppuru, line 72: Actually if snow works, then I don't see why cloudstack would not work with the proxy config. @jiayiwang7 we don't need wireguard connection to connect to proxy for snow as well?

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
```diff
@@ -49,6 +49,9 @@ env:
     T_HTTP_PROXY: "proxy-config-data:httpProxy"
```
jiayiwang7, line 49: can we update this to be `T_HTTP_PROXY_VSPHERE`, same for values and below

test/e2e/proxy_test.go
vivek-koppuru: ohh interesting, because that would allow us to avoid provider specific proxies. But I guess it's not a huge deal to introduce new ones now

jiayiwang7 (approved):

PR #2610, opened 2022-06-29
Modify description of MetalLB to avoid confusion

lewisdiamond:
Upstream's description uses the term "bare metal" which might be
confusing to EKS-A customers since we currently consider "bare metal" as
being non-virtualized environment.


TerryHowe (approved):
/lgtm
/approve

Issue #1047, opened 2022-01-26
Autoscaling design phase


jaxesn:
**What would you like to be added**:
Start with a design using CAS and work out plan. Feature due ~Sept
**Why is this needed**:

PR #2630, opened 2022-06-30
Add cluster autoscaling design

jiayiwang7:
*Issue #, if available:*

#1047 

*Description of changes:*

*Testing (if applicable):*


designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+* create clusters in a streamlined fashion with autoscaling capabilities
+* add autoscaling capabilities to existing EKS-A clusters
+* see my clusters to auto scale up and down worker nodes within the boundaries I limited
+
+### Statement of Scope
+
+#### In Scope
+
+* Include Cluster Autoscaler as an EKS-A curated package.
```
maxdrib, line 25: it might be helpful here to link to https://anywhere.eks.amazonaws.com/docs/concepts/packages/ to explain what a "curated package" is

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+At launch, EKS Anywhere user can specify a fixed number of nodes for each worker node group in the cluster. Once a cluster is created, EKS-A cluster controller watches this configuration for changes, generates CAPI resources, and re-applies them in the cluster. Since EKS Anywhere does not support user interaction with CAPI directly, most manual CAPI resource changes made by the user will be overwritten by the EKS-A cluster controller. This behavior prevents users from enabling Cluster Autoscaler themselves, since the required autoscaling annotations manually applied to CAPI resources will be removed by EKS-A cluster controller.
+
+This document aims to solve this limitation and provides **autoscaling** **compatibility** in EKS Anywhere, so the user can either choose to install Cluster Autoscaler as an EKS-A curated package during or after cluster creation time, or install themselves from [upstream](https://github.com/kubernetes/autoscaler/releases) after a cluster is created.
+
+### Goals and Objectives
+
+As an EKS Anywhere user, I want to:
+
+* create clusters in a streamlined fashion with autoscaling capabilities
+* add autoscaling capabilities to existing EKS-A clusters
```
maxdrib, line 18: I understand autoscaling in the context of workload clusters. A user is running some workload and consuming too much CPU/memory, so they need to scale up. But does autoscaling make sense in the context of management clusters? If so, it might be good to specify that autoscaling for both is in scope

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+* Update the autoscaling configuration to set up the min/max number of nodes in each worker node group.
+
+#### Out of Scope
+
+* Compare cluster autoscaler products (e.g. [Karpenter](https://karpenter.sh/))
+* Autoscale at the pod level
+* Control Plane nodes autoscaling
+
+#### Future Scope
+
```
maxdrib, line 40: you might also consider adding support for additional configurations on the Cluster Autoscaler. It sounds like the initial scope is just using default configurations but I'm sure there are tweaks that users would like to modify outside of those

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+Note: the cluster will not have any autoscaling functionality until the Cluster Autoscaler controller is properly configured and installed in the cluster. Setting up the MachineDeployment annotations only guarantees that Cluster Autoscaler can recognize the worker node group and its autoscaling configuration. We will add a warning message in CLI logs when the autoscaling is configured in EKS-A cluster spec but autoscaler is not installed in the cluster.
+
+### Cluster Autoscaler Controller
+
+Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster. Specifically, it can:
+
+* increase the size of the cluster when there are pending pods due to insufficient resources.
+* decrease the size of the cluster when some nodes have low utilization for a significant amount of time.
```
maxdrib, line 112: how are "insufficient resources" and "low utilization" defined? Is this something a user would be able to control? Things like time durations, specific metrics to monitor (avg vs max utilization), etc.

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+Cluster Autoscaler provides the flexibility to enable autoscaling, whether it is within the cluster, or for a remote workload cluster. Cluster Autoscaler can be:
+
+* deployed in the management cluster to autoscale the management cluster itself
+* deployed in the management cluster to autoscale a remote workload cluster
+* deployed in the workload cluster to autoscale the workload cluster itself
+
+Since EKS Anywhere follows a standard management/workload clusters setup where all the CAPI machine objects are defined in the management cluster, we choose to **deploy autoscalers in the management cluster as well — to watch and manage remote workload clusters, as default option**. In this way, we have a centralized place — management cluster, to deploy and manage all the autoscaler controllers for workload clusters.
```
maxdrib, line 124: This does not describe how EKS-A might support autoscaling of management clusters. Is that out of scope?

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+### EKS-A Curated Package
+
+We will build the Cluster Autoscaler as an EKS-A curated package with an opinionated package bundle with default values. Once built and included in the package list, the autoscaler installation should fit into the standard curated package workflow. 
+
+## User Experience
+
+There are two requirements to enable cluster autoscaling:
+
+1. annotate the CAPI MachineDeployment — configurable through EKS-A cluster spec
```
maxdrib, line 138: This would require changes to the providers' template-md.yaml files, right, to conditionally add some annotations?

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+## Testing
+
+Add E2E tests for the below cases:
+
+1. Create an autoscaling enabled management cluster
+    1. generate autoscaling package config
+    2. create a cluster with `AutoScalingConfiguration` in cluster spec and package config above
+    3. scale up workload applications to create pending pods
```
maxdrib, line 498: how can we simulate load to trigger a scaleup?

maxdrib (commented):
Great work! Looking forward to seeing this in action

designs/autoscaler.md
jiayiwang7: management cluster does need autoscaling. To me it's more like an implementation detail to talk abt mgmt vs workload clusters.  And for goals we wanna stay more high level. I have covered mgmt autoscaling in solution section.

designs/autoscaler.md
jiayiwang7: We don't have more config we wanna add to eks-a spec. We support all the Cluster Autoscaler config through eks-a curated pkg which exposes all the config values Helm chart supports. Or user can install autoscaler through upstream.

designs/autoscaler.md
jiayiwang7: we do support it. i will add a sentense to be more clear.

designs/autoscaler.md
jiayiwang7: yup. with the new apibuiler, it's provider agnostic so we don't need to add in provider code in future.

designs/autoscaler.md
jiayiwang7: brutal way u can have initially 1 worker node. then deploy an application w 20 replicas and watch node scaling up

designs/autoscaler.md
jiayiwang7: details are covered here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-scale-down-work
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders

some metric you can configure in CA https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+This document aims to solve this limitation and provides **autoscaling** **compatibility** in EKS Anywhere, so the user can either choose to install Cluster Autoscaler as an EKS-A curated package during or after cluster creation time, or install themselves from [upstream](https://github.com/kubernetes/autoscaler/releases) after a cluster is created.
+
+### Goals and Objectives
+
+As an EKS Anywhere user, I want to:
+
+* create clusters in a streamlined fashion with autoscaling capabilities
+* add autoscaling capabilities to existing EKS-A clusters
+* see my clusters to auto scale up and down worker nodes within the boundaries I limited
```
TerryHowe, line 19:
```suggestion
* have clusters automatically scale worker nodes up and down within the set boundaries
```

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+### Statement of Scope
+
+#### In Scope
+
+* Include Cluster Autoscaler as an [EKS-A curated package](https://anywhere.eks.amazonaws.com/docs/concepts/packages/).
+* Integrate Cluster Autoscaler to all CAPI providers since the feature is provider agnostic.
+* Integrate autoscaling configuration in cluster spec to make EKS-A cluster autoscaling-aware.
```
TerryHowe, line 27: Why would the autoscaling configuration be in the cluster spec?

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+* Generate an EKS-A opinionated Cluster Autoscaler curated package template with default configurations.
+* Install Cluster Autoscaler controller as a curated package during or after management/workload cluster creation.
+* Support upstream k8s Autoscaler installed/managed outside of EKS-A Curated Package.
```
TerryHowe, line 30: This would mean the helm chart and container images would need to be available in both private and public ECR repo. I personally like that idea, but may be extra work for the prod deployment. Maybe make that use case future scope?

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster. Specifically, it can:
+
+* increase the size of the cluster when there are pending pods due to insufficient resources.
+* decrease the size of the cluster when some nodes have low utilization for a significant amount of time.
+
+Just like any other application, Cluster Autoscaler can be installed as a `Deployment` in any Kubernetes cluster. In order to connect Cluster Autoscaler to Cluster API managed EKS Anywhere cluster, a user needs to configure Cluster Autoscaler to use `clusterapi` as the cloud provider, and provide the path to the kubeconfig for the target cluster to autoscale.
+
+Currently, it is impossible to have one autoscaler controller to watch multiple workload clusters with different kubeconfig at the same time, as there is no way currently to separate the pods between the workload clusters autoscaler is watching. The recommended method is to run multiple autoscalers in the management cluster with different options for how they discover the scalable resource sets.
+
```
TerryHowe, line 117: I don't know what customers want, but it seems like it would be way easier to choose one deployment. The easiest thing I think would be run the autoscaler on the cluster it is managing which would cover both management and workload clusters.

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+## Overview of Solution
+
+We provide a solution to add autoscaling configuration in the EKS Anywhere API, which marks the worker node groups to be Cluster Autoscaler compatible. EKS Anywhere user can then install Cluster Autoscaler in any EKS-A cluster with proper configuration, and expect node autoscaling to work for the associated resource group.
+
+1. Introduce an optional type `AutoScalingConfiguration` in the cluster and populate it into each worker node group configuration.
+2. If `AutoScalingConfiguration` is specified (not nil) for a worker node group, EKS-A converts its fields to CAPI autoscaler annotations and applies them to the underlying CAPI Machine Deployment resource. In this way, the specified worker node group will be discoverable by the Cluster Autoscaler controller once installed.
+3. Install Cluster Autoscaler either (A) as an EKS-A curated package, or (B) from upstream; with configuration that uses Cluster API as cloud provider, and auto-discovers node groups in the target cluster.
+
+### API Design
+
```
TerryHowe, line 53: I would expect these to be configurations in the package custom resource.

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+1. annotate the CAPI MachineDeployment — configurable through EKS-A cluster spec
+2. install Cluster Autoscaler controller in the cluster
+
+Although it is not required, we recommend that users install the autoscaler as EKS-A curated package through the EKS-A CLI for a more streamlined installation experience and better support. There are a few possible scenarios when enabling autoscaling. 
+
+**Enable Cluster Autoscaler as a Curated Package During Cluster Creation**
+
+First, define a cluster spec file with autoscaling config enabled for selected worker node group(s):
+
+```yaml
```
TerryHowe, line 147: I guess if the min/max was in the package configuration, it would also need this machine group name.

designs/autoscaler.md
jiayiwang7: One major reason is that we want to guarantee user to be able to install autoscaler from upstream, without curated package. Exposing the machinedeployment annotations to cluster spec so that user does not need to interact w CAPI objects directly and eks-a cluster controller won't revert changes.

designs/autoscaler.md
jiayiwang7: That's another reason I don't think autoscaler config should be part of package config. We need to map the worker node group and min/max together. It creates dependencies to eks-a main workflow/cluster spec. Plus the reason https://github.com/aws/eks-anywhere/pull/2630#discussion_r913893248. I feel we should have autoscaler config as part of the cluster spec.

designs/autoscaler.md
jiayiwang7: That is true. Happy to discuss if there is a better way to support upstream cluster autoscaler without changing our cluster spec. A couple of ways to make it less confusing:
1. add warning in cli log if min/max specified but autoscaler controller is not installed
2. document the whole autoscaler installation process
3. default to install autoscaler curated pkg when detecting min/max in cluster spec during cluster creation (tho this approach introduce another problem which is -- should we auto install curated pkg w/o `--install-packages` arg, since it can be a gated feature)

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+* deployed in the workload cluster to autoscale the workload cluster itself
+
+Since EKS Anywhere follows a standard management/workload clusters setup where all the CAPI machine objects are defined in the management cluster, we choose to **deploy autoscalers in the management cluster as well — to watch and manage management cluster, and all the remote workload clusters, as default option**. In this way, we have a centralized place — management cluster, to deploy and manage all the autoscaler controllers for workload clusters. 
+
+However, it is not uncommon that the management cluster size defined by the user is small, but the workload clusters are rather big, then it makes more sense to distribute the workload and run the autoscaler in each workload cluster to reduce resource usage in one management cluster. This configuration is also supported by us. Though this approach requires extra steps to create the management cluster’s `kubeconfig` in the workload cluster, comparing to the first approach where each workload cluster’s kubeconfig by default exists in the management cluster. Having management cluster kubeconfig in workload cluster also draws security concerns.
```
chrisdoherty4, line 126: Are we able to elaborate a little on the security concerns of external controllers given its a valid deployment model? It'll be useful for docs given we're saying its supported.

jiayiwang7: yup, updated

designs/autoscaler.md
```diff
@@ -0,0 +1,515 @@
...
+
+```yaml
+// cluster.yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: `${CLUSTER_NAME}`
+spec:
+  workerNodeGroupConfigurations:
+  - count: 2
```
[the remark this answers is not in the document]
jiayiwang7: my initial thought was to map those eks-a fields to capi machine deployment:
- `cluster.spec.workerNodeGroupConfigurations[0].count` -> `machinedeployment.spec.replicas`
- `cluster.spec.workerNodeGroupConfigurations[0].autoscalingConfiguration.minCount` -> `machinedeployment.metadata.annotations.cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size`
- `cluster.spec.workerNodeGroupConfigurations[0].autoscalingConfiguration.maxCount` -> `machinedeployment.metadata.annotations.cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size`

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+
+However, it is not uncommon that the management cluster size defined by the user is small, but the workload clusters are rather big, then it makes more sense to distribute the workload and run the autoscaler in each workload cluster to reduce resource usage in one management cluster. This configuration is also supported by us. Though this approach requires extra steps to create the management cluster’s kubeconfig in the workload cluster, comparing to the first approach where each workload cluster’s kubeconfig by default exists in the management cluster. Some limitations and concerns:
+
+  1. limiting user permission to the raw CAPI components: having management cluster kubeconfig in workload cluster means workload cluster user can provision and manage clusters that are not built in a conformant, reviewed or secured manner.
+  2. limiting user access to the underlying infrastructure: once non-admin workload cluster users have management cluster admin kubeconfig, which puts very high access credentials needed by CAPI. They then have high permission to the provider infrastructure (e.g vSphere: admin access to vCenter).
```
g-gaston, line 134: Is this true? I think we can use a different kubeconfig with less permissions
Probably something like
https://github.com/kubernetes/autoscaler/blob/042badcc37f49824fbd735d87431c02230d0eaff/cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml#L156-L172

Still something to take into consideration, since it gives access to modify other workload (or even the management) clusters.

This risk can be reduced with namespace separation of workload clusters. But that's not something that eksa supports right now.

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+  3. auditing and compliance of autoscaler: as the number of workload clusters arises, it becomes a mess of autoscaler deployments with various versions and configurations. Centralizing configuration makes sure compliance and configuration can be audited.
+  4. lack of visibility and management: it is hard to list and visualize all the autoscaler deployed in each workload cluster at the same time without user building its own monitoring tools. User needs to find credentials for each cluster and run kubectl separately to get and manage each deployment.
```
g-gaston, line 136: I would maybe add something about network connectivity. One benefit of separating management and workload clusters is that you can completely cut any connection from workload to management.

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+
+```sh
+$ eksctl anywhere create cluster -f workload-cluster.yaml --install-packages cas-package.yaml --kubeconfig mgmt-cluster.kubeconfig
+```
+
+This command goes through the following steps:
+
+1. create a new EKS-A cluster with proper autoscaling annotations on worker resources
+2. install curated package controller on the management cluster
+3. store the management cluster’s kubeconfig as a secret in the workload cluster (through curated package webhook, TBD)
```
g-gaston, line 260: Do we need to bake this into our product? I'm not sure we have to, it seems like extra work that can be avoided. Imo we shouldn't recommend this option at all, let alone build it.

If in the future there are enough users who want this, we could add it. But for the first iteration I think it's best to only provide as an option our preferred topology.

Also, if we go with this, we need to take into consideration the security implications of adding that capability to the curated packages controller. In order to create that kubeconfig, the controller needs access to read it, which gives it admin access. There is a big difference between just having access to create deployments (or whatever else) and having admin access.

designs/autoscaler.md
chrisdoherty4: I had a similar thought. It might be useful to think through an independent CRD and controller that targets worker node groups using label selectors, identifies associated objects and applies the annotations. 

I also thought about this in the inverse. I install cluster auto scaler because we're saying customers can do that as its supported. After getting it working having applied annotations, scaling stops working due to a different unrelated change - something's seems a miss here with the experience. The above solution would at least identify missing annotations and apply them again re-enabling auto scaling.

---

Typing out the above got me thinking, we're modelling an abstract API, yet we're telling customers to interact with the concrete implementation of auto scaling. Is it me or does that seem odd? Perhaps we shouldn't be overwriting the annotations like we are? Has that been explored?

jiayiwang7: The goal of the API design is to fit into the current product decision: we want to support both upstream autoscaler and autoscaler as an eks-a curated package. Just like any other package (load balancer, ingress controller, harbor), user can choose between installing it from upstream, or through eks-a gated, scanned curated package. To support upstream, we will need to expose the underlying CAPI `machinedeployment` min/max annotation to user so that the eks-a cluster controller reconciles and always applies it.

Now to expose the min/max annotation, there are two ways:
1. include it in cluster spec, as part of the worker node group config (chosen in this doc)
2. define an eks-a autoscaling CRD, map min/max to the worker node group. e.g.

```go
kind: Autoscaler
metadata:
  name: cluster-1-autoscaler
spec:
  minCount: 1
  maxCount: 3
  workerNodeGroupRef:
    kind: VSphereMachineConfig
    name: md-0
```

The one benefit of this approach is to isolate cluster autoscaler config from the cluster spec. But it's more complicated and less straightforward API wise. One autoscaler CR maps to only one worker node group. Way more complicated implementation for both CLI and controller. And this still does not solve the confusion mentioned above. In the future even if we integrate with Karpenter, this CRD will not be used at all since Karpenter does not scale through node group and it has its own CRDs user can use.

So in general, the design API is the most straightforward, easy to implement solution. As for confusion, EKS follows similar pattern where user needs to create a cluster with autoscaling capability first:`eksctl create cluster --asg-access`, which will set up the EC2 ASG group, and the IAM permissions. Then manually install cluster autoscaler themselves from upstream to enable autoscaling function.

designs/autoscaler.md
jiayiwang7: Imo we should, because there is no way for us to prevent user from installing autoscaler in workload cluster themselves, even through curated package. To me it is a choice made based on different use cases: resource allocation, infrastructure and deployment strategy, etc. We tell the user pros and cons for each approach, they determine which option they go with. Plus I believe we will support creating cluster with installing pkg on the cluster itself first. Treating autoscaler as a package to me means user has the flexibility to choose how and where to install this package. We can have opinionated suggestion, but also provide instructions for each option that is durable.

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+* Integrate Cluster Autoscaler to all CAPI providers since the feature is provider agnostic.
+* Integrate autoscaling configuration in cluster spec to make EKS-A cluster autoscaling-aware.
+* Generate an EKS-A opinionated Cluster Autoscaler curated package template with default configurations.
+* Install Cluster Autoscaler controller as a curated package during or after management/workload cluster creation.
+* Support upstream k8s Autoscaler installed/managed outside of EKS-A Curated Package.
+* Update the autoscaling configuration to set up the min/max number of nodes in each worker node group.
+
+#### Out of Scope
+
+* Compare cluster autoscaler products (e.g. [Karpenter](https://karpenter.sh/))
```
vivek-koppuru, line 35: Did we do some analysis on this before to determine to go with this native approach that we want to call out?

designs/autoscaler.md
g-gaston: I see what you are saying, however, there is a difference between providing instructions to do something and baking everything into the code.

The more functionality we add to the codebase, the more we need to maintain. And that's a fee we need to pay for life. Hence my preference to start simple and add more features as users request them.

designs/autoscaler.md
vivek-koppuru: If we use a CRD, we can reuse the CRD for different worker node groups right?

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+### API Design
+
+```go
+type WorkerNodeGroupConfiguration struct {
+    // Count defines the number of desired worker nodes.
+    Count *int `json:"count,omitempty"`
+    
+    // AutoScalingConfiguration defines the auto scaling configuration
+    // +optional
+    AutoScalingConfiguration *AutoScalingConfiguration `json:"autoscalingConfiguration,omitempty"`
```
vivek-koppuru, line 60: So this is only used if the user actually installs the AutoScaler?

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+  - autoscalingConfiguration:
+    - minCount: 1
+    - maxCount: 5
+    machineGroupRef:
+      kind: VSphereMachineConfig
+      name: worker-machine-a
+    name: md-0
+  - count: 1
+    autoscalingConfiguration:
+    - minCount: 1
```
vivek-koppuru, line 174: This isn't an array right?

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+    // AutoScalingConfiguration defines the auto scaling configuration
+    // +optional
+    AutoScalingConfiguration *AutoScalingConfiguration `json:"autoscalingConfiguration,omitempty"`
+    ...
+}
+
+type AutoScalingConfiguration struct {
+    // MinCount defines the minimum number of nodes for the associated resource group.
+    // +optional
+    MinCount int `json:"minCount,omitempty"`
```
ewollesen, line 67: Just curious, but on line 98, it seem that the cluster autoscaler already uses the terms "min-size" and "max-size". Would it be worthwhile to maintain consistency and use minSize / maxSize in our code?

designs/autoscaler.md
ewollesen: I agree, at least at first thought. Packages shouldn't be reliant on the cluster controller to perform their functions, at least not in terms of configuration. It wouldn't support the concept of packages on a non-eks-a cluster, for example, or it would make for multiple avenues of obtaining that information would have to be available.

designs/autoscaler.md
```diff
@@ -0,0 +1,526 @@
...
+    MaxCount int `json:"maxCount,omitempty"`
+}
+```
+
+### `AutoScalingConfiguration` Defaults and Validation
+
+We consider the following semantics when setting initial Count that is passed down to MachineDeployment:
+
+* If `Count` is specified, regardless the existence of `AutoScalingconfiguration`, `Count` is used as the initial count.
+* If `AutoScalingconfiguration` is specified, then `Count` is optional. When `Count` is not specified: `MinCount` is used as the initial count.
```
ewollesen, line 80: Having this documented is great. I can see it helping give us a defined path to take in the event that someone configures for an autoscaler, but doesn't have paid curated packages support. This way you've designed in a clearly defined and workable fallback strategy. 👍🏻

designs/autoscaler.md
ewollesen: If the cluster controller wants to be sole point of access for anything in the domain of cluster api, I think it will need to be flexible in allowing curated packages access to at least parts of that functionality. It will be a major detriment to curated packages independent development and release cycles if they require custom coding in the cluster controller for every package that needs to touch something in the cluster controller's domain.

designs/autoscaler.md
jiayiwang7: internally, out of scope for this doc

designs/autoscaler.md
jiayiwang7: eks-a spec is the only one we expose to the user, not CAPI. It is more important to stay in consistent w eks-a field rather than underlying CAPI field. Similar to `Count`, underlying it's named `replicas` in CAPI.

TerryHowe (approved):

PR #2646, opened 2022-07-01
Add C4 diagrams for cluster controller

g-gaston:


designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,59 @@
...
+    Rel(cilimHandler, remoteClient, "Reads/writes Cilium DS/deployment and CRDs")
+
+    Rel(cilimHandler, helm, "Generates Cilium resources in yaml")
+}
+
+Container(capiControllers, "CAPI controllers", "Go", "Reoncile workload clusters")
+Rel(snowHandler, capiControllers, "Creates/updates snow clusters", $tags="indirect relantionship")
+Rel(vsphereHandler, capiControllers, "Creates/updates vSphere clusters", $tags="indirect relantionship")
+Rel(capiControllers, cilimHandler, "Marks control plane as ready", $tags="indirect relantionship")
+Rel(capiControllers, workloadKubeApiServer, "Creates", $tags="indirect relantionship")
```
chrisdoherty4, line 46: For the purpose of this diagram, I think you can drop arrows between components/containers _outside_ the boundary. These arrows tend to add clutter and detract from communicating the relationship of the components in the subject container.

(not to be confused with a user of the system and how they interact with it).

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,59 @@
...
+
+Rel(user, kubeApiServer, "Writes EKS-A API objects", "JSON/HTTPS")
+
+Rel(clusteController, kubeApiServer, "Subscribes to API events", "JSON/HTTPS")
```
chrisdoherty4, line 50: On component diagrams there's no need to depict higher level relationships like container to container. The focus should be on _what component_ communicates with an external container/system.

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,59 @@
...
+    AddElementTag("port", $fontColor="#ffffbf", $borderColor="#ffffbf")
+    AddElementTag("busines logic", $fontColor="#01ffff", $borderColor="#01ffff")
+
+    Component(clusterReconciler, "Cluster Reconciler", "K8s controller-runtime Reconciler", "Validates input and orchestrates Cluster reconcile requests", $tags="busines logic")
+    Component(vsphereHandler, "vSphere Cluster Reconciler", "ProviderClusterReconciler", "Reconciles a vSphere Cluster to the desired state", $tags="busines logic")
+    Component(snowHandler, "Snow Cluster Reconciler", "ProviderClusterReconciler", "Reconciles a Snow Cluster to the desired state", $tags="busines logic")
+
+    Component(cilimHandler, "Cilium Reconciler", "", "Reonciles the Cilium CNI installation to the desired state", $tags="busines logic")
+
+    Component(client, "Kubernetes API client", "controller-runtime client.Client", "Kubernetes client provided by controller-runtime", $tags="port")
```
chrisdoherty4, line 23: I'm thinking we can drop these components. Looking at this now, it feels like it detracts from the core component set (where you'd find the business rules). Instead we can depict a relationship between the components and the KupAPI server (the fact it happens via a client is probably an even lower level diagram).

designs/images/controller-manager-deployment.puml
```diff
@@ -0,0 +1,41 @@
...
+        Deployment_Node(controllerManager, "Controller Manager", "Binary"){
+            Container(clusterController, "Cluster controller", "controller-runtime Controller", "Reconciles EKS-A Clusters")
+            Container(vsphereDatacenterController, "VSphereDatacenterConfig controller", "controller-runtime Controller", "Runs validations for VSphereDatacenterConfig that require external API calls (eg. vSphere endpoint)")
+            Container(snowMachineConfigController, "SnowMachineConfig controller", "controller-runtime Controller", "Runs validations for SnowMAchineConfig that require external API calls (eg. Snow endpoint)")
+
+            Container(clusterMutation, "Cluster mutation webhook", "K8s mutation webhook", "Sets defaults in Cluster objects before they are validated")
+            Container(clusterValidation, "Cluster validation webhook", "K8s validation webhook", "Validates Cluster objects before they are accepted by the kube api server")
+        }
+
+        Container(govc, "govc", "Binary")
```
chrisdoherty4, line 19: I don't think we should care about clients on a deployment diagram. They possibly raise more questions than provide answers as its unclear what they're doing given they aren't 'deployable' in their own right.

designs/images/snowmachineconfig-controller-container-components.puml
```diff
@@ -0,0 +1,24 @@
...
+
+    Rel(snowMachineConfigReconciler, awsCli, "Uses")
+}
+
+Component_Ext(snowAPI, "Snow API endpoint", "AWS?")
+Rel(awsCli, snowAPI, "Makes API calls", "JSON/HTTPS ??")
+
+Rel(user, kubeApiServer, "Writes EKS-A API objects", "JSON/HTTPS")
+
+Rel(snowMachineConfigController, kubeApiServer, "Subscribes to API events", "JSON/HTTPS")
```
chrisdoherty4, line 21: Similar to a previous comment, arrows from the subject boundary should really be coming from a component within the boundary so we understand _what performs the communication_.

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,59 @@
...
+
+Rel(user, kubeApiServer, "Writes EKS-A API objects", "JSON/HTTPS")
+
+Rel(clusteController, kubeApiServer, "Subscribes to API events", "JSON/HTTPS")
+
+Rel(client, kubeApiServer, "Reads/Writes Objects", "JSON/HTTPS")
+
+Rel(remoteClient, workloadKubeApiServer, "Reads/Writes Objects", "JSON/HTTPS")
+
+Rel(helm, ociRegistry, "Reads Helm charts", "JSON/HTTPS??")
```
chrisdoherty4, line 56: Question marks.

designs/images/controller-manager-deployment.puml
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+Container_Ext(vCenter, "vCenter Server", "VMware", "vSphere Automation API")
+Rel(govc, vCenter, "Makes API calls", "JSON/HTTPS ??")
+
+Container_Ext(ociRegistry, "OCI registry", "Public ECR or other", "OCI compliant registry to store container images and Helm charts")
+Rel(helm, ociRegistry, "Reads Helm charts", "JSON/HTTPS??")
+
+Container_Ext(snowAPI, "Snow API endpoint", "AWS?")
```
chrisdoherty4, line 30: Question mark.

designs/images/controller-manager-deployment.puml
```diff
@@ -0,0 +1,41 @@
...
+        Container(govc, "govc", "Binary")
+        Container(helm, "helm", "Binary")
+    }
+}
+
+Container_Ext(vCenter, "vCenter Server", "VMware", "vSphere Automation API")
+Rel(govc, vCenter, "Makes API calls", "JSON/HTTPS ??")
+
+Container_Ext(ociRegistry, "OCI registry", "Public ECR or other", "OCI compliant registry to store container images and Helm charts")
+Rel(helm, ociRegistry, "Reads Helm charts", "JSON/HTTPS??")
```
chrisdoherty4, line 28: Question marks.

chrisdoherty4 (changes requested):
Nice

designs/Makefile
```diff
@@ -0,0 +1,31 @@
...
+		$(PLANTUML_IMG_TAGGED) \
+		-v /workdir/$(shell echo '$^' | sed -e 's,.*designs/,,g')
+
+.PHONY: plantuml-image
+plantuml-image:
+	docker build --build-arg PLANTUML_VERSION=$(PLANTUML_VERSION) -t $(PLANTUML_IMG_TAGGED) .
+
+.PHONY: push-plantuml-image
+push-plantuml-image: plantuml-image
+	aws ecr-public create-repository --repository-name $(IMAGE_NAME) --region us-east-1 || true
```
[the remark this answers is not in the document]
g-gaston: Yeah, but why?

designs/images/cluster-controller-container-components.puml
g-gaston: Fair

designs/images/cluster-controller-container-components.puml
g-gaston: Yeah, that's fair.

I felt similarly, however, I wanted to represent this relationship, I think it's useful to understand the input of events to the components inside the container.

I saw 3 options:
* What I have here
* Start the relationship in the cluster reconciler. Which is directionally correct but technically incorrect, since the subscription is managed by the controller itself (you define the subscription when registering the controller with the controller manager).
* Add another component, like a "generic" `controller-runtime`. This would perform the subscription and send the reconcile request to the cluster reconciler. This seemed going a bit too deep into how the controller-runtime framework works. 

What do you think?

designs/images/snowmachineconfig-controller-container-components.puml
g-gaston: Same answer. I will address this with the same solution we agreed for the other controller

designs/images/cluster-controller-container-components.puml
g-gaston: I see what you are saying, however I don't think they detract.

In fact I believe they help understand where the boundaries of the "business rules" components are. These are what decouple the business rules from the protocols and "languages" of the external containers.

designs/Makefile
```diff
@@ -0,0 +1,31 @@
+ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+SOURCES := $(shell find ${ROOT_DIR} -name \*.puml)
+DIAGRAMS := $(SOURCES:%.puml=%.png)
+
+REGISTRY ?= public.ecr.aws/i0f3w2d9
```
vivek-koppuru, line 6: So is this a shared registry just for this? Should we create an official one to use from build tooling?

designs/Makefile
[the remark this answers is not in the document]
vivek-koppuru: If this is a command to push an image based on the user's choice, then they would the docker run command with the image overridden? I am assuming the default is to not run this command in favor of using the default for the registry set above.

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,58 @@
...
+    Component(snowHandler, "Snow Cluster Reconciler", "ProviderClusterReconciler", "Reconciles a Snow Cluster to the desired state", $tags="busines logic")
+
+    Component(cilimHandler, "Cilium Reconciler", "", "Reonciles the Cilium CNI installation to the desired state", $tags="busines logic")
+
+    Component(client, "Kubernetes API client", "controller-runtime client.Client", "Kubernetes client provided by controller-runtime", $tags="port")
+    Component(remoteClient, "Remote Kubernetes API client", "controller-runtime client.Client", "Kubernetes client for external cluster authenticated with kubeconfig", $tags="port")
+
+    Component(helm, "Helm client", "Helm cli binary [Go]", "Client to interact with Helm charts", $tags="port")
+
+    Rel(vsphereHandler, cilimHandler, "Delegates")
```
vivek-koppuru, line 28: Why does the individual providers delegate to cilium and not the cluster itself? Can't cluster wait for the provider to create the clusters and then delegate to the cilium controller to reconcile?

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,58 @@
...
+title Component diagram for EKS-A Cluster controller
+
+AddRelTag("indirect relantionship", $textColor="orange", $lineColor="orange", $lineStyle = DashedLine())
+
+Person(user, "EKS-A user", "Manages EKS-A clusters")
+Container(kubeApiServer, "Kubernetes API server", "Go", "Provides read/write access to the kubernetes API objects stored in the etcd server and allows to subscribe to events when they change")
+Container_Ext(workloadKubeApiServer, "Workload Cluster Kubernetes API server", "Go", "Provides read/write access to the kubernetes API objects stored in the etcd server")
+Container_Ext(ociRegistry, "OCI registry", "Public ECR or other", "OCI compliant registry to store container images and Helm charts")
+
+Container_Boundary(clusteController, "Cluster Controller") {
```
vivek-koppuru, line 13:
```suggestion
Container_Boundary(clusterController, "Cluster Controller") {
```

designs/images/cluster-controller-container-components.puml
```diff
@@ -0,0 +1,58 @@
+@startuml cluster-controller-container-components
+!include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Component.puml
+
+title Component diagram for EKS-A Cluster controller
+
+AddRelTag("indirect relantionship", $textColor="orange", $lineColor="orange", $lineStyle = DashedLine())
```
vivek-koppuru, line 6:
```suggestion
AddRelTag("indirect relationship", $textColor="orange", $lineColor="orange", $lineStyle = DashedLine())
```

designs/images/cluster_reconcile.puml
```diff
@@ -1,4 +1,4 @@
-@startuml
+@startuml cluster_reconcile
```
vivek-koppuru, line 1: What does this show to create the above png diff?

designs/images/snowmachineconfig-controller-container-components.puml
```diff
@@ -0,0 +1,24 @@
...
+
+Container_Boundary(snowMachineConfigController, "Snow MachineConfig Controller") {
+    Component(snowMachineConfigReconciler, "SnowMachineConfig Reconciler", "K8s controller-runtime Reconciler", "Runs validations for SnowMAchineConfig that require external API calls (eg. Snow endpoint)")
+    Component(awsCli, "AWS API client", "awscli - Python")
+
+    Rel(snowMachineConfigReconciler, awsCli, "Uses")
+}
+
+Component_Ext(snowAPI, "Snow API endpoint")
+Rel(awsCli, snowAPI, "Makes API calls", "JSON/HTTPS")
```
vivek-koppuru, line 17: Just curious, only the cli works here and not the go sdk? Probably because it's not written for snow?

designs/Makefile
g-gaston: This is my own account
I don't think we should be pushing manually to a build account and I didn't think it was worth to setup post submit infrastructure to automate this, given that this image will rarely change.

designs/images/cluster-controller-container-components.puml
g-gaston: Because the provider specific reconcilers have control over the flow of operations

We learned that providers need to hook into cluster workflows at different moments. When having a generic flow that calls into providers, you can end up with providers implementing operations they don't need. This just inverts the responsibility and gives each provider control over the high level orchestration.

designs/images/snowmachineconfig-controller-container-components.puml
g-gaston: awsCli, should have been `awsClient`. Updated and clarified on the component itself

designs/images/cluster_reconcile.puml
g-gaston: This defined the name of the image file that plantuml generates (`cluster_reconcile.png`)
It should be required for all diagrams, this one was missing it

designs/Makefile
vivek-koppuru: True but I think the main benefit is that we can have an official place that this image is stored rather than your account, otherwise it will end up becoming hard to figure out what account this is from when people want to actually pull the image.

g-gaston: Does it matter what account this is in to pull the image?
Imagine this image was already available in docker.io, would that make the pull process more difficult?

Building automation for this image seems complete overkill imo. But I think the way this works right now allows to do that without mayor changes here. Basically just change the registry url.

designs/images/cluster-controller-container-components.puml
g-gaston: With the way things work right now, yes. But we made the same assumption on the cli and i ended up not being accurate. This just tries to avoid that.

Plus it's always easier to start with some minimal "duplication" (even though in this case I wouldn't even call it that) and then find a better common abstraction with time that try to abstract it right way and then have to split it later on.

designs/images/cluster-controller-container-components.puml
chrisdoherty4: I think the relationship belongs on a container diagram, not a component diagram :).

As is, this denotes a component in the container talks to kube API but doesn't tell us what component and that defeats the point of diving a level deeper than the container.

chrisdoherty4 (approved):

Issue #2649, opened 2022-07-04
docs: links with access denied 


bryanasdev000:
**What happened**:

Extra links at https://anywhere.eks.amazonaws.com/docs/getting-started/local-environment/ like OIDC, etcd, proxy and gitops leads to a standard Access Denied like https://anywhere.eks.amazonaws.com/docs/reference/clusterspec/oidc/ :

![Captura de tela_2022-07-04_03-32-37](https://user-images.githubusercontent.com/53131727/177095211-05964874-3db8-4f1c-b05b-db6156e76f0d.png)

**What you expected to happen**:

Access the etcd docs.

**How to reproduce it (as minimally and precisely as possible)**:

1. Go to https://anywhere.eks.amazonaws.com/docs/getting-started/local-environment/ 
2. Click on etcd link above step 2.

**Anything else we need to know?**:

**Environment**: N/A
- EKS Anywhere Release: N/A
- EKS Distro Release: N/A

PR #2650, opened 2022-07-04
Fix local environment doc links

TerryHowe:
Closes: https://github.com/aws/eks-anywhere/issues/2649


chrisnegus (approved):
Thanks for catching those Terry!

PR #2654, opened 2022-07-05
Controller code reorg

g-gaston:
### Proposal
* Un-nest `controllers/controllers`. In order to avoid having a main.go in the root folder (which was one the goals of the nested controllers folder), we will move `main.go` to a `manager` folder.
  * Move everything except the actual controllers and their tests to `pkg`. This increases the coherence with the rest of the codebase and avoids having a “separate” project under the same module.
* All the packages that are exclusive to controllers functionality will go nested under `pkg/controller`. This gives contextual information about the package and helps with name collision.
    * For the provider cluster reconcilers (currently in `controllers/controllers/cluster`), move them to their respective `pkg/providers/[provider name]/reconciler`.
    * Common reconciler functionality shared among providers will go in `pkg/controller/clusters`.
    * Package `controllers/controllers/reconciler` should be split into:
      * `pkg/unstructuredutil`: util methods to convert yaml to `unstructured.Unstructured`.
      * `pkg/controller/clientutil`: util methods to convert yaml to `client.Object`.
      * `pkg/controller`: just the `Result` struct.
      * `pkg/controller/serverside`: server side apply “reconcile” methods.
  * Leave `resource`` package under `controllers`. This package will be removed once we promote the new controller to the default one, so we can avoid the extra work and leave it untouched.

### Aprox package structure

```
controllers
|- resource
|- cluster_controller.go
|- cluster_controller_legacy.go
|- cluster_controller_test.go
|- vsphere_datacenter_controller.go

manager
|- main.go
|- buildspec.yaml
|- README.md
|- Tiltfile
|- docker
   |- Dockerfile

pkg
|- controller
   |- clusters
   |- clientutil
   |- serverside
   |- result.go
|- providers
   |- vsphere
      |- reconciler
   |- snow
      |- reconciler
   |- docker
      |- reconciler
|- unstructuredutil
```


pkg/unstructuredutil/yaml.go
```diff
@@ -0,0 +1,11 @@
+package unstructuredutil
```
jiayiwang7, line 1: why not having it part of the `pkg/controller` ?

manager/Tiltfile
```diff
@@ -8,7 +8,8 @@ default_registry(settings.get('default_registry'))
 IMG = 'cluster-controller:latest'
 PROJECT_ROOT='./../'
 CONTROLLERS_FOLDER = PROJECT_ROOT + 'controllers'
-CONTROLLERS_BIN_FOLDER = CONTROLLERS_FOLDER + '/bin'
+MANAGER_FOLDER = PROJECT_ROOT + 'manager'
+MANAGER_BIN_FOLDER = MANAGER_FOLDER + '/bin'
 BIN_FOLDER = PROJECT_ROOT + 'bin'
```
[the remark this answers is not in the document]
g-gaston: I haven't touched that variable or any of its usages
I just replaced `CONTROLLERS_BIN_FOLDER` by `MANAGER_BIN_FOLDER`

pkg/unstructuredutil/yaml.go
g-gaston: Because `unstructured.Unstructured` is not something exclusive to controllers or the `controller-runtime`, so it can potentially be used in other places of our codebase

manager/Tiltfile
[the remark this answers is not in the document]
jiayiwang7: understood, but i dont think `BIN_FOLDER` is used anywhere so can we remove it since u r close to it :)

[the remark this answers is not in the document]
g-gaston: I don't think we should mix that with a PR that only moves code around, specially one as big as this one. It makes tracking changes very difficult.

https://github.com/aws/eks-anywhere/pull/2658

pkg/controller/clusters/provider.go
```diff
@@ -0,0 +1,30 @@
...
+	providerClient client.Client
+}
+
+func NewProviderClusterReconciler(client client.Client) *ProviderClusterReconciler {
+	return &ProviderClusterReconciler{
+		providerClient: client,
+	}
+}
+
+func (p *ProviderClusterReconciler) GetEksdRelease(ctx context.Context, name, namespace string) (*eksdv1alpha1.Release, error) {
```
vivek-koppuru, line 21: How is this a provider specific method? Maybe I am a bit confused by the interface.

pkg/controller/handlerutil/childobjects_test.go
```diff
@@ -11,8 +11,8 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	"github.com/aws/eks-anywhere/controllers/controllers/utils/handlerutil"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/controller/handlerutil"
```
vivek-koppuru, line 15: Can this be `handlers` also?

pkg/controller/serverside/reconcile.go
```diff
@@ -1,16 +1,18 @@
-package reconciler
+package serverside
```
vivek-koppuru, line 1: Are these controller reconcilers always server side operations? Do we want to make this distinction because these methods aren't always called and used in special cases?

jiayiwang7 (approved):

controllers/cluster_controller.go
```diff
@@ -43,14 +45,18 @@ type ClusterReconciler struct {
 	validator               *vsphere.Validator
 	defaulter               *vsphere.Defaulter
 	tracker                 *remote.ClusterCacheTracker
-	buildProviderReconciler BuildProviderReconciler
+	buildProviderReconciler ProviderReconcilerBuilder
+}
+
+type ProviderClusterReconciler interface {
```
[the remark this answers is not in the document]
g-gaston: I see what you are saying. However, these are not reconciling a datacenter config. They are reconciling a cluster, but following a different logic depending on the provider. If we had something like a `VSphereCluster` CRD, this would be its controller.

You make a good point, we don't have a resource to pair this reconcilers. But that extends to our API in general: we didn't design it with kubernetes controllers in mind (one controller per resource) but with the CLI and how the user interacts with config yaml file. But addressing that (which is doable) is a whole other topic.

I still think [Provider]ClusterReconciler is the right name for what these entities do. However, if we think we need to rename them, I would vote to continue that conversation outside of this PR. The goal here was to reorganize the code with a more sound package structure. We can add renaming refactors on top of it.

pkg/controller/clientutil/kubernetes.go
```diff
@@ -1,4 +1,4 @@
-package clients
+package clientutil
```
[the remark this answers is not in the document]
g-gaston: `clients` didn't feel right, we are not implementing multiple clients for controllers
This package includes util methods to work with the kubernetes client package provided by controller runtime, so `clientutils` felt more appropriate.

pkg/controller/clusters/provider.go
g-gaston: It is not. I'm not sure these method should be part of the `ProviderClusterReconciler`. But I'm just moving code around.

I think the idea was that `ProviderClusterReconciler` was gonna be embed by all provider cluster reconciler, so we could implement common functionality in it. We might want to revisit that idea as we start using this more and probably create less generic entities.

pkg/controller/serverside/reconcile.go
g-gaston: Sorry, I don't follow, what do you mean exactly by making a distinction?

In case it helps, this refers to server side apply provided by the kuber api server.

vivek-koppuru (approved):

PR #2656, opened 2022-07-05
Snow machine config mutation webhook for simple defaults

mitalipaygude:
*Issue #, if available:*
#2310 

*Description of changes:*
Setting the defaults values for `InstanceType` and `PhysicalNetworkConnector` for Snow Machine Config.

*Testing (if applicable):*
- Create docker cluster with tilt and attach a snow machine config on the cluster without defaults values set. Check once attached if the values are set.
- Unit tests

### DIFF
--- a/pkg/api/v1alpha1/snowmachineconfig_webhook.go
+++ b/pkg/api/v1alpha1/snowmachineconfig_webhook.go
@@ -1,0 +1,40 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


pkg/api/v1alpha1/snowmachineconfig_webhook.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+//+kubebuilder:webhook:path=/mutate-anywhere-eks-amazonaws-com-v1alpha1-snowmachineconfig,mutating=true,failurePolicy=fail,sideEffects=None,groups=anywhere.eks.amazonaws.com,resources=snowmachineconfigs,verbs=create;update,versions=v1alpha1,name=mutation.snowmachineconfig.anywhere.amazonaws.com,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Defaulter = &SnowMachineConfig{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *SnowMachineConfig) Default() {
+	snowmachineconfiglog.Info("default", "name", r.Name)
+	snowmachineconfiglog.Info("Setting up Snow Machine Config defaults")
```
g-gaston, line 39: Do we need two log lines or could we combine them?

g-gaston (approved):

pkg/api/v1alpha1/snowmachineconfig_webhook.go
mitalipaygude: First is the default one that displays the resource name, so added an extra log saying we're seeing the defaults for that resource.

Issue #2464, opened 2022-06-21
Refactor CloudStack provider to pass secrets to CAPC instead of env var


maxdrib:
**What would you like to be added**:
Cloudstack provider should create secrets (like vsphere does) and apply those to the cluster instead of relying on CAPC to generate the secret(s) itself from an env var we pass to it.

**Why is this needed**:
As CAPC is moving to support multiple Cloudstack endpoints, it also needs to support multiple sets of Cloudstack credentials. With the current mechanism, the credentials are call captured as a single env var so if we want to change one, we have to change all of them via this env var. If we split them into independent secrets, however, then we can modify each credential independently of the others and reduce the blast radius.

rejoshed:
👍 

However, I want to add that I think the input from a user's perspective should be a yaml file of k8s secrets that's just verified by EKS-A.

Even if it's not desirable to use k8s secrets as the end user's input format, then a yaml file of secrets without the k8s boilerplate would still be better than INI.

We are looking to abandon the INI format in CAPC anyway, so I believe EKS-A should abandon it as well.

PR #2666, opened 2022-07-06
Generate k8s secrets from cloudstack cloud config

wongni:
*Issue #, if available:* #2406

*Description of changes:*
Instead of creating a new environment variable with base64 encoded INI file content, this PR creates k8s secrets based on the provided CloudStack cloud configuration to support multiple endpoints

*Testing (if applicable):*


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,6 +94,34 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
 }
 
 func (p *cloudstackProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster) error {
+	var contents bytes.Buffer
```
maxdrib, line 97: It feels kind of weird to pass in a buffer to be filled inside `createSecrets`. Why don't we just initialize the buffer inside createSecrets instead of passing it in?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -47,6 +49,9 @@ var defaultClusterConfigMD string
 //go:embed config/machine-health-check-template.yaml
 var mhcTemplate []byte
 
+//go:embed config/secrets.yaml
+var defaultSecretsTemplate string
```
maxdrib, line 53: it looks like this change will be backwards compatible, right? We're still passing the env var to CAPC in addition to the secret

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,6 +94,34 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
 }
 
 func (p *cloudstackProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster) error {
+	var contents bytes.Buffer
+	err := p.createSecrets(ctx, cluster, &contents)
+	if err != nil {
+		return err
+	}
+
+	err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, contents.Bytes())
```
maxdrib, line 103: We need more validation here - see https://github.com/aws/eks-anywhere/blob/main/designs/cloudstack-multiple-endpoints.md#cloudstack-credentials. Basically we want to make these secrets write-once so they can't be updated in this code. If the secret already exists on the cluster, we should not modify it in any way. Further, we should validate that the contents of the secret are the same as the contents of the secret locally generated to fail fast if anyone is trying to change them.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,6 +94,34 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+	if err != nil {
+		return fmt.Errorf("loading secrets object: %v", err)
+	}
+	return nil
+}
+
+func (p *cloudstackProvider) createSecrets(ctx context.Context, cluster *types.Cluster, contents *bytes.Buffer) error {
+	if err := p.providerKubectlClient.GetNamespace(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
+		if err := p.providerKubectlClient.CreateNamespace(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
```
maxdrib, line 112: can we avoid hardcoding the namespace here? Would it be possible to use the same namespace as the eksa cluster object?

pkg/providers/cloudstack/cloudstack.go
wongni: Good point

pkg/providers/cloudstack/cloudstack.go
wongni: Yes.

pkg/providers/cloudstack/cloudstack.go
wongni: I think I can use p.clusterConfig.Namespace instead of the constant

pkg/providers/cloudstack/cloudstack.go
vivek-koppuru: Who is allowed to modify these secrets then, and at what point?

wongni: I changed the caller hook calling UpdateSecrets function to PostBootstrapSetup. As far as I know PostBootstrapSetup is called after creating a bootstrap cluster on creating a management cluster. In this way, the secrets are created only once when creating a management cluster using the eks-a CLI. The operator of the management cluster can modify the secret directly using the kubeconfig of the cluster if they want to update the credentials of a certain ACS.

pkg/providers/cloudstack/cloudstack.go
g-gaston: `p.clusterConfig.Namespace` is not always going to be `EksaSystemNamespace`

g-gaston, line 112: I think I have seen this code in multiple places of the code base. Can just move it to somewhere common? If having this namespace has become a prerequisite to our code, we should just make sure it's always created.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,6 +94,34 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+	}
+
+	err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, contents.Bytes())
+	if err != nil {
+		return fmt.Errorf("loading secrets object: %v", err)
+	}
+	return nil
+}
+
+func (p *cloudstackProvider) createSecrets(ctx context.Context, cluster *types.Cluster, contents *bytes.Buffer) error {
```
g-gaston, line 110: Interesting, why pass a buffer as opposed to just return `[]byte`?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,56 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+		return nil, err
+	}
+
+	secrets := [][]byte{}
+	for _, profile := range p.execConfig.Profiles {
+		secret := createSecret(profile)
+		if bytes, err := yaml.Marshal(secret); err == nil {
+			secrets = append(secrets, bytes)
+		} else {
+			return nil, fmt.Errorf("marshalling secret: %v: %v", secret, err)
```
maxdrib, line 117: I'm not sure if we want to log the secret, maybe just the error

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,56 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: corev1.SchemeGroupVersion.Version,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: constants.EksaSystemNamespace,
+			Name:      profile.Name,
+		},
+		Immutable: &immutable,
```
maxdrib, line 134: I don't think we want to set this attribute - users should be able to modify/rotate the values here, but that's impossible with this attribute set as far as I can tell

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -364,6 +417,7 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
 			return fmt.Errorf("unable to set %s: %v", eksaLicense, err)
 		}
 	}
+	p.execConfig = execConfig
```
maxdrib, line 420: what's the point of the temporary `execConfig` variable? Why not just use `p.execConfig` throughout?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,56 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
 }
 
 func (p *cloudstackProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster) error {
+	contents, err := p.createSecrets(ctx, cluster)
+	if err != nil {
+		return fmt.Errorf("creating secrets object: %v", err)
+	}
+
+	err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, contents)
```
maxdrib, line 99: could you add more validation here to make sure the secret does exist and that it aligns with the local environment's secret values?

pkg/providers/cloudstack/cloudstack.go
wongni: Fixed

pkg/providers/cloudstack/cloudstack.go
wongni: Removed Immutable: true from secret creation

pkg/providers/cloudstack/cloudstack.go
wongni: Fixed

pkg/providers/cloudstack/cloudstack.go
wongni: Added a step checking the secrets are aligned with the env vars in the upgrade cluster preflight checks. 
https://github.com/aws/eks-anywhere/pull/2666/commits/0e663e16eee6462009bb65b6b917e469b81c9a86

maxdrib:
Looks like this PR is also related to https://github.com/aws/eks-anywhere/issues/2464

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,54 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+	if err != nil {
+		return fmt.Errorf("applying secrets object: %v", err)
+	}
 	return nil
 }
 
+func (p *cloudstackProvider) createSecrets(ctx context.Context, cluster *types.Cluster) ([]byte, error) {
+	if err := p.providerKubectlClient.CreateNamespaceIfNotExists(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
```
g-gaston, line 107: could we move this to the create workflow? 🙏🏻

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,54 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+	}
+
+	err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, contents)
+	if err != nil {
+		return fmt.Errorf("applying secrets object: %v", err)
+	}
 	return nil
 }
 
+func (p *cloudstackProvider) createSecrets(ctx context.Context, cluster *types.Cluster) ([]byte, error) {
```
g-gaston, line 106: nit:
```suggestion
func (p *cloudstackProvider) generateSecrets(ctx context.Context, cluster *types.Cluster) ([]byte, error) {
```

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,54 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+		if bytes, err := yaml.Marshal(secret); err == nil {
+			secrets = append(secrets, bytes)
+		} else {
+			return nil, fmt.Errorf("marshalling secret for profile %s: %v", profile.Name, err)
+		}
+	}
+	return templater.AppendYamlResources(secrets...), nil
+}
+
+func createSecret(profile decoder.CloudStackProfileConfig) *corev1.Secret {
```
g-gaston, line 123: nit:
```suggestion
func generateSecret(profile decoder.CloudStackProfileConfig) *corev1.Secret {
```

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -184,7 +233,9 @@ func (p *cloudstackProvider) RunPostControlPlaneUpgrade(ctx context.Context, old
 
 type ProviderKubectlClient interface {
 	ApplyKubeSpecFromBytes(ctx context.Context, cluster *types.Cluster, data []byte) error
+	GetNamespace(ctx context.Context, kubeconfig string, namespace string) error
```
g-gaston, line 236: I think we can remove this now

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -367,6 +419,26 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
...
+		}
+		if secretDifferentFromProfile(secret, profile) {
+			return fmt.Errorf("profile '%s' is different from the secret", profile.Name)
+		}
+	}
+	return nil
+}
+
+func secretDifferentFromProfile(secret *corev1.Secret, profile decoder.CloudStackProfileConfig) bool {
+	return string(secret.Data["uri"]) != profile.ManagementUrl ||
```
g-gaston, line 436: Maybe I'm wrong, but aren't the values of data encoded in base64?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -367,6 +419,26 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
 	return nil
 }
 
+func (p *cloudstackProvider) validateSecretsUnchanged(ctx context.Context, cluster *types.Cluster) error {
+	for _, profile := range p.execConfig.Profiles {
+		secret, err := p.providerKubectlClient.GetSecret(ctx, profile.Name, executables.WithCluster(cluster), executables.WithNamespace(constants.EksaSystemNamespace))
```
g-gaston, line 424: I think this will fail when the secret doesn't exit, right?
But we don't want to fail in that case, we do allow for new secrets.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +91,54 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
 }
 
+func (p *cloudstackProvider) createSecrets(ctx context.Context, cluster *types.Cluster) ([]byte, error) {
+	if err := p.providerKubectlClient.CreateNamespaceIfNotExists(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
+		return nil, err
+	}
+
+	secrets := [][]byte{}
+	for _, profile := range p.execConfig.Profiles {
+		secret := createSecret(profile)
```
g-gaston, line 113: can we avoid adding the secret to the list if the secret already exists?

pkg/providers/cloudstack/cloudstack.go
wongni: Moved to CreateBootstrapCluster in bootstrapper.go and removed creating eksa-system namespace from providers code.

pkg/providers/cloudstack/cloudstack.go
wongni: Fixed

pkg/providers/cloudstack/cloudstack.go
wongni: Fixed

pkg/providers/cloudstack/cloudstack.go
wongni: Done

pkg/providers/cloudstack/cloudstack.go
wongni: When I tested this code the a plain text in byte array format was returned.

pkg/providers/cloudstack/cloudstack.go
wongni: Got it

pkg/providers/cloudstack/cloudstack.go
wongni: Yes

pkg/executables/kubectl.go
```diff
@@ -167,6 +167,13 @@ func (k *Kubectl) CreateNamespace(ctx context.Context, kubeconfig string, namesp
 	return nil
 }
 
+func (k *Kubectl) CreateNamespaceIfNotExists(ctx context.Context, kubeconfig string, namespace string) error {
```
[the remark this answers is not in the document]
wongni: How about `CreateNamespaceIfNotPresent`? I referred to `imagePullPolicy: IfNotPresent`.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -367,6 +419,30 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
 	return nil
 }
 
+func (p *cloudstackProvider) validateSecretsUnchanged(ctx context.Context, cluster *types.Cluster) error {
```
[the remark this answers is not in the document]
wongni: The current validator does not have access to kubectl client. It looks like it is responsible for validations via cmk. 
```
type Validator struct {
	cmk ProviderCmkClient
}
```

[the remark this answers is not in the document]
maxdrib: I feel like the validator should support all cloudstack provider validations, so I'd be inclined to add this functionality there and keep the cloudstack.go provider class as simple as we can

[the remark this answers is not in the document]
wongni: Moving the function to validator requires some level of refactoring because the current validator does not have access to kubectl client. I will add a TODO comment and let's revisit this refactoring later.

vivek-koppuru (approved):

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +92,61 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
 }
 
 func (p *cloudstackProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster) error {
+	if err := p.providerKubectlClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
```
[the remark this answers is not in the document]
wongni: When creating a bootstrap cluster it create the eksa-system namespace, however, when creating a workload cluster it does not create it. That's why I added back the creation logic to UpgradeSecrets function.

pkg/providers/vsphere/vsphere.go
```diff
@@ -508,6 +507,10 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
 }
 
 func (p *vsphereProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster) error {
+	if err := p.providerKubectlClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace); err != nil {
```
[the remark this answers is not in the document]
wongni: Please refer to https://github.com/aws/eks-anywhere/pull/2666#discussion_r921639989

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -87,9 +92,60 @@ func (p *cloudstackProvider) PostWorkloadInit(ctx context.Context, cluster *type
...
+		}
+	}
 	return nil
 }
 
+func (p *cloudstackProvider) generateSecrets(ctx context.Context, cluster *types.Cluster) ([]byte, error) {
+	secrets := [][]byte{}
+	for _, profile := range p.execConfig.Profiles {
+		_, err := p.providerKubectlClient.GetSecretFromNamespace(ctx, cluster.KubeconfigFile, profile.Name, constants.EksaSystemNamespace)
+		if err == nil {
```
g-gaston, line 116: should we check here the type of error to see if it's a not found error?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -367,6 +424,31 @@ func (p *cloudstackProvider) validateEnv(ctx context.Context) error {
...
+func (p *cloudstackProvider) validateSecretsUnchanged(ctx context.Context, cluster *types.Cluster) error {
+	for _, profile := range p.execConfig.Profiles {
+		secret, err := p.providerKubectlClient.GetSecretFromNamespace(ctx, cluster.KubeconfigFile, profile.Name, constants.EksaSystemNamespace)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// When the secret is not found we allow for new secrets
+				continue
+			}
+			return fmt.Errorf("getting secret for profile %s: %v", profile.Name, err)
+		}
```
g-gaston, line 437: nit:

```suggestion
        if apierrors.IsNotFound(err) {
		    // When the secret is not found we allow for new secrets
		    continue
	    }
			
		if err != nil {
			return fmt.Errorf("getting secret for profile %s: %v", profile.Name, err)
		}
```

Issue #2665, opened 2022-07-06
EKS-A upgrade triggers move before workload cluster's RollingUpgrade is complete and fails


maxdrib:
**What happened**:
While upgrading from an older EKS-A release on a management cluster with a single workload cluster, there is a step where the new EKS-A controller Deployment is applied to the cluster being upgraded, where a rolling upgrade starts on a workload cluster. A new eks-a controller comes up (unpaused) and applies the new CAPI/CAPC cluster spec, which has a modified value for EtcdadmCluster. The new resource removes the field `etcdadmBuiltin: true` and adds a new field for `etcdadmInstallCommands`.

As a result, the upgrade fails either on the move to the bootstrap cluster, or it times out waiting for ControlPlaneReady on the workload cluster after the move because the move happened in the middle of the workload cluster's upgrade.

**What you expected to happen**:
I think a RollingUpgrade is expected in this case, so my proposal would be that we should wait for all the clusters associated with a given management cluster to finish upgrading before proceeding with the clusterctl move

**How to reproduce it (as minimally and precisely as possible)**:
Create a Cloudstack management cluster with 3 etcd, 2 CP, 3 worker nodes with an old eks-a release (v0.8.3-dev). Then create a workload cluster with the same configuration using the management cluster. Proceed to upgrade the management cluster using the v0.10.0 release of EKS-A.

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: v0.10.0
- EKS Distro Release:

maxdrib:
Discussing with @vivek-koppuru, he suggested performing a check prior to executing the move where we call `kubectl get clusters.cluster.x-k8s.io` and then looking at the `Status` field to make sure they are in the `Ready` state before executing `clusterctl move`. For management clusters, loop over all the results.

PR #2668, opened 2022-07-06
Adding check before move to make sure workload clusters are ready

maxdrib:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/2665

*Description of changes:*
This PR adds a validation step prior to executing a clusterctl move to make sure the workload clusters are ready for a self-managed cluster. 

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -160,8 +160,10 @@ func (c *ClusterManager) MoveCAPI(ctx context.Context, from, to *types.Cluster, 
 	if err := c.waitForNodesReady(ctx, from, clusterName, labels, checkers...); err != nil {
 		return err
 	}
-	if err := c.waitForClustersReady(ctx, from, clusterName, labels, types.WithClusterReady()); err != nil {
-		return err
+	if clusterSpec.Cluster.IsSelfManaged() {
+		if err := c.waitForManagedClustersReady(ctx, from, types.WithClusterReady()); err != nil {
+			return err
+		}
 	}
 
 	err := c.clusterClient.MoveManagement(ctx, from, to)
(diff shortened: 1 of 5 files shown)


pkg/clustermanager/cluster_manager.go
```diff
@@ -799,6 +802,47 @@ func (c *ClusterManager) waitForNodesReady(ctx context.Context, managementCluste
...
+		logger.V(4).Info("Clusters ready", "total", totalClusters)
+		return nil
+	}
+
+	err := areClustersReady()
+	if err == nil {
+		return nil
+	}
+
+	timeout := time.Duration(totalClusters) * c.machineMaxWait
```
maxdrib, line 833: move off of using machineMaxWait here for some cluster max wait instead

pkg/clustermanager/cluster_manager.go
```diff
@@ -828,6 +872,30 @@ func (c *ClusterManager) countNodesReady(ctx context.Context, managementCluster
 	return ready, total, nil
 }
 
+func (c *ClusterManager) countClustersReady(ctx context.Context, managementCluster *types.Cluster, clusterName string, labels []string, checkers ...types.ClusterReadyChecker) (ready, total int, err error) {
```
wongni, line 875: clusterName and labels are not used

pkg/clustermanager/cluster_manager.go
vivek-koppuru: Is this a future action item? I would suggest deciding on something here rather than using machineMaxWait if possible.

wongni: Added durations for cluster ready check

pkg/clustermanager/cluster_manager.go
```diff
@@ -53,6 +56,9 @@ type ClusterManager struct {
 	machineMaxWait     time.Duration
 	machineBackoff     time.Duration
 	machinesMinWait    time.Duration
+	clusterMaxWait     time.Duration
+	clusterBackoff     time.Duration
+	clustersMinWait    time.Duration
```
vivek-koppuru, line 61: Is this supposed to be `clusterMinWait` instead?

PR #2671, opened 2022-07-07
Hardware availability validations for scale up/down and rolling upgrades

ptrivedi:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/workflows/upgrade.go
```diff
@@ -127,8 +133,8 @@ func (s *setupAndValidateTasks) validations(ctx context.Context, commandContext
 	return []validations.Validation{
 		func() *validations.ValidationResult {
 			return &validations.ValidationResult{
-				Name: fmt.Sprintf("%s Provider setup is valid", commandContext.Provider.Name()),
-				Err:  commandContext.Provider.SetupAndValidateUpgradeCluster(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec),
+				Name: fmt.Sprintf("%s is provider setup valid", commandContext.Provider.Name()),
```
abhinavmpandey08, line 136: what you have currently is going to output this on error which doesn't sound right
`Tinkerbell is provider setup valid`

Maybe we can do something like this so the output will be 
`Tinkerbell provider validation`
```suggestion
				Name: fmt.Sprintf("%s provider validation", commandContext.Provider.Name()),
```

ptrivedi: Yes, was confusing for me too. Apparently that is the 'name' of the validation. This was existing, but I agree, would be better to make it clearer.

abhinavmpandey08 (approved):
/approve

PR #2683, opened 2022-07-07
Add option to use native binaries in dep factory

g-gaston:
*Description of changes:*
TODO
*Testing (if applicable):*
Manual testing


pkg/dependencies/factory.go
```diff
@@ -190,23 +193,51 @@ func (f *Factory) WithExecutableMountDirs(mountDirs ...string) *Factory {
...
 
+func (f *Factory) WithLocalExecutables() *Factory {
+	f.executablesInDocker = false
+	return f
+}
+
+// UserExecutablesDockerClient forces a specific DockerClient to build
+// Executables as opposed to follow the normal building flow
+// This is only for testing
+func (f *Factory) UserExecutablesDockerClient(client executables.DockerClient) *Factory {
```
vivek-koppuru, line 204: Is it supposed to be this?
```suggestion
func (f *Factory) UseExecutablesDockerClient(client executables.DockerClient) *Factory {
```

vivek-koppuru: If this is only used for testing, it might be clearer to mention in the method name that it is for Testing. But I would understand not doing that if we want to extend it to be able to pass it in other situations

pkg/dependencies/factory.go
```diff
@@ -104,8 +104,10 @@ func ForSpec(ctx context.Context, clusterSpec *cluster.Spec) *Factory {
 }
 
 type Factory struct {
-	executableBuilder        *executables.ExecutableBuilder
+	executableBuilder        *executables.ExecutablesBuilder
 	executablesImage         string
+	executablesInDocker      bool
```
vivek-koppuru, line 109: If we have a `WithDocker`, why not call this something like `containerExecutables`, and then if we have `WithPodman` in the future, it would use the same logic here? Just want to see if that is possible vs calling everything `executableInDocker` or `InitInDocker` or something, but I understand that it would be more refactoring and something we can pick up later when we actually support the different ones.

g-gaston: `containerExecutables` seems a bit ambiguous. If you prefer `executablesInContainer`, I'm ok with that.

However, as you are saying, if we want to support more container runtimes/clients, there would more stuff we would need to change.

pkg/executables/builder.go
```diff
@@ -77,67 +86,49 @@ func BuildDockerExecutable() *Docker {
...
-		return true
+		return false
 	}
-	return false
+	return true
 }
 
-func NewExecutableBuilder(ctx context.Context, image string, mountDirs ...string) (*ExecutableBuilder, Closer, error) {
-	currentDir, err := os.Getwd()
+// InitInDockerExecutablesBuilder builds an inits a default executables builder
```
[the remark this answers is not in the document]
g-gaston: Curious why? this method does both build the object and calls init on it

pkg/dependencies/factory.go
```diff
@@ -104,8 +104,10 @@ func ForSpec(ctx context.Context, clusterSpec *cluster.Spec) *Factory {
 }
 
 type Factory struct {
-	executableBuilder        *executables.ExecutableBuilder
+	executableBuilder        *executables.ExecutablesBuilder
```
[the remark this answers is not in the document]
g-gaston: I'm not sure having nested fields helps with readability, at least to me. But if you find it easier to read, I'll change it.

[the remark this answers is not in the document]
g-gaston: Done

jiayiwang7 (approved):

vivek-koppuru (approved):

PR #2692, opened 2022-07-08
Fix bug of textAnalyzer

YoyoTT:
*Issue #, if available:*
capi Kubeadm Control Plane System Log Analyzers cannot locate the log file due to the wrong file path. Output of the analyzer showed `message: No matching files`.

*Description of changes:*
Fix the bug by giving the analyzer the full manager pod log path and removing the collector name which should not be partial of it.

*Testing (if applicable):*
- run `make` to complie binaries
- then run `./bin/eksctl-anywhere generate support-bundle -f myCluster.yaml`
- the output of that analyzer shows "message: API server pods launched correctly" instead of "No matching files". 
- Details show below: 
`- URI: ""
  isFail: false
  isPass: true
  isWarn: false
  message: API server pods launched correctly
  title: 'log analysis:: API server pod missing. Log: logs/capi-kubeadm-control-plane-system/capi-kubeadm-control-plane-controller-manager-*.log'
`

### DIFF
--- a/pkg/diagnostics/analyzers.go
+++ b/pkg/diagnostics/analyzers.go
@@ -225,7 +225,7 @@ func (a *analyzerFactory) namespaceLogTextAnalyzersMap() map[string][]*Analyze {
 
 func (a *analyzerFactory) capiKubeadmControlPlaneSystemLogAnalyzers() []*Analyze {
 	capiCpManagerPod := "capi-kubeadm-control-plane-controller-manager-*"
-	capiCpManagerContainerLogFile := path.Join(capiCpManagerPod, "manager.log")
+	capiCpManagerContainerLogFile := capiCpManagerPod + ".log"
 	fullManagerPodLogPath := path.Join(logpath(constants.CapiKubeadmControlPlaneSystemNamespace), capiCpManagerContainerLogFile)
 	return []*Analyze{
 		{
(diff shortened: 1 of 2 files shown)


mitalipaygude:
Can you add some description and how you tested this?

YoyoTT:
> Can you add some description and how you tested this?
Yes, I have updated the description. Could you please take a look? Let me know anything I need to improve. Thanks!

pkg/diagnostics/analyzers.go
```diff
@@ -225,17 +225,16 @@ func (a *analyzerFactory) namespaceLogTextAnalyzersMap() map[string][]*Analyze {
...
-	capiCpManagerContainerLogFile := path.Join(capiCpManagerPod, "manager.log")
+	capiCpManagerContainerLogFile := capiCpManagerPod + ".log"
 	fullManagerPodLogPath := path.Join(logpath(constants.CapiKubeadmControlPlaneSystemNamespace), capiCpManagerContainerLogFile)
 	return []*Analyze{
 		{
 			TextAnalyze: &textAnalyze{
 				analyzeMeta: analyzeMeta{
 					CheckName: fmt.Sprintf("%s: API server pod missing. Log: %s", logAnalysisAnalyzerPrefix, fullManagerPodLogPath),
 				},
-				CollectorName: constants.CapiKubeadmControlPlaneSystemNamespace,
```
g-gaston, line 236: Why are we removing the collector name?

pkg/diagnostics/analyzers_test.go
```diff
@@ -26,3 +27,15 @@ func getDeploymentStatusAnalyzer(analyzers []*diagnostics.Analyze, name string)
...
 }
+
+func TestEksaLogTextAnalyzers(t *testing.T) {
+	collectorFactory := diagnostics.NewDefaultCollectorFactory()
+	collectors := collectorFactory.ManagementClusterCollectors()
+	analyzerFactory := diagnostics.NewAnalyzerFactory()
+	expectAnalzyers := analyzerFactory.EksaLogTextAnalyzers(collectors)
+	for _, analyzer := range expectAnalzyers {
+		if analyzer == nil {
+			t.Errorf("EksaLogTextAnalyzers: %s failed", analyzer.TextAnalyze.CheckName)
```
g-gaston, line 38: What are we testing here? Just that the analyzer is not nil? In that case I would make that info part of the error message

YoyoTT: Eksa Log Text Analyzers will check namespaced log collectors are present and then return the analyzers associated with that namespace. So I would say it is testing if there is any returned analyzers. If error message is changed to "None analyzer returned, analyzer.TextAnalyze.CheckName.", do you think it make more sense?

YoyoTT: It won't catch specific bug but test the functionality. I have change the error message by removing out the `analyzer.TextAnalyzer.CheckName`.

g-gaston:
I think we need to clean up those extra commits :)

ATTRIBUTION.txt
```diff
@@ -1,7 +1,4 @@
```
mitalipaygude, line 1: Is this an accidental change?

YoyoTT: I don't think I made that change.

PR #2712, opened 2022-07-11
Support for upgrading Cilium

mdsgabriel:
*Issue #, if available:*

*Description of changes:*
This will introduce a new reconciler for Cilium.
The code in this PR tried to reuse CLI code as much as possible.

*Testing (if applicable):*

### DIFF
--- a/pkg/controller/cni/cilium_reconciler.go
+++ b/pkg/controller/cni/cilium_reconciler.go
@@ -27,7 +27,6 @@ import (
 const defaultRequeueTime = time.Second * 10
 
 type CiliumReconciler struct {
-	Client  client.Client
 	Log     logr.Logger
 	tracker *remote.ClusterCacheTracker
 }
@@ -234,11 +233,8 @@ func getCiliumDeployment(ctx context.Context, client client.Client) (*v1.Deploym
 	return deployment, nil
 }
 
-func NewCiliumReconciler(
-	client client.Client, log logr.Logger, tracker *remote.ClusterCacheTracker,
-) *CiliumReconciler {
+func NewCiliumReconciler(log logr.Logger, tracker *remote.ClusterCacheTracker) *CiliumReconciler {
 	return &CiliumReconciler{
-		Client:  client,
 		Log:     log,
 		tracker: tracker,
 	}
--- a/pkg/controller/cni/cilium_reconciler.go
+++ b/pkg/controller/cni/cilium_reconciler.go
@@ -88,7 +88,7 @@ func (cr *CiliumReconciler) Reconcile(ctx context.Context, log logr.Logger, clus
 	}
 
 	if err := cilium.CheckDaemonSetReady(ciliumDS); err != nil {
-		log.Info("Cilium DS not ready")
+		log.Error(err, "error checking if Cilium DaemonSet is ready")
 		return controller.Result{Result: &ctrl.Result{
 			RequeueAfter: defaultRequeueTime,
 		}}, nil
@@ -97,14 +97,14 @@ func (cr *CiliumReconciler) Reconcile(ctx context.Context, log logr.Logger, clus
 	preFlightCiliumDS := &v1.DaemonSet{}
 	preFlightCiliumDSName := types.NamespacedName{Namespace: "kube-system", Name: cilium.PreflightDaemonSetName}
 	if err := remoteClient.Get(ctx, preFlightCiliumDSName, preFlightCiliumDS); err != nil {
-		log.Info("Preflight Cilium DS not found.")
+		log.Error(err, "error getting Preflight Cilium DS")
 		return controller.Result{Result: &ctrl.Result{
 			RequeueAfter: defaultRequeueTime,
 		}}, err
 	}
 
 	if err := cilium.CheckPreflightDaemonSetReady(ciliumDS, preFlightCiliumDS); err != nil {
-		log.Info("Preflight DS not ready ")
+		log.Error(err,"error checking if Preflight DS is ready ready")
 		return controller.Result{Result: &ctrl.Result{
 			RequeueAfter: defaultRequeueTime,
 		}}, err
@@ -112,7 +112,7 @@ func (cr *CiliumReconciler) Reconcile(ctx context.Context, log logr.Logger, clus
 
 	log.Info("Deleting Preflight Cilium objects")
 	if err := serverside.DeleteYaml(ctx, remoteClient, preflight); err != nil {
-		log.Info("Error deleting Preflight Cilium objects")
+		log.Error(err,"error deleting Preflight Cilium objects")
 		return controller.Result{}, err
 	}
 
--- a/pkg/controller/cni/cilium_reconciler.go
+++ b/pkg/controller/cni/cilium_reconciler.go
@@ -42,9 +42,7 @@ func (cr *CiliumReconciler) Reconcile(ctx context.Context, log logr.Logger, clus
 	}
 
 	log.Info("Applying CNI")
-	ciliumDS := &v1.DaemonSet{}
-	ciliumDSName := types.NamespacedName{Namespace: "kube-system", Name: cilium.DaemonSetName}
-	err = remoteClient.Get(ctx, ciliumDSName, ciliumDS)
+	ciliumDS, err := getCiliumDS(ctx, remoteClient)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			log.Info("Deploying Cilium DS")
@@ -86,13 +84,6 @@ func (cr *CiliumReconciler) Reconcile(ctx context.Context, log logr.Logger, clus
 
 	log.Info("Installing Cilium upgrade preflight manifest")
 	if err := serverside.ReconcileYaml(ctx, remoteClient, preflight); err != nil {
-		return controller.Result{}, err
-	}
-
-	ciliumDS = &v1.DaemonSet{}
-	ciliumDSName = types.NamespacedName{Namespace: "kube-system", Name: "cilium"}
-	if err := remoteClient.Get(ctx, ciliumDSName, ciliumDS); err != nil {
-		log.Info("Cilium DS not found")
 		return controller.Result{}, err
 	}
 
--- a/pkg/controller/cni/cilium_reconciler.go
+++ b/pkg/controller/cni/cilium_reconciler.go
@@ -10,8 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	"sigs.k8s.io/cluster-api/controllers/remote"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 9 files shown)


pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
+package cni
```
g-gaston, line 1: I would vote to move this to `pkg/networking/cilium` or `pkg/networking/cilium/reconciler`

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/aws/eks-anywhere/pkg/controller"
+	"github.com/aws/eks-anywhere/pkg/controller/serverside"
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/networking/cilium"
+)
+
+const defaultRequeueTime = time.Second * 10
+
+type CiliumReconciler struct {
+	Client  client.Client
```
g-gaston, line 30: What is this needed for?

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	Log     logr.Logger
```
g-gaston, line 31: I would vote to receive the logger as part of the reconcile method, so higher level entities can set values and pass it down

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	tracker *remote.ClusterCacheTracker
+}
+
+func (cr *CiliumReconciler) Reconcile(ctx context.Context, cluster *anywherev1.Cluster, capiCluster *clusterv1.Cluster, specWithBundles *eksacluster.Spec) (controller.Result, error) {
```
g-gaston, line 35: Instead of passing the capi cluster, it might be more flexible if we pass a `client.Client` directly. That way this reconciler becomes agnostic from the fact that it will be used with a remote client (provided by the tracker.

And we can implement the client generation one level higher.

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	remoteClient, err := cr.tracker.GetClient(ctx, key)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	cr.Log.Info("Applying CNI")
+	ciliumDS := &v1.DaemonSet{}
+	ciliumDSName := types.NamespacedName{Namespace: "kube-system", Name: cilium.DaemonSetName}
+	err = remoteClient.Get(ctx, ciliumDSName, ciliumDS)
```
g-gaston, line 49: why not user `getCiliumDS`?

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
g-gaston, line 51: I think we can un-nest this and have `apierrors.IsNotFound(err)` at the top level

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+			cr.Log.Info("Deploying Cilium DS")
+			helm := executables.NewHelm(executables.NewExecutable("helm"), executables.WithInsecure())
+
+			ci := cilium.NewCilium(nil, helm)
```
g-gaston, line 55: I think we should build this beforehand and inject it in the constructor, probably as an interface

And probably avoid having that nil in the constructor. That makes me think we need a smaller entity than `Cilium`

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+
+			ciliumSpec, err := ci.GenerateManifest(ctx, specWithBundles, []string{constants.CapvSystemNamespace})
```
g-gaston, line 57: we need to make this dynamic depending on the provider

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+			if err != nil {
+				return controller.Result{}, err
+			}
+			if err := serverside.ReconcileYaml(ctx, remoteClient, ciliumSpec); err != nil {
+				return controller.Result{}, err
+			}
+			return controller.Result{}, err
```
g-gaston, line 64: should this be 
```suggestion
			return controller.Result{}, nil
```

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+		}
```
g-gaston, line 65: probably worth extracting the "install" block to a method

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// upgrade cilium
+	cr.Log.Info("Upgrading Cilium")
+	needsUpgrade, err := ciliumNeedsUpgrade(ctx, cr.Log, remoteClient, specWithBundles)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	if !needsUpgrade {
+		cr.Log.Info("Skipping Cilium")
```
g-gaston, line 78: maybe something like `Cilium already reconciled`

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	log.Info("Checking if Cilium operator deployment needs upgrade")
+	needsUpgrade, err = ciliumOperatorNeedsUpgrade(ctx, log, client, clusterSpec)
+	if err != nil {
+		return false, err
+	}
+
+	if needsUpgrade {
+		log.Info("Cilium operator deployment needs upgrade")
```
g-gaston, line 161: maybe worth logging this one level up instead of here

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+
+	cr.Log.Info("Installing Cilium upgrade preflight manifest")
+	if err := serverside.ReconcileYaml(ctx, remoteClient, preflight); err != nil {
+		return controller.Result{}, err
+	}
+
+	ciliumDS = &v1.DaemonSet{}
+	ciliumDSName = types.NamespacedName{Namespace: "kube-system", Name: "cilium"}
+	if err := remoteClient.Get(ctx, ciliumDSName, ciliumDS); err != nil {
+		cr.Log.Info("Cilium DS not found")
```
g-gaston, line 97: why not found? it could be that there was a network error or anything else

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if err := cilium.CheckPreflightDaemonSetReady(ciliumDS, preFlightCiliumDS); err != nil {
+		cr.Log.Info("Preflight DS not ready ")
+		return controller.Result{Result: &ctrl.Result{
+			RequeueAfter: defaultRequeueTime,
+		}}, err
+	}
+
+	cr.Log.Info("Deleting Preflight Cilium objects")
+	if err := serverside.DeleteYaml(ctx, remoteClient, preflight); err != nil {
```
g-gaston, line 125: I think it might be safer to delete this after the upgrade manifest has been applied
that way we guarantee that the container images are not deleted

pkg/controller/cni/cni_reconciler.go
```diff
@@ -0,0 +1,26 @@
...
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	eksacluster "github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/controller"
+)
+
+type CNIReconciler interface {
+	Reconcile(ctx context.Context, cluster *anywherev1.Cluster, capiCluster *clusterv1.Cluster, specWithBundles *eksacluster.Spec) (controller.Result, error)
+}
+
+func BuildCNIReconciler(cniName string, client client.Client, log logr.Logger, tracker *remote.ClusterCacheTracker) (CNIReconciler, error) {
```
g-gaston, line 21: What is this for?

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: nothing. It was a copy and paste from other reconcilers. removed

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: good point. using getCiliumDS now

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+		cr.Log.Info("Cilium DS not ready")
+		return controller.Result{Result: &ctrl.Result{
+			RequeueAfter: defaultRequeueTime,
+		}}, nil
+	}
+
+	preFlightCiliumDS := &v1.DaemonSet{}
+	preFlightCiliumDSName := types.NamespacedName{Namespace: "kube-system", Name: cilium.PreflightDaemonSetName}
+	if err := remoteClient.Get(ctx, preFlightCiliumDSName, preFlightCiliumDS); err != nil {
+		cr.Log.Info("Preflight Cilium DS not found.")
```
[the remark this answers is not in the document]
mdsgabriel: using log.Error now

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: using.log.Error now to include the error message

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: done

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+	log.Info("Installing Cilium upgrade preflight manifest")
+	if err := serverside.ReconcileYaml(ctx, remoteClient, preflight); err != nil {
+		return controller.Result{}, err
+	}
+
+	if err := cilium.CheckDaemonSetReady(ciliumDS); err != nil {
+		log.Error(err, "error checking if Cilium DaemonSet is ready")
+		return controller.Result{Result: &ctrl.Result{
+			RequeueAfter: defaultRequeueTime,
+		}}, nil
```
mitalipaygude, line 94: Should we be returning err here?

pkg/networking/cilium/client.go
```diff
@@ -11,10 +11,10 @@ import (
 )
 
 const (
-	ciliumDaemonSetName           = "cilium"
-	ciliumPreflightDaemonSetName  = "cilium-pre-flight-check"
-	ciliumDeploymentName          = "cilium-operator"
-	ciliumPreflightDeploymentName = "cilium-pre-flight-check"
+	DaemonSetName           = "cilium"
```
mitalipaygude, line 14: just out of curiosity, why did we remove the cilium prefix?

mdsgabriel: To use "cilium.DaemonSetname" instead of "cilium.ciliumDaemonSetName" outside the package. The package name already serves as namespace for the constants.

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: The goal here is to reschedule reconciliation after defaultRequeueTime, but the current code returns an empty ctrl.Result if an error is returned by clusterProviderReconciler.Reconcile()

https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L164

pkg/controller/cni/cilium_reconciler.go
mdsgabriel: done

pkg/controller/cni/cni_reconciler.go
mdsgabriel: it generates a CNI reconciler

pkg/controller/cni/cilium_reconciler.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	if !needsUpgrade {
+		log.Info("Skipping Cilium")
+		return controller.Result{}, nil
+	}
+
+	helm := executables.NewHelm(executables.NewExecutable("helm"))
```
jiayiwang7, line 64: can we build helm exec on a higher level and use it in this method instead of building twice?

PR #2724, opened 2022-07-13
Add snowmachineconfig controller

vivek-koppuru:
*Issue #, if available:*
#2688 
*Description of changes:*
These changes add a snow machine config controller with dynamic validations that we have today for create cluster. 

The following things are still action items:
* Creating the clients and setting up the creds in the controller reconciliation. Either we need to construct the clients with the creds before constructing the reconciler or construct the clients and pass in the creds for validation calls every time we reconcile
* Pulling the creds from a secret in the cluster to be able to properly build the aws clients
* Add defaulter for snow machine config to generate default ssh keys and have them retrievable through a secret on the cluster (could be a future feature)
* Get skipIp attribute and implement this check in the controller https://github.com/aws/eks-anywhere/blob/main/pkg/providers/snow/snow.go#L74 (could be a future feature)
* Validate this workflow apart from just unit testing

*Testing (if applicable):*
unit testing


vivek-koppuru:
/hold

will do some additional testing

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if err := validator.ValidateEC2ImageExistsOnDevice(ctx, snowMachineConfig); err != nil {
+		log.Error(err, "failed validating ec2 image on devices")
+		return ctrl.Result{}, err
+	}
+	if err := validator.ValidateEC2SshKeyNameExists(ctx, snowMachineConfig); err != nil {
+		log.Error(err, "failed validating ec2 ssh key name")
+		return ctrl.Result{}, err
+	}
+	// TODO: Add Defaulter for snow machine config to generate default ssh keys and store as secrets on cluster
```
g-gaston, line 102: IMO we should skip this for now. We decided something similar for vsphere: we don't generate keys from the controller, the user needs to pass them.

If there are user's requests later, we can added.

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	// TODO: need to figure out how to load creds in controller
+	deviceClientMap, err := r.ClientBuilder.BuildSnowAwsClientMap(ctx)
+	if err != nil {
+		log.Error(err, "failed to build aws client map")
+		return ctrl.Result{}, err
+	}
+	// Setting the aws client map on every reconcile based on the secrets at that point of time
+	validator := r.Validator.WithAwsClients(deviceClientMap)
+	if err := validator.ValidateMachineDeviceIPs(ctx, snowMachineConfig); err != nil {
+		log.Error(err, "failed validating machine device ips")
```
g-gaston, line 91: We don't need to log the error, `controller-runtime` will do it if a reconcile method returns an error.

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+
+// SnowMachineConfigReconciler reconciles a SnowMachineConfig object
+type SnowMachineConfigReconciler struct {
+	Client        client.Client
+	Log           logr.Logger
+	ClientBuilder aws.ClientBuilder
+	Validator     *snow.Validator
+}
+
+func NewSnowMachineConfigReconciler(client client.Client, log logr.Logger, clientBuilder aws.ClientBuilder, validator *snow.Validator) *SnowMachineConfigReconciler {
```
g-gaston, line 25: I think we should try to receive interfaces defined in this file if possible

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	patchHelper, err := patch.NewHelper(snowMachineConfig, r.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	defer func() {
+		// Always attempt to patch the object and status after each reconciliation.
+		patchOpts := []patch.Option{}
+		if reterr == nil {
+			patchOpts = append(patchOpts, patch.WithStatusObservedGeneration{})
```
g-gaston, line 61: I don't think we have observed generation for for this crd
But we could add it

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	// There's no need to go any further if the SnowMachineConfig is marked for deletion.
+	if !snowMachineConfig.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, snowMachineConfig, log)
+	}
+
+	result, err := r.reconcile(ctx, snowMachineConfig, log)
+	if err != nil {
+		log.Error(err, "Failed to reconcile SnowMachineConfig")
+	}
+	return result, nil
```
g-gaston, line 78: why not return the error?

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/aws/eks-anywhere/pkg/aws"
+	"github.com/aws/eks-anywhere/pkg/providers/snow"
+)
+
+// SnowMachineConfigReconciler reconciles a SnowMachineConfig object
+type SnowMachineConfigReconciler struct {
+	Client        client.Client
+	Log           logr.Logger
+	ClientBuilder aws.ClientBuilder
+	Validator     *snow.Validator
```
g-gaston, line 22: I think we can make private all of these

pkg/api/v1alpha1/snowmachineconfig_types.go
```diff
@@ -17,6 +17,7 @@ type PhysicalNetworkConnectorType string
 type SnowInstanceType string
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// Important: Run "make generate" to regenerate code after modifying this file
```
g-gaston, line 20: should we test this in a pre-submit?

pkg/dependencies/factory_test.go
```diff
@@ -130,6 +130,7 @@ func TestFactoryBuildWithMultipleDependencies(t *testing.T) {
 		WithCmk().
 		WithVSphereDefaulter().
 		WithVSphereValidator().
+		WithSnowValidatorForController().
```
g-gaston, line 133: nil check?

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (r *SnowMachineConfigReconciler) reconcile(ctx context.Context, snowMachineConfig *anywherev1.SnowMachineConfig, log logr.Logger) (_ ctrl.Result, reterr error) {
+	// TODO: need to figure out how to load creds in controller
+	deviceClientMap, err := r.ClientBuilder.BuildSnowAwsClientMap(ctx)
+	if err != nil {
+		log.Error(err, "failed to build aws client map")
+		return ctrl.Result{}, err
+	}
+	// Setting the aws client map on every reconcile based on the secrets at that point of time
+	validator := r.Validator.WithAwsClients(deviceClientMap)
```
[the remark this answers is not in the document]
vivek-koppuru: Yea I was debating between the two but wanted to see if this implementation would work with the assumption that we don't have to worry about thread safety. I think I prefer creating the validator dynamically so that we don't affect the regular snow flow, but will try it out first and see if the passing in the client is better.

controllers/snow_machineconfig_controller.go
vivek-koppuru: Got it, moved ClientBuilder interface to this file and the Validator will be dynamically created in reconcile

controllers/snow_machineconfig_controller.go
vivek-koppuru: Just kept it as we have this for the other controllers, plus this would help in the future when we actually implement observed generation to know the real status of the spec, so no reason to omit it now right?

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		if err := patchHelper.Patch(ctx, snowMachineConfig, patchOpts...); err != nil {
+			log.Error(reterr, "Failed to patch snowmachineconfig")
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+	}()
+
+	// There's no need to go any further if the SnowMachineConfig is marked for deletion.
+	if !snowMachineConfig.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, snowMachineConfig, log)
```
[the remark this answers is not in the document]
vivek-koppuru: No, I can delete it. Kept it because the vsphere datacenter config controller had it too

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	result, err := r.reconcile(ctx, snowMachineConfig, log)
```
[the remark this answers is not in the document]
vivek-koppuru: Good catch, realized once I wrote unit tests for this that it was failing because of it

controllers/snow_machineconfig_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	// Setting the aws client map on every reconcile based on the secrets at that point of time
+	validator := r.Validator.WithAwsClients(deviceClientMap)
+	if err := validator.ValidateMachineDeviceIPs(ctx, snowMachineConfig); err != nil {
+		log.Error(err, "failed validating machine device ips")
+		return ctrl.Result{}, err
+	}
+	if err := validator.ValidateEC2ImageExistsOnDevice(ctx, snowMachineConfig); err != nil {
+		log.Error(err, "failed validating ec2 image on devices")
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
vivek-koppuru: We are using `FailureMessage` in `ClusterStatus` as well as it being referenced below, so I will do `FailureMessage` for now unless we want to suggest a change to `ValidationError`

https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md#new-vspheredatacenterconfigstatus

pkg/api/v1alpha1/snowmachineconfig_types.go
vivek-koppuru: Created this issue: https://github.com/aws/eks-anywhere/issues/2734

pkg/dependencies/factory.go
```diff
@@ -1043,6 +1026,20 @@ func (f *Factory) WithVSphereValidator() *Factory {
 	return f
 }
 
+func (f *Factory) WithSnowValidatorForController() *Factory {
```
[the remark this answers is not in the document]
vivek-koppuru: To differentiate the way we were building the validators, but changed the logic now so it's not needed

pkg/dependencies/factory_test.go
vivek-koppuru: Not relevant anymore

controllers/snow_machineconfig_controller.go
vivek-koppuru: Let me know if I am misunderstanding this tho

jiayiwang7 (approved):
Looks great

PR #2726, opened 2022-07-14
Cluster validations in webhook

mitalipaygude:
*Issue #, if available:*
#1090 

*Description of changes:*
Add cluster validations webhook.

*Testing (if applicable):*
- Unit tests
- Manual testing with tilt
- 

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -19,6 +19,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -54,19 +55,31 @@ var _ webhook.Validator = &Cluster{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *Cluster) ValidateCreate() error {
 	clusterlog.Info("validate create", "name", r.Name)
-	if r.IsReconcilePaused() {
-		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
-		return nil
-	}
+	var allErrs []error
+
 	if !features.IsActive(features.FullLifecycleAPI()) {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
-	}
-	if r.IsSelfManaged() {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+		allErrs = append(allErrs,
+			fmt.Errorf("creating new cluster on existing cluster is not supported"))
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if err := r.Validate(); err != nil {
+		allErrs = append(
+			allErrs,
+			fmt.Errorf("invalid cluster %v", err.Error()))
+	}
+
+	if r.IsSelfManaged() {
+		if !r.IsReconcilePaused() {
+			allErrs = append(
+				allErrs, fmt.Errorf("creating new cluster on existing cluster is not supported for self managed clusters"))
+		} else {
+			clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
+		}
+	}
+
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return apierrors.NewBadRequest(fmt.Sprintf("validating cluster spec: %v", aggregate))
 	}
 
 	return nil
@@ -88,20 +101,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
 	}
 
-	// Test for both taints and labels
-	if err := validateWorkerNodeGroups(r); err != nil {
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "workerNodeGroupConfigurations"), r.Spec.WorkerNodeGroupConfigurations, err.Error()))
-	}
-
-	// Control plane configuration is mutable if workload cluster
-	if !r.IsSelfManaged() {
-		if err := validateControlPlaneLabels(r); err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneConfiguration", "labels"), r.Spec, err.Error()))
-		}
-	}
-
-	if len(allErrs) != 0 {
-		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
+	if err := r.Validate(); err != nil {
+		return err
 	}
 
 	return nil
--- a/pkg/api/v1alpha1/cluster_webhook_test.go
+++ b/pkg/api/v1alpha1/cluster_webhook_test.go
@@ -1,6 +1,7 @@
 package v1alpha1_test
 
 import (
+	"strings"
 	"testing"
 
 	. "github.com/onsi/gomega"
@@ -11,6 +12,8 @@ import (
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/features"
 )
+
+const pausedAnnotation = "anywhere.eks.amazonaws.com/paused"
 
 func TestClusterDefault(t *testing.T) {
 	cOld := &v1alpha1.Cluster{}
@@ -36,14 +39,7 @@ func TestClusterValidateUpdateManagementValueImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementOldNilNewTrueSuccess(t *testing.T) {
-	cOld := &v1alpha1.Cluster{
-		Spec: v1alpha1.ClusterSpec{
-			WorkerNodeGroupConfigurations: []v1alpha1.WorkerNodeGroupConfiguration{{
-				Count: 1,
-				Name:  "test",
-			}},
-		},
-	}
+	cOld := createCluster()
 	c := cOld.DeepCopy()
 	c.SetSelfManaged()
 
@@ -61,14 +57,7 @@ func TestClusterValidateUpdateManagementOldNilNewFalseImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementBothNilImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{
-		Spec: v1alpha1.ClusterSpec{
-			WorkerNodeGroupConfigurations: []v1alpha1.WorkerNodeGroupConfiguration{{
-				Count: 1,
-				Name:  "test",
-			}},
-		},
-	}
+	cOld := createCluster()
 	c := cOld.DeepCopy()
 
 	g := NewWithT(t)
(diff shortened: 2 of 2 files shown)


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -58,14 +58,16 @@ func (r *Cluster) ValidateCreate() error {
 		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
 		return nil
```
g-gaston, line 59: this will skip the validation, I don't think we want that

Maybe we need to reorder all these statement in validate create

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -88,20 +90,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
-	if err := validateWorkerNodeGroups(r); err != nil {
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "workerNodeGroupConfigurations"), r.Spec.WorkerNodeGroupConfigurations, err.Error()))
-	}
-
-	// Control plane configuration is mutable if workload cluster
-	if !r.IsSelfManaged() {
-		if err := validateControlPlaneLabels(r); err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneConfiguration", "labels"), r.Spec, err.Error()))
-		}
-	}
```
g-gaston, line 101: For my own understanding, why are we removing these two?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -88,20 +90,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
-
-	if len(allErrs) != 0 {
-		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
+	if err := r.Validate(); err != nil {
+		return apierrors.NewBadRequest(err.Error())
```
g-gaston, line 94: why not use `allErrs`?

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1423,3 +1157,75 @@ func TestClusterUpdateControlPlaneLabelsInvalid(t *testing.T) {
...
+
+func TestClusterValidateCreateActiveFeature(t *testing.T) {
+	cOld := createCluster()
+
+	g := NewWithT(t)
+	g.Expect(cOld.ValidateCreate()).NotTo(Succeed())
+}
+
+func TestClusterValidateCreatePass(t *testing.T) {
+	os.Setenv(features.FullLifecycleAPIEnvVar, "true")
```
g-gaston, line 1178:
```suggestion
	t.Setenv(features.FullLifecycleAPIEnvVar, "true")
```

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1423,3 +1157,75 @@ func TestClusterUpdateControlPlaneLabelsInvalid(t *testing.T) {
...
 	g.Expect(c.ValidateUpdate(cOld)).NotTo(Succeed())
 }
+
+func TestClusterValidateCreateFail(t *testing.T) {
+	cOld := createCluster()
+	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{}
+	cOld.SetManagedBy("management-cluster")
+
+	g := NewWithT(t)
+	g.Expect(cOld.ValidateCreate()).NotTo(Succeed())
```
g-gaston, line 1167: Can we validate part of the error message?

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: Because these are already validated in the cluster Validate() method.

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: even with the pause annotation we want the validations to be checked for?
I can move the logic around then.

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: We dont add anything to the allErrs after this statement, so just returning from here.

pkg/api/v1alpha1/cluster_webhook.go
g-gaston: I would vote to always validate, even with annotation, yes

pkg/api/v1alpha1/cluster_webhook.go
g-gaston: Is there a reason to do that? I wonder if we should, that's the idea of having an array of errors

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,33 @@ var _ webhook.Validator = &Cluster{}
...
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+		allErrs = append(
+			allErrs,
+			fmt.Errorf("creating new cluster on existing cluster is not supported for self managed clusters"))
+	}
+
+	if err := r.Validate(); err != nil {
+		allErrs = append(
+			allErrs,
+			fmt.Errorf("cluster validation failed %v", err.Error()))
```
g-gaston, line 74:
```suggestion
			fmt.Errorf("invalid cluster %v", err.Error()))
```

maybe we should use some constructor for invalid field?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,33 @@ var _ webhook.Validator = &Cluster{}
...
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return fmt.Errorf("validating cluster spec: %v", aggregate)
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if r.IsReconcilePaused() {
+		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
+		return nil
```
g-gaston, line 84: I think this is broken now
I believe we want to use this flag in combination with `r.IsSelfManaged()`

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -84,20 +99,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
-
-	// Control plane configuration is mutable if workload cluster
-	if !r.IsSelfManaged() {
-		if err := validateControlPlaneLabels(r); err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneConfiguration", "labels"), r.Spec, err.Error()))
-		}
+	if err := r.Validate(); err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(field.NewPath("Cluster validation failed"), err.Error(), ""))
```
g-gaston, line 105: I don't think we use use `NewPath` for an error message, I believe that is only used to specify the invalid path

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1150,90 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+
+func TestClusterValidateCreateFail(t *testing.T) {
+	t.Setenv(features.FullLifecycleAPIEnvVar, "true")
+	cOld := createCluster()
+	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{}
+	cOld.SetManagedBy("management-cluster")
+
+	g := NewWithT(t)
+	g.Expect(cOld.ValidateCreate()).NotTo(Succeed())
+	strings.Contains(cOld.ValidateCreate().Error(), "cluster validation failed")
```
g-gaston, line 1175: I believe this is not testing anything, we are not even checking the result of Contains

I would advice using gomega with an expect for error contains substring. I think there should already be examples in our codebase

Also, why call `ValidateCreate()` twice?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -88,20 +101,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
-	// Control plane configuration is mutable if workload cluster
-	if !r.IsSelfManaged() {
-		if err := validateControlPlaneLabels(r); err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneConfiguration", "labels"), r.Spec, err.Error()))
-		}
-	}
-
-	if len(allErrs) != 0 {
-		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
+	if err := r.Validate(); err != nil {
```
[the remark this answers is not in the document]
mitalipaygude: In ValidateUpdate, we currently have a lot of fields getting validated where the errList makes more sense. But if we end up having more validation added here, we could use the aggregate errors here in the future.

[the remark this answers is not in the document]
jiayiwang7: not sure if i get it, because there are many validations already so that we don't add error returned by `r.Validate()` to the errList?

jiayiwang7 (approved):

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
mitalipaygude: Continued discussion offline.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,31 @@ var _ webhook.Validator = &Cluster{}
...
 	if !features.IsActive(features.FullLifecycleAPI()) {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+		allErrs = append(allErrs,
+			fmt.Errorf("creating new cluster on existing cluster is not supported"))
 	}
+
+	if err := r.Validate(); err != nil {
+		allErrs = append(
+			allErrs,
+			fmt.Errorf("invalid cluster %v", err.Error()))
```
g-gaston, line 68:
```suggestion
			fmt.Errorf("invalid cluster: %v", err))
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,31 @@ var _ webhook.Validator = &Cluster{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *Cluster) ValidateCreate() error {
 	clusterlog.Info("validate create", "name", r.Name)
-	if r.IsReconcilePaused() {
-		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
-		return nil
-	}
+	var allErrs []error
+
 	if !features.IsActive(features.FullLifecycleAPI()) {
```
g-gaston, line 60: I don't think this is correct. This will add the error even for self-managed clusters

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,31 @@ var _ webhook.Validator = &Cluster{}
...
+
+	if err := r.Validate(); err != nil {
+		allErrs = append(
+			allErrs,
+			fmt.Errorf("invalid cluster %v", err.Error()))
+	}
+
 	if r.IsSelfManaged() {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+		if !r.IsReconcilePaused() {
```
g-gaston, line 72: No need to nest, I would put the two conditions in one clause

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,31 @@ var _ webhook.Validator = &Cluster{}
...
+		} else {
+			clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
+		}
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return apierrors.NewBadRequest(fmt.Sprintf("validating cluster spec: %v", aggregate))
```
g-gaston, line 82: do we need another top level message? I would just use the error from the aggregate directly

g-gaston: Also, it might be a good idea to look at how StatusErrors are build from field.ErrorList, I think that process can help us do the same for `[]error`

mitalipaygude:
Fixed in #2787

Issue #2657, opened 2022-07-05
WorkerNodeGroupConfiguration.count default to 1


jiayiwang7:
**What happened**:

https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L233

We claim to default the `workerNodeGroupConfiguration.count` to 1 but the current code sets `machinedeployment.spec.replicas` = 0 if count not specified.

**What you expected to happen**:

We should
1. set `workerNodeGroupConfiguration.count` default to 1
2. add validation logic: `workerNodeGroupConfiguration.count != 0` or the cluster creation fails with cert manager not able to be installed on workload cluster since by default it's deployed to worker node.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

taneyland:
#2728

PR #2728, opened 2022-07-14
Validate worker node group count

taneyland:
*Issue #, if available:*
#2657 

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_defaults.go
+++ b/pkg/api/v1alpha1/cluster_defaults.go
@@ -50,11 +50,6 @@ func setWorkerNodeGroupDefaults(cluster *Cluster) error {
 		logger.V(1).Info("First worker node group name not specified. Defaulting name to md-0.")
 		cluster.Spec.WorkerNodeGroupConfigurations[0].Name = "md-0"
 	}
-	for i := range cluster.Spec.WorkerNodeGroupConfigurations {
-		if cluster.Spec.WorkerNodeGroupConfigurations[i].Count <= 0 {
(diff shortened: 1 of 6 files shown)


pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -50,6 +50,11 @@ func setWorkerNodeGroupDefaults(cluster *Cluster) error {
 		logger.V(1).Info("First worker node group name not specified. Defaulting name to md-0.")
 		cluster.Spec.WorkerNodeGroupConfigurations[0].Name = "md-0"
 	}
+	for i := range cluster.Spec.WorkerNodeGroupConfigurations {
```
jiayiwang7, line 53:
```suggestion
	for _, wng := range cluster.Spec.WorkerNodeGroupConfigurations {
	    if wng.Count <= 0 {
	      wng.Count = 1
	    }
	
	}
	
```
```suggestion
	for i := range cluster.Spec.WorkerNodeGroupConfigurations {
```

taneyland: I tried the first suggestion initially, but it doesn't actually update the count in the cluster spec. That's why I used the index

TerryHowe (approved):
/lgtm

Doesn't actually set the worker node group count to 1, but I think this fits the intention

taneyland:
> /lgtm
> 
> Doesn't actually set the worker node group count to 1, but I think this fits the intention

Yes I talked to Joey offline & we decided to just add this validation instead of setting the default to 1. Will update the PR title for clarity

jiayiwang7 (approved):

PR #2729, opened 2022-07-14
Increase support bundle pod logs collection 'since' time to 3 hours.

ptrivedi:
Issue:
In certain failure cases, the logs for tinkerbell infrastructure pods
were empty. This was because the support bundle log collection duration
was set to one hour. For certain timeout scenarios in bare metal world,
the failure and hence the support bundle collection could take more an
hour since start of cluster creation. Tinkerbell infrastructure pods are
mostly inactive once provisioning finishes or fails, and may not be
actively logging resulting in empty hourly logs for these pods.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/diagnostic_bundle.go
+++ b/pkg/diagnostics/diagnostic_bundle.go
@@ -370,7 +370,9 @@ func ParseTimeOptions(since string, sinceTime string) (*time.Time, error) {
 	var sinceTimeValue time.Time
 	var err error
 	if sinceTime == "" && since == "" {
-		return &sinceTimeValue, nil
+		return &sinceTimeValue, nil // returning an uninitialized (zero) Time value here results in a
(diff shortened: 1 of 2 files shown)


pkg/diagnostics/diagnostic_bundle.go
```diff
@@ -370,7 +370,9 @@ func ParseTimeOptions(since string, sinceTime string) (*time.Time, error) {
 	var sinceTimeValue time.Time
 	var err error
 	if sinceTime == "" && since == "" {
-		return &sinceTimeValue, nil
+		return &sinceTimeValue, nil    // returning an uninitialized (zero) Time value here results in a
```
danbudris, line 373: thanks for including this clarification

danbudris (approved):

PR #2730, opened 2022-07-14
Preflight Checks doc

raymond-zhang00:
*Issue #, if available:*
Preflight check design doc

*Description of changes:*
Adding preflight check design doc


designs/preflight-checks.md
```diff
@@ -0,0 +1,163 @@
...
+`        ``├──`` api``/``v1alpha1`
+`        ``│``   ``└──`` cluster``.``go`
+`        ``├──`` providers`
+`        ``│``   ``├──`` cloudstack`
+`        ``│``   ``├──`` docker`
+`        ``│``   ``├──`` snow`
+`        ``│``   ``├──`` tinkerbell`
+`        ``│``   ``└──`` vsphere`
+`        ``└──`` validations`
+`            ``├── validatevalidations`
```
g-gaston, line 89: in this package new? The name is a bit weird

designs/preflight-checks.md
```diff
@@ -0,0 +1,163 @@
...
+
+**Problem:** Users do not have a method to validate configuration files to be used with EKS Anywhere without attempting to create a cluster. 
+
+With more complex provider configurations, users may need to review and modify their configuration multiple times before arriving at a valid configuration for cluster creation. Running the create cluster command for validation commits the user to attempting to create a cluster and generates artifacts that need to be removed.
+
+### Tenets
+
+* ****Simple:**** simple to use, simple to understand, simple to maintain
+* **Informative**: output should lead user to take action and guide next steps 
+* **Non-invasive:** minimal permanent changes should be left as a result of this comman**d**
```
chrisdoherty4, line 15:
```suggestion
* **Non-invasive:** minimal permanent changes should be left as a result of this command
```

designs/preflight-checks.md
```diff
@@ -0,0 +1,163 @@
...
+
+* Validate EKS Anywhere configurations without running `create cluster`
+* Validate required resources are available prior to cluster creation 
+* Receive informative feedback about configuration and resource readiness
+
+### Statement of scope
+
+**In scope**
+
+* Preflight validation checks can be run independently of `create cluster`
```
chrisdoherty4, line 30:
```suggestion
```

This is a goal :)

designs/preflight-checks.md
g-gaston: can they just be in `validations`?
Or maybe depending on what they are validating they can go in there own packages? Like if it's a docker validation, in a docker package, etc.

@chrisdoherty4 is better than be at naming packages

chrisdoherty4: If its a set of validation helpers `validation` is fine. Initially a `validation.go` is fine and if we start finding groups of algorithms like 'network validation' then we can isolate that in its own source file down the road.

I wouldn't follow the `createvalidations` thing - that's rather odd.

raymond-zhang00: targeted validation helpers under `validation` folder but ended up making something pretty similar to the `createvalidations` flow - which as Chris mentioned may not be what we're shooting for. Will follow up to see what makes sense to move/restructure.

raymond-zhang00: Updated to move away from the `createvalidations` flow, currently this just consists of a `createcluster` pkg in validations that runs the associated validations in sequence.

designs/preflight-checks.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Addonmanager/addonclients validation ([link](https://github.com/aws/eks-anywhere/blob/b4a4eb84c03091d1a646c1e49b9760b8b63961d3/pkg/addonmanager/addonclients/fluxaddonclient.go#L289))
+    * fluxaddonclient
+* Additional individual validation functions in validations/createcluster called directly to replicate `create cluster` validation tasks
+
+#### Usage:
+
+`exp` - subcommand, introduce all subcommands as experimental
+
+`eks-a exp validate create cluster [flags]`
```
taneyland, line 101:
```suggestion
`eksctl anywhere exp validate create cluster [flags]`
```

PR #2742, opened 2022-07-15
Get latest ECR tag for Tinkerbell charts

abhay-krishna:
We don't push `latest` tag for Tinkerbell charts, so need to use the ECR API to get the latest tag pushed by time.


pokearu (approved):
/lgtm
/approve

PR #2743, opened 2022-07-16
Remove kube-rbac-proxy from EKS-A controller deployment

abhay-krishna:
We don't need kube-rbac-proxy in the EKS-A cluster controller `Deployment` manifest. We should have done this way back during v1alpha4 migration.


abhay-krishna:
Force-merging to fix E2E tests.

PR #2752, opened 2022-07-19
Add CertManagerExits logic for PackageController Installation

ivyostosh:
The existence of cert-manager is required to ensure proper operation of the PackageController. We modify the cluster creation workflow and imperative package installation command to add in the CertManagerExits logic, and guide users to handle the case where cert-manager is not present in the cluster.

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/278

*Description of changes:* Modified the cluster creation workflow and imperative package installation command to add in the CertManagerExits logic

*Testing (if applicable):* Tested in cluster creation workflow and imperative package installation

### DIFF
--- a/cmd/eksctl-anywhere/cmd/installpackagecontroller.go
+++ b/cmd/eksctl-anywhere/cmd/installpackagecontroller.go
@@ -2,14 +2,12 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
 
 	"github.com/spf13/cobra"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
-	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 	"github.com/aws/eks-anywhere/pkg/utils/urls"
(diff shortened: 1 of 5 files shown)


pkg/curatedpackages/packageinstaller.go
```diff
@@ -49,6 +53,18 @@ func newPackageController(installer ChartInstaller, runner KubectlRunner, spec *
 
 func (pi *Installer) InstallCuratedPackages(ctx context.Context) error {
 	PrintLicense()
+
+	// If cert-manager does not exist, instruct users to follow instructions in
+	// PrintCertManagerDoesNotExistMsg to install packages manually.
+	// Note although we passed in a namespace parameter in the kubectl command, the GetResource command will be
+	// performed in all namespaces since CRDs are not bounded by namespaces.
+	kubeConfig := kubeconfig.FromClusterName(pi.spec.Cluster.Name)
+	certManagerExists, _ := pi.kubectl.GetResource(ctx, "crd", "certificates.cert-manager.io", kubeConfig, constants.CertManagerNamespace)
```
g-gaston, line 62: What's the reason to ignore the error? In general, that's not good practice

pkg/curatedpackages/packageinstaller.go
```diff
@@ -49,6 +53,18 @@ func newPackageController(installer ChartInstaller, runner KubectlRunner, spec *
...
+	if !certManagerExists {
+		PrintCertManagerDoesNotExistMsg()
+		return errors.New("cert-manager is not present in the cluster")
+	}
```
g-gaston, line 66: This block of code is duplicated, could we, for example, extract it to a separate method?

pkg/curatedpackages/curatedpackages.go
```diff
@@ -70,9 +74,22 @@ func PrintLicense() {
...
+
+func PrintCertManagerDoesNotExistMsg() {
+	// Currently, use the width of the longest line to repeat the dashes
+	// Sample Output
+	//----------------------------------------------------------------------------------------------------------------
+	//Curated packages cannot be installed as cert-manager is not present in the cluster. This is most likely caused
+	//by an action to install Curated packages at a workload cluster. Refer to
+	//https://anywhere.eks.amazonaws.com/docs/tasks/packages/ for how to resolve this issue.
+	//----------------------------------------------------------------------------------------------------------------
+	fmt.Println(strings.Repeat(constants.UserMsgSeparator, width))
```
g-gaston, line 90: is it worth to store the separator in a const/variable? It seems like we know what this need to be at compile time

ivyostosh (commented):
Comments addressed

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -79,6 +79,11 @@ func installPackageController(ctx context.Context) error {
 		helmChart.Tag(),
 	)
 
+	err = curatedpackages.VerifyCertManagerExists(ctx, deps.Kubectl, kubeConfig)
+	if err != nil {
```
g-gaston, line 83: nit:
```suggestion
	if err = curatedpackages.VerifyCertManagerExists(ctx, deps.Kubectl, kubeConfig); err != nil {
```

ivyostosh: Updated

g-gaston (approved):

PR #2759, opened 2022-07-19
Install flux controllers during components upgrade if not exist before

jiayiwang7:
*Issue #, if available:*

First part of #2755 

*Description of changes:*

This PR only handles the case of installing new Flux controllers in a Flux disabled cluster through upgrade command.

During components upgrade, when FluxConfig does not exist in the existing cluster, but specified in the new cluster spec, we install the flux controllers **without** adding EKS-A spec in the git repo. `UpdateGitEksaSpec` in the "updateClusterAndGitResources" step will then add the new EKS-A spec to the repo once cluster is fully upgraded. 

Logic behind this setting:
- It makes logical sense to install controllers as part of the EKS-A components during components upgrade step
- Only install the Flux controller with its controller configs during components upgrade so that none of the EKS-A resources will be reconciled by GitOps during upgrade. The EKS-A cluster controller will not interfere with CLI upgrade process.
- EKS-A spec will only be added after cluster is fully upgraded with latest spec. So that the cluster spec in Git repo will reflect the actual cluster config. If upgrade failed, the cluster spec will not be pushed. 

Also refactored addon unit tests a bit.

We will have upcoming PRs to handle
- Update Flux configs in a cluster
- Disable Flux in a cluster
- Add proper validation logic for Flux immutable fields

*Testing (if applicable):*

- Unit tests
- Manual local test to enable Flux in vsphere cluster with upgrade command.

### DIFF
--- a/pkg/addonmanager/addonclients/upgrader.go
+++ b/pkg/addonmanager/addonclients/upgrader.go
@@ -51,6 +51,10 @@ func FluxChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 		logger.V(1).Info("Skipping Flux upgrades, not a self-managed cluster")
 		return nil
 	}
+	if currentSpec.Cluster.Spec.GitOpsRef == nil && newSpec.Cluster.Spec.GitOpsRef != nil {
+		logger.V(1).Info("Skipping Flux upgrades, no previous flux installed in the cluster")
+		return nil
+	}
 	if currentSpec.Cluster.Spec.GitOpsRef == nil {
 		logger.V(1).Info("Skipping Flux upgrades, GitOps not enabled")
 		return nil
@@ -68,6 +72,13 @@ func FluxChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 				},
 			},
 		}
+	}
+	return nil
+}
+
+func (f *FluxAddonClient) Install(ctx context.Context, cluster *types.Cluster, oldSpec, newSpec *cluster.Spec) error {
+	if oldSpec.Cluster.Spec.GitOpsRef == nil && newSpec.Cluster.Spec.GitOpsRef != nil {
+		return f.InstallGitOps(ctx, cluster, newSpec, nil, nil)
 	}
 	return nil
 }
--- a/pkg/workflows/upgrade.go
+++ b/pkg/workflows/upgrade.go
@@ -238,6 +238,11 @@ func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.Co
 	}
 	commandContext.UpgradeChangeDiff.Append(changeDiff)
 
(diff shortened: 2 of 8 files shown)


pkg/workflows/upgrade.go
```diff
@@ -238,6 +238,11 @@ func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.Co
 	}
 	commandContext.UpgradeChangeDiff.Append(changeDiff)
 
+	if err = commandContext.AddonManager.Install(ctx, commandContext.ManagementCluster, commandContext.CurrentClusterSpec, commandContext.ClusterSpec); err != nil {
```
danbudris, line 241: nit: consistency with above `if err != nil` error handling as opposed to `if err =`

pkg/addonmanager/addonclients/upgrader.go
```diff
@@ -51,6 +51,10 @@ func FluxChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 		logger.V(1).Info("Skipping Flux upgrades, not a self-managed cluster")
 		return nil
 	}
+	if currentSpec.Cluster.Spec.GitOpsRef == nil && newSpec.Cluster.Spec.GitOpsRef != nil {
+		logger.V(1).Info("Skipping Flux upgrades, already installed latest")
+		return nil
```
mitalipaygude, line 56: Can we add an unit test for these lines?

pkg/addonmanager/addonclients/upgrader.go
```diff
@@ -51,6 +51,10 @@ func FluxChangeDiff(currentSpec, newSpec *cluster.Spec) *types.ChangeDiff {
 		logger.V(1).Info("Skipping Flux upgrades, not a self-managed cluster")
 		return nil
 	}
+	if currentSpec.Cluster.Spec.GitOpsRef == nil && newSpec.Cluster.Spec.GitOpsRef != nil {
+		logger.V(1).Info("Skipping Flux upgrades, already installed latest")
```
danbudris, line 55: should we add a comment here the explains why the diff does not take into account the addition of a 'new' flux version? I would find it a little confusing to be new to this code and see that the diff did not take into account the difference between no version and an existing version

```suggestion
		logger.V(1).Info("Skipping Flux upgrades, already installed latest version")
```

danbudris (approved):
approved, one thought and a nit

pkg/workflows/upgrade.go
jiayiwang7: reason not updating other methods in this step is bc all other methods return `changeDiff, err` instead of just err

pkg/addonmanager/addonclients/upgrader.go
jiayiwang7: i updated to
```
logger.V(1).Info("Skipping Flux upgrades, no previous flux installed in the cluster")
```
hopefully it's less confusing

mitalipaygude (approved):

PR #2761, opened 2022-07-19
Add get packagebundlecontroller command

TerryHowe:


cmd/eksctl-anywhere/cmd/getpackagebundlecontroller.go
```diff
@@ -0,0 +1,28 @@
...
+func init() {
+	getCmd.AddCommand(getPackageBundleControllerCommand)
+	getPackageBundleControllerCommand.Flags().StringVarP(&gpbco.output, "output", "o", "", "Specifies the output format (valid option: json, yaml)")
+}
+
+var getPackageBundleControllerCommand = &cobra.Command{
+	Use:          "packagebundlecontroller(s) [flags]",
+	Aliases:      []string{"packagebundlecontroller", "packagebundlcontrolleres", "pbc"},
+	Short:        "Get packagebundlecontroller(s)",
+	Long:         "This command is used to display the currently packagebundlecontrollers",
```
ivyostosh, line 22: currently active packagebundlecontrollers?

TerryHowe:
```suggestion
	Long:         "This command is used to display the current packagebundlecontrollers",
```

a-cool-train:
Why do we want to expose our internal workings to the client? I think packagebundlecontroller should not be exposed.

TerryHowe:
> Why do we want to expose our internal workings to the client? I think packagebundlecontroller should not be exposed.

If the package controller runs on the management cluster, the PBC will be the only way to tell the active bundle. Each cluster could have a different active bundle.

a-cool-train (approved):

Issue #2710, opened 2022-07-11
Cluster components upgrade can trigger machine rolling upgrade on management cluster itself


jiayiwang7:
**What happened**:

https://github.com/aws/eks-anywhere/issues/2665#issuecomment-1178208038

Upgrading a cloudstack management cluster with latest v0.10.x CLI triggers its control plane and unstacked etcd node machines to rollout and recreate in the management cluster itself, during the core components upgrade, before creating the bootstrap cluster and move CAPI components.

This happens because during management cluster upgrade, EKS-A conducts its core components upgrade before the actual target cluster upgrade. During the core components upgrade, we update EKS-A CRDs and the cluster controller which reconciles the EKS-A spec and converts it into the underlying latest version of CAPI templates using the new controller code, and applies them on the management cluster. This can potentially trigger a rolling upgrade of the nodes.

**What you expected to happen**:

Management cluster's etcd and control plane machine upgrade should not happen on the cluster itself. Instead, we should always move the CAPI components to a separate bootstrap cluster and provision a cluster from there. This is the principle we follow to upgrade management cluster, so that we always have the CAPI management components in the bootstrap cluster even if the target cluster is down and goes into an irreversible/unrecoverable state.

**How to reproduce it (as minimally and precisely as possible)**:

Create a management cluster and a workload cluster with cloudstack provider using EKS-A CLI version v0.8.3. Upgrade the management cluster using EKS-A CLI version v0.10.x. Watch management cluster machines rollout before upgrading the actual cluster.

**Anything else we need to know?**:

Potential solution:

We should pause EKS-A cluster controller reconciliation (https://github.com/aws/eks-anywhere/blob/main/pkg/workflows/upgrade.go#L86) before the core components upgrade (https://github.com/aws/eks-anywhere/blob/main/pkg/workflows/upgrade.go#L203), so that the controller will not apply any CAPI spec and affect the components upgrade process, which the CLI should be in charge of.

We should also add validation checks in the EKS-A cluster controller to never apply management cluster's control plane, and unstacked etcd templates through the controller code, on top of the existing upgrade webhook validations.

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

PR #2767, opened 2022-07-20
Pause reconcile before upgrading core components

taneyland:
*Issue #, if available:*
#2710 

*Description of changes:*
We are pausing reconciliation before the upgrade-core-components task to avoid premature rolling upgrades. See issue above for more details.

We also added a validation in the controller to never apply the CP and etcd templates on the management cluster through the controller.

*Testing (if applicable):*

### DIFF
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -13,8 +13,8 @@ import (
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
-	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/features"
+	"github.com/aws/eks-anywhere/pkg/providers/common"
 	anywhereTypes "github.com/aws/eks-anywhere/pkg/types"
 )
 
@@ -168,7 +168,7 @@ func (cor *clusterReconciler) applyTemplates(ctx context.Context, cs *anywherev1
 	for _, resource := range resources {
 		kind := resource.GetKind()
 		name := resource.GetName()
-		if cs.IsSelfManaged() && (strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, constants.ControlPlaneTemplatePrefix)) || strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, constants.EtcdTemplatePrefix))) {
+		if cs.IsSelfManaged() && (strings.HasPrefix(name, common.CPMachineTemplateBase(cs.Name)) || strings.HasPrefix(name, common.EtcdMachineTemplateBase(cs.Name))) {
 			continue
 		}
 		cor.Log.Info("applying object", "kind", kind, "name", name, "dryRun", dryRun)
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -3,6 +3,7 @@ package resource
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -12,6 +13,7 @@ import (
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/features"
 	anywhereTypes "github.com/aws/eks-anywhere/pkg/types"
 )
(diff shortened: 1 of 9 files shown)


controllers/resource/reconciler.go
```diff
@@ -159,13 +165,16 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
...
 }
 
-func (cor *clusterReconciler) applyTemplates(ctx context.Context, resources []*unstructured.Unstructured, dryRun bool) error {
+func (cor *clusterReconciler) applyTemplates(ctx context.Context, cs *anywherev1.Cluster, resources []*unstructured.Unstructured, dryRun bool) error {
 	for _, resource := range resources {
 		kind := resource.GetKind()
 		name := resource.GetName()
+		if cs.IsSelfManaged() && (strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, cpTemplatePrefix)) || strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, etcdTemplatePrefix))) {
+			continue
+		}
```
maxdrib, line 177: wouldn't this prevent the cp template and etcd templates from ever getting updated? How do we update them? Where in the process would that happen? I'm guessing it'll happen when it's on the bootstrap cluster as part of upgrade, I'm just a bit confused where

Let's also make sure the new logic is unit tested

controllers/resource/reconciler.go
```diff
@@ -16,6 +17,11 @@ import (
 	anywhereTypes "github.com/aws/eks-anywhere/pkg/types"
 )
 
+const (
+	cpTemplatePrefix   = "-control-plane-template"
+	etcdTemplatePrefix = "-etcd-template"
```
maxdrib, line 22: should we also use these constants in https://github.com/aws/eks-anywhere/blob/main/pkg/providers/common/common.go?

maxdrib (commented):
Great to see this is being fixed! Has this been tested e2e in a cloudstack environment?

controllers/resource/reconciler.go
taneyland: After we generate the new capi spec & apply it, these templates should get updated. [Here](https://github.com/aws/eks-anywhere/blob/main/pkg/clustermanager/cluster_manager.go#L364) we are calling apply on the control plane capi content, which includes the templates for both the control plane and etcd.

& ok yes I'll be sure to add a unit test too

maxdrib (approved):
When testing with v0.8.3 -> latest release upgrade, the rolling upgrade on the management cluster did not start until after everything was moved to the bootstrap cluster.

pkg/providers/common/common.go
```diff
@@ -74,12 +75,12 @@ func GenerateSSHAuthKey(writer filewriter.FileWriter) (string, error) {
 
 func CPMachineTemplateName(clusterName string, now types.NowFunc) string {
```
jiayiwang7, line 76: we can define a new method

```go
CPMachineTemplateBase(clusterName string) string {
  return fmt.Sprintf("%s-control-plane-template", clusterName)
}
```

`CPMachineTemplateName` can use this method, the controller can also use it by `strings.HasPrefix(name, CPMachineTemplateBase)`

Same with Etcd. Preferably we should update worker name as well to follow same pattern.

controllers/resource/reconciler.go
```diff
@@ -159,13 +161,16 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
...
-	return cor.applyTemplates(ctx, resources, dryRun)
+	return cor.applyTemplates(ctx, cs, resources, dryRun)
 }
 
-func (cor *clusterReconciler) applyTemplates(ctx context.Context, resources []*unstructured.Unstructured, dryRun bool) error {
+func (cor *clusterReconciler) applyTemplates(ctx context.Context, cs *anywherev1.Cluster, resources []*unstructured.Unstructured, dryRun bool) error {
 	for _, resource := range resources {
 		kind := resource.GetKind()
 		name := resource.GetName()
+		if cs.IsSelfManaged() && (strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, constants.ControlPlaneTemplatePrefix)) || strings.HasPrefix(name, fmt.Sprintf("%s%s", cs.Name, constants.EtcdTemplatePrefix))) {
```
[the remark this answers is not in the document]
jiayiwang7: Can we add unit tests?

pkg/workflows/upgrade.go
```diff
@@ -590,20 +599,23 @@ func (s *resumeFluxReconcile) Run(ctx context.Context, commandContext *task.Comm
 		commandContext.SetError(err)
 		return &writeClusterConfigTask{}
 	}
+	if !s.eksaSpecDiff {
```
jiayiwang7, line 602: what's the use case of this? if eksa spec is different, we skip writeclusterconfig and delete bootstrap?

taneyland: If there is no difference in the eksaSpec, then we skip writeclusterconfig and delete bootstrap. Since we pause reconciliation before we decide whether or not the spec changed, we have to resume reconciliation if it didn't change & we wouldn't have to write a new cluster config & there would be no new bootstrap cluster to delete

jiayiwang7 (approved):

PR #2768, opened 2022-07-20
Add containers volume option in snow machine config

jiayiwang7:
*Issue #, if available:*

#1105 

*Description of changes:*

`containerVolumes` allow user to add extra container root volumes to the node instance.

*Testing (if applicable):*

Unit tests


designs/snow-integration.md
```diff
@@ -60,6 +60,9 @@ type SnowMachineConfigSpec struct {
 	// Devices contains a device ip list assigned by the user to provision machines.
 	Devices []string `json:"devices,omitempty"`
 
+	// ContainersVolume provides the configuration options for the containers data storage volumes.
```
g-gaston, line 63:
```suggestion
	// ContainersVolume provides the configuration options for the containers data storage volume.
```

pkg/api/v1alpha1/snowmachineconfig_types.go
```diff
@@ -38,6 +42,9 @@ type SnowMachineConfigSpec struct {
 
 	// Devices contains a device ip list assigned by the user to provision machines.
 	Devices []string `json:"devices,omitempty"`
+
+	// ContainersVolume provides the configuration options for the containers data storage volumes.
```
g-gaston, line 46:
```suggestion
	// ContainersVolume provides the configuration options for the containers data storage volume.
```

pkg/api/v1alpha1/snowmachineconfig_types.go
```diff
@@ -38,6 +42,9 @@ type SnowMachineConfigSpec struct {
...
+	ContainersVolume *snowv1.Volume `json:"containersVolume,omitempty"`
```
g-gaston, line 47: we just have to be careful with this so we don't update our API by accident when updating the dependency on the snow module

Maybe it would be wise to define our own volume struct

pkg/providers/snow/apibuilder.go
```diff
@@ -45,6 +45,12 @@ func KubeadmControlPlane(clusterSpec *cluster.Spec, snowMachineTemplate *snowv1.
 		fmt.Sprintf("/etc/eks/bootstrap.sh %s %s", clusterSpec.VersionsBundle.Snow.KubeVip.VersionedImage(), clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host),
 	)
 
+	if snowMachineTemplate.Spec.Template.Spec.ContainersVolume != nil {
+		kcp.Spec.KubeadmConfigSpec.PreKubeadmCommands = append(kcp.Spec.KubeadmConfigSpec.PreKubeadmCommands,
```
g-gaston, line 49: I was under the impression that we only wanted to support this for worker nodes and not for CP

jiayiwang7: i confirmed w snow team, cp volume is also supported

designs/snow-integration.md
jiayiwang7: We chatted offline. Will keep the volume type from upstream. Whenever there is API change upstream, it should be a breaking change and we will handle it on our end too.

pkg/api/v1alpha1/snowmachineconfig_types.go
jiayiwang7: We chatted offline. Will keep the volume type from upstream. Whenever there is API change upstream, it should be a breaking change and we will handle it on our end too.

g-gaston (approved):

PR #2771, opened 2022-07-20
Update the cloudstack templates to support failure domains and CAPC v1beta2 API

wongni:
*Issue #, if available:*

Description of changes:
Updated the cloudstack CP and MD templates to support failure domains and v1beta2 API

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/cloudstack.go
+++ b/pkg/cluster/cloudstack.go
@@ -18,6 +18,9 @@ func cloudstackEntry() *ConfigManagerEntry {
 		},
 		Defaulters: []Defaulter{
 			func(c *Config) error {
+				if c.CloudStackDatacenter != nil {
+					c.CloudStackDatacenter.SetDefaults()
+				}
 				return nil
 			},
 		},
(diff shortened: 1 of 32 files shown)


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -782,9 +784,10 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec, datacenterConfigSpec v1alpha1
 
 		// Add no-proxy defaults
 		noProxyList = append(noProxyList, clusterapi.NoProxyDefaults()...)
-		cloudStackManagementApiEndpointHostname, err := getHostnameFromUrl(datacenterConfigSpec.ManagementApiEndpoint)
-		if err == nil {
-			noProxyList = append(noProxyList, cloudStackManagementApiEndpointHostname)
+		for _, az := range datacenterConfigSpec.AvailabilityZones {
```
jiayiwang7, line 787: Can we have one single proxy setup method which fills all the proxy configs? it can be used in both `buildTemplateMapCP` and `buildTemplateMapMD`. Looks like a lot of duplicate code here: line 788 vs 875

pkg/cluster/cloudstack.go
```diff
@@ -18,6 +18,9 @@ func cloudstackEntry() *ConfigManagerEntry {
 		},
 		Defaulters: []Defaulter{
 			func(c *Config) error {
+				if c.CloudStackDatacenter != nil {
+					c.CloudStackDatacenter.SetDefaults()
```
jiayiwang7, line 22: do we have unit test for this? if not can we add

jiayiwang7 (approved):

PR #2779, opened 2022-07-20
Feature config validation

a-cool-train:
*Description of changes:*
- Configuration structure has changed in bundles. Each package bundle currently contains a base64 encoded string of configurations instead of the plain text originally used in the package bundle.

*Testing (if applicable):*


a-cool-train:
This PR is no longer required since we won't be supporting defaulting configurations.

PR #2781, opened 2022-07-21
Add a collector and an analyzer into diagnostic bundle to detect Control Plane IP problem

YoyoTT:
*Issue #, if available:*

Create a vsphere cluster with a controlPlaneIP that is not in the subnet range of the physical network. The bootstrap cluster will fail to connect to the newly created API server and stall.

*Description of changes:*

- Adding a collector that will deploy a pod to ping control plane with the given IP(in the cluster-definition.yaml), and output of ping with an exit code(0 is fine, other# will be a problem) would indicate whether the API server is up or the controlPlane IP is bad. We already have an analyzer to analyze API server health condition. So, by combining the two results, we are able to know the reason of failure connection between bootstrap cluster and API server is due to a bad control plane IP. 

- Also, adding a new analyzer to generate message and show us the diagnostic results for reference. 

*Testing (if applicable):*


pkg/diagnostics/analyzers.go
```diff
@@ -220,22 +220,22 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
...
 	return []*Analyze{
 		{
 			TextAnalyze: &textAnalyze{
 				analyzeMeta: analyzeMeta{
 					CheckName: fmt.Sprintf("%s: API server pod missing. Log: %s", logAnalysisAnalyzerPrefix, fullManagerPodLogPath),
 				},
-				CollectorName: constants.CapiKubeadmControlPlaneSystemNamespace,
-				FileName:      capiCpManagerContainerLogFile,
-				RegexPattern:  `machine (.*?) reports APIServerPodHealthy condition is false \(Error, Pod kube-apiserver-(.*?) is missing\)`,
+				FileName:     fullManagerPodLogPath,
```
mitalipaygude, line 237: Wasn't this bug already fixed and merged? are these changes rebase from main?

pkg/diagnostics/analyzers.go
```diff
@@ -220,22 +220,22 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
...
-	capiCpManagerContainerLogFile := path.Join(capiCpManagerPod, "manager.log")
+	capiCpManagerContainerLogFile := capiCpManagerPod + ".log"
 	fullManagerPodLogPath := path.Join(logpath(constants.CapiKubeadmControlPlaneSystemNamespace), capiCpManagerContainerLogFile)
 	return []*Analyze{
 		{
 			TextAnalyze: &textAnalyze{
 				analyzeMeta: analyzeMeta{
 					CheckName: fmt.Sprintf("%s: API server pod missing. Log: %s", logAnalysisAnalyzerPrefix, fullManagerPodLogPath),
 				},
-				CollectorName: constants.CapiKubeadmControlPlaneSystemNamespace,
```
danbudris, line 236: why remove the collector name here?

pkg/diagnostics/analyzers.go
```diff
@@ -255,6 +255,40 @@ func (a *analyzerFactory) capiKubeadmControlPlaneSystemLogAnalyzers() []*Analyze
...
+				analyzeMeta: analyzeMeta{
+					CheckName: fmt.Sprintf("%s:  Destination Host Unreachable. Log: %s", logAnalysisAnalyzerPrefix, fullRunPingPodLogPath),
+				},
+				FileName:     fullRunPingPodLogPath,
+				RegexPattern: `exit code: 0`,
+				Outcomes: []*outcome{
+					{
+						Fail: &singleOutcome{
+							When:    "false",
+							Message: fmt.Sprintf("Pod failed to launch, API server is missing or the route to API Server is unvalid. See %s", fullRunPingPodLogPath),
```
mitalipaygude, line 276: for exit code 0, that comes from pinging the control plane IP right? In that case the failure message seems misleading where we say pod failed to launch. The temporary pod launched and did ping the ip. We just indicate instead that the IP wasn't reachable/had issues.

pkg/diagnostics/collectors.go
```diff
@@ -380,6 +380,56 @@ func (c *collectorFactory) crdCollector(crdType string) *Collect {
...
+	// collectors = append(collectors, c.pingLogCollectors()...)
+	return collectors
+}
+
+func (c *collectorFactory) generateRunPodsCollector(controlPlaneIP string) []*Collect {
+	return c.runPodsCollectors(controlPlaneIP)
+}
+
+func (c *collectorFactory) runPodsCollectors(controlPlaneIP string) []*Collect {
+	tempIPRequest := "yum install -y nmap && for port in 6443 22; do nc -z -v -w5 " + controlPlaneIP + " $port; done"
```
danbudris, line 396: nit: use `fmt.Sprintf` instead of concatenation

mitalipaygude, line 396: Any reason we don't run these arguments at once? as opposed to creating 2 requests?

pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -145,6 +152,11 @@ func TestGenerateBundleConfigWithOidc(t *testing.T) {
 			TypeMeta:   metav1.TypeMeta{},
 			ObjectMeta: metav1.ObjectMeta{},
 			Spec: eksav1alpha1.ClusterSpec{
+				ControlPlaneConfiguration: eksav1alpha1.ControlPlaneConfiguration{
+					Endpoint: &eksav1alpha1.Endpoint{
+						Host: "195.16.99.76",
```
mitalipaygude, line 157: nit: the test ips can be 1.1.1.1 for unit tests.

pkg/diagnostics/collectors.go
danbudris, line 396: rather than run `yum install` here, we can use `yum` to install `nmap` directly on the base image that is used for the Diagnostic Collector image.

Check out the dockerfile here:

https://github.com/aws/eks-anywhere-build-tooling/blob/a27f3bd332d5374b4519ed48aedc2144c107996c/projects/aws/eks-anywhere/docker/linux/eksa-diagnostic-collector/Dockerfile

Currently it only has tar installed but we could also install NMAP as well; it seems like it would be generically useful.

Let me know if you need any assistance here.

pkg/diagnostics/analyzers.go
danbudris, line 276:
```suggestion
							Message: fmt.Sprintf("API server is unavailable or the route to API Server is invalid. See %s", fullRunPingPodLogPath),
```

pkg/diagnostics/analyzers.go
```diff
@@ -255,6 +255,40 @@ func (a *analyzerFactory) capiKubeadmControlPlaneSystemLogAnalyzers() []*Analyze
...
+						},
+					},
+
+					{
+						Pass: &singleOutcome{
+							When:    "true",
+							Message: "Control plane IP is verified, cleaning up pod...",
```
danbudris, line 283:
```suggestion
							Message: "Control plane IP verified",
```

pkg/diagnostics/collectors.go
```diff
@@ -380,6 +380,56 @@ func (c *collectorFactory) crdCollector(crdType string) *Collect {
 	}
 }
 
+// APIServerCollector collect connection info when running a pod on an existing cluster
+func (c *collectorFactory) APIServerCollector(controlPlaneIP string) []*Collect {
+	var collectors []*Collect
+	collectors = append(collectors, c.generateRunPodsCollector(controlPlaneIP)...)
+	// collectors = append(collectors, c.pingLogCollectors()...)
```
danbudris, line 387:
```suggestion
```

pkg/diagnostics/collectors.go
```diff
@@ -380,6 +380,56 @@ func (c *collectorFactory) crdCollector(crdType string) *Collect {
...
+	return collectors
+}
+
+func (c *collectorFactory) generateRunPodsCollector(controlPlaneIP string) []*Collect {
+	return c.runPodsCollectors(controlPlaneIP)
+}
+
+func (c *collectorFactory) runPodsCollectors(controlPlaneIP string) []*Collect {
```
danbudris, line 395: Can I recommend that we break this up into two, separate generic collector methods that can be more easily reused in other use cases?

for example:

`pingHostCollector(hostIP string) *Collect`
`hostPortsCollector([]string ports, hostIP string) *Collect`

and then return these as part of a parent method, such as:

`controlPlaneNetworkPathCollector(controlPlaneIP string) []*Collect`

that way we can re-use the functionality for purposes other than just the CP IP availability.

pkg/diagnostics/collectors.go
```diff
@@ -380,6 +380,56 @@ func (c *collectorFactory) crdCollector(crdType string) *Collect {
...
+	tempIPRequest := "yum install -y nmap && for port in 6443 22; do nc -z -v -w5 " + controlPlaneIP + " $port; done"
+	argsIP := []string{tempIPRequest}
+	tempPingRequest := "yum install -y iputils && ping -c5 " + controlPlaneIP + "; echo exit code: $?"
```
danbudris, line 398: same aobut installation of `iputils` as `nmap` -- let's install these directly on the diagnostic collector base image.

danbudris (commented):
This is great, Yue, and will help our customers a lot.

I've left a few comments about the structure and the installation of dependencies on the collector image. Please let me know if I can help.

pkg/diagnostics/analyzers.go
mitalipaygude: I believe this is from the bug Yue ended up fixing. We get the analyzer data from logs, and not from a collector. This was causing the bug where the analyzer was looking for the log file within a collector instead of under logs folder.

pkg/diagnostics/collectors.go
mitalipaygude: I agree with this approach.

YoyoTT:
> This is great, Yue, and will help our customers a lot.
> 
> I've left a few comments about the structure and the installation of dependencies on the collector image. Please let me know if I can help.

 I am working on it! Appreciate Dan!

pkg/diagnostics/collectors.go
mitalipaygude: https://github.com/aws/eks-anywhere-build-tooling/pull/1077

abhinavmpandey08, line 396: Why not use a Golang net library instead of exec'ing nc? Here's an example of how we use it currently https://github.com/aws/eks-anywhere/blob/main/pkg/networkutils/networkutils.go#L32-L44

pkg/diagnostics/analyzers.go
```diff
@@ -254,6 +255,40 @@ func (a *analyzerFactory) capiKubeadmControlPlaneSystemLogAnalyzers() []*Analyze
...
+				analyzeMeta: analyzeMeta{
+					CheckName: fmt.Sprintf("%s:  Destination Host Unreachable. Log: %s", logAnalysisAnalyzerPrefix, fullRunPingPodLogPath),
+				},
+				FileName:     fullRunPingPodLogPath,
+				RegexPattern: `exit code: 0`,
+				Outcomes: []*outcome{
+					{
+						Fail: &singleOutcome{
+							When:    "false",
+							Message: fmt.Sprintf("API server is unavailable or the route to API Server is invalid. See %s", fullRunPingPodLogPath),
```
mitalipaygude, line 276: Let's be more specific and say the control plane endpoint host is unreachable instead.

pkg/diagnostics/collectors.go
jaxesn: this is running in the container in the pod

abhinavmpandey08: So is the container going to just run these commands directly through a shell script or something?
Or is there a custom go binary that it's going to execute?
I was thinking it was the latter but if it isn't then ignore my comment

YoyoTT (commented):
/approve

danbudris (approved):

PR #2787, opened 2022-07-22
Cluster webhook validation for create

mitalipaygude:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1090

Description of changes:
Add cluster validations that we use in the CLI to the cluster webhook for creation on CRs.

Testing (if applicable):

- Unit tests
- Manual testing with tilt

-By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -16,9 +16,13 @@ package v1alpha1
 
 import (
 	"fmt"
+	"net/http"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -54,19 +58,21 @@ var _ webhook.Validator = &Cluster{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *Cluster) ValidateCreate() error {
 	clusterlog.Info("validate create", "name", r.Name)
-	if r.IsReconcilePaused() {
-		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
-		return nil
-	}
-	if !features.IsActive(features.FullLifecycleAPI()) {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
-	}
-	if r.IsSelfManaged() {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+	var allErrs []error
+	if err := r.Validate(); err != nil {
+		allErrs = append(allErrs, fmt.Errorf("cluster is not valid: %v ", err))
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if !r.IsReconcilePaused() {
+		if r.IsSelfManaged() {
+			allErrs = append(allErrs, fmt.Errorf("creating new cluster on existing cluster is not supported for self managed clusters"))
+		} else if !features.IsActive(features.FullLifecycleAPI()) {
+			allErrs = append(allErrs, fmt.Errorf("creating new managed cluster on existing cluster is not supported"))
+		}
+	}
+
+	if len(allErrs) > 0 {
+		return BuildStatusError(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
 	}
 
 	return nil
@@ -218,4 +224,31 @@ func (r *Cluster) ValidateDelete() error {
 	clusterlog.Info("validate delete", "name", r.Name)
 
 	return nil
+}
+
+func BuildStatusError(qualifiedKind schema.GroupKind, name string, errs []error) *apierrors.StatusError {
+	causes := make([]metav1.StatusCause, 0, len(errs))
+	for _, err := range errs {
+		causes = append(causes, metav1.StatusCause{
+			Message: err.Error(),
+		})
+	}
+
+	err := &apierrors.StatusError{ErrStatus: metav1.Status{
+		Status: metav1.StatusFailure,
+		Code:   http.StatusUnprocessableEntity,
+		Reason: metav1.StatusReasonInvalid,
+		Details: &metav1.StatusDetails{
+			Group:  qualifiedKind.Group,
+			Kind:   qualifiedKind.Kind,
+			Name:   name,
+			Causes: causes,
+		},
+	}}
+	aggregatedErrs := utilerrors.NewAggregate(errs)
+
+	if aggregatedErrs != nil {
+		err.ErrStatus.Message = fmt.Sprintf("%s %q is invalid: %v", qualifiedKind.String(), name, aggregatedErrs)
+	}
+	return err
 }
--- a/pkg/api/v1alpha1/cluster_webhook_test.go
+++ b/pkg/api/v1alpha1/cluster_webhook_test.go
@@ -1,11 +1,13 @@
 package v1alpha1_test
 
 import (
+	"fmt"
 	"testing"
 
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	_ "k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
@@ -36,14 +38,7 @@ func TestClusterValidateUpdateManagementValueImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementOldNilNewTrueSuccess(t *testing.T) {
-	cOld := &v1alpha1.Cluster{
-		Spec: v1alpha1.ClusterSpec{
-			WorkerNodeGroupConfigurations: []v1alpha1.WorkerNodeGroupConfiguration{{
-				Count: 1,
-				Name:  "test",
-			}},
-		},
-	}
+	cOld := createCluster()
 	c := cOld.DeepCopy()
 	c.SetSelfManaged()
 
@@ -61,14 +56,7 @@ func TestClusterValidateUpdateManagementOldNilNewFalseImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementBothNilImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{
-		Spec: v1alpha1.ClusterSpec{
-			WorkerNodeGroupConfigurations: []v1alpha1.WorkerNodeGroupConfiguration{{
-				Count: 1,
-				Name:  "test",
-			}},
-		},
-	}
+	cOld := createCluster()
 	c := cOld.DeepCopy()
 
 	g := NewWithT(t)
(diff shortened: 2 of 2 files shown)


pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1152,134 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+}
+
+func TestClusterValidateCreateActiveFeatureWorkload(t *testing.T) {
+	features.ClearCache()
+	cOld := createCluster()
+	g := NewWithT(t)
+	cOld.SetManagedBy("management-cluster")
+	validateCreateError := cOld.ValidateCreate()
+	g.Expect(validateCreateError).NotTo(Succeed())
+	g.Expect(validateCreateError.Error()).To(ContainSubstring("cluster on existing cluster is not supported"))
```
g-gaston, line 1196: You can do this with one expect

```suggestion
	g.Expect(cOld.ValidateCreate()).To(MatchError(ContainSubstring("cluster on existing cluster is not supported")))
```

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1152,134 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+	g := NewWithT(t)
+	cOld.SetManagedBy("mgmt")
+	cOld.Annotations = map[string]string{}
+	cOld.Annotations[pausedAnnotation] = "true"
+	g.Expect(cOld.ValidateCreate()).To(Succeed())
+}
+
+func TestClusterValidateCreateActiveFeatureWorkload(t *testing.T) {
+	features.ClearCache()
+	cOld := createCluster()
```
g-gaston, line 1191: nit: I'm not sure `cOld` make a lot of sense, there is no old vs new here
Maybe just `cluster`

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1152,134 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+	features.ClearCache()
+	t.Setenv(features.FullLifecycleAPIEnvVar, "true")
+	cOld := createCluster()
+	cOld.SetManagedBy("management-cluster")
+
+	g := NewWithT(t)
+	g.Expect(cOld.ValidateCreate()).To(Succeed())
+}
+
+func TestClusterValidateCreateFail(t *testing.T) {
```
g-gaston, line 1219:
```suggestion
func TestClusterValidateCreateInvalidCluster(t *testing.T) {
```

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1152,134 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+	cOld := createCluster()
+	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{}
+	cOld.SetManagedBy("management-cluster")
+
+	g := NewWithT(t)
+	g.Expect(cOld.ValidateCreate()).NotTo(Succeed())
+	strings.Contains(cOld.ValidateCreate().Error(), "control plane node count must be positive")
+}
+
+func TestClusterValidateCreatePassSelfManagedFail(t *testing.T) {
```
g-gaston, line 1231:
```suggestion
func TestClusterValidateCreateSelfManagedNotPaused(t *testing.T) {
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,21 +55,26 @@ var _ webhook.Validator = &Cluster{}
...
 	if r.IsSelfManaged() {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+		allErrs = append(allErrs, fmt.Errorf("creating new cluster on existing cluster is not supported for self managed clusters"))
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return apierrors.NewBadRequest(fmt.Sprintf("invalid cluster spec: %v", aggregate.Error()))
```
g-gaston, line 76: can we just pass `aggregate.Error()`?
`invalid cluster spec` is not always accurate

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,21 +55,26 @@ var _ webhook.Validator = &Cluster{}
...
 func (r *Cluster) ValidateCreate() error {
 	clusterlog.Info("validate create", "name", r.Name)
+	var allErrs []error
+	if err := r.Validate(); err != nil {
+		allErrs = append(allErrs, err)
+	}
+
 	if r.IsReconcilePaused() {
 		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
 		return nil
```
g-gaston, line 65: This is wrong
It means that even if `r.Validate()` fails, we will accept the request

Which makes me think we are missing a set of tests

I would advise to write the logic in a way where we only return at the end of the method. Having errors accumulating in a slice but also intertwined with early returns makes the code more difficult to read.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,22 @@ var _ webhook.Validator = &Cluster{}
...
 	clusterlog.Info("validate create", "name", r.Name)
-	if r.IsReconcilePaused() {
-		clusterlog.Info("cluster is paused, so allowing create", "name", r.Name)
-		return nil
-	}
-	if !features.IsActive(features.FullLifecycleAPI()) {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+	var allErrs []error
+	if err := r.Validate(); err != nil {
+		allErrs = append(allErrs, fmt.Errorf("invalid cluster: %v ", err))
```
taneyland, line 60: Do we want to append all errors from the `Validate()` method? Because it seems that it will return after the first error & only append that one error to `allErrs`

mitalipaygude: Yes we only append the first error from Validate. Appending everything from Validate would require changes in the Validate method itself, which I believe are not exactly in scope for this task. We could explore it though in the future.

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1149,242 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+	}
+	cluster.SetManagedBy("management-cluster")
+	c := cluster.DeepCopy()
+	c.Spec.ControlPlaneConfiguration.Labels["test"] = "val1/val2"
+
+	g := NewWithT(t)
+	g.Expect(c.ValidateUpdate(cluster)).NotTo(Succeed())
+}
+
+func TestClusterValidateCreateSelfManagedUnpause(t *testing.T) {
```
g-gaston, line 1166:
```suggestion
func TestClusterValidateCreateSelfManagedUnpaused(t *testing.T) {
```

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1149,242 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+			features.ClearCache()
+			if tt.featureGateEnabled {
+				t.Setenv(features.FullLifecycleAPIEnvVar, "true")
+			}
+
+			// Invalid control plane configuration
+			tt.cluster.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{}
+
+			g := NewWithT(t)
+			g.Expect(tt.cluster.ValidateCreate()).To(MatchError(ContainSubstring("invalid cluster")))
```
g-gaston, line 1274: nit: why don't we look for the specific error we get with an invalid control plane? It seems a bit more reliable

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +55,22 @@ var _ webhook.Validator = &Cluster{}
...
+		} else if !features.IsActive(features.FullLifecycleAPI()) {
+			allErrs = append(allErrs, fmt.Errorf("creating new managed cluster on existing cluster is not supported"))
+		}
 	}
 
-	if err := validateCNIPlugin(r.Spec.ClusterNetwork); err != nil {
-		return apierrors.NewBadRequest(err.Error())
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return apierrors.NewBadRequest(aggregate.Error())
```
g-gaston, line 73: What would you think about doing something similar to this:
https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L292

mitalipaygude: That looks like a good idea.

We would need to combine the [NewInvalid](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L282) and the [BadRequest](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L313) errors into the Status Error, which the library does not support at the moment. We could write our own logic for this, and we might use it for future PRs for returning similar errors from webhooks as well. 

Created this task https://github.com/aws/eks-anywhere/issues/2813.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -219,3 +225,31 @@ func (r *Cluster) ValidateDelete() error {
 
 	return nil
 }
+
+func buildStatusError(qualifiedKind schema.GroupKind, name string, errs []error) *apierrors.StatusError {
+	causes := make([]metav1.StatusCause, 0, len(errs))
+	for i := range errs {
+		err := errs[i]
```
g-gaston, line 232: any reason to not get the element from the range directly?
```suggestion
	for _, err := range errs {
```

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1442,23 +1149,242 @@ func TestClusterUpdateControlPlaneTaintsAndLabelsSuccess(t *testing.T) {
...
+	features.ClearCache()
+	cluster := createCluster()
+	g := NewWithT(t)
+	cluster.SetSelfManaged()
+	validateCreateError := cluster.ValidateCreate()
+	g.Expect(validateCreateError).NotTo(Succeed())
+	g.Expect(cluster.ValidateCreate()).To(MatchError(ContainSubstring("cluster on existing cluster is not supported")))
+}
+
+func TestClusterValidateCreateManagedUnpause(t *testing.T) {
```
g-gaston, line 1176:
```suggestion
func TestClusterValidateCreateManagedUnpaused(t *testing.T) {
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -219,3 +225,31 @@ func (r *Cluster) ValidateDelete() error {
 
 	return nil
 }
+
+func buildStatusError(qualifiedKind schema.GroupKind, name string, errs []error) *apierrors.StatusError {
```
g-gaston, line 229: this is important/big enough that we might want to test it individually

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -54,19 +58,21 @@ var _ webhook.Validator = &Cluster{}
...
+	var allErrs []error
+	if err := r.Validate(); err != nil {
+		allErrs = append(allErrs, fmt.Errorf("invalid cluster: %v ", err))
 	}
-	if r.IsSelfManaged() {
-		return apierrors.NewBadRequest("Creating new cluster on existing cluster is not supported")
+
+	if !r.IsReconcilePaused() {
+		if r.IsSelfManaged() {
+			allErrs = append(allErrs, fmt.Errorf("creating new cluster on existing cluster is not supported for self managed clusters"))
```
vivek-koppuru, line 68: Why do all these errors need to be appended? From a user perspective, since the apiserver is already doing minimal validations, I would think that if it really isn't supported, we don't need to validate the fields if the user is not allowed to create a new cluster anyways.

vivek-koppuru (approved):
looks good apart from my previous comment about needing to aggregate the errors, but we will be discussing that separately

PR #2792, opened 2022-07-25
E2E test for re-running upgrade command

taneyland:
*Issue #, if available:*

*Description of changes:*
Adds a new env var `EKSA_EXTERNAL_ETCD_TIMEOUT` to configure the timeout when waiting for external etcd to be ready

The new upgrade e2e test uses an invalid resource pool on the first attempt to upgrade the cluster, we then change the resource pool field to valid option & resume upgrade, in which it should now succeed. 

*Testing (if applicable):*

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -14,6 +14,7 @@ const (
 	EksaGitKnownHostsFileEnv    = "EKSA_GIT_KNOWN_HOSTS"
 	SshKnownHostsEnv            = "SSH_KNOWN_HOSTS"
 	EksaReplicasReadyTimeoutEnv = "EKSA_REPLICAS_READY_TIMEOUT"
+	ExternalEtcdTimeoutEnv      = "EKSA_EXTERNAL_ETCD_TIMEOUT"
 )
 
 type CliConfig struct {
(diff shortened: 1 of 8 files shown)


cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -24,6 +24,7 @@ type upgradeClusterOptions struct {
 	forceClean            bool
 	hardwareCSVPath       string
 	tinkerbellBootstrapIP string
+	e2eTimeout            bool
```
vivek-koppuru, line 27: Why do we need to expose this as an `e2e-timeout`? Ideally we wouldn't have any flag that refers to something as e2e, but if this is something that a user can see, can we call it `external-etcd-timeout`, or have an overridable timeout for all waits like `node-wait-timeout`?

vivek-koppuru: If we don't want to make that decision now, I would maybe opt in for this to be an envvar configuration rather than a flag.

pkg/config/config.go
```diff
@@ -36,3 +40,13 @@ func GetMaxWaitPerMachine() time.Duration {
 	}
 	return DefaultMaxWaitPerMachine
 }
+
+func GetExternalEtcdTimeout() string {
+	if env, found := os.LookupEnv(ExternalEtcdTimeoutEnv); found {
+		if _, err := time.ParseDuration(env); err == nil {
+			return env
+		}
+		logger.V(3).Info(fmt.Sprintf("Invalid EKSA_EXTERNAL_ETCD_TIMEOUT value: %s Use the default timeout: %s", env, ExternalEtcdTimeoutEnv))
```
vivek-koppuru, line 49: I know we didn't do this for the others, but can we use the constant defined above here?

test/e2e/upgrade_test.go
```diff
@@ -4,6 +4,8 @@
 package e2e
 
 import (
+	"github.com/aws/eks-anywhere/pkg/config"
```
vivek-koppuru, line 7: This needs to be gofumpt

vivek-koppuru (approved):

PR #2793, opened 2022-07-25
Fix ip conflict issues during tinkerbell e2e tests

gwesterfieldjr:


internal/test/e2e/setup.go
```diff
@@ -136,6 +136,11 @@ func (e *E2ESession) setup(regex string) error {
 		return err
 	}
 
+	ipPool := e.ipPool.ToString()
+	if ipPool != "" {
```
abhinavmpandey08, line 140: Might be an edge case but should we error here if the `ipPool` is empty?

abhinavmpandey08 (approved):
looks good, just 1 minor comment

internal/test/e2e/setup.go
gwesterfieldjr: No because not all providers currently use ip pool. For instance, docker and cloudstack are not using it yet, so it's ok to be empty on those test cases.

abhinavmpandey08: oh right, ignore this comment then. This should be good to merge

PR #2795, opened 2022-07-25
Rename flux package and methods

jiayiwang7:
*Issue #, if available:*

#2755 

*Description of changes:*

Rename Flux package and methods to something simpler and clear. Remove concept "addon" from Flux code.

*Testing (if applicable):*


pkg/gitops/flux/flux.go
```diff
@@ -53,35 +53,35 @@ const (
...
 		retrier:   retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
 	}
 }
 
-// Flux is an interface that abstracts the basic commands of flux executable.
-type Flux interface {
-	// BootstrapToolkitsComponentsGithub bootstraps toolkit components in a GitHub repository.
-	BootstrapToolkitsComponentsGithub(ctx context.Context, cluster *types.Cluster, fluxConfig *v1alpha1.FluxConfig) error
+// Client is an interface that abstracts the basic commands of flux executable.
+type Client interface {
```
mitalipaygude, line 73: nit: should we rename it to FluxClient?

jiayiwang7: I think it's ok since out of this pkg we will call `flux.Client` instead of `flux.FluxClient` which is duplicate

danbudris: package name is Flux, so Client is the go-prefered name in this case

Issue #2614, opened 2022-06-29
Failed creating library for template when there are more than one datacenter in vsphere


jiayiwang7:
**What happened**:

CLI fails when creating a vsphere cluster for vCenter with multiple datacenters, without `template` field specified.

:x: Validation failed  {"validation": "vsphere Provider setup is valid", "error": "failed setting default values for vsphere machine configs: failed creating library for new template: creating library eks-a-templates: govc: please specify a datacenter\n", "remediation": ""}

**What you expected to happen**:

CLI to successfully create library.

**How to reproduce it (as minimally and precisely as possible)**:

1. Not specifying `template` field in `vspheremachineconfig`
2. vCenter with more than one datacenters and the user has access to all of them.

Run `eksctl anywhere create cluster` cmd

**Anything else we need to know?**:

Cause: https://github.com/aws/eks-anywhere/blob/main/pkg/executables/govc.go#L287-L292

When user does not specify the template, EKS-A creates a library and then imports template from upstream. `CreateLibrary` method does not specify which datacenter to use, thus preflight validation fails when there are multiple datacenters.

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

vivek-koppuru:
Might be as simple as adding a `GOVC_DATACENTER` env var to our commands, trying that as a potential solution

jaxesn:
Can we use the datacenter from the spec instead of introducing a new env var?

vivek-koppuru:
This is regarding the env var to pass into `govc`, not a new user specified env var. We would populate that env var from what you are suggesting.

PR #2796, opened 2022-07-26
Add datacenter to govc env vars

vivek-koppuru:
*Issue #, if available:*
#2614 

*Description of changes:*
Specify the datacenter as an env var to avoid having govc infer based on the datacenters it is able to retrieve.

*Testing (if applicable):*
unit test and e2e

### DIFF
--- a/internal/test/e2e/testRunner.go
+++ b/internal/test/e2e/testRunner.go
@@ -26,6 +26,7 @@ const (
 	govcPasswordKey            string = "GOVC_PASSWORD"
 	govcURLKey                 string = "GOVC_URL"
 	govcInsecure               string = "GOVC_INSECURE"
+	govcDatacenterKey          string = "GOVC_DATACENTER"
 	ssmActivationCodeKey       string = "ssm_activation_code"
 	ssmActivationIdKey         string = "ssm_activation_id"
 	ssmActivationRegionKey     string = "ssm_activation_region"
@@ -133,6 +134,11 @@ func (v *VSphereTestRunner) setEnvironment() (map[string]string, error) {
 		return nil, fmt.Errorf("unable to set %s: %v", govcURLKey, err)
 	}
 	envMap[govcInsecure] = strconv.FormatBool(v.Insecure)
+
+	if err := os.Setenv(govcDatacenterKey, v.Datacenter); err != nil {
+		return nil, fmt.Errorf("unable to set %v: %v", govcDatacenterKey, err)
+	}
+	envMap[govcDatacenterKey] = v.Datacenter
 
 	v.envMap = envMap
 	return envMap, nil
--- a/pkg/executables/govc.go
+++ b/pkg/executables/govc.go
@@ -30,6 +30,7 @@ const (
 	govcPasswordKey      = "GOVC_PASSWORD"
 	govcURLKey           = "GOVC_URL"
 	govcInsecure         = "GOVC_INSECURE"
+	govcDatacenterKey    = "GOVC_DATACENTER"
 	govcTlsHostsFile     = "govc_known_hosts"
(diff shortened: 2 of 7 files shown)


abhinavmpandey08 (approved):

pkg/executables/govc.go
```diff
@@ -533,6 +534,9 @@ func (g *Govc) validateAndSetupCreds() (map[string]string, error) {
 	} else if govcURL, ok := os.LookupEnv(govcURLKey); !ok || len(govcURL) <= 0 {
 		return nil, fmt.Errorf("%s is not set or is empty: %t", govcURLKey, ok)
 	}
+	if govcDatacenter, ok := os.LookupEnv(govcDatacenterKey); !ok || len(govcDatacenter) <= 0 {
+		return nil, fmt.Errorf("%s is not set or is empty: %t", govcDatacenterKey, ok)
```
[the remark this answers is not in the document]
vivek-koppuru: Will do, wanted to see what the coverage was

TerryHowe (approved):
/lgtm

gwesterfieldjr:
This will break tinkerbell e2e tests as we use govc.deployTemplate to deploy the test runner vm in the lab. With this change the new dc env var will be required and cause it to fail since we are not setting it.

gwesterfieldjr (changes requested):
See my previous comment...

gwesterfieldjr:
https://github.com/aws/eks-anywhere/blob/main/internal/test/e2e/testRunner.go#L107 update this to ensure the new datacenter env var gets set to v.Datacenter

vivek-koppuru:
@gwesterfieldjr Good catch, made the changes now

internal/test/e2e/testRunner.go
```diff
@@ -134,6 +135,10 @@ func (v *VSphereTestRunner) setEnvironment() (map[string]string, error) {
 	}
 	envMap[govcInsecure] = strconv.FormatBool(v.Insecure)
 
+	if err := os.Setenv(govcDatacenterKey, v.Datacenter); err != nil {
```
gwesterfieldjr, line 138: Actually, all you need is to set the value in the map.

`envMap[govcDatacenterKey] = v.Datacenter`

gwesterfieldjr: Just FYI, this map gets passed directly to the govc container environment: https://github.com/aws/eks-anywhere/blob/5653615245e6bf719ff2057f23200ee8d4d57037/pkg/executables/govc.go#L490

I pass it directly to ensure it picks up the vpshere info for the lab rather than vmc. Otherwise if we are all setting/referencing the same default env vars there will be race conditions during the build since the test runners are created on same CodeBuild agent machine. This is easier to explain via voice so I'm open to chat if you have questions.

pkg/executables/govc.go
```diff
@@ -533,6 +534,9 @@ func (g *Govc) validateAndSetupCreds() (map[string]string, error) {
 	} else if govcURL, ok := os.LookupEnv(govcURLKey); !ok || len(govcURL) <= 0 {
 		return nil, fmt.Errorf("%s is not set or is empty: %t", govcURLKey, ok)
 	}
+	if govcDatacenter, ok := os.LookupEnv(govcDatacenterKey); !ok || len(govcDatacenter) <= 0 {
```
gwesterfieldjr, line 537: During e2e this will fail for vsphere tests as well since we are not setting this envar in code build. You either need to update code build job file to set it or follow pattern for other vars where we read the eksa value and set it.

internal/test/e2e/testRunner.go
vivek-koppuru: Forgot about this sorry

gwesterfieldjr (approved):

PR #2806, opened 2022-07-27
Add create validations to OIDC webhook

taneyland:
*Issue #, if available:*

*Description of changes:*
Adding validations in oidc webhook

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/oidcconfig.go
+++ b/pkg/api/v1alpha1/oidcconfig.go
@@ -3,6 +3,9 @@ package v1alpha1
 import (
 	"fmt"
 	"net/url"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 const OIDCConfigKind = "OIDCConfig"
@@ -12,8 +15,8 @@ func GetAndValidateOIDCConfig(fileName string, refName string, clusterConfig *Cl
 	if err != nil {
 		return nil, err
 	}
-	if err = validateOIDCConfig(config); err != nil {
-		return nil, err
+	if errs := validateOIDCConfig(config); len(errs) != 0 {
+		return nil, apierrors.NewInvalid(GroupVersion.WithKind(OIDCConfigKind).GroupKind(), config.Name, errs)
 	}
 	if err = validateOIDCRefName(config, refName); err != nil {
 		return nil, err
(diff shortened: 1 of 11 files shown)


pkg/api/v1alpha1/oidcconfig.go
```diff
@@ -38,31 +41,35 @@ func getOIDCConfig(fileName string) (*OIDCConfig, error) {
...
+func validateOIDCConfig(config *OIDCConfig) field.ErrorList {
+	var errs field.ErrorList
+
 	if config == nil {
 		return nil
 	}
 
 	if config.Spec.ClientId == "" {
-		return fmt.Errorf("OIDCConfig clientId is required")
+		errs = append(errs, field.Invalid(field.NewPath("OIDCConfig", "spec", "ClientId"), config.Spec.ClientId, "OIDCConfig clientId is required"))
```
g-gaston, line 52: shouldn't this be
```suggestion
		errs = append(errs, field.Invalid(field.NewPath("spec", "OIDCConfig", "ClientId"), config.Spec.ClientId, "OIDCConfig clientId is required"))
```

pkg/cluster/oidc.go
```diff
@@ -18,7 +20,7 @@ func oidcEntry() *ConfigManagerEntry {
 			func(c *Config) error {
 				for _, o := range c.OIDCConfigs {
 					if err := o.Validate(); err != nil {
```
g-gaston, line 22: nit:
```suggestion
					if errs := o.Validate(); len(errs) != 0 {
```

pkg/cluster/oidc.go
```diff
@@ -18,7 +20,7 @@ func oidcEntry() *ConfigManagerEntry {
 			func(c *Config) error {
```
g-gaston, line 20: I like the direction you are taking with this
Maybe we should consider changing the signature of these validators to return an slice of errors in the long term. Wdyt?

pkg/cluster/oidc.go
```diff
@@ -18,7 +20,7 @@ func oidcEntry() *ConfigManagerEntry {
...
 				for _, o := range c.OIDCConfigs {
 					if err := o.Validate(); err != nil {
-						return err
+						return apierrors.NewInvalid(anywherev1.GroupVersion.WithKind(anywherev1.OIDCConfigKind).GroupKind(), o.Name, err)
```
g-gaston, line 23: I know this code was already here, but could we add a test for this case? Since we are changing the logic to return an error.

pkg/api/v1alpha1/oidcconfig_webhook.go
```diff
@@ -29,7 +29,15 @@ var _ webhook.Validator = &OIDCConfig{}
 func (r *OIDCConfig) ValidateCreate() error {
 	oidcconfiglog.Info("validate create", "name", r.Name)
 
-	return nil
+	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.Validate()...)
```
g-gaston, line 34: do we need this? or can we do
```suggestion
	allErrs := r.Validate()
```

g-gaston (commented):
love this, great idea!

pkg/cluster/oidc.go
taneyland: That makes sense! Something like changing [this line](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/manager_entry.go#L11) to return a slice of errors? Then refactoring the other validators to follow this structure?

pkg/api/v1alpha1/oidcconfig.go
```diff
@@ -38,31 +41,35 @@ func getOIDCConfig(fileName string) (*OIDCConfig, error) {
...
+func validateOIDCConfig(config *OIDCConfig) field.ErrorList {
+	var errs field.ErrorList
+
 	if config == nil {
 		return nil
 	}
 
 	if config.Spec.ClientId == "" {
-		return fmt.Errorf("OIDCConfig clientId is required")
+		errs = append(errs, field.Invalid(field.NewPath("spec", "OIDCConfig", "ClientId"), config.Spec.ClientId, "OIDCConfig clientId is required"))
```
g-gaston, line 52: sorry to come back to this again
do you know if the standard is to capitalize the field names or not?
I guess we either capitalize all or none?

pkg/cluster/oidc_test.go
```diff
@@ -56,3 +56,28 @@ func TestDefaultConfigClientBuilderOIDC(t *testing.T) {
...
 }
+
+func TestConfigManagerValidateOIDCConfigSuccess(t *testing.T) {
+	g := NewWithT(t)
+	c := clusterConfigFromFile(t, "testdata/docker_cluster_oidc_awsiam_flux.yaml")
+	m, err := cluster.NewDefaultConfigManager()
+	g.Expect(err).To(BeNil())
+
+	err = m.Validate(c)
+	g.Expect(err).To(BeNil())
```
g-gaston, line 67:
```suggestion
	g.Expect(m.Validate(c)).To(Succeed())
```

g-gaston (approved):

pkg/api/v1alpha1/oidcconfig.go
taneyland: I don't think there is a standard in place. I see that `spec` is usually all lowercase but the capitalization for the field names to follow vary across our files. It's also a weird case here because OIDC is capitalized in our config since it is an acronym

PR #2816, opened 2022-07-28
Newer Ubuntu versions not supported

chrisnegus:
*Description of changes:*
This moves the note that describes how newer versions of Ubuntu that have cgroup v2 enabled are not supported. It is now right next to where the supported Ubuntu version is listed.


TerryHowe (approved):
/lgtm

PR #2819, opened 2022-07-28
Cluster webhook validation for update

mitalipaygude:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1090

Description of changes:
Add cluster validations that we use in the CLI to the cluster webhook for updates.

Testing (if applicable):

-Unit tests
-Manual testing with tilt

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -86,6 +86,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 		return apierrors.NewBadRequest(fmt.Sprintf("expected a Cluster but got a %T", old))
 	}
 
+	if r.IsSelfManaged() && !r.IsReconcilePaused() && features.IsActive(features.FullLifecycleAPI()) {
+		return apierrors.NewBadRequest(fmt.Sprintf("upgrading self managed clusters is not supported: %s", r.Name))
+	}
+
 	var allErrs field.ErrorList
 
 	allErrs = append(allErrs, validateImmutableFieldsCluster(r, oldCluster)...)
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -94,16 +98,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "workerNodeGroupConfigurations"), r.Spec.WorkerNodeGroupConfigurations, err.Error()))
-	}
-
-	// Control plane configuration is mutable if workload cluster
-	if !r.IsSelfManaged() {
-		if err := validateControlPlaneLabels(r); err != nil {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneConfiguration", "labels"), r.Spec, err.Error()))
-		}
+	if err := r.Validate(); err != nil {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("Cluster"), r.Spec, err.Error()))
```
vivek-koppuru, line 102: In a future pr, I'm assuming we will separate each error based on the specific field within `Validate` right?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -86,6 +86,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 		return apierrors.NewBadRequest(fmt.Sprintf("expected a Cluster but got a %T", old))
 	}
 
+	if r.IsSelfManaged() && !r.IsReconcilePaused() && features.IsActive(features.FullLifecycleAPI()) {
+		return apierrors.NewBadRequest(fmt.Sprintf("upgrading self managed clusters is not supported: %s", r.Name))
```
vivek-koppuru, line 90: We need to update this for the `Create` function too right?

mitalipaygude: yes, for create the logic will be updated to return the requests, and then validate the spec, similar to this logic. Will be doing it as a part of a separate PR though. This is for Update validations.

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: yes, it will be a future refactor.

pkg/api/v1alpha1/cluster_webhook.go
vivek-koppuru: Also, what difference does it make checking for the envar in this case? We don't want to support it when the feature is enabled but when not, workload cluster is supported? Should it be separate conditions then?

vivek-koppuru (approved):

jiayiwang7 (approved):

pkg/api/v1alpha1/cluster_webhook.go
g-gaston, line 102: what about this?
```suggestion
		allErrs = append(allErrs, field.Invalid(field.NewPath("spec"), r.Spec, err.Error()))
```

jiayiwang7 (approved):

g-gaston (approved):

PR #2825, opened 2022-07-29
curatedpackages/packageclient: idiomatic go Writer

ewollesen:
The DisplayPackages and DisplayConfigurationOptions functions are modified to
take an io.Writer into which they'll write. This better conforms with go
idioms, makes testing easier, and will make it easier to add CLI flags to
write to a file in the future.

In the process of doing so, a customized tabwriter is implemented to make
working with tabwriter more uniform. Helpers for instantiation and writing a
table from a 2-D slice of strings make the code easier to read, and should
make changes to the output less error prone (less fiddling with '\t' and other
spacings).

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/installpackages.go
+++ b/cmd/eksctl-anywhere/cmd/installpackages.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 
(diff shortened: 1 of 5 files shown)


ewollesen:
replaced by #3533

PR #2826, opened 2022-07-29
remove datacenter var as vm.power does not work with relative vm path…

gwesterfieldjr:


vivek-koppuru:
Manually merging as the code coverage is complaining about log statements and the e2e test flow is verified

PR #2831, opened 2022-08-01
Set bundlesRef for new workload clusters if not set

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -143,6 +143,12 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ 
 		return ctrl.Result{}, nil
 	}
 
+	if cluster.Spec.BundlesRef == nil {
+		if err := r.setBundlesRef(ctx, cluster); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
 	if err = r.ensureClusterOwnerReferences(ctx, cluster); err != nil {
 		return ctrl.Result{}, err
 	}
@@ -216,5 +222,14 @@ func (r *ClusterReconciler) ensureClusterOwnerReferences(ctx context.Context, cl
 		}
 	}
 
+	return nil
+}
+
+func (r *ClusterReconciler) setBundlesRef(ctx context.Context, clus *anywherev1.Cluster) error {
+	mgmtCluster := &anywherev1.Cluster{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
+		return err
+	}
+	clus.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
 	return nil
 }
--- a/controllers/cluster_controller_test_test.go
+++ b/controllers/cluster_controller_test_test.go
@@ -24,6 +24,13 @@ func TestClusterReconcilerEnsureOwnerReferences(t *testing.T) {
 	g := NewWithT(t)
 	ctx := context.Background()
 
+	managementCluster := &anywherev1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-management-cluster",
+			Namespace: "my-namespace",
+		},
+	}
+
 	cluster := &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "my-cluster",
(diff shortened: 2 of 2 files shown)


controllers/cluster_controller.go
```diff
@@ -111,6 +111,16 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
 		return ctrl.Result{}, err
 	}
 
+	if cluster.IsManaged() && cluster.Spec.BundlesRef == nil {
```
g-gaston, line 114: I think we should do this later in the flow, after we check for `if cluster.IsSelfManaged()`
And that way we can avoid checking `cluster.IsManaged()`

controllers/cluster_controller.go
```diff
@@ -111,6 +111,16 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+		mgmtCluster := &anywherev1.Cluster{}
+		if err := r.client.Get(ctx, types.NamespacedName{Name: cluster.ManagedBy(), Namespace: req.Namespace}, mgmtCluster); err != nil {
+			if apierrors.IsNotFound(err) {
+				return reconcile.Result{}, nil
```
g-gaston, line 118: why doesn't this return or log the error?

g-gaston: Also, why don't we handle the other types of errors?

controllers/cluster_controller_test_test.go
```diff
@@ -88,6 +95,33 @@ func TestClusterReconcilerSetupWithManager(t *testing.T) {
 	g.Expect(r.SetupWithManager(env.Manager())).To(Succeed())
 }
 
+func TestClusterReconcilerManagementClusterNotFound(t *testing.T) {
```
g-gaston, line 98: can we have a test that asserts that when the management cluster exists, we set the bundle ref?

controllers/cluster_controller.go
taneyland: Updated to handle the other errors. We don't log or return the error when the obj isn't found. I'm not sure why but that's the behavior in other places for the controller. It seems pretty intentional so perhaps there are cases where sometimes the objects aren't found but should warrant an error

controllers/cluster_controller.go
```diff
@@ -143,6 +143,15 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
 		return ctrl.Result{}, nil
 	}
 
+	if cluster.Spec.BundlesRef == nil {
+		if err := r.SetBundlesRef(ctx, cluster); err != nil {
+			if apierrors.IsNotFound(err) {
+				return reconcile.Result{}, nil
```
g-gaston, line 149: Are you doing this because other parts of the controller do the same? If that's the case, either those other parts are wrong or they have a valid reason to do it. I don't think there is one in this case, but I might be missing something :)

I think we should return an error for this scenario as well

controllers/cluster_controller.go
```diff
@@ -218,3 +227,12 @@ func (r *ClusterReconciler) ensureClusterOwnerReferences(ctx context.Context, cl
 
 	return nil
 }
+
+func (r *ClusterReconciler) SetBundlesRef(ctx context.Context, clus *anywherev1.Cluster) error {
```
g-gaston, line 231: does this need to be public?

controllers/cluster_controller_test_test.go
g-gaston: gentle bump :)

controllers/cluster_controller.go
taneyland: I made this public for unit testing

g-gaston: let's test from reconcile and make this private

controllers/cluster_controller_test_test.go
```diff
@@ -88,6 +95,70 @@ func TestClusterReconcilerSetupWithManager(t *testing.T) {
...
+	}
+	cluster.SetManagedBy("my-management-cluster")
+
+	objs := []runtime.Object{cluster, managementCluster}
+	cb := fake.NewClientBuilder()
+	cl := cb.WithRuntimeObjects(objs...).Build()
+
+	r := controllers.NewClusterReconciler(cl, nullLog(), newRegistryForDummyProviderReconciler())
+	_, err := r.Reconcile(ctx, clusterRequest(cluster))
+	g.Expect(err).To(HaveOccurred())
```
g-gaston, line 122: might be worth asserting the error message here

g-gaston (approved):
perfect!

PR #2838, opened 2022-08-01
Replace kube-rbac-proxy in CAPI provider components manifest

abhay-krishna:
At the time of release tooling refactor, the CAPI providers had dropped usage of `kube-rbac-proxy` in their manifests, so we removed the functionality of replacing that URI in components manifests with the release URI for the `kube-rbac-proxy` image. Now, CAPC has a new release that uses kube-rbac-proxy for authz, so need to bring back that replacement functionality.


vignesh-goutham (approved):
/approve

PR #2839, opened 2022-08-01
Simplify cilium templater code

g-gaston:
*Description of changes:*

Reduced the templater api to just one method for manifests and added
options for the different scenarios. This reduces the complexity and
allows for better reusability, specially from the controller side.

*Testing (if applicable):*

Unit tests 

### DIFF
--- a/pkg/networking/cilium/templater.go
+++ b/pkg/networking/cilium/templater.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/semver"
 	"github.com/aws/eks-anywhere/pkg/templater"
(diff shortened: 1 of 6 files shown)


pkg/networking/cilium/templater.go
```diff
@@ -43,57 +44,81 @@ func (c *Templater) GenerateUpgradePreflightManifest(ctx context.Context, spec *
...
+
+	uri, version := getChartUriAndVersion(spec)
+
+	manifest, err := t.helm.Template(ctx, uri, version, namespace, c.values, c.kubeVersion)
 	if err != nil {
 		return nil, fmt.Errorf("failed generating cilium manifest: %v", err)
 	}
 
+	if spec.Cluster.Spec.ClusterNetwork.CNIConfig.Cilium.PolicyEnforcementMode == anywherev1.CiliumPolicyModeAlways {
+		networkPolicyManifest, err := t.GenerateNetworkPolicyManifest(spec, c.namespaces)
```
taneyland, line 111: Since we only set `c.namespaces` in `WithPolicyAllowedNamespaces` now, what would be the behavior if there are no namespaces? 
Alternatively, if there will always be namespaces here, what's the use case for making this a ManifestOpt?

g-gaston: The behavior is that no traffic in extra namespaces will be allowed with network policies.

The controller will use this without the namespaces.

taneyland (approved):

PR #2843, opened 2022-08-01
Refactor Flux code

jiayiwang7:
*Issue #, if available:*

#2755 

*Description of changes:*

Flux code refactoring: separate out git client, flux client used in gitops flux package. Clean up existing flux methods. In pkg/gitops/flux package:
- flux.go - methods being used by other packages (e.g. workflow)
- cluster.go - moved all the fluxForCluster methods from flux.go to here. Clean up the code a bit
- files.go - a fileGenerator that writes flux related files (eks-a configs, flux sync/patch, etc.) to local
- client.go - flux client that includes both flux executable and kubectl executable methods with retry logic
- git.go - git client that contains git and git provider client methods with partial retry logic
- upgrade.go - handles flux components upgrade

Also add generic git upgrade during components upgrade (was missing from current implementation).

*Testing (if applicable):*

Unit tests


pkg/gitops/flux/client.go
```diff
@@ -0,0 +1,116 @@
...
+		flux:    flux,
+		kube:    kube,
+		Retrier: retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
+	}
+}
+
+func (c *fluxClient) BootstrapGithub(ctx context.Context, cluster *types.Cluster, fluxConfig *v1alpha1.FluxConfig) error {
+	return c.Retry(
+		func() error {
+			return c.flux.BootstrapGithub(ctx, cluster, fluxConfig)
```
danbudris, line 53: not for this pr, just a thought: it would be good to understand the specific circumstances in which these commands (the bootstrap and uninstall cmds) fail, and if there are cases where we should and shouldn't retry; temporary loss of connectivity between the admin machine and the cluster leading to failure to call the API, for example, is a good case for a retry. But I wonder if there are other cases outside of this that we're accounting for here which should just fail and not be retried. Ideally those would be caught earlier in the validations tho...

pkg/gitops/flux/cluster.go
```diff
@@ -0,0 +1,259 @@
...
+	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/git"
+	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/validations"
+)
+
+type fluxForCluster struct {
```
danbudris, line 18: coudl be worth adding a godoc here as this struct isn't very intuitive

pkg/gitops/flux/git.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return &gitClient{
+		git:         gitTools.Client,
+		gitProvider: gitTools.Provider,
+		Retrier:     retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
+	}
+}
+
+func (c *gitClient) GetRepo(ctx context.Context) (repo *git.Repository, err error) {
```
danbudris, line 29: the git factor will return a `nil` provider for the generic git case; https://github.com/aws/eks-anywhere/blob/main/pkg/git/factory/gitfactory.go#L57 

it could be worth adding a nil check for gitTools.Provider in these methods to make that explicit

pkg/gitops/flux/upgrader.go
```diff
@@ -103,18 +97,17 @@ func (f *Flux) upgradeFilesAndCommit(ctx context.Context, newSpec *cluster.Spec)
...
 
-	logger.V(3).Info("Generating flux custom manifest files...")
-	if err := fc.writeFluxUpgradeFiles(); err != nil {
+	if err := g.WriteFluxPatch(fc.clusterSpec); err != nil {
 		return err
 	}
 
-	if err := fc.gitTools.Client.Add(fc.path()); err != nil {
-		return &ConfigVersionControlFailedError{Err: fmt.Errorf("adding %s to git: %v", fc.path(), err)}
+	if err := fc.gitClient.Add(fc.path()); err != nil {
```
danbudris, line 109: yeah this is much clearer, thanks for doing this

danbudris (approved):

pkg/gitops/flux/cluster.go
```diff
@@ -0,0 +1,261 @@
...
+// commitFluxAndClusterConfigToGit commits the cluster configuration file to the flux-managed git repository.
+// If the remote repository does not exist it will initialize a local repository and push it to the configured remote.
+// It will generate the kustomization file and marshal the cluster configuration file to the required locations in the repo.
+// These will later be used by Flux and our controllers to reconcile the repository contents and the cluster configuration.
+func (fc *fluxForCluster) commitFluxAndClusterConfigToGit(ctx context.Context) error {
+	logger.Info("Adding cluster configuration files to Git")
+	config := fc.clusterSpec.FluxConfig
+
+	if err := fc.validateLocalConfigPathDoesNotExist(); err != nil {
+		return fmt.Errorf("validating local config path not exists: %v", err)
```
mitalipaygude, line 45: nit: "local config path does not exist" ?

jiayiwang7: this actually we need "validating local config path not exists"

pkg/gitops/flux/cluster.go
```diff
@@ -0,0 +1,261 @@
...
+		Name:        fc.repository(),
+		Owner:       fc.owner(),
+		Description: "EKS-A cluster configuration repository",
+		Personal:    fc.personal(),
+		Privacy:     true,
+	}
+
+	logger.V(4).Info("Creating remote Github repo", "options", opts)
+	if err := fc.gitClient.CreateRepo(ctx, opts); err != nil {
+		return fmt.Errorf("creating repo: %w", err)
```
mitalipaygude, line 165: nit: %v
```suggestion
		return fmt.Errorf("creating repo: %v", err)
```

pkg/gitops/flux/cluster.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+// validateLocalConfigPathDoesNotExist returns an exception if the cluster configuration file exists.
+// This is done so that we avoid clobbering existing cluster configurations in the user-provided git repository.
+func (fc *fluxForCluster) validateLocalConfigPathDoesNotExist() error {
+	if fc.clusterSpec.Cluster.IsSelfManaged() {
+		p := path.Join(fc.writer.Dir(), fc.path())
+		if validations.FileExists(p) {
+			return fmt.Errorf("a cluster configuration file already exists at path %s", p)
```
mitalipaygude, line 195: Just out of curiosity, would this be good user experience if we return from this point? can we store the config with a different name maybe? Also I suppose this validation can't be moved up in the code?

pkg/gitops/flux/cluster.go
jiayiwang7: the method has proper error info. gonna remove the error statement here

pkg/gitops/flux/cluster.go
jiayiwang7: I didn't write this method but only move the code to different file. But you are right, I think we can move this validation up to preflight and fail early if local path exists. We can refactor this after this PR.

mitalipaygude (approved):

PR #2845, opened 2022-08-02
Feature gated packages

a-cool-train:
*Description of changes:*
- Curated Packages will be gated for GA. 
- Refactor to simplify dependency management.

*Testing (if applicable):*


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -38,7 +53,21 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
 	sourceRegistry := fmt.Sprintf("sourceRegistry=%s", registry)
 	values := []string{sourceRegistry}
-	return pc.chartInstaller.InstallChart(ctx, pc.chartName, ociUri, pc.chartVersion, pc.kubeConfig, values)
+	err := pc.chartInstaller.InstallChart(ctx, pc.chartName, ociUri, pc.chartVersion, pc.kubeConfig, values)
+	if err != nil {
+		return err
+	}
+
+	err = pc.ApplySecret(ctx)
+	if err != nil {
```
g-gaston, line 62: nit:
```suggestion
	if err = pc.ApplySecret(ctx); err != nil {
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -38,7 +53,21 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
+		logger.Info("Warning: not able to create secret. Package installation might fail.", "error", err)
+	}
+
+	err = pc.TriggerCronJob(ctx)
+	if err != nil {
```
g-gaston, line 67:
```suggestion
	if err = pc.TriggerCronJob(ctx); err != nil {
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -38,7 +53,21 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
 	values := []string{sourceRegistry}
-	return pc.chartInstaller.InstallChart(ctx, pc.chartName, ociUri, pc.chartVersion, pc.kubeConfig, values)
+	err := pc.chartInstaller.InstallChart(ctx, pc.chartName, ociUri, pc.chartVersion, pc.kubeConfig, values)
+	if err != nil {
+		return err
+	}
+
+	err = pc.ApplySecret(ctx)
+	if err != nil {
+		logger.Info("Warning: not able to create secret. Package installation might fail.", "error", err)
```
g-gaston, line 63: would it make sense to use `logger.Error`?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
 	}
 	return nil
 }
+
+func (pc *PackageControllerClient) ApplySecret(ctx context.Context) error {
+	eksaAccessKeyId, eksaSecretAccessKey := os.Getenv(config.EksaAccessKeyIdEnv), os.Getenv(config.EksaSecretAcessKeyEnv)
```
g-gaston, line 82: Ideally, I think we should inject these two in `NewPackageControllerClient`
It makes the business logic agnostic from how the config is read

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+	eksaRegion := eksaDefaultRegion
+	if region, found := os.LookupEnv(config.EksaRegionEnv); found {
```
g-gaston, line 84: same comment about injecting the config value in the constructor :)

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+	templateValues := map[string]string{
+		"eksaAccessKeyId":     base64.StdEncoding.EncodeToString([]byte(eksaAccessKeyId)),
+		"eksaSecretAccessKey": base64.StdEncoding.EncodeToString([]byte(eksaSecretAccessKey)),
+		"eksaRegion":          base64.StdEncoding.EncodeToString([]byte(eksaRegion)),
+	}
+
+	result, err := templater.Execute(awsSecretYaml, templateValues)
+	if err != nil {
+		return fmt.Errorf("replacing template values %v", err)
+	}
```
g-gaston, line 97: nit: I would just define the secret using go structs and just marshal it to yaml. I think it makes the code a bit more clear since you don't need to go read the template yaml file and you have everything defined here in code

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+
+	params := []string{"create", "-f", "-", "--kubeconfig", pc.kubeConfig}
+	stdOut, err := pc.kubectl.CreateFromYaml(ctx, result, params...)
```
g-gaston, line 100: i know this is not part of this PR, but I think we should revisit this method signature. It feels non intuitive that we need to pass "create" in the params when the method is already named create.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+	if err != nil {
+		return fmt.Errorf("creating secret %v", err)
+	}
+
+	fmt.Print(&stdOut)
```
g-gaston, line 105: nit: any specific reason to use `Print` vs `Println`? If the text you print doesn't end with a line break, this will mess the terminal a bit

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -1,18 +1,33 @@
...
 	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages/mocks"
+	"github.com/aws/eks-anywhere/pkg/templater"
+)
+
+//go:embed config/awssecret.yaml
+var awsSecretYaml string
```
g-gaston, line 25: I'm guessing we have this to be able to generate secrets in the tests. I would expect us to have files in testdata with the final expected result so we don't need to generate them on the fly. This makes the tests more robust: if there is a bug in the template and we use the same logic to generate the yaml in the code and in the tests, we won't catch the bug.

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -54,14 +75,52 @@ func TestInstallControllerSuccess(t *testing.T) {
...
 	sourceRegistry := fmt.Sprintf("sourceRegistry=%s", registry)
 	values := []string{sourceRegistry}
+	params := []string{"create", "-f", "-", "--kubeconfig", tt.kubeConfig}
+	templateValues := map[string]string{
+		"eksaAccessKeyId":     "",
+		"eksaSecretAccessKey": "",
+		"eksaRegion":          base64.StdEncoding.EncodeToString([]byte(eksaDefaultRegion)),
+	}
+	result, err := templater.Execute(awsSecretYaml, templateValues)
+	tt.Expect(err).To(BeNil())
```
g-gaston, line 85: nit: I tend to use this, a bit more expressive imo but not big deal
```suggestion
	tt.Expect(err).NotTo(HaveOcurred())
```

pkg/dependencies/factory.go
```diff
@@ -827,15 +830,16 @@ func (f *Factory) WithGitOpsFlux(clusterConfig *v1alpha1.Cluster, fluxConfig *v1
 }
 
 func (f *Factory) WithPackageInstaller(spec *cluster.Spec, packagesLocation string) *Factory {
-	f.WithHelmInsecure().WithKubectl()
+	f.WithHelmInsecure().WithKubectl().WithPackageControllerClient(spec).WithPackageClient()
```
g-gaston, line 833: if it doesn't use helm, should be remove it from here?
```suggestion
	f.WithKubectl().WithPackageControllerClient(spec).WithPackageClient()
```

g-gaston (commented):
Looks great, left a few comments but I believe all minor

pkg/curatedpackages/config/awssecret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+data:
+  ID: {{.eksaAccessKeyId}}
```
vivek-koppuru, line 3: This is basically like hardcoding credentials. I know this was discussed in the design doc, but have we considered making IRSA a requirement for this and leveraging that so we don't need a whole user for pull the images?

pkg/curatedpackages/packagecontrollerclient.go
vivek-koppuru: Is this something where if it fails, the user can rerun this command or will it be messy to clean up if package installation fails?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+
+func (pc *PackageControllerClient) ApplySecret(ctx context.Context) error {
+	eksaAccessKeyId, eksaSecretAccessKey := os.Getenv(config.EksaAccessKeyIdEnv), os.Getenv(config.EksaSecretAcessKeyEnv)
+	eksaRegion := eksaDefaultRegion
+	if region, found := os.LookupEnv(config.EksaRegionEnv); found {
+		eksaRegion = region
+	}
+
+	templateValues := map[string]string{
+		"eksaAccessKeyId":     base64.StdEncoding.EncodeToString([]byte(eksaAccessKeyId)),
```
vivek-koppuru, line 89: If you save it in the secret as `stringData`, you may not need to base64 encode it here but I think this is fine.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -48,3 +77,41 @@ func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Co
...
+	stdOut, err := pc.kubectl.CreateFromYaml(ctx, result, params...)
+	if err != nil {
+		return fmt.Errorf("creating secret %v", err)
+	}
+
+	fmt.Print(&stdOut)
+	return nil
+}
+
+func (pc *PackageControllerClient) TriggerCronJob(ctx context.Context) error {
```
vivek-koppuru, line 109: This is really a Create and not a Trigger right?

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -54,14 +75,52 @@ func TestInstallControllerSuccess(t *testing.T) {
...
+	registry := curatedpackages.GetRegistry(tt.ociUri)
+	sourceRegistry := fmt.Sprintf("sourceRegistry=%s", registry)
+	values := []string{sourceRegistry}
+	params := []string{"create", "-f", "-", "--kubeconfig", tt.kubeConfig}
+	templateValues := map[string]string{
+		"eksaAccessKeyId":     "",
+		"eksaSecretAccessKey": "",
+		"eksaRegion":          base64.StdEncoding.EncodeToString([]byte(tt.eksaRegion)),
+	}
+	os.Setenv(config.EksaRegionEnv, tt.eksaRegion)
```
vivek-koppuru, line 109: You can make use of https://pkg.go.dev/testing#T.Setenv now to not have to worry about cleaning it up at the end.

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: The reason it is a warning instead of an error is because there are scenarios where this failure doesn't cause any issue. For example if the user is using registry mirror. We don't want to mention an error in that case. It is also possible the secret was already created before and this error doesn't really impact anything as such.

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: Yes the result from kubectl already sends the result with a new line character. I didn't see the necessity to add an extra /ln to it.

pkg/curatedpackages/config/awssecret.yaml
a-cool-train: @junshun might have a good answer on this. Thoughts?

pkg/config/config.go
```diff
@@ -15,6 +15,9 @@ const (
 	SshKnownHostsEnv            = "SSH_KNOWN_HOSTS"
 	EksaReplicasReadyTimeoutEnv = "EKSA_REPLICAS_READY_TIMEOUT"
 	ExternalEtcdTimeoutEnv      = "EKSA_EXTERNAL_ETCD_TIMEOUT"
```
TerryHowe, line 17: maybe this has been discussed already, but should this map directly to the AWS CLI env vars like EKSA_REGION would be EKSA_AWS_REGION etc?

TerryHowe (commented):
it looks like the AWS creds are only used to create the secret, so we won't have to pass them to the CLI tool which is nice.

pkg/curatedpackages/config/awssecret.yaml
junshun: For now we are currently suggesting that they use this method. IRSA could work however we won't be suggesting. IRSA also makes it more difficult as we migrate towards the long term solution. Since we are suggesting that the customer will create a role with limited scope (just pulling images/getting token) there should be minimal risks.

pkg/config/config.go
junshun: So the current naming is replacing AWS with EKSA, is your suggestion prefixing EKSA_ in front of the variable names?

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: Is this a data type? I wasn't able to find stringData.

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: I will be having a PR to remove feature flag. I'll revisit this when that PR is ready.

pkg/config/config.go
a-cool-train: Yeah I think we should have EKSA_ as prefix. Made the change.

pkg/curatedpackages/config/awssecret.yaml
vivek-koppuru: I understand, but there is a difference between creating a role and creating a user, and here it is specifying creating a user. But if IRSA makes it more difficult maybe this is an okay step in that direction.

pkg/curatedpackages/packagecontrollerclient.go
vivek-koppuru: okay as long as even if it fails the cluster state can be recovered somehow

pkg/curatedpackages/packagecontrollerclient.go
vivek-koppuru: yea its a field in the yaml: https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret

a-cool-train: aaah I see. Let me try that.

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: This failure doesn't cause any impact. Even install package controller failing doesn't have an impact on the cluster. Otherwise, that would be very concerning.

vivek-koppuru (approved):

Issue #2837, opened 2022-08-01
download images command does not work with proxy


wongni:
**What happened**:

```
$ eksctl anywhere download images  --output artifacts.tar
Pulling images from origin, this might take a while
Writing images to disk
Pulling images from origin, this might take a while
Writing images to disk
Saving helm chart to disk	{"chart": "public.ecr.aws/eks-anywhere/eks-anywhere-packages:0.1.13-eks-a-12"}
Error: downloading chart [public.ecr.aws/eks-anywhere/eks-anywhere-packages:0.1.13-eks-a-12] from registry: Error: failed to do request: Head "https://public.ecr.aws/v2/eks-anywhere/eks-anywhere-packages/manifests/0.1.13-eks-a-12": dial tcp 99.83.145.10:443: connect: connection timed out
```

**What you expected to happen**:
The command should work with proxy

**How to reproduce it (as minimally and precisely as possible)**:
Run `eksctl anywhere download images  --output artifacts.tar` with proxy

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: v0.10.1
- EKS Distro Release: 0.106.0

jiayiwang7:
We did fix the issue of passing proxy ENVs from clusterSpec into helm executable: https://github.com/aws/eks-anywhere/pull/2032

But since `download images` command does not use clusterSpec, we need to pass the proxy ENVs from the user environment instead, to run helm commands.

PR #2849, opened 2022-08-02
Derive Proxy Config From Env Vars

jonathanmeier5:
#2837

*Description of changes:*

This PR fixes a bug in our proxy configuration logic. Prior to this PR, we only derived proxy configs from cluster config files.

This meant commands such as `eksctl anywhere download images` could not be used with a proxy.

This PR allows the CLI to derive a proxy config from the following env vars:
```
HTTPS_PROXY
HTTP_PROXY
NO_PROXY
```

*Testing (if applicable):*

I tested the `eksctl anywhere download images` command manually on an ec2 instance that could only communicate to the internet via a second ec2 instance that ran a proxy service.

I'm more than happy to add unit tests, I just wanted a review on the implementation first in case I was very off base.

### DIFF
--- a/pkg/config/proxy.go
+++ b/pkg/config/proxy.go
@@ -1,0 +1,19 @@
+package config
+
+import (
+	"os"
+)
+
+const (
+	HttpsProxyKey = "HTTPS_PROXY"
+	HttpProxyKey  = "HTTP_PROXY"
+	NoProxyKey    = "NO_PROXY"
+)
+
+func GetProxyConfigFromEnv() map[string]string {
+	return map[string]string{
+		HttpsProxyKey: os.Getenv(HttpsProxyKey),
+		HttpProxyKey:  os.Getenv(HttpProxyKey),
+		NoProxyKey:    os.Getenv(NoProxyKey),
+	}
+}
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -104,7 +104,7 @@ func ForSpec(ctx context.Context, clusterSpec *cluster.Spec) *Factory {
 	return NewFactory().
 		UseExecutableImage(eksaToolsImage.VersionedImage()).
 		WithRegistryMirror(clusterSpec.Cluster.RegistryMirror()).
-		WithProxyConfiguration(clusterSpec.Cluster.ProxyConfiguration()).
+		UseProxyConfiguration(clusterSpec.Cluster.ProxyConfiguration()).
 		WithWriterFolder(clusterSpec.Cluster.Name).
 		WithDiagnosticCollectorImage(clusterSpec.VersionsBundle.Eksa.DiagnosticCollector.VersionedImage())
 }
@@ -165,8 +165,25 @@ func (f *Factory) WithRegistryMirror(mirror string) *Factory {
 	return f
 }
 
-func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory {
+func (f *Factory) UseProxyConfiguration(proxyConfig map[string]string) *Factory {
 	f.proxyConfiguration = proxyConfig
+	return f
+}
+
+func (f *Factory) GetProxyConfiguration() map[string]string {
+	return f.proxyConfiguration
+}
+
+func (f *Factory) WithProxyConfiguration() *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.proxyConfiguration == nil {
+			proxyConfig := config.GetProxyConfigFromEnv()
+			f.UseProxyConfiguration(proxyConfig)
+		}
+		return nil
+	},
+	)
+
 	return f
 }
 
@@ -580,7 +597,7 @@ func (f *Factory) WithTroubleshoot() *Factory {
 }
 
 func (f *Factory) WithHelmSecure() *Factory {
-	f.WithExecutableBuilder()
+	f.WithExecutableBuilder().WithProxyConfiguration()
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		if f.dependencies.HelmSecure != nil {
@@ -604,7 +621,7 @@ func (f *Factory) WithHelmSecure() *Factory {
 }
 
 func (f *Factory) WithHelmInsecure() *Factory {
-	f.WithExecutableBuilder()
+	f.WithExecutableBuilder().WithProxyConfiguration()
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		if f.dependencies.HelmInsecure != nil {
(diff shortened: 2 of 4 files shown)


pkg/dependencies/factory.go
```diff
@@ -167,6 +168,20 @@ func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory
 	return f
 }
 
+func (f *Factory) WithProxyConfigurationFromEnv() *Factory {
+	
+	if f.proxyConfiguration == nil {
+		proxyConfig := map[string]string{
+			"HTTP_PROXY":  os.Getenv("HTTP_PROXY"),
```
jonathanmeier5, line 175: Should these all be constants?

pkg/dependencies/factory.go
```diff
@@ -167,6 +168,20 @@ func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory
...
+			"HTTPS_PROXY": os.Getenv("HTTPS_PROXY"),
+			"NO_PROXY":    os.Getenv("NO_PROXY"),
```
jonathanmeier5, line 177: Should I add documentation for these env vars?

pkg/dependencies/factory.go
```diff
@@ -167,6 +168,19 @@ func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory
 	return f
 }
 
+func (f *Factory) WithProxyConfigurationFromEnv() *Factory {
+	if f.proxyConfiguration == nil {
```
g-gaston, line 172: I think we could treat this as another dependency
for example: rename `WithProxyConfiguration` to `UseProxyConfiguration`
And have this new method named `WithProxyConfiguration`

Then any dependency that wants the proxy config, will call `WithProxyConfiguration` as part of their dependencies. And at the build time, if the config hasn't been set manually already (with `UseProxyConfiguration`) then we read it from the environment

pkg/dependencies/factory.go
g-gaston: Looking at this, I wonder if we should make this map part of the config in the `config` package

vivek-koppuru: Yea maybe this isn't something we need to keep in factory and have to specify for all dependencies where we want to share this logic

pkg/dependencies/factory.go
```diff
@@ -601,7 +615,7 @@ func (f *Factory) WithHelmSecure() *Factory {
 }
 
 func (f *Factory) WithHelmInsecure() *Factory {
-	f.WithExecutableBuilder()
+	f.WithExecutableBuilder().WithProxyConfigurationFromEnv()
```
vivek-koppuru, line 618: Also, if we call this here, does it mean that the values from env can override the values specified in cluster spec, or vice versa? If it's the other way, maybe it is okay? But ideally we can read the values from the clusterspec if we anyways need to take it in as input, but this would make it more consistent with how the docker pull command would read the env vars

pkg/dependencies/factory.go
jonathanmeier5: @g-gaston I *think* I did what you asked, let me know if it wasn't what you were imagining.

@vivek-koppuru re [this comment](https://github.com/aws/eks-anywhere/pull/2849#discussion_r937035140), I was concerned about the same thing.

I think that with the approach Guillermo outlines above the proxy config from the cluster config file will always take precedence over the env vars as long as `UseProxyConfiguration` is explicitly called as a dependency. I think this is what we want, right?

pkg/config/proxy.go
```diff
@@ -0,0 +1,21 @@
...
+	"os"
+)
+
+const (
+	httpsProxyKey = "HTTPS_PROXY"
+	httpProxyKey  = "HTTP_PROXY"
+	noProxyKey    = "NO_PROXY"
+)
+
+func GetProxyConfigFromEnv() (map[string]string, error) {
```
g-gaston, line 13: any reason why this needs to return an error?

pkg/dependencies/factory.go
```diff
@@ -162,11 +162,25 @@ func (f *Factory) WithRegistryMirror(mirror string) *Factory {
 	return f
 }
 
-func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory {
+func (f *Factory) UseProxyConfiguration(proxyConfig map[string]string) *Factory {
 	f.proxyConfiguration = proxyConfig
 	return f
 }
 
+func (f *Factory) WithProxyConfiguration() *Factory {
```
g-gaston, line 170: It should be already tested by I would add this explicitly to a test and assert not-nil

pkg/config/proxy.go
g-gaston, line 13: Maybe add a unit test for this guy?

g-gaston (commented):
Super clean!

pkg/config/proxy.go
jonathanmeier5: Nope, removed it 👍

pkg/dependencies/factory.go
```diff
@@ -162,11 +162,32 @@ func (f *Factory) WithRegistryMirror(mirror string) *Factory {
 	return f
 }
 
-func (f *Factory) WithProxyConfiguration(proxyConfig map[string]string) *Factory {
+func (f *Factory) UseProxyConfiguration(proxyConfig map[string]string) *Factory {
 	f.proxyConfiguration = proxyConfig
 	return f
 }
 
+func (f *Factory) GetProxyConfiguration() *map[string]string {
```
g-gaston, line 170: curious: what is this for and why would it return a pointer?

pkg/config/proxy.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func GetProxyConfigFromEnv() map[string]string {
+	config := map[string]string{
+		HttpsProxyKey: os.Getenv(HttpsProxyKey),
+		HttpProxyKey:  os.Getenv(HttpProxyKey),
+		NoProxyKey:    os.Getenv(NoProxyKey),
+	}
+
+	return config
```
g-gaston, line 20: nit:
```suggestion
	return map[string]string{
		HttpsProxyKey: os.Getenv(HttpsProxyKey),
		HttpProxyKey:  os.Getenv(HttpProxyKey),
		NoProxyKey:    os.Getenv(NoProxyKey),
	}
```

pkg/dependencies/factory.go
jonathanmeier5: The `Factory` struct uses an unpublished `proxyConfiguration` attribute to track proxy information in the `dependencies` package. To test it in `dependencies_testing` I needed a way to access the attribute.

I don't mind passing by value here. Given the method's intent is introspective I think a pointer is the first thing that came to mind, but I don't need it. After a quick google it looks like returning smaller objects by value is more the norm in golang.

If you're getting at "why are we testing an unpublished attribute when go doesn't want us to", there are a few other solutions, but they all felt equally weird to me and involved more work so I went with this one. Happy to discuss though.

PR #2858, opened 2022-08-03
Preflight validate cmd POC

raymond-zhang00:
*Issue #, if available:*
Add validate cmd

*Description of changes:*
Add validate cmd to run create cluster validations

*Testing (if applicable):*
To be added


cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,207 @@
...
+
+	runner.Register(cmdvalidations.PackageCreatePreflight(ctx, createValidations)...)
+	runner.Register(cmdvalidations.PackageProviderValidations(ctx, clusterSpec, deps.Provider)...)
+	runner.Register(deps.GitOpsFlux.Validations(ctx, clusterSpec)...)
+
+	err = runner.StoreValidationResults()
+	runner.ReportResults()
+
+	cleanup(deps, &err)
+	deps.Writer.CleanUp()
```
raymond-zhang00, line 180: This would unintentionally remove any existing folder and contents sharing the cluster name right?

raymond-zhang00:
Usage - `./eksctl-anywhere exp validate -f <yaml config file>`

Output - valid config example:
```✅ Validate docker executable
✅ Validate cluster
✅ Validate kubeconfig path
✅ Validate supported provider
✅ Validate certificate for registry mirror
✅ Create preflight validations pass
✅ Connected to server
✅ Authenticated to vSphere
✅ Datacenter validated
✅ Network validated
✅ Datastore validated
✅ Folder validated
✅ Resource pool validated
✅ Control plane and Workload templates validated
Provided control plane sshAuthorizedKey is not set or is empty, auto-generating new key pair...
Private key saved to tmpValidate1230314345/eks-a-id_rsa. Use 'ssh -i tmpValidate1230314345/eks-a-id_rsa <username>@<Node-IP-Address>' to login to your cluster node
✅ Vsphere Provider setup is valid```

Output - invalid config example:
```✅ Validate docker executable
❌ Validation failed	{"validation": "validate cluster", "error": "failed to validate cluster config name:  is not a valid cluster name, cluster names must start with lowercase/uppercase letters and can include numbers and dashes. For instance 'testCluster-123' is a valid name but '123testCluster' is not. ", "remediation": ""}
Error: validations failed```

File: cmd/eksctl-anywhere/cmd/validate.go [hunk 2]
```diff
@@ -0,0 +1,166 @@
...
+	clusterOptions
+	skipIpCheck           bool
+	hardwareCSVPath       string
+	tinkerbellBootstrapIP string
+}
+
+var valOpt = &validateOptions{}
+
+var validateCmd = &cobra.Command{
+	Use:          "validate",
```
g-gaston, line 33: I wonder if we should have a `eks-a exp validate create cluster` command
or at least `esk-a exp validate create`
I can def see us implementing another command for upgrades as well

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+		TinkerbellHardwareCSVFlagDescription,
+	)
+	validateCmd.Flags().StringVar(&valOpt.tinkerbellBootstrapIP, "tinkerbell-bootstrap-ip", "", "Override the local tinkerbell IP in the bootstrap cluster")
+	validateCmd.Flags().BoolVar(&valOpt.skipIpCheck, "skip-ip-check", false, "Skip check for whether cluster control plane ip is in use")
+	validateCmd.Flags().StringVar(&valOpt.bundlesOverride, "bundles-override", "", "Override default Bundles manifest (not recommended)")
+	validateCmd.Flags().StringVar(&valOpt.managementKubeconfig, "kubeconfig", "", "Management cluster kubeconfig file")
+
+	if err := validateCmd.MarkFlagRequired("filename"); err != nil {
+		log.Fatalf("Error marking flag as required: %v", err)
+	}
```
g-gaston, line 58: We should probably look into moving these flags to a common place so we avoid copy-paste from the create command

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+		&valOpt.hardwareCSVPath,
+		TinkerbellHardwareCSVFlagName,
+		TinkerbellHardwareCSVFlagAlias,
+		"",
+		TinkerbellHardwareCSVFlagDescription,
+	)
+	validateCmd.Flags().StringVar(&valOpt.tinkerbellBootstrapIP, "tinkerbell-bootstrap-ip", "", "Override the local tinkerbell IP in the bootstrap cluster")
+	validateCmd.Flags().BoolVar(&valOpt.skipIpCheck, "skip-ip-check", false, "Skip check for whether cluster control plane ip is in use")
+	validateCmd.Flags().StringVar(&valOpt.bundlesOverride, "bundles-override", "", "Override default Bundles manifest (not recommended)")
+	validateCmd.Flags().StringVar(&valOpt.managementKubeconfig, "kubeconfig", "", "Management cluster kubeconfig file")
```
g-gaston, line 54: Do we need all of these?
I think we can remove at least skipIpCheck and bundlesOverride

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+	})
+	return nil
+}
+
+func (valOpt *validateOptions) validateCluster(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+
+	validateCluster := workflows.NewValidate(ctx)
+
+	validateCluster.RunDockerValidations()
```
g-gaston, line 76: why does this run separetly?

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return err
+	}
+
+	cleanup(deps, &err)
+	deps.Writer.CleanUp()
+	return err
+}
+
+func (valOpt *validateOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig *config.CliConfig) ([]string, error) {
```
g-gaston, line 143: is this copy paste from a different command?

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+	validateCluster := workflows.NewValidate(ctx)
+
+	validateCluster.RunDockerValidations()
+
+	clusterConfig, err := cluster.ParseConfigFromFile(valOpt.fileName)
+	if err != nil {
+		return err
+	}
+
+	err = validateCluster.RunConfigValidations(clusterConfig)
```
g-gaston, line 83: why don't make this part of the `validateCluster.RunSpecValidations`?

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	cliConfig := buildCliConfig(clusterSpec)
+	dirs, err := valOpt.directoriesToMount(clusterSpec, cliConfig)
+	if err != nil {
+		return err
+	}
+
+	tmpPath, err := os.MkdirTemp("./", "tmpValidate")
```
g-gaston, line 117: the write should create this for you (I think)

pkg/cluster/cloudstack.go
```diff
@@ -62,6 +62,9 @@ func cloudstackEntry() *ConfigManagerEntry {
 func processCloudStackDatacenter(c *Config, objects ObjectLookup) {
 	if c.Cluster.Spec.DatacenterRef.Kind == anywherev1.CloudStackDatacenterKind {
 		datacenter := objects.GetFromRef(c.Cluster.APIVersion, c.Cluster.Spec.DatacenterRef)
+		if datacenter == nil {
```
g-gaston, line 65: We should probably move this to a separate PR, this is useful
Maybe add a few tests

pkg/validations/runner.go
```diff
@@ -6,12 +6,15 @@ var errRunnerValidation = errors.New("validations failed")
 
 type Validation func() *ValidationResult
 
+type ValidationFunc func() error
```
g-gaston, line 9: what is this for?

pkg/validations/cmdvalidations/cmdvalidations.go
```diff
@@ -0,0 +1,124 @@
+package cmdvalidations
```
g-gaston, line 1: What is this package supposed to contain? I mix match of different things that I'm not sure belong in the same place.

pkg/validations/runner.go
```diff
@@ -27,6 +30,7 @@ func (r *Runner) Run() error {
 			failed = true
 		}
 	}
+	r.validations = make([]Validation, 0)
```
g-gaston, line 33: what is this for?

pkg/workflows/validate.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/validations"
+	"github.com/aws/eks-anywhere/pkg/validations/cmdvalidations"
+	"github.com/aws/eks-anywhere/pkg/validations/createvalidations"
+)
+
+type Validate struct {
```
g-gaston, line 16: I'm not sure I understand what this is supposed to be
Usually it's easier to understand what an entity is when named with a noun

pkg/workflows/validate.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx    context.Context
```
g-gaston, line 17: it's usually better to pass the context as an argument to methods

pkg/workflows/validate.go
```diff
@@ -0,0 +1,91 @@
+package workflows
```
g-gaston, line 1: It's difficult to know without understanding what this `Validate` is, but I think we shouldn't add anything to the workflows package
This probably belong in a `validations` subpackage?

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func preRunValidate(cmd *cobra.Command, args []string) error {
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		err := viper.BindPFlag(flag.Name, flag)
+		if err != nil {
+			log.Fatalf("Error initializing flags: %v", err)
+		}
+	})
+	return nil
```
chrisdoherty4, line 68: I know this is copied but since this is new code lets do some simple things better.

```suggestion
    var err error
	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		if err != nil {
		  return
		}
		err = viper.BindPFlag(flag.Name, flag)
	})
	return err
```

cmd/eksctl-anywhere/cmd/validate.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (valOpt *validateOptions) validateCluster(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+
+	validateCluster := workflows.NewValidate(ctx)
+
+	validateCluster.RunDockerValidations()
+
+	clusterConfig, err := cluster.ParseConfigFromFile(valOpt.fileName)
```
chrisdoherty4, line 78: @g-gaston This makes me think we should move the parsing and validation logic (orthogonal to this work) to a `config` package.

chrisdoherty4:
I haven't reviewed this in full but great work!

Do we think its necessary to output so much information? I suspect consumers don't really care that a specific portion of validation succeeded - that feels more like a debug level of logging. I care whether or not the invocation I made makes sense or not and I can get that from a single line (ideally) saying "Configuration valid".

pkg/workflows/validate.go
taneyland: Ah ok I figured that it would make sense to have this in `workflows` since we are using a runner & the structure of this could be similar to our other workflows that use a runner (create, delete, etc.) but I see your point. Makes sense to add it to its own subpackage

cmd/eksctl-anywhere/cmd/validate.go
raymond-zhang00: I definitely agree, will split this out

cmd/eksctl-anywhere/cmd/validate.go
raymond-zhang00: I think it's pass/fail won't get in the way of the cluster and spec dependent validations so I wanted to just run this independently of the others.

cmd/eksctl-anywhere/cmd/validate.go
raymond-zhang00: Yeah, this is taken directly from the createCluster cmd. Should I move it out into a separate function that both cmds can access?
https://github.com/aws/eks-anywhere/blob/9b1601f90b00ba5fbb39a0fd00cc06de53235c71/cmd/eksctl-anywhere/cmd/createcluster.go#L211

cmd/eksctl-anywhere/cmd/validate.go
raymond-zhang00: I was hoping to break up the validations to ones that were only dependent on cluster.Config so that they can be run regardless of if cluster.Spec parse succeeds. Since cluster.Spec does the same parsing and validating though, it does run through a lot of the same logic twice.

pkg/cluster/cloudstack.go
raymond-zhang00: Sounds good, will move this out

pkg/validations/cmdvalidations/cmdvalidations.go
raymond-zhang00: I'm packaging the validations into validation results and grouping some individual checks into single validation. Should grouping the validations be moved out to the broader validations package?

pkg/validations/runner.go
raymond-zhang00: This is so I can register() and run() validations multiple times using the same runner. Just trying to clear out the r.validations array after a run.

pkg/validations/runner.go
raymond-zhang00: This was left by mistake from some previous changes I made. Removing!

raymond-zhang00:
Closing this POC PR

PR #2859, opened 2022-08-03
Load snow env vars from secret for controller

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Adding ability to read credentials from the credentials secret provided by the capas provider in order to construct the aws client maps.

*Testing (if applicable):*
unit testing

### DIFF
--- a/pkg/providers/snow/reconciler/clientbuilder.go
+++ b/pkg/providers/snow/reconciler/clientbuilder.go
@@ -1,0 +1,97 @@
+package reconciler
+
+import (
+	"bytes"
+	"context"
+	"net"
+
+	awstypes "github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	awscredentials "github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/pkg/errors"
+	"gopkg.in/ini.v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/aws"
+)
+
+type AwsClientBuilder struct {
+	client client.Client
+}
+
+func NewAwsClientBuilder(client client.Client) *AwsClientBuilder {
+	return &AwsClientBuilder{
+		client: client,
+	}
+}
+
+func (b *AwsClientBuilder) Build(ctx context.Context) (aws.Clients, error) {
+	credentials, certificates, err := getSnowCredentials(ctx, b.client)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting snow credentials")
+	}
+
+	return createAwsClients(ctx, credentials, certificates)
+}
+
+type credentialConfiguration struct {
+	AccessKey string `ini:"aws_access_key_id"`
+	SecretKey string `ini:"aws_secret_access_key"`
+	Region    string `ini:"region"`
+}
+
+func createAwsClients(ctx context.Context, credentials []byte, certificates []byte) (aws.Clients, error) {
+	var deviceIps []string
+	credsCfg, err := ini.Load(credentials)
+	if err != nil {
+		return nil, errors.Wrap(err, "loading values from credentials")
+	}
+	for _, ip := range credsCfg.SectionStrings() {
+		if net.ParseIP(ip) != nil {
+			deviceIps = append(deviceIps, ip)
+		}
+	}
+
+	deviceClientMap := make(aws.Clients, len(deviceIps))
+	for _, ip := range deviceIps {
+		ipCfg, err := parseIpConfiguration(credsCfg, ip)
+		if err != nil {
+			return nil, errors.Wrapf(err, "parsing configuration for %v", ip)
+		}
+		clientCfg, err := awsconfig.LoadDefaultConfig(ctx,
+			awsconfig.WithCustomCABundle(bytes.NewReader(certificates)),
+			awsconfig.WithRegion(ipCfg.Region),
+			awsconfig.WithCredentialsProvider(awscredentials.StaticCredentialsProvider{
+				Value: awstypes.Credentials{
+					AccessKeyID:     ipCfg.AccessKey,
+					SecretAccessKey: ipCfg.SecretKey,
+				},
+			}),
+			awsconfig.WithEndpointResolverWithOptions(aws.SnowEndpointResolver(ip)),
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "setting up aws client")
+		}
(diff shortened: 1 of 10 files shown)


vivek-koppuru:
Patch coverage is showing up as 100% right now because no test files are present in the package

pkg/providers/snow/reconciler/clientbuilder.go
```diff
@@ -0,0 +1,97 @@
...
+	return createAwsClients(ctx, credentials, certificates)
+}
+
+type credentialConfiguration struct {
+	AccessKey string `ini:"aws_access_key_id"`
+	SecretKey string `ini:"aws_secret_access_key"`
+	Region    string `ini:"region"`
+}
+
+func createAwsClients(ctx context.Context, credentials []byte, certificates []byte) (aws.Clients, error) {
```
vivek-koppuru, line 43: Separated this function to make it easy to move to aws package if we wanted

pkg/providers/snow/reconciler/clientbuilder.go
```diff
@@ -0,0 +1,97 @@
...
+func NewAwsClientBuilder(client client.Client) *awsClientBuilder {
+	return &awsClientBuilder{
+		client: client,
+	}
+}
+
+func (b *awsClientBuilder) BuildSnowAwsClientMap(ctx context.Context) (aws.Clients, error) {
+	credentials, certificates, err := SnowCredentials(ctx, b.client)
+	if err != nil {
+		return nil, fmt.Errorf("getting snow credentials: %v", err)
```
g-gaston, line 31: I have been trying to use `errors.Wrapf` in the controller code

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -0,0 +1,25 @@
...
+
+	apiv1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+const BoostrapSecretName = "capas-manager-bootstrap-credentials"
+
+func SnowCredentials(ctx context.Context, cli client.Client) ([]byte, []byte, error) {
```
g-gaston, line 14: Does this need to be public?

pkg/providers/snow/reconciler/clientbuilder_test.go
```diff
@@ -0,0 +1,99 @@
...
+			secretFile: "testdata/invalid_secret_missing_region",
+			wantErr:    "parsing configuration for 1.2.3.5: unable to set region",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			ctx := context.Background()
+
+			secret, err := readSecret(tt.secretFile)
```
g-gaston, line 55: I feel like it would be more clear to just define the secrets with structs in the test code and avoid testdata files

pkg/providers/snow/reconciler/clientbuilder.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	deviceClientMap := make(aws.Clients, len(deviceIps))
+	for _, ip := range deviceIps {
+		ipCfg, err := parseIpConfiguration(credsCfg, ip)
+		if err != nil {
+			return nil, fmt.Errorf("parsing configuration for %v: %v", ip, err)
+		}
+		clientCfg, err := awsconfig.LoadDefaultConfig(ctx,
```
g-gaston, line 61: curious why we use a different config than when building the clients from the cli?

pkg/providers/snow/reconciler/clientbuilder.go
g-gaston, line 43: This seems to be similar to `BuildSnowAwsClientMap`. Is there a way to reuse code do they have to work very differently?

pkg/providers/snow/reconciler/clientbuilder.go
```diff
@@ -0,0 +1,97 @@
...
+	client client.Client
+}
+
+func NewAwsClientBuilder(client client.Client) *awsClientBuilder {
+	return &awsClientBuilder{
+		client: client,
+	}
+}
+
+func (b *awsClientBuilder) BuildSnowAwsClientMap(ctx context.Context) (aws.Clients, error) {
```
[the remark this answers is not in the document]
vivek-koppuru: This is to be able to interchange it with the following, but I can probably change the name for both: https://github.com/aws/eks-anywhere/blob/main/pkg/aws/snow.go#L23

pkg/providers/snow/reconciler/clientbuilder.go
vivek-koppuru: We load it from a secret as data instead of files, so the whole logic of creating the client actually changes.

pkg/providers/snow/reconciler/clientbuilder.go
vivek-koppuru: As described in the above comment, I couldn't find a way to reuse the logic so I decided to keep it completely separate until we have a better idea on how to share the logic.

pkg/providers/snow/reconciler/clientbuilder.go
g-gaston: Got it. But once we load the data, shouldn't the rest of the logic be the same. In other words, is it possible to separate "data load" from "client construction"? And then reuse the latter in both scenarios

pkg/providers/snow/reconciler/clientbuilder_test.go
```diff
@@ -0,0 +1,178 @@
...
+	. "github.com/onsi/gomega"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere/pkg/providers/snow/reconciler"
+)
+
+const validSecret = `apiVersion: v1
```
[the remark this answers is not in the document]
vivek-koppuru: Yea I had a feeling that was what you were mentioning but went with this approach to keep base64 data. I don't think it makes a different, I should have went with this to begin with

pkg/providers/snow/reconciler/clientbuilder.go
vivek-koppuru: It doesn't because of how you load this config from this sdk. I am using static credentials provider instead of a shared credentials provider through file inputs. Here is the comparison: https://github.com/aws/eks-anywhere/blob/0c56cd3dfe592057433c76977e96172afd5438d4/pkg/aws/snow.go#L79

PR #2862, opened 2022-08-03
Skeleton for snow reconciler

taneyland:
*Issue #, if available:*

*Description of changes:*
Skeletion for the snow cluster reconciler

*Testing (if applicable):*


pkg/controller/clusters/registry.go
```diff
@@ -44,9 +44,9 @@ func NewProviderClusterReconcilerRegistryBuilder() *ProviderClusterReconcilerReg
 	}
 }
 
-// Add accumulates a pair of datacenter kind a reconciler to be included in the final registry
-func (b *ProviderClusterReconcilerRegistryBuilder) Add(datacenterKind string, reconciler ProviderClusterReconciler) *ProviderClusterReconcilerRegistryBuilder {
-	b.reconciler.add(datacenterKind, reconciler)
+// Add accumulates a pair of kind and reconciler to be included in the final registry
+func (b *ProviderClusterReconcilerRegistryBuilder) Add(kind string, reconciler ProviderClusterReconciler) *ProviderClusterReconcilerRegistryBuilder {
```
taneyland, line 48: Changed this to `kind` because it seems to vary. For Snow, we focus on the the MachineConfig kind, right? @g-gaston

vivek-koppuru: My understanding was that we use datacenterKind from the Cluster spec to determine what provider to use to keep the logic we had right now. But following this thread as well.

g-gaston: Yeah, we should use the datacenterkind. As vivek was saying, that's the only thing we can use to determine the provider with the current api. Not all providers have machine configs.

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/aws/eks-anywhere/pkg/controller"
+)
+
+type SnowReconciler struct{}
+
+func NewSnowReconciler() *SnowReconciler {
+	return &SnowReconciler{}
+}
+
+func (v *SnowReconciler) Reconcile(ctx context.Context, cluster *anywherev1.Cluster) (controller.Result, error) {
```
g-gaston, line 16: this is already in the snow (sub) package
```suggestion
type Reconciler struct{}

func New() *Reconciler {
	return &Reconciler{}
}

func (r *Reconciler) Reconcile(ctx context.Context, cluster *anywherev1.Cluster) (controller.Result, error) {
```

pkg/controller/clusters/registry.go
taneyland: Got it. Makes sense. Will change it back

g-gaston (approved):

PR #2863, opened 2022-08-03
E2E for metallb

lewisdiamond:
*Description of changes:*
MetalLB e2e tests

(Some reformatting done by golines which I can remove if need be but I think we should use golines)

### DIFF
--- a/test/e2e/metallb_test.go
+++ b/test/e2e/metallb_test.go
@@ -1,0 +1,278 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/test/framework"
+	"github.com/stretchr/testify/suite"
+)
+
+type Suite struct {
+	suite.Suite
+	cluster *framework.ClusterE2ETest
+}
+
+func TestPackagesMetalLB(t *testing.T) {
+	suite.Run(t, new(Suite))
+}
+
+func (suite *Suite) SetupSuite() {
+	t := suite.T()
+	suite.cluster = framework.NewClusterE2ETest(t, framework.NewDocker(t),
+		framework.WithPackageConfig(t, EksaPackageBundleURI, EksaPackageControllerHelmChartName,
+			EksaPackageControllerHelmURI, EksaPackageControllerHelmVersion,
+			EksaPackageControllerHelmValues),
+	)
+}
+
+func (suite *Suite) TestPackagesMetalLB() {
+	// This should be split into multiple tests with a cluster setup in `SetupSuite`.
+	// This however requires the creation of utilites managing cluster creation.
+	t := suite.T()
+	suite.cluster.WithPersistentCluster(func(test *framework.ClusterE2ETest) {
+		test.InstallCuratedPackagesController()
+		ctx := context.Background()
+		namespace := "metallb-system"
+		t.Cleanup(func() { test.DeleteNamespace(namespace) })
+		test.CreateNamespace(namespace)
+		packageName := "metallb"
+		packagePrefix := "test"
+		err := WaitForLatestBundleToBeAvailable(test, ctx, 2*time.Minute)
+		if err != nil {
+			t.Fatalf("waiting for latest bundle: %s", err)
+		}
+		bundle, err := GetLatestBundleFromCluster(test)
+		if err != nil {
+			t.Fatal(err)
+		}
+		UpgradePackages(test, bundle)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Run("Basic installation", func(t *testing.T) {
+			t.Cleanup(func() { test.UninstallCuratedPackage(packagePrefix) })
+			test.InstallCuratedPackage(packageName, packagePrefix, "--kube-version=1.22")
+			err = WaitForPackageToBeInstalled(test, ctx, packagePrefix, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb package to be installed: %s", err)
+			}
+			err = test.KubectlClient.WaitForDeployment(context.Background(),
+				kubeconfig.FromClusterName(test.ClusterName), "1m", "Available", "test-metallb-controller", namespace)
+			if err != nil {
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+			err = WaitForDaemonset(test, ctx, "test-metallb-speaker", namespace, 2, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+		})
+
+		t.Run("Address pool configuration", func(t *testing.T) {
+			address := "10.100.100.1"
+			t.Cleanup(func() { test.UninstallCuratedPackage(packagePrefix) })
+			test.CreateResource(ctx, fmt.Sprintf(
+				`
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: Package
+metadata:
+  name: test
+  namespace: eksa-packages
+spec:
+  packageName: metallb
+  config: |
+    IPAddressPools:
+      - name: default
+        addresses:
+          - %s/32
+    L2Advertisements:
+      - IPAddressPools:
+        - default
+`, address))
+			err = WaitForPackageToBeInstalled(test, ctx, packagePrefix, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb package to be installed: %s", err)
+			}
+			err = test.KubectlClient.WaitForDeployment(context.Background(),
+				kubeconfig.FromClusterName(test.ClusterName), "1m", "Available", "test-metallb-controller", namespace)
+			if err != nil {
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+			err = WaitForDaemonset(test, ctx, "test-metallb-speaker", namespace, 2, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb speaker deployment to be available: %s", err)
+			}
+
+			expectedConfig := fmt.Sprintf(
+				`address-pools:
+  - name: default
+    addresses:
+      - %s/32
+    protocol: layer2
+`, address)
+			err = WaitForResource(
+				test,
+				ctx,
+				"configmap/test-metallb",
+				namespace,
+				"{.data.config}",
+				20*time.Second,
+				NoErrorPredicate,
+				StringMatchPredicate(expectedConfig),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			t.Cleanup(func() {
+				test.KubectlClient.Delete(ctx, "service", "my-service", "default", kubeconfig.FromClusterName(test.ClusterName))
+			})
+			test.CreateResource(ctx, `
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+spec:
+  type: LoadBalancer
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+`)
+			err = WaitForResource(
+				test,
+				ctx,
+				"service/my-service",
+				"default",
+				"{.status.loadBalancer.ingress[0].ip}",
+				20*time.Second,
+				NoErrorPredicate,
+				StringMatchPredicate(address),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+		})
+		t.Run("BGP configuration", func(t *testing.T) {
+			address := "10.100.100.2"
+			t.Cleanup(func() { test.UninstallCuratedPackage(packagePrefix) })
+			test.CreateResource(ctx, fmt.Sprintf(
+				`
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: Package
+metadata:
+  name: test
+  namespace: eksa-packages
+spec:
+  packageName: metallb
+  config: |
+    IPAddressPools:
+      - name: default
+        addresses:
+          - 10.100.0.1/32
+        autoAssign: false
+      - name: bgp
+        addresses:
+          - %s/32
+    L2Advertisements:
+      - IPAddressPools:
+        - default
+    BGPAdvertisements:
+      - IPAddressPools:
+          - bgp
+        localPref: 123
+        aggregationLength: 32
+        aggregationLengthV6: 32
+    BGPPeers:
+      - myASN: 123
+        peerASN: 55001
+        peerAddress: 12.2.4.2
+        keepaliveTime: 30s
+`, address))
+			err = WaitForPackageToBeInstalled(test, ctx, packagePrefix, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb package to be installed: %s", err)
+			}
+			err = test.KubectlClient.WaitForDeployment(context.Background(),
+				kubeconfig.FromClusterName(test.ClusterName), "1m", "Available", "test-metallb-controller", namespace)
+			if err != nil {
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+			err = WaitForDaemonset(test, ctx, "test-metallb-speaker", namespace, 2, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb speaker deployment to be available: %s", err)
+			}
+
+			expectedConfig := fmt.Sprintf(
+				`address-pools:
+  - name: default
+    addresses:
+      - 10.100.0.1/32
+    protocol: layer2
+    auto-assign: false
+  - name: bgp
+    addresses:
+      - %s/32
+    protocol: bgp
+    bgp-advertisements:
+      - aggregation-length: 32
+        aggregation-length-v6: 32
+        localpref: 123
+peers:
+  - peer-address: 12.2.4.2
+    peer-asn: 55001
+    my-asn: 123
+    keepalive-time: 30s
+`, address)
+			err = WaitForResource(
+				test,
+				ctx,
+				"configmap/test-metallb",
+				namespace,
+				"{.data.config}",
+				20*time.Second,
+				NoErrorPredicate,
+				StringMatchPredicate(expectedConfig),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			t.Cleanup(func() {
+				test.KubectlClient.Delete(ctx, "service", "my-service", "default", kubeconfig.FromClusterName(test.ClusterName))
+			})
+			test.CreateResource(ctx, `
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+spec:
+  type: LoadBalancer
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+`)
+			err = WaitForResource(
+				test,
+				ctx,
+				"service/my-service",
+				"default",
+				"{.status.loadBalancer.ingress[0].ip}",
+				20*time.Second,
+				NoErrorPredicate,
+				StringMatchPredicate(address),
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+		})
+	})
+}
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -161,7 +161,11 @@ func ExpectFailure(expected bool) ClusterE2ETestOpt {
 }
 
 func WithControlPlaneHardware(requiredCount int) ClusterE2ETestOpt {
-	return withHardware(requiredCount, api.ControlPlane, map[string]string{api.HardwareLabelTypeKeyName: api.ControlPlane})
+	return withHardware(
+		requiredCount,
+		api.ControlPlane,
+		map[string]string{api.HardwareLabelTypeKeyName: api.ControlPlane},
+	)
 }
 
 func WithWorkerHardware(requiredCount int) ClusterE2ETestOpt {
@@ -173,7 +177,11 @@ func WithCustomLabelHardware(requiredCount int, label string) ClusterE2ETestOpt 
 }
 
 func WithExternalEtcdHardware(requiredCount int) ClusterE2ETestOpt {
-	return withHardware(requiredCount, api.ExternalEtcd, map[string]string{api.HardwareLabelTypeKeyName: api.ExternalEtcd})
+	return withHardware(
+		requiredCount,
+		api.ExternalEtcd,
+		map[string]string{api.HardwareLabelTypeKeyName: api.ExternalEtcd},
+	)
 }
 
 func (e *ClusterE2ETest) GetHardwarePool() map[string]*api.Hardware {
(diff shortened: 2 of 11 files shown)


cmd/eksctl-anywhere/cmd/upgradepackages.go
```diff
@@ -46,20 +46,5 @@ func upgradePackages(ctx context.Context) error {
...
-	activeController, err := b.GetActiveController(ctx)
-	if err != nil {
-		return err
-	}
-	err = b.UpgradeBundle(ctx, activeController, upo.bundleVersion)
-	if err != nil {
-		return err
-	}
-	return nil
+	return curatedpackages.UpgradePackages(ctx, kubeConfig, deps.Kubectl, string(upo.source), upo.bundleVersion)
```
a-cool-train, line 49: Is string conversion required here? The field is already a string.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -153,3 +154,25 @@ func GetRegistry(uri string) string {
...
+	targetBundleVersion string,
+) error {
+	b := NewBundleReader(
+		kubeConfig,
+		BundleSource(source),
+		kubectl,
+		nil,
+		nil,
+	)
+	activeController, err := b.GetActiveController(ctx)
```
a-cool-train, line 172: We might want to have some test coverage for this method.

test/e2e/metallb_test.go
```diff
@@ -0,0 +1,279 @@
...
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+		})
+
+		t.Run("Address pool configuration", func(t *testing.T) {
+			address := "10.100.100.1"
+			t.Cleanup(func() { test.UninstallCuratedPackage(packagePrefix) })
+			test.CreateResource(ctx, fmt.Sprintf(
+				`
+apiVersion: packages.eks.amazonaws.com/v1alpha1
```
a-cool-train, line 84: The current that is followed in eks-anywhere is to use files for testdata. Should we follow that here?

pkg/curatedpackages/curatedpackages.go
a-cool-train: The general structure is to have tests in pkg/ but we don't have any tests for cmd/ package. The orchestration is done in cmd/ and the business functionality done in pkg/. I can add the needed tests accordingly since this is just orchestration.

cmd/eksctl-anywhere/cmd/upgradepackages.go
lewisdiamond: It's required if no modification is made to bundlesource. I can add it to this PR.

test/e2e/metallb_test.go
lewisdiamond: I decided to keep it like this given that we'll move all these e2e tests out (hopefully sooner than later).

pkg/curatedpackages/curatedpackages.go
```diff
@@ -153,3 +154,25 @@ func GetRegistry(uri string) string {
...
+	if err != nil {
+		return err
+	}
+	err = b.UpgradeBundle(ctx, activeController, targetBundleVersion)
+	return err
```
chrisdoherty4, line 177:
```suggestion
	return b.UpgradeBundle(ctx, activeController, targetBundleVersion)
```

test/e2e/curated_packages_test.go
```diff
@@ -4,37 +4,174 @@
...
+	EksaPackageBundleURI               = "oci://" + EksaPackagesSourceRegistry + "/eks-anywhere-packages-bundles"
 )
 
-var (
-	eksaPackageControllerHelmValues = []string{}
-	eksAnywherePackagesHelmValues   = []string{"sourceRegistry=public.ecr.aws/l0g8r8j6"}
-)
+var EksaPackageControllerHelmValues = []string{"sourceRegistry=public.ecr.aws/l0g8r8j6"}
+
+type resourcePredicate func(string, error) bool
```
chrisdoherty4, line 30: Need to export this if its used by the public API (`WaitForResource` consumes it).

chrisdoherty4, line 30: Why does a predicate need to care about an error?

test/e2e/curated_packages_test.go
```diff
@@ -4,37 +4,174 @@
...
+	packageName string,
+	timeout time.Duration,
+) error {
+	//--for=jsonpath isn't supported in v1.22. Update once it's supported
+	//_, err = test.KubectlClient.Execute(
+	//    ctx, "wait", "--timeout", "1m",
+	//    "--for", "jsonpath='{.status.state}'=installed",
+	//    "package", packagePrefix, "--kubeconfig", kubeconfig,
+	//    "-n", "eksa-packages",
+	//)
```
chrisdoherty4, line 132: Is the PR in draft?

test/e2e/curated_packages_test.go
```diff
@@ -4,37 +4,174 @@
...
+			allPredicates = f(trimmed, err) && allPredicates
+		}
+		if allPredicates {
+			return nil
+		}
+		time.Sleep(2 * time.Second)
+		now = time.Now()
+	}
+	return fmt.Errorf(
+		"Timed out waiting for resource: %s [namespace: %s, jsonpath: %s, timeout: %s]",
```
chrisdoherty4, line 82:
```suggestion
		"timed out waiting for resource: %s [namespace: %s, jsonpath: %s, timeout: %s]",
```

pkg/curatedpackages/curatedpackages.go
lewisdiamond: This commit only moves this code from https://github.com/aws/eks-anywhere/pull/2863/files/069fb7b7afe5364851ce8e91bcca8f3792a335c9#diff-55361c4daff82b0371cefa04eada4c480ef3b0a92e8cc93edd7dcbeee37438cfL49
I'd use another PR for this as this is not related to metallb E2E.

test/e2e/curated_packages_test.go
lewisdiamond: A predicate could be `hasError(...)` for example.

test/e2e/curated_packages_test.go
lewisdiamond: It can be removed but I'd like to keep it as we'll use that implementation once the cli image includes the kubernetes v1.23 client.

pkg/curatedpackages/curatedpackages.go
lewisdiamond: I actually had removed some redundancy already but forgot some after moving the code.

test/e2e/curated_packages_test.go
chrisdoherty4: Given `resourcePredicate` is unexported meaning we have to provide the predicates from this package, and that we don't have any predicates that act on the error at all, I suggest we simplify by removing it.

lewisdiamond: `NoErrorPredicate` does use the error input. We could indeed export `resourcePredicate` as well.

lewisdiamond: https://github.com/aws/eks-anywhere/pull/2863/files#diff-e04741f3b0c5b687abeb91625437530ddf182779c1b1b5fa4df95b9c59717c5eR32

chrisdoherty4: I recall seeing that and thinking it should be general error handling for `Execute(...)`. Why do we need a separate predicate for it?

lewisdiamond: It just allows to have test logic based on non-zero exit code and/or `stderr` from the forked process. If we use `kubectl wait --for` and it times out, we'll get a non-zero exit code with a stderr message which will be part of `error`.

chrisdoherty4: OK. Perhaps there's a broader change required. There's still something off here though - it sounds like a hard to use error return is dictating a package API.

lewisdiamond: I suppose it being an `error` might not be ideal compared to having `stdout`, `stderr` and the return code.

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,173 @@
...
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
 const (
-	eksaPackageControllerHelmChartName = "eks-anywhere-packages"
-	eksaPackageControllerHelmURI       = "oci://public.ecr.aws/eks-anywhere/eks-anywhere-packages"
-	eksaPackageControllerHelmVersion   = "0.1.10-eks-a-10"
-	eksaPackageBundleBaseURI           = "oci://public.ecr.aws/eks-anywhere/eks-anywhere-packages-bundles"
+	EksaPackagesNamespace              = "eksa-packages"
```
ewollesen, line 20: These are test-specific values and should not be exported. The could probably be renamed and make it obvious that they're meant for the hello-eks-anywhere test, but these were not exported for a reason---they don't apply to other tests!

At the time I wrote these tests, there were no other artifacts against which I could test, so these were hard-coded in. It wasn't the best decision, but it was the only choice I had at the time.

Borrowing these values is not the right way forward, unless it's also the only way forward for metallb as well? And even still, I wouldn't recommend exporting these and using them _anywhere_ else.

test/e2e/curated_packages_test.go
ewollesen: Two thoughts:

1. it seems poorly named, and should be documented if a more obvious name isn't forthcoming
2. if it's exported, it should probably
    a. have more documentation for future possible users 
    b. live in a shared utils package, for better discoverability

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,173 @@
...
+	timeout time.Duration,
+	predicates ...resourcePredicate,
+) error {
+	start := time.Now()
+	end := start.Add(timeout)
+	if !strings.HasPrefix(jsonpath, "jsonpath") {
+		jsonpath = fmt.Sprintf("jsonpath='%s'", jsonpath)
+	}
+	command := []string{}
+	for now := start; now.Before(end); now = time.Now() {
```
ewollesen, line 65:
```suggestion
	for time.Now().Before(end) {
```

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,173 @@
...
+	test *framework.ClusterE2ETest,
+	ctx context.Context,
+	resource string,
+	namespace string,
+	jsonpath string,
+	timeout time.Duration,
+	predicates ...resourcePredicate,
+) error {
+	start := time.Now()
+	end := start.Add(timeout)
```
ewollesen, line 60:
```suggestion
	end := time.Now().Add(timeout)
```
Assumes you accept the change below.

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,173 @@
...
+	command := []string{}
+	for now := start; now.Before(end); now = time.Now() {
+		out, err := test.KubectlClient.Execute(ctx, "get", "-n", namespace,
+			"--kubeconfig="+test.Cluster().KubeconfigFile,
+			"-o", jsonpath, resource)
+		outStr := out.String()
+		trimmed := strings.Trim(outStr, "'")
+		allPredicates := true
+		for _, f := range predicates {
+			allPredicates = f(trimmed, err) && allPredicates
```
ewollesen, line 73:
```suggestion
			allPredicates = allPredicates && f(trimmed, err)
```
Nitpick, but allows for short-circuiting.

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,173 @@
...
+		}
+		if allPredicates {
+			return nil
+		}
```
ewollesen, line 77:
```suggestion
		for _, f := range predicates {
			if !f(trimmed, err) {
				allPredicates = false
				break
			}                        
		}
		if allPredicates {
			return nil
		}
```

It's an extra line of code, but maybe it's a little simpler to read?

I don't have a strong feeling either way, you can ignore both of the comments in the area really.

cmd/eksctl-anywhere/cmd/upgradepackages.go
```diff
@@ -46,20 +46,5 @@ func upgradePackages(ctx context.Context) error {
...
-	activeController, err := b.GetActiveController(ctx)
-	if err != nil {
-		return err
-	}
-	err = b.UpgradeBundle(ctx, activeController, upo.bundleVersion)
-	if err != nil {
-		return err
-	}
-	return nil
+	return curatedpackages.UpgradePackages(ctx, kubeConfig, deps.Kubectl, ipo.source, upo.bundleVersion)
```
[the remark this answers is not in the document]
lewisdiamond: I matched the input from the code that was already there. Can you elaborate on what the bug is?

pkg/executables/helm.go
```diff
@@ -61,13 +62,27 @@ func NewHelm(executable Executable, opts ...HelmOpt) *Helm {
 	return h
 }
 
-func (h *Helm) Template(ctx context.Context, ociURI, version, namespace string, values interface{}, kubeVersion string) ([]byte, error) {
+func (h *Helm) Template(
```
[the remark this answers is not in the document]
lewisdiamond: I mentioned it in the PR first comment:
`(Some reformatting done by golines which I can remove if need be but I think we should use golines)`

Regardless of preference, using an automated approach is useful to avoid changes hidden by reformatting. This is included because I modified this file and `golines` made the modifications and people using different formats will be a source of similar issues going forward.

pkg/executables/helm.go
```diff
@@ -142,6 +161,18 @@ func (h *Helm) InstallChartWithValuesFile(ctx context.Context, chart, ociURI, ve
...
 
+func (h *Helm) ListCharts(ctx context.Context, kubeconfigFilePath string) ([]string, error) {
+	params := []string{"list", "-q", "--kubeconfig", kubeconfigFilePath}
+	out, err := h.executable.Command(ctx, params...).WithEnvVars(h.env).Run()
+	if err != nil {
+		return nil, err
+	}
+	charts := strings.FieldsFunc(out.String(), func(c rune) bool {
+		return c == '\n'
+	})
```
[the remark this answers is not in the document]
lewisdiamond: It's not quite equivalent as `strings.Split` returns an empty element if the first character is `\n`.

pkg/executables/helm.go
```diff
@@ -142,6 +161,18 @@ func (h *Helm) InstallChartWithValuesFile(ctx context.Context, chart, ociURI, ve
...
+	return charts, err
```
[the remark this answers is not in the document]
lewisdiamond: I have no opinion either way.

cmd/eksctl-anywhere/cmd/upgradepackages.go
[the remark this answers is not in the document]
ewollesen: In the suggestion I made, you can see that I've changed "ipo" to "upo". That's basically the bug. The variable "ipo", is a package-scoped variable in the "curatedpackages" package. It's name, "ipo", stands for "install package options". It's a struct containing a number of configuration options used when installing packages (obviously, I guess :)). Anyway, it has a member, "source". Meanwhile, there's another package-scoped variable in the same package, named "upo". It stands for "upgrade package options". I imagine you can guess what it's purpose is. It too contains a member named "source". However, it's value for source is different from "ipo"'s value for source. Since they're both package-scoped variables that are structs containing members named source, the compiler can't know that you want the one versus the other, so it doesn't even know to warn you. All of our CLI commands use this kind of package-scope variable for options and the command definitions, and they really should be refactored not to, for this very reason.

Since this line of code (line #49) is calling "UpgradePackages", it should be using the upgrade package options, ie "upo.source", as opposed to the install package options source, which would be "ipo.source". If for no other reason, than because "ipo.source" is probably a default or zero-value, whereas "upo.source" should contain whatever value was indicated by the user on the command-line.

pkg/executables/helm.go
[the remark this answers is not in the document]
ewollesen: As I see it there are two issues here. The first is the use of the golines tool. I think this is the wrong way to introduce the tool. I think it should be brought up with and discussed with the team. If the team decides to use it, someone should then reformat all the files in one fell swoop, so that reviewers know that the only changes expected are non-functional whitespace changes.

The second issue is that this change is unrelated to the subject of the PR. These kinds of changes shouldn't be included in PRs that have functional changes, which, as you wrote (and yes, I read what you wrote), you know, yet you did it anyway (why?). I'm not asking you to remove this change; I've already looked at it, verified that it's a whitespace only change and unrelated. My comment was meant to reinforce what I'm guessing you already know; that these aren't good changes to include, and I'd be grateful if we all tried not to include them in the future.

As an aside, you wrote:

> using an automated approach is useful to avoid changes hidden by reformatting

Your automated tool is the culprit here, introducing "false" changes that distract from what really is changing. Yes, people having different settings will cause these problems. Right now, the person with different settings... is you. 😛 

Nothing automated is required. People just need to not alter the formatting of things they aren't changing, and there won't be any such problems.

pkg/executables/helm.go
ewollesen, line 172:
```suggestion
	charts := strings.Split(strings.Trim(out.String()), "\n")
```

test/e2e/curated_packages_test.go
ewollesen: We had a voice chat, end result, we're leaving these here. I'd prefer if they were at least commented with a warning that they are here for convenience and any given test should use the values that make sense for them, not necessarily these values. But even that is hopefully moot since a good chunk of this will eventually change as we further develop our e2e test infrastructure.

test/e2e/metallb_test.go
```diff
@@ -0,0 +1,279 @@
...
+			err = WaitForPackageToBeInstalled(test, ctx, packagePrefix, 30*time.Second)
+			if err != nil {
+				t.Fatalf("waiting for metallb package to be installed: %s", err)
+			}
+			err = test.KubectlClient.WaitForDeployment(context.Background(),
+				test.Cluster(), "1m", "Available", "test-metallb-controller", namespace)
+			if err != nil {
+				t.Fatalf("waiting for metallb controller deployment to be available: %s", err)
+			}
+			err = WaitForDaemonset(test, ctx, "test-metallb-speaker", namespace, 2, 30*time.Second)
```
ewollesen, line 73: Nitpick: The "2" here is a magic number. It's in no way clear what it's purpose is without knowing the function arguments, and even knowing what it is, it's not clear why we're waiting for 2 nodes?

ewollesen: The 30 seconds at least we can imply from the function name.

test/e2e/metallb_test.go
```diff
@@ -0,0 +1,279 @@
...
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+`)
+			err = WaitForResource(
+				test,
+				ctx,
+				"service/my-service",
+				"default",
+				"{.status.loadBalancer.ingress[0].ip}",
```
ewollesen, line 155: This is brittle. I wonder if something using a function like [assert.JSONEq](https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq) could make this more flexible?

test/e2e/metallb_test.go
```diff
@@ -0,0 +1,279 @@
...
+`, address)
+			err = WaitForResource(
+				test,
+				ctx,
+				"configmap/test-metallb",
+				namespace,
+				"{.data.config}",
+				20*time.Second,
+				NoErrorPredicate,
+				StringMatchPredicate(expectedConfig),
```
ewollesen, line 243: This is brittle. Using [assert.YAMLEq](https://pkg.go.dev/github.com/stretchr/testify/assert#YAMLEq), it's cousin in the require package, or other YAML-aware comparator, would help this be more resilient.

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,172 @@
...
+	predicates ...resourcePredicate,
+) error {
+	end := time.Now().Add(timeout)
+	if !strings.HasPrefix(jsonpath, "jsonpath") {
+		jsonpath = fmt.Sprintf("jsonpath='%s'", jsonpath)
+	}
+	command := []string{}
+	for time.Now().Before(end) {
+		out, err := test.KubectlClient.Execute(ctx, "get", "-n", namespace,
+			"--kubeconfig="+test.Cluster().KubeconfigFile,
```
ewollesen, line 66:
```suggestion
			"--kubeconfig="+kubeconfig.FromClusterName(test.ClusterName),
```
This way you don't have to export Cluster().

test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,172 @@
...
+	)
+}
+
+func GetLatestBundleFromCluster(test *framework.ClusterE2ETest) (string, error) {
+	bundleBytes, err := test.KubectlClient.ExecuteCommand(
+		context.Background(),
+		"get",
+		"packagebundle",
+		"-n", "eksa-packages",
+		"--kubeconfig="+test.Cluster().KubeconfigFile,
```
ewollesen, line 149: See https://github.com/aws/eks-anywhere/pull/2863/files#r968881732

test/framework/cluster.go
```diff
@@ -727,7 +755,7 @@ func (e *ClusterE2ETest) cleanup(f func()) {
 	})
 }
 
-func (e *ClusterE2ETest) cluster() *types.Cluster {
+func (e *ClusterE2ETest) Cluster() *types.Cluster {
```
ewollesen, line 758: I'm not a huge fan of exporting this. It feels like it probably shouldn't exist anyway, because all it's doing for these tests is wrapping together a cluster name and a kubeconfig file. The former is already exported, and the latter can be determined via kubeconfig.FromClusterName, an already exported function. Exporting it just feels like reinforcing an already iffy behavior.

It seems that you wanted to export it to use it in WaitForResource and related functions, which I think would be a good candidate to move into the e2e framework package, and thereby wouldn't need cluster() to be exported? That or just don't use cluster() would be my preference.

test/framework/cluster.go
```diff
@@ -820,7 +848,11 @@ func setEksctlVersionEnvVar() error {
 	if eksctlVersionEnv == "" {
 		err := os.Setenv(eksctlVersionEnvVar, eksctlVersionEnvVarDummyVal)
```
ewollesen, line 849: Not your code, but this should almost certainly be using [e.T.Setenv](https://pkg.go.dev/testing#T.Setenv). If you feel like tweaking while you're there.

test/framework/cluster.go
```diff
@@ -830,22 +862,58 @@ func (e *ClusterE2ETest) InstallHelmChart() {
...
 	// TODO Add a test that installs the controller via the CLI.
 	ctx := context.Background()
-	err := e.PackageConfig.HelmClient.InstallChart(ctx,
-		e.PackageConfig.chartName, e.PackageConfig.chartURI,
-		e.PackageConfig.chartVersion, kubeconfig, e.PackageConfig.chartValues)
+	charts, err := e.PackageConfig.HelmClient.ListCharts(ctx, kubeconfig)
 	if err != nil {
-		e.T.Fatalf("Error installing %s helm chart on the cluster: %v",
-			e.PackageConfig.chartName, err)
+		e.T.Fatalf("Unable to list charts: %v", err)
```
ewollesen, line 900: This is better handled with two separate functions, IMO:

IsCuratedPackagesControllerInstalled and InstallCuratedPackagesController

The behavior of this function doesn't align well with its name. Having two functions is simpler, and reads more obviously.

test/e2e/metallb_test.go
```diff
@@ -0,0 +1,279 @@
...
+			EksaPackageControllerHelmValues),
+	)
+}
+
+func (suite *Suite) TestPackagesMetalLB() {
+	// This should be split into multiple tests with a cluster setup in `SetupSuite`.
+	// This however requires the creation of utilites managing cluster creation.
+	t := suite.T()
+	suite.cluster.WithCluster(func(test *framework.ClusterE2ETest) {
+		test.InstallCuratedPackagesController()
```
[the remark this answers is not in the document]
ewollesen: OK, it won't because you've changed the behavior of InstallCuratedPackagesController, which makes this really more of a MaybeInstallCuratedPackagesController.

I'd prefer to see:

```go
if !test.IsCuratedPackagesControllerInstalled(){
    test.InstallCuratedPackagesController()
}
```

cmd/eksctl-anywhere/cmd/upgradepackages.go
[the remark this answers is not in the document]
lewisdiamond: As it is today, `upgradePackageOptions` only has one field (`bundleVersion`) and no `source` so I can't apply the suggestiong as it won't compile. Re-structuring the CLI options is something I'd consider out-of-scope for this PR.

lewisdiamond:
Will open a new one this week with updated tests for newer versions. Closing this.

PR #2864, opened 2022-08-03
Add nil check to clusterConfig processors

raymond-zhang00:
*Issue #, if available:*

Objects.GetFromRef() can return a nil value that cannot be cast to a provider datacenter datatype.

*Description of changes:*
Added nil check to prevent casting in provider processors.

*Testing (if applicable):*


raymond-zhang00:
closing - this change will be added to preflight validate cmd

raymond-zhang00:
reopening - breaking this out of preflight validate into it's own branch

PR #2872, opened 2022-08-04
Make gitOpsRef mutable in upgrade preflight check

jiayiwang7:
*Issue #, if available:*

#2755 

*Description of changes:*

Enable flux upgrade in preflight flux immutability validation:
- old gitOpsRef nil, new gitOpsRef not nil -- validation succeeds, as we support installing flux in existing cluster
- old gitOpsRef not nil, new gitOpsRef nil -- validation fails, as we don't support removing flux in existing cluster rn
- both old and new gitOpsRef nil -- validation succeeds
- both old and new gitOpsRef not nil -- compare objects and fail if diff

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/validations/upgradevalidations/immutableFields.go
+++ b/pkg/validations/upgradevalidations/immutableFields.go
@@ -126,13 +126,13 @@ func ValidateGitOpsImmutableFields(ctx context.Context, k validations.KubectlCli
 		if prevGitOps.Spec.Flux.Github.Personal != clusterSpec.GitOpsConfig.Spec.Flux.Github.Personal {
 			return errors.New("gitOps spec.flux.github.personal is immutable")
 		}
-		if clusterSpec.GitOpsConfig.Spec.Flux.Github.FluxSystemNamespace != "" && prevGitOps.Spec.Flux.Github.FluxSystemNamespace != clusterSpec.GitOpsConfig.Spec.Flux.Github.FluxSystemNamespace {
+		if prevGitOps.Spec.Flux.Github.FluxSystemNamespace != clusterSpec.GitOpsConfig.Spec.Flux.Github.FluxSystemNamespace {
 			return errors.New("gitOps spec.flux.github.fluxSystemNamespace is immutable")
 		}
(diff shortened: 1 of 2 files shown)


pkg/validations/upgradevalidations/immutableFields.go
```diff
@@ -169,3 +100,83 @@ func ValidateImmutableFields(ctx context.Context, k validations.KubectlClient, c
...
+		if prevGitOps.Spec.Flux.Github.Owner != clusterSpec.GitOpsConfig.Spec.Flux.Github.Owner {
+			return errors.New("gitOps spec.flux.github.owner is immutable")
+		}
+		if prevGitOps.Spec.Flux.Github.Repository != clusterSpec.GitOpsConfig.Spec.Flux.Github.Repository {
+			return errors.New("gitOps spec.flux.github.repository is immutable")
+		}
+		if prevGitOps.Spec.Flux.Github.Personal != clusterSpec.GitOpsConfig.Spec.Flux.Github.Personal {
+			return errors.New("gitOps spec.flux.github.personal is immutable")
+		}
+		if clusterSpec.GitOpsConfig.Spec.Flux.Github.FluxSystemNamespace != "" && prevGitOps.Spec.Flux.Github.FluxSystemNamespace != clusterSpec.GitOpsConfig.Spec.Flux.Github.FluxSystemNamespace {
```
danbudris, line 129: what happens if the new cluster config `GitOpsConfig` spec includes a `""` `FluxSystemNamespace` (or `Branch`, or `ClusterConfigPath`) but the previous cluster spec `prevGitOps` has it set to a non-nil value? I believe that golang short-circuts logical AND so we'd only check if the new value is `""` and then never check if the old value was non-nil or equal, right? I realize this is just moving the old validations code and I very well might have written these validations :)

danbudris: talked to Joey -- will confirm that the defaults are set prior to the validation execution which would make this a moot point

danbudris (approved):
/lgtm

PR #2875, opened 2022-08-04
Add cilium upgrade planner

g-gaston:
*Description of changes:*
In combination with the cilium installation, this allows to represent
the currently installed Cilium components and make an upgrade plan based
on the desired state defined in a cluster Spec.

*Testing (if applicable):*
Unit tests

### DIFF
--- a/pkg/networking/cilium/installation.go
+++ b/pkg/networking/cilium/installation.go
@@ -1,0 +1,14 @@
+package cilium
+
+import appsv1 "k8s.io/api/apps/v1"
+
+// Installation represents the Cilium components installed in a cluster
+type Installation struct {
+	DaemonSet *appsv1.DaemonSet
+	Operator  *appsv1.Deployment
+}
+
+// Installed determines if all Cilium components are present
+func (i Installation) Installed() bool {
+	return i.DaemonSet != nil && i.Operator != nil
+}
--- a/pkg/networking/cilium/installation_test.go
+++ b/pkg/networking/cilium/installation_test.go
@@ -1,0 +1,52 @@
+package cilium_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+
+	"github.com/aws/eks-anywhere/pkg/networking/cilium"
+)
+
+func TestInstallationInstalled(t *testing.T) {
+	tests := []struct {
+		name         string
+		installation cilium.Installation
+		want         bool
+	}{
+		{
+			name: "installed",
+			installation: cilium.Installation{
+				DaemonSet: &appsv1.DaemonSet{},
+				Operator:  &appsv1.Deployment{},
+			},
+			want: true,
+		},
+		{
+			name: "ds not installed",
+			installation: cilium.Installation{
+				Operator: &appsv1.Deployment{},
+			},
+			want: false,
+		},
+		{
+			name: "operator not installed",
+			installation: cilium.Installation{
+				DaemonSet: &appsv1.DaemonSet{},
+			},
+			want: false,
+		},
+		{
+			name:         "none installed",
+			installation: cilium.Installation{},
+			want:         false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			g.Expect(tt.installation.Installed()).To(Equal(tt.want))
+		})
+	}
+}
--- a/pkg/networking/cilium/upgrade_plan.go
+++ b/pkg/networking/cilium/upgrade_plan.go
@@ -1,0 +1,118 @@
+package cilium
+
+import (
+	"fmt"
+	"strings"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+)
+
+// UpgradePlan contains information about a Cilium installation upgrade
+type UpgradePlan struct {
+	DaemonSet ComponentUpgradePlan
+	Operator  ComponentUpgradePlan
+}
+
+// Needed determines if an upgrade is needed or not
+// Returns true if any of the installation components needs an upgrade
+func (c UpgradePlan) Needed() bool {
+	return c.DaemonSet.Needed() || c.Operator.Needed()
+}
+
+// Reason returns the reason why an upgrade might be needed
+// If no upgrade needed, returns empty string
+// For multiple components with needed upgrades, it composes their reasons into one
+func (c UpgradePlan) Reason() string {
+	if !c.Needed() {
+		return ""
+	}
+
+	s := make([]string, 0, 2)
+	if c.DaemonSet.UpgradeReason != "" {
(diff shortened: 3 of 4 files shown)


pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func daemonSetUpgradeInfo(ds *appsv1.DaemonSet, clusterSpec *cluster.Spec) ComponentUpgradeInfo {
+	dsImage := clusterSpec.VersionsBundle.Cilium.Cilium.VersionedImage()
+	i := ComponentUpgradeInfo{
+		NewImage: dsImage,
+	}
+
+	if ds == nil {
+		i.Reason = "Cilium needs upgrade, DS doesn't exist"
```
taneyland, line 72:
```suggestion
		i.Reason = "Cilium needs upgrade, DaemonSet doesn't exist"
```

pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+	oldDSImage := ds.Spec.Template.Spec.Containers[0].Image
+	i.OldImage = oldDSImage
+
+	containers := make([]corev1.Container, 0, len(ds.Spec.Template.Spec.Containers)+len(ds.Spec.Template.Spec.InitContainers))
+	containers = append(containers, ds.Spec.Template.Spec.Containers...)
+	containers = append(containers, ds.Spec.Template.Spec.InitContainers...)
+	for _, c := range containers {
+		if c.Image != dsImage {
+			i.OldImage = c.Image
+			i.Reason = fmt.Sprintf("Cilium needs upgrade, DS container %s doesn't match image", c.Name)
```
taneyland, line 85:
```suggestion
			i.Reason = fmt.Sprintf("Cilium needs upgrade, DaemonSet container %s doesn't match image", c.Name)
```

pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+	if len(operator.Spec.Template.Spec.Containers) == 0 {
+		i.Reason = "Operator deployment doesn't have any containers"
+		return i
+	}
+
+	oldImage := operator.Spec.Template.Spec.Containers[0].Image
+	i.OldImage = oldImage
+
+	if oldImage != newImage {
+		i.Reason = "Operator container doesn't match image"
```
d8660091, line 113: Sounds a little blunt, "Image used in Operator doesn't match the image specified by clusterSpec"?

pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	oldDSImage := ds.Spec.Template.Spec.Containers[0].Image
+	i.OldImage = oldDSImage
+
+	containers := make([]corev1.Container, 0, len(ds.Spec.Template.Spec.Containers)+len(ds.Spec.Template.Spec.InitContainers))
+	containers = append(containers, ds.Spec.Template.Spec.Containers...)
+	containers = append(containers, ds.Spec.Template.Spec.InitContainers...)
+	for _, c := range containers {
+		if c.Image != dsImage {
```
d8660091, line 83: Looks like Cilium always needs upgrade when the daemonSet has more than 1 containers. Is this intentional?

pkg/networking/cilium/upgrade_checker.go
d8660091, line 85: "Cilium needs upgrade, DS container %s doesn't match the image **specified in clusterSpec**", some context about "image"?

pkg/networking/cilium/upgrade_checker.go
d8660091: Assuming containers have different images

pkg/networking/cilium/upgrade_checker.go
g-gaston: I prefer not to include terms like clusterSpec in information messages, that's an implementation detail. IMO logs should help the reader understand what the code is doing not how the code is written.

pkg/networking/cilium/upgrade_checker.go
g-gaston: The image information is already included in the NewImage and OldImage fields

pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+// ComponentUpgradeInfo contains upgrade information for a Cilium component
+type ComponentUpgradeInfo struct {
+	Reason   string
+	OldImage string
+	NewImage string
+}
+
+// Needed determines if an upgrade is needed or not
+func (c ComponentUpgradeInfo) Needed() bool {
+	return c.Reason != ""
```
vivek-koppuru, line 53: Why use Reason than another field that specifies if it's latest or if it needs an upgrade or something?

pkg/networking/cilium/upgrade_checker.go
g-gaston: They shouldn't use different images, that's the thing. They all have to use the same image.

pkg/networking/cilium/upgrade_checker.go
g-gaston: Not sure  I get the question. Are you saying why we don't have a specific field to indicate if upgrade is needed? If that's the case: I started with a boolean for that reason, but then `Reason` and that flag always were set at the same time so the flag seemed redundant.

If we ever want to add one, we just need to change this method, that's why I added (vs checking `c.Reason != ""` all over the code).

pkg/networking/cilium/upgrade_checker.go
g-gaston: You can check how the cilium-cli does it:
https://github.com/cilium/cilium-cli/blob/592745bd8fc4be3ba024f1732fca8301ffb0c954/install/upgrade.go#L53-L63

pkg/networking/cilium/upgrade_checker.go
```diff
@@ -0,0 +1,118 @@
...
+			return i
+		}
+	}
+
+	return i
+}
+
+func operatorSetUpgradeInfo(operator *appsv1.Deployment, clusterSpec *cluster.Spec) ComponentUpgradeInfo {
+	newImage := clusterSpec.VersionsBundle.Cilium.Operator.VersionedImage()
+	i := ComponentUpgradeInfo{
```
mitalipaygude, line 95: nit: can we name the variable something more descriptive?

pkg/networking/cilium/upgrade_checker.go
vivek-koppuru: Yea I was more wondering whether having a Reason always means an upgrade is needed, which in that case, would `UpgradeReason` make more sense? Maybe I am having a trouble time visualizing but I assume Reason is more general.

g-gaston: Yes, `Reason` is only set when an upgrade is needed. I like `UpgradeReason`, changing

taneyland (approved):

d8660091 (approved):

vivek-koppuru (approved):

PR #2877, opened 2022-08-04
Add tooling to delete kube objects defined in yaml

g-gaston:
*Description of changes:*
Intended to be used from controllers with `client.Client`

*Testing (if applicable):*
Unit tests

### DIFF
--- a/pkg/controller/clientutil/delete.go
+++ b/pkg/controller/clientutil/delete.go
@@ -1,0 +1,35 @@
+package clientutil
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func DeleteYaml(ctx context.Context, c client.Client, yaml []byte) error {
+	objs, err := YamlToClientObjects(yaml)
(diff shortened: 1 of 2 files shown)


mitalipaygude (approved):

pkg/controller/clientutil/delete.go
```diff
@@ -0,0 +1,35 @@
+package clientutil
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func DeleteYaml(ctx context.Context, c client.Client, yaml []byte) error {
```
vivek-koppuru, line 10: Do these need to exactly match to be deleted? Or is it based on the resource kind and name? I know you are adding this tooling but what is an example in which this is used?

g-gaston: It will be used in the cilium reconciler, to delete the preflight objects we generate with helm.

It should only need to match resource type (kind and api version), name and namespace

vivek-koppuru (approved):

Issue #2878, opened 2022-08-04
--insecure-skip-tls-verify is always used when downloading helm charts from `download images` command


wongni:
**What would you like to be added**:
Add --insecure flag to `download images` command

**Why is this needed**:
In a certain customer environment, the HTTPS_PROXY environment is not honored by helm when --insecure-skip-tls-verify flag is enabled. Also, we should disable insecure flags by default from secure coding perspective

PR #2879, opened 2022-08-04
Add --insecure flag to download images command

wongni:
*Issue #, if available:*
#2878 

*Description of changes:*
Add --insecure flag to download images command

*Testing (if applicable):*
Manual test with and without --insecure on `download images` command


pkg/dependencies/factory.go
```diff
@@ -579,39 +578,19 @@ func (f *Factory) WithTroubleshoot() *Factory {
...
 
-func (f *Factory) WithHelmSecure() *Factory {
+func (f *Factory) WithHelm(insecure bool) *Factory {
 	f.WithExecutableBuilder()
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
-		if f.dependencies.HelmSecure != nil {
+		if f.dependencies.Helm != nil {
 			return nil
 		}
```
abhinavmpandey08, line 587: The insecure flag here won't work because once the helm executable is built, it is not going to build it again

vivek-koppuru: What do you mean by this? It gets built on every command execution?

abhinavmpandey08: There was an if condition which would prevent the executable from being re-built if wasn't nil. I think this is resolved in the latest commit

abhinavmpandey08 (approved):
looks good, thanks!

pkg/dependencies/factory.go
```diff
@@ -596,39 +595,15 @@ func (f *Factory) WithTroubleshoot() *Factory {
 	return f
 }
 
-func (f *Factory) WithHelmSecure() *Factory {
+func (f *Factory) WithHelm(insecure bool) *Factory {
```
jiayiwang7, line 598: can we make it optional arg? 
```suggestion
func (f *Factory) WithHelm(opts ...HelmOpts) *Factory {
```

jiayiwang7 (approved):

PR #2883, opened 2022-08-04
Isolated the api server collector and analyzer to vsphere cluster only

YoyoTT:
*Issue #, if available:*

*Description of changes:*
Move the the api server collector to be one of collectors of vSphere provider. This collector will only work when need to generate support bundle for the vSphere cluster. It won't bother the process of creating/getting support bundle for docker cluster. Same change on control plane analyzer. 

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/analyzers.go
+++ b/pkg/diagnostics/analyzers.go
@@ -165,11 +165,14 @@ func (a *analyzerFactory) DataCenterConfigAnalyzers(datacenter v1alpha1.Ref) []*
 }
 
 func (a *analyzerFactory) eksaVsphereAnalyzers() []*Analyze {
+	var analyzers []*Analyze
 	crds := []string{
 		fmt.Sprintf("vspheredatacenterconfigs.%s", v1alpha1.GroupVersion.Group),
 		fmt.Sprintf("vspheremachineconfigs.%s", v1alpha1.GroupVersion.Group),
 	}
-	return a.generateCrdAnalyzers(crds)
+	analyzers = append(analyzers, a.generateCrdAnalyzers(crds)...)
+	analyzers = append(analyzers, a.vsphereDiagnosticAnalyzers()...)
+	return analyzers
 }
 
 func (a *analyzerFactory) eksaCloudstackAnalyzers() []*Analyze {
@@ -211,13 +214,6 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
 				analyzers = append(analyzers, analyzer...)
 			}
 		}
-		// the Run Pod collector generates logs as well but not in the default logs folder.
-		if collector.RunPod != nil {
-			rp_analyzer, rp_ok := analyzersMap[collector.RunPod.Namespaces]
-			if rp_ok {
-				analyzers = append(analyzers, rp_analyzer...)
-			}
-		}
 	}
 	return analyzers
 }
@@ -227,7 +223,6 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
 func (a *analyzerFactory) namespaceLogTextAnalyzersMap() map[string][]*Analyze {
 	return map[string][]*Analyze{
 		constants.CapiKubeadmControlPlaneSystemNamespace: a.capiKubeadmControlPlaneSystemLogAnalyzers(),
-		constants.EksaDiagnosticsNamespace:               a.eksaDiagnosticsLogAnalyzers(),
 	}
 }
 
@@ -254,40 +249,6 @@ func (a *analyzerFactory) capiKubeadmControlPlaneSystemLogAnalyzers() []*Analyze
 						Pass: &singleOutcome{
 							When:    "false",
 							Message: "API server pods launched correctly",
-						},
-					},
-				},
-			},
-		},
-	}
-}
-
-// EksaDiagnosticsLogAnalyzers currently can help to analyze whether user is using a valid control plane IP to connect
-// to API server
-func (a *analyzerFactory) eksaDiagnosticsLogAnalyzers() []*Analyze {
-	runPingPod := "run-ping"
-	runPingPodLog := fmt.Sprintf("%s.log", runPingPod)
-	fullRunPingPodLogPath := path.Join(runPingPod, runPingPodLog)
-	return []*Analyze{
-		{
-			TextAnalyze: &textAnalyze{
-				analyzeMeta: analyzeMeta{
-					CheckName: fmt.Sprintf("%s:  Destination Host Unreachable. Log: %s", logAnalysisAnalyzerPrefix, fullRunPingPodLogPath),
-				},
-				FileName:     fullRunPingPodLogPath,
-				RegexPattern: `exit code: 0`,
-				Outcomes: []*outcome{
-					{
-						Fail: &singleOutcome{
-							When:    "false",
-							Message: fmt.Sprintf("The control plane endpoint host is unavailable. See %s", fullRunPingPodLogPath),
-						},
-					},
-
-					{
-						Pass: &singleOutcome{
-							When:    "true",
-							Message: "Control plane IP verified.",
 						},
 					},
 				},
@@ -359,6 +320,45 @@ func (a *analyzerFactory) crdAnalyzer(crdName string) *Analyze {
 				},
 			},
 			CustomResourceDefinitionName: crdName,
+		},
+	}
+}
+
+// vsphereDiagnosticAnalyzers will return diagnostic analyzers to analyze the condition of vSphere cluster
+func (a *analyzerFactory) vsphereDiagnosticAnalyzers() []*Analyze {
+	return a.controlPlaneIPAnalyzer()
+}
+
+// controlPlaneIPAnalyzer analyze whether a valid control plane IP is used to connect
+// to API server
+func (a *analyzerFactory) controlPlaneIPAnalyzer() []*Analyze {
+	runPingPod := "ping-host-ip"
+	runPingPodLog := fmt.Sprintf("%s.log", runPingPod)
+	fullRunPingPodLogPath := path.Join(runPingPod, runPingPodLog)
+	return []*Analyze{
+		{
+			TextAnalyze: &textAnalyze{
+				analyzeMeta: analyzeMeta{
+					CheckName: fmt.Sprintf("%s:  Destination Host Unreachable. Log: %s", logAnalysisAnalyzerPrefix, fullRunPingPodLogPath),
+				},
+				FileName:     fullRunPingPodLogPath,
+				RegexPattern: `exit code: 0`,
+				Outcomes: []*outcome{
+					{
+						Fail: &singleOutcome{
+							When:    "false",
+							Message: fmt.Sprintf("The control plane endpoint host is unavailable. See %s", fullRunPingPodLogPath),
+						},
+					},
+
+					{
+						Pass: &singleOutcome{
+							When:    "true",
+							Message: "Control plane IP verified.",
+						},
+					},
+				},
+			},
 		},
 	}
 }
--- a/pkg/diagnostics/analyzers_test.go
+++ b/pkg/diagnostics/analyzers_test.go
@@ -3,18 +3,22 @@ package diagnostics_test
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	. "github.com/onsi/gomega"
 
+	eksav1alpha1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/diagnostics"
 )
 
 func TestManagementClusterAnalyzers(t *testing.T) {
+	g := NewGomegaWithT(t)
 	factory := diagnostics.NewAnalyzerFactory()
 	analyzers := factory.ManagementClusterAnalyzers()
-	assert.Equal(t, len(analyzers), 12, "DataCenterConfigCollectors() mismatch between desired collectors and actual")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager"), "capc controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capv-controller-manager"), "capv controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capt-controller-manager"), "capt controller manager analyzer should be present")
+	g.Expect(analyzers).To(HaveLen(12), "DataCenterConfigCollectors() mismatch between desired collectors and actual")
+	g.Expect(getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager")).ToNot(BeNil(), "capc controller manager analyzer should be present")
+	g.Expect(getDeploymentStatusAnalyzer(analyzers, "capv-controller-manager")).ToNot(BeNil(), "capv controller manager analyzer should be present")
+	g.Expect(getDeploymentStatusAnalyzer(analyzers, "capt-controller-manager")).ToNot(BeNil(), "capt controller manager analyzer should be present")
+	g.Expect(analyzers[10].CustomResourceDefinition.CheckName).To(Equal("clusters.anywhere.eks.amazonaws.com"))
+	g.Expect(analyzers[11].CustomResourceDefinition.CheckName).To(Equal("bundles.anywhere.eks.amazonaws.com"))
 }
 
 func getDeploymentStatusAnalyzer(analyzers []*diagnostics.Analyze, name string) *diagnostics.Analyze {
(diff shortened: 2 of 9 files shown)


pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -450,12 +450,24 @@ func machineConfigs() []providers.MachineConfig {
 func TestGenerateManagementClusterBundle(t *testing.T) {
 	kubeconfig := "testcluster.kubeconfig"
 
-	spec := test.NewClusterSpec(func(s *cluster.Spec) {
+	bs_spec := test.NewClusterSpec(func(s *cluster.Spec) {
```
jiayiwang7, line 453: nit: can we use camelCase
```suggestion
	bootstrapSpec := test.NewClusterSpec(func(s *cluster.Spec) {
```

pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -450,12 +450,24 @@ func machineConfigs() []providers.MachineConfig {
...
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "bootstrap-cluster",
 			},
+			Spec: eksav1alpha1.ClusterSpec{
+				ControlPlaneConfiguration: eksav1alpha1.ControlPlaneConfiguration{},
+			},
+			Status: eksav1alpha1.ClusterStatus{},
+		}
+	})
+	wl_spec := test.NewClusterSpec(func(s *cluster.Spec) {
```
jiayiwang7, line 465:
```suggestion
	workloadSpec := test.NewClusterSpec(func(s *cluster.Spec) {
```

pkg/diagnostics/diagnostic_bundle.go
```diff
@@ -72,9 +72,12 @@ func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactor
 	b.WithDefaultCollectors().
 		WithDefaultAnalyzers().
 		WithManagementCluster(true).
-		WithDatacenterConfig(spec.Cluster.Spec.DatacenterRef).
-		WithAPIServerCollectors(spec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host). // Add a collector here to runPod and ping control plane IP once bootstrap cluster was created
-		WithLogTextAnalyzers()
+		WithDatacenterConfig(spec.Cluster.Spec.DatacenterRef)
+
+	if spec.Cluster.Spec.ControlPlaneConfiguration.Endpoint != nil && spec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host != "" {
+		b.WithAPIServerCollectors(spec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host)
```
jiayiwang7, line 78: im not familiar w this api server collector -- does it make sense that if no cp endpoint host specified, we don't check the ap server?

pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -468,19 +480,20 @@ func TestGenerateManagementClusterBundle(t *testing.T) {
...
+		a.EXPECT().DataCenterConfigAnalyzers(bs_spec.Cluster.Spec.DatacenterRef).AnyTimes().Return(nil)
+		a.EXPECT().DataCenterConfigAnalyzers(wl_spec.Cluster.Spec.DatacenterRef).AnyTimes().Return(nil)
 		a.EXPECT().EksaLogTextAnalyzers(gomock.Any()).AnyTimes().Return(nil)
 
 		c := givenMockCollectorsFactory(t)
-		c.EXPECT().DefaultCollectors().Return(nil)
-		c.EXPECT().ManagementClusterCollectors().Return(nil)
-		c.EXPECT().DataCenterConfigCollectors(spec.Cluster.Spec.DatacenterRef).Return(nil)
-		c.EXPECT().APIServerCollectors(spec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host).AnyTimes().Return(nil)
+		c.EXPECT().DefaultCollectors().AnyTimes().Return(nil)
```
jiayiwang7, line 490: Do we need AnyTimes() here for retry? If not can we limit it to One? Same below

pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -489,6 +502,7 @@ func TestGenerateManagementClusterBundle(t *testing.T) {
 		}
 
 		f := diagnostics.NewFactory(opts)
-		_, _ = f.DiagnosticBundleManagementCluster(spec, kubeconfig)
+		_, _ = f.DiagnosticBundleManagementCluster(bs_spec, kubeconfig)
```
jiayiwang7, line 505: do we need to validate the return for the method?

pkg/diagnostics/diagnostic_bundle_test.go
YoyoTT: Sure. Thanks!

pkg/diagnostics/diagnostic_bundle.go
YoyoTT: If aim to create a workload cluster and no cp endpoint provided, it won't pass the preflight check. The change here is to inform the api collector that the current task is for a workload cluster.

jaxesn: I'm not sure I follow. A workload cluster would still have control plane nodes.  In the case of docker there is no control plan ip specified since it's just handled automatically by docker. We still would want to collect api server logs in that case.

pkg/diagnostics/diagnostic_bundle_test.go
YoyoTT: Yeah, I will fix it shortly.

pkg/diagnostics/diagnostic_bundle.go
YoyoTT: This APISever collector will deploy a pod on bootstrap cluster and ping cp IP. From the result of ping, we are able to know the ip is valid or not. 
So even the docker takes over the cp ip, do we still need to run a pod and ping that cp ip?  Except this one, we still can get regular logs.

pkg/diagnostics/diagnostic_bundle_test.go
YoyoTT: > Do we need AnyTimes() here for retry? If not can we limit it to One? Same below

Yep, we need retry here.

pkg/diagnostics/diagnostic_bundle.go
jaxesn: The ping collector isnt needed for the docker provider. Can we just only add it for vsphere?

YoyoTT: Now it works for vsphere and even other providers as long as the spec contains the cp IP. If no cp host specified for vsphere cluster, the preflight check will filter it out.

pkg/diagnostics/diagnostic_bundle_test.go
jiayiwang7: then can you specify the `Times()` instead of `AnyTimes()`?

pkg/diagnostics/diagnostic_bundle.go
YoyoTT: To add the collector only for vsphere,  it needs to specify the provider=vsphere. There is no such field in cluster-definition.yaml. I just found that the kind of datacenter in the yaml could help us distinguish the providers. We already had a DataCenterConfig function to distribute collectors for vsphere, docker, and cloudstack. Is that ok that we add this collector to vsphere there? 
If we are allowed to do that, there is another problem that we couldnt pass the control plane IP for this collector. emmm, block here.

pkg/diagnostics/diagnostic_bundle_test.go
YoyoTT: Sorry about this late update. Took some time to learn the go mock and unite test. We do not need retry here and 1 time is fine. Fixed it in the recent commit.

pkg/diagnostics/diagnostic_bundle_test.go
```diff
@@ -462,6 +462,9 @@ func TestGenerateManagementClusterBundle(t *testing.T) {
 						Host: "1.1.1.1",
 					},
 				},
+				DatacenterRef: eksav1alpha1.Ref{
+					Kind: "VSphereDatacenterConfig",
```
mitalipaygude, line 466: can we add a test for docker data center config and confirm that it does not get triggered for docker?

YoyoTT: I did try to test for docker but since we need test two spec in this unit test, I couldn't make it work. Is it possible to test this in a separated unit test?

mitalipaygude:
ok-to-test

pkg/diagnostics/collectors.go
```diff
@@ -402,8 +407,8 @@ func (c *collectorFactory) hostPortCollector(ports []string, hostIP string) *Col
 	argsIP := []string{tempIPRequest}
 	return &Collect{
 		RunPod: &runPod{
-			Name:       "run-ip",
-			Namespaces: constants.EksaDiagnosticsNamespace,
+			Name:      "run-ip",
```
jaxesn, line 410: we should probably rename these to something more clear instead of just `run-ip`

pkg/diagnostics/collectors.go
```diff
@@ -421,8 +426,8 @@ func (c *collectorFactory) pingHostCollector(hostIP string) *Collect {
 	argsPing := []string{tempPingRequest}
 	return &Collect{
 		RunPod: &runPod{
-			Name:       "run-ping",
-			Namespaces: constants.EksaDiagnosticsNamespace,
+			Name:      "run-ping",
```
jaxesn, line 429: same here

jaxesn (commented):
Small naming thing, otherwise this looks reasonable.

pkg/diagnostics/collectors.go
YoyoTT: How about `check-host-port` and next could be `ping-host-ip`? Any suggestions?

pkg/diagnostics/analyzers.go
```diff
@@ -362,3 +323,43 @@ func (a *analyzerFactory) crdAnalyzer(crdName string) *Analyze {
...
 	}
 }
+
+// vsphereDiagnosticAnalyzers will return diagnostic analyzers to analyze the condition of vSphere cluster
+func (a *analyzerFactory) vsphereDiagnosticAnalyzers() []*Analyze {
+	var analyzers []*Analyze
+	return append(analyzers, a.controlPlaneIPAnalyzer()...)
+}
+
+// controlPlaneIPAnalyzer currently can help to analyze whether a valid control plane IP is used to connect
```
jiayiwang7, line 333:
```suggestion
// controlPlaneIPAnalyzer analyzes whether a valid control plane IP is used to connect
```

pkg/diagnostics/collectors.go
```diff
@@ -381,7 +386,7 @@ func (c *collectorFactory) crdCollector(crdType string) *Collect {
 }
 
 // APIServerCollectors collect connection info when running a pod on an existing cluster
```
jiayiwang7, line 388:
```suggestion
// apiServerCollectors collects connection info when running a pod on an existing cluster
```

jiayiwang7 (approved):

pkg/diagnostics/analyzers_test.go
```diff
@@ -41,3 +40,24 @@ func TestEksaLogTextAnalyzers(t *testing.T) {
 		}
 	}
 }
+
+func TestVsphereDataCenterConfigAnalyzers(t *testing.T) {
+	datacenter := eksav1alpha1.Ref{Kind: eksav1alpha1.VSphereDatacenterKind}
+	analyzerFactory := diagnostics.NewAnalyzerFactory()
+	analyzers := analyzerFactory.DataCenterConfigAnalyzers(datacenter)
+	assert.Equal(t, len(analyzers), 3, "DataCenterConfigAnalyzers() mismatch between desired analyzers and actual")
```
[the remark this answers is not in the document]
YoyoTT: Could you please tell me how the gomega is better here? I am not quite familiar gomega but I'd like to learn, just need a little time.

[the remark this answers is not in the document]
YoyoTT: Yes, let me add content check for each analyzer.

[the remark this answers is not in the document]
g-gaston: Yeah! I don't think it's better, but it's the assertion library the team decided to use. I think the tests that were already here were using testify before we made that decision. You don't need to change existing tests, but it's better if we write new tests using gomega.

pkg/diagnostics/analyzers.go
```diff
@@ -362,3 +323,43 @@ func (a *analyzerFactory) crdAnalyzer(crdName string) *Analyze {
 		},
 	}
 }
+
+// vsphereDiagnosticAnalyzers will return diagnostic analyzers to analyze the condition of vSphere cluster
+func (a *analyzerFactory) vsphereDiagnosticAnalyzers() []*Analyze {
+	var analyzers []*Analyze
+	return append(analyzers, a.controlPlaneIPAnalyzer()...)
```
g-gaston, line 330: you don't need to create a new variable for this

```suggestion
	return a.controlPlaneIPAnalyzer()
```

pkg/diagnostics/analyzers_test.go
```diff
@@ -3,18 +3,20 @@ package diagnostics_test
...
 func TestManagementClusterAnalyzers(t *testing.T) {
+	g := NewGomegaWithT(t)
 	factory := diagnostics.NewAnalyzerFactory()
 	analyzers := factory.ManagementClusterAnalyzers()
-	assert.Equal(t, len(analyzers), 12, "DataCenterConfigCollectors() mismatch between desired collectors and actual")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager"), "capc controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capv-controller-manager"), "capv controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capt-controller-manager"), "capt controller manager analyzer should be present")
+	g.Expect(len(analyzers)).To(Equal(12), "DataCenterConfigCollectors() mismatch between desired collectors and actual")
+	g.Expect(getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager")).ToNot(Equal(nil), "capc controller manager analyzer should be present")
```
g-gaston, line 17: nit

```suggestion
	g.Expect(getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager")).ToNot(BeNil(), "capc controller manager analyzer should be present")
```

g-gaston, line 17: why only nil check? aren't we able to predict the content of the structs? Happy to take this offline if you prefer!

pkg/diagnostics/analyzers_test.go
```diff
@@ -3,18 +3,20 @@ package diagnostics_test
...
 
 func TestManagementClusterAnalyzers(t *testing.T) {
+	g := NewGomegaWithT(t)
 	factory := diagnostics.NewAnalyzerFactory()
 	analyzers := factory.ManagementClusterAnalyzers()
-	assert.Equal(t, len(analyzers), 12, "DataCenterConfigCollectors() mismatch between desired collectors and actual")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capc-controller-manager"), "capc controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capv-controller-manager"), "capv controller manager analyzer should be present")
-	assert.NotNilf(t, getDeploymentStatusAnalyzer(analyzers, "capt-controller-manager"), "capt controller manager analyzer should be present")
+	g.Expect(len(analyzers)).To(Equal(12), "DataCenterConfigCollectors() mismatch between desired collectors and actual")
```
g-gaston, line 16:
```suggestion
	g.Expect(analyzers).To(HaveLen(12), "DataCenterConfigCollectors() mismatch between desired collectors and actual")
```

pkg/diagnostics/analyzers_test.go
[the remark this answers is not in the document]
YoyoTT: Thanks for letting me know this. I already changed both new tests and previous tests, just want to do more practice and also make them looks consistently. Do they look good now?

pkg/diagnostics/analyzers_test.go
YoyoTT: This test is obtained from previous work. It seems that the following `getDeploymentStatusAnalyzer()` tests the expected content by checking the `deploymentStatus.Name == name`. And the number of returned analyzers in each namespace is variable. I can add two more test regarding the crd content. Does it make sense?

pkg/diagnostics/collectors_test.go
```diff
@@ -4,44 +4,83 @@ import (
...
+			},
+			Status: eksav1alpha1.ClusterStatus{},
+		}
+	})
+	datacenter := eksav1alpha1.Ref{Kind: eksav1alpha1.VSphereDatacenterKind}
+	factory := diagnostics.NewDefaultCollectorFactory()
+	collectors := factory.DataCenterConfigCollectors(datacenter, spec)
+	g.Expect(collectors).To(HaveLen(10), "DataCenterConfigCollectors() mismatch between number of desired collectors and actual")
+	g.Expect(collectors[0].Logs.Namespace).To(Equal(constants.CapvSystemNamespace))
+	g.Expect(collectors[0].Logs.Name).To(Equal(fmt.Sprintf("logs/%s", constants.CapvSystemNamespace)))
```
g-gaston, line 49: What I was suggesting is comparing the whole slice content

```go
g.Expect(collectors).To(ConsistOf([]*Collector{
    {
     ...
    },
}))
```

Maybe I'm missing something and comparing individual field values is enough

g-gaston (approved):

PR #2817, opened 2022-07-28
Improve error handling for snowmachineconfig controller

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Address further comments from https://github.com/aws/eks-anywhere/pull/2724

TODO:
- Work on a builder so that we can mock the validator functions instead of the client ones https://github.com/aws/eks-anywhere/pull/2724#discussion_r931324391

*Testing (if applicable):*
unit test

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2887, opened 2022-08-05
Use client registry for snow validator

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Continuation of addressing comments from https://github.com/aws/eks-anywhere/pull/2724 and https://github.com/aws/eks-anywhere/pull/2817 for the snow machine config controller work.

Specifically addressing https://github.com/aws/eks-anywhere/pull/2724#discussion_r931324391

*Testing (if applicable):*
unit test


pkg/providers/snow/validator.go
```diff
@@ -12,23 +12,39 @@ const (
 	defaultAwsSshKeyName = "eksa-default"
 )
 
-type Validator struct {
+type validator struct {
```
g-gaston, line 15: I think we can call this `AwsClientValidator` (or something like that)
It follows the `Validator` interface using an aws client as the underlying implementation

pkg/providers/snow/validator.go
```diff
@@ -12,23 +12,39 @@ const (
...
+
+func NewValidator(aws aws.Clients) Validator {
+	return &validator{
 		awsClientMap: NewAwsClientMap(aws),
 	}
 }
 
-func NewValidatorFromAwsClientMap(awsClientMap AwsClientMap) *Validator {
-	return &Validator{
+func NewValidatorFromAwsClientMap(awsClientMap AwsClientMap) Validator {
```
g-gaston, line 41: I think this should still return the struct
We should try to "return structs, receive interfaces" when possible (except for the builder)

controllers/snow_machineconfig_controller.go
```diff
@@ -19,18 +19,24 @@ type ClientBuilder interface {
 	BuildSnowAwsClientMap(ctx context.Context) (aws.Clients, error)
 }
 
+type ValidatorBuilder interface {
+	Build(aws aws.Clients) snow.Validator
```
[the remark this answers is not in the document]
vivek-koppuru: This would require a lot more refactoring (I think in `entry.go`) from the cli workflow side too because right now we don't need to maintain context of the aws clients I think. Yea if we can avoid the builder that is probably best, but maybe we have a better way to solve this once we change how we use validator for both workflows?

pkg/providers/snow/reconciler/validatorbuilder.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func NewValidatorBuilder(clientRegistry ClientRegistry) *validatorBuilder {
+	return &validatorBuilder{
+		clientRegistry: clientRegistry,
+	}
+}
+
+func (b *validatorBuilder) Build(ctx context.Context) (snow.Validator, error) {
+	deviceClientMap, err := b.clientRegistry.Get(ctx)
```
g-gaston, line 25: I wonder if we should call this from inside the "validate" methods in the validator
I think that would allow us to avoid having a validator builder, right? We could just inject the validator directly in the reconciler

pkg/providers/snow/clientregistry.go
```diff
@@ -0,0 +1,38 @@
+package snow
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/aws"
+)
+
+type ClientRegistry interface {
```
g-gaston, line 10: nit: I would define this next to where it's used

pkg/providers/snow/validator.go
```diff
@@ -5,35 +5,39 @@ import (
...
-	awsClientMap AwsClientMap
+type AwsClientValidator struct {
+	clientRegistry ClientRegistry
 }
 
-func NewValidator(aws aws.Clients) *Validator {
-	return &Validator{
-		awsClientMap: NewAwsClientMap(aws),
-	}
+type Validator interface {
```
g-gaston, line 18: Do we need this interface here anymore?

g-gaston (approved):

pkg/providers/snow/clientregistry.go
vivek-koppuru: I have it defined here because it is used in both validator and defaulter

pkg/providers/snow/validator.go
vivek-koppuru: Yes but moving it to the controller

g-gaston (approved):

PR #2893, opened 2022-08-05
Continue e2e test after failure when checkpoint is enabled

taneyland:
*Issue #, if available:*

*Description of changes:*
For checkpoint tests, the first run of `upgrade cluster` will fail. If we return the error in the framework, this will signal the e2e test to end there. So if the checkpoint feature is enabled, we won't return the error in the framework. The error will still happen in the logs as usual, but the e2e test will continue, allowing us to accurately test the checkpoint feature.

Example output after the first upgrade failure with these changes:
```
2022-08-05T17:26:04.543Z	V4	Task finished	{"task_name": "collect-cluster-diagnostics", "duration": "2m41.519058791s"}
2022-08-05T17:26:04.543Z	V4	----------------------------------
2022-08-05T17:26:04.543Z	V4	Saving checkpoint	{"file": "eksa-test-5d81cdf-checkpoint.yaml"}
2022-08-05T17:26:04.544Z	V4	Tasks completed	{"duration": "4m48.184183815s"}
2022-08-05T17:26:04.544Z	V3	Cleaning up long running container	{"name": "eksa_1659720075911723793"}
Error: failed to upgrade cluster: waiting for external etcd for workload cluster to be ready: executing wait: error: timed out waiting for the condition on clusters/eksa-test-5d81cdf

    cluster.go:646: Running shell command [ eksctl anywhere upgrade cluster -f eksa-test-5d81cdf/cluster.yaml -v 4 ]
2022-08-05T17:26:04.816Z	V4	Logger init completed	{"vlevel": 4}
2022-08-05T17:26:04.989Z	V4	Reading bundles manifest
2022-08-05T17:26:05.064Z	V2	Pulling docker image
2022-08-05T17:26:05.294Z	V3	Initializing long running container
2022-08-05T17:26:05.483Z	V4	Checkpoint feature enabled
2022-08-05T17:26:05.483Z	V4	Reading checkpoint	{"file": "eksa-test-5d81cdf/generated/eksa-test-5d81cdf-checkpoint.yaml"}
2022-08-05T17:26:05.483Z	V4	Restoring task	{"task_name": "setup-and-validate"}
2022-08-05T17:26:05.483Z	V0	docker Provider setup is valid
2022-08-05T17:26:06.948Z	V4	Restoring task	{"task_name": "update-secrets"}
2022-08-05T17:26:06.948Z	V4	Restoring task	{"task_name": "ensure-etcd-capi-components-exist"}
...
```

*Testing (if applicable):*

Tested with docker checkpoint e2e test

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -82,6 +82,7 @@ type ClusterE2ETest struct {
 	ProxyConfig            *v1alpha1.ProxyConfiguration
 	AWSIamConfig           *v1alpha1.AWSIamConfig
 	eksaBinaryLocation     string
+	ExpectFailure          bool
 }
 
 type ClusterE2ETestOpt func(e *ClusterE2ETest)
@@ -149,6 +150,12 @@ func withHardware(requiredCount int, hardareType string, labels map[string]strin
 func WithNoPowerActions() ClusterE2ETestOpt {
 	return func(e *ClusterE2ETest) {
 		e.WithNoPowerActions = true
+	}
+}
+
+func ExpectFailure(expected bool) ClusterE2ETestOpt {
+	return func(e *ClusterE2ETest) {
+		e.ExpectFailure = expected
 	}
 }
 
@@ -675,6 +682,10 @@ func (e *ClusterE2ETest) Run(name string, args ...string) {
 		}
 
 		if errorMessage != "" {
+			if e.ExpectFailure {
+				e.T.Logf("This error was expected. Continuing...")
+				return
+			}
 			e.T.Fatalf("Command %s %v failed with error: %v: %s", name, args, err, errorMessage)
 		}
 
@@ -691,12 +702,6 @@ func (e *ClusterE2ETest) RunEKSA(args []string, opts ...CommandOpt) {
 		}
 	}
 	e.Run(binaryPath, args...)
-}
-
-func (e *ClusterE2ETest) StopIfSucceeded() {
-	if !e.T.Failed() {
-		e.T.FailNow()
-	}
 }
 
 func (e *ClusterE2ETest) StopIfFailed() {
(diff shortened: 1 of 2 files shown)


test/framework/cluster.go
```diff
@@ -675,6 +676,9 @@ func (e *ClusterE2ETest) Run(name string, args ...string) {
 		}
 
 		if errorMessage != "" {
+			if features.IsActive(features.CheckpointEnabled()) {
```
g-gaston, line 679: could we make it some kind of option? This is mixing eks-a feature toggles with the framework and it stays a bit hidden. I would prefer it we could make it more explicit, where when we execute the command from the test we specify we expect that command to fail.

taneyland: Good point. Added it as a ClusterE2ETestOpt now. Thoughts?

vivek-koppuru:
Are your unit tests succeeding locally?

taneyland:
> 

I'm unable to test vsphere e2e tests locally but I have the same test locally using Docker provider & it passes

test/framework/cluster.go
```diff
@@ -675,6 +682,9 @@ func (e *ClusterE2ETest) Run(name string, args ...string) {
 		}
 
 		if errorMessage != "" {
+			if e.ExpectFailure {
+				return
```
vivek-koppuru, line 686: Can we add a print statement here saying that we expected failure and some context?

test/framework/cluster.go
taneyland: Had a few conversations offline on potentially using `CommandOpt` here instead. Ultimately decided to keep this as a `ClusterE2ETestOpt` since the CommandOpts are used for changes to the command itself or the binary that we use to run that command. Whether or not a failure is expected does not change those factors, so it seems to make more sense as is. Merging it like this for now to get test runs in with the feature, but we can do a fast follow-up as we see fit to decide on the best route to take for this.

cc @vivek-koppuru @g-gaston

vivek-koppuru (approved):

PR #2894, opened 2022-08-05
Add Cilium reconciler

g-gaston:
*Description of changes:*
Reconciles the Cilium installation in a cluster to the desired state
defined in a cluster Spec. Intended to be used from a kubernetes
controller.

*Testing (if applicable):*
Unit tests


pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -0,0 +1,183 @@
...
+	return &Reconciler{
+		templater: templater,
+	}
+}
+
+// Reconcile takes the CIlium CNI in a cluster to the desired state defined in a cluster Spec
+// It uses a controller.Result to indicate when requeues are needed
+// Intended to be used in a kubernetes controller
+func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client client.Client, spec *cluster.Spec) (controller.Result, error) {
+	installation, err := getInstallation(ctx, client)
```
maxdrib, line 44: somewhat related but do we need a CNI health check to periodically monitor the status of Cilium and make sure it's all well and good? In my scaling testing, I encountered a few scenarios where Cilium had gone bad for a workload cluster and had to be reinstalled

g-gaston: that should be done by default, since this will run in a controller-runtime controller: CRs get reconciled periodically (every x hours), even without change

That said, this won't reinstall if it already exists, it would just check its health. I think we should investigate why Cilium needed to be reinstalled. Any more details on what was broken and why it needed to be reinstalled?

maxdrib: No more details. I just observed that after the upgrade, a few random workload clusters would be stuck with the machinedeployments not being ready and upon closer inspection, it showed that Cilium was down for those clusters. I never identified a root cause.

maxdrib (approved):

taneyland (approved):

pkg/networking/cilium/reconciler/reconciler.go
g-gaston: Got you. Then I would propose to move forward with this code. If we see those problems again, we can dive deep and find the root cause. That should allow us to write the code here to detect such scenario and fix it (even if the fix is reinstalling).

pkg/networking/cilium/reconciler/fetch.go
```diff
@@ -0,0 +1,97 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/networking/cilium"
+)
+
+func getInstallation(ctx context.Context, client client.Client) (*cilium.Installation, error) {
```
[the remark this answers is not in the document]
g-gaston: oh this package is just for the cilium reconciler -> `cilium`/`reconciler`

d8660091 (approved):

PR #2897, opened 2022-08-05
Use kubernetes client-go in a few places

ewollesen:
In general, the EKS-Anywhere project has shelled out to kubectl to interact
with the cluster's API. However, given that the EKS-A kubectl binary is run
via a docker image, it can be slow. For cluster creation/teardown, this
slowness is amortized and isn't nearly as noticeable. But for commands like
"list packages" which should be rather fast, it's very noticeable.

So we're starting to use API client libraries where it makes sense to speed up
these calls that are more interactive in nature, or where faster responses
make sense.

There will be more commits of this nature in the near future. So if you see
patterns here that you don't like, or think could be improved, now would be a
great time to bring them up for discussion. :)

As a historical note, the choice to shell out to kubectl was made because it
was planned that EKS-A would support more versions of kubernetes than the
upstream project would, and it was thought that the ability to swap out
executable versions to match the kubernetes version would be an easier or more
flexible way to manage the differences when compared with linking to / using
multiple versions of the client libraries. Whether or not that is true remains
to be seen, but that fact is being used as guidance here, in that—at least at
first—we will replace only calls that are stable and therefore less likely to
change or require multiple library versions. (Like the list and get calls
replaced here).

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgradepackages.go
+++ b/cmd/eksctl-anywhere/cmd/upgradepackages.go
@@ -45,11 +45,10 @@ func upgradePackages(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("unable to initialize executables: %v", err)
 	}
-
 	b := curatedpackages.NewBundleReader(
 		kubeConfig,
 		ipo.source,
-		deps.Kubectl,
+		deps.KubeRESTClient,
 		nil,
 		nil,
 	)
--- a/pkg/curatedpackages/bundle.go
+++ b/pkg/curatedpackages/bundle.go
@@ -2,17 +2,27 @@ package curatedpackages
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 
 	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(packagesv1.AddToScheme(scheme))
+}
 
 const (
 	ImageRepositoryName = "eks-anywhere-packages-bundles"
@@ -29,18 +39,20 @@ type BundleRegistry interface {
 type BundleReader struct {
 	kubeConfig    string
 	source        BundleSource
-	kubectl       KubectlRunner
 	bundleManager Manager
 	registry      BundleRegistry
+	restClient    rest.Interface
 }
 
-func NewBundleReader(kubeConfig string, source BundleSource, k KubectlRunner, bm Manager, reg BundleRegistry) *BundleReader {
+func NewBundleReader(kubeConfig string, source BundleSource,
+	restClient rest.Interface, bm Manager, reg BundleRegistry,
+) *BundleReader {
 	return &BundleReader{
 		kubeConfig:    kubeConfig,
 		source:        source,
-		kubectl:       k,
 		bundleManager: bm,
 		registry:      reg,
+		restClient:    restClient,
 	}
 }
 
@@ -64,7 +76,6 @@ func (b *BundleReader) getLatestBundleFromRegistry(ctx context.Context, kubeVers
 }
 
 func (b *BundleReader) getActiveBundleFromCluster(ctx context.Context) (*packagesv1.PackageBundle, error) {
-	// Active BundleReader is set at the bundle Controller
 	bundleController, err := b.GetActiveController(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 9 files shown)


pkg/dependencies/factory.go
```diff
@@ -914,6 +921,33 @@ func (f *Factory) WithCuratedPackagesRegistry(registryName, kubeVersion string,
 	return f
 }
 
+func (f *Factory) WithRESTClient(kubeConfigFilename string) *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		loader := clientcmd.NewDefaultClientConfigLoadingRules()
+		if f.kubeConfigFilename != "" {
```
g-gaston, line 927: What's the idea here? I got a bit lost

It seems like we can store the kubeconfig as part of the factory properties and then the rest client build logic uses it (if set) to configure the client. But we are also taking  `kubeConfigFilename` as an argument to `WithRESTClient`, except we seem to not use it, unless I'm missing something.

I think we can go with the simplest solution for now and just use the input argument to `WithRESTClient`? That way we avoid having to store config in the factory.

pkg/dependencies/factory.go
```diff
@@ -914,6 +921,33 @@ func (f *Factory) WithCuratedPackagesRegistry(registryName, kubeVersion string,
...
+			loader.ExplicitPath = f.kubeConfigFilename
+		}
+		cfg, err := clientcmd.BuildConfigFromKubeconfigGetter("", loader.Load)
+		if err != nil {
+			return err
+		}
+		scheme := runtime.NewScheme()
+		config := cfg
```
g-gaston, line 935: curious why the copy?

pkg/curatedpackages/bundle.go
```diff
@@ -2,18 +2,27 @@ package curatedpackages
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 
 	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+var scheme = runtime.NewScheme()
```
g-gaston, line 19: is this being used?

pkg/curatedpackages/bundle.go
```diff
@@ -76,43 +86,45 @@ func (b *BundleReader) getActiveBundleFromCluster(ctx context.Context) (*package
...
 
 func (b *BundleReader) UpgradeBundle(ctx context.Context, controller *packagesv1.PackageBundleController, newBundle string) error {
 	controller.Spec.ActiveBundle = newBundle
-	controllerYaml, err := yaml.Marshal(controller)
-	if err != nil {
-		return err
-	}
-	params := []string{"apply", "-f", "-", "--kubeconfig", b.kubeConfig}
-	_, err = b.kubectl.CreateFromYaml(ctx, controllerYaml, params...)
+	err := b.restClient.Put().
```
g-gaston, line 118: IMO this is ok for this usecase. But worth noting we are changing the behavior here from moving from apply (3 way merge, etc.) to PUT.

g-gaston (commented):
Looks good, minor suggestions :)

pkg/dependencies/factory.go
```diff
@@ -86,6 +91,7 @@ type Dependencies struct {
 	VSphereValidator          *vsphere.Validator
 	VSphereDefaulter          *vsphere.Defaulter
 	SnowValidator             *snow.Validator
+	RESTClient                rest.Interface
```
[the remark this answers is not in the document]
ewollesen: Done

cmd/eksctl-anywhere/cmd/upgradepackages.go
```diff
@@ -41,15 +41,15 @@ var upgradePackagesCommand = &cobra.Command{
 
 func upgradePackages(ctx context.Context) error {
 	kubeConfig := kubeconfig.FromEnvironment()
-	deps, err := NewDependenciesForPackages(ctx, WithMountPaths(kubeConfig))
+	deps, err := NewDependenciesForPackages(ctx, WithRegistryName(gpOptions.registry),
+		WithKubeVersion(gpOptions.kubeVersion), WithMountPaths(kubeConfig))
```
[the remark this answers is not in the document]
ewollesen: I believe so. I can't see any reason why I would need that change... Probably me trying to revert a botched earlier attempt and missed this. Thanks for the eagle eyes.

pkg/dependencies/factory.go
ewollesen: Oof yes, this was a "should I do it this way?", no I should do it this other way... But then goofed.

My goal was to have the kubeconfig only specified in one place. But the trick is that it's optional in its previous existence, but withrestclient requires it, so yeah, it's probably best to specify it here, since it's required here.

pkg/dependencies/factory.go
ewollesen: Mistake. Fixed.

pkg/curatedpackages/bundle.go
ewollesen: It should be used a few lines below, in the init() method. Github only shows the one line you highlighted, but if you were to look below, it should look like this:

```
var scheme = runtime.NewScheme()

func init() {
	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
	utilruntime.Must(packagesv1.AddToScheme(scheme))
}
```

pkg/curatedpackages/bundle.go
ewollesen: Interesting. I guess I'm not 100% up to date on the semantics of the POST/PUT/PATCH vs kubectl commands. Do you happen to have a good link to docs on that so I can brush up? I'd be curious how the command line is handling 3-way merges and general race issues.

I guess I should read the kubectl code.

pkg/curatedpackages/bundle.go
g-gaston: Ahh yeah I see that. However, aside from adding those crds to the scheme, are we using the scheme for anything else? It seems like we are creating a new one, so it's not necessarily modifying a global one (global at the client level, obv this one is global for this package).

ewollesen: Oh yes, it's doing nasty side-effecty-stateful things inside of the utilruntime package (hey, I didn't write it). As I'm sure you know, without it we couldn't marshal/unmarshal the packagesv1 (aka curated packages) API types. We need to do that now since we're using the go client directly, whereas before it was done via manual calls to `yaml.Unmarshal` etc.

pkg/curatedpackages/bundle.go
```diff
@@ -29,18 +39,20 @@ type BundleRegistry interface {
 type BundleReader struct {
 	kubeConfig    string
 	source        BundleSource
-	kubectl       KubectlRunner
 	bundleManager Manager
 	registry      BundleRegistry
+	restClient    rest.Interface
```
chrisdoherty4, line 44: Shouldn't we be able to specify the behavior we need and the client implementation is then irrelevant to this consumer?

ewollesen: Just had a good chat over chime, which I'll summarize here:

+ We agreed that the code would be improved if this used a more task-centered interface of our own definition.
+ We also agreed that such an interface wouldn't be entirely in the scope of this PR, and is probably best added in a follow-up.

Furthermore, part of the reason for the 2nd point above is because we don't have a good grasp of what our interface would look like. Using the k8s client-go interface is still somewhat exploratory, so the above gives us freedom for now, though it does also leave exposed some rough edges that if not handled carefully could bite us later.

... hopefully I can get to the follow on soon.

ewollesen:
/hold

I'm seeing some issues when no KUBECONFIG is provided (thanks to @a-cool-train) that I need to get sorted before this merges.

pkg/dependencies/factory.go
```diff
@@ -7,7 +7,12 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
+	cpv1alpha1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
```
g-gaston, line 15: nit: I would name this `packagesv1`. That kind of follows the naming we have been using for other api packages. The idea of not using `v1alpha1` is that it makes easier to bump versions (like to v1alpha2)

pkg/curatedpackages/bundle.go
g-gaston: I see. But that's what's tripping me, how is this having side effects when we are creating our own scheme? This is mostly out of curiosity btw. But every time I have seen this been used is either usually over `scheme.Scheme` or if the new created scheme was being passed to a client. But it doesn't seem like here we are doing either.

PR #2899, opened 2022-08-05
Flux webhook validation and preflight check

jiayiwang7:
*Issue #, if available:*

#2755 

*Description of changes:*

The recent #2856 causes cluster creation failed when Flux git is enabled, since the webhook validation checks the git related ENV vars which should only be checked in CLI validation.

Remove the env checks from webhook validation, update CLI preflight validation:
- Move `ValidateAuthenticationForGitProvider` to validation pkg
- Check the git envs for both create and upgrade command, for both mgmt and workload clusters.

This PR mainly move some methods and tests around instead of refactoring any of them.

*Testing (if applicable):*

Unit tests


d8660091 (approved):
Discussed in Slack, LGTM. Thanks for the quick fix!

PR #2900, opened 2022-08-05
Use Println instead of Print for kubectl output

g-gaston:
*Description of changes:*
This messes up with the terminal output and the go test output when the
printed text doesn't end with a break line.

### DIFF
--- a/pkg/curatedpackages/packageclient.go
+++ b/pkg/curatedpackages/packageclient.go
@@ -127,7 +127,7 @@ func (pc *PackageClient) InstallPackage(ctx context.Context, bp *packagesv1.Bund
 	if err != nil {
 		return err
 	}
-	fmt.Print(&stdOut)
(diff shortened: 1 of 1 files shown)


a-cool-train (commented):
This will print an extra line unnecessarily. I would recommend to refrain from it unless you have seen a need for it.

vivek-koppuru:
@a-cool-train Is this a breaking change for you? Currently, does the output properly end with an extra line, and this would just add another?

a-cool-train:
> @a-cool-train Is this a breaking change for you? Currently, does the output properly end with an extra line, and this would just add another?

This doesn't break but it adds an extra line to the output.

g-gaston:
Closing in favor of a different fix @a-cool-train has in mind

g-gaston:
Closing in favor of a different fix @a-cool-train has in mind

PR #2901, opened 2022-08-08
Add snowmachineconfig check in snow reconcilier

d8660091:
*Issue #, if available:*

*Description of changes:* 1) Add snowmachineconfig check 2) add unit test for snow reconciler

*Testing (if applicable):* unit test

### DIFF
--- a/pkg/providers/snow/reconciler/main_test.go
+++ b/pkg/providers/snow/reconciler/main_test.go
@@ -1,0 +1,14 @@
+package reconciler_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/aws/eks-anywhere/internal/test/envtest"
+)
+
+var env *envtest.Environment
+
+func TestMain(m *testing.M) {
+	os.Exit(envtest.RunWithEnvironment(m, envtest.WithAssignment(&env)))
+}
--- a/pkg/providers/snow/reconciler/reconciler.go
+++ b/pkg/providers/snow/reconciler/reconciler.go
@@ -2,19 +2,45 @@ package reconciler
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	c "github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/controller"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
 )
 
-type Reconciler struct{}
-
-func New() *Reconciler {
-	return &Reconciler{}
+type Reconciler struct {
+	client client.Client
 }
 
-func (v *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+func New(client client.Client) *Reconciler {
+	return &Reconciler{
+		client: client,
+	}
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	specWithBundles, err := c.BuildSpec(ctx, clientutil.NewKubeClient(r.client), cluster)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	for _, machineConfig := range specWithBundles.SnowMachineConfigs {
+		if !machineConfig.Status.SpecValid {
+			failureMessage := fmt.Sprintf("SnowMachineConfig %s is invalid", machineConfig.Name)
+			if machineConfig.Status.FailureMessage != nil {
+				failureMessage += ": " + *machineConfig.Status.FailureMessage
+			}
+
+			log.Error(nil, failureMessage)
+			cluster.Status.FailureMessage = &failureMessage
+			return controller.Result{}, nil
+		}
+	}
+
 	return controller.Result{}, nil
 }
--- a/pkg/providers/snow/reconciler/reconciler_test.go
+++ b/pkg/providers/snow/reconciler/reconciler_test.go
@@ -1,0 +1,281 @@
+package reconciler_test
+
+import (
+	"context"
+	"testing"
+
+	eksdv1alpha1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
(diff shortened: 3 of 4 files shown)


pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,38 @@ package reconciler
...
+func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	for _, ref := range cluster.MachineConfigRefs() {
+		machineConfig := &anywherev1.SnowMachineConfig{}
+		if err := r.client.Get(ctx,
+			types.NamespacedName{Namespace: cluster.Namespace, Name: ref.Name},
+			machineConfig); err != nil {
+			return controller.Result{}, err
+		}
+		if !machineConfig.Status.SpecValid {
+			return controller.Result{}, fmt.Errorf("SnowMachineConfig %s is invalid", ref.Name)
```
g-gaston, line 34: I think we should try to include the status error in the message

However, I don't think we should return an error here. That will trigger another reconciliation loop. But it will continue failing forever until the machine config is not changed.

So maybe just log it?

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,38 @@ package reconciler
...
-	return &Reconciler{}
+func New(client client.Client) *Reconciler {
+	return &Reconciler{
+		client: client,
+	}
 }
 
-func (v *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	for _, ref := range cluster.MachineConfigRefs() {
```
g-gaston, line 26: we could just use `cluster.BuildSpec` here, which should already have all the snow config machines
and it will come in handy for the rest of the reconciliation loop

pkg/providers/snow/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/providers/snow/reconciler"
+)
+
+func TestReconcileSuccess(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	cluster := &v1alpha1.Cluster{}
+	client := mocks.NewMockClient(ctrl)
```
g-gaston, line 24: This is not blocking but we don't need to use mocks, we can use the kube fake client
I personally think it makes writing tests for controller code easier. But if you think otherwise, I'm good with this

pkg/providers/snow/reconciler/reconciler.go
d8660091: Good catch, the cluster reconciler has already watched change for SnowMachineConfig. We should ignore validation errors here as it cannot be fixed by requeue and just wait for another notification.

We may want to change the vSphere reconciler to match this behaviour

pkg/providers/snow/reconciler/reconciler_test.go
d8660091: Good idea, it's simpler to use the built-in fake client.

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,41 @@ package reconciler
...
-func (v *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	specWithBundles, err := c.BuildSpec(ctx, clientutil.NewKubeClient(r.client), cluster)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	for _, machineConfig := range specWithBundles.SnowMachineConfigs {
+		if !machineConfig.Status.SpecValid {
+			failureMessage := fmt.Sprintf("SnowMachineConfig %s is invalid", machineConfig.Name)
```
g-gaston, line 34: could we include here the error from the machine config?

controllers/cluster_controller_test.go
```diff
@@ -277,6 +278,127 @@ func TestClusterReconcilerDeleteNoCAPIClusterSuccess(t *testing.T) {
 	}
 }
 
+func TestSnowClusterReconcileSuccess(t *testing.T) {
```
g-gaston, line 281: could we test this from `pkg/providers/snow/reconciler/reconciler_test.go`?

g-gaston (commented):
looks good!

controllers/cluster_controller_test.go
d8660091: Good idea, I'll try and see if it can fix the code coverage report

pkg/providers/snow/reconciler/main_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"os"
+	"testing"
+
+	"github.com/aws/eks-anywhere/internal/test/envtest"
+)
+
+var env *envtest.Environment
+
+func TestMain(m *testing.M) {
+	os.Exit(envtest.RunWithEnvironment(m, envtest.WithAssignment(&env)))
```
g-gaston, line 13: We don't need this unless the tests use envtest. I think you are only using fake client, right?

pkg/providers/snow/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,281 @@
...
+	client := fake.NewClientBuilder().WithObjects(cluster, managementCluster, datacenterConfig, bundle, machineConfigCP, machineConfigWN, eksd).Build()
+
+	_, e := reconciler.New(client).Reconcile(context.Background(), log.Log, cluster)
+	g.Expect(e).To(BeNil(), "error should be nil to prevent requeue")
+
+	g.Expect(cluster.Status.FailureMessage).ToNot(BeZero())
+	g.Expect(*cluster.Status.FailureMessage).To(ContainSubstring("SnowMachineConfig snow-test-cp is invalid"))
+}
+
+func createSnowCluster() *anywherev1.Cluster {
```
g-gaston, line 115: nit: for all these method I would suggest to just use the name of the object you are returning, like 
```suggestion
func snowCluster() *anywherev1.Cluster {
```

These methods are not creating any object in the cluster, they just return a struct.

g-gaston (approved):

pkg/providers/snow/reconciler/main_test.go
d8660091: Without setting environment, fake client will not work, as it will refuse the objects we pass.

g-gaston: Ahhh I see, we don't need init envtest, just import a package

try adding `_ "github.com/aws/eks-anywhere/internal/test/envtest"` to the imports

d8660091: Great, I'll make another PR for it

Issue #2885, opened 2022-08-05
aws-iam-authenticator not working in baremtal


helloyeops:
**What happened**:
`aws-iam-authenticator` pods are not working with error as below.
```bash
$ kubectl -n kube-system logs aws-iam-authenticator-5qtf
...
W0729 07:36:06.265507       1 client_config.go:615] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2022-07-29T07:36:06Z" level=info msg="starting mapper \"EKSConfigMap\""
time="2022-07-29T07:36:06Z" level=fatal msg="could not load/generate a certificate" error="read /var/aws-iam-authenticator/cert.pem: is a directory"
```
It seems like that the files(`kubeconfig.yaml`, `cert.pem`, `key.pem`) are normally not generated.
And I got some find related issues.

- https://github.com/aws/eks-anywhere/pull/1803
- #2749 

Is working the workflow for `aws-iam-athutenticator` in baremetal?

**What you expected to happen**:
`aws-iam-authenticator` pods' state will be running.

**How to reproduce it (as minimally and precisely as possible)**:
EKS-A deploy a new cluster with `aws-iam-authenticator`
https://anywhere.eks.amazonaws.com/docs/tasks/cluster/cluster-iam-auth/

**Anything else we need to know?**:
Thanks

**Environment**:
- EKS Anywhere Release: 0.10.1
- EKS Distro Release:  1.22-9

pokearu:
@helloyeops Hi, IAM Authenticator integration is not supported for bare metal at the moment. We are working to add support in the upcoming release.

SeungyeopShin:
Thanks for reply.

helloyeops:
Thanks for reply.

danbudris:
Look for this functionality in the next minor release, which will be available soon

Issue #2749, opened 2022-07-18
AWS IAM Authenticator for bare-metal clusters


pokearu:
### Description
EKS-A provides integration for `aws-iam-authenticator`. We need to enable and validate the integration for bare-metal cluster creation workflow.

SeungyeopShin:
Hi @pokearu 
I'm trying to install EKS-A for bare-metal on own bare-metal cluster with aws-iam-authenticator.
Does EKS-A for bare-meta not yet support aws-iam-authenticator?

helloyeops:
Hi @pokearu
I'm trying to install EKS-A for bare-metal on own bare-metal cluster with aws-iam-authenticator.
Does EKS-A for bare-meta not yet support aws-iam-authenticator?

pokearu:
@helloyeops Hi, apologies for the late response. We are currently working to add IAM Authenticator integrated support for bare metal clusters in the upcoming release.

PR #2902, opened 2022-08-08
IAM Authenticator for provider Tinkerbell

pokearu:
*Issue #2749*
Also addresses #2885 

*Description of changes:*
Changes to enable support for `aws-iam-authenticator` integration for EKS-A with provider tinkerbell.

### DIFF
--- a/pkg/providers/tinkerbell/config/template-cp.yaml
+++ b/pkg/providers/tinkerbell/config/template-cp.yaml
@@ -72,6 +72,17 @@ spec:
         extraArgs:
 {{ .apiserverExtraArgs.ToYaml | indent 10 }}
 {{- end }}
+{{- if .awsIamAuth}}
+        extraVolumes:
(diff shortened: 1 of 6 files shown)


d8660091:
I see the test is doing ValidateAWSIamAuth(), but I am not sure if this line is invoked, https://github.com/d8660091/eks-anywhere/blob/snow-machine-config-validate-in-reconciler/pkg/clusterapi/apibuilder.go#L205, without it, the iam authenticator would not work correctly, right?

pokearu:
> I see the test is doing ValidateAWSIamAuth(), but I am not sure if this line is invoked, https://github.com/d8660091/eks-anywhere/blob/snow-machine-config-validate-in-reconciler/pkg/clusterapi/apibuilder.go#L205, without it, the iam authenticator would not work correctly, right?

I am actually not sure. Let me check. Right now all the e2e tests for IAM Auth invoke that validate.

PR #2904, opened 2022-08-09
Change immutable field errors to type Forbidden

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Change immutable field errors to type Forbidden instead of Invalid so as to follow CAPI.
Reference: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta1/machinehealthcheck_webhook.go#L135

*Testing (if applicable):*
- Unit tests

### DIFF
--- a/pkg/api/v1alpha1/awsiamconfig_webhook.go
+++ b/pkg/api/v1alpha1/awsiamconfig_webhook.go
@@ -78,7 +78,7 @@ func validateImmutableAWSIamFields(new, old *AWSIamConfig) field.ErrorList {
 	if !new.Spec.Equal(&old.Spec) {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "AWSIamConfig"), new, "config is immutable"),
+			field.Forbidden(field.NewPath(AWSIamConfigKind), "config is immutable"),
 		)
 	}
 
--- a/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
+++ b/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
@@ -108,6 +108,7 @@ func isCapcV1beta1ToV1beta2Upgrade(new, old *CloudStackDatacenterConfigSpec) boo
 
 func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConfig) field.ErrorList {
 	var allErrs field.ErrorList
+	specPath := field.NewPath("spec")
 
 	// Check for CAPC v1beta1 -> CAPC v1beta2 upgrade
 	if isCapcV1beta1ToV1beta2Upgrade(&new.Spec, &old.Spec) {
@@ -124,7 +125,7 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 			if !newAz.Equal(&oldAz) {
 				allErrs = append(
 					allErrs,
-					field.Invalid(field.NewPath("spec", "availabilityZone", oldAz.Name), newAz, "availabilityZone is immutable"),
+					field.Forbidden(specPath.Child("availabilityZone", oldAz.Name), "availabilityZone is immutable"),
 				)
 			}
 		}
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -113,48 +113,49 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 	}
 
 	var allErrs field.ErrorList
+	specPath := field.NewPath("spec")
 
 	if !old.ManagementClusterEqual(new) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "managementCluster", new.Spec.ManagementCluster.Name), fmt.Sprintf("field is immutable %v", new.Spec.ManagementCluster)))
+			field.Forbidden(specPath.Child("managementCluster", new.Spec.ManagementCluster.Name), fmt.Sprintf("field is immutable %v", new.Spec.ManagementCluster)))
 	}
 
 	if !new.Spec.ControlPlaneConfiguration.Endpoint.Equal(old.Spec.ControlPlaneConfiguration.Endpoint) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "ControlPlaneConfiguration.endpoint"), fmt.Sprintf("field is immutable %v", new.Spec.ControlPlaneConfiguration.Endpoint)))
+			field.Forbidden(specPath.Child("ControlPlaneConfiguration.endpoint"), fmt.Sprintf("field is immutable %v", new.Spec.ControlPlaneConfiguration.Endpoint)))
 	}
 
 	if !new.Spec.DatacenterRef.Equal(&old.Spec.DatacenterRef) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "datacenterRef"), fmt.Sprintf("field is immutable %v", new.Spec.DatacenterRef)))
+			field.Forbidden(specPath.Child("datacenterRef"), fmt.Sprintf("field is immutable %v", new.Spec.DatacenterRef)))
 	}
 
 	if !new.Spec.ClusterNetwork.Equal(&old.Spec.ClusterNetwork) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "ClusterNetwork"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
+			field.Forbidden(specPath.Child("ClusterNetwork"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
 	}
 
 	if !new.Spec.ProxyConfiguration.Equal(old.Spec.ProxyConfiguration) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "ProxyConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ProxyConfiguration)))
+			field.Forbidden(specPath.Child("ProxyConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ProxyConfiguration)))
 	}
 
 	if new.Spec.ExternalEtcdConfiguration != nil && old.Spec.ExternalEtcdConfiguration == nil {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec.externalEtcdConfiguration"), "cannot switch from local to external etcd topology"),
+			field.Forbidden(specPath.Child("externalEtcdConfiguration"), "cannot switch from local to external etcd topology"),
 		)
 	}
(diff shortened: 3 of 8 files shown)


pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -87,7 +87,7 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 	if old.Spec.Domain != new.Spec.Domain {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "domain"), new.Spec.Domain, "field is immutable"),
+			field.Forbidden(field.NewPath("spec", "domain", new.Spec.Domain), "field is immutable"),
```
maxdrib, line 90: What's the motivation for this change?

mitalipaygude: CAPI uses forbidden for immutable fields, instead of Invalid. Invalid fields can generally mean its a string instead of int, but we want to forbid changing the value at all for immutable fields.

maxdrib: Gotcha. In that case we should be good from Cloudstack, just need to resolve the conflicts

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -104,21 +104,21 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 	if zonesMutated {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "zone"), new.Spec.Zones, "field is immutable"),
+			field.Forbidden(field.NewPath("spec", "zone"), "field is immutable"),
```
vivek-koppuru, line 107: Should this be the following? But check the spec fields in the api package
```suggestion
			field.Forbidden(field.NewPath("spec", "zones"), "field is immutable"),
```

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -104,21 +104,21 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
...
 		)
 	}
 
 	if old.Spec.Account != new.Spec.Account {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "account"), new.Spec.Account, "field is immutable"),
+			field.Forbidden(field.NewPath("spec", "account", new.Spec.Account), "field is immutable"),
```
vivek-koppuru, line 114: ?
```suggestion
			field.Forbidden(field.NewPath("spec", "account"), "field is immutable"),
```

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -104,21 +104,21 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
...
 		)
 	}
 
 	if old.Spec.ManagementApiEndpoint != new.Spec.ManagementApiEndpoint {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "managementApiEndpoint"), new.Spec.ManagementApiEndpoint, "field is immutable"),
+			field.Forbidden(field.NewPath("spec", "managementApiEndpoint", new.Spec.ManagementApiEndpoint), "field is immutable"),
```
vivek-koppuru, line 121: ?
```suggestion
			field.Forbidden(field.NewPath("spec", "managementApiEndpoint"), "field is immutable"),
```

pkg/api/v1alpha1/awsiamconfig_webhook.go
```diff
@@ -78,7 +78,7 @@ func validateImmutableAWSIamFields(new, old *AWSIamConfig) field.ErrorList {
 	if !new.Spec.Equal(&old.Spec) {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "AWSIamConfig"), new, "config is immutable"),
+			field.Forbidden(field.NewPath("spec", "AWSIamConfig", new.Name), "config is immutable"),
```
vivek-koppuru, line 81: I wonder if this is better? However we should see what capi does as an example
```suggestion
			field.Forbidden(field.NewPath("AWSIamConfig", new.Name), "spec is immutable"),
```

pkg/api/v1alpha1/fluxconfig_webhook.go
```diff
@@ -90,7 +90,7 @@ func validateImmutableFluxFields(new, old *FluxConfig) field.ErrorList {
 	if !new.Spec.Equal(&old.Spec) {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", FluxConfigKind), new, "config is immutable"),
+			field.Forbidden(field.NewPath("spec", FluxConfigKind, new.Name), "config is immutable"),
```
vivek-koppuru, line 93: Same as above comment for `AWSIamConfig`, as well as all the other files below

pkg/api/v1alpha1/vspheredatacenterconfig_webhook.go
```diff
@@ -93,35 +93,35 @@ func validateImmutableFieldsVSphereCluster(new, old *VSphereDatacenterConfig) fi
 	if old.Spec.Server != new.Spec.Server {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "server"), new.Spec.Server, "field is immutable"),
+			field.Forbidden(field.NewPath("spec", "server", new.Spec.Server), "field is immutable"),
```
vivek-koppuru, line 96: Same goes here on whether we want to remove that field if we are already specifying "server", as well as the ones below in this file.

pkg/api/v1alpha1/awsiamconfig_webhook.go
```diff
@@ -78,7 +78,7 @@ func validateImmutableAWSIamFields(new, old *AWSIamConfig) field.ErrorList {
 	if !new.Spec.Equal(&old.Spec) {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "AWSIamConfig"), new, "config is immutable"),
+			field.Forbidden(field.NewPath("AWSIamConfig", new.Name), "config is immutable"),
```
g-gaston, line 81: What is the logic we follow to define the path?
It seems like we are removing spec from all of them, was that redundant?
Also, why is `new.Name` part of the path?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -113,56 +113,57 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
...
 			allErrs,
-			field.Forbidden(field.NewPath("spec", "ProxyConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ProxyConfiguration)))
+			field.Forbidden(specPath.Child("ProxyConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ProxyConfiguration)))
 	}
 
 	if new.Spec.ExternalEtcdConfiguration != nil && old.Spec.ExternalEtcdConfiguration == nil {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(field.NewPath("spec.externalEtcdConfiguration"), "cannot switch from local to external etcd topology"),
+			field.Forbidden(specPath.Child("spec.externalEtcdConfiguration"), "cannot switch from local to external etcd topology"),
```
g-gaston, line 151: this already contains spec

pkg/api/v1alpha1/awsiamconfig_webhook.go
```diff
@@ -74,11 +74,12 @@ func (r *AWSIamConfig) ValidateDelete() error {
 
 func validateImmutableAWSIamFields(new, old *AWSIamConfig) field.ErrorList {
 	var allErrs field.ErrorList
+	specPath := field.NewPath("spec")
 
 	if !new.Spec.Equal(&old.Spec) {
 		allErrs = append(
 			allErrs,
-			field.Invalid(field.NewPath("spec", "AWSIamConfig"), new, "config is immutable"),
+			field.Forbidden(specPath.Child("AWSIamConfig"), "config is immutable"),
```
vivek-koppuru, line 82: Does `spec.AWSIamConfig` make sense though? Can we somehow just say "AWSIamConfig <name>: config is immutable"?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -186,7 +187,7 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 		if !oldAWSIamConfig.Equal(newAWSIamConfig) {
 			allErrs = append(
 				allErrs,
-				field.Forbidden(field.NewPath("spec", "AWS Iam Config"), fmt.Sprintf("field is immutable %v", newAWSIamConfig.Kind)))
+				field.Forbidden(specPath.Child("AWS Iam Config"), fmt.Sprintf("field is immutable %v", newAWSIamConfig.Kind)))
```
vivek-koppuru, line 190: This is not a field right? It's the entire config, unless this is referring to the ref field, which we should refer to that instead.

mitalipaygude: Not a field technically, its a ref. But when users edit it, we treat it as a field that is a reference.

pkg/api/v1alpha1/awsiamconfig_webhook.go
mitalipaygude: So capi adds the spec path to all these fields we call out, so to stick with capi, we add spec as well. Which means the entire AWS IAM config spec is not mutable, with all the sub fields.

pkg/api/v1alpha1/cluster_webhook.go
vivek-koppuru:
```suggestion
				field.Forbidden(specPath.Child("identityProviderRefs", "kind", "AWSIamConfig"), fmt.Sprintf("field is immutable %v", newAWSIamConfig.Kind)))
```

vivek-koppuru (approved):

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -124,7 +125,7 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 			if !newAz.Equal(&oldAz) {
 				allErrs = append(
 					allErrs,
-					field.Invalid(field.NewPath("spec", "availabilityZone", oldAz.Name), newAz, "availabilityZone is immutable"),
+					field.Forbidden(specPath.Child("availabilityZone", oldAz.Name), "availabilityZone is immutable"),
```
maxdrib, line 128: there's another field.Invalid reference lower in this file. Should that change as well?

mitalipaygude: that reference isnt for an immutable error/field. We are updating the Invalid to Forbidden only for immutable cases.

Issue #2744, opened 2022-07-18
Preflight validation of vsphere creds for required perms


vivek-koppuru:
Currently users need to add the following permissions to their users/roles when creating an EKS-A cluster. 
https://anywhere.eks.amazonaws.com/docs/reference/vsphere/vsphere-preparation/

However, we currently don't have robust validations on whether the user they provide has the appropriate permissions that we are asking for in the documentation. `govc` has a few commands that would help be able to do this programmatically. Here is the usage instructions for govc: https://github.com/vmware/govmomi/blob/master/govc/USAGE.md

The list of permissions we ask above can be turned into a list that we can check for by running the `govc role.ls` command, which would list out all the applicable permissions defined for the role. It is described differently through the cli as follows:

If described as the following in the documentation:
```
> Content Library
* Add library item
```
It translates to the following in the permissions list for the cli (running the command for an `Admin` user should list all the possible values, and we might have to map them accordingly to the documentation):
```
ContentLibrary.AddLibraryItem
```

There are also the `permissions.ls` command and the `sso.user.ls` command that might list out what roles are associated with the user, however, the documentation on the usage is sparse, so we need to see if `permissions.ls` command supports `-principal` flag and if the `sso.user.ls` supports listing permissions based on passing in the name of the user.

vivek-koppuru:
This example may help understand how users, groups and roles might all be associated with each other, but am not sure if it applies to our usecase.

jonathanmeier5:
Based on my research so far we can implement this ticket but the check will require admin credentials.

We can view the roles associated with an object using:
```
govc permissions.ls  /Datacenter/datastore/datastore2
```

You can ignore inherited roles using the `-a` flag:
```
govc permissions.ls -a=false  /Datacenter/datastore/datastore2
```

This gives the information we need about a particular object's associated roles.

We can check that roles privileges with `role.ls`:
```
govc role.ls EKS\ Anywhere\ User
```

However, the only way I have found to establish which roles a *user* is associated to is with `sso.user.id`:
```
govc sso.user.id jwmeier
```

But this command requires a user have the Administrator role.

I'm trying to determine if lower levels of access may be able to run this command but haven't found any yet.

jonathanmeier5:
I timeboxed some additional exploration to determine if it is possible to list a user's groups without the Administrator role and have not come up with anything.

[...]

PR #2907, opened 2022-08-09
Validate VSphere User Privs

jonathanmeier5:
*Issue #, if available:*
#2744 

*Description of changes:*

Add a preflight check to validate vSphere user's permissions.

*Testing (if applicable):*

I have tested creating clusters using the permissions I defined following our docs. I'm concerned that my test environment might not reflect real world use cases in some way that I am not aware of.


pkg/providers/vsphere/validator.go
```diff
@@ -12,15 +14,51 @@ import (
...
 type Validator struct {
-	govc      ProviderGovcClient
-	netClient networkutils.NetClient
+	govc             ProviderGovcClient
+	netClient        networkutils.NetClient
+	newVSphereClient NewVSphereClientCallable
 }
 
-func NewValidator(govc ProviderGovcClient, netClient networkutils.NetClient) *Validator {
+func NewValidator(govc ProviderGovcClient, netClient networkutils.NetClient, f NewVSphereClientCallable) *Validator {
```
jonathanmeier5, line 52: I feel like I might be abusing golang here. I want a way to do the following:
- inject a vsphere client mock when testing
- when not testing, I want to automatically specify a sane default vsphere client

Is there a better way to achieve this?

pkg/providers/vsphere/validator.go
```diff
@@ -385,3 +423,184 @@ func (v *Validator) validateControlPlaneIpUniqueness(spec *Spec) error {
...
+			"path":         spec.datacenterConfig.Spec.Network,
+		},
+
+		// validate Administrator role (all privs) on VM folder and Template folder
+		{
+			"type":         vsphereTypeFolder,
+			"privsContent": vsphereAdminPrivsFile,
+			"path":         controlPlaneMachineConfig.Spec.Folder,
+		},
+		// Not implementing a check on the template because I'm not sure if we always require/provide/allow template permissions
```
jonathanmeier5, line 461: I don't understand what this section means:
https://anywhere.eks.amazonaws.com/docs/reference/vsphere/vsphere-preparation/#deploy-an-ova-template

Does it mean that a user doesn't necessarily have to have access to the templates folder?

pkg/providers/vsphere/vsphere.go
```diff
@@ -680,31 +710,26 @@ func (vs *VsphereTemplateBuilder) GenerateCAPISpecWorkers(clusterSpec *cluster.S
...
-	kubeletExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs().
-		Append(clusterapi.ResolvConfExtraArgs(clusterSpec.Cluster.Spec.ClusterNetwork.DNS.ResolvConf)).
-		Append(clusterapi.ControlPlaneNodeLabelsExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration))
-	apiServerExtraArgs := clusterapi.OIDCToExtraArgs(clusterSpec.OIDCConfig).
-		Append(clusterapi.AwsIamAuthExtraArgs(clusterSpec.AWSIamConfig)).
-		Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig)).
-		Append(sharedExtraArgs)
-	controllerManagerExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs().
-		Append(clusterapi.NodeCIDRMaskExtraArgs(&clusterSpec.Cluster.Spec.ClusterNetwork))
+type VSphereUserConfig struct {
```
jonathanmeier5, line 713: I bundled these things together because we were reading from env vars all through the code. I was hoping to consolidate the config into some kind of object we could pass around instead. Eventually, I'd love to move these configs into the CliConfig object.

As is, I can move this into the `config/` package.

pkg/providers/vsphere/vsphere.go
```diff
@@ -714,6 +739,35 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec, datacenterSpec v1alpha1.VSphe
...
+	kubeletExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs().
+		Append(clusterapi.ResolvConfExtraArgs(clusterSpec.Cluster.Spec.ClusterNetwork.DNS.ResolvConf)).
+		Append(clusterapi.ControlPlaneNodeLabelsExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration))
+	apiServerExtraArgs := clusterapi.OIDCToExtraArgs(clusterSpec.OIDCConfig).
+		Append(clusterapi.AwsIamAuthExtraArgs(clusterSpec.AWSIamConfig)).
+		Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig)).
+		Append(sharedExtraArgs)
+	controllerManagerExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs().
+		Append(clusterapi.NodeCIDRMaskExtraArgs(&clusterSpec.Cluster.Spec.ClusterNetwork))
+
```
jonathanmeier5, line 768: I didn't add this code. I'm not sure if my git history is goofed somehow. I'll see if I can remove it from the diff when I rebase and squash.

pkg/providers/vsphere/vsphere_test.go
```diff
@@ -429,17 +430,35 @@ func newProviderWithGovc(t *testing.T, datacenterConfig *v1alpha1.VSphereDatacen
...
+		var privs []string
+		err := json.Unmarshal([]byte(vsphereAdminPrivsFile), &privs)
+		if err != nil {
+			return nil, err
+		}
+
+		vsc.EXPECT().GetPrivsOnEntity(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(privs, nil).AnyTimes()
+		return vsc, nil
+	}
+	validator := NewValidator(govc, netClient, vSphereClientCallable)
```
jonathanmeier5, line 448: I feel like this code is messy. I needed a way to inject a dummy vsphere client into the validator for my tests. I am extremely open to alternative suggestions.

pkg/providers/vsphere/vsphere.go
```diff
@@ -431,6 +437,30 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
...
 
+	vuc := NewVsphereUserConfig()
+
+	if err := p.validator.validateUserPrivs(ctx, vSphereClusterSpec, vuc); err != nil {
+		return err
+	} else {
+		logger.MarkPass("EKSA_VSPHERE_USER vSphere privileges validated")
+	}
+
+	if len(vuc.EksaVsphereCPUsername) > 0 {
```
jonathanmeier5, line 448: Verify that CP/CSI values exist *and* usernames are different from `EKSA_VSPHERE_USERNAME`.

pkg/providers/vsphere/govmomiclient.go
```diff
@@ -0,0 +1,103 @@
...
+	vsc := VMOMIClient{gvmc, f, username}
+
+	return &vsc, err
+}
+
+func (vsc *VMOMIClient) Username() string {
+	return vsc.username
+}
+
+func (vsc *VMOMIClient) GetPrivsOnEntity(ctx context.Context, path string, objType string, username string) ([]string, error) {
```
jonathanmeier5, line 63: I should probably add a unit test on this function.

pkg/providers/vsphere/validator.go
jonathanmeier5: This validation can be conditional. If the template field in vsphere config object has been configured, we can *skip* the validation.

If we *do* apply the validation, we need to do so on:
- control plane template directory
- worker node template directory
- external etcd directory

pkg/providers/vsphere/validator.go
```diff
@@ -385,3 +403,184 @@ func (v *Validator) validateControlPlaneIpUniqueness(spec *Spec) error {
...
+			"type":         vsphereTypeFolder,
+			"privsContent": config.VSphereCnsSearchAndSpbmPrivsFile,
+			"path":         vsphereRootPath,
+		},
+		{ // CNS-VM role
+			"type":         vsphereTypeFolder,
+			"privsContent": config.VSphereCnsVmPrivsFile,
+			"path":         controlPlaneMachineConfig.Spec.Folder,
+		},
+		// CNS-HOST-CONFIG-STORAGE I don't know what this applies to yet
```
jonathanmeier5, line 487: Ask Vignesh or Abhinav about where this should be applied.

pkg/providers/vsphere/validator.go
jonathanmeier5: Have a look at this, might be some inspo: https://github.com/aws/eks-anywhere/pull/2887

pkg/providers/vsphere/validator.go
jonathanmeier5: Based on this conversation, I think this role only needs to be applied to the datastore.

pkg/config/vsphereuser.go
```diff
@@ -0,0 +1,82 @@
...
+var VSphereCnsVmPrivsFile string
+
+//go:embed static/cnsHostConfigStorage.json
+var VSphereCnsHostConfigStorageFile string
+
+//go:embed static/readOnlyPrivs.json
+var VSphereReadOnlyPrivs string
+
+func NewVsphereUserConfig() *VSphereUserConfig {
+	eksaVsphereUsername := os.Getenv(EksavSphereUsernameKey)
```
vivek-koppuru, line 53: Do we want to do validations of whether they are empty or not, or will the govmomi api handle it?

pkg/dependencies/factory.go
```diff
@@ -1046,8 +1046,12 @@ func (f *Factory) WithVSphereValidator() *Factory {
 		if f.dependencies.VSphereValidator != nil {
 			return nil
 		}
-
-		f.dependencies.VSphereValidator = vsphere.NewValidator(f.dependencies.Govc, &networkutils.DefaultNetClient{})
+		vb := vsphere.NewValidatorBuilder(
```
vivek-koppuru, line 1049: Take a look at what we adopted now, we decided not to use a validator builder but instead pull the clients as part of the validator methods: https://github.com/aws/eks-anywhere/pull/2887

pkg/providers/vsphere/envars.go
```diff
@@ -6,23 +6,24 @@ import (
...
 	} else {
-		return fmt.Errorf("%s is not set or is empty", EksavSphereUsernameKey)
+		return fmt.Errorf("%s is not set or is empty", config.EksavSphereUsernameKey)
 	}
 
-	if vSpherePassword, ok := os.LookupEnv(EksavSpherePasswordKey); ok && len(vSpherePassword) > 0 {
+	if vSpherePassword, ok := os.LookupEnv(config.EksavSpherePasswordKey); ok && len(vSpherePassword) > 0 {
 		if err := os.Setenv(vSpherePasswordKey, vSpherePassword); err != nil {
-			return fmt.Errorf("unable to set %s: %v", EksavSpherePasswordKey, err)
+			return fmt.Errorf("unable to set %s: %v", config.EksavSpherePasswordKey, err)
```
vivek-koppuru, line 23: Is there where the above env vars get validated?

pkg/providers/vsphere/govmomiclient.go
```diff
@@ -0,0 +1,107 @@
...
+	var objRef types.ManagedObjectReference
```
vivek-koppuru, line 68: What does this object ref refer to?

pkg/providers/vsphere/validator.go
```diff
@@ -385,3 +419,234 @@ func (v *Validator) validateControlPlaneIpUniqueness(spec *Spec) error {
...
+			return err
+		} else if len(missingPrivs) > 0 {
+			return fmt.Errorf("User %s missing privileges on %s: %v", vsc.Username(), path, missingPrivs)
+		}
+	}
+
+	return nil
+}
+
+func diffPrivs(requiredPrivs []string, hasPrivs []string) []string {
```
vivek-koppuru, line 621: Diff makes it sound like we are running a diff, but this is almost a subset of a diff imo
```suggestion
func checkRequiredPrivs(requiredPrivs []string, hasPrivs []string) []string {
```

vivek-koppuru (commented):
I think I understand the validations overall and seems like the right approach. Let's make sure that we are able to test it against our vcenter as well.

pkg/config/vsphereuser.go
```diff
@@ -0,0 +1,82 @@
+package config
+
+import (
+	_ "embed"
+	"os"
+)
+
+const (
+	EksavSphereUsernameKey = "EKSA_VSPHERE_USERNAME"
```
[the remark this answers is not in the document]
jonathanmeier5: Looks like they already existed in two places:
https://github.com/aws/eks-anywhere/blob/b1df4a1b328b192eded61663f3e352682eb6abe9/pkg/executables/govc.go#L36-L36

https://github.com/aws/eks-anywhere/blob/b1df4a1b328b192eded61663f3e352682eb6abe9/pkg/providers/vsphere/vsphere.go#L41-L41

It's also defined in many different test files. I feel like a good first step is making `executables/govc` import and use the const from config to button things up, and we can get the test usages later?

pkg/config/vsphereuser.go
```diff
@@ -0,0 +1,82 @@
...
+	EksaVsphereUsername    string
+	EksaVspherePassword    string
+	EksaVsphereCPUsername  string
+	EksaVsphereCPPassword  string
+	EksaVsphereCSIUsername string
+	EksaVsphereCSIPassword string
+}
+
+//go:embed static/globalPrivs.json
+var VSphereGlobalPrivsFile string
```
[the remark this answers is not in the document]
jonathanmeier5: I actually did that originally, but it meant I didn't get linting errors if a particular file wasn't present. I'm kind of paranoid about forgetting to bundle static assets in builds because python makes it suuuper easy to shoot yourself in the foot this way when creating packages.

Agreed it's not great to need to add each new file here.

pkg/providers/vsphere/govmomiclient.go
jonathanmeier5: This will be a ref to an object in vsphere. I've tried renaming it, let me know if you think it's clearer?

pkg/providers/vsphere/validator.go
jonathanmeier5: committed locally 👍

pkg/providers/vsphere/envars.go
jonathanmeier5: yup 👍

pkg/config/vsphereuser.go
jonathanmeier5: already in place 👍 
https://github.com/aws/eks-anywhere/pull/2907#discussion_r949279785

pkg/config/vsphereuser.go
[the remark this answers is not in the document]
vivek-koppuru: I think this is a case where we need to keep the validation of env vars in govc somehow and pass in parameters to the check setup method to not have multiple instances of these.

pkg/providers/vsphere/govmomiclient.go
vivek-koppuru: What is the official name that vcenter uses? I might like `resource` more otherwise

pkg/config/vsphereuser.go
```diff
@@ -0,0 +1,82 @@
...
+//go:embed static/cnsVmPrivs.json
+var VSphereCnsVmPrivsFile string
+
+//go:embed static/cnsHostConfigStorage.json
+var VSphereCnsHostConfigStorageFile string
+
+//go:embed static/readOnlyPrivs.json
+var VSphereReadOnlyPrivs string
+
+func NewVsphereUserConfig() *VSphereUserConfig {
```
jonathanmeier5, line 52: Add tests for this function

pkg/providers/vsphere/govmomiclient.go
jonathanmeier5: I don't have a strong opinion here and am happy to go with whatever you think is best.

In general, I really don't like "object" for variable names because it's so vague. At the same time, golang's naming conventions go against a lot of what I'm used to, so I've been tossing my personal preferences out the window lately. 😅

I unfortunately do see "object" in a lot of the docs, but again am happy to go with whatever you think is best:
https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenterhost.doc/GUID-031BDB12-D3B2-4E2D-80E6-604F304B4D0C.html

https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenterhost.doc/GUID-7FDFBDBE-F8AC-4D00-AE5E-3F14D7472FAF.html

https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenterhost.doc/GUID-E8E854DD-AA97-4E0C-8419-CE84F93C4058.html

I am guessing that these persistent references to "objects" in the govmomi code derive from the "object" language in the vsphere docs.

pkg/config/vsphereuser.go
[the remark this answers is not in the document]
jonathanmeier5: Agreed. I've filed concerns in a ticket [here](https://github.com/aws/eks-anywhere/issues/3071).

pkg/providers/vsphere/vsphere.go
```diff
@@ -149,22 +142,34 @@ type ClusterResourceSetManager interface {
 	ForceUpdate(ctx context.Context, name, namespace string, managementCluster, workloadCluster *types.Cluster) error
 }
 
+type ValidatorBuilder interface {
```
vivek-koppuru, line 145: Is it possible to avoid a `ValidatorBuilder` now that #2887 is merged?

vivek-koppuru (commented):
Left some more comments, let me know what you think

pkg/providers/vsphere/vsphere.go
jonathanmeier5: Oh i meant to strip this out, yes 👍

pkg/providers/vsphere/validator.go
```diff
@@ -385,3 +406,234 @@ func (v *Validator) validateControlPlaneIpUniqueness(spec *Spec) error {
...
+			path:         vsphereRootPath,
+		},
+		{
+			objectType:   vsphereTypeNetwork,
+			privsContent: config.VSphereUserPrivsFile,
+			path:         spec.datacenterConfig.Spec.Network,
+		},
+	}
+
+	var newRas []RequiredAccess
```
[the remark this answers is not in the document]
jonathanmeier5: These names are so bad, haha. I'll make them better.

pkg/providers/vsphere/vsphere.go
```diff
@@ -431,6 +437,30 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
...
+		} else {
+			logger.MarkPass("EKSA_VSPHERE_CP_USER vSphere privileges validated")
+		}
+	}
+
+	if len(vuc.EksaVsphereCSIUsername) > 0 && vuc.EksaVsphereCSIUsername != vuc.EksaVsphereUsername {
+		if err := p.validator.validateCSIUserPrivs(ctx, vSphereClusterSpec, vuc); err != nil {
+			return err
+		} else {
+			logger.MarkPass("EKSA_VSPHERE_CSI_USER vSphere privileges validated")
```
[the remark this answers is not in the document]
jonathanmeier5: The documentation mentions a few permissions that are not included in the set assigned to `EKSA_VSPHERE_USERNAME` per our docs. However, I was not able to find anything broken in the driver when deploying without them. Presumably the majority of our users have been deploying without them and we haven't heard any complaints?

Aside from those one or two perms, the CSI and CP user privs are just subsets of EKSA_VSPHERE_USER's perms.

I think in this situation I'd rather the validator be more permissive than restrictive until we can find an example of the software breaking without the permissions. What do you think?

pkg/providers/vsphere/govmomiclient.go
```diff
@@ -0,0 +1,152 @@
...
+	Network(ctx context.Context, path string) (object.NetworkReference, error)
+	ResourcePool(ctx context.Context, path string) (*object.ResourcePool, error)
+	VirtualMachine(ctx context.Context, path string) (*object.VirtualMachine, error)
+}
+
+type VMOMIAuthorizationManager interface {
+	FetchUserPrivilegeOnEntities(ctx context.Context, entities []types.ManagedObjectReference, userName string) ([]types.UserPrivilegeResult, error)
+}
+
+type VSphereClient interface {
```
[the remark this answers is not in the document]
jonathanmeier5: I generate a mnock object off of it for tests in `vsphere_test.go`. https://github.com/aws/eks-anywhere/blob/73b232f33062de067ef0a8668ebb0d2fd6db9386/pkg/providers/vsphere/validator_test.go#L19-L19

pkg/providers/vsphere/govmomiclient.go
```diff
@@ -0,0 +1,152 @@
...
+	AuthorizationManager VMOMIAuthorizationManager
+}
+
+type VSphereClientBuilder interface {
+	Build(ctx context.Context, host string, username string, password string, insecure bool, datacenter string) (VSphereClient, error)
+}
+
+type VMOMIClientBuilder struct{}
+
+func (*VMOMIClientBuilder) Build(ctx context.Context, host string, username string, password string, insecure bool, datacenter string) (VSphereClient, error) {
```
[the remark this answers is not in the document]
jonathanmeier5: Yeah, I've been going back and forth on this. The test will have to just verify that a bunch of mocks are called correctly. It seemed like a lot of extra code to test ~15loc, but you're right that we should have something around it. 

Also has the benefit of pushing us towards structuring the builder better with a DI approach.

pkg/providers/vsphere/vsphere.go
```diff
@@ -431,6 +435,30 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
 		logger.Info("Skipping check for whether control plane ip is in use")
 	}
 
+	vuc := config.NewVsphereUserConfig()
+
+	if err := p.validator.validateUserPrivs(ctx, vSphereClusterSpec, vuc); err != nil {
+		return err
+	} else {
+		logger.MarkPass("EKSA_VSPHERE_USER vSphere privileges validated")
```
vivek-koppuru, line 443: All of these should probably reference "USERNAME" if that's how we documented and read them, and then maybe we can say the following instead. Or better if we can just output the value of the env var below instead
```suggestion
		logger.MarkPass("EKSA_VSPHERE_USERNAME user vSphere privileges validated")
```

pkg/providers/vsphere/govmomiclient.go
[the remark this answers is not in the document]
vivek-koppuru: I am wondering if we are interfacing too much here but I can't think of an alternate solution except for maybe passing in values like datacenter on each validate command so we can potentially remove the need of a builder. We can brainstorm maybe as I want to avoid overcomplicating and it getting stuck in code.

vivek-koppuru (approved):

pkg/providers/vsphere/vsphere.go
```diff
@@ -431,6 +435,30 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
 		logger.Info("Skipping check for whether control plane ip is in use")
 	}
 
+	vuc := config.NewVsphereUserConfig()
+
+	if err := p.validator.validateUserPrivs(ctx, vSphereClusterSpec, vuc); err != nil {
+		return err
+	} else {
+		logger.MarkPass("EKSA_VSPHERE_USERNAME vSphere privileges validated")
```
vivek-koppuru, line 443: Nit: Let's print the values of this env var and add "user" after that name

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #2915, opened 2022-08-10
Capc v1beta2

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR re-introduces the changes required to support CAPC v1beta2 (originally reverted in https://github.com/aws/eks-anywhere/pull/2818), and fixes a few bugs identified through further testing. The overall goal of the migration is to support distributing an EKS-A cluster across multiple CloudStack endpoints with multiple sets of credentials.

The PR is large, but the **majority of the changes are reverts of a revert and were already reviewed**, just blocked by a CAPC release. An official release candidates tag for CAPC is expected by EOD 8/10/22 and then we can change the go.mod file to use it instead.

*Testing (if applicable):*
Manual runs of the following e2e tests passed when using the custom-built EKS-A controller, and the CAPC controller/components bundle overrides:

1. TestCloudStackKubernetes121SimpleFlow (create/delete)
2. TestCloudStackKubernetes120RedhatTo121Upgrade (K8s version upgrade)
3. TestCloudStackKubernetes121WorkloadCluster (create/delete with workload cluster)
4. TestCloudStackKubernetes120GitopsOptionsFluxLegacy (flux upgrade)
5. TestCloudStackKubernetes121StackedEtcdRedhat


pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -81,45 +83,67 @@ func (r *CloudStackDatacenterConfig) ValidateUpdate(old runtime.Object) error {
...
+	r := regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
+	return r.MatchString(uuid)
+}
 
-	if old.Spec.Domain != new.Spec.Domain {
-		allErrs = append(
-			allErrs,
-			field.Invalid(field.NewPath("spec", "domain"), new.Spec.Domain, "field is immutable"),
-		)
+func isCapcV1beta1ToV1beta2Upgrade(new, old *CloudStackDatacenterConfigSpec) bool {
```
maxdrib, line 91: TODO: Add unit tests

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -81,45 +83,67 @@ func (r *CloudStackDatacenterConfig) ValidateUpdate(old runtime.Object) error {
...
-		allErrs = append(
-			allErrs,
-			field.Invalid(field.NewPath("spec", "managementApiEndpoint"), new.Spec.ManagementApiEndpoint, "field is immutable"),
-		)
+func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConfig) field.ErrorList {
+	var allErrs field.ErrorList
+
+	// Check for CAPC v1beta1 -> CAPC v1beta2 upgrade
+	if isCapcV1beta1ToV1beta2Upgrade(&new.Spec, &old.Spec) {
+		return allErrs
```
g-gaston, line 114: Unit test for this?

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -81,45 +83,67 @@ func (r *CloudStackDatacenterConfig) ValidateUpdate(old runtime.Object) error {
...
+	}
+	newAzMap := make(map[string]*CloudStackAvailabilityZone)
```
g-gaston, line 116: do these need to be pointers?

pkg/executables/kubectl.go
```diff
@@ -128,6 +128,7 @@ func (k *Kubectl) GetEksaCloudStackDatacenterConfig(ctx context.Context, cloudst
 	if err != nil {
 		return nil, fmt.Errorf("getting eksa cloudstack datacenterconfig: %v", err)
 	}
+	response.SetDefaults()
```
g-gaston, line 131: Why are we doing this? IMO, this is a mistake, we shouldn't have these kind of logic at the client level.

We either set the defaults in a mutation webhook, or we set them from the cli when we create/update the object or we have some extra logic at the usecase level (maybe the provider?)

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -641,6 +641,9 @@ func (cs *CloudStackTemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *c
 	if clusterSpec.Cluster.Spec.ExternalEtcdConfiguration != nil {
 		etcdMachineSpec = *cs.etcdMachineSpec
 	}
+	if clusterSpec.CloudStackDatacenter == nil {
+		return nil, fmt.Errorf("provided clusterSpec CloudStackDatacenter is nil. Unable to build template map CP")
```
g-gaston, line 645: If this is a concerning scenario, we should probably have this as part of the validation layer, not here.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1003,6 +1006,7 @@ func (p *cloudstackProvider) generateCAPISpecForUpgrade(ctx context.Context, boo
 	if err != nil {
 		return nil, nil, err
 	}
+	currentSpec.CloudStackDatacenter = csdc
```
g-gaston, line 1009: why are we modifying the cluster spec here? This should be set during the object construction. This creates a dangerous side effect/invariant

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1169,6 +1173,7 @@ func (p *cloudstackProvider) UpgradeNeeded(ctx context.Context, newSpec, current
 	if err != nil {
 		return false, err
 	}
+	currentSpec.CloudStackDatacenter = existingCsdc
```
g-gaston, line 1176: same comment here about side effect

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
maxdrib: Nope they certainly don't, will change

pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
maxdrib: Will add today

pkg/providers/cloudstack/cloudstack.go
maxdrib: From what I can tell, currentSpec's plumbing into this method can be traced back through

* generateCAPISpecForUpgrade (cloudstack.go)
* GenerateCAPISpecForUpgrade (cloudstack.go)
* clusterManager calling GetCurrentClusterSpec (cluster_manager.go)

So the cluster spec is originally constructed in GetCurrentClusterSpec. However, there is no logic anywhere to account for setting the provider-specific specs, like spec.CloudStackDatacenter, which resulted in currentSpec.CloudStackDatacenter being `nil`. I'm not sure where else this logic could exist, and without persisting CloudStackDatacenterConfig in the provider, we need to have currentSpec.CloudStackDatacenter set somewhere.

pkg/executables/kubectl.go
maxdrib: How would we go about implementing a mutation webhook? The need for this call is when upgrading older clusters which are not using AvailabilityZones to the new AvailabilityZones schema, which is used internally in EKS-A throughout both the CLI and controller, as we are deprecating the old schema.

wongni: When upgrading a cluster created from v1beta1 capc to eksa CLI with v1beta2 the retrieved cloudstack datacenter config is v1beta1 so we need conversion so I used SetDefaults() here in the original PR. If we think this is not a good place to call the function then we have to call it from the callers. What do you think?

pkg/providers/cloudstack/cloudstack.go
wongni: We used to pass the cloudstack datacenter config (aka. csdc) to other functions as a separate parameter. However, as per Joey's suggestion I removed it because it is a part of cloudstack cluster spec (aka. spec). During the upgrade test I found the spec had a nil for its csdc so I set it to avoid nil pointer dereference error. If we don't want to do this we should go back to the original function prototypes where csdc is explicitly passed in.

pkg/providers/cloudstack/cloudstack.go
wongni: Please refer to https://github.com/aws/eks-anywhere/pull/2915#discussion_r943642635

pkg/executables/kubectl.go
g-gaston: @maxdrib you can find mutation webhook for other CRDs: in the `_webhook.go` files in the api package

@wongni I'm ok calling this from the caller. I hope we only need to do it once. If we require this new format even in old objects, we should be able to pull, update and pull again.

maxdrib: I don't know if mutation webhooks will work for our usecase. From what I can tell, mutation webhooks are intended to modify resources _before_ they're persisted in etcd. In our case, we have _existing_ resources in etcd from an older cluster which need to be converted to a new schema to be used in the EKS-A codebase, which can't seem to be done with mutating webhooks.

pkg/providers/cloudstack/cloudstack.go
g-gaston: I see. I think we just need to modify the code that builds a clusterSpec from the CRs in the cluster (what's called here `currentSpec`) and make sure it pulls the cloudstack datacenterconfig and set it in the `cluster.Config`

maxdrib: >  I think we just need to modify the code that builds a clusterSpec from the CRs

There doesn't appear to be any provider specific logic in [the code that builds the clusterSpecs](https://github.com/aws/eks-anywhere/blob/23b12d10a9614b62edc86720f587c09f32ca0149/pkg/clustermanager/cluster_manager.go#L1038) which would enable setting this field. Are you suggesting we modify this function somehow? Otherwise it seems like the provider-specific implementation should happen here.

controllers/resource/template.go
```diff
@@ -175,6 +175,7 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster
 	for _, wnConfig := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		workerNodeGroupMachineSpecs[wnConfig.MachineGroupRef.Name] = workerCsmcs[wnConfig.MachineGroupRef.Name].Spec
 	}
+	clusterSpec.CloudStackDatacenter = &csdc
```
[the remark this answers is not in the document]
maxdrib: That's what I've been going back and forth with @g-gaston about. It seems like the CloudStackDatacenter is never present when we build the clusterSpec and I don't see why it would be. The code to build the clusterSpec is provider agnostic, so I'm not sure where the CloudStackDatacenter would be set, unless it has something to do with Config Processors/ConfigManagerEntry stuff? I'm not familiar with those pieces.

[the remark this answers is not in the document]
jiayiwang7: I think the reason is we build the spec in controller by https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/spec.go#L262 which does not update `Config.cloudstackdatacenter` object. The build spec method used in CLI is https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/spec.go#L184 where we use config manager and properly set up the provider spec.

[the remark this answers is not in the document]
jiayiwang7: https://github.com/aws/eks-anywhere/issues/2929

pkg/providers/cloudstack/cloudstack.go
maxdrib: Created https://github.com/aws/eks-anywhere/issues/2929 per Joey's advice

Makefile
```diff
@@ -599,7 +599,7 @@ release-manifests: $(KUSTOMIZE) generate-manifests $(RELEASE_DIR) $(CONTROLLER_M
 	cp $(RELEASE_DIR)/$(RELEASE_MANIFEST_TARGET) $(CONTROLLER_MANIFEST_OUTPUT_DIR)
 
 .PHONY: run-controller # Run eksa controller from local repo with tilt
-run-controller:
+run-controller: $(KUSTOMIZE)
```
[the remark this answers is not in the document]
maxdrib: Yes we do. `manager/Tiltfile` expects a Kustomize binary to be present.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -166,10 +166,11 @@ func (p *cloudstackProvider) ValidateNewSpec(ctx context.Context, cluster *types
 		return err
 	}
 
-	datacenter := p.datacenterConfig
+	prevDatacenter.SetDefaults()
 
-	oSpec := prevDatacenter.Spec
-	nSpec := datacenter.Spec
+	if err = clusterSpec.CloudStackDatacenter.ValidateUpdate(prevDatacenter); err != nil {
```
jiayiwang7, line 171: currently `ValidateUpdate` is only used for webhook. It makes sense to call it in CLI here since all the validations are same (as for now). We should be careful about adding new validation logic in the webhook since it also affects CLI.

jiayiwang7 (approved):

PR #2916, opened 2022-08-10
[Bundles] Move etcdadm and crictl to be os family independent

vignesh-goutham:
*Description of changes:*
Etcdadm and Crictl are under eksD/<image-format>/image-name/. We currently do not build a separate versions of etcdadm & crictl for each OS/format type.

This change makes etcdadm and crictl independent of os family and os image format type.

/hold


pokearu (approved):
/approve

PR #2922, opened 2022-08-11
Separating clusterConfig validation from creation

raymond-zhang00:
*Issue #, if available:*

*Description of changes:*
Splitting cluster config validation out from clusterSpec creation

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/spec.go
+++ b/pkg/cluster/spec.go
@@ -190,9 +190,6 @@ func NewSpecFromClusterConfig(clusterConfigPath string, cliVersion version.Info,
 	if err = configManager.SetDefaults(clusterConfig); err != nil {
 		return nil, err
 	}
-	if err = ValidateConfig(clusterConfig); err != nil {
-		return nil, err
-	}
 
 	versionsBundle, err := GetVersionsBundle(clusterConfig.Cluster, bundlesManifest)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/cluster/spec.go
```diff
@@ -168,6 +168,17 @@ func newWithCliVersion(cliVersion version.Info, opts ...SpecOpt) *Spec {
 	return NewSpec(opts...)
 }
 
+func NewSpecFromClusterConfigWithValidate(clusterConfigPath string, cliVersion version.Info, opts ...SpecOpt) (*Spec, error) {
```
g-gaston, line 171: Curious, do we need to implement this in the `cluster` package? This package already provides the building blocks to implement it and it seems like all the valid usecases come from the cmd package, so I wonder if we should just put it there. My main concern is we are trying to separate these 2 operations, reading and validating. By offering this in the package interface we are kind of perpetuating that pattern. 

I see that test/framework and internal/test both were calling this method, but I don't think either of them need the validation part.

wdyt?

g-gaston, line 171: another way of naming this
```suggestion
func ReadAndValidateSpecFromClusterConfig(clusterConfigPath string, cliVersion version.Info, opts ...SpecOpt) (*Spec, error) {
```

g-gaston (commented):
thanks a lot for doing this!

pkg/cluster/spec.go
raymond-zhang00: Think you're right and going with the cmd pkg will be a good idea to keep the validate and spec creation separate. Adding the function together also resulted in some redundant testing being recommended.

Will move packages.

raymond-zhang00: Naming the new cmd pkg function readAndValidateClusterSpec

g-gaston (approved):

PR #2923, opened 2022-08-11
Revert changes on where to pull harbor helm chart

vincentni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


ewollesen (approved):
We should probably try to resolve these names, as a few of them at least are pointing at the same things.

/lgtm

PR #2939, opened 2022-08-15
Adding token refresher to EKS-A release bundle

jonahjon:

Current tests are failing  since the new image isn't being pushed to Public ECR. This is adding it to the bundle.


release/api/v1alpha1/bundle_types.go
```diff
@@ -235,6 +235,7 @@ type FluxBundle struct {
 type PackageBundle struct {
 	Version    string `json:"version,omitempty"`
 	Controller Image  `json:"packageController"`
+	Refreshser Image  `json:"tokenRefresher"`
```
abhay-krishna, line 238: typo
```suggestion
	Refresher Image  `json:"tokenRefresher"`
```

abhay-krishna: I would actually prefer to use `TokenRefresher` as the field name to make intentions clear.

TerryHowe:
conflicts!

TerryHowe:
I resolved the conflicts

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
```diff
@@ -39,12 +39,13 @@ spec:
               of CloudStackDatacenterConfig
             properties:
               account:
```
TerryHowe, line 41: This change looks like it shouldn't be here

config/manifest/eksa-components.yaml
```diff
@@ -3056,12 +3086,13 @@ spec:
               of CloudStackDatacenterConfig
             properties:
               account:
```
TerryHowe, line 3088: and these changes

config/crd/bases/anywhere.eks.amazonaws.com_cloudstackdatacenterconfigs.yaml
TerryHowe, line 41: It would be nice to get rid of these cloudstack changes.

TerryHowe:
tests borked

jonahjon:
Looks to be an issue due to a recent cloudtstack PR https://github.com/aws/eks-anywhere-build-tooling/pull/1170

release/pkg/bundles/package-controller.go
```diff
@@ -23,42 +23,50 @@ import (
 	anywherev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 	"github.com/aws/eks-anywhere/release/pkg/constants"
 	releasetypes "github.com/aws/eks-anywhere/release/pkg/types"
+	bundleutils "github.com/aws/eks-anywhere/release/pkg/util/bundles"
 	"github.com/aws/eks-anywhere/release/pkg/version"
 )
 
 func GetPackagesBundle(r *releasetypes.ReleaseConfig, imageDigests map[string]string) (anywherev1alpha1.PackageBundle, error) {
-	artifacts := r.BundleArtifactsTable["eks-anywhere-packages"]
+	//artifacts := r.BundleArtifactsTable["eks-anywhere-packages"]
```
a-cool-train, line 31: Can we remove this from the code if it is not necessary?

a-cool-train (changes requested):
@jonahjon, we need to also extract the ecr refresher cron job in case of download images and import images. This is to support the disconnected area.

jonahjon:
> @jonahjon, we need to also extract the ecr refresher cron job in case of download images and import images. This is to support the disconnected area.

What does that mean?

jonahjon:
> @jonahjon, we need to also extract the ecr refresher cron job in case of download images and import images. This is to support the disconnected area.

Also can we just do that in another PR since this one is blocking the release pipeline.

jonahjon:
tide

jonahjon:
Why prow decides to re-test 7 jobs that passed a few minutes ago is a mystery to me.

PR #2940, opened 2022-08-16
Fix dev-release S3 URI for tarball archives

abhay-krishna:
The release version var has the `+` that ends up in the archive URI, which results in an invalid S3 URI. The dev release URI version has `+` replaced with `-` so we should use this to construct S3/ECR URIs.


jaxesn (approved):

TerryHowe (approved):
/lgtm
/approve

vignesh-goutham (approved):
/lgtm

PR #2947, opened 2022-08-16
Added prereqs to docs for each curated package

chrisnegus:
*Description of changes:*
Created a short prereqs section for curated packages to be read into docs for each package. To do this, I moved content out of tasks/packages/_index.md.

### DIFF
--- a/docs/content/en/docs/tasks/packages/prereq.md
+++ b/docs/content/en/docs/tasks/packages/prereq.md
@@ -1,0 +1,33 @@
+## Prerequisites
+Before installing any curated packages for EKS Anywhere, do the following:
+
+* Check that the version of `eksctl anywhere` is `v0.9.0` or above with the following commands:
+    ```bash
+    eksctl anywhere version
+    ```
+    Example command output
+    ```
+    NAME                                       READY   STATUS     RESTARTS   AGE
+    eks-anywhere-packages-57778bc88f-587tq     2/2     Running    0          16h
+    ```
+* Make sure cert-manager is up and running in the cluster. Note cert-manager is not installed on workload clusters by default. If cert-manager is not installed, you can manually install cert-manager and follow the instructions below to finish the package controller installation.
+
+* Check the existence of package controller:
+
+    ```bash
+    kubectl get pods -n eksa-packages | grep "eks-anywhere-packages"
+    ```
+    If the returned result is empty, you need to install the package controller.
+
+* If the packages controller is not yet installed, follow these steps:
+
+    1. Install the package controller
+        ```bash
+        export CURATED_PACKAGES_SUPPORT=true
+        eksctl anywhere install packagecontroller -f $CLUSTER_NAME.yaml
+        ```
+
+    1. Check the package controller
+        ```bash
+        kubectl get pods -n eksa-packages
+        ```
--- a/docs/content/en/docs/tasks/packages/_index.md
+++ b/docs/content/en/docs/tasks/packages/_index.md
@@ -9,41 +9,4 @@ description: >
 
 The main goal of EKS Anywhere curated packages is to make it easy to install, configure and maintain operational components in an EKS Anywhere cluster. EKS Anywhere curated packages offers to run secure and tested operational components on EKS Anywhere clusters. Please check out [EKS Anywhere curated packages concepts]({{< relref "../../concepts/packages" >}}) and [EKS Anywhere curated packages configurations]({{< relref "../../reference/clusterspec/packages.md" >}}) for more details.
 
-### Check the existence of package controller
-```bash
-kubectl get pods -n eksa-packages | grep "eks-anywhere-packages"
-```
-Skip the following installation steps if the returned result is not empty.
-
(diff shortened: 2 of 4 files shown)


docs/content/en/docs/tasks/packages/prereq.md
```diff
@@ -0,0 +1,33 @@
...
+    ```bash
+    kubectl get pods -n eksa-packages | grep "eks-anywhere-packages"
+    ```
+    If the returned result is empty, you need to install the package controller.
+
+* If the packages controller is not yet installed, follow these steps:
+
+    1. Install the package controller
+        ```bash
+        export CURATED_PACKAGES_SUPPORT=true
```
binyamz, line 26: feature flag CURATED_PACKAGES_SUPPORT is no longer needed after this release.

docs/content/en/docs/tasks/packages/prereq.md
```diff
@@ -0,0 +1,33 @@
+## Prerequisites
+Before installing any curated packages for EKS Anywhere, do the following:
+
+* Check that the version of `eksctl anywhere` is `v0.9.0` or above with the following commands:
+    ```bash
+    eksctl anywhere version
```
rothgar, line 6: The below output isn't checking the version. It's for the package controller output

docs/content/en/docs/tasks/packages/prereq.md
```diff
@@ -0,0 +1,33 @@
...
+    kubectl get pods -n eksa-packages | grep "eks-anywhere-packages"
+    ```
+    If the returned result is empty, you need to install the package controller.
+
+* If the packages controller is not yet installed, follow these steps:
+
+    1. Install the package controller
+        ```bash
+        export CURATED_PACKAGES_SUPPORT=true
+        eksctl anywhere install packagecontroller -f $CLUSTER_NAME.yaml
```
rothgar, line 27: You're missing a step where you add the package controller to the cluster config.

docs/content/en/docs/tasks/packages/prereq.md
```diff
@@ -0,0 +1,33 @@
+## Prerequisites
+Before installing any curated packages for EKS Anywhere, do the following:
+
+* Check that the version of `eksctl anywhere` is `v0.9.0` or above with the following commands:
```
TerryHowe, line 4:
```suggestion
* Check that the version of `eksctl anywhere` is `v0.11.0` or above with the following commands:
```

TerryHowe: This whole thing is a bit of a toubleshooting doc. I like the include of this partial, but wonder if we should just link to troubleshooting given one or two checks here

docs/content/en/docs/tasks/packages/prereq.md
TerryHowe: This command is going to be deprecated or maybe it is already too.

docs/content/en/docs/tasks/packages/prereq.md
a-cool-train: @chrisnegus, I am working on doc changes which will be updating these items. I assume this PR is to just move items around and not update the new changes that will be associated with this release. @binyamz, the change specified above is not required now since I am working on those updates.

chrisnegus: @a-cool-train Yes. The intention is to just move items around. Would you prefer I close this PR and have you handle it or would you like to work with this as a base?

docs/content/en/docs/tasks/packages/prereq.md
chrisnegus: Well, I guess it doesn't have to look like a troubleshooting doc. The idea was to let someone know that they must have some things in place before they just try to install a package. By putting it in one file, we can just read it into each packages task doc.

docs/content/en/docs/tasks/packages/prereq.md
chrisnegus: That was my mistake. I'll fix that.

docs/content/en/docs/tasks/packages/prereq.md
chrisnegus: As I noted above, we should decide if we want to update this PR to reflect the new way of setting up the package controller or just close it and add the content separately. Either way, I think having it in a separate prereq file that's read into each task is a good idea.

docs/content/en/docs/tasks/packages/prereq.md
rothgar: We should make sure the docs are versioned properly so this step remains in docs that _do_ require this env

a-cool-train: @chrisnegus, Feel free to get this PR in, then I'll use it as a base.

chrisnegus: Okay. @a-cool-train I'll go ahead and merge this PR and let you make the fixes that are needed before the release.

PR #2950, opened 2022-08-16
Snow control plane reconcile phase

taneyland:
*Issue #, if available:*

*Description of changes:*
Add snow CP reconcile phase

TODO: add unit tests

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/snowdatacenterconfig_types.go
+++ b/pkg/api/v1alpha1/snowdatacenterconfig_types.go
@@ -9,7 +9,17 @@ type SnowDatacenterConfigSpec struct { // Important: Run "make generate" to rege
 }
 
 // SnowDatacenterConfigStatus defines the observed state of SnowDatacenterConfig
-type SnowDatacenterConfigStatus struct{}
+type SnowDatacenterConfigStatus struct {
+	// SpecValid is set to true if snow datacenterconfig is validated.
+	SpecValid bool `json:"specValid,omitempty"`
+
+	// ObservedGeneration is the latest generation observed by the controller.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// FailureMessage indicates that there is a fatal problem reconciling the
+	// state, and will be set to a descriptive error message.
+	FailureMessage *string `json:"failureMessage,omitempty"`
+}
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -84,6 +84,7 @@ type Dependencies struct {
 	VSphereValidator          *vsphere.Validator
 	VSphereDefaulter          *vsphere.Defaulter
 	SnowValidator             *snow.AwsClientValidator
+	SnowDefaulters            *snow.Defaulters
 }
 
 func (d *Dependencies) Close(ctx context.Context) error {
--- a/pkg/providers/snow/reconciler/reconciler.go
+++ b/pkg/providers/snow/reconciler/reconciler.go
@@ -2,19 +2,151 @@ package reconciler
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"time"
 
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/types"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	c "github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/controller"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
+	"github.com/aws/eks-anywhere/pkg/controller/serverside"
+	"github.com/aws/eks-anywhere/pkg/providers/snow"
 )
 
-type Reconciler struct{}
+const defaultRequeueTime = time.Minute
 
-func New() *Reconciler {
-	return &Reconciler{}
+const (
+	controlSpecPlaneAppliedCondition clusterv1.ConditionType = "ControlPlaneSpecApplied"
+	controlPlaneReadyCondition       clusterv1.ConditionType = "ControlPlaneReady"
+	eksaLicense                      string                  = "EKSA_LICENSE"
+)
+
+type Reconciler struct {
+	client     client.Client
+	validator  *snow.AwsClientValidator
+	defaulters *snow.Defaulters
 }
 
-func (v *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+func New(client client.Client, validator *snow.AwsClientValidator, defaulters *snow.Defaulters) *Reconciler {
+	return &Reconciler{
+		client:     client,
+		validator:  validator,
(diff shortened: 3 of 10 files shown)


pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+	specWithBundles, err := c.BuildSpec(ctx, kubeClient, cluster)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	configManager := snow.NewConfigManager(r.defaulters, r.validator)
+	if err = configManager.SetDefaultsAndValidate(ctx, specWithBundles.Config); err != nil {
+		return controller.Result{}, err
+	}
+	log.Info("cluster", "name", cluster.Name)
```
g-gaston, line 80: I think we can skip everything above this line

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -9,7 +9,17 @@ type SnowDatacenterConfigSpec struct { // Important: Run "make generate" to rege
...
+type SnowDatacenterConfigStatus struct {
+	// SpecValid is set to true if snow datacenterconfig is validated.
+	SpecValid bool `json:"specValid,omitempty"`
+
+	// ObservedGeneration is the latest generation observed by the controller.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// FailureMessage indicates that there is a fatal problem reconciling the
+	// state, and will be set to a descriptive error message.
+	FailureMessage *string `json:"failureMessage,omitempty"`
```
g-gaston, line 21: I don't think we need any of these, the datacenter spec for snow doesn't even have fields in the spec, so it can't be invalid :)

pkg/dependencies/factory.go
```diff
@@ -84,6 +84,7 @@ type Dependencies struct {
 	VSphereValidator          *vsphere.Validator
 	VSphereDefaulter          *vsphere.Defaulter
 	SnowValidator             *snow.AwsClientValidator
+	SnowDefaulters            *snow.Defaulters
```
g-gaston, line 87: Interesting, why are we adding this field here if it's not being built in this factory?

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+			}}, err
+		}
+		conditions.MarkTrue(cluster, controlSpecPlaneAppliedCondition)
+	}
+	return controller.Result{}, nil
+}
+
+func (r *Reconciler) getCAPICluster(ctx context.Context, cluster *anywherev1.Cluster, log logr.Logger) (*clusterv1.Cluster, controller.Result, error) {
+	capiCluster := &clusterv1.Cluster{}
+	capiClusterName := types.NamespacedName{Namespace: constants.EksaSystemNamespace, Name: cluster.Name}
```
g-gaston, line 120: We should probably add a method in apibuilder that gives us the capi cluster name based on a eksa cluster

That said, I wonder if we should reference the capi cluster in our cluster status

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+		controlPlaneSpec, err := snow.ControlPlaneSpec(ctx, specWithBundles, kubeClient)
+		if err != nil {
+			return controller.Result{}, err
+		}
+		if err := serverside.ReconcileYaml(ctx, r.client, controlPlaneSpec); err != nil {
+			return controller.Result{Result: &ctrl.Result{
+				RequeueAfter: defaultRequeueTime,
+			}}, err
+		}
+		conditions.MarkTrue(cluster, controlSpecPlaneAppliedCondition)
```
g-gaston, line 113: I think we can get rid of these conditions for now

pkg/providers/snow/snow.go
```diff
@@ -103,27 +103,44 @@ func (p *SnowProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster
...
-	controlPlaneSpec, err = templater.ObjectsToYaml(controlPlaneObjs...)
+	workersSpec, err = WorkersSpec(ctx, clusterSpec, kubeClient)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	return controlPlaneSpec, workersSpec, nil
+}
+
+func ControlPlaneSpec(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) (controlPlaneSpec []byte, err error) {
```
g-gaston, line 119: I don't think we need this, we can reconcile `unstructured.Unstructured` directly

g-gaston (approved):

g-gaston (commented):
Undoing the approve, wrong button :(

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+	if err := r.client.Get(ctx, dataCenterName, dataCenterConfig); err != nil {
+		return controller.Result{}, err
+	}
+
+	if err := setupEnvVars(ctx, r.client); err != nil {
+		log.Error(err, "Failed to set up snow env vars")
+		return controller.Result{}, err
+	}
+
+	if !dataCenterConfig.Status.SpecValid {
```
vivek-koppuru, line 59: So based on @g-gaston comment above, we might not even need this check? But we should check for the machineconfig too

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+	if err := r.client.Get(ctx, capiClusterName, capiCluster); err != nil {
+		return nil, controller.Result{Result: &ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: defaultRequeueTime,
+		}}, err
+	}
+	return capiCluster, controller.Result{}, nil
+}
+
+func setupEnvVars(ctx context.Context, cli client.Client) error {
```
vivek-koppuru, line 131: Unfortunately this won't work for the controller as there is no way to create files in the filesystem. I had to solve this differently by using a `ClientBuilder` which is different for the controller and loads the creds from the secret.

https://github.com/aws/eks-anywhere/blob/main/pkg/providers/snow/reconciler/clientbuilder.go

https://github.com/aws/eks-anywhere/blob/main/pkg/providers/snow/validator.go#L29

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+
+	if err := os.Setenv(snow.EksaSnowCredentialsFileKey, string(credentials)); err != nil {
+		return fmt.Errorf("failed setting env %s: %v", snow.EksaSnowCredentialsFileKey, err)
+	}
+
+	if err := os.Setenv(snow.EksaSnowCABundlesFileKey, string(caBundle)); err != nil {
+		return fmt.Errorf("failed setting env %s: %v", snow.EksaSnowCABundlesFileKey, err)
+	}
+
+	if _, ok := os.LookupEnv(eksaLicense); !ok {
```
vivek-koppuru, line 145: We can probably pull this from the secret in `eksa-system` though, if it exists

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -2,19 +2,151 @@ package reconciler
...
+		return nil, controller.Result{Result: &ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: defaultRequeueTime,
+		}}, err
+	}
+	return capiCluster, controller.Result{}, nil
+}
+
+func setupEnvVars(ctx context.Context, cli client.Client) error {
+	credentials, caBundle, err := getSnowCredentials(ctx, cli)
```
mitalipaygude, line 132:
```suggestion
	if credentials, caBundle, err := getSnowCredentials(ctx, cli); err != nil {
```

To keep it consistent with the code below

PR #2951, opened 2022-08-16
Change secret creation failure messaging

a-cool-train:
*Description of changes:*
Remove noisy messaging during secret creation failure and cron job failure.

*Testing (if applicable):*

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -69,11 +69,11 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error 
 	}
 
 	if err = pc.ApplySecret(ctx); err != nil {
-		logger.Info("warning: no aws key/license provided, package controller will be installed but curated packages might not be installed")
+		logger.Info("Warning: No AWS key/license provided. Please be aware this will prevent the package controller from installing curated packages.")
 	}
 
 	if err = pc.CreateCronJob(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -69,11 +69,11 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
 	}
 
 	if err = pc.ApplySecret(ctx); err != nil {
-		logger.Info("Warning: not able to create secret. Package installation might fail.", "error", err)
+		logger.Info("warning: no aws key/license provided, package controller will be installed but curated packages might not be installed")
```
jaxesn, line 72: Lets run this by @binyamz for final wording. I dont have a strong opinion, just want it to be something that doesnt look so error-y. I also dont mind if this shows up under the default log output, could be a good way to call out the existence of curated packages, I just don't want people to mistake it for an actual error.

TerryHowe (commented):
/lgtm

pkg/curatedpackages/packagecontrollerclient.go
binyamz: Can keep it consistent with the current iam authenticator messaging. Recommend to change it to "Warning: No AWS key/license provided. Please be aware this will prevent the package controller from installing curated packages."

TerryHowe:
presubmit still busted

TerryHowe (approved):
/lgtm
/approve

PR #2959, opened 2022-08-16
add options to override machine wait timeouts

gwesterfieldjr:
add options to override machine wait timeouts


cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -114,10 +133,32 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
 		return err
 	}
 
+	_, err = time.ParseDuration(uc.cpWaitTimeout)
+	if err != nil {
```
abhinavmpandey08, line 137: you can combine these two in a single line

```suggestion
	if _, err := time.ParseDuration(uc.cpWaitTimeout); err != nil {
```

cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -114,10 +133,32 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
...
+
+	_, err = time.ParseDuration(uc.externalEtcdWaitTimeout)
+	if err != nil {
+		logger.V(0).Info(fmt.Sprintf("Warning: Invalid %s value: %s, Using the default timeout instead: %s",
+			externalEtcdWaitTimeoutFlag, uc.externalEtcdWaitTimeout, clustermanager.DefaultEtcdWaitStr))
+		uc.externalEtcdWaitTimeout = clustermanager.DefaultEtcdWaitStr
+	}
+
+	var perMachineWaitTimeout time.Duration
+	perMachineWaitTimeout, err = time.ParseDuration(uc.perMachineWaitTimeout)
```
abhinavmpandey08, line 151: nit: you don't need to declare the var separately
```suggestion
	perMachineWaitTimeout, err := time.ParseDuration(uc.perMachineWaitTimeout)
```

cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -114,10 +133,32 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
...
+	_, err = time.ParseDuration(uc.cpWaitTimeout)
+	if err != nil {
+		logger.V(0).Info(fmt.Sprintf("Warning: Invalid %s value: %s, Using the default timeout instead: %s",
+			cpWaitTimeoutFlag, uc.cpWaitTimeout, clustermanager.DefaultControlPlaneWaitStr))
+		uc.cpWaitTimeout = clustermanager.DefaultControlPlaneWaitStr
+	}
+
+	_, err = time.ParseDuration(uc.externalEtcdWaitTimeout)
+	if err != nil {
+		logger.V(0).Info(fmt.Sprintf("Warning: Invalid %s value: %s, Using the default timeout instead: %s",
```
abhinavmpandey08, line 145: Since this message `Warning: Invalid %s value: %s, Using the default timeout instead: %s` is repeated multiple times, can you define it as a const and just the const variable instead?

Something like

```go
# This can go at the top or in the constants file
const invalidTimeMessage = "Warning: Invalid %s value: %s, Using the default timeout instead: %s"

logger.V(0).Info(fmt.Sprintf(invalidTimeMessage, ...)
```

test/e2e/tinkerbell_test.go
```diff
@@ -34,7 +51,7 @@ func TestTinkerbellKubernetes122SkipPowerActions(t *testing.T) {
 	test.PowerOffHardware()
 	test.PXEBootHardware()
 	test.PowerOnHardware()
-	test.CreateCluster(framework.WithForce())
+	test.CreateCluster(framework.WithForce(), framework.WithControlPlaneWaitTimeout("15m"))
```
abhinavmpandey08, line 54: Is this 15m for just a single control plane or for all control planes?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -143,10 +168,32 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		return err
 	}
 
+	_, err = time.ParseDuration(cc.cpWaitTimeout)
+	if err != nil {
```
abhinavmpandey08, line 172: nit 
```suggestion
	if _, err := time.ParseDuration(cc.cpWaitTimeout); err != nil {
```

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -143,10 +168,32 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
+	_, err = time.ParseDuration(cc.externalEtcdWaitTimeout)
+	if err != nil {
+		logger.V(0).Info(fmt.Sprintf("Warning: Invalid %s value: %s, Using the default timeout instead: %s",
+			externalEtcdWaitTimeoutFlag, cc.externalEtcdWaitTimeout, clustermanager.DefaultEtcdWaitStr))
+		cc.externalEtcdWaitTimeout = clustermanager.DefaultEtcdWaitStr
+	}
+
+	var perMachineWaitTimeout time.Duration
+	perMachineWaitTimeout, err = time.ParseDuration(cc.perMachineWaitTimeout)
+	if err != nil {
```
abhinavmpandey08, line 187: nit
```suggestion
	perMachineWaitTimeout, err := time.ParseDuration(cc.perMachineWaitTimeout)
	if err != nil {
```

abhinavmpandey08 (commented):
Looks good! just some minor nits and suggestions. Also, the lint job seems to be failing so you might have to look into that

test/e2e/tinkerbell_test.go
gwesterfieldjr: Just the initial control plane node. In case of replicas they use the per machine timeout

abhinavmpandey08 (approved):
Looks good!

pkg/clustermanager/cluster_manager.go
```diff
@@ -34,29 +33,35 @@ import (
...
+	writer                  filewriter.FileWriter
+	networking              Networking
+	diagnosticsFactory      diagnostics.DiagnosticBundleFactory
+	Retrier                 *retrier.Retrier
+	machineMaxWait          time.Duration
+	machineBackoff          time.Duration
+	machinesMinWait         time.Duration
+	awsIamAuth              AwsIamAuth
+	controlPlaneWaitTimeout string
+	externalEtcdWaitTimeout string
```
g-gaston, line 64: Could we store these two as `time.Duration` as well?

pkg/clustermanager/cluster_manager.go
```diff
@@ -118,20 +123,22 @@ type AwsIamAuth interface {
 
 type ClusterManagerOpt func(*ClusterManager)
 
-func New(clusterClient ClusterClient, networking Networking, writer filewriter.FileWriter, diagnosticBundleFactory diagnostics.DiagnosticBundleFactory, awsIamAuth AwsIamAuth, maxWaitPerMachine time.Duration, opts ...ClusterManagerOpt) *ClusterManager {
-	retrier := retrier.NewWithMaxRetries(maxRetries, backOffPeriod)
+func New(clusterClient ClusterClient, networking Networking, writer filewriter.FileWriter, diagnosticBundleFactory diagnostics.DiagnosticBundleFactory, awsIamAuth AwsIamAuth, opts ...ClusterManagerOpt) *ClusterManager {
+	retrier := retrier.NewWithMaxRetries(maxRetries, DefaultBackOffPeriod)
```
g-gaston, line 127: This is changing the behavior, right? I thought we were just adding the ability to configure some timeouts that were hardcoded before

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -64,6 +68,21 @@ func init() {
...
+
+	createClusterCmd.Flags().StringVar(&cc.timeoutOptions.externalEtcdWaitTimeout, externalEtcdWaitTimeoutFlag, clustermanager.DefaultEtcdWaitStr, "Override the default external etcd wait timeout (60m)")
+	if err := createClusterCmd.Flags().MarkHidden(externalEtcdWaitTimeoutFlag); err != nil {
+		logger.V(5).Info("Warn: Failed to mark flag as hidden: " + externalEtcdWaitTimeoutFlag)
+	}
+
+	createClusterCmd.Flags().StringVar(&cc.timeoutOptions.perMachineWaitTimeout, perMachineWaitTimeoutFlag, clustermanager.DefaultPerMachineWaitStr, "Override the default machine wait timeout (10m)/per machine ")
+	if err := createClusterCmd.Flags().MarkHidden(perMachineWaitTimeoutFlag); err != nil {
+		logger.V(5).Info("Warn: Failed to mark flag as hidden: " + perMachineWaitTimeoutFlag)
+	}
```
g-gaston, line 84: These are duplicated in the upgrade command, could we move them to a common place?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -143,10 +162,30 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
 	}
 
+	if _, err = time.ParseDuration(cc.timeoutOptions.cpWaitTimeout); err != nil {
+		logger.V(0).Info(fmt.Sprintf(timeoutWarningTemplate,
+			cpWaitTimeoutFlag, cc.timeoutOptions.cpWaitTimeout, clustermanager.DefaultControlPlaneWaitStr))
+		cc.timeoutOptions.cpWaitTimeout = clustermanager.DefaultControlPlaneWaitStr
+	}
+
+	if _, err = time.ParseDuration(cc.timeoutOptions.externalEtcdWaitTimeout); err != nil {
+		logger.V(0).Info(fmt.Sprintf(timeoutWarningTemplate,
```
g-gaston, line 172: Why wouldn't this just return an error if the value is invalid?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -143,10 +162,30 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
+			externalEtcdWaitTimeoutFlag, cc.timeoutOptions.externalEtcdWaitTimeout, clustermanager.DefaultEtcdWaitStr))
+		cc.timeoutOptions.externalEtcdWaitTimeout = clustermanager.DefaultEtcdWaitStr
+	}
+
+	perMachineWaitTimeout, err := time.ParseDuration(cc.timeoutOptions.perMachineWaitTimeout)
+	if err != nil {
+		logger.V(0).Info(fmt.Sprintf(timeoutWarningTemplate,
+			perMachineWaitTimeoutFlag, cc.timeoutOptions.perMachineWaitTimeout, clustermanager.DefaultPerMachineWaitStr))
+		perMachineWaitTimeout = clustermanager.DefaultMaxWaitPerMachine
+	}
```
g-gaston, line 182: Can we move this parsing logic somewhere else and share it between commands? It could even be a method in `timeoutOptions`

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -143,10 +162,30 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
...
+
 	deps, err := dependencies.ForSpec(ctx, clusterSpec).WithExecutableMountDirs(dirs...).
 		WithBootstrapper().
 		WithCliConfig(cliConfig).
-		WithClusterManager(clusterSpec.Cluster).
+		WithClusterManager(clusterSpec.Cluster, clustermanager.WithControlPlaneWaitTimeout(cc.timeoutOptions.cpWaitTimeout), clustermanager.WithExternalEtcdWaitTimeout(cc.timeoutOptions.externalEtcdWaitTimeout),
```
g-gaston, line 187: Maybe have some method that generates cluster manager options for a given `timeoutOptions`?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -234,9 +273,7 @@ func (cc *createClusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cl
 }
 
 func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
-	cliConfig := &config.CliConfig{
-		MaxWaitPerMachine: config.GetMaxWaitPerMachine(),
-	}
+	cliConfig := &config.CliConfig{}
```
g-gaston, line 276: Looking at this, should those timeout options go in this struct?

Def not blocking: I feel like we probably need a general agreement on how to handle flags/configuration, from reading to passing them down to other entities. That conversation is not for this PR, but we should have it.

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
```diff
@@ -5,7 +5,7 @@ env:
...
+    SKIPPED_TESTS: "TestTinkerbellKubernetes122SkipPowerActions,TestTinkerbellKubernetes120AWSIamAuth,TestTinkerbellKubernetes120SimpleFlow,TestTinkerbellKubernetes121ExternalEtcdSimpleFlow,TestVSphereKubernetes122BottlerocketAutoimport,TestCloudStackKubernetes120OIDC,TestCloudStackKubernetes121OIDC,TestCloudStackKubernetes121RedhatRegistryMirrorAndCert,TestSnowKubernetes121SimpleFlow,TestSnowKubernetes122SimpleFlow,TestSnowKubernetes123SimpleFlow,TestSnowKubernetes121OIDC,TestSnowKubernetes121UbuntuProxyConfig,TestSnowKubernetes123UbuntuRemoveWorkerNodeGroups,TestSnowKubernetes123LabelsUbuntu,TestSnowKubernetes123TaintsUbuntu,TestSnowKubernetes121To122UbuntuUpgrade,TestSnowKubernetes122To123UbuntuMultipleFieldsUpgrade"
```
g-gaston, line 8: Is this related to this PR?

pkg/config/config.go
```diff
@@ -1,55 +1,17 @@
...
-	MaxWaitPerMachine   time.Duration
-}
-
-const (
-	DefaultMaxWaitPerMachine = 10 * time.Minute
-	DefaultEtcdWaitStr       = "60m"
-)
-
-func GetMaxWaitPerMachine() time.Duration {
-	if env, found := os.LookupEnv(EksaReplicasReadyTimeoutEnv); found {
```
g-gaston, line 36: We should probably let the team know we are removing support for these env vars so they start using the flags if needed.

test/e2e/upgrade_test.go
```diff
@@ -499,13 +499,17 @@ func TestVSphereKubernetes121UbuntuTo122UpgradeWithCheckpoint(t *testing.T) {
 		framework.WithClusterFiller(api.WithWorkerNodeCount(1)),
 	)
 	clusterOpts = append(clusterOpts, framework.WithClusterUpgrade(api.WithKubernetesVersion(v1alpha1.Kube122)),
-		provider.WithProviderUpgrade(framework.UpdateUbuntuTemplate122Var(), api.WithResourcePoolForAllMachines(vsphereInvalidResourcePoolUpdateVar)), framework.WithEnvVar(features.CheckpointEnabledEnvVar, "true"), framework.WithEnvVar(config.ExternalEtcdTimeoutEnv, "10m"), framework.WithEnvVar(framework.CleanupVmsVar, "false"))
+		provider.WithProviderUpgrade(framework.UpdateUbuntuTemplate122Var(), api.WithResourcePoolForAllMachines(vsphereInvalidResourcePoolUpdateVar)), framework.WithEnvVar(features.CheckpointEnabledEnvVar, "true"), framework.WithEnvVar(framework.CleanupVmsVar, "false"))
+
+	commandOpts = append(commandOpts, framework.WithExternalEtcdWaitTimeout("10m"))
```
g-gaston, line 504:
```suggestion
	commandOpts := []framework.CommandOpt{framework.WithExternalEtcdWaitTimeout("10m"))}
```

pkg/clustermanager/cluster_manager.go
gwesterfieldjr: I made them strings because they are all actually consumed by cluster manager as a string. The string form of these are what gets passed to kubectl wait so I left them as strings since that's what the library currently uses them as.

pkg/clustermanager/cluster_manager.go
gwesterfieldjr: No I just renamed the constant on this one.

cmd/eksctl-anywhere/cmd/createcluster.go
gwesterfieldjr: Yeah that bothered me as well.. We actually have this problem sprinkled through all the commands with similar args. We duplicate them in each command.  Let me give it some thought on how to use shared options.

cmd/eksctl-anywhere/cmd/createcluster.go
gwesterfieldjr: So the previous env vars functioned this way in that they did not error out. Instead they just defaulted to the default values so I kept the same functionality.

cmd/eksctl-anywhere/cmd/createcluster.go
gwesterfieldjr: Yes will do

cmd/eksctl-anywhere/cmd/createcluster.go
gwesterfieldjr: Yeah I wrestled with idea of just providing a WithTimeoutOptions arg so they can all be passed to it. I will make this change.

cmd/eksctl-anywhere/cmd/createcluster.go
gwesterfieldjr: Agreed we need to have a common way to handle flag config as right now we do not. So I'm completely up to talking it through. 

I decided against using CliConfig because the timeout are only used by ClusterManager. They are not used anywhere else and I did not care for the approach of passing all args to the factory in CliConfig and then having the factory responsible for parsing out the logs to ClusterManager. I thought it would be better to just pass them directly to cluster manager. But again I am open to doing it differently. Whatever the best practice is I want to follow it.

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
gwesterfieldjr: Sort of.. This PR started as an effort to add more resiliency to the tinkerbell e2e tests which ultimately led to wanting ability to override timeouts so when tink test fails we don't have to wait whole 60m. 

This is a test that is flaky and I don't have a great option to fix it since to do so I would need to reformat the disk on each test run. Therefore, I decided to skip it for now so it stops polluting the build with failures until I can come up with a way to resolve the intermittency associated with it.

gwesterfieldjr: I can submit in different PR if you prefer.

pkg/config/config.go
gwesterfieldjr: Yes I did communicate in CLI standup that I would be moving timeout env vars to cli options. I will re-communicate it to the team.

pkg/clustermanager/cluster_manager.go
g-gaston: Ahhh I see, yeah. However I would vote to keep them all as duration for api consistency. I think converting from duration to string is easy, right?

cmd/eksctl-anywhere/cmd/createcluster.go
g-gaston: Sounds good! Let's take this offline and involve folks who are interested in the topic :)

cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
g-gaston: Non blocking, Up to you :)
Different PRs is always better, but I'm ok with this, as long there was a reason for it and there is

pkg/clustermanager/cluster_manager.go
```diff
@@ -34,29 +33,35 @@ import (
...
+	defaultMaxWaitPerMachine   = 10 * time.Minute
+	clusterWaitStr             = "60m"
+	DefaultControlPlaneWaitStr = "60m"
+	deploymentWaitStr          = "30m"
+	controlPlaneInProgressStr  = "1m"
+	etcdInProgressStr          = "1m"
+	DefaultEtcdWaitStr         = "60m"
 )
 
+var DefaultPerMachineWaitStr = defaultMaxWaitPerMachine.String()
```
g-gaston, line 50: I would advice against making a global var public
Can we just export `defaultMaxWaitPerMachine` and make the importers call `.String()` as needed?

cmd/eksctl-anywhere/cmd/createcluster.go
g-gaston: What about this: a method where you can pass a cobra command and it sets the timeout flags

like `setupTimeoutHiddenFlags(command *cobra.Command)`

It's not the most elegant solution but it solved the duplication and it's quite easy to change/undo in the future.

pkg/clustermanager/cluster_manager.go
gwesterfieldjr: Yes will do

pkg/clustermanager/cluster_manager.go
gwesterfieldjr: Yes easy peasy I'll make them the same

abhinavmpandey08 (approved):
/approve

PR #2979, opened 2022-08-17
Fix validation on osFamily for Tinkerbell

vignesh-goutham:

### DIFF
--- a/pkg/providers/tinkerbell/template.go
+++ b/pkg/providers/tinkerbell/template.go
@@ -63,6 +63,9 @@ func (tb *TemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *cluster.Spe
 				return nil, fmt.Errorf("getting control plane disk type of the hardware selector: %v", err)
 			}
 		}
+		if tb.controlPlaneMachineSpec.OSFamily != v1alpha1.Bottlerocket && tb.datacenterSpec.OSImageURL == "" {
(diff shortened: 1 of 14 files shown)


pkg/providers/tinkerbell/template.go
```diff
@@ -63,6 +63,9 @@ func (tb *TemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *cluster.Spe
 				return nil, fmt.Errorf("getting control plane disk type of the hardware selector: %v", err)
 			}
 		}
+		if tb.controlPlaneMachineSpec.OSFamily != v1alpha1.Bottlerocket && tb.datacenterSpec.OSImageURL == "" {
+			return nil, fmt.Errorf("failed generating Control Plane TinkerbellTemplateConfig: cannot import image for osFamily: %s, please use bottlerocket as osFamily for auto-importing or provide a valid osImageURL", tb.controlPlaneMachineSpec.OSFamily)
+		}
```
abhinavmpandey08, line 68: The CAPI spec is generated after the bootstrap cluster is created so this will fail after the kind cluster is created and CAPI is installed on it.
Can we add it as a preflight and fail before that?
Maybe somewhere here https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/validate.go#L32

pokearu: We would need to add a preflight and also register the assertion here I think
https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/cluster.go#L98

pokearu (approved):

abhinavmpandey08 (approved):
/approve
/lgtm

PR #2980, opened 2022-08-17
Add CNI reconciler

g-gaston:
*Description of changes:*
It calls the different CNI reconciler implementations, depending on the
CNI configured for the cluster. It only supports Cilium for now.

Also added logic to build it in factories and inject it in the snow
cluster reconciler.

*Testing (if applicable):*
Unit tests


pkg/networking/reconciler/reconciler.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Reconcile takes the specified CNI in a cluster to the desired state defined in a cluster Spec
+// It uses a controller.Result to indicate when requeues are needed
+// Intended to be used in a kubernetes controller
+// Only Cilium CNI is supported for now
+func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client client.Client, spec *cluster.Spec) (controller.Result, error) {
+	if spec.Cluster.Spec.ClusterNetwork.CNIConfig.Cilium != nil {
+		return r.ciliumReconciler.Reconcile(ctx, logger, client, spec)
+	} else {
+		return controller.Result{}, errors.New("unsupported CNI, only Cilium is supported at this time")
```
d8660091, line 36: When we return this error, the controller will requeue the request to be processed again, should we return nil err here?

g-gaston: Great question. I'm on the fence, because this line is mostly for correctness, I'm not sure that this can happen in the real world, since the cluster won't pass the validations.

Maybe we can just write this error to the cluster status and return nil? Let's take this offline.

vivek-koppuru: The next time that it would get reconciled however is when there is a change to the cluster resource, or would the snow reconciler always invoke this reconciler anyways? Just want to clarify the trigger here

pkg/networking/reconciler/reconciler.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/controller"
+)
+
+type CiliumReconciler interface {
```
vivek-koppuru, line 14: Should this be called the `CiliumReconciler`? Let's say we have a new network addon that we change to, then we need to change this name here. Instead, shouldn't this be called a `CniReconciler` and the interface one level up the `NetworkReconciler`? That may not be the best naming still, but it seems like we don't have to call it cilium here

vivek-koppuru: I am also wondering if this shows whether we need these two networking reconcilers and only need one, where we call the cilium reconciler in the snow reconciler, but that's probably not the best because we want to separate responsibilities here. 

I think my concern is more the name of the interface is our reconcile signature may not look different anyways, and then we decide to split it up based on the names after we realize that the signature is going to be different.

pkg/networking/reconciler/reconciler.go
g-gaston: The snow cluster reconciler would get called when the cluster object changes. And that reonciler would call the CNI one.

pkg/networking/reconciler/reconciler.go
g-gaston: This interface is representing the Cilium reconciler, not the cni one. They just happen to have the same signature.

This first level cni reconciler is in charge of knowing which implementation of a reconciler needs to be called. We only support Cilium right now, but once we add docker support we will need to add kindnetd as well.

vivek-koppuru: Got it, so we will have `CiliumReconciler` and a `KindnetdReconciler` interface later which may have the same signature. I can't think of a reason that they would have a different signature which was why I was wondering if we wanted to generalize the interface more

controllers/factory.go
```diff
@@ -210,16 +213,34 @@ func (f *Factory) withVSphereClusterReconciler() *Factory {
...
 
 	return f
 }
+
+func (f *Factory) withCNIReconciler() *Factory {
+	f.dependencyFactory.WithCiliumTemplater()
+
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.cniReconciler != nil {
+			return nil
```
mitalipaygude, line 237: Does it always return nil?

pkg/networking/reconciler/reconciler.go
g-gaston: Yeah, they might have the same signature, that's fair. However, there is no benefit in enforcing that right now. And if we do and we are wrong, we would need to refactor a few things.

IMO it's probably safer to wait until we are sure we have an abstraction before actually trying to enforce one.

controllers/factory.go
g-gaston: Yes, that's correct.

vivek-koppuru (approved):

PR #2982, opened 2022-08-17
Resolving eks-a controller spec discrepancy

maxdrib:
*Issue #, if available:*

*Description of changes:*
in the e2e tests, we observed two issues:
1. Many errors in the EKS-A controller log like `cloudstackmachinetemplate not found`
2. Replacement of worker nodes because

After setting up the debugger, the issue was isolated to the moment the EKS-A controller was paused, and we were able to analyze that comparison between the CSDC generated from the CloudStackCluster and the CSDC in the cluster would not be apples-to-apples, and would be missing the ManagementAPIEndpoint attribute

*Testing (if applicable):*
Unit testing, and e2e testing behavior is improved. No more errors in EKS-A controller logs, and no more recycling of worker machines.

### DIFF
--- a/pkg/providers/cloudstack/cloudstack.go
+++ b/pkg/providers/cloudstack/cloudstack.go
@@ -576,10 +576,30 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
 	return true, nil
 }
 
+// AnyImmutableFieldChanged Used by EKS-A controller and CLI upgrade workflow to compare generated CSDC/CSMC's from
+// CAPC resources in fetcher.go with those already on the cluster when deciding whether or not to generate and apply
+// new CloudStackMachineTemplates
 func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig) bool {
-	if !oldCsdc.Spec.Equal(&newCsdc.Spec) {
+	fmt.Printf("Comparing old csdc spec (%s) and new csdc spec (%s)", oldCsdc.Spec, newCsdc.Spec)
+	if len(newCsdc.Spec.AvailabilityZones) != len(oldCsdc.Spec.AvailabilityZones) {
 		return true
 	}
+	oAzsMap := map[string]v1alpha1.CloudStackAvailabilityZone{}
+	for _, oAz := range oldCsdc.Spec.AvailabilityZones {
+		oAzsMap[oAz.Name] = oAz
+	}
+	for _, sAz := range newCsdc.Spec.AvailabilityZones {
+		oAz, found := oAzsMap[sAz.Name]
+		if !found || !(sAz.Zone.Equal(&oAz.Zone) &&
+			sAz.Name == oAz.Name &&
+			sAz.CredentialsRef == oAz.CredentialsRef &&
+			sAz.Account == oAz.Account &&
+			sAz.Domain == oAz.Domain) {
+
+			return true
+		}
+	}
+
 	if oldCsmc.Spec.Template != newCsmc.Spec.Template {
 		return true
 	}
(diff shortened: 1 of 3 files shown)


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -576,10 +576,29 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
 	return true, nil
 }
 
+// AnyImmutableFieldChanged Used by EKS-A controller and CLI upgrade workflow to compare generated CSDC/CSMC's from
+// CAPC resources in fetcher.go with those already on the cluster when deciding whether or not to generate and apply
+// new CloudStackMachineTemplates
 func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig) bool {
-	if !oldCsdc.Spec.Equal(&newCsdc.Spec) {
+	if len(newCsdc.Spec.AvailabilityZones) != len(oldCsdc.Spec.AvailabilityZones) {
```
wongni, line 583:
```suggestion
	if len(oldCsdc.Spec.AvailabilityZones) != len(newCsdc.Spec.AvailabilityZones) {
```
Let's be consistent on the order of `old` and `new`

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -576,10 +576,29 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
...
 		return true
 	}
+	oAzsMap := map[string]v1alpha1.CloudStackAvailabilityZone{}
+	for _, oAz := range oldCsdc.Spec.AvailabilityZones {
+		oAzsMap[oAz.Name] = oAz
+	}
+	for _, sAz := range newCsdc.Spec.AvailabilityZones {
```
wongni, line 590:
```suggestion
	for _, newAz := range newCsdc.Spec.AvailabilityZones {
```
If oAz means old availability zone then nAz or newAz may be more appropriate.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -576,10 +576,29 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
...
 
+// AnyImmutableFieldChanged Used by EKS-A controller and CLI upgrade workflow to compare generated CSDC/CSMC's from
+// CAPC resources in fetcher.go with those already on the cluster when deciding whether or not to generate and apply
+// new CloudStackMachineTemplates
 func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig) bool {
-	if !oldCsdc.Spec.Equal(&newCsdc.Spec) {
+	if len(newCsdc.Spec.AvailabilityZones) != len(oldCsdc.Spec.AvailabilityZones) {
 		return true
 	}
+	oAzsMap := map[string]v1alpha1.CloudStackAvailabilityZone{}
```
g-gaston, line 586: I thought we already had a method in the api package to make this comparison.

g-gaston (approved):

pkg/providers/cloudstack/cloudstack.go
maxdrib: Unfortunately, that was the whole problem. When we make this comparison in this method, we actually compare the existing CSDC with the one that was generated from a CloudStackCluster resource in fetcher.go. That generated CSDC is missing an attribute - ManagementApiEndpoint (because it’s not easily available in the CloudStackCluster), so we can’t do a full comparison with .Equal()

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -576,10 +576,29 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
...
 		return true
 	}
+	oldAzsMap := map[string]v1alpha1.CloudStackAvailabilityZone{}
+	for _, oldAz := range oldCsdc.Spec.AvailabilityZones {
+		oldAzsMap[oldAz.Name] = oldAz
+	}
+	for _, newAz := range newCsdc.Spec.AvailabilityZones {
+		oldAz, found := oldAzsMap[newAz.Name]
+		if !found || !(oldAz.Zone.Equal(&newAz.Zone) &&
+			oldAz.Name == newAz.Name &&
```
wanyufe, line 593: oldAz.Name and newAz.Name must be equal here, right? 

Also, I think a comment to explain the logic is easier to understand. 
Are we allowing the changes under the same name now?

PR #2988, opened 2022-08-17
Document Proxy Env Vars

jonathanmeier5:
*Issue #, if available:*
Follows https://github.com/aws/eks-anywhere/pull/2849

*Description of changes:*
This change documents support for using EKS-Anywhere behind a Proxy via environment variables.

Previously, only commands using a cluster config file could use a proxy.

Now, you can use a proxy for any command by setting the documented environment variables.

Additionally, I added documentation that to successfully use a proxy you will need to update your docker daemon's configuration as described in docker's own documentation.

*Testing (if applicable):*


chrisnegus (commented):
/lgtm

PR #3003, opened 2022-08-17
Fix list from cluster

a-cool-train:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/curatedpackages/bundle.go
+++ b/pkg/curatedpackages/bundle.go
@@ -90,7 +90,12 @@ func (b *BundleReader) getPackageBundle(ctx context.Context, activeBundle string
 }
 
 func (b *BundleReader) GetActiveController(ctx context.Context) (*packagesv1.PackageBundleController, error) {
-	params := []string{"get", "packageBundleController", "-o", "json", "--kubeconfig", b.kubeConfig, "--namespace", constants.EksaPackagesName, packagesv1.PackageBundleControllerName}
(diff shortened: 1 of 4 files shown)


pkg/curatedpackages/bundle.go
```diff
@@ -90,7 +90,12 @@ func (b *BundleReader) getPackageBundle(ctx context.Context, activeBundle string
 }
 
 func (b *BundleReader) GetActiveController(ctx context.Context) (*packagesv1.PackageBundleController, error) {
-	params := []string{"get", "packageBundleController", "-o", "json", "--kubeconfig", b.kubeConfig, "--namespace", constants.EksaPackagesName, packagesv1.PackageBundleControllerName}
+	params := []string{"get", "kubeadmcontrolplane", "--no-headers", "-o", "custom-columns=:metadata.name", "--kubeconfig", b.kubeConfig, "--namespace", constants.EksaSystemNamespace}
```
ewollesen, line 93: I don't see any tests around this. Do you know that it handles having multiple results? I suspect that the following call to get the package bundle controller will fail if the name is passed as 
```
name1
name2
```
Which is what I suspect would happen with multiple clusters.

Is there something I'm not seeing that is determining which of the kubeadm control planes that gets returned is the "active" one?

TerryHowe: This command will need a --cluster option and KUBECONFIG will point to the management cluster when we remotely manage, but yeh, this might be broken for a management cluster managing multiple clusters.

ewollesen: Doh, yes, of course, that makes perfect sense. Thanks!

PR #3014, opened 2022-08-18
Remove kubeadm apiVersion from kind config

g-gaston:
*Description of changes:*

When not specified, kind will use the right apiVersion for the desired
kubernetes version. However, when specified, the patch will only apply
if the kubernetes version requires the specified kubeadm version. This
made it a problem when supporting 1.20 and 1.23 and the same time, since
1.20 requires v1beta2 and 1.23 requires v1beta3. By removing this line,
we let kind select the right one dynamically. This only works if the
specified fields in patch are compatible with both kubeadm versions,
otherwise, one patch per apiVersion is required. For now, we only use
stable fields available in both versions, so this is the easiest
solution.

More info:
https://github.com/kubernetes-sigs/kind/issues/1839#issuecomment-1148968204
https://github.com/cert-manager/cert-manager/pull/5187

### DIFF
--- a/pkg/executables/config/kind.yaml
+++ b/pkg/executables/config/kind.yaml
@@ -7,7 +7,6 @@ networking:
 {{- end }}
 kubeadmConfigPatches:
   - |
-    apiVersion: kubeadm.k8s.io/v1beta2
     kind: ClusterConfiguration
     dns:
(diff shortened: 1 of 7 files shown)


jaxesn (approved):
Thanks!

PR #3040, opened 2022-08-18
Upload EKS-A tarballs before running tests

abhay-krishna:
Tests take a while to run, so we don't want the release to be blocked by them.


pokearu (approved):
/lgtm

Issue #1524, opened 2022-03-18
Multi-zone CloudStack E2E Test


maxdrib:
**What would you like to be added**:
We should introduce an E2E test for CloudStack using multiple zones. It will require some refactoring of the E2E test implementation for CloudStack so that multiple zones and networks can be passed in, and also require an environment where multiple zones are supported with a load balancer between them.

**Why is this needed**:
 This is large and complex feature we're offering to customers and should be tested end to end in an automated fashion.

PR #3049, opened 2022-08-19
Adding CloudStack e2e test for clusters spanning multiple Cloudstack zones

maxdrib:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/1524

*Description of changes:*
This PR adds an e2e test to create a cluster in a single Cloudstack zone, then update it to be spread across two Cloudstack availability zones. Each availability zone can have its own network, account, domain, managementApiEndpoint, and credentialsRef, but for now we are focusing on testing with separate networks in separate zones.

I believe this will require some additional changes in secrets manager, which I could use some assistance with.

I plan to follow up these changes with another PR which moves all the remaining tests to the new AvailabilityZone model

*Testing (if applicable):*
Executed test against CI Cloudstack env from laptop


internal/pkg/api/cloudstack.go
```diff
@@ -67,6 +68,33 @@ func WithCloudStackManagementServer(value string) CloudStackFiller {
...
 
+func WithCloudStackAz(az anywherev1.CloudStackAvailabilityZone) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.datacenterConfig.Spec.AvailabilityZones = append(config.datacenterConfig.Spec.AvailabilityZones, az)
+	}
+}
+
+func WithFirstCloudStackAz(az anywherev1.CloudStackAvailabilityZone) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.datacenterConfig.Spec.AvailabilityZones[0] = az
```
vivek-koppuru, line 79: Same here, but what is the purpose of this? Instead of making it first cloudstack az, can we pass in index and attempt to set it at that spot? Otherwise if it is just for modifying one of them, I would just reset the cloudstack azs and re-add them.

internal/pkg/api/cloudstack.go
```diff
@@ -67,6 +68,33 @@ func WithCloudStackManagementServer(value string) CloudStackFiller {
...
+	}
+}
+
+func WithoutCloudStackAz(az anywherev1.CloudStackAvailabilityZone) CloudStackFiller {
```
vivek-koppuru, line 83: Should this just pass in the name then?

vivek-koppuru: I am just wondering if this is needed vs having something like `RemoveCloudstackAzs` and allowing you to recreate them

internal/pkg/api/cloudstack.go
```diff
@@ -111,23 +139,27 @@ func WithCloudStackConfigNamespace(ns string) CloudStackFiller {
 func WithCloudStackSSHAuthorizedKey(value string) CloudStackFiller {
 	return func(config CloudStackConfig) {
 		for _, m := range config.machineConfigs {
-			if len(m.Spec.Users) == 0 {
+			if len(m.Spec.Users) == 0 || m.Spec.Users == nil {
```
vivek-koppuru, line 142: `len` already checks for nil

internal/pkg/api/cloudstack.go
```diff
@@ -147,6 +179,23 @@ func WithCloudStackStringFromEnvVar(envVar string, opt func(string) CloudStackFi
 	return opt(os.Getenv(envVar))
 }
 
+func WithCloudStackAzFromEnvVars(cloudstackAccountVar, cloudstackDomainVar, cloudstackZoneVar, cloudstackNetworkVar, cloudstackManagementServerVar string, opt func(zone anywherev1.CloudStackAvailabilityZone) CloudStackFiller) CloudStackFiller {
```
vivek-koppuru, line 182: Do we want to do any validations on these env vars?

internal/pkg/api/cloudstack_test.go
```diff
@@ -0,0 +1,118 @@
+package api
```
vivek-koppuru, line 1: Nice thanks for adding unit tests here!

internal/pkg/api/cloudstack.go
```diff
@@ -67,6 +68,33 @@ func WithCloudStackManagementServer(value string) CloudStackFiller {
 	}
 }
 
+func WithCloudStackAz(az anywherev1.CloudStackAvailabilityZone) CloudStackFiller {
+	return func(config CloudStackConfig) {
+		config.datacenterConfig.Spec.AvailabilityZones = append(config.datacenterConfig.Spec.AvailabilityZones, az)
```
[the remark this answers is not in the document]
maxdrib: it should be getting initialized when you parse the result of `eksctl anywhere generate clusterconfig`

internal/pkg/api/cloudstack.go
maxdrib: Refactored so that instead of removing one at a time, we instead remove all the AZ's, and then can add it back incrementally

vivek-koppuru (approved):

PR #3076, opened 2022-08-19
grouping common command functions

raymond-zhang00:
*Issue #, if available:*

*Description of changes:*
- Moving directoriesToMount and buildCliConfig from createcluster.go to options.go and updating references
- Creating bindAllFlags function in flags.go and updating references

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -96,7 +96,7 @@ func (dc *deleteClusterOptions) deleteCluster(ctx context.Context) error {
 	}
 
 	cliConfig := buildCliConfig(clusterSpec)
-	dirs, err := cc.directoriesToMount(clusterSpec, cliConfig)
+	dirs, err := dc.directoriesToMount(clusterSpec, cliConfig)
 	if err != nil {
 		return err
 	}
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -3,11 +3,8 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"log"
 
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
-	"github.com/spf13/viper"
 
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/dependencies"
@@ -32,7 +29,7 @@ var deleteClusterCmd = &cobra.Command{
 	Use:          "cluster (<cluster-name>|-f <config-file>)",
 	Short:        "Workload cluster",
 	Long:         "This command is used to delete workload clusters created by eksctl anywhere",
-	PreRunE:      preRunDeleteCluster,
+	PreRunE:      bindFlagsToViper,
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if err := dc.validate(cmd.Context(), args); err != nil {
@@ -43,16 +40,6 @@ var deleteClusterCmd = &cobra.Command{
 		}
 		return nil
 	},
-}
-
-func preRunDeleteCluster(cmd *cobra.Command, args []string) error {
-	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
-		err := viper.BindPFlag(flag.Name, flag)
-		if err != nil {
-			log.Fatalf("Error initializing flags: %v", err)
-		}
-	})
-	return nil
 }
 
 func init() {
--- a/cmd/eksctl-anywhere/cmd/flags.go
+++ b/cmd/eksctl-anywhere/cmd/flags.go
@@ -1,0 +1,18 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func bindFlagsToViper(cmd *cobra.Command, args []string) error {
+	var err error
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if err != nil {
+			return
+		}
+		err = viper.BindPFlag(flag.Name, flag)
+	})
+	return err
+}
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -2,14 +2,20 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"time"
 
+	"strings"
+
 	"github.com/spf13/cobra"
 
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/clustermanager"
+	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 	"github.com/aws/eks-anywhere/pkg/version"
 )
 
@@ -59,6 +65,7 @@ type clusterOptions struct {
 	fileName             string
 	bundlesOverride      string
 	managementKubeconfig string
+	installPackages      string
 }
 
 func (c clusterOptions) mountDirs() []string {
@@ -107,4 +114,47 @@ func markFlagHidden(cmd *cobra.Command, flagName string) {
 	if err := cmd.Flags().MarkHidden(flagName); err != nil {
 		logger.V(5).Info("Warning: Failed to mark flag as hidden: " + flagName)
 	}
+}
+func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
+	cliConfig := &config.CliConfig{}
+	if clusterSpec.FluxConfig != nil && clusterSpec.FluxConfig.Spec.Git != nil {
+		cliConfig.GitSshKeyPassphrase = os.Getenv(config.EksaGitPassphraseTokenEnv)
+		cliConfig.GitPrivateKeyFile = os.Getenv(config.EksaGitPrivateKeyTokenEnv)
+		cliConfig.GitKnownHostsFile = os.Getenv(config.EksaGitKnownHostsFileEnv)
+	}
+
+	return cliConfig
+}
+
+func (c *clusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig *config.CliConfig) ([]string, error) {
+	dirs := c.mountDirs()
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig != nil && fluxConfig.Spec.Git != nil {
+		dirs = append(dirs, filepath.Dir(cliConfig.GitPrivateKeyFile))
+		dirs = append(dirs, filepath.Dir(cliConfig.GitKnownHostsFile))
+		dirs = append(dirs, filepath.Dir(c.installPackages))
+	}
+
+	if clusterSpec.Config.Cluster.Spec.DatacenterRef.Kind == v1alpha1.CloudStackDatacenterKind {
+		if extraDirs, err := c.extraDirectoriesToMount(); err == nil {
+			dirs = append(dirs, extraDirs...)
+		}
+	}
+
+	return dirs, nil
+}
+
+func (c *clusterOptions) extraDirectoriesToMount() ([]string, error) {
+	dirs := []string{}
+	env, found := os.LookupEnv(decoder.EksaCloudStackHostPathToMount)
+	if found && len(env) > 0 {
+		mountDirs := strings.Split(env, ",")
+		for _, dir := range mountDirs {
+			if _, err := os.Stat(dir); err != nil {
+				return nil, fmt.Errorf("invalid host path to mount: %v", err)
+			}
+			dirs = append(dirs, dir)
+		}
+	}
+	return dirs, nil
 }
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -4,9 +4,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"time"
-
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -115,6 +114,7 @@ func markFlagHidden(cmd *cobra.Command, flagName string) {
 		logger.V(5).Info("Warning: Failed to mark flag as hidden: " + flagName)
 	}
 }
+
 func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
 	cliConfig := &config.CliConfig{}
 	if clusterSpec.FluxConfig != nil && clusterSpec.FluxConfig.Spec.Git != nil {
@@ -138,7 +138,7 @@ func (c *clusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig
 	}
 
 	if clusterSpec.Config.Cluster.Spec.DatacenterRef.Kind == v1alpha1.CloudStackDatacenterKind {
-		if extraDirs, err := c.extraDirectoriesToMount(); err == nil {
+		if extraDirs, err := c.cloudStackDirectoriesToMount(); err == nil {
 			dirs = append(dirs, extraDirs...)
 		}
 	}
@@ -146,7 +146,7 @@ func (c *clusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig
 	return dirs, nil
 }
 
-func (c *clusterOptions) extraDirectoriesToMount() ([]string, error) {
+func (c *clusterOptions) cloudStackDirectoriesToMount() ([]string, error) {
 	dirs := []string{}
 	env, found := os.LookupEnv(decoder.EksaCloudStackHostPathToMount)
 	if found && len(env) > 0 {
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -132,7 +132,9 @@ func (c *clusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig
 	if fluxConfig != nil && fluxConfig.Spec.Git != nil {
 		dirs = append(dirs, filepath.Dir(cliConfig.GitPrivateKeyFile))
 		dirs = append(dirs, filepath.Dir(cliConfig.GitKnownHostsFile))
-		dirs = append(dirs, filepath.Dir(c.installPackages))
+		if c.installPackages != "" {
+			dirs = append(dirs, filepath.Dir(c.installPackages))
+		}
 	}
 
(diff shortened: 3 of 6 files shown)


cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -2,16 +2,22 @@ package cmd
...
+	"github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 	"github.com/aws/eks-anywhere/pkg/version"
 )
 
 type clusterOptions struct {
 	fileName             string
 	bundlesOverride      string
 	managementKubeconfig string
+	installPackages      string
```
raymond-zhang00, line 20: One of my concerns is whether including the installPackages in clusterOptions makes sense or if this should be wrapped up in a separate structure. 

Having it in clusterOptions helps ensure directoriesToMount runs which is used by create, update, and delete cmds.

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	if _, ok := overriddenFlags["kubeconfig"]; !ok {
+		cc.Flags().StringVar(&clusterOpt.managementKubeconfig, "kubeconfig", "", "Management cluster kubeconfig file")
+	}
+	if _, ok := overriddenFlags["install-packages"]; !ok {
+		cc.Flags().StringVar(&clusterOpt.installPackages, "install-packages", "", "Location of curated packages configuration files to install to the cluster")
+	}
+}
+
+func bindAllFlags(cmd *cobra.Command, args []string) error {
```
raymond-zhang00, line 29: Moved the bindAllFlags to a common function between create, delete, upgrade. If defining them individually is intentional and we want to keep those separate I can revert out this section

cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -45,23 +42,12 @@ var deleteClusterCmd = &cobra.Command{
...
-		if err != nil {
-			log.Fatalf("Error initializing flags: %v", err)
-		}
-	})
-	return nil
-}
-
 func init() {
 	deleteCmd.AddCommand(deleteClusterCmd)
+	addClusterFlags(deleteClusterCmd, dc.clusterOptions, "filename", "install-packages")
```
taneyland, line 47: why do we override filename here? the next line sets that flag

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,39 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func addClusterFlags(cc *cobra.Command, clusterOpt clusterOptions, override ...string) {
```
taneyland, line 9: can we add `--force-cleanup` to this? it looks like all 3 commands use this flag

cmd/eksctl-anywhere/cmd/deletecluster.go
raymond-zhang00: I overrode it because the deleteCmd usage differs from create and upgrade cluster. StringVarP panics when I try and just overwrite an existing flag so I wasn't able to just leave it in

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: I wanted to group the addClusterFlags together based on fields defined by clusterOptions. I think functionally it would help to add this since all commands use it though. Would it make sense to rename the function and group them all together?

raymond-zhang00: Updated the func to add --force-cleanup and rebased to resolve a merge conflict.

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type clusterCmdOptions struct {
+	clusterOptions
+	forceClean bool
```
taneyland, line 11:
```suggestion
	forceCleanup bool
```

cmd/eksctl-anywhere/cmd/options.go
taneyland, line 20: should we actually just move the `forceCleanup` here too, instead of in `clusterCmdOptions`?

taneyland (commented):
Looks great!

cmd/eksctl-anywhere/cmd/options.go
taneyland: I think this makes total sense. Especially since it can be passed in as a flag to our commands, similar to every other field in this struct

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Thanks Taylor, updated forceCleanup field name and made it part of clusterOptions

cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -45,23 +41,11 @@ var deleteClusterCmd = &cobra.Command{
...
-		if err != nil {
-			log.Fatalf("Error initializing flags: %v", err)
-		}
-	})
-	return nil
-}
-
 func init() {
 	deleteCmd.AddCommand(deleteClusterCmd)
+	addClusterFlags(deleteClusterCmd, &dc.clusterOptions, "filename", "install-packages")
```
g-gaston, line 46: we are adding flags here that wasn't there before, like `install-packages`

cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -45,23 +41,11 @@ var deleteClusterCmd = &cobra.Command{
...
 	deleteClusterCmd.Flags().StringVarP(&dc.fileName, "filename", "f", "", "Filename that contains EKS-A cluster configuration, required if <cluster-name> is not provided")
```
g-gaston, line 47: this one is gonna be added twice

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func bindAllFlags(cmd *cobra.Command, args []string) error {
+	var err error
+
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if err != nil {
+			return
+		}
+		err = viper.BindPFlag(flag.Name, flag)
```
g-gaston, line 39: this would override err if a the first flag fails but the second succeeds

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -103,8 +116,48 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
...
+	if clusterSpec.Config.Cluster.Spec.DatacenterRef.Kind == v1alpha1.CloudStackDatacenterKind {
+		if extraDirs, err := c.extraDirectoriesToMount(); err == nil {
+			dirs = append(dirs, extraDirs...)
+		}
+	}
+
+	return dirs, nil
+}
+
+func (c *clusterOptions) extraDirectoriesToMount() ([]string, error) {
```
g-gaston, line 150:
```suggestion
func (c *clusterOptions) cloudStackDirectoriesToMount() ([]string, error) {
```

cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -44,23 +41,10 @@ var upgradeClusterCmd = &cobra.Command{
...
-}
-
 func init() {
 	upgradeCmd.AddCommand(upgradeClusterCmd)
-	upgradeClusterCmd.Flags().StringVarP(&uc.fileName, "filename", "f", "", "Filename that contains EKS-A cluster configuration")
+	addClusterFlags(upgradeClusterCmd, &uc.clusterOptions, "install-packages")
 	upgradeClusterCmd.Flags().StringVarP(&uc.wConfig, "w-config", "w", "", "Kubeconfig file to use when upgrading a workload cluster")
-	upgradeClusterCmd.Flags().BoolVar(&uc.forceClean, "force-cleanup", false, "Force deletion of previously created bootstrap cluster")
-	upgradeClusterCmd.Flags().StringVar(&uc.bundlesOverride, "bundles-override", "", "Override default Bundles manifest (not recommended)")
-	upgradeClusterCmd.Flags().StringVar(&uc.managementKubeconfig, "kubeconfig", "", "Management cluster kubeconfig file")
```
g-gaston, line 63: we are going to be adding things like `install-packages`

cmd/eksctl-anywhere/cmd/deletecluster.go
raymond-zhang00: I was using the last "filename" and "install-packages" argument as an override that will cause addClusterFlags to exclude the `install-packages` and `filename` flag so they can be added in deletecluster.go. 

I realize it doesn't seem very intuitive to list out the flags that should NOT be included, but I thought it'd be more concise than including all the ones that should. Any suggestions on getting this to make more sense?

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: I grabbed this logic from Chris's comment on the preflight POC: https://github.com/aws/eks-anywhere/pull/2858#discussion_r938733162

From my understanding this initializes err in scope of the bindAllFlags parent function and stops binding flags if that err value is not nil.

cmd/eksctl-anywhere/cmd/deletecluster.go
raymond-zhang00: added filename as an override to addClusterFlags and left it to be initialized by deletecluster.go since the usage text for the delete cmd is slightly different

cmd/eksctl-anywhere/cmd/upgradecluster.go
raymond-zhang00: added install-packages as an override

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,42 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func addClusterFlags(cc *cobra.Command, clusterOpt *clusterOptions, override ...string) {
```
chrisdoherty4, line 9: The fact there's an override (that should be called 'exclude') tells me this isn't common code. I'd rather see individual functions such as `applyBundlesOverrideFlag(...)` called in the various commands its required because its abundantly clear whats happening and makes that flag consistent across commands.

cmd/eksctl-anywhere/cmd/flags.go
chrisdoherty4: It won't override. After the first error is observed it returns on l.37 for all subsequent visits and finally returns the original error on l.41.

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	if _, ok := overriddenFlags["install-packages"]; !ok {
+		cc.Flags().StringVar(&clusterOpt.installPackages, "install-packages", "", "Location of curated packages configuration files to install to the cluster")
+	}
+	if _, ok := overriddenFlags["force-cleanup"]; !ok {
+		cc.Flags().BoolVar(&clusterOpt.forceCleanup, "force-cleanup", false, "Force deletion of previously created bootstrap cluster")
+	}
+}
+
+func bindAllFlags(cmd *cobra.Command, args []string) error {
```
chrisdoherty4, line 32: Can we call this `bindFlagsToViper` given we're using the Viper singleton. I.e. there's nothing in the signature that indicates we're doing anything with Viper as is.

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -55,10 +60,18 @@ func buildClusterManagerOpts(t timeoutOptions) ([]clustermanager.ClusterManagerO
 	}, nil
 }
 
+func markFlagHidden(cmd *cobra.Command, flagName string) {
+	if err := cmd.Flags().MarkHidden(flagName); err != nil {
+		logger.V(5).Info("Warning: Failed to mark flag as hidden: " + flagName)
```
chrisdoherty4, line 65: This seems wrong. Typically this will only error if the flag doesn't exist so warning on it strikes me as a significant issue, possibly a cause for a panic given the end-user can't do anything about it.

chrisdoherty4 (changes requested):
See inline comments.

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Updated the naming to `bindFlagsToViper`

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Thanks Chris, I removed the override arg and reduced addClusterFlags to only flags used on all 3 (create, upgrade, delete) commands. The `delete cluster` command does have a different usage statement for filename so I added an overrideFlagNameAndUsage in flags.go.

Let me know if that looks like a better way of handling the common flags, or if it may be better to just keep the flag initialization separate per command.

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: markFlagHidden was merged in a separate PR (https://github.com/aws/eks-anywhere/pull/2959/), think git diff picked it up since I moved my functions underneath it. Is that something that I should address in this PR? Or should I make sure that the diff doesn't regard this as a change?

cmd/eksctl-anywhere/cmd/flags.go
```diff
@@ -0,0 +1,38 @@
...
+	if f != nil {
+		f.Name = newName
+		f.Shorthand = shorthand
+		f.Usage = usage
+	}
+}
+
+func bindFlagsToViper(cmd *cobra.Command, args []string) error {
+	var err error
+
```
chrisdoherty4, line 30: Whitespace

cmd/eksctl-anywhere/cmd/options.go
chrisdoherty4: If it has nothing to do with your PR then I would recommend rebasing your changes on-top of main so we can see its got nothing to do with this.

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Decided to not group the flags in a separate function and not use this struct. Ended up not touching forceCleanup.

cmd/eksctl-anywhere/cmd/deletecluster.go
raymond-zhang00: Decided to not touch the flag grouping in this PR. I can make a separate PR as needed if it helps to group the flags.

cmd/eksctl-anywhere/cmd/deletecluster.go
raymond-zhang00: Backed out the flag grouping changes

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: Updated the naming of this func

cmd/eksctl-anywhere/cmd/upgradecluster.go
raymond-zhang00: backed out of flag grouping changes

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Decided to back out of the flag grouping changes. If needed, they can be handled in a separate PR

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: Rebased changes and backed out of flag changes. This should no longer show up as modified

cmd/eksctl-anywhere/cmd/flags.go
raymond-zhang00: Removed whitespace in bindFlagsToViper func

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -59,6 +64,7 @@ type clusterOptions struct {
 	fileName             string
 	bundlesOverride      string
 	managementKubeconfig string
+	installPackages      string
```
chrisdoherty4, line 67: This is going to be available to `delete cluster` commands - that seems wrong? Its also a little unclear why we need package installation on an abstract set of cluster options. This is only used in cluster creation so perhaps we leave it on `createClusterOptions` for now.

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -108,3 +114,49 @@ func markFlagHidden(cmd *cobra.Command, flagName string) {
...
+
+func (c *clusterOptions) directoriesToMount(clusterSpec *cluster.Spec, cliConfig *config.CliConfig) ([]string, error) {
+	dirs := c.mountDirs()
+	fluxConfig := clusterSpec.FluxConfig
+	if fluxConfig != nil && fluxConfig.Spec.Git != nil {
+		dirs = append(dirs, filepath.Dir(cliConfig.GitPrivateKeyFile))
+		dirs = append(dirs, filepath.Dir(cliConfig.GitKnownHostsFile))
+		if c.installPackages != "" {
+			dirs = append(dirs, filepath.Dir(c.installPackages))
+		}
```
chrisdoherty4, line 137: These lines have been added from the original - whats the motivation here? I'm also struggling some to understand why we conditionally append the base path of `installPackages` on flux configuration being present?

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: The delete and upgrade commands previously both called directoriesToMount using the createClusterOptions variable (cc) initialized in createcluster.go. I added this so that each command could use it's own options but I think you're right and I ca work this out of the shared function - looking into that.

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: I just added the if statement for installPackages. Previously delete and upgrade commands just called the createClusterOptions installPackages directory. I'll see if this can be split out and if it is still necessary for createCluster.

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: Moved this out of the shared clusterOptions and back into `createClusterOptions`

cmd/eksctl-anywhere/cmd/options.go
raymond-zhang00: Followed up with Abdul, you're right the installPackages mount can be moved out of the flux if statement. Updated the function to accept additional directories to mount, `create cluster` will pass c.installPackages as an additional directory.

chrisdoherty4 (approved):

PR #3100, opened 2022-08-22
Update Go version to 1.18

g-gaston:
*Description of changes:*

* Add doc for Go version update process
* Add entry to changelog about SHA-1 signed certs deprecation


.github/workflows/golangci-lint.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2
         with:
-          go-version: 1.17.8
+          go-version: 1.18.5
```
chrisdoherty4, line 16: Is there any particular reason we specify the patch version?

docs/developer/go-version.md
```diff
@@ -0,0 +1,10 @@
+# Go version
+
+## How to update the repo's Go Version
+
+1. Make sure the [image builder](https://github.com/aws/eks-anywhere-build-tooling/blob/main/build/lib/install_go_versions.sh#L37) supports the new Go minor version. If it doesn't, you will need to add it there first.
```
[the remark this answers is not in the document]
g-gaston: No, as long the image builder supports the same minor version (`1.18`), it works

.github/workflows/golangci-lint.yml
g-gaston: That's what they did last time: https://github.com/aws/eks-anywhere/pull/1841/files#diff-d9a956120ac84289d2650137f64bd8f0893331de05e44cc41899dd984c9e0d48
That said, I just followed the same process and documented it, I don't know if there is a reason or we were just following what was there.

I would like to think it's to ensure consistency and make sure we always use the same version across tests, but maybe not.

@a-cool-train @vivek-koppuru you guys know?

chrisdoherty4: If we don't strive for binary reproducibility I don't think this is necessary; leaving the patch version off will reduce maintenance burden.

vivek-koppuru: It was initially added by @danbudris here: https://github.com/aws/eks-anywhere/pull/1732

I am not opposed to excluding the patch version. I think that the changes don't change too much between the patch versions to affect this functionality.

g-gaston: Changed

release/pkg/test/testdata/main-bundle-release.yaml
```diff
@@ -323,7 +323,7 @@ spec:
         imageDigest: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
         name: eks-anywhere-cli-tools
         os: linux
-        uri: public.ecr.aws/release-container-registry/eks-anywhere-cli-tools:v0.11.0-eks-a-v0.0.0-dev-build.1
+        uri: public.ecr.aws/release-container-registry/eks-anywhere-cli-tools:v0.11.1-eks-a-v0.0.0-dev-build.1
```
g-gaston, line 326: none of these changes are due to the go update. These files have to change almost every time we update the `release` folder since they depend on whatever is defined in the build tooling repo

vivek-koppuru (approved):

docs/content/en/docs/reference/changelog.md
```diff
@@ -4,6 +4,11 @@ linkTitle: "What's New?"
 weight: 35
 ---
 
+## Unreleased
+
+### ⚠️ Breaking changes
+- Certificates signed with SHA-1 are not supported anymore for Registry Mirror. Users with a registry mirror and providing a custom CA cert will need to rotate the certificate served by the registry mirror endpoint before using the new EKS-A version. This is true for both new clusters (`create cluster` command) and existing clusters (`upgrade cluster` command).
```
jaxesn, line 10: Is this more accurate/clear?
```suggestion
- Certificates signed with SHA-1 are not supported anymore for Registry Mirror. Users with a registry mirror and providing a custom CA cert signed with SHA-1 will need to rotate the certificate served by the registry mirror endpoint before using the new EKS-A version. This is true for both new clusters (`create cluster` command) and existing clusters (`upgrade cluster` command).
```

release/pkg/test/testdata/main-bundle-release.yaml
jaxesn: @abhay-krishna we should think a better long term solution here

PR #3104, opened 2022-08-22
Adding retries for reading remote EKS-D CRD's/manifests

maxdrib:
*Issue #, if available:*

*Description of changes:*
While performing some scaling testing, I observed several occasions where EKS-A was unable to fetch EKS-D CRDs/manifests and never retried. This connection failure resulted in cluster creation failing. This felt like a defect as we work to make EKS-A more robust, so this PR adds retry logic to the reader calls.

Adding this retry logic does slow down the unit tests to account for the retries and waits.

This logic could also be added at the Reader level itself. However, that leads to a myriad of other issues, including unit tests failing in the cluster_manager because the gomock matcher no longer recognizes two cluster.Specs as equivalent because their nested Reader resources differ. 

*Testing (if applicable):*
Created cluster using debugger in goland and observed that the EKS-D resources were correctly retrieved.

### DIFF
--- a/pkg/eksd/installer.go
+++ b/pkg/eksd/installer.go
@@ -27,22 +27,21 @@ type Reader interface {
 
 type Installer struct {
 	client  EksdInstallerClient
-	// Exposing for unit testing purposes
-	Retrier *retrier.Retrier
+	retrier *retrier.Retrier
 	reader  Reader
 }
 
 func NewEksdInstaller(client EksdInstallerClient, reader Reader) *Installer {
 	return &Installer{
 		client:  client,
-		Retrier: retrier.NewWithMaxRetries(maxRetries, backOffPeriod),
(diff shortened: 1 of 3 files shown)


taneyland (approved):

pkg/eksd/installer.go
```diff
@@ -56,14 +62,25 @@ func (i *Installer) InstallEksdCRDs(ctx context.Context, clusterSpec *cluster.Sp
 	return nil
 }
 
+// SetRetrier Exposing retrier for unit testing purposes
```
g-gaston, line 65:
```suggestion
// SetRetrier allows to modify the internal retrier
// For unit testing purposes only. It is not thread safe
```

g-gaston (approved):

PR #3106, opened 2022-08-22
Fix Cilium reconciler flaky tests

g-gaston:
*Description of changes:*
Add util to create objects with a kube client and wait until the client
cache gets refreshed with them. This avoid race conditions where the
tested code gets executed before those objects are available, making it
seem like they don't exist.

*Testing (if applicable):*
Ran all the package tests in loop 200 times.

### DIFF
--- a/internal/test/envtest/client.go
+++ b/internal/test/envtest/client.go
@@ -1,0 +1,45 @@
+package envtest
+
+import (
+	"context"
+	"testing"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


internal/test/envtest/client.go
```diff
@@ -0,0 +1,48 @@
...
+		if err := client.Create(ctx, obj); err != nil {
+			t.Fatal(err)
+		}
+		obj.SetGroupVersionKind(objKind)
+	}
+
+	found := make([]bool, len(objs))
+	for i, obj := range objs {
+		if found[i] {
+			continue
```
d8660091, line 37: I am not sure how this block work, I guess you meant to use "while" here (for condition {}).

Instead of creating our own "CreateObjs" helper function. I would suggest us to consider a more "well-known" solution by the official kubebuilder doc:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go#L116-L123

vivek-koppuru (approved):

Issue #3091, opened 2022-08-21
Flux fails to upgrade if flux-system kustomization is suspended


echel0n:
**What happened**: 
attempted to perform a cluster upgrade but fails cause flux is unable to upgrade due to flux-system being suspended before flux bootstrap command executed

**What you expected to happen**:
flux-system kustomization to be unsuspended after flux-system secret is replaced and before flux bootstrap command is issued

**How to reproduce it (as minimally and precisely as possible)**:
attempt to upgrade a cluster that has a older version of flux then what the latest version is

**Anything else we need to know?**:
I was able to work around this by manually issuing the command `flux resume kustomization flux-system` before the flux bootstrap command was issued.

However, even after the cluster and flux upgrade finished successfully, flux version still reports the following old version
`Flux change diff        {"oldVersion ": "v0.29.4+e670ae4", "newVersion ": "v0.31.3+15d160e"}`

**Environment**:
- EKS Anywhere Release: 0.11.0
- EKS Distro Release: 1.22

echel0n:
The cluster was originally created with EKS Anywhere v0.10.1 and was Flux-enabled, Flux kustomization is suspended automatically by the eksctl upgrade command it appears prior to it replacing the flux-system secret, also I did notice that when it fails to upgrade the cluster it does not resume the Flux kustomization.

Here is what I get when I run `eksctl anywhere upgrade plan cluster -f eksa-cluster.yaml`, keep in mind I have already upgraded the cluster, it seems to be improperly reporting versions as well.
```
Warning: VSphereDatacenterConfig configured in insecure mode
Checking new release availability...
NAME                 CURRENT VERSION   NEXT VERSION
EKS-A                v0.10.1+c59d2cc   v0.11.0+fcdfc21
Flux                 v0.29.4+e670ae4   v0.31.3+15d160e
cert-manager         v1.7.2+d4fa7e0    v1.8.2+543ab1d
cluster-api          v1.1.3+d32ac08    v1.2.0+06ada8a
kubeadm              v1.1.3+973f8a2    v1.2.0+a15cbd5
vsphere              v1.1.1+628bf01    v1.1.1+8aff43e
kubeadm              v1.1.3+9094d93    v1.2.0+dbcb744
etcdadm-bootstrap    v1.0.3+bfb8d15    v1.0.5+238c366
etcdadm-controller   v1.0.1+842bc86    v1.0.4+33fd30d
```

Here is what I get when I run `flux version`
```
[...]

PR #3108, opened 2022-08-22
Fix upgrade failure when flux enabled

jiayiwang7:
*Issue #, if available:*

#3091 

In EKS-A v0.10.0 release, during upgrade we pause the eks-a and flux reconciliation before core components upgrade to reduce upgrade side affect. Related PR: #2761. During Flux version upgrade,`flux bootstrap` preflight will fail when flux is already suspended, which causes the entire upgrade command fails when GitOps is enabled.

*Description of changes:*

Instead of suspending the entire Flux Kustomization, we add annotations to let Flux ignore EKS-A resources during the upgrade process. For example, below annotation is added to EKS-A cluster, provider datacenter and provider machineconfigs before components upgrade:

```bash
kustomize.toolkit.fluxcd.io/reconcile: disabled
```

Reference: https://fluxcd.io/docs/components/kustomize/kustomization/#reconciliation

*Testing (if applicable):*

Unit tests. Also manually tested Flux enabled cluster upgrade from 0.10 to latest.

Will add e2e in a separate PR.


echel0n:
If flux-system kustomization is suspended, the flux bootstrap will fail, I tested this by running the bootstrap command standalone while flux-system was suspended.

g-gaston (approved):

pkg/gitops/flux/flux.go
```diff
@@ -149,6 +151,78 @@ func (f *Flux) Uninstall(ctx context.Context, cluster *types.Cluster, clusterSpe
 	return nil
 }
 
+func (f *Flux) PauseClusterResourcesReconcile(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec, provider providers.Provider) error {
+	if f.shouldSkipFlux() {
+		logger.Info("GitOps field not specified, pause cluster resources reconcile skipped")
```
g-gaston, line 156: nit: maybe do this for level 4?
It doesn't look like something we should print by default

pkg/gitops/flux/flux.go
```diff
@@ -149,6 +151,78 @@ func (f *Flux) Uninstall(ctx context.Context, cluster *types.Cluster, clusterSpe
...
+		return nil
+	}
+
+	c, err := f.fluxClient.GetCluster(ctx, cluster, clusterSpec)
```
g-gaston, line 160: can we just pass the currentSpec to this method? And avoid an extra get

pkg/gitops/flux/flux.go
```diff
@@ -149,6 +151,78 @@ func (f *Flux) Uninstall(ctx context.Context, cluster *types.Cluster, clusterSpe
...
+		return fmt.Errorf("disable resource %s %s from Flux reconcile: %v", provider.DatacenterResourceType(), clusterSpec.Cluster.Spec.DatacenterRef.Name, err)
+	}
+
+	if provider.MachineResourceType() != "" {
+		for _, machineConfigRef := range clusterSpec.Cluster.MachineConfigRefs() {
+			if err := f.fluxClient.DisableResourceReconcile(ctx, cluster, provider.MachineResourceType(), machineConfigRef.Name, clusterSpec.Cluster.Namespace); err != nil {
+				return fmt.Errorf("disable resource %s %s from Flux reconcile: %v", provider.MachineResourceType(), machineConfigRef.Name, err)
+			}
+		}
+	}
```
g-gaston, line 185: should we include here the rest of the objects, like OIDC, AWSIam, etc?

pkg/workflows/upgrade.go
```diff
@@ -327,34 +327,35 @@ func (s *upgradeNeeded) Restore(ctx context.Context, commandContext *task.Comman
 	return &createBootstrapClusterTask{}, nil
 }
 
-func (s *pauseEksaAndFluxReconcile) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func (s *pauseEksaAndGitOpsReconcile) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
```
g-gaston, line 330: should this be named `pauseEksaReconcile`?

pkg/workflows/interfaces/interfaces.go
```diff
@@ -45,6 +45,8 @@ type GitOpsManager interface {
 	InstallGitOps(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec, datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig) error
 	PauseGitOpsKustomization(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 	ResumeGitOpsKustomization(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
```
g-gaston, line 47: Can we remove some of these if we don't use them anymore? same for their implementations

pkg/gitops/flux/flux.go
jiayiwang7: yea actually we don't need this check

pkg/workflows/interfaces/interfaces.go
jiayiwang7: i mentioned in the description there will be separate PR to remove unused methods

pkg/gitops/flux/flux.go
jiayiwang7: i prefer to create issue to follow up, since it is not just in flux, but same for cluster_manager add pause annotation.

jiayiwang7: https://github.com/aws/eks-anywhere/issues/3127

PR #3116, opened 2022-08-23
Add Nutanix API types

thunderboltsid:
Add types for Nutanix resources in pkg/api/v1alpha1


pkg/api/v1alpha1/nutanixdatacenterconfig_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// +kubebuilder:validation:Minimum=9440
+	Port int `json:"port"`
+
+	// User is the User name for Nutanix Prism Central
+	// +kubebuilder:validation:Required
+	User string `json:"user"`
+
+	// Password is the Password for Nutanix Prism Central
+	// +kubebuilder:validation:Required
+	Password string `json:"password"`
```
abhinavmpandey08, line 30: Can we read username and password from environment variables and store them in a Kubernetes secret?
Ideally, we shouldn't expose credentials in the cluster config

thunderboltsid: We indeed plan on using secret but were using the config as an intermediary object; this is not really necessary so we can move it out.

abhay-krishna:
Can you add the generated CRD changes for these types?

abhay-krishna:
> Can you add the generated CRD changes for these types?

You can generate these by running the following commands
```
# This will generate two CRD bases at 
# config/crd/bases/anywhere.eks.amazonaws.com_nutanixdatacenterconfigs.yaml
# and config/crd/bases/anywhere.eks.amazonaws.com_nutanixmachineconfigs.yaml
make generate-manifests 

# Edit config/crd/kustomization.yaml to include these two CRDs in the resources section

# This will update the releases manifest at config/manifest/eksa-components.yaml
make release-manifests

File: pkg/api/v1alpha1/nutanix_types.go [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+	NutanixIdentifierUUID NutanixIdentifierType = "uuid"
+
+	// NutanixIdentifierName is a resource identifier identifying the object by Name.
+	NutanixIdentifierName NutanixIdentifierType = "name"
+)
+
+// NutanixResourceIdentifier holds the identity of a Nutanix Prism resource (cluster, image, subnet, etc.)
+//
+// +union
+type NutanixResourceIdentifier struct {
```
pokearu, line 17: Could we move the struct definition to the place of consumption? Since they are used inside `NutanixMachineConfig`.

pokearu:
We use kubebuilder for scaffolding APIs, so could we also add these types to the PROJECT file?
https://github.com/aws/eks-anywhere/blob/main/PROJECT

pkg/api/v1alpha1/nutanixdatacenterconfig_test.go
```diff
@@ -0,0 +1,140 @@
...
+			name:        "invalid-kind",
+			fileName:    "testdata/nutanix/invalid-kind.yaml",
+			expectedErr: "does not contain kind NutanixDatacenterConfig",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			conf, err := GetNutanixDatacenterConfig(test.fileName)
+			if test.expectedErr == "" {
```
pokearu, line 119: Love the use of table tests. 
But could we avoid `if else` clauses and break these into different unit tests? Table tests are best used for varying input for a single scenario.

pkg/api/v1alpha1/nutanixmachineconfig_test.go
```diff
@@ -0,0 +1,170 @@
...
+			fileName:    "testdata/nutanix/invalid-kind.yaml",
+			machineConf: nil,
+			expectedErr: "unable to find kind NutanixMachineConfig in file",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			conf, err := GetNutanixMachineConfigs(test.fileName)
+			if test.expectedErr == "" {
```
pokearu, line 145: Same as [above](https://github.com/aws/eks-anywhere/pull/3116/files#r955265868)

pkg/api/v1alpha1/nutanixdatacenterconfig_types.go
```diff
@@ -0,0 +1,140 @@
...
+	Endpoint string `json:"endpoint"`
+
+	// Port is the Port of Nutanix Prism Central
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=9440
+	Port int `json:"port"`
+
+	// AdditionalTrustBundle is the trust bundle for Prism Centrals using non-publicly trusted certificates
+	// +kubebuilder:validation:Required
+	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
```
pokearu, line 28: We have `Required` and `omitempty` for this field. Does that work together?

pkg/api/v1alpha1/nutanixmachineconfig.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+const (
+	NutanixMachineConfigKind        = "NutanixMachineConfig"
+	DefaultNutanixOSFamily          = Ubuntu
+	DefaultNutanixSystemDiskSizeGi  = "40Gi"
+	DefaultNutanixMemorySizeGi      = "4Gi"
+	DefaultNutanixVCPUsPerSocket    = 1
+	DefaultNutanixVCPUSockets       = 2
+	DefaultNutanixMachineConfigUser = "nutanix-user"
```
pokearu, line 20: Do we expect to use these defaults outside of the `NewNutanixMachineConfigGenerate` method?
If not we could make them non exported.

thunderboltsid:
> make release-manifests

This didn't generate any new changes.

pkg/api/v1alpha1/nutanixdatacenterconfig_types.go
thunderboltsid: Good catch! Removing required. This field is for providing a cert bundle that you would like to trust in case of self-signed certs.

pkg/api/v1alpha1/nutanixmachineconfig.go
thunderboltsid: Good catch; unexported them. Can always make it exported later if need arises.

pkg/api/v1alpha1/nutanixdatacenterconfig_test.go
thunderboltsid: Not sure I subscribe to that ideology but I'll break it up for the purpose of moving the PR forward.

PROJECT
```diff
@@ -101,4 +101,13 @@ resources:
   webhooks:
     validation: true
     webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: eks.amazonaws.com
+  group: anywhere
+  kind: NutanixConfig
```
pokearu, line 109: This should be the `Kind` of the types we are adding. So `NutanixDatacenterConfig` and `NutanixMachineConfig`.

abhinavmpandey08 (approved):
looks good!

pokearu (approved):

pokearu:
/approve
🚀

PR #3119, opened 2022-08-23
Preflight validate

raymond-zhang00:
*Issue #, if available:*

*Description of changes:*
Add experimental preflight validation command and associated command validation package.

*Testing (if applicable):*
Unit tests added to cover validation command workflow and added functions


cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+
+	tmpPath, err := os.MkdirTemp("./", "tmpValidate")
+	if err != nil {
+		return err
+	}
+	deps, err := dependencies.ForSpec(ctx, clusterSpec).WithExecutableMountDirs(dirs...).
+		WithWriterFolder(tmpPath).
```
taneyland, line 94: Can we just use `.WithWriter()` here? It has a built-in temp folder

pkg/validations/commandvalidations/commandvalidations.go
```diff
@@ -0,0 +1,82 @@
...
+	return err
+}
+
+func (v *ValidateCreateCluster) generateCreateValidations(ctx context.Context) []validations.Validation {
+	return []validations.Validation{
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name:   "validate docker executable",
+				Err:    ValidateDockerExecutable(ctx, v.dockerExec, runtime.GOOS),
+				Silent: true,
```
taneyland, line 45: Just curious: why are we setting this to true? We don't want to log when they succeed?

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: From what I found `WithWriter()` makes the temp folder under the `factory.writerFolder` location, which is set to the cluster name in the spec by the `dependencies.ForSpec` func.

I used the temp path to avoid accidentally cleaning up any existing directory named with the cluster name. Is there a way to avoid using the cluster name directory with `WithWriter()`?

pkg/validations/commandvalidations/commandvalidations.go
raymond-zhang00: I ended up marking the validations that are silent during our `create cluster` silent during the `validate` as well.  This was to address some feedback and help keep the output from being too verbose if these validations succeed.

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+type validateOptions struct {
+	clusterOptions
+	hardwareCSVPath       string
+	tinkerbellBootstrapIP string
+}
+
+var valOpt = &validateOptions{}
+
+var validateCreateCmd = &cobra.Command{
```
g-gaston, line 31: I think we need a another command?
Looking at this file `validatecreatecluster`, shouldn't the command be exp -> validate -> create -> cluster?

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+	Short:        "Validate create cluster",
+	Long:         "This command is used to validate the create cluster action",
+	PreRunE:      bindFlagsToViper,
+	SilenceUsage: true,
+	RunE:         valOpt.validateCreateCluster,
+}
+
+func init() {
+	validateCmd.AddCommand(validateCreateCmd)
+	validateCreateCmd.Flags().StringVarP(&valOpt.fileName, "filename", "f", "", "Filename that contains EKS-A cluster configuration")
```
g-gaston, line 42: were you going to put this in a method to get consistent flags?

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (valOpt *validateOptions) validateCreateCluster(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+
+	clusterSpec, err := cluster.NewSpecFromClusterConfig(valOpt.fileName, version.Get())
+	if err != nil {
+		return err
+	}
+
+	if clusterSpec.Config.Cluster.Spec.DatacenterRef.Kind == v1alpha1.TinkerbellDatacenterKind {
```
g-gaston, line 65: if this is copy paste from `create` I would vote to at least put it in a method

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+	dirs, err := valOpt.directoriesToMount(clusterSpec, cliConfig)
+	if err != nil {
+		return err
+	}
+
+	tmpPath, err := os.MkdirTemp("./", "tmpValidate")
+	if err != nil {
+		return err
+	}
+	deps, err := dependencies.ForSpec(ctx, clusterSpec).WithExecutableMountDirs(dirs...).
```
g-gaston, line 93:
```suggestion
	deps, err := dependencies.ForSpec(ctx, clusterSpec).
		WithExecutableMountDirs(dirs...).
```

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+		WithWriterFolder(tmpPath).
+		WithKubectl().
+		WithProvider(valOpt.fileName, clusterSpec.Cluster, false, valOpt.hardwareCSVPath, true, valOpt.tinkerbellBootstrapIP).
+		WithGitOpsFlux(clusterSpec.Cluster, clusterSpec.FluxConfig, cliConfig).
+		Build(ctx)
+	if err != nil {
+		deps.Writer.CleanUp()
```
g-gaston, line 100: If there is an error, it's not safe to rely on the first value.

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+		cluster = &types.Cluster{
+			Name:           clusterSpec.Cluster.Name,
+			KubeconfigFile: kubeconfig.FromClusterName(clusterSpec.Cluster.Name),
+		}
+	} else {
+		cluster = &types.Cluster{
+			Name:           clusterSpec.ManagementCluster.Name,
+			KubeconfigFile: clusterSpec.ManagementCluster.KubeconfigFile,
+		}
+	}
```
g-gaston, line 116: I think this can also be in a method if it's copy paste

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+			KubeconfigFile: kubeconfig.FromClusterName(clusterSpec.Cluster.Name),
+		},
+		ManagementCluster: cluster,
+		Provider:          deps.Provider,
+		CliConfig:         cliConfig,
+	}
+
+	createValidations := createvalidations.New(validationOpts)
+
+	docker := executables.BuildDockerExecutable()
```
g-gaston, line 132: why wouldn't we build this with the factory?

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
```diff
@@ -0,0 +1,140 @@
...
+
+	commandVal := commandvalidations.New(clusterSpec, deps, createValidations, docker)
+	err = commandVal.ValidateCreate(ctx)
+
+	cleanup(deps, &err)
+	deps.Writer.CleanUp()
```
g-gaston, line 138: doesn't `cleanup` do this?

pkg/validations/commandvalidations/commandvalidations.go
g-gaston: This makes me think that we might not want to use the current implementation of the validation runner then and just write something that fits our need better

pkg/validations/commandvalidations/commandvalidations.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type ValidateCreateCluster struct {
+	clusterSpec    *cluster.Spec
+	deps           *dependencies.Dependencies
+	cmdValidations interfaces.Validator
+	dockerExec     validations.DockerExecutable
+}
+
+func New(clusterSpec *cluster.Spec, deps *dependencies.Dependencies, cmdValidations interfaces.Validator, dockerExec validations.DockerExecutable) *ValidateCreateCluster {
```
g-gaston, line 21: Using New as the name for a constructor is only a good idea when the name of the package tells you already what that constructor is building. In this case `commandvalidations.New()` would be telling a lie

g-gaston, line 21: instead of taking `dependencies.Dependencies` I would just take whatever this objects needs
If it becomes a lot of arguments, then feel free to create an input struct just for that

g-gaston, line 21: I would vote against importing `interfaces.Validator`, that's under the workflows package and I think we should avoid that dependency direction

pkg/validations/commandvalidations/dockerexec.go
```diff
@@ -0,0 +1,24 @@
+package commandvalidations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/validations"
+)
+
+func ValidateDockerExecutable(ctx context.Context, docker validations.DockerExecutable, os string) error {
```
g-gaston, line 10: should this validations be in `pkg/docker`?
IMO validations fit better in packages that inform about their domain (in general)

That said, `pkg/docker` shouldn't probably depend on `pkg/validations` so maybe we need to change that too?

pkg/validations/commandvalidations/kubeconfig.go
```diff
@@ -0,0 +1,20 @@
+package commandvalidations
+
+import (
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/validations"
+)
+
+func ValidateKubeconfigPath(clusterName string) error {
```
g-gaston, line 10: same comment here about having this in `pkg/kubeconfig`

pkg/validations/commandvalidations/commandvalidations.go
```diff
@@ -0,0 +1,82 @@
+package commandvalidations
```
g-gaston, line 1: should this package be `cluster`? or `createcluster`? (one word is always better than two but not sure if `cluster` is too generic

that way we get `pkg/validations/createcluster`

pkg/validations/commandvalidations/commandvalidations.go
```diff
@@ -0,0 +1,82 @@
...
+			return &validations.ValidationResult{
+				Name:   "validate kubeconfig path",
+				Err:    ValidateKubeconfigPath(v.clusterSpec.Cluster.Name),
+				Silent: true,
+			}
+		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name:   "validate cluster",
+				Err:    v.clusterSpec.Cluster.Validate(),
```
g-gaston, line 58: should we validate the whole cluster config?

pkg/validations/commandvalidations/commandvalidations.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name: "create preflight validations pass",
+				Err:  v.cmdValidations.PreflightValidations(ctx),
+			}
+		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name: fmt.Sprintf("%s Provider setup is valid", v.deps.Provider.Name()),
```
g-gaston, line 77:
```suggestion
				Name: fmt.Sprintf("validate %s Provider", v.deps.Provider.Name()),
```

pkg/validations/commandvalidations/validateprovider.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/features"
+	"github.com/aws/eks-anywhere/pkg/providers"
+)
+
+func ValidateSupportedProvider(provider providers.Provider) error {
+	if !features.IsActive(features.CloudStackProvider()) && provider.Name() == constants.CloudStackProviderName {
```
g-gaston, line 12: just curious, where were we running this before?

g-gaston:
Great job! Glad to see the whole thing coming together 🚀

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: Yeah, I'll break that into a separate PR and rebase it in if that makes it in first.

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: Will do, I'll group this in with the cmd flags PR

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: I totally missed that in the dependency factory, that'd be much better to use. Changing

pkg/validations/commandvalidations/commandvalidations.go
raymond-zhang00: That sounds good

pkg/validations/commandvalidations/validateprovider.go
raymond-zhang00: Yes, this func just wraps up the checks we were doing individually in create cluster.

pkg/validations/commandvalidations/commandvalidations.go
raymond-zhang00: Good catch, using cluster.ValidateConfig instead

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: updated to use `exp validate create cluster`

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: updated to use cleanupDirectory() instead

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: Ended up removing both and just using cleanupDirectory()

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: https://github.com/aws/eks-anywhere/pull/3227

cmd/eksctl-anywhere/cmd/validatecreatecluster.go
raymond-zhang00: I placed this in cmd/options since that already had the cluster pkg. Another place to put it would be in cmd/createcluster.go since this is specific to `create cluster` and `validate create cluster`. Would that be better?

pkg/validations/commandvalidations/dockerexec.go
raymond-zhang00: moved to pkg/validations/docker.go - where all the other docker validations are for now

pkg/validations/commandvalidations/kubeconfig.go
raymond-zhang00: Moved to pkg/kubeconfig

pkg/validations/commandvalidations/validateprovider.go
g-gaston: Ahhh we do that in the cmd package 🤦🏻‍♂️
Ok, another thing that we should get our of there, eventually

cmd/eksctl-anywhere/cmd/cleanup.go
```diff
@@ -20,3 +21,10 @@ func close(ctx context.Context, closer types.Closer) {
 		logger.Error(err, "Closer failed", "closerType", fmt.Sprintf("%T", closer))
 	}
 }
+
+func cleanupDirectory(directory string) {
+	_, err := os.Stat(directory)
+	if err == nil {
```
g-gaston, line 27:
```suggestion
	if _, err := os.Stat(directory); err == nil {
```

pkg/kubeconfig/kubeconfig.go
```diff
@@ -37,3 +37,15 @@ func NewMissingFileError(path string) error {
 func (m missingFileError) Error() string {
 	return fmt.Sprintf("kubeconfig file missing: path=%v", m.Path)
 }
+
+func ValidateKubeconfigPath(clusterName string) error {
```
g-gaston, line 41: test?

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/validations"
+	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
+)
+
+type validationManager struct {
+	clusterSpec    *cluster.Spec
+	provider       providers.Provider
+	gitOpsFlux     *flux.Flux
+	cmdValidations interfaces.Validator
```
g-gaston, line 20: can we just define our own interface for this?

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,86 @@
...
+			return &validations.ValidationResult{
+				Name:   "validate supported provider",
+				Err:    ValidateSupportedProvider(v.provider),
+				Silent: true,
+			}
+		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name: "create preflight validations pass",
+				Err:  v.cmdValidations.PreflightValidations(ctx),
```
g-gaston, line 76: this is running all the validations in `PreflightValidations` as one validation here
Could we just register all those validations individually here?
We might not be able to do this for everything (specially providers) without a big refactor, but this one should be using the same types for validations, right?

pkg/validations/createcluster/createcluster.go
g-gaston, line 20: instead of `cmdValidations`, `createValidations`?

pkg/validations/createcluster/createcluster_test.go
```diff
@@ -0,0 +1,86 @@
...
+	c.flux = flux.NewFluxFromGitOpsFluxClient(nil, nil, nil, nil)
+}
+
+func (c *createClusterValidationTest) expectPreflightValidationsToPass() {
+	c.validator.EXPECT().PreflightValidations(c.ctx).Return(nil).AnyTimes()
+}
+
+func TestCreateClusterValidationsSuccess(t *testing.T) {
+	test := newValidateTest(t)
+	test.parseClusterSpecFromFile(t, "testdata/cluster_vsphere.yaml")
```
g-gaston, line 70: do we need to parse from file or can we just generate the cluster.Spec directly in go code?

pkg/validations/createcluster/validateprovider.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/features"
+	"github.com/aws/eks-anywhere/pkg/providers"
+)
+
+func ValidateSupportedProvider(provider providers.Provider) error {
```
g-gaston, line 11: should we move this to `pkg/provider`?

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,86 @@
...
+
+type validationManager struct {
+	clusterSpec    *cluster.Spec
+	provider       providers.Provider
+	gitOpsFlux     *flux.Flux
+	cmdValidations interfaces.Validator
+	dockerExec     validations.DockerExecutable
+}
+
+func NewValidationManager(clusterSpec *cluster.Spec, provider providers.Provider, gitOpsFlux *flux.Flux, cmdValidations interfaces.Validator, dockerExec validations.DockerExecutable) *validationManager {
```
g-gaston, line 24: maybe this can be called `New` now? lol
Or maybe `NewValidations`

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,86 @@
...
+	return &validationManager{
+		clusterSpec:    clusterSpec,
+		provider:       provider,
+		gitOpsFlux:     gitOpsFlux,
+		cmdValidations: cmdValidations,
+		dockerExec:     dockerExec,
+	}
+}
+
+func (v *validationManager) ValidateCreate(ctx context.Context) error {
```
g-gaston, line 34: I think `Validate` is enough, since this is running in a validation manager in the validations/createcluster package

pkg/validations/createcluster/createcluster.go
raymond-zhang00: updated to `NewValidations`

pkg/validations/createcluster/createcluster.go
raymond-zhang00: Perfect! Updated this

pkg/validations/createcluster/createcluster.go
raymond-zhang00: Updated this to `createValidations` and got rid of the interfaces.Validator reference

pkg/validations/createcluster/validateprovider.go
raymond-zhang00: Moved this into the providers pkg as `ValidateActive()`. I created this validation from the `create cluster` cmd but noticed `delete cluster` only checks Cloudstack. Should I make this `ValidateActiveCreate()`?

pkg/validations/createcluster/createcluster_test.go
raymond-zhang00: Updated this to be generated within go code and removed config file

pkg/kubeconfig/kubeconfig.go
raymond-zhang00: Added a test, to do so I did need to remove `*.kubeconfig` from our gitignore

pkg/validations/createcluster/createcluster.go
raymond-zhang00: Defined the interface for createValidations and made a mock for it.

pkg/validations/createcluster/createcluster.go
raymond-zhang00: I think this would be good to do, but it would also make `validate create cluster` more independent of `create cluster` since `create cluster` is still reliant on the `createValidations.PreflightValidations()`. 

Could this be done in a separate PR, maybe accompanying  `create cluster` using this or an updated set of validations?

g-gaston: Not necessarily. We can still have all the validations defined where they are. Just allow to return a slice with all validations as opposed to only have one method that returns them

pkg/kubeconfig/kubeconfig.go
taneyland: Tweaked the tests a bit so that we can keep this in gitignore

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/gitops/flux"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/providers/validator"
+	"github.com/aws/eks-anywhere/pkg/validations"
+)
+
+type validationManager struct {
```
g-gaston, line 16: can we make this struct public?

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,90 @@
...
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name: "create preflight validations pass",
+				Err:  v.createValidations.PreflightValidations(ctx),
+			}
+		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name: fmt.Sprintf("validate %s Provider", v.provider.Name()),
+				Err:  v.provider.SetupAndValidateCreateCluster(ctx, v.clusterSpec),
```
g-gaston, line 86: do these two need to be silent as well?

pkg/validations/createcluster/createcluster.go
```diff
@@ -0,0 +1,90 @@
...
+type validationManager struct {
+	clusterSpec       *cluster.Spec
+	provider          providers.Provider
+	gitOpsFlux        *flux.Flux
+	createValidations Validator
+	dockerExec        validations.DockerExecutable
+}
+
+type Validator interface {
+	PreflightValidations(ctx context.Context) error
```
g-gaston, line 25: Is it very difficult to have a `GetValidations() []validations.Validation` method?
I'm assuming the other object is creating a slice of validations and executing them. It should be as easy as splitting those two steps, right?

pkg/validations/createcluster/createcluster.go
taneyland: No. We didn't add Silent here to keep the output consistent with what is shown when running the create cmd

pkg/validations/createcluster/createcluster.go
taneyland: Why public? We don't need to access the struct outside of this package

g-gaston (approved):

PR #3120, opened 2022-08-23
Update metallb link to their github page

lewisdiamond:
The column defined in the parent template is called "GitHub" so update the link to their github page.


binyamz (approved):
/lgtm

PR #3122, opened 2022-08-23
Disable apiServer anonymous auth

ahreehong:
*Description of changes:*
apiServer has anonymous auth enabled by default. These changes disables this for both vsphere and baremetal

### DIFF
--- a/pkg/providers/tinkerbell/config/template-cp.yaml
+++ b/pkg/providers/tinkerbell/config/template-cp.yaml
@@ -67,9 +67,10 @@ spec:
         imageRepository: {{.bottlerocketBootstrapRepository}}
         imageTag: {{.bottlerocketBootstrapVersion}}
 {{- end }}
-{{- if .apiserverExtraArgs }}
       apiServer:
         extraArgs:
+          anonymous-auth: "false"
+{{- if .apiserverExtraArgs }}
 {{ .apiserverExtraArgs.ToYaml | indent 10 }}
 {{- end }}
 {{- if .awsIamAuth}}
--- a/pkg/providers/tinkerbell/testdata/expected_results_cluster_tinkerbell_cp_awsiam.yaml
+++ b/pkg/providers/tinkerbell/testdata/expected_results_cluster_tinkerbell_cp_awsiam.yaml
@@ -41,6 +41,7 @@ spec:
         imageTag: v1.8.3-eks-1-21-4
       apiServer:
         extraArgs:
+          anonymous-auth: "false"
           authentication-token-webhook-config-file: /etc/kubernetes/aws-iam-authenticator/kubeconfig.yaml
           feature-gates: ServiceLoadBalancerClass=true
         extraVolumes:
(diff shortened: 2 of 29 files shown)


jaxesn:
I think this is probably the correct change, however this is technically a breaking change since we are changing the default.  Lets make sure we have a discussion between @g-gaston @jiayiwang7 @vivek-koppuru @vignesh-goutham and @micahhausler if possible to make sure we consider the potential side effects here

vivek-koppuru:
What is the motivation behind this change?

chrisdoherty4:
@vivek-koppuru I think this came out of the bare metal security review.

> When enabled, requests that are not rejected by other configured authentication methods are treated as anonymous requests, and given a username of system:anonymous and a group of system:unauthenticated.
> 
> For example, on a server with token authentication configured, and anonymous access enabled, a request providing an invalid bearer token would receive a 401 Unauthorized error. A request providing no bearer token would be treated as an anonymous request.
> 
> In 1.5.1-1.5.x, anonymous access is disabled by default, and can be enabled by passing the --anonymous-auth=true option to the API server.
> 
> In 1.6+, anonymous access is enabled by default if an authorization mode other than AlwaysAllow is used, and can be disabled by passing the --anonymous-auth=false option to the API server. Starting in 1.6, the ABAC and RBAC authorizers require explicit authorization of the system:anonymous user or the system:unauthenticated group, so legacy policy rules that grant access to the * user or * group do not include anonymous users.

- https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests


It seems users are evaluated as anonymous if no configured authenticating methods reject the request and there's no user identifying information in the request. In 1.6+ setting authentication mode to something other than AlwaysAllow enables anonymous mode so we still need to explicitly disable it. Disabling anonymous mode just means anyone invoking `kubectl` needs to provide an identity in some way, such as a cert, which I think is standard practice in most environments (KinD even does this). It is different behavior but I'm not sure I'd qualify this as a breaking change.

vignesh-goutham:
Yea, this should not affect our cluster operations. @ahreehong is currently testing this change on both Ubuntu and Bottlerocket.

/hold

ndeksa:
@jaxesn @vivek-koppuru  is there any particular use case you suggest to verify before planning on including this fix? We will run CI and conformance testing.

vivek-koppuru:
> @jaxesn @vivek-koppuru is there any particular use case you suggest to verify before planning on including this fix? We will run CI and conformance testing.

I am not sure if testing will show anything here, it's more a change in the user's behavior if they run an upgrade to this version of the cluster from a previous version and realize that they can't make anonymous calls anymore. We do have to list this as a breaking change in our changelog because of this I think.

vignesh-goutham:
It looks like kubeadm join with bootstrap token gets redirected/classified as `system:anonymous` / `system:unauthenticated` on first control plane api-server that comes up. This is considered secure as the bootstrap token itself has an expiry on it. This is a known issue and has no fix. Kubeadm join doesn't work with anonymous auth on api-server. https://github.com/kubernetes/kubeadm/issues/1105

While testing @ahreehong  and I also figured out another problem. Health/liveness/readiness probes will now need authentication. While bootstrapping Bottlerocket we check for the static pod health to confirm the bootstrapping progress and those will fail with anonymous auth set as false.

The scope to fix this is quite big, so I'd suggest to close this PR for now, while we can engage kubeadm and see if they have a fix planned or if we can contribute a fix.

PR #3134, opened 2022-08-23
Add analyzer to analyze if vms have access to vsphere api server

YoyoTT:
*Issue #, if available:*

*Description of changes:*

To extent the functionality of support bundle by 
- adding a run-pod collector--vmsAccessCollector. It supposes to collect logs from vsphere cloud controller manager on control plane. 
    - since the pod need to be deployed on control plane node, specific tolerations have been set. 
    - It is also able to accept customized taints that given by user in cluster defination yaml. 
- adding a vmsAccess analyzer. Based on the log from the collector, the analyzer is able to analyze if the vms have access to vsphere api server. 
   - Now, the analyzer starts to work from bootstrap cluster, although we expect that the analyzer only does its work on workload cluster. A clarification was temporarily added to the output to avoid confusion. 

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -33,7 +33,6 @@ import (
 )
 
 const (
-<<<<<<< HEAD
 	maxRetries                = 30
 	defaultBackOffPeriod      = 5 * time.Second
 	machineBackoff            = 1 * time.Second
@@ -46,18 +45,6 @@ const (
 	controlPlaneInProgressStr = "1m"
 	etcdInProgressStr         = "1m"
 	DefaultEtcdWait           = 60 * time.Minute
-=======
-	maxRetries             = 30
-	backOffPeriod          = 5 * time.Second
-	machineBackoff         = 1 * time.Second
-	machinesMinWait        = 30 * time.Minute
-	moveCAPIWait           = 15 * time.Minute
-	clusterWaitStr         = "60m"
-	ctrlPlaneWaitStr       = "60m"
-	deploymentWaitStr      = "30m"
-	ctrlPlaneInProgressStr = "1m"
-	etcdInProgressStr      = "1m"
->>>>>>> 7bcd36d1 (Add new analyzer and collector to support bundle)
 )
 
 type ClusterManager struct {
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -283,7 +283,8 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 				return err
 			},
 		)
-		return workloadCluster, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
+		// add warning here *******
+		return nil, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
 	}
 
 	logger.V(3).Info("Waiting for workload kubeconfig generation", "cluster", workloadCluster.Name)
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -33,6 +33,7 @@ import (
 )
 
 const (
+<<<<<<< HEAD
 	maxRetries                = 30
 	defaultBackOffPeriod      = 5 * time.Second
 	machineBackoff            = 1 * time.Second
@@ -45,6 +46,18 @@ const (
 	controlPlaneInProgressStr = "1m"
 	etcdInProgressStr         = "1m"
 	DefaultEtcdWait           = 60 * time.Minute
+=======
+	maxRetries             = 30
+	backOffPeriod          = 5 * time.Second
+	machineBackoff         = 1 * time.Second
+	machinesMinWait        = 30 * time.Minute
+	moveCAPIWait           = 15 * time.Minute
+	clusterWaitStr         = "60m"
+	ctrlPlaneWaitStr       = "60m"
+	deploymentWaitStr      = "30m"
+	ctrlPlaneInProgressStr = "1m"
+	etcdInProgressStr      = "1m"
+>>>>>>> 7bcd36d1 (Add new analyzer and collector to support bundle)
 )
 
 type ClusterManager struct {
@@ -277,7 +290,13 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	logger.V(3).Info("Waiting for control plane to be ready")
 	err = c.clusterClient.WaitForControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), workloadCluster.Name)
 	if err != nil {
-		return nil, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
+		err = c.Retrier.Retry(
+			func() error {
+				workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
+				return err
+			},
+		)
+		return workloadCluster, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
 	}
 
 	logger.V(3).Info("Waiting for workload kubeconfig generation", "cluster", workloadCluster.Name)
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -277,14 +277,17 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	logger.V(3).Info("Waiting for control plane to be ready")
 	err = c.clusterClient.WaitForControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), workloadCluster.Name)
 	if err != nil {
-		err = c.Retrier.Retry(
+		logger.Info("Warning: attempting to create kubeconfig")
+		retryErr := c.Retrier.Retry(
 			func() error {
 				workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
 				return err
 			},
 		)
-		// add warning here *******
-		return nil, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
+		if retryErr != nil {
+			logger.Info("Warning: unable to generate workload kubeconfig")
+		}
+		return workloadCluster, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
 	}
 
 	logger.V(3).Info("Waiting for workload kubeconfig generation", "cluster", workloadCluster.Name)
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -278,14 +278,16 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	err = c.clusterClient.WaitForControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), workloadCluster.Name)
 	if err != nil {
 		logger.Info("Warning: attempting to create kubeconfig")
+
 		retryErr := c.Retrier.Retry(
 			func() error {
-				workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
-				return err
+				var getKubconfigErr error
+				workloadCluster.KubeconfigFile, getKubconfigErr = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
+				return getKubconfigErr
 			},
 		)
 		if retryErr != nil {
-			logger.Info("Warning: unable to generate workload kubeconfig")
+			logger.Info("Warning: unable to generate workload kubeconfig.")
 		}
 		return workloadCluster, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
 	}
--- a/pkg/diagnostics/analyzers.go
+++ b/pkg/diagnostics/analyzers.go
@@ -223,6 +223,7 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
 func (a *analyzerFactory) namespaceLogTextAnalyzersMap() map[string][]*Analyze {
 	return map[string][]*Analyze{
 		constants.CapiKubeadmControlPlaneSystemNamespace: a.capiKubeadmControlPlaneSystemLogAnalyzers(),
+		constants.KubeSystemNamespace:                    a.kubeSystemLogAnalyzers(),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


jiayiwang7:
@YoyoTT can u please add PR description, also make it less generic of the PR title? Thanks!

pkg/clustermanager/cluster_manager.go
```diff
@@ -277,7 +277,13 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	logger.V(3).Info("Waiting for control plane to be ready")
 	err = c.clusterClient.WaitForControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), workloadCluster.Name)
 	if err != nil {
-		return nil, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
+		err = c.Retrier.Retry(
```
vivek-koppuru, line 280: If this `err` is not nil, we probably should add a warning message and return nil instead of workloadCluster

pkg/diagnostics/analyzers.go
```diff
@@ -223,6 +223,7 @@ func (a *analyzerFactory) EksaLogTextAnalyzers(collectors []*Collect) []*Analyze
 func (a *analyzerFactory) namespaceLogTextAnalyzersMap() map[string][]*Analyze {
 	return map[string][]*Analyze{
 		constants.CapiKubeadmControlPlaneSystemNamespace: a.capiKubeadmControlPlaneSystemLogAnalyzers(),
+		// constants.KubeSystemNamespace:                    a.kubeSystemLogAnalyzers(),
```
mitalipaygude, line 226: can we remove the commented code?

pkg/diagnostics/collectors.go
```diff
@@ -435,7 +437,70 @@ func (c *collectorFactory) pingHostCollector(hostIP string) *Collect {
...
 					Args:    argsPing,
 				}},
+			}, Timeout: "20s",
+		},
+	}
+}
+
+// vmsAccessCollector will connect to API server first, then collect vsphere-cloud-controller-manager logs
+// on control plane node
+func (c *collectorFactory) vmsAccessCollector(controlPlaneEndpoint string) *Collect {
```
vivek-koppuru, line 447: This sounds more like a vsphereCloudControllerCollector right?

YoyoTT: > This sounds more like a vsphereCloudControllerCollector right?

Right, it's vsphere's.

pkg/clustermanager/cluster_manager.go
YoyoTT: Yeah, i will change the return of workloadCluster to nil. Could the warning message be like "Control plane is not ready, generating workload kubeconfig" ?

pkg/diagnostics/collectors.go
danbudris: yeah it's a bit different than our standard log collectors as it's actually making use directly of a pod execution and explicitly invoking `kubectl`, instead of using the built-in troubleshoot log collectors. It might be good to extend the comment on this collector explaining why we need to target the control plane nodes and use this specific approach

danbudris: iirc and I might be messing this up, the specific log line that we're looking for in these results only occurs on the vsphere-cloud-controller-manager pod on the CP node; if we collect from multiple nodes, such as the worker nodes, it adds a significant amount of noise to the analysis results which could lead to customer confusion. Is that right @YoyoTT ?

danbudris (commented):
love this, should help solve some otherwise tricky-to-diagnose customer issues. Left a couple comments but nothing huge.

pkg/diagnostics/analyzers.go
```diff
@@ -362,3 +366,33 @@ func (a *analyzerFactory) controlPlaneIPAnalyzer() []*Analyze {
...
+
+// vmsAccessAnalyzer will analyze if vms have access to the API server of vSphere cluster
+func (a *analyzerFactory) vmsAccessAnalyzer() *Analyze {
+	runBashPod := "check-cloud-controller"
+	runBashPodLog := fmt.Sprintf("%s.log", runBashPod)
+	vSphereCloudControllerPodLogPath := path.Join(runBashPod, runBashPodLog)
+	return &Analyze{
+		TextAnalyze: &textAnalyze{
+			analyzeMeta: analyzeMeta{
+				CheckName: fmt.Sprintf("%s: VMs has no access to vSphere APi. Logs: %s", logAnalysisAnalyzerPrefix, vSphereCloudControllerPodLogPath),
```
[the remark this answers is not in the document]
YoyoTT: Thanks! Will make that change and update it in next commit.

pkg/clustermanager/cluster_manager.go
vivek-koppuru: Well I am referring to if kubeconfig generation fails, so a warning like "attempting to create kubeconfig: <error>"

pkg/clustermanager/cluster_manager.go
```diff
@@ -277,7 +277,13 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
...
+			func() error {
+				workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
+				return err
+			},
+		)
+		return workloadCluster, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
```
jiayiwang7, line 286: we can just `return nil, fmt.Errorf()`.
also in this case, you are overriding the `err` from line 279, with err generated from line 280. There is no way to indicate the `WaitForControlPlaneReady` error then.

pkg/clustermanager/cluster_manager.go
YoyoTT: > Well I am referring to if kubeconfig generation fails, so a warning like "attempting to create kubeconfig: "

Okay, I probably need to add that warning after this block( `err=c.Retrier.Retry(     )`), right?

pkg/diagnostics/collectors.go
YoyoTT: > 

Yes, that's right! That's also the reason why we want to get the log of vsphere-cloud-controller-manager pod directly from control plane node instead of using the built-in troubleshoot log collector.

pkg/diagnostics/collectors.go
```diff
@@ -435,7 +437,70 @@ func (c *collectorFactory) pingHostCollector(hostIP string) *Collect {
...
+			PodSpec: &v1.PodSpec{
+				Containers: []v1.Container{{
+					Name:    "check-cloud-controller",
+					Image:   c.DiagnosticCollectorImage,
+					Command: []string{"/bin/sh", "-c"},
+					Args:    args,
+				}},
+				ServiceAccountName: "default",
+				HostNetwork:        true,
+				Tolerations: []v1.Toleration{
```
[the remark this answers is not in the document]
YoyoTT: Great catch! I got these tolerations by trial-and-error on a standard cluster. So once user customized the taints, then the pod may not be deployed as we expected. 
Thanks for reminding me this. I'll try to figure it out and extract the taint.

pkg/clustermanager/cluster_manager.go
vivek-koppuru: If kubeconfig is set, this would allow the support bundle to be created for workload cluster, otherwise if you return nil it won't

YoyoTT: Yeah, I made that change to return a nil instead of workloadcluster. 
And right, I probably need to re-name the `err` to something like `error` to distinguish the two?

YoyoTT: I am little bit confused here. If we don't return workLoadCluster at line 286, how can we get the kubeconfig? It seems like that kubeconfigFile is one of the attributes of workLoadCluster, right?

jiayiwang7: i see, make sense. the overriding error msg still valid right

pkg/diagnostics/analyzers.go
```diff
@@ -362,3 +366,33 @@ func (a *analyzerFactory) controlPlaneIPAnalyzer() []*Analyze {
...
+			},
+			FileName:     vSphereCloudControllerPodLogPath,
+			RegexPattern: `Failed to create new client. err: Post (.*) dial tcp (.*) connect: connection timed out\n(.*)Failed to create govmomi client. err: Post (.*) dial tcp (.*) connect: connection timed out`,
+			Outcomes: []*outcome{
+				{
+					Fail: &singleOutcome{
+						When:    "true",
+						Message: fmt.Sprintf("Failed to create client, Virtural Machines have no access to vSphere API. See the cloud controller log in master node: %s", vSphereCloudControllerPodLogPath),
```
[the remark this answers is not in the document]
YoyoTT: > I wonder if there's a relevent link to the documentation we can provide. I don't know if we have docs which lay out the required connection to Vsphere but that would be ideal.

Do you mind to tell more about which kind of doc would make this better? I probably didn't get the point when you said `required connection to vsphere` . 
My understanding is that it's reasonable if we assume that it should connect to vsphere because of that this analyzer supposes to do its job when it's confirmed that a vSphere cluster will be created.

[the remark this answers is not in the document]
YoyoTT: I'm also trying to add a warning message to this analyzer. Because the analyzer should only work on workload cluster instead of bootstrap cluster. Currently, it's not that smart to be able to figure out the condition it is woking on. Hopefully, a warning message could guide user to make a better use of it.

pkg/clustermanager/cluster_manager.go
vivek-koppuru: In order to make this work though, we need to also add this line in the err block in this file right? https://github.com/aws/eks-anywhere/blob/249f1abdf2a3c740ef3e05fdbf4031353c7460bb/pkg/workflows/create.go#L219

pkg/clustermanager/cluster_manager.go
```diff
@@ -277,7 +277,17 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	logger.V(3).Info("Waiting for control plane to be ready")
 	err = c.clusterClient.WaitForControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), workloadCluster.Name)
 	if err != nil {
-		return nil, fmt.Errorf("waiting for workload cluster control plane to be ready: %v", err)
+		logger.Info("Warning: attempting to create kubeconfig")
+		retryErr := c.Retrier.Retry(
+			func() error {
+				workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
```
g-gaston, line 283: Didn't we move this somewhere else on purpose?
What's the goal of adding it back again?

pkg/clustermanager/cluster_manager.go
g-gaston: I would vote to find an alternative to this.

This method now returns:
* (valid cluster, nil)
* (nil, error)
* (valid cluster, error)
* (invalid cluster without kubeconfig, error)

It's a bit confusing, it requires the caller to know about the internal implementation and the returned values are not easy to handle.

I would start by listing the scenarios where `WaitForControlPlaneReady` might fail but we might still be able to generate the kubeconfig.

pkg/diagnostics/collectors.go
```diff
@@ -435,9 +437,98 @@ func (c *collectorFactory) pingHostCollector(hostIP string) *Collect {
...
+				HostNetwork:        true,
+				Tolerations:        tolerations,
+				Affinity: &v1.Affinity{
+					NodeAffinity: &v1.NodeAffinity{
+						PreferredDuringSchedulingIgnoredDuringExecution: []v1.PreferredSchedulingTerm{
+							{
+								Weight: 10,
+								Preference: v1.NodeSelectorTerm{
+									MatchExpressions: []v1.NodeSelectorRequirement{{
+										Key:      "node-role.kubernetes.io/control-plane",
```
g-gaston, line 476: curious, why does this collector need to run in the cp?

pkg/clustermanager/cluster_manager.go
YoyoTT: > Didn't we move this somewhere else on purpose? What's the goal of adding it back again?

We want to get the kubeconfig even thought control plane is not ready. That kubeconfig would be used for generating support bundle on the workload cluster. Am I answering your questions?

pkg/clustermanager/cluster_manager.go
vivek-koppuru: I think it does make sense though right? If we get an error, the workflow will go through the error path, and if not, it won't. All we want from this method is whatever form of cluster is ready, and we already have those existing validations for generating the support bundle in the diagnostic workflow. But do you feel that we should instead be splitting this up more to only return a combination of `nil, error`?

pkg/diagnostics/collectors.go
YoyoTT: To deal with the problem that vms have no access to api server, the analyzer need info from vsphere-cloud-controller-manager.log on cp, which should be collected by a log collecter, ideally. However, that log collector must collect vsphere-cloud-controller-manager.log from control plane and worker nodes, then generate corresponding  amount of log files with almost the same title. The analyzer is not able to figure out which log file it should work with. Therefore, we deploy a pod and run it in the cp to get the exact log that the analyzer needs to analyze for this issue.

pkg/clustermanager/cluster_manager.go
g-gaston: I think we need a better definition of what this method does. Does it just send the request to create the cluster? Does it wait until the kubeconfig is ready? Does it wait until the CP is ready?

Based on the previous implementation, a CP ready was the minimum we needed to consider the cluster created. Now it seems we think we could have a kubeconfig without the control plane being ready. But a kubeconfig without a control plane doesn't do anything. So that makes me think we need to analyze a bit more that scenario and list where `WaitForControlPlaneReady` might fail but we might still be able to generate the kubeconfig. I think that will point us into the right direction, which probably involves making some more granular condition checks (more granular than `WaitForControlPlaneReady`).

pkg/clustermanager/cluster_manager.go
g-gaston: Yes, but when is a kubeconfig valuable if the control plane is not ready?

pkg/diagnostics/collectors.go
g-gaston: What do you mean exactly by `vsphere-cloud-controller-manager.log`? Are you reading the logs from the node? I thought you were using `kubectl logs` to retrieve the logs from the cloud manager pod.

pkg/clustermanager/cluster_manager.go
YoyoTT: > Yes, but when is a kubeconfig valuable if the control plane is not ready?

Please correct me if my understanding is wrong, the statement that the control plane is not ready means the control plane has been created but not in a health condition. Once it has been created, the kubeconfig is valuable, at least for diagnostic process of support bundle. We can get data about current cluster status through the kubeconfig, then start to analyze. Does it make sense?

pkg/diagnostics/collectors.go
YoyoTT: > What do you mean exactly by `vsphere-cloud-controller-manager.log`? Are you reading the logs from the node? I thought you were using `kubectl logs` to retrieve the logs from the cloud manager pod.

Sorry I didn't make it clear. Once the issue that VMs have no access to api server happens, only log that contains the corresponding error message is the vsphere-cloud-controller-manager.log from the Control Plane Node. Ideally, log collector should read logs from the nodes by using `kubectl logs` and it's doing this job as well now. Then analyzer will open the log files and catch the error message. However, the log files we got not only from control plane but also from worker nodes. The analyzer cannot figure out which log file it should analyze so it does its job on every log files, and gives us multiple results, which is confused and unexpected. Therefore, we deploy a pod and run it in the cp to directly get the vsphere-cloud-controller-manager.log from Control Plane Node. The analyzer can analyze for this issue and give us accurate diagnostic result.

pkg/clustermanager/cluster_manager.go
YoyoTT: > I think we need a better definition of what this method does. Does it just send the request to create the cluster? Does it wait until the kubeconfig is ready? Does it wait until the CP is ready?
> 
> Based on the previous implementation, a CP ready was the minimum we needed to consider the cluster created. Now it seems we think we could have a kubeconfig without the control plane being ready. But a kubeconfig without a control plane doesn't do anything. So that makes me think we need to analyze a bit more that scenario and list where `WaitForControlPlaneReady` might fail but we might still be able to generate the kubeconfig. I think that will point us into the right direction, which probably involves making some more granular condition checks (more granular than `WaitForControlPlaneReady`).

Just want to clarify that we do need a kubeconfig without the control plane being ready. The fact got from multiple tests has verified that, without the cp being ready, the kubeconfig can help support bundle to diagnose cluster. What my unsolicited thought about this condition is, as long as control plane is created, kubeconig should be valuable.

g-gaston: I understand that. But what I think we need are the condition under which the control plane might not be ready but the kubeconfig is available and useful.

pkg/clustermanager/cluster_manager.go
g-gaston: That depends why the CP is not ready. For example, if the kube api server sis not running or not available, the kubeconfig won't help us.
I think this turned into the same convo as the other thread, let's move this there.

pkg/diagnostics/collectors.go
g-gaston: Then I'm missing something. Isn't this pod collector just running `kubectl logs`?

YoyoTT: > Then I'm missing something. Isn't this pod collector just running `kubectl logs`?

You're good. I should clarify it first. Yes and No to answer your question. Yes, the collector run "kubectl logs" to obtain logs. No, becasue it's not just run "kubectl logs", it also need to connect control plane first.

PR #3153, opened 2022-08-24
Specify vsphere template in latest release tests

g-gaston:
*Description of changes:*
Ubuntu doesn't support autoimport anymore so we can't rely on it to set
the template when creating clusters using the latest minor release.
Bottlerocket doesn't strictly need this but making the same change as
well for consistency. Plus it's a more robust solution since it doesn't
need to write to vsphere, avoiding race conditions with other test runs.

Using env vars for consistency with the rest of the input data. The env
var name is constructed from the OS and EKS-D version. These env vars
need to be set prior to running the tests.

When we release a new eks-a version with new eks-d versions, we need to
import template for those and add new variables for the tests to work
properly.

New variables that we need to add:

```sh
T_VSPHERE_UBUNTU_KUBERNETES_1_20_EKS_19
T_VSPHERE_UBUNTU_KUBERNETES_1_21_EKS_17
T_VSPHERE_UBUNTU_KUBERNETES_1_22_EKS_9
T_VSPHERE_UBUNTU_KUBERNETES_1_23_EKS_4

T_VSPHERE_BOTTLEROCKET_KUBERNETES_1_20_EKS_19
T_VSPHERE_BOTTLEROCKET_KUBERNETES_1_21_EKS_17
T_VSPHERE_BOTTLEROCKET_KUBERNETES_1_22_EKS_9
T_VSPHERE_BOTTLEROCKET_KUBERNETES_1_23_EKS_4
```

*Testing (if applicable):*
Ran a couple of ubuntu and a bottlerocket tests.


jaxesn (approved):
/approve

PR #3174, opened 2022-08-25
Add phase for workers in snow cluster reconciler

g-gaston:
*Description of changes:*

*Testing (if applicable):*
Unit tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aws/eks-anywhere
 
-go 1.17
+go 1.18
 
 require (
 	github.com/Masterminds/sprig v2.22.0+incompatible
--- a/internal/test/envtest/client.go
+++ b/internal/test/envtest/client.go
@@ -4,42 +4,81 @@ import (
 	"context"
 	"testing"
 
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type Object interface {
-	client.Object
-	schema.ObjectKind
-}
+type Object client.Object
 
 // CreateObjs creates Objects using the provided kube client and waits until its cache
 // has been updated with those objects
-func CreateObjs(ctx context.Context, t testing.TB, client client.Client, objs ...Object) {
+func CreateObjs(ctx context.Context, t testing.TB, c client.Client, objs ...Object) {
 	t.Helper()
-	for _, obj := range objs {
-		// client.Create cleans the group version and kind of the objects
-		// We just save it before calling and it and restore it later,
-		// just so we can use it to perform the get later
-		objKind := obj.GroupVersionKind()
-		if err := client.Create(ctx, obj); err != nil {
+	for _, o := range objs {
+		// we copy objects because the client modifies them while making creating/updating calls
+		obj := copyObject(t, o)
+
+		if err := c.Create(ctx, obj); isNamespace(obj) && apierrors.IsAlreadyExists(err) {
+			// namespaces can't be deleted
+			// assuming most tests just want the namespace to exist, since it already does
+			// we ignore the error
+			// for more advance usecases, handle namespaces manually outside of this helper
+			continue
+		} else if err != nil {
 			t.Fatal(err)
 		}
-		obj.SetGroupVersionKind(objKind)
 	}
 
-	for _, obj := range objs {
-		for {
-			o := &unstructured.Unstructured{}
-			o.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
-			if err := client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, o); err == nil {
-				break
-			} else if !apierrors.IsNotFound(err) {
-				t.Fatal(err)
-			}
+	for _, o := range objs {
+		obj := copyObject(t, o)
+		objReady := waitForObjectReady(ctx, t, c, obj)
+
+		// We need to update the status independently, kubernetes doesn't allow to create the main objects and
+		// its subresources all at once
+		obj.SetResourceVersion(objReady.GetResourceVersion())
+		if err := c.Status().Update(ctx, obj); apierrors.IsNotFound(err) {
+			// There is not easy way to check if an object has a status subresource
+			// So we just try and if it fails with a 404, we ignore the error
+			t.Logf(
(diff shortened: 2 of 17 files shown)


internal/test/envtest/environment.go
```diff
@@ -201,6 +214,10 @@ func (e *Environment) CreateNamespaceForTest(ctx context.Context, t *testing.T)
 }
 
 func getRootPath() string {
+	return path.Join(currentDir(), "..", "..", "..")
```
vivek-koppuru, line 217: Can this work instead of guessing the 4 dirs? https://stackoverflow.com/a/32163888

pkg/controller/serverside/applier.go
```diff
@@ -0,0 +1,40 @@
...
+	client client.Client
+}
+
+func NewObjectApplier(client client.Client) *ObjectApplier {
+	return &ObjectApplier{
+		client: client,
+	}
+}
+
+func (a *ObjectApplier) Apply(ctx context.Context, generateObjects ObjectGenerator) (controller.Result, error) {
```
vivek-koppuru, line 25: Why are we passing in a func instead of just passing in kubernetes.Object as an array or opts?

pkg/providers/snow/snow.go
```diff
@@ -118,14 +118,23 @@ func CAPIObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kube
...
-	workersSpec, err = templater.ObjectsToYaml(workersObjs...)
+	workersSpec, err = templater.ObjectsToYaml(kubernetesToRuntimeObjects(workersObjs)...)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	return controlPlaneSpec, workersSpec, nil
 }
 
+func kubernetesToRuntimeObjects(objs []kubernetes.Object) []runtime.Object {
```
vivek-koppuru, line 129: Can you explain the difference between kubernetes and runtime objects?

pkg/controller/clientutil/object.go
```diff
@@ -0,0 +1,14 @@
+package clientutil
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ObjectsToClientObjects[T client.Object](objs []T) []client.Object {
```
taneyland, line 7: why not make these objs `client.Object` from the beginning? 
Does `kubernetes.Object` have more methods that we need to use before we get to this point?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/eks-anywhere
 
-go 1.17
+go 1.18
```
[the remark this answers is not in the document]
g-gaston: The `go.mod` wasn't changes in the original 1.18 PR. This line in go.mod is not indicate which version of go to use but to set a minimum requirement. Until now, we didn't have a 1.18 requirement, but this PR uses a feature only available in 1.18, hence the change.

internal/test/envtest/environment.go
g-gaston: That's what we use to get `currentDir()`. But from there, we need to go to the root of the repo. And we know that's 3 folders up the stack because we know where this file is.

pkg/controller/serverside/applier.go
g-gaston: That's just for convenience. If I passed the objects, I wouldn't need this applier, I could just use the `serverside.ReconcileObjecst` method.

This allows me to pass the generator of objects (method implemented in the snow package) and let this applier do the error handling.

pkg/providers/snow/snow.go
g-gaston: Good question. `runtime.Object` is the basic kubernetes interface for api objects. `client.Object` is the interface used by the `controller-runtime` `Client`, which is a superset of `runtime.Object`. Which is the reason why we can't use `runtime.Object` with the controller code.

`kubernetes.Object` is the interface we created to represent api objects in our eks-a code, in particular when dealing with share code between the controller and CLI. `kubernetes.Object` should always implement `client.Object`

We could use `client.Client` directly in the CLI code as well but that seemed like a dangerous commitment. By having our own interface, we can add more stuff to it without necessarily having to update the signature of all methods in those pieces of shared code.

pkg/controller/clientutil/object.go
g-gaston: Ahh great question again, moved convo to https://github.com/aws/eks-anywhere/pull/3174#discussion_r955519684

internal/test/envtest/client.go
```diff
@@ -4,42 +4,81 @@ import (
...
-		// We just save it before calling and it and restore it later,
-		// just so we can use it to perform the get later
-		objKind := obj.GroupVersionKind()
-		if err := client.Create(ctx, obj); err != nil {
+	for _, o := range objs {
+		// we copy objects because the client modifies them while making creating/updating calls
+		obj := copyObject(t, o)
+
+		if err := c.Create(ctx, obj); isNamespace(obj) && apierrors.IsAlreadyExists(err) {
+			// namespaces can't be deleted
```
mitalipaygude, line 25: The comments are really helpful! Thank you for adding those.

pkg/providers/snow/objects.go
```diff
@@ -32,10 +31,10 @@ func ControlPlaneObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeCli
 	}
 	capiCluster := CAPICluster(clusterSpec, snowCluster, kubeadmControlPlane)
 
-	return []runtime.Object{capiCluster, snowCluster, kubeadmControlPlane, new}, nil
+	return []kubernetes.Object{capiCluster, snowCluster, kubeadmControlPlane, new}, nil
 }
 
-func WorkersObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]runtime.Object, error) {
+func WorkersObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]kubernetes.Object, error) {
```
mitalipaygude, line 37: nit: should this be named `WorkerObjects` like the `ControlPlaneObjects` above?

g-gaston: ummm I don't have a strong opinion
I think the idea was that there is only one CP but possibly multiple workers

mitalipaygude:
Can we add more unit tests so that the code coverage goes up?

pkg/controller/serverside/applier.go
vivek-koppuru: Yea I guess that's what I was confused about, I feel it makes it a bit harder to understand the flow because when I think of `Apply` I think of passing in the objects rather than a func that generates them. Does it make sense to call this `ApplyWithGenerator`, however I am assuming that having "with" in naming is probably not ideal.

g-gaston: I see I see. Given that the only purpose of this object is to server as a helper for this, isn't it just redundant info?

What about renaming the struct to `ObjectGenerateApplier` or `ApplierForGenerator`?

vivek-koppuru (approved):

PR #3175, opened 2022-08-26
add omitempty for optional strings in disk offering spec

wongni:
*Issue #, if available:*

*Description of changes:*
A customer reported the empty strings under disk offering spec are not omitted. This may cause unwanted cluster upgrade to occur. 

```
kind: CloudStackMachineConfig
metadata:
  creationTimestamp: null
  name: cluster-name-26z-cp
  namespace: cluster-ns
spec:
  computeOffering:
    name: 6 vCore 8GB Memory
  diskOffering:
    device: ""
    filesystem: ""
    label: ""
    mountPath: ""
  template:
```

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
@@ -56,13 +56,13 @@ type CloudStackResourceDiskOffering struct {
 	// +optional
 	CustomSize int64 `json:"customSizeInGB,omitempty"`
 	// path the filesystem will use to mount in VM
-	MountPath string `json:"mountPath"`
+	MountPath string `json:"mountPath,omitempty"`
 	// device name of the disk offering in VM, shows up in lsblk command
-	Device string `json:"device"`
+	Device string `json:"device,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -56,13 +56,13 @@ type CloudStackResourceDiskOffering struct {
 	// +optional
 	CustomSize int64 `json:"customSizeInGB,omitempty"`
 	// path the filesystem will use to mount in VM
-	MountPath string `json:"mountPath"`
+	MountPath string `json:"mountPath,omitempty"`
```
maxdrib, line 59: I think we'll need some preflight validation in addition to this to make sure the fields are set as expected. Previously, it was assumed that all these fields were present and that may now not be the case.

wanyufe: I remembered that MountPath/device/fs/label is required fields if disk offering is configured.

wongni:
This fix does not solve the customer problem and they are making a workaround.

PR #3177, opened 2022-08-26
Build cluster-controller image for consumption in E2E presubmit

abhay-krishna:
This PR adds logic to build the cluster-controller image in the E2E presubmit and consume it for the E2E tests. This enables the controller image to be tested in presubmits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,16 @@ DOCKERFILE_FOLDER = ./manager/docker/linux/eks-anywhere-cluster-controller
 
 IMAGE_REPO?=$(if $(AWS_ACCOUNT_ID),$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com,localhost:5000)
 IMAGE_TAG?=$(GIT_TAG)-$(shell git rev-parse HEAD)
+# Tagging with the commit hash and the branch the PR is raised against
+ifeq ($(JOB_NAME),eks-anywhere-e2e-presubmit)
+	IMAGE_PLATFORMS=linux/amd64
+	IMAGE_TAG=$(PULL_PULL_SHA).$(PULL_BASE_REF)
+	IMAGES_TO_BUILD?=$(CLUSTER_CONTROLLER_IMAGE)
+endif
 CLUSTER_CONTROLLER_IMAGE_NAME=eks-anywhere-cluster-controller
 CLUSTER_CONTROLLER_IMAGE=$(IMAGE_REPO)/$(CLUSTER_CONTROLLER_IMAGE_NAME):$(IMAGE_TAG)
 CLUSTER_CONTROLLER_LATEST_IMAGE=$(IMAGE_REPO)/$(CLUSTER_CONTROLLER_IMAGE_NAME):$(LATEST)
+IMAGES_TO_BUILD?=$(CLUSTER_CONTROLLER_IMAGE),$(CLUSTER_CONTROLLER_LATEST_IMAGE)
 
 # Branch builds should look at the current branch latest image for cache as well as main branch latest for cache to cover the cases
 # where its the first build from a new release branch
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -385,8 +392,9 @@ capd-test: ## Run default e2e capd test locally
 	$(MAKE) local-e2e LOCAL_E2E_TESTS=$(DOCKER_E2E_TEST)
 
 .PHONY: docker-e2e-test
-docker-e2e-test: e2e ## Run docker integration test in new ec2 instance
+docker-e2e-test: release-cluster-controller e2e ## Run docker integration test in new ec2 instance
 	scripts/e2e_test_docker.sh $(DOCKER_E2E_TEST) $(BRANCH_NAME)
+	aws ecr-public batch-delete-image --repository-name eks-anywhere-cluster-controller --image-ids=imageTag=$(PULL_PULL_SHA).$(PULL_BASE_REF) --region us-east-1
```
jaxesn, line 397: can we make this always run, regardless of failure?

Makefile
```diff
@@ -91,9 +91,16 @@ DOCKERFILE_FOLDER = ./manager/docker/linux/eks-anywhere-cluster-controller
 
 IMAGE_REPO?=$(if $(AWS_ACCOUNT_ID),$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com,localhost:5000)
 IMAGE_TAG?=$(GIT_TAG)-$(shell git rev-parse HEAD)
+# Tagging with the commit hash and the branch the PR is raised against
+ifeq ($(JOB_NAME),eks-anywhere-e2e-presubmit)
```
jaxesn, line 95: i dont love it, but making these env vars in the job doesnt make a lot of sense either since the pull_sha will have to be like this anyway

Makefile
```diff
@@ -505,12 +513,11 @@ eks-a-e2e:
 			make eks-a-release GIT_VERSION=$(DEV_GIT_VERSION); \
 			scripts/get_bundle.sh; \
 		else \
-			make check-eksa-components-override; \
```
jaxesn, line 508: never going to be an override here?

Makefile
abhay-krishna: That was my original thought and I'm willing to go either way, but I tested it from the Makefile because doing it in the prowjob would require another PR and approval.

Makefile
abhay-krishna: That override was solely for the eks-a components not being in S3 at the time of running e2e tests. I figured that now since we always generate the bundle before running tests, we should have it in S3 so won't need overrides.

Makefile
abhay-krishna: I'll put it in the trap

jaxesn (approved):

scripts/e2e_test_docker.sh
```diff
@@ -18,6 +18,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+delete_cluster_controller_image() {
+    aws ecr-public batch-delete-image --repository-name eks-anywhere-cluster-controller --image-ids=imageTag=${PULL_PULL_SHA}.${PULL_BASE_REF} --region us-east-1
```
vivek-koppuru, line 22: So the latest will never be cleaned up, it will always be based on any PR open right?

Makefile
```diff
@@ -91,9 +91,16 @@ DOCKERFILE_FOLDER = ./manager/docker/linux/eks-anywhere-cluster-controller
 
 IMAGE_REPO?=$(if $(AWS_ACCOUNT_ID),$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com,localhost:5000)
 IMAGE_TAG?=$(GIT_TAG)-$(shell git rev-parse HEAD)
+# Tagging with the commit hash and the branch the PR is raised against
+ifeq ($(JOB_NAME),eks-anywhere-e2e-presubmit)
+	IMAGE_PLATFORMS=linux/amd64
+	IMAGE_TAG=$(PULL_PULL_SHA).$(PULL_BASE_REF)
+	IMAGES_TO_BUILD?=$(CLUSTER_CONTROLLER_IMAGE)
```
vivek-koppuru, line 98: Oh so for e2e, we are not pushing the latest?

vivek-koppuru (approved):

Makefile
abhay-krishna: Yes. We're pushing only one tagged with a separate tagging format COMMIT.BRANCH and then deleting it at the end.

PR #3189, opened 2022-08-26
Improved validation of kubeconfig files

ewollesen:
Instead of just checking that the file exists, the k8s client-go library is
used to load the file's data, which validates its basic structure, as well as
its existence.

This small step is a precursor to further work to concentrate how kubeconfig
files are handled across the various eksctl-anywhere subcommands.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -21,12 +21,9 @@ type deleteClusterOptions struct {
 	forceCleanup          bool
 	hardwareFileName      string
 	tinkerbellBootstrapIP string
-	kubeconfigValidator   kubeconfig.Validator
 }
 
-var dc = &deleteClusterOptions{
-	kubeconfigValidator: kubeconfig.NewValidator(),
-}
+var dc = &deleteClusterOptions{}
 
 var deleteClusterCmd = &cobra.Command{
 	Use:          "cluster (<cluster-name>|-f <config-file>)",
@@ -72,7 +69,7 @@ func (dc *deleteClusterOptions) validate(ctx context.Context, args []string) err
 	}
 
 	kubeconfigPath := getKubeconfigPath(clusterConfig.Name, dc.wConfig)
-	if err := dc.kubeconfigValidator.ValidateFile(kubeconfigPath); err != nil {
+	if err := kubeconfig.ValidateFile(kubeconfigPath); err != nil {
 		return err
 	}
 
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -69,8 +69,8 @@ func (dc *deleteClusterOptions) validate(ctx context.Context, args []string) err
 	}
 
 	kubeconfigPath := getKubeconfigPath(clusterConfig.Name, dc.wConfig)
-	if !validations.FileExistsAndIsNotEmpty(kubeconfigPath) {
-		return kubeconfig.NewMissingFileError(kubeconfigPath)
+	if err := validations.KubeConfigFile(kubeconfigPath); err != nil {
+		return err
 	}
 
 	return nil
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -21,9 +21,12 @@ type deleteClusterOptions struct {
 	forceCleanup          bool
 	hardwareFileName      string
 	tinkerbellBootstrapIP string
+	kubeconfigValidator   kubeconfig.Validator
 }
 
-var dc = &deleteClusterOptions{}
+var dc = &deleteClusterOptions{
+	kubeconfigValidator: kubeconfig.NewValidator(),
+}
 
 var deleteClusterCmd = &cobra.Command{
 	Use:          "cluster (<cluster-name>|-f <config-file>)",
@@ -69,7 +72,7 @@ func (dc *deleteClusterOptions) validate(ctx context.Context, args []string) err
 	}
 
 	kubeconfigPath := getKubeconfigPath(clusterConfig.Name, dc.wConfig)
-	if err := validations.KubeConfigFile(kubeconfigPath); err != nil {
+	if err := dc.kubeconfigValidator.ValidateFile(kubeconfigPath); err != nil {
 		return err
 	}
 
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -2,13 +2,13 @@ package kubeconfig
 
 import (
 	"fmt"
-	"io"
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path/filepath"
 
 	"k8s.io/client-go/tools/clientcmd"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
 // FromClusterFormat defines the format of the kubeconfig of the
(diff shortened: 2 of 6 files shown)


cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -82,8 +82,8 @@ func (dc *deleteClusterOptions) validate(ctx context.Context, args []string) err
 	}
 
 	kubeconfigPath := getKubeconfigPath(clusterConfig.Name, dc.wConfig)
-	if !validations.FileExistsAndIsNotEmpty(kubeconfigPath) {
-		return kubeconfig.NewMissingFileError(kubeconfigPath)
+	if err := validations.KubeConfigFile(kubeconfigPath); err != nil {
```
a-cool-train, line 85: This still validates whether the data contained in the file matches a kubeconfig structure. We should probably name the method as such? validations.KubeConfigFile is confusing as to what this is supposed to perform.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -37,3 +41,40 @@ func NewMissingFileError(path string) error {
...
+
+	return nil
+}
+
+// validator can be overridden for easier testing.
+var validator = func(b []byte) error { _, err := clientcmd.Load(b); return err }
+
+// ValidateFile is a helper for Validate.
+func ValidateFile(filename string) error {
+	f, err := os.Open(filename)
```
a-cool-train, line 69: Should we defer close right after opening?

cmd/eksctl-anywhere/cmd/deletecluster.go
ewollesen: I'm confused as to what you're confused about? The function clearly states that it's validating the kubeconfig file. If you want to know exactly what that entails, it's very easy to look at the godoc for the function (your editor probably displays is somewhere either on mouseover, or via a handy keybinding) and if that fails, you have the source.

Putting the specific validations in the name gets unwieldy very quickly. For example, this function would be named: ValidateFileExistsAndIsNotEmptyAndHasValidYAMLKubeconfigStructure or something like that.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -37,3 +41,40 @@ func NewMissingFileError(path string) error {
...
+	// much. :(
+	if err := validator(kubeConfigData); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validator can be overridden for easier testing.
+var validator = func(b []byte) error { _, err := clientcmd.Load(b); return err }
```
g-gaston, line 65: nit: maybe overkill but maybe pass the validation as an option to the validate method? just trying to avoid global state

cmd/eksctl-anywhere/cmd/deletecluster.go
g-gaston, line 85: nit: I feel like I would just call `kubeconfig.ValidateFile` here directly and avoid that var in `validations`
Tbh, `validations` as a package sounds too broad (unless it's for validation tooling or something like that). I like more your idea of having the validation on an specific package for that domain

g-gaston (approved):

pkg/kubeconfig/kubeconfig.go
ewollesen: Lemme look at this and see what makes sense. I def. don't love the package-level scope either.

cmd/eksctl-anywhere/cmd/deletecluster.go
ewollesen: I'd be glad to do that. I did it the way I did to try to maintain the pattern, cuz I didn't know a whole lot about the validation package.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -25,15 +29,46 @@ func FromEnvironment() string {
...
-type missingFileError struct {
-	Path string
+type Validator interface {
+	ValidateFile(filename string) error
+	Validate(r io.Reader) error
+}
+
+type validator struct {
+	load func([]byte) (*clientcmdapi.Config, error)
+}
```
chrisdoherty4, line 39: Have we considered modelling this as just a function (`func Validate(...)`)? This feels quite heavy and I'm not sure we really need an object at consumer sites.

ewollesen: Oh boy, have we! Let me tell you, the permutations we've been through.

I tend to agree, that this is all quite over-engineered at this point. I'd say I don't know how it got this far, but that's not true, cuz I was there for every step.

So, to your question, I did have a Validate function for a while. However, in order to test it, I had to be able to mock out the call to clientcmd.Load(), and to do that with a function, I had to use a package-level variable. This wasn't ideal, as it wasn't thread-safe. Nevermind that every one of the EKS-A CLI commands does the same (is that normal with cobra.Command's, and if so, why does anyone use that package?)

So to try to remove the package-level variable, I added a parameter to Validate that let one specify a validator to use. That led to a discussion around the use of nil vs variadic parameters, which lead to a bunch of interfaces and then eventually ended up ... here.

I guess a lot of that discussion ended up happening over Slack for speed's sake. It might have been interesting if it had been captured in this PR. Well, interesting in some weird academic meets train wreck sort way.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -25,15 +30,90 @@ func FromEnvironment() string {
 	return os.Getenv(EnvName)
 }
 
-type missingFileError struct {
-	Path string
+// Loader abstracts the unmarshaling of a kubeconfig file into a Config.
+//
+// This is essentially building an interface around the clientcmd.Load
+// function so that it can be swapped out for tests.
+type Loader interface {
```
g-gaston, line 37: Crazy annoying idea, I wonder if we can just omit the concept of loader? I get why this is useful for testing, so we don't need to depend on the `client-go` implementation logic. But would it be that bad if we do?

It shouldn't be that difficult to pass a valid/invalid kubeconfig body. And I think it would simplify this code a lot.

pkg/kubeconfig/kubeconfig.go
chrisdoherty4: Isn't loading the configuration the crux of validation? Without it, any testing seems rather redundant.

I think we should go back to the beginning of the road and just have a `func Validate(io.Reader) error`. Creating invalid configurations should be trivial as would generating an io error and we'd actually test that a `kubeconfig.Validate(reader)` functions as expected. Given this is generally serving main type functions that are operating on path strings I would even suggest `Validate(path string) error` and only introduce the `io.Reader` when its needed. Using the path simplifies our function further to just `clientcmd.LoadFromFile(...)`.

I do empathize with a desire to push the `io` package but given the only consumer is a main type func right now it doesn't seem appropriate in this case.

Sorry to step in last minute here, but I don't think what we have in the current iteration is the right approach.

pkg/kubeconfig/kubeconfig.go
ewollesen: It seems that both you and Chris have had the same idea here, and it sounds like a good one to me. I'll whip something up.

pkg/kubeconfig/kubeconfig.go
ewollesen: Yeah, that seems like a good way to remove a fair bit of cruft.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -25,15 +29,26 @@ func FromEnvironment() string {
...
 }
 
-type missingFileError struct {
-	Path string
-}
+// ValidateFile loads a file to validate it's basic contents.
+//
+// The values of the fields within aren't validated, but the file's existence
+// and basic structure are checked.
+func ValidateFile(filename string) error {
```
chrisdoherty4, line 36: Do we want to simplify this to just `Validate`?

chrisdoherty4 (approved):

pkg/kubeconfig/kubeconfig.go
ewollesen: What can I say, I want to keep the hope alive for a reader based implementation some day. :)

pkg/kubeconfig/kubeconfig_test.go
```diff
@@ -0,0 +1,96 @@
+package kubeconfig
```
g-gaston, line 1: nit: `kubeconfig_test`

g-gaston (approved):

PR #3192, opened 2022-08-29
Reconcile CNI in snow cluster reconciler

g-gaston:
*Testing (if applicable):*
Unit tests

### DIFF
--- a/controllers/factory.go
+++ b/controllers/factory.go
@@ -213,14 +213,18 @@ func (f *Factory) withVSphereClusterReconciler() *Factory {
 }
 
 func (f *Factory) withSnowClusterReconciler() *Factory {
-	f.withCNIReconciler()
+	f.withCNIReconciler().withTracker()
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		if f.snowClusterReconciler != nil {
 			return nil
 		}
 
-		f.snowClusterReconciler = snowreconciler.New(f.manager.GetClient(), f.cniReconciler)
+		f.snowClusterReconciler = snowreconciler.New(
+			f.manager.GetClient(),
+			f.cniReconciler,
+			f.tracker,
+		)
 		f.registryBuilder.Add(anywherev1.SnowDatacenterKind, f.snowClusterReconciler)
 
 		return nil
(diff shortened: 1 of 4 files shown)


controllers/factory.go
```diff
@@ -213,14 +213,18 @@ func (f *Factory) withVSphereClusterReconciler() *Factory {
 }
 
 func (f *Factory) withSnowClusterReconciler() *Factory {
-	f.withCNIReconciler()
+	f.withCNIReconciler().withTracker()
```
vivek-koppuru, line 216: What is this?

g-gaston: tracker? You can see it here https://github.com/aws/eks-anywhere/blob/079ca2f5b838e81b9f7446519b0e895cc3205bf1/controllers/factory.go#L132-L155

It builds a remote client tracker. It builds kube clients for workload clusters from kubeconfigs stored in secrets.

mitalipaygude (approved):

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -75,11 +104,10 @@ func (s *Reconciler) ReconcileWorkers(ctx context.Context, log logr.Logger, clus
...
 }
 
-func (s *Reconciler) ReconcileControlPlane(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec) (controller.Result, error) {
-	log = log.WithValues("phase", "reconcileControlPlane")
-	log.Info("Applying control plane CAPI objects")
-
-	return s.Apply(ctx, func() ([]kubernetes.Object, error) {
-		return snow.ControlPlaneObjects(ctx, clusterSpec, clientutil.NewKubeClient(s.client))
-	})
+func capiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
```
taneyland, line 107: Should we also define this in one of the capi files that you are introducing in #3193? 
Maybe in `pkg/controller/clusters/clusterapi.go`

g-gaston: we probably should yeah
what about we make this a follow up? I don't have time right now to move it and I want to start the testing as soon as possible. Or maybe you can move the method and add a test? And just push to this branch :)

g-gaston: Talked about it offline, we will have a follow up PR

Issue #3195, opened 2022-08-29
gitops FluxConfig does not support private git ssh urls


dashkan:
**What happened**:
Updating a cluster with gitops FluxConfig does not work if using a private ssh url such as:
ssh://git@github.com:USER/repo.git

Get the following error:
Error: failed to display upgrade plan: unable to get cluster config from file: invalid cluster config: unable to parse repository url: parse "ssh://git@github.com:USER/repo.git": invalid port ":USER" after host

Please not that i replaced USER and repo name.

**What you expected to happen**:
Should work with any valid git ssh url

**How to reproduce it (as minimally and precisely as possible)**:

Add gitops section and reference section in cluster config:

```yaml
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: Cluster
metadata:
  name: management
spec:
  gitOpsRef:
    kind: FluxConfig
    name: gitops

---
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: FluxConfig
metadata:
  name: gitops
spec:
    git:
      repositoryUrl: ssh://git@github.com:USER/repo.git
      sshKeyAlgorithm: ecdsa

```

**Anything else we need to know?**:
https://github.com/aws/eks-anywhere/blob/079ca2f5b838e81b9f7446519b0e895cc3205bf1/pkg/api/v1alpha1/fluxconfig.go#L3-L9

https://github.com/aws/eks-anywhere/blob/079ca2f5b838e81b9f7446519b0e895cc3205bf1/pkg/api/v1alpha1/fluxconfig.go#L77-L86

Looks like code uses net/url which tries to parse the USER portion of the git ssh url as a port.

**Environment**:
- EKS Anywhere Release: v0.11.1
- EKS Distro Release: 0.109.0

dashkan:
Private repo can be specified using the following format ssh://git@github.com/ORG/repo.git

PR #3197, opened 2022-08-30
Allow Private SSH Git Urls using FluxConfig

dashkan:
*Issue #, if available:*
#3195 

*Description of changes:*
Allows private ssh git urls

*Testing (if applicable):*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tinkerbell/rufio v0.0.0-20220606134123-599b7401b5cc
 	github.com/tinkerbell/tink v0.6.1-0.20220509141453-30fe9e015575
+	github.com/whilp/git-urls v1.0.0
 	go.uber.org/zap v1.19.1
 	golang.org/x/crypto v0.0.0-20220517005047-85d78b3ac167
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8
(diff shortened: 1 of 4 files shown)


jaxesn:
Thanks for the submission! Well make sure someone reviews

dashkan:
This patch fails to bootstrap gitops because it cannot validate the git ssh url. I need to figure out how to test this end to end. Any guidance is appreciated as this is holding me back from using gitops.

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tinkerbell/rufio v0.0.0-20220606134123-599b7401b5cc
 	github.com/tinkerbell/tink v0.6.1-0.20220509141453-30fe9e015575
+	github.com/whilp/git-urls v1.0.0
```
chrisdoherty4, line 32: Thanks for the contribution. Are there any alternatives to this lib - it doesn't seem to be maintained? We could also consider copying the functions we need into our repository.

dashkan: How about a regex to validate the ssh url?

dashkan:
Private repo can be provided using the following url format ssh://git@github.com/ORG/repo.git

danbudris:
> 

Hey there @dashkan 

Is there anything we could do to improve the docs to make sure that the URL format is apparent for EKS A users?

dashkan:
> > 
> 
> Hey there @dashkan
> 
> Is there anything we could do to improve the docs to make sure that the URL format is apparent for EKS A users?

I think a lot of people will go to their repo click the copy button add prefix with ssh://, which is a valid format. I think it would help to put an example of expected ssh url format for public and private repos.

PR #3198, opened 2022-08-30
Bugfix download images

a-cool-train:
*Description of changes:*
- Currently Curated Packages images are not getting fetched. This could potentially cause an issue when creating a cluster during disconnected scenarios
- This PR includes the extraction of images from curated packages bundles.
- Temporary Solution: Currently private curated packages don't have a root level project hence this PR adds a root path to the images of the packages
- Currently the images of curated packages are referenced through digest, this PR removes the @ included in the digest
- Temporary Solution: Currently curated packages are in a private account, this PR adds constants to use for getting the images from the private account

*Testing (if applicable):*


jaxesn (commented):
These are for the curated packages images themselves? Will this only trigger if the user is running with a valid license?  I just want to make sure we arent breaking import/download images for uses without enterprise and therefore no access to the private ecr repos

pkg/docker/registry.go
```diff
@@ -3,10 +3,21 @@ package docker
...
 )
 
+// These constants are temporary since currently there is a limitation on harbor
+// Harbor requires root level projects but curated packages private account currently
+// doesn't have support for root level
+const (
+	packageProdLocation = "783794618700.dkr.ecr.us-west-2.amazonaws.com"
+	packageDevLocation  = "857151390494.dkr.ecr.us-west-2.amazonaws.com"
+	publicProdECRName   = "eks-anywhere"
+	publicDevECRName    = "l0g8r8j6"
```
g-gaston, line 18: could we just use the same "folder" for both prod and dev? I would vote for `eks-anywhere`

pkg/executables/docker.go
```diff
@@ -80,7 +84,9 @@ func (d *Docker) CgroupVersion(ctx context.Context) (int, error) {
 }
 
 func (d *Docker) TagImage(ctx context.Context, image string, endpoint string) error {
-	localImage := strings.ReplaceAll(image, defaultRegistry, endpoint)
+	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdLocation, endpoint, packageDevLocation, endpoint)
+	localImage := replacer.Replace(image)
+	localImage = removeDigestReference(localImage)
```
g-gaston, line 89: I would vote to do this in the registry mover

pkg/executables/docker.go
```diff
@@ -80,7 +84,9 @@ func (d *Docker) CgroupVersion(ctx context.Context) (int, error) {
 }
 
 func (d *Docker) TagImage(ctx context.Context, image string, endpoint string) error {
-	localImage := strings.ReplaceAll(image, defaultRegistry, endpoint)
+	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdLocation, endpoint, packageDevLocation, endpoint)
```
g-gaston, line 87: Can we do this in the registry mover as well? That place already has knowledge about the different registries and it's doing some host/path manipulation

pkg/executables/docker.go
```diff
@@ -89,7 +95,9 @@ func (d *Docker) TagImage(ctx context.Context, image string, endpoint string) er
 }
 
 func (d *Docker) PushImage(ctx context.Context, image string, endpoint string) error {
-	localImage := strings.ReplaceAll(image, defaultRegistry, endpoint)
+	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdLocation, endpoint, packageDevLocation, endpoint)
+	localImage := replacer.Replace(image)
+	localImage = removeDigestReference(localImage)
```
g-gaston, line 100: Same, move this to the registry mover

pkg/executables/docker.go
```diff
@@ -89,7 +95,9 @@ func (d *Docker) TagImage(ctx context.Context, image string, endpoint string) er
 }
 
 func (d *Docker) PushImage(ctx context.Context, image string, endpoint string) error {
-	localImage := strings.ReplaceAll(image, defaultRegistry, endpoint)
+	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdLocation, endpoint, packageDevLocation, endpoint)
```
g-gaston, line 98: same comment about doing this in the registry mover

pkg/curatedpackages/reader.go
```diff
@@ -3,53 +3,79 @@ package curatedpackages
...
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// Temporary: Curated packages dev and prod accounts are currently hard coded
+// This is because there is no mechanism to extract these values as of now
 const (
-	defaultRegistry = "783794618700.dkr.ecr.us-west-2.amazonaws.com"
+	publicProdECR       = "public.ecr.aws/eks-anywhere"
+	packageProdLocation = "783794618700.dkr.ecr.us-west-2.amazonaws.com"
+	packageDevLocation  = "857151390494.dkr.ecr.us-west-2.amazonaws.com"
```
g-gaston, line 20: for this two, maybe `package[Prod/Dev]Domain`?

pkg/curatedpackages/reader.go
```diff
@@ -61,14 +87,51 @@ func fetchPackages(ctx context.Context, versionsBundle releasev1.VersionsBundle,
...
 	images := make([]releasev1.Image, 0, len(bundle.Spec.Packages))
 	for _, p := range bundle.Spec.Packages {
-		pI := releasev1.Image{
+		pHC := releasev1.Image{
 			Name:        p.Name,
 			Description: p.Name,
 			OS:          ctrl.OS,
 			OSName:      ctrl.OSName,
-			URI:         fmt.Sprintf("%s/%s:%s", defaultRegistry, p.Source.Repository, p.Source.Versions[0].Name),
+			URI:         fmt.Sprintf("%s/%s:%s", GetRegistry(ctrl.URI), p.Source.Repository, p.Source.Versions[0].Name),
```
g-gaston, line 95: I got lost here. If we are doing some domain manipulation in the image mover, why is this necessary here?

pkg/curatedpackages/reader.go
```diff
@@ -61,14 +87,51 @@ func fetchPackages(ctx context.Context, versionsBundle releasev1.VersionsBundle,
...
+
+	for _, p := range bundle.Spec.Packages {
+		// each package will have at least one version
+		for _, version := range p.Source.Versions[0].Images {
+			image := releasev1.Image{
+				Name:        version.Repository,
+				Description: version.Repository,
+				OS:          ctrl.OS,
+				OSName:      ctrl.OSName,
+				URI:         fmt.Sprintf("%s/%s@%s", getRegistry(ctrl.URI), version.Repository, version.Digest),
```
g-gaston, line 124: Same question here about why we do this here in addition to what we do in the image mover

pkg/manifests/reader.go
```diff
@@ -80,7 +80,7 @@ func (r *Reader) ReadImages(eksaVersion string) ([]releasev1.Image, error) {
 	return bundles.ReadImages(r, bundle)
 }
 
-func (r *Reader) ReadImagesFromBundles(b *releasev1.Bundles) ([]releasev1.Image, error) {
+func (r *Reader) ReadImagesFromBundles(ctx context.Context, b *releasev1.Bundles) ([]releasev1.Image, error) {
```
g-gaston, line 83: nit:
```suggestion
func (r *Reader) ReadImagesFromBundles(_ context.Context, b *releasev1.Bundles) ([]releasev1.Image, error) {
```

pkg/curatedpackages/reader.go
a-cool-train: This is not doing any manipulation. It just extracts the image uri

pkg/docker/registry.go
a-cool-train: This will not be ideal because for dev, the controller will be using l0gr... to fetch images, and if we make this change, it will break the pull of images.

pkg/curatedpackages/reader.go
a-cool-train: I responded above but this doesn't perform any manipulation, it merely just extracts the images.

pkg/executables/docker.go
a-cool-train: I don't see any host/path manipulation in the registry mover.

pkg/executables/docker.go
a-cool-train: Added an issue for this: https://github.com/aws/eks-anywhere/issues/3206

pkg/executables/docker.go
a-cool-train: Added an issue for this: https://github.com/aws/eks-anywhere/issues/3206

PR #3201, opened 2022-08-30
add e2e cli option to set the branch name for baremetal tests to run on  @gwesterfieldjr

gwesterfieldjr:
add e2e cli option to set the branch name for baremetal tests to run on

### DIFF
--- a/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
+++ b/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
@@ -98,6 +98,7 @@ env:
     T_TINKERBELL_SSH_AUTHORIZED_KEY: "vsphere_ci_beta_connection:ssh_authorized_key"
     T_TINKERBELL_CP_NETWORK_CIDR: "tinkerbell_ci:cp_network_cidr"
     T_TINKERBELL_S3_INVENTORY_CSV_KEY: "tinkerbell_ci:s3_inventory_csv"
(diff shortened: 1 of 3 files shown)


ewollesen:
Adding flags for each thing you want to enable/disable doesn't seem sustainable in the long-term. Have you given any thought to what kind of system could be designed that might allow for any given set of criteria to be applied to tests to enable or disable them?

gwesterfieldjr:
> Adding flags for each thing you want to enable/disable doesn't seem sustainable in the long-term. Have you given any thought to what kind of system could be designed that might allow for any given set of criteria to be applied to tests to enable or disable them?

Yes, the sustainable solution is to either have a hardware reservation system (assuming we have enough hardware to handle the load of multiple builds running simultaneously otherwise all builds will be backed up waiting for reservations) or a separate hardware bucket for release builds to utilize for tinkerbell e2e tests.

For now, I'm adding this option and updating the release runbook doc to note that we need to disable bare metal tests on main to avoid conflicting tests runs in upcoming releases so it does not get missed as it did in previous release.

Once I'm able to add our HP machines to the CI pool I'll have enough hardware to provide release builds their own pool, but I can't do this until BR is fixed on HP. This will be a follow up PR to further improve this process, but for now this is an improvement to the current process which is tribal knowledge.

abhinavmpandey08 (approved):

pokearu (approved):
/approve

PR #3205, opened 2022-08-30
Move capi obj key method to controller capi file

taneyland:
*Issue #, if available:*

*Description of changes:*
Follow up from this [discussion](https://github.com/aws/eks-anywhere/pull/3192#discussion_r957774271)

*Testing (if applicable):*

### DIFF
--- a/pkg/controller/clusterapi.go
+++ b/pkg/controller/clusterapi.go
@@ -31,3 +31,13 @@ func GetCAPICluster(ctx context.Context, client client.Client, cluster *anywhere
 	}
 	return capiCluster, nil
 }
+
+// CapiClusterObjectKey generates an ObjectKey for the CAPI cluster owned by
+// the provided eks-a cluster
+func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
+	// TODO: we should consider storing a reference to the CAPI cluster in the eksa cluster status
+	return client.ObjectKey{
+		Name:      clusterapi.ClusterName(cluster),
+		Namespace: constants.EksaSystemNamespace,
+	}
+}
--- a/pkg/controller/clusterapi_test.go
+++ b/pkg/controller/clusterapi_test.go
@@ -7,6 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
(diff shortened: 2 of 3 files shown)


pkg/controller/clusters/clusterapi.go
```diff
@@ -37,3 +38,11 @@ func CheckControlPlaneReady(ctx context.Context, client client.Client, log logr.
 	log.Info("CAPI control plane is ready")
 	return controller.Result{}, nil
 }
+
+func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
```
g-gaston, line 42: nit: add doc comment?
```suggestion
// CapiClusterObjectKey generates and ObjectKey for the CAPI cluster owned by
// the provided eks-a cluster
func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
```

pkg/controller/clusters/clusterapi_test.go
```diff
@@ -94,6 +96,19 @@ func eksaCluster() *anywherev1.Cluster {
...
 }
 
+func TestGetCapiClusterObjectKey(t *testing.T) {
+	g := NewWithT(t)
+	eksaCluster := eksaCluster()
+
+	expected := types.NamespacedName{
+		Name:      clusterapi.ClusterName(eksaCluster),
+		Namespace: constants.EksaSystemNamespace,
+	}
```
g-gaston, line 106: IMO we shouldn't use the same code being tested to generate the expected results

```suggestion
	expected := types.NamespacedName{
		Name:      "my-cluster",
		Namespace: "eksa-system",
	}
```

pkg/controller/clusters/clusterapi.go
g-gaston, line 42: I would maybe move this to `pkg/controller/clusterapi.go`
It seems something closer to what we have there, I think.

taneyland: Makes sense. I was on the fence on which file to use so I definitely see your point, updated it!

pkg/controller/clusterapi.go
```diff
@@ -31,3 +31,13 @@ func GetCAPICluster(ctx context.Context, client client.Client, cluster *anywhere
 	}
 	return capiCluster, nil
 }
+
+// CapiClusterObjectKey generates an ObjectKey for the CAPI cluster owned by
+// the provided eks-a cluster
+func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
```
g-gaston, line 37: sorry i missed this
```suggestion
// CAPIClusterObjectKey generates an ObjectKey for the CAPI cluster owned by
// the provided eks-a cluster
func CAPIClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
```

g-gaston (approved):

PR #3215, opened 2022-08-31
Fix GitHub upgrade tests

danbudris:
*Issue #, if available:*

*Description of changes:*
alright, let's see if this addresses the issue with the name overrides -- we're adding an option to explicitly set the name in the FluxConfig, so that it uses the same name as we're expecting in the upgrade filler.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/flux_test.go
+++ b/test/e2e/flux_test.go
@@ -311,8 +311,8 @@ func TestDockerInstallGithubFluxDuringUpgrade(t *testing.T) {
 	runUpgradeFlowWithFlux(
 		test,
 		v1alpha1.Kube122,
-		framework.WithFluxGithub(),
-		framework.WithClusterUpgrade(),
+		framework.WithFluxGithub(api.WithFluxConfigName(framework.DefaultFluxConfigName)),
+		framework.WithClusterUpgrade(api.WithGitOpsRef(framework.DefaultFluxConfigName, v1alpha1.FluxConfigKind)),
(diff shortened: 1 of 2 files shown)


jaxesn (approved):

test/e2e/flux_test.go
```diff
@@ -311,8 +311,8 @@ func TestDockerInstallGithubFluxDuringUpgrade(t *testing.T) {
 	runUpgradeFlowWithFlux(
 		test,
 		v1alpha1.Kube122,
-		framework.WithFluxGithub(),
-		framework.WithClusterUpgrade(),
+		framework.WithFluxGithub(api.WithFluxConfigName(framework.DefaultFluxConfigName)),
+		framework.WithClusterUpgrade(api.WithGitOpsRef(framework.DefaultFluxConfigName, v1alpha1.FluxConfigKind)),
```
g-gaston, line 315: why not implement this in `WithFluxGithub`?

g-gaston (approved):

test/e2e/flux_test.go
danbudris: yeah, that would be ideal. However, at the moment, the WithClusterUpgrade seems to override the GitOpsRef name with its own fillers derived from the initial cluster spec. So this is an attempt to verify that is what's happening and get the names aligned, so that we don't end up with a GitOpsRef that points to nothing

PR #3216, opened 2022-08-31
GOVC Business Logic for vSphere Setup User Command

jonathanmeier5:
*Issue #, if available:*

*Description of changes:*
Adds govc business logic required to support setting up a vsphere user.

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/govc.go
+++ b/pkg/executables/govc.go
@@ -973,11 +973,9 @@ func getDeployOptions(network string) ([]byte, error) {
 	return deployOpts, err
 }
 
-func (g *Govc) CreateUser(ctx context.Context, username string) error {
+func (g *Govc) CreateUser(ctx context.Context, username string, password string) error {
 	params := []string{
-		"sso.user.create",
-		"-p", "foobar",
-		username,
+		"sso.user.create", "-p", password, username,
 	}
 
 	if _, err := g.exec(ctx, params...); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/executables/govc.go
```diff
@@ -972,3 +972,95 @@ func getDeployOptions(network string) ([]byte, error) {
 
 	return deployOpts, err
 }
+
+func (g *Govc) CreateUser(ctx context.Context, username string, password string) error {
```
chrisdoherty4, line 976: We need function comments on all these even if they are simple. `AddGroupUser` for example feels a little odd. I had to look up what `sso.group.update` does which suggests the name isn't clear with its intent.

chrisdoherty4 (changes requested):

pkg/executables/govc.go
chrisdoherty4: Sorry, I should've been more elaborate. There's a specific format to Go comments too. This ensures tooling such as `go doc` can render code comments and it provides a really handy wiki for understanding the code base over the CLI. Tools like `godoc` (the web server version) also parse and publish a website similar to `pkg.go.dev`. 

https://tip.golang.org/doc/comment

chrisdoherty4: Heres an example that I wrote today for reference (this code may or may not get merged, eventually) https://github.com/aws/eks-anywhere/pull/3246/files

jonathanmeier5: Ah, thanks! I've updated the docstrings to match the style that I *think* these examples/docs are demonstrating.

chrisdoherty4 (approved):
lgtm

PR #3218, opened 2022-08-31
Update the generated go code for pkg/api/v1alpha1

thunderboltsid:
PR #3181 updated the CloudStack types but did not trigger `make generate`
to update the `zz_generated.deepcopy.go`. This causes a few unit tests
to fail when others try to regenerate the file. This PR updates the generated
code and fixes the broken testcases for CloudStackMachineConfig.

### DIFF
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_test.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_test.go
@@ -312,40 +312,40 @@ func TestCloudStackMachineNotEqualDiskOfferingName(t *testing.T) {
 func TestCloudStackMachineNotEqualDiskOfferingId(t *testing.T) {
 	g := NewWithT(t)
 	cloudStackMachineConfigSpec2 := cloudStackMachineConfigSpec1.DeepCopy()
-	cloudStackMachineConfigSpec2.DiskOffering.Id = "newDiskOffering"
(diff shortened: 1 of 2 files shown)


abhinavmpandey08 (approved):

pkg/api/v1alpha1/cloudstackmachineconfig_test.go
```diff
@@ -312,40 +312,40 @@ func TestCloudStackMachineNotEqualDiskOfferingName(t *testing.T) {
 func TestCloudStackMachineNotEqualDiskOfferingId(t *testing.T) {
 	g := NewWithT(t)
 	cloudStackMachineConfigSpec2 := cloudStackMachineConfigSpec1.DeepCopy()
-	cloudStackMachineConfigSpec2.DiskOffering.Id = "newDiskOffering"
 	cloudStackMachineConfigSpec2.DiskOffering = (*cloudStackMachineConfigSpec1.DiskOffering).DeepCopy()
```
pokearu, line 315: Isn't this line redundant everywhere?  `cloudStackMachineConfigSpec1.DeepCopy()` would already handle this right? Since the `zz_generated` file was updated correctly.

thunderboltsid: Indeed, the line was being redundant as the test is comparing a deepcopy, replacing a value, and expecting the comparison to result in false. This would only work if the line was moved after the deepcopy. If the line wasn't moved after deepcopy, the comparison would ofc have resulted in True.

PR #3220, opened 2022-08-31
Remove all references to AWS bundle

g-gaston:
*Description of changes:*
We made AWS a pointer and we don't include it in the bundle manifest
anymore, which means that all the code the references that field (while
assuming it was a struct a not a pointer) blows up when with a nil
pointer when facing the new bundle.


pokearu (approved):
/lgtm

PR #3222, opened 2022-08-31
cp e2e: Define constants where they're used.

ewollesen:
A short term goal of the CP team is to have our E2E tests run against the
current builds of our packages. These hard-coded registries are preventing
that, so we're planning to replace them. As a first step towards that,
separating these variables will allow for smaller-step refactoring towards
that goal.


vincentni (approved):
/lgtm
/approve

PR #3236, opened 2022-09-01
Helm chart release fixes

jonahjon:
*Issue #, if available:*

*Description of changes:*

Adds change to do 2 Things

1. Modify Helm Chart.yaml version to be what the destination Tag will be be, re-package, and push to destination repository.
2. Override the Image Digest references to the images digests inside the helm chart so they will always be in sync.


release/pkg/aws/ecr/ecr.go
```diff
@@ -59,6 +61,16 @@ func GetAuthToken(ecrClient *ecr.ECR) (string, error) {
 	return authToken, nil
 }
 
+func GetAuthTokenPublic(ecrClient *ecrpublic.ECRPublic) (string, error) {
```
abhay-krishna, line 64: We already have this method in the ecrpublic package [here](https://github.com/aws/eks-anywhere/blob/main/release/pkg/aws/ecrpublic/ecrpublic.go#L50).

jonahjon: O ok I'll update that

PR #3239, opened 2022-09-01
Set bundles refs when it's not specified

wongni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -63,6 +63,13 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 	if err != nil {
 		return err
 	}
+
+	if cs.Spec.BundlesRef == nil {
+		if err := cor.setBundlesRef(ctx, cs); err != nil {
+			return err
+		}
+	}
+
 	spec, err := cor.FetchAppliedSpec(ctx, cs)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


controllers/resource/reconciler.go
```diff
@@ -164,6 +171,16 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 	return cor.applyTemplates(ctx, cs, resources, dryRun)
 }
 
+func (cor *clusterReconciler) setBundlesRef(ctx context.Context, cluster *anywherev1.Cluster) error {
+	mgmtClusterKey := types.NamespacedName{Name: cluster.ManagedBy(), Namespace: cluster.Namespace}
+	mgmtCluster, err := cor.FetchCluster(ctx, mgmtClusterKey)
+	if err != nil {
+		return err
+	}
+	cluster.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
```
maxdrib, line 180: for a self-managed cluster, this will just set BundlesRef to itself, won't it? So if it was nil before, it'll be nil still.

g-gaston:
Interesting. How would the system reach this state? In order to use the new version of the code, the user has to run an upgrade command. And that upgrade command will set the bundles ref if it's null.

maxdrib:
>Interesting. How would the system reach this state? In order to use the new version of the code, the user has to run an upgrade command. And that upgrade command will set the bundles ref if it's null.

I agree, the system should not be in this state. I think the real issue we've observed stems from this fallback logic: https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/fetch.go#L78-L84

If a cluster is running the new EKS-A, I don't think it makes sense to fall back to legacy bundles and trigger a confusing rollback. Instead, it makes more sense to throw an error, since the BundlesRef field is now a required field.

TerryHowe:
Reopen if there is something here. Sounds like this "shouldn't happen"

PR #3241, opened 2022-09-01
Setup vSphere User Config Read & Validation

jonathanmeier5:
*Issue #, if available:*

*Description of changes:*
This adds business logic to read and validate a config file for the vSphere setup user command.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/vsphere/setupuser/validator.go
+++ b/pkg/providers/vsphere/setupuser/validator.go
@@ -4,10 +4,9 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	"gopkg.in/yaml.v2"
-
-	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
 const (
@@ -82,56 +81,24 @@ func readConfig(ctx context.Context, filepath string) (*VSphereSetupUserConfig, 
 }
 
 func validate(c *VSphereSetupUserConfig) error {
-	results := []validations.ValidationResult{
-		{
-			Name:        "validate datacenter",
-			Remediation: "",
-			Err:         validateDatacenter(c),
-		},
-		{
-			Name:        "validate vspheredomain",
-			Remediation: "",
-			Err:         validateVSphereDomain(c),
-		},
-		{
-			Name:        "validate connection",
-			Remediation: "",
-			Err:         validateConnection(c),
-		},
+	errs := []string{}
+
+	if c.Spec.Datacenter == "" {
+		errs = append(errs, "datacenter cannot be empty")
 	}
 
-	errs := []string{}
-	for _, r := range results {
-		if r.Err != nil {
-			errs = append(errs, r.Err.Error())
-		}
+	if c.Spec.VSphereDomain == "" {
+		errs = append(errs, "vSphereDomain cannot be empty")
+	}
+
+	if c.Spec.Connection.Server == "" {
+		errs = append(errs, "server cannot be empty")
 	}
 
 	if len(errs) > 0 {
-		return &validations.ValidationError{Errs: errs}
+		return fmt.Errorf("validations failed: %s", strings.Join(errs[:], ","))
 	}
 
-	return nil
-}
-
-func validateDatacenter(c *VSphereSetupUserConfig) error {
-	if c.Spec.Datacenter == "" {
-		return fmt.Errorf("datacenter cannot be empty")
-	}
-	return nil
-}
-
-func validateVSphereDomain(c *VSphereSetupUserConfig) error {
-	if c.Spec.VSphereDomain == "" {
-		return fmt.Errorf("vSphereDomain cannot be empty")
-	}
-	return nil
-}
-
-func validateConnection(c *VSphereSetupUserConfig) error {
-	if c.Spec.Connection.Server == "" {
-		return fmt.Errorf("server cannot be empty")
-	}
 	return nil
 }
 
--- a/pkg/providers/vsphere/setupuser/validator.go
+++ b/pkg/providers/vsphere/setupuser/validator.go
@@ -155,32 +155,4 @@ func setDefaults(c *VSphereSetupUserConfig) {
 	if c.Spec.Username == "" {
 		c.Spec.Username = DefaultUsername
 	}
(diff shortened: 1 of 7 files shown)


pkg/config/vspheresetupuser.go
```diff
@@ -0,0 +1,41 @@
+package config
```
chrisdoherty4, line 1: Why does this live separately to the `setupuser` code?

pkg/providers/vsphere/setupuser/validator.go
```diff
@@ -0,0 +1,117 @@
+package setupuser
```
chrisdoherty4, line 1: How do we intend on someone using this?

jonathanmeier5: I'm imagining adding another file, perhaps `pkg/providers/vsphere/setupuser/setup.go`, that defines something like `SetupVSphereUser` and calls this validation logic.

The command could then call `setupuser.SetupVSphereUser(ctx, filepath, ...)`.

I have little to no commitment to most of this structure. The only goal I was trying to achieve with a new package was some level of logical separation between vsphere code oriented around cluster create/upgrade/delete and this new workflow.

What do you have in mind?

pkg/config/vspheresetupuser.go
jonathanmeier5: I've been leaning into `config` a lot lately trying to extract globally accessed environment variables into config objects. None of this is deriving from the environment though, so yeah I'm happy to put it in `setupuser`.

chrisdoherty4: I would start by keeping related types close together. These types seem very specific to setting up a user so I would keep them there. 

A useful trick: consider the touch points for changing this functionality. If you _have_ to touch multiple packages to alter the business logic (wiring up doesn't count as business logic) there might be a smell.

pkg/providers/vsphere/setupuser/validator.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if err = yaml.Unmarshal(file, &c); err != nil {
+		return nil, fmt.Errorf("Failed to parse %s, err = %v", filepath, err)
+	}
+
+	return &c, nil
+}
+
+func validate(ctx context.Context, c *VSphereSetupUserConfig) error {
+	results := []validations.ValidationResult{
```
chrisdoherty4, line 85: Use of the `ValidationResult` seems redundant here given it doesn't tie into some grander framework.

pkg/providers/vsphere/setupuser/validator.go
```diff
@@ -0,0 +1,186 @@
...
+	if c.Spec.GroupName == "" {
+		c.Spec.GroupName = DefaultGroup
+	}
+
+	if c.Spec.Username == "" {
+		c.Spec.Username = DefaultUsername
+	}
+
+	if c.Spec.Objects.Networks == nil {
+		c.Spec.Objects.Networks = []string{}
```
chrisdoherty4, line 160: You shouldn't need to initialize slices as their zero value is valid.

jonathanmeier5: Oh wow, I didn't realize you can iterate over nil slices. Cool 👍

pkg/providers/vsphere/setupuser/validator.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+func validateConnection(ctx context.Context, c *VSphereSetupUserConfig) error {
+	if c.Spec.Connection.Server == "" {
+		return fmt.Errorf("server cannot be empty")
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
jonathanmeier5: My only thought for doing it this way is that it would be nice to have and use a ~standard framework for validating inputs in our CLI.

I was riffing off of [this PR ](https://github.com/aws/eks-anywhere/pull/3119/files) and trying to keep things similar.

pkg/providers/vsphere/setupuser/validator.go
jonathanmeier5: Same answer here: https://github.com/aws/eks-anywhere/pull/3241#discussion_r962001331

pkg/providers/vsphere/setupuser/validator.go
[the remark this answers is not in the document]
jonathanmeier5: Do you mind if we stick to the framework? What do you think?

pkg/providers/vsphere/setupuser/validator.go
```diff
@@ -0,0 +1,186 @@
...
+func setDefaults(ctx context.Context, c *VSphereSetupUserConfig) {
```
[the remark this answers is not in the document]
jonathanmeier5: I pulled it out of the validators as well

pkg/providers/vsphere/setupuser/validator.go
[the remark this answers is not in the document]
chrisdoherty4: A `validate` function could still be used in some framework - that's really about signatures etc. In any case, this isn't being used with a framework as is so I still feel like it would be simpler and clearer to have a `validate(cfg *VSphereSetupUserConfig) error`.

Re 'standard frameworks' we don't really have one that serves our needs. There have been various attempts but we've never got it quite right.

pkg/providers/vsphere/setupuser/validator.go
chrisdoherty4: I think the 2 comments are a little differentiates. What I mean to say here is there's now consumer of `ValidationResult` in this package nor does anything return a set of validation results.

What we have here is semantically equivalent to:

```go
results := []error{
  validateDatacenter(...),
  validateDomain(...),
}

for _, result := range results {
  if result != nil {
    ...
  }
}
```

pkg/providers/vsphere/setupuser/validator.go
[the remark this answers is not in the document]
jonathanmeier5: Updated 👍

chrisdoherty4 (approved):

PR #3243, opened 2022-09-01
Add analyer to analyze if vms have access to api server

YoyoTT:
*Issue #, if available:*

*Description of changes:*

Extend the functionality of support bundle by

- adding a run-pod collector--vmsAccessCollector. It supposes to collect logs from vsphere cloud controller manager on control plane.
   - since the pod need to be deployed on control plane node, specific tolerations have been set.

   - It is also able to accept customized taints that given by user in cluster defination yaml.

- adding a vmsAccess analyzer. Based on the log from the collector, the analyzer is able to analyze if the vms have access to vsphere api server.

    - Now, the analyzer starts to work from bootstrap cluster, although we expect that the analyzer only does its work on workload cluster. A clarification was temporarily added to the output to avoid confusion.
    - Also, more work on cluster_manager.go is needed to generate kubeconfig for a cluster. Therefore, current analyzer is implemented but hasn't been used.  

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/analyzers.go
+++ b/pkg/diagnostics/analyzers.go
@@ -362,3 +362,33 @@ func (a *analyzerFactory) controlPlaneIPAnalyzer() []*Analyze {
 		},
 	}
 }
+
+// vmsAccessAnalyzer will analyze if vms have access to the API server of vSphere cluster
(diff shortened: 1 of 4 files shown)


mitalipaygude:
ok-to-test

g-gaston (approved):

PR #3244, opened 2022-09-02
Package proxy fix

a-cool-train:
*Description of changes:*
- Currently curated packages don't work behind a proxy. This enables curated packages to be installed behind a proxy.

*Testing (if applicable):*

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -94,12 +94,12 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error 
 	return nil
 }
 
-func (pc *PackageControllerClient) IsInstalled(ctx context.Context) error {
+func (pc *PackageControllerClient) IsInstalled(ctx context.Context) (bool, error) {
 	found, _ := pc.kubectl.GetResource(ctx, "packageBundleController", packagesv1.PackageBundleControllerName, pc.kubeConfig, constants.EksaPackagesName)
 	if found {
-		return errors.New("curated Packages controller exists in the current cluster")
+		return true, errors.New("curated Packages controller exists in the current cluster")
 	}
-	return nil
+	return false, nil
 }
 
 func (pc *PackageControllerClient) ApplySecret(ctx context.Context) error {
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -162,7 +162,7 @@ func WithHTTPProxy(httpProxy string) func(client *PackageControllerClient) {
 	}
 }
 
-func WithHttpsProxy(httpsProxy string) func(client *PackageControllerClient) {
+func WithHTTPSProxy(httpsProxy string) func(client *PackageControllerClient) {
 	return func(config *PackageControllerClient) {
 		config.httpsProxy = httpsProxy
 	}
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"strings"
 
-	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/templater"
@@ -93,9 +92,8 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error 
 	return nil
 }
 
-func (pc *PackageControllerClient) IsInstalled(ctx context.Context) bool {
-	found, _ := pc.kubectl.GetResource(ctx, "packageBundleController", packagesv1.PackageBundleControllerName, pc.kubeConfig, constants.EksaPackagesName)
-	return found
+func (pc *PackageControllerClient) IsInstalled(ctx context.Context) (bool, error) {
+	return pc.kubectl.GetResource(ctx, "packageBundleController", pc.clusterName, pc.kubeConfig, constants.EksaPackagesName)
 }
 
 func (pc *PackageControllerClient) ApplySecret(ctx context.Context) error {
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -67,12 +67,18 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error 
 
 	sourceRegistry := fmt.Sprintf("sourceRegistry=%s", registry)
 	clusterName := fmt.Sprintf("clusterName=%s", pc.clusterName)
-	httpProxy := fmt.Sprintf("proxy.HTTP_PROXY=%s", pc.httpProxy)
-	httpsProxy := fmt.Sprintf("proxy.HTTPS_PROXY=%s", pc.httpsProxy)
-	noProxy := fmt.Sprintf("proxy.NO_PROXY=%s", strings.Join(pc.noProxy, "\\,"))
-	logger.Info(fmt.Sprintf("httpProxy: %s, httpsProxy: %s, noProxy: %s", pc.httpProxy, pc.httpsProxy, pc.noProxy))
+	values := []string{sourceRegistry, clusterName}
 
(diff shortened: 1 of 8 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -139,3 +155,21 @@ func WithEksaRegion(eksaRegion string) func(client *PackageControllerClient) {
 		}
 	}
 }
+
+func WithHttpProxy(httpProxy string) func(client *PackageControllerClient) {
```
chrisdoherty4, line 159:
```suggestion
func WithHTTPProxy(httpProxy string) func(client *PackageControllerClient) {
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -139,3 +155,21 @@ func WithEksaRegion(eksaRegion string) func(client *PackageControllerClient) {
...
+	return func(config *PackageControllerClient) {
+		config.httpProxy = httpProxy
+	}
+}
+
+func WithHttpsProxy(httpsProxy string) func(client *PackageControllerClient) {
```
chrisdoherty4, line 165:
```suggestion
func WithHTTPSProxy(httpsProxy string) func(client *PackageControllerClient) {
```

pkg/curatedpackages/packageinstaller.go
```diff
@@ -10,6 +10,7 @@ import (
 
 type PackageController interface {
 	InstallController(ctx context.Context) error
+	ValidateControllerDoesNotExist(ctx context.Context) error
```
chrisdoherty4, line 13: You could abbreviate this to `IsInstalled` and then use `!controller.IsInstalled(ctx)`. Typically these functions are better defined as affirming something _is_ rather than _is not_. That's because its easier to understand `!Isinstalled()` and `IsInstalled()` as opposed to `IsNotInstalled()` and `!IsNotInstalled()` (the latter being the confusing part because its a double negative).

pkg/dependencies/factory.go
```diff
@@ -1087,3 +1092,11 @@ func (f *Factory) WithVSphereDefaulter() *Factory {
 
 	return f
 }
+
+func getProxyConfiguration(clusterSpec *cluster.Spec) (string, string, []string) {
```
g-gaston, line 1096: I would name these returns
```suggestion
func getProxyConfiguration(clusterSpec *cluster.Spec) (httpProxy, httpsProxy string, noProxy []string) {
```

pkg/curatedpackages/packageinstaller.go
g-gaston: Agreed with @chrisdoherty4 
However, where is this being used?

a-cool-train: It is currently used here: https://github.com/aws/eks-anywhere/blob/72414be6c5bfe3caccaf4cbf8fc2de157c2a2c82/cmd/eksctl-anywhere/cmd/installpackagecontroller.go#L92.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -78,7 +94,7 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
 	return nil
 }
 
-func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Context) error {
+func (pc *PackageControllerClient) IsInstalled(ctx context.Context) error {
 	found, _ := pc.kubectl.GetResource(ctx, "packageBundleController", packagesv1.PackageBundleControllerName, pc.kubeConfig, constants.EksaPackagesName)
 	if found {
```
chrisdoherty4, line 99: You'll need to change the logic here too given we're now inverting the names semantics.

pkg/curatedpackages/packageinstaller.go
g-gaston: Ahh. I think we should probably change the factory so it stores a concrete implementation as opposed to the interface.

pkg/dependencies/factory.go
```diff
@@ -1087,3 +1092,11 @@ func (f *Factory) WithVSphereDefaulter() *Factory {
 
 	return f
 }
+
+func getProxyConfiguration(clusterSpec *cluster.Spec) (httpProxy, httpsProxy string, noProxy []string) {
+	proxyConfiguration := clusterSpec.Cluster.Spec.ProxyConfiguration
+	if proxyConfiguration != nil {
+		return proxyConfiguration.HttpProxy, proxyConfiguration.HttpsProxy, proxyConfiguration.NoProxy
+	}
+	return "", "", []string{}
```
g-gaston, line 1101: you can return nil for the slice

pkg/curatedpackages/packageinstaller.go
a-cool-train: What would the benefit of storing the concrete implementation rather than the interface?

pkg/dependencies/factory.go
TerryHowe: Fine for now, but just the kind of convenience function I'd want to have with clusterSpec as the receiver

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: The logic doesn't change actually. The logic is throw an error if the controller is already installed.

pkg/curatedpackages/packageinstaller.go
g-gaston: The factory is not using this entity, is only building it. And it's building a concrete type directly. In general, is better to return concrete implementations and accept interfaces.

One pretty straight forward benefit: we had to add/change this method in an interface defined in the `curatedpackages` package when the (I think?) only receiver is the cmd package. Which, in fact, doesn't care about receiving interfaces or not because is not even being tested and already depends on the `curatedpackages`. If the factory was returning a concrete type, we wouldn't need to update that interface.

IMO interfaces in the deps struct make sense only when we conditionally build different implementations depending on some external configuration (like we do with providers).

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -78,7 +94,7 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
 	return nil
 }
 
-func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Context) error {
+func (pc *PackageControllerClient) IsInstalled(ctx context.Context) error {
```
g-gaston, line 97: Based on the name, should't this return a boolean?
Like:
```suggestion
func (pc *PackageControllerClient) IsInstalled(ctx context.Context) (bool, error) {
```

g-gaston (commented):
Looks good, just minor suggestion for the method signature

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -78,12 +94,12 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
-func (pc *PackageControllerClient) ValidateControllerDoesNotExist(ctx context.Context) error {
+func (pc *PackageControllerClient) IsInstalled(ctx context.Context) (bool, error) {
 	found, _ := pc.kubectl.GetResource(ctx, "packageBundleController", packagesv1.PackageBundleControllerName, pc.kubeConfig, constants.EksaPackagesName)
 	if found {
-		return errors.New("curated Packages controller exists in the current cluster")
+		return true, errors.New("curated Packages controller exists in the current cluster")
 	}
-	return nil
+	return false, nil
 }
```
g-gaston, line 103: Ah interesting
I was thinking something like

```go
func (pc *PackageControllerClient) IsInstalled(ctx context.Context) (bool, error) {
    found, err := pc.kubectl.GetResource(ctx, "packageBundleController", packagesv1.PackageBundleControllerName, pc.kubeConfig, constants.EksaPackagesName)
    if err != nil {
        return false, err
    }

    return found, nil
}
```

chrisdoherty4: Ditto. This ties in with my [other comment](https://github.com/aws/eks-anywhere/pull/3244#discussion_r961932500).

a-cool-train: Ok I think there is some confusion here. IsInstalled should lead to an error due to the fact that there is a controller that is already installed in the cluster. We wouldn't want to install the controller if there is already an installation. I hope this clears this up.

g-gaston (approved):

PR #3245, opened 2022-09-02
Add support for weekly bundle creation and Github release

abhay-krishna:
This PR adds a new weekly release mode to the release tool that generates a bundle release every week with a different tagging scheme. It also adds a script that will publish this bundle release manifest as a GitHub release with a date-based tag on main branch. The release notes will contain the commit information pertaining to this release.

### DIFF
--- a/release/Makefile
+++ b/release/Makefile
@@ -28,6 +28,9 @@ SOURCE_BUCKET=projectbuildpipeline-857-pipelineoutputartifactsb-10ajmk30khe3f
 else
 DRY_RUN=false
 endif
+DATE_YYYYMMDD=$(shell date "+%F")
+WEEKLY?=false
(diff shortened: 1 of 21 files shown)


TerryHowe:
If we update the GIT_TAG in build tooling for the package controller, will this weekly bundle add the new controller?

abhay-krishna:
Yeah it would, depending on when the tag is updated. It does a full build of all projects and generates a bundle manifest once a week. If the packages update happens _after_ the weekly pipeline has run for that week, it will be included in the following week's build.

abhay-krishna:
> If we update the GIT_TAG in build tooling for the package controller, will this weekly bundle add the new controller?

Yeah it would, depending on when the tag is updated. The weekly release pipeline does a full build of all projects and generates a bundle manifest once a week. If the packages update happens _after_ this pipeline has run for that week, it will be included in the following week's build.

abhinavmpandey08 (approved):

PR #3246, opened 2022-09-02
Add new workflow definition framework

### LINKED ISSUE #3286: Implement workflow framework

chrisdoherty4:
Add workflow implementation and template the management cluster create, upgrade and delete workflows.

Error handling is sketched out but may be revised as part of a separate story. Checkpoint-recovery and profiling are not implemented. Package documentation has purposely been left out as we expect to iterate on the framework so wish to minimize doc churn.

Closes #3286


pkg/workflow/workflow.go
```diff
@@ -0,0 +1,94 @@
...
+type TaskID string
+
+// Task represents an individual step within a workflow that can be run.
+type Task interface {
+	// ID returns a unique identifier for the task.
+	ID() TaskID
+
+	// RunTask executes the task. Tasks may return a context that should be used in subsequent task
+	// execution.
+	RunTask(context.Context) (context.Context, error)
```
chrisdoherty4, line 15: Returning a context was necessary for information not available on construction, such as the bootstrap cluster config. 

We could write a custom context that lets us populate the parameter passed instead of treating it as immutable.

pkg/workflow/context.go
```diff
@@ -0,0 +1,23 @@
+package workflow
+
+import "context"
+
+// targetCluster is used to store and retrieve a target cluster kubeconfig.
+var targetCluster contextKey = "targetCluster"
+
+// WithTargetCluster returns a context based on ctx containing the target cluster kubeconfig.
```
chrisdoherty4, line 8: This code illustrates how to populate a context with information as workflows execute.

pkg/workflow/hook.go
```diff
@@ -0,0 +1,40 @@
+package workflow
+
+import (
+	"context"
+)
+
+// Hook defines an algorithm to be run before or after a task is executed.
+type Hook interface {
```
chrisdoherty4, line 8: Change to using Tasks and make clear tasks can operate on the context

pkg/workflow/context.go
```diff
@@ -0,0 +1,23 @@
...
+func WithTargetCluster(ctx context.Context, kubeconfig string) context.Context {
```
chrisdoherty4, line 9: Draft something using a generic context.

pkg/workflow/management/create.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Build builds the create cluster workflow.
+func (cfg *CreateClusterBuilder) Build() (*workflow.Workflow, error) {
+	wflw, err := workflow.New(workflow.Config{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, r := range cfg.HookRegistrars {
```
g-gaston, line 30: This got me thinking, do we actually need to expose methods to "add" registrars to the builder? If the builder is the entity who knows how to build a workflow and what are the tasks that should be included in it, shouldn't it also know which are right the hook registrars for it?

That would remove the need to have a `WithHookRegistrar` method, which eliminates the debate generics vs code duplication for it.

Maybe we got the point where we need to go and try build one workflow? Maybe the answers would come easier then.

pkg/workflow/management/upgrade.go
```diff
@@ -0,0 +1,41 @@
...
+	RegisterUpgradeManagementClusterHooks(workflow.HookBinder)
+}
+
+// UpgradeClusterBuilder defines the configuration for a management cluster upgrade workflow.
+type UpgradeClusterBuilder struct {
+	HookRegistrars []UpgradeClusterHookRegistrar
+}
+
+// WithHookRegistrar adds a hook registrar to the upgrade cluster workflow builder.
+func (b *UpgradeClusterBuilder) WithHookRegistrar(registrar UpgradeClusterHookRegistrar) *UpgradeClusterBuilder {
```
g-gaston, line 22: If we were gonna have this kind of method (see my comment above), this would have my vote. It does require to re-implement this method for every workflow builder, but given it's two lines and we don't expect the number of workflows to wrong significantly, it a good tradeoff for the cognitive complexity it removes.

That said, if y'all agree the generics approach is better, it wouldn't concern me much.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,127 @@
...
+		preHooks:  make(map[TaskName][]Task),
+		postHooks: make(map[TaskName][]Task),
+	}
+
+	return wflw, nil
+}
+
+// AppendTask appends t to the list of workflow tasks.
+func (w *Workflow) AppendTask(t Task) {
+	w.tasks = append(w.tasks, t)
```
g-gaston, line 42: should we make sure task IDs are unique?

pkg/workflow/management/create.go
chrisdoherty4: I don't think so. The design stipulates an intent to disambiguate the concern of workflows where a workflow should only include tasks that create, upgrade or delete a cluster - nothing more, nothing less. 

Embedding knowledge of registrars here suggests workflows _also_ know about concrete hook capabilities and I think that conflates the scope of a workflow.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"context"
+)
+
+// Config is the configuration for constructing a Workflow instance.
+type Config struct {
+	// ErrorHandler is handler called when a workflow experiences an error. The error may originate
+	// from hook or from a task. The original error is alwasy returned from the workflow's Execute.
+	// Optional. Defaults to a no-op handler.
+	ErrorHandler ErrorHandler
```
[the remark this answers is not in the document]
chrisdoherty4: I would expect the error handler to define those rules. We would need to provide information on the current task and previous tasks which shouldn't be difficult via the context. That might even be a good justification for a custom context given it will be applicable to everything.

pkg/workflow/workflow.go
chrisdoherty4: We probably should, yes.

pkg/workflow/workflow.go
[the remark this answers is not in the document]
g-gaston: Ah I see. Like the we would have an error handler to orchestrates other handlers, sounds good.

If we know we need to pass that kind info (tasks executed, etc.) to the handler, doesn't that guarantee a dedicated argument for that?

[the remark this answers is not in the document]
chrisdoherty4: Possibly. I've purposely left this rather plain for this PR. I included it so we have an appreciation for how this could be done. I have an open issue to actually implement this including the design work/pocing.

pkg/workflow/management/upgrade.go
chrisdoherty4: I've updated everything with this approach. Its easy enough to return to a generic approach if we want.

jacobweinstock (approved):

g-gaston (approved):

vincentni (approved):

PR #3247, opened 2022-09-02
Adding validation to bundleRef

maxdrib:
*Issue #, if available:*

*Description of changes:*
When upgrading clusters from an old version of EKS-A to one using the new BundleRef field, we observed an issue where the bundleRef field could be wiped out unintentionally with some flux or kubectl apply operations resulting in an unintended rolling upgrade to an older set of bundles. After further thought and discussion, also agreed to make them monotonically increasing by default, or settable to a custom name as a workaround in case of issues.

This PR makes the field required if it was already set.

*Testing (if applicable):*
Created a cluster, deployed this controller onto it, and tried removing bundleRef. The kubectl operation was rejected, as desired. Also unit tests.


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -213,6 +216,23 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 			field.Forbidden(specPath.Child("ControlPlaneConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ControlPlaneConfiguration)))
 	}
 
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef == nil {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("BundlesRef"), fmt.Sprintf("field cannot be removed after setting. Previous value %v", old.Spec.BundlesRef)))
+	}
+
+	standardBundleRef := regexp.MustCompile(`^bundle-[0-9]+$`)
```
maxdrib, line 225: this prefix is incorrect - should be `bundles` (https://github.com/aws/eks-anywhere/blob/main/release/pkg/bundles/bundles.go#L33)

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -213,6 +216,23 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 			field.Forbidden(specPath.Child("ControlPlaneConfiguration"), fmt.Sprintf("field is immutable %v", new.Spec.ControlPlaneConfiguration)))
 	}
 
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef == nil {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("BundlesRef"), fmt.Sprintf("field cannot be removed after setting. Previous value %v", old.Spec.BundlesRef)))
+	}
+
+	standardBundleRef := regexp.MustCompile(`^bundles-[0-9]+$`)
```
g-gaston, line 225: `MustCompile` will panic, which will crash the controller
If we want to construct the regex in the method scope, we should use the regexp method that returns an error
If not, I would vote to put this in a package private var

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -213,6 +216,23 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
...
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef != nil && standardBundleRef.MatchString(old.Spec.BundlesRef.Name) && standardBundleRef.MatchString(new.Spec.BundlesRef.Name) {
+		oldIndex, _ := strconv.Atoi(strings.Split(old.Spec.BundlesRef.Name, "-")[1])
```
g-gaston, line 227: instead of doing this, can we use capturing groups in the regex? That way we use the regex to extract the number and avoid doing the string manipulation here

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -213,6 +216,23 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
...
+		newIndex, _ := strconv.Atoi(strings.Split(new.Spec.BundlesRef.Name, "-")[1])
+		if newIndex < oldIndex {
+			allErrs = append(
+				allErrs,
+				field.Forbidden(specPath.Child("BundlesRef"), fmt.Sprintf("bundle id must be monotonically increasing. Previous value %v, new value %v", old.Spec.BundlesRef, new.Spec.BundlesRef)))
+		}
+	}
```
g-gaston, line 234: I don't think this validations belong in `validateImmutableFieldsCluster`
I would vote to have them in a separate method and call it from `ValidateUpdate`

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -213,6 +216,23 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
...
+
+	standardBundleRef := regexp.MustCompile(`^bundles-[0-9]+$`)
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef != nil && standardBundleRef.MatchString(old.Spec.BundlesRef.Name) && standardBundleRef.MatchString(new.Spec.BundlesRef.Name) {
+		oldIndex, _ := strconv.Atoi(strings.Split(old.Spec.BundlesRef.Name, "-")[1])
+		newIndex, _ := strconv.Atoi(strings.Split(new.Spec.BundlesRef.Name, "-")[1])
+		if newIndex < oldIndex {
+			allErrs = append(
+				allErrs,
+				field.Forbidden(specPath.Child("BundlesRef"), fmt.Sprintf("bundle id must be monotonically increasing. Previous value %v, new value %v", old.Spec.BundlesRef, new.Spec.BundlesRef)))
```
g-gaston, line 232: I don't think this a forbidden error, maybe invalid?

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -833,6 +833,24 @@ func TestClusterValidateUpdateAWSIamNameImmutableAddConfig(t *testing.T) {
...
+func TestClusterValidateUpdateUnsetBundlesRefImmutable(t *testing.T) {
+	cOld := createCluster()
+	c := cOld.DeepCopy()
+	c.Spec.BundlesRef = nil
+
+	g := NewWithT(t)
+	g.Expect(c.ValidateUpdate(cOld)).NotTo(Succeed())
+}
+
+func TestClusterValidateUpdateBundleRefRollbackInvalid(t *testing.T) {
```
g-gaston, line 845: Could we add a test for:
* bundle moving up
* bundle staying the same
* updating bundle to a custom name

release/pkg/bundles/bundles.go
```diff
@@ -29,6 +29,8 @@ import (
 	sliceutils "github.com/aws/eks-anywhere/release/pkg/util/slices"
 )
 
+// NewBundlesName provides a strict format for bundle names, which is validated against in the Cluster webhook
+// Numbers must be monotonically increasing to be upgraded.
```
g-gaston, line 33: 🙏🏻

pkg/api/v1alpha1/cluster_webhook_test.go
maxdrib: Will add test for bundle moving up and updating bundle to custom name. Bundle staying the same should be covered by other tests.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -107,6 +112,36 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+	var allErrs field.ErrorList
+	bundlesRefPath := field.NewPath("spec").Child("BundlesRef")
+
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef == nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(bundlesRefPath, new.Spec.BundlesRef, fmt.Sprintf("field cannot be removed after setting. Previous value %v", old.Spec.BundlesRef)))
+	}
+
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef != nil && standardBundleRef.MatchString(old.Spec.BundlesRef.Name) && standardBundleRef.MatchString(new.Spec.BundlesRef.Name) {
```
g-gaston, line 125: is it necessary to run the regex twice per bundle? Can't we consolidate it into one?
When running `FindStringSubmatch`, if the matches lenght is not 2, then we know it doesn't match. We can either return -1 from the `getBundleIdFromBundleRefName` (common in the go standard library) or we could even return `(bool, int)`

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -29,6 +31,7 @@ import (
 
 // log is for logging in this package.
 var clusterlog = logf.Log.WithName("cluster-resource")
+var standardBundleRef = regexp.MustCompile(`bundles-(?P<bundleid>\d+)`)
```
g-gaston, line 34:
```suggestion
var standardBundleRef = regexp.MustCompile(`^bundles-(?P<bundleid>\d+)$`)
```

pkg/api/v1alpha1/cluster_webhook.go
maxdrib: Refactored to only run regex once per bundle

g-gaston (approved):

PR #3250, opened 2022-09-02
Add cluster lifecycle providers integration guide

g-gaston:

### DIFF
--- a/docs/developer/cluster-lifecycle-api-provider-integration.md
+++ b/docs/developer/cluster-lifecycle-api-provider-integration.md
@@ -1,0 +1,71 @@
+# Cluster Lifecycle API provider integration guide
+
+> `THIS IS A WORK IN PROGRESS`
+
+## Context
+> Note: if you haven't yet, reading the [original design doc](https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md) is a must to execute this integration.
+
+The main reconciliation logic is ran in the cluster controller. This watches all eks-a CRs and reacts to their changes. However, currently the cluster controller itself only performs some config/setup (controller watches, patch the `Cluster` CR at the end of the reconcile method, set owner refs, etc.) and reconciles deleted `Cluster`s. The main reconciliation logic is delegated to the `ProviderClusterReconciler`'s, one per provider.
+
+`ProviderClusterReconciler` has to handle all the actions needed to reconcile the workload cluster to the state defined in the eks-a CRs. It has complete control over how this is implemented and there are no restrictions imposed in this regard. That said, there are 3 main tasks generic tasks that every `ProviderClusterReconciler` has to perform, in this order:
+1. Reconcile the control plane to the eks-a CR specifications
+1. Reconcile the workload cluster worker nodes to the eks-a CR specifications
+1. Reconcile the CNI to the eks-a CR specifications
+
+Each provider can add any other additional steps, in order: before, after or even in between of these 3.
+
+The CNI reconciliation should be executed by using the Cilium reconciler (and other in the future), orchestrated by the main [CNI Reconciler](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/reconciler/reconciler.go). These are designed to work in a reconciliation loop provided a stable Control Plane and a client to interact with.
+
+## Requirements
+
+* Implement the [`ProviderClusterReconciler` interface](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusters/registry.go#L12)
+* Build and register the new reconciler in the [controller factory](https://github.com/aws/eks-anywhere/blob/main/controllers/factory.go#L170)
+* Add the necessary watches to the [cluster controller](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L54)
+* Update [cluster controller RBAC](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L94) accordingly. You might need permissions to watch the provider eks-a CRDs or read/create/update/delete the provider infra CAPI CRDs.
+* If they haven't been added yet, add the infra CAPX API for your provider to the [scheme](https://github.com/aws/eks-anywhere/blob/main/manager/main.go#L45).
+* Add your provider specific API structs to [`cluster.Config`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10). Add them as [child objects](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L88) if necessary.
+* Implement the proper [client processors](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/client_builder.go#L18) for your API structs and register them in the [default `ConfigClientBuilder`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10)
+* Add validation webhooks for all provider specific CRDs
+* Add mutation webhooks for all provider specific CRDs when defaults are needed
+* If context aware defaults or validations (those that require any kind of API or disk call) are needed, they should be implemented in a controller for that CRD ([example](https://github.com/aws/eks-anywhere/blob/main/controllers/snow_machineconfig_controller.go)). For any extra validations, the `ProviderClusterReconciler` will need to check if the spec of the eks-a provider CRs is valid before starting the reconciliation loop. In case they are not, bubble that error to the `Cluster.status`.
+
+## Recommendations
+
+### Generating CAPI specs from EKS-A objects
+
+#### Templating vs API structs
+
+Currently, most providers make use of `go` templates to generate `yaml` documents with all the CAPI objects based on a particular eks-a spec. This has resulted in quite complex templates, with a lot of duplication between providers. While it's possible to fix the duplication issue by using reusable template fragments, and reconciling yaml documents from the controller shouldn't be an issue (using for example [this](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/serverside/reconcile.go#L14)), we recommend moving away from this and using `go` API structs directly.
+
+A common set of generator helpers are available in the [`clusterapi`](https://github.com/aws/eks-anywhere/blob/main/pkg/clusterapi/apibuilder.go) package. We encourage all providers implementers to enrich the package with anything not provider specific so everyone can take advantage of it.
+
+#### Immutable objects
+
+The biggest challenge when generating CAPI specs is the generation of new objects for immutable CRDs. The current pattern the CLI implements is: retrieving the current state of eks-a CRs and based on the diff with the new desired spec and determine if such immutable objects would change after executing the conversion logic or not. This comparison is mostly and implicit version of the reverse transformation logic, one that tries to reverse the eks-a -> CAPI conversion logic.
+
+Unfortunately this is not possible in the controller world since there is no old and new version of objects, only current. To solve this, the original cluster implements a reverse logic, trying to infer the past state of eks-a objects from the current state of CAPI objects. Then it plugs into the main CLI logic, running the previously mentioned comparison logic. This is extremely prone to errors and very hard to maintain.
+
+We encourage all providers to follow the reverse algorithm: instead of comparing the old and new inputs and inferring if the outputs would differ, compare directly the old and new output. This is a way simpler process: retrieve the current status of the immutable CAPI object, run the conversion logic to generate that same object based on the current EKS-A spec and just compare them.
+
+When used in combination with API structs, the comparison step becomes trivial most of the times, being able to take advantage of `equality.Semantic.DeepDerivative` provided by the `apimachinery` module. Examples of this can be found in the [Snow](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/snow/objects.go) provider, which reuses the same code/logic for both the controller and the CLI.
+
+
+## Development
+
+### Enabling full cluster lifecycle API support
+
+Right now, there are two different implementations of the cluster controller: the legacy one, with limited operations support, and the new one, under development. This is gated by a feature gate. In order to enable it you have several alternatives:
+* Set `FULL_LIFECYCLE_API` to true before creating the management cluster. (`export FULL_LIFECYCLE_API=true`)
+* Set `FULL_LIFECYCLE_API` to true before running `tilt`
+* Manually add a flag arg to the `manager` command in the `eksa-controller-manager` deployment: `--feature-gates FullLifecycleAPI=true`
+
+### Tooling
+
+The are some packages that can be useful when implementing a `ProviderClusterReconciler`. Their use is encouraged but not enforced.
+
+* [`PhaseRunner`](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/runner.go). Useful for running reconciliation logic that can divided in phases where each phase can abort the execution by either erroring out or requesting a reconcile request requeue (for example, to wait some time and so it check again for a particular event).
+* [Helpers to retrieve CAPI objects](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusterapi.go)
+* [Helpers to use server-side apply](https://github.com/aws/eks-anywhere/tree/main/pkg/controller/serverside)
+* [Helpers to extract data from CAPI objects](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusters/clusterapi.go)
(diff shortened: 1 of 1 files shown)


docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+> `THIS IS A WORK IN PROGRESS`
+
+## Context
+> Note: if you haven't yet, reading the [original design doc](https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md) is a must to execute this integration.
+
+Then main reconciliation logic is ran in the cluster controller. This watches all eks-a CRs and reacts to their changes. However, currently the cluster controller itself only performs some config/setup (controller watches, patch the `Cluster` CR at the end of the reconcile method, set owner refs, etc.) and reconciles deleted `Cluster`s. The main reconciliation logic is delegated to the `ProviderClusterReconciler`'s, one per provider.
+
+`ProviderClusterReconciler` has to handle all the actions needed to reconcile the workload cluster to the state defined in the eks-a CRs. It has complete control over how this is implemented and there are no restrictions imposed in this regard. That said, there are 3 main tasks generic tasks that every `ProviderClusterReconciler` has to perform, in this order:
+1. Reconcile the workload cluster Control Plane to the eks-a CR specifications
+1. Reconcile the worker nodes to the eks-a CR specifications
```
jiayiwang7, line 12: should it be 
```suggestion
1. Reconcile the workload cluster worker nodes to the eks-a CR specifications
```

or are we supporting mgmt cluster reconciliation in controller?

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+
+> `THIS IS A WORK IN PROGRESS`
+
+## Context
+> Note: if you haven't yet, reading the [original design doc](https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md) is a must to execute this integration.
+
+Then main reconciliation logic is ran in the cluster controller. This watches all eks-a CRs and reacts to their changes. However, currently the cluster controller itself only performs some config/setup (controller watches, patch the `Cluster` CR at the end of the reconcile method, set owner refs, etc.) and reconciles deleted `Cluster`s. The main reconciliation logic is delegated to the `ProviderClusterReconciler`'s, one per provider.
+
+`ProviderClusterReconciler` has to handle all the actions needed to reconcile the workload cluster to the state defined in the eks-a CRs. It has complete control over how this is implemented and there are no restrictions imposed in this regard. That said, there are 3 main tasks generic tasks that every `ProviderClusterReconciler` has to perform, in this order:
+1. Reconcile the workload cluster Control Plane to the eks-a CR specifications
```
vivek-koppuru, line 11: We are already referring to workload cluster above right? And why capitalize?
```suggestion
1. Reconcile the control plane to the eks-a CR specifications
```

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+1. Reconcile the worker nodes to the eks-a CR specifications
+1. Reconcile the CNI to the eks-a CR specifications
+
+Each provider can add any other additional steps, in order: before, after ir even in between of these 3.
```
vivek-koppuru, line 15: How so tho? 
```suggestion
Each provider can add any other additional steps, in order: before, after or even in between of these 3.
```

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+
+The CNI reconciliation should be executed by using the Cilium reconciler (and other in the future), orchestrated by the main [CNI Reconciler](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/reconciler/reconciler.go). These are designed to work in a reconciliation loop provided a stable Control Plane and and a client to interact with.
```
vivek-koppuru, line 17:
```suggestion
The CNI reconciliation should be executed by using the Cilium reconciler (and other in the future), orchestrated by the main [CNI Reconciler](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/reconciler/reconciler.go). These are designed to work in a reconciliation loop provided a stable Control Plane and a client to interact with.
```

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Requirements
+
+* Implement the [`ProviderClusterReconciler` interface](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusters/registry.go#L12)
+* Build and register the new reconciler in the [controller factory](https://github.com/aws/eks-anywhere/blob/main/controllers/factory.go#L170)
+* Add the necessary watches to the [cluster controller](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L54)
+* Update [cluster controller RBAC](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L94) accordingly. You might need permissions to watch the provider eks-a CRDs or read/create/update/delete the provider infra CAPI CRDs.
+* Add your provider specific API structs to [`cluster.Config`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10). Add them ass [child objects](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L88) if necessary.
```
jiayiwang7, line 25: wow 
```suggestion
* Add your provider specific API structs to [`cluster.Config`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10). Add them as [child objects](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L88) if necessary.
```

vivek-koppuru:
```suggestion
* Add your provider specific API structs to [`cluster.Config`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10). Add them as [child objects](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L88) if necessary.
```

jiayiwang7 (approved):
Super! minor comments

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+* If context aware defaults or validations (those that require any kind of API or disk call) are needed, they should be implemented in a controller for that CRD ([example](https://github.com/aws/eks-anywhere/blob/main/controllers/snow_machineconfig_controller.go)). For any extra validations, the `ProviderClusterReconciler` will need to check if the spec of the eks-a provider CRs is valid before starting the reconciliation loop. In case they are not, bubble that error to the `Cluster.status`.
+
+## Recommendations
+
+### Generating CAPI specs from EKS-A objects
+
+#### Templating vs API structs
+
+Currently, most providers make use of `go` templates to generate `yaml` documents with all the CAPI objects based on a particular eks-a spec. This has resulted in quite complex templates, with a lot of duplication between providers. While it's possible to fix the duplication issue by using reusable template fragments and reconcile yaml documents from the controller shouldn't be an issue (using for example [this](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/serverside/reconcile.go#L14)), we recommend moving away from this and using `go` API structs directly.
```
vivek-koppuru, line 37:
```suggestion
Currently, most providers make use of `go` templates to generate `yaml` documents with all the CAPI objects based on a particular eks-a spec. This has resulted in quite complex templates, with a lot of duplication between providers. While it's possible to fix the duplication issue by using reusable template fragments, and reconciling yaml documents from the controller shouldn't be an issue (using for example [this](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/serverside/reconcile.go#L14)), we recommend moving away from this and using `go` API structs directly.
```

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+
+* [`PhaseRunner`](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/runner.go). Useful for running reconciliation logic that can divided in phases where each phase can abort the execution by either erroring out or requesting a reconcile request requeue (for example, to wait some time and so it check again for a particular event).
+* [Helpers to retrieve CAPI objects](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusterapi.go)
+* [Helpers to use server-side apply](https://github.com/aws/eks-anywhere/tree/main/pkg/controller/serverside)
+* [Helpers to extract data from CAPI objects](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusters/clusterapi.go)
+* [Build cluster.Spec from a client using `cluster.BuildSpec`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/fetch.go#L159)
+* [`ObjectApplier`](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/serverside/applier.go#L16): helpful when generating the CAPI spec as API structs (`client.Object`s), allowing to easily reconcile them using server-side apply.
```
maxdrib, line 60: This may be a larger issue, but I think we're missing some details on how debug the code with this model. I found it to be very valuable to use the GoLand debugger when observing issues with cluster creation, for example. What would be the equivalent approach when all operations are performed on the EKS-A controller? Is the expectation to use Tilt with a debugger? If so, are there any instructions for setting that up?

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+1. Reconcile the workload cluster worker nodes to the eks-a CR specifications
+1. Reconcile the CNI to the eks-a CR specifications
+
+Each provider can add any other additional steps, in order: before, after ir even in between of these 3.
+
+The CNI reconciliation should be executed by using the Cilium reconciler (and other in the future), orchestrated by the main [CNI Reconciler](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/reconciler/reconciler.go). These are designed to work in a reconciliation loop provided a stable Control Plane and and a client to interact with.
+
+## Requirements
+
+* Implement the [`ProviderClusterReconciler` interface](https://github.com/aws/eks-anywhere/blob/main/pkg/controller/clusters/registry.go#L12)
```
maxdrib, line 21: what is the migration process for providers using the legacy reconciler? Is there a way to "dry-run" this ProviderClusterReconciler during development? How do we decide when to switch over to the new reconciler?

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
+# Cluster Lifecycle API provider integration guide
+
+> `THIS IS A WORK IN PROGRESS`
+
+## Context
+> Note: if you haven't yet, reading the [original design doc](https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md) is a must to execute this integration.
```
maxdrib, line 6: is the expectation to continue to support CLI-based operations as well, or stop supporting that in favor of controller-based operations?

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+`ProviderClusterReconciler` has to handle all the actions needed to reconcile the workload cluster to the state defined in the eks-a CRs. It has complete control over how this is implemented and there are no restrictions imposed in this regard. That said, there are 3 main tasks generic tasks that every `ProviderClusterReconciler` has to perform, in this order:
+1. Reconcile the workload cluster Control Plane to the eks-a CR specifications
+1. Reconcile the workload cluster worker nodes to the eks-a CR specifications
+1. Reconcile the CNI to the eks-a CR specifications
+
+Each provider can add any other additional steps, in order: before, after ir even in between of these 3.
+
+The CNI reconciliation should be executed by using the Cilium reconciler (and other in the future), orchestrated by the main [CNI Reconciler](https://github.com/aws/eks-anywhere/blob/main/pkg/networking/reconciler/reconciler.go). These are designed to work in a reconciliation loop provided a stable Control Plane and and a client to interact with.
+
+## Requirements
```
taneyland, line 19: Should we also mention that they will need to ensure that the proper resources are added to the scheme [here](https://github.com/aws/eks-anywhere/blob/main/manager/main.go#L58)? Else it'll fail to build the reconcilers

docs/developer/cluster-lifecycle-api-provider-integration.md
```diff
@@ -0,0 +1,60 @@
...
+* Add your provider specific API structs to [`cluster.Config`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10). Add them as [child objects](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L88) if necessary.
+* Implement the proper [client processors](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/client_builder.go#L18) for your API structs and register them in the [default `ConfigClientBuilder`](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/config.go#L10)
+* Add validation webhooks for all provider specific CRDs
+* Add mutation webhooks for all provider specific CRDs when defaults are needed
+* If context aware defaults or validations (those that require any kind of API or disk call) are needed, they should be implemented in a controller for that CRD ([example](https://github.com/aws/eks-anywhere/blob/main/controllers/snow_machineconfig_controller.go)). For any extra validations, the `ProviderClusterReconciler` will need to check if the spec of the eks-a provider CRs is valid before starting the reconciliation loop. In case they are not, bubble that error to the `Cluster.status`.
```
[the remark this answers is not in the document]
maxdrib: I'm confused because the ProviderClusterReconciler for snow doesn't do anything. And the vsphere validations all seem to happen inside this ProviderClusterReconciler. Which approach is correct between these?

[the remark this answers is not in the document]
maxdrib: It would also help to add a note that these ProviderClusterReconcilers need to be registered in https://github.com/aws/eks-anywhere/blob/main/manager/main.go#L112

docs/developer/cluster-lifecycle-api-provider-integration.md
g-gaston: The providers are in full control of the reconcile logic flow. They can implement that logic however the please, in which over they need.

docs/developer/cluster-lifecycle-api-provider-integration.md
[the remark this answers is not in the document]
g-gaston: re preflight checks: that depends on what preflight checks we are talking about. In general, anything that validates the data stored in the objects and can prevent some kind of failure should be included.

re snow: the snow cluster reconciler checks that the spec for the machine configs is valid. In general, I wouldn't use the vsphere implementation as a guide, it needs to change in some places, snow is probably a better example. That said, each provider can have different requirements/constrains.

re registering in main: where do you see that? I probably missed it. In theory, `ProviderClusterReconcilers` only need to be registered in the factory. It's only the controller-runtime reconcilers (which in fact are controllers) that need to be registered in main (with the controller manager).

docs/developer/cluster-lifecycle-api-provider-integration.md
g-gaston: That's a good topic. However, I am of the opinion that tooling should be built when it's needed.

It's definitely possible to run tilt with a debugger, but we haven't had the need yet. And yes, I would say that's probably the right solution to the problem you are describing. We will just need to build it when we get there.

docs/developer/cluster-lifecycle-api-provider-integration.md
g-gaston: What do you mean by "dry-run"?

Once the new implementation is ready for a provider, we will update the CLI to enable it when installing the controller manager deployment.

For development, we have a a feature flag. You can enable it by setting an env var either before management cluster creation or before running tilt.

- [ ] I'll add some details

docs/developer/cluster-lifecycle-api-provider-integration.md
g-gaston: Great question

TBD if we continue to support CLI commands to manage workload clusters. My guess is that we will. However, we should change the internal implementation of CLI to make use of the controller, by just applying the changes with kubectl.

maxdrib (approved):

taneyland (approved):

mitalipaygude (approved):

PR #3259, opened 2022-09-03
Add snowball aws go sdk v2

jiayiwang7:
*Issue #, if available:*

#3158 

*Description of changes:*

We need to run some of the snow validations with aws snowball go sdk methods such as
- DescribeDevice
- DescribeDNI

Currently the snowball client is not publicly available in the [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2) module. We need to manually add the snowball client code in eks-a repo as a temporary solution.

*Testing (if applicable):*

N/A


go.mod
```diff
@@ -178,6 +179,9 @@ require (
 
 // TODO: Once the repo is public, remove this so we use a versioned module
 replace (
+	github.com/aws/eks-anywhere/internal/aws-sdk-go-v2/internal/configsources => ./internal/aws-sdk-go-v2/internal/configsources
+	github.com/aws/eks-anywhere/internal/aws-sdk-go-v2/internal/endpoints/v2 => ./internal/aws-sdk-go-v2/internal/endpoints/v2
+	github.com/aws/eks-anywhere/internal/aws-sdk-go-v2/service/snowballdevice => ./internal/aws-sdk-go-v2/service/snowballdevice
```
vivek-koppuru, line 184: Why are these necessary? Isn't that path already under the module?

g-gaston, line 184: Can't we just do a replace for the whole module?
```suggestion
	github.com/aws/eks-anywhere/internal/aws-sdk-go-v2 => ./internal/aws-sdk-go-v2
```

.golangci.yml
```diff
@@ -5,6 +5,7 @@ run:
   skip-dirs:
   - ".*/mocks"
   - "manager/tilt_modules"
+  - "internal/aws-sdk-go-v2"
```
g-gaston, line 8: Can we add this to the codecov config too?

go.mod
jiayiwang7: it needs to be replaced or it will look for the internal/aws-sdk-go-v2 from github which does not exist yet

jiayiwang7: there a 3 submodules under `internal/aws-sdk-go-v2` thats why i replace all

g-gaston (approved):

PR #3281, opened 2022-09-06
kubeconfig: Modify FromEnvironment to handle lists

ewollesen:
The KUBECONFIG environment variable is defined to contain a list of filenames. If it is a list, return the first one. If it's not a list, returns the entire string (as before).

In theory, a program using this environment variable would merge the kubeconfig files listed into a larger structure containing contexts, users, and clusters. At this time however, the eksctl-anywhere command is always expecting a single user/context/cluster to be defined in a single kubeconfig file.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -5,6 +5,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #3288, opened 2022-09-07
Handle Nutanix Provider in create cluster and generate clusterconifg commands

thunderboltsid:
Add the case for adding Nutanix DatacenterConfig and MachineConfig
to cluster config options when generate clusterconfig and create cluster commands
are run for Nutanix provider with NUTANIX_PROVIDER environment variable
set to true.


abhinavmpandey08 (approved):
looks good!

PR #3289, opened 2022-09-07
Handle the nutanix provider case in cluster reconciler

thunderboltsid:
Currently, this is more of a placeholder to start with as the TemplateResources method
just returns an empty array.

### DIFF
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -30,6 +30,7 @@ type clusterReconciler struct {
 	cloudStackTemplate   CloudStackTemplate
 	dockerTemplate       DockerTemplate
 	awsIamConfigTemplate AWSIamConfigTemplate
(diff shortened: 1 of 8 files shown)


thunderboltsid:
@abhinavmpandey08 the test are passing, anything else we need here?

controllers/resource/reconciler.go
```diff
@@ -146,6 +151,37 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 			return err
 		}
 		resources = append(resources, r...)
+	case anywherev1.NutanixDatacenterKind:
```
abhinavmpandey08, line 154: Can you add a feature gate here similar to this? https://github.com/aws/eks-anywhere/blob/main/controllers/resource/reconciler.go#L109-L111

abhinavmpandey08 (approved):
/approve

PR #3291, opened 2022-09-07
Add the implementation for Nutanix provider

thunderboltsid:
Quite a few methods have not been implemented in the provider and have been tagged with `TODO(nutanix)` to be implemented in future PRs.


chrisdoherty4:
Holding until we have the CI in place.

/hold

thunderboltsid:
> Holding until we have the CI in place.
> 
> /hold

Can you please elaborate on what's needed here?

chrisdoherty4:
@thunderboltsid We don't have a functional Nutanix CI yet. Once we have that and can validate the core provider logic, we can merge. @abhinavmpandey08 Is making great progress on getting the CI functional so it shouldn't be long.

abhinavmpandey08:
We have an environment where we can test Nutanix Provider now. Unholding
/unhold

pkg/api/v1alpha1/awsiamconfig_test.go
```diff
@@ -10,7 +10,7 @@ import (
 func TestGetAndValidateAWSIamConfig(t *testing.T) {
 	c := &Cluster{
 		ObjectMeta: metav1.ObjectMeta{
-			ClusterName: "eksa-unit-test-cluster",
```
thunderboltsid, line 13: Was getting linter error here as the field does not exist.

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
+piVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
abhinavmpandey08, line 1:
```suggestion
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
...
+kind: NutanixCluster
+metadata:
+  name: "{{.clusterName}}"
+  namespace: "{{.eksaSystemNamespace}}"
+spec:
+  prismCentral:
+    address: "{{.nutanixEndpoint}}"
+    port: {{.nutanixPort}}
+    insecure: true
```
abhinavmpandey08, line 10: Can we make this configurable instead of hardcoding it to true?

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
...
+    credentialRef:
+      name: "{{.clusterName}}"
+      kind: Secret
+  controlPlaneEndpoint:
+    host: "{{.controlPlaneEndpointIp}}"
+    port: 6443
+---
+apiVersion: v1
+kind: Secret
```
abhinavmpandey08, line 19: Instead of keeping the creds in this file, which gets written to disk, can we have it in a different file where we can just apply it from bytes?
Here's how we handle vSphere creds for reference https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/config/secret.yaml#L1-L9

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
...
+metadata:
+  name: "{{.clusterName}}"
+  namespace: "{{.eksaSystemNamespace}}"
+stringData:
+  NUTANIX_PASSWORD: "{{.nutanixPassword}}"
+  NUTANIX_USER: "{{.nutanixUser}}"
```
abhinavmpandey08, line 25: Instead of using `stringData`, can we just use `data` and base64 encode the creds? This ensures the creds are not displayed in plain text and also handles any special characters that the creds might have.
Here's an example of how we do it for other creds https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/config/secret.yaml#L8-L9

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
...
+  endpoints: []
+  caFile: "/etc/kubernetes/pki/etcd/ca.crt"
+  certFile: "/etc/kubernetes/pki/apiserver-etcd-client.crt"
+  keyFile: "/etc/kubernetes/pki/apiserver-etcd-client.key"
+  {{- else }}
+local:
+  imageRepository: {{.etcdRepository}}
+  imageTag: {{.etcdImageTag}}
+  {{- end }}
+files:
```
abhinavmpandey08, line 89: The indentation looks off here

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,196 @@
...
+                  - NET_RAW
+            volumeMounts:
+              - mountPath: /etc/kubernetes/admin.conf
+                name: kubeconfig
+            resources: {}
+        hostNetwork: true
+        hostAliases:
+          - hostnames:
+              - kubernetes
+            ip: 127.0.0.1
```
abhinavmpandey08, line 141: We patch the `kube-vip` we use [here](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kube-vip/kube-vip/patches/0001-use-hostname-instead-of-kubernetes-to-contact-apiser.patch) so this isn't needed

abhinavmpandey08 (changes requested):

pkg/providers/nutanix/env.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"os"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+const (
+	nutanixUsernameKey = "NUTANIX_USER"
+	nutanixPasswordKey = "NUTANIX_PASSWORD"
+	nutanixEndpointKey = "NUTANIX_ENDPOINT"
```
chrisdoherty4, line 13: Are these well known Nutanix variables or something we've decided on for EKS-A?

pkg/providers/nutanix/validator.go
```diff
@@ -0,0 +1,195 @@
...
+// NewValidator returns a new validator client
+func NewValidator(config *anywherev1.NutanixDatacenterConfig, creds basicAuthCreds) (*Validator, error) {
+	url := fmt.Sprintf("%s:%d", config.Spec.Endpoint, config.Spec.Port)
+	nutanixCreds := prismgoclient.Credentials{
+		URL:      url,
+		Username: creds.username,
+		Password: creds.password,
+		Endpoint: config.Spec.Endpoint,
+	}
+	client, err := v3.NewV3Client(nutanixCreds)
```
chrisdoherty4, line 30: Can we inject the client instead of constructing it in-line. In doing so, we create a more testable data structure and avoid twiddling internals in tests as we have to now.

pkg/providers/nutanix/validator.go
```diff
@@ -0,0 +1,195 @@
...
+			return fmt.Errorf("missing cluster name")
+		} else {
+			clusterName := *identifier.Name
+			clusterUUID, err := findClusterUUIDByName(ctx, v.v3Client, clusterName)
+			if err != nil {
+				return fmt.Errorf("failed to find cluster with name %q: %v", clusterName, err)
+			}
+
+			identifier.Type = anywherev1.NutanixIdentifierUUID
+			identifier.UUID = clusterUUID
```
chrisdoherty4, line 69: Whats the intent with these lines? It seems like we're updating some configuration, but this is a validation function so I'm not sure why it would be mutating data.

This seems to be happening in several places below.

pkg/providers/nutanix/validator.go
```diff
@@ -0,0 +1,195 @@
...
+func findSubnetUUIDByName(ctx context.Context, v3Client client, subnetName string) (*string, error) {
+	res, err := v3Client.ListSubnet(ctx, &v3.DSMetadata{
+		Filter: utils.StringPtr(fmt.Sprintf("name==%s", subnetName)),
+	})
+	if err != nil || len(res.Entities) == 0 {
+		return nil, fmt.Errorf("failed to find subnet by name %q: %v", subnetName, err)
+	}
+
+	if len(res.Entities) > 1 {
+		logr.FromContextOrDiscard(ctx).Info("Found more than one (%v) subnet with name %q.", len(res.Entities), subnetName)
```
chrisdoherty4, line 159: Avoid extracting loggers from context. Instead, can we inject it as an explicit parameter so its clear there's a dependency here.

Ditto for elsewhere in the code.

pkg/providers/nutanix/client.go
```diff
@@ -0,0 +1,18 @@
+package nutanix
+
+import (
+	"context"
+
+	v3 "github.com/nutanix-cloud-native/prism-go-client/v3"
+)
+
+type client interface {
```
chrisdoherty4, line 9: This client is used in various places and nearly always referenced with a `v3Client` variable. Can we invert that and call the interface `v3Client` and the variables just `client`. This way, the emphasis is on the expected type, not the variable name.

pkg/providers/nutanix/kubectl.go
```diff
@@ -0,0 +1,33 @@
...
+	corev1 "k8s.io/api/core/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	kubeadmv1beta1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/types"
+)
+
+type ProviderKubectlClient interface {
```
chrisdoherty4, line 16: This looks like a copy-n-paste interface with several of the behaviors not relevant to the Nutanix provide; lets clean it up!

pkg/cluster/config.go
```diff
@@ -84,6 +92,20 @@ func (c *Config) DeepCopy() *Config {
...
 	}
 
+	if c.NutanixMachineConfigs != nil {
+		c2.NutanixMachineConfigs = make(map[string]*anywherev1.NutanixMachineConfig, len(c.NutanixMachineConfigs))
+	}
+	for k, v := range c.NutanixMachineConfigs {
+		c2.NutanixMachineConfigs[k] = v.DeepCopy()
+	}
+
+	if c.SnowMachineConfigs != nil {
```
chrisdoherty4, line 102: Were the Snow adjustments intentional?

chrisdoherty4 (changes requested):

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: good catch

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: good catch; for sure. We want this configurable only for the case where user doesn't have publicly trusted CA certs.

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: Good catch, the `if/else` blocks seem to be messing with my editor.

pkg/providers/nutanix/provider.go
```diff
@@ -0,0 +1,371 @@
...
+	eksaNutanixMachineResourceType    = fmt.Sprintf("nutanixmachineconfigs.%s", v1alpha1.GroupVersion.Group)
+	requiredEnvs                      = []string{nutanixEndpointKey, nutanixUsernameKey, nutanixPasswordKey}
+)
+
+type basicAuthCreds struct {
+	username string
+	password string
+}
+
+type nutanixProvider struct {
```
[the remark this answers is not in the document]
thunderboltsid: Sure

pkg/providers/nutanix/env.go
thunderboltsid: Same env vars as we use for CAPX.

pkg/cluster/config.go
thunderboltsid: Yes, they were causing deep equal to fail.

pkg/providers/nutanix/validator.go
thunderboltsid: Good catch; we don't need that here.

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: We will be redoing the creds in another PR (including the format) in a week or so hopefully based on the standardized creds format we use in other k8s distros. Can we punt on this for now?

Makefile
```diff
@@ -480,6 +480,7 @@ mocks: ## Generate mocks
 	${GOPATH}/bin/mockgen -destination=pkg/helm/mocks/download.go -package=mocks -source "pkg/helm/download.go"
 	${GOPATH}/bin/mockgen -destination=pkg/aws/mocks/ec2.go -package=mocks -source "pkg/aws/ec2.go"
 	${GOPATH}/bin/mockgen -destination=pkg/aws/mocks/snowballdevice.go -package=mocks -source "pkg/aws/snowballdevice.go"
+	${GOPATH}/bin/mockgen -destination=pkg/providers/nutanix/mock_client.go -package=nutanix -source "pkg/providers/nutanix/client.go"
```
chrisdoherty4, line 483: Was this intentionally added to the `nutanix` package making it part of its API?

go.mod
```diff
@@ -51,6 +52,21 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
```
chrisdoherty4, line 55: Does this need a `go mod tidy`? Its peculiar to have more than 2 `require` blocks.

thunderboltsid: I did run go mod tidy and that didn't do anything. I can manually move it into two blocks: one for direct and one for indirect.

pkg/cluster/nutanix_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/yaml"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+//go:embed testdata/nutanix/eksa-cluster.yaml
+var nutanicClusterConfigSpec string
```
abhinavmpandey08, line 16: nit
```suggestion
var nutanixClusterConfigSpec string
```

Makefile
abhinavmpandey08: This is just to add generate mocks. This doesn't add `nutanix` package to any APIs

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: NutanixCluster
+metadata:
+  name: "{{.clusterName}}"
+  namespace: "{{.eksaSystemNamespace}}"
+spec:
+  prismCentral:
+    address: "{{.nutanixEndpoint}}"
+    port: {{.nutanixPort}}
+    insecure: {{.nutanixInsecure}}
```
abhinavmpandey08, line 10: Is there a way to use TLS even with self-signed certs instead of `insecure`? 
Doesn't have to block this PR but just curious if this is something that can be handled without skipping TLS by passing the CA cert in the config

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+      users:
+        - name: "{{ .controlPlaneSshUsername }}"
+          lockPassword: false
+          sudo: ALL=(ALL) NOPASSWD:ALL
+          sshAuthorizedKeys:
+            - "{{.controlPlaneSshAuthorizedKey}}"
+      preKubeadmCommands:
+        - echo "before kubeadm call" > /var/log/prekubeadm.log
+      postKubeadmCommands:
+        - echo "after kubeadm call" > /var/log/postkubeadm.log
```
abhinavmpandey08, line 182: This seems like a debugging this. Is this needed for something else?

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+      verbosity: 10
```
abhinavmpandey08, line 183: Is this needed? Seems like a really high verbosity. I would leave it out entirely

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+      - name: "{{.controlPlaneSshUsername }}"
+        lockPassword: false
+        sudo: ALL=(ALL) NOPASSWD:ALL
+        sshAuthorizedKeys:
+          - "{{.controlPlaneSshAuthorizedKey}}"
+    preKubeadmCommands:
+      - echo "before kubeadm call" > /var/log/prekubeadm.log
+    postKubeadmCommands:
+      - echo export KUBECONFIG=/etc/kubernetes/admin.conf >> /root/.bashrc
+      - echo "after kubeadm call" > /var/log/postkubeadm.log
```
abhinavmpandey08, line 154: Any of this required?

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+    useExperimentalRetryJoin: true
+    verbosity: 10
```
abhinavmpandey08, line 156: Let's take this out

pkg/providers/nutanix/config/cp-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
```
abhinavmpandey08, line 159: can we move this to template-md.yaml?

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: That's the plan, Abhinav! However, at the moment, CAPX doesn't support taking in a additional trust bundle. EKS-A codebase already has the wiring needed for whenever CAPX will support an additional trust bundle. Hopefully in the coming weeks.

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: yeah, for debugging. we can remove it.

pkg/providers/nutanix/config/cp-template.yaml
thunderboltsid: not really.

pkg/providers/nutanix/config/cp-template.yaml
chrisdoherty4: Could you raise an issue to address explicitly so we don't miss. Thanks.

Makefile
chrisdoherty4: Not quite. Its generating a file used in `go build` with the package name `nutanix` which means the mock data structures are part of the `nutanix` package API. I'm not sure if that was on purpose or accidental. Normally, you'd generate a `mock_client_test.go` (`_test` etc) or put it in a separate `mocks` package (which I'm not a fan of). 

What's the intent here? Having it as part of the package API seems odd but there may be a use-case if its needed externally.

thunderboltsid: That's a fair point given the mock is being checked in rather than being generated on the fly. I've renamed the generated mock file with a `_test` suffix. Not a fan of a separate mocks package either.

pkg/dependencies/factory.go
```diff
@@ -403,6 +407,38 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+				URL:      url,
+				Username: os.Getenv("NUTANIX_USER"),
+				Password: os.Getenv("NUTANIX_PASSWORD"),
+				Endpoint: datacenterConfig.Spec.Endpoint,
+				Port:     fmt.Sprintf("%d", datacenterConfig.Spec.Port),
+			}
+			client, err := v3.NewV3Client(nutanixCreds)
+			if err != nil {
+				return fmt.Errorf("error creating nutanix client: %v", err)
+			}
```
abhinavmpandey08, line 432: Can you move this section to a `With..()` func?
Maybe you can call it `WithPrismClient() *Factory` and make prism client a part of the `Dependencies` struct https://github.com/aws/eks-anywhere/blob/5dffa8f12fb2e666b4c04b9d5d450395d3ddad3b/pkg/dependencies/factory.go#L48

Also, can we check if the env vars are actually set using `os.LookupEnv` instead of `os.Getenv()`?

pkg/dependencies/factory.go
```diff
@@ -403,6 +407,38 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+
+			validator, err := nutanix.NewValidator(client.V3)
+			if err != nil {
+				return err
+			}
```
abhinavmpandey08, line 437: any reason to initialize this here instead of from within the provider? It only needs the client which is already being passed to the provider

pkg/executables/config/clusterctl.yaml
```diff
@@ -81,6 +85,9 @@ images:
   infrastructure-docker/capd-manager:
     repository: {{.DockerManagerRepository}}
     tag: {{.DockerManagerTag}}
+  infrastructure-nutanix/manager:
+    repository: {{ .ClusterApiNutanixControllerRepository }}
+    tag: {{ .ClusterApiNutanixControllerTag }}
```
abhinavmpandey08, line 90: nit: can you move this below `infrastructure-docker/kube-rbac-proxy` so the two docker ones are together

Also, does CAPX need `kube-rbac-proxy`? If so, you might wanna add that here as well

pkg/executables/testdata/nutanix/datacenterConfig.json
```diff
@@ -0,0 +1,12 @@
...
+  "kind": "NutanixDatacenterConfig",
+  "metadata": {
+    "name": "eksa-unit-test",
+    "namespace": "default"
+  },
+  "spec": {
+    "endpoint": "prism.nutanix.com",
+    "port": 9440
+  }
+}
```
abhinavmpandey08, line 12: nit: add new line at the end of the files to avoid the red circle

pkg/executables/testdata/nutanix/datacenterConfigs.json
```diff
@@ -0,0 +1,16 @@
...
+        "name": "eksa-unit-test",
+        "namespace": "default"
+      },
+      "spec": {
+        "endpoint": "prism.nutanix.com",
+        "port": 9440
+      }
+    }
+  ]
+}
```
abhinavmpandey08, line 16: nit: add new line at the end of the files to avoid the red circle

pkg/executables/testdata/nutanix/machineConfig.json
```diff
@@ -0,0 +1,35 @@
...
+    "users": [
+      {
+        "name": "mySshUsername",
+        "sshAuthorizedKeys": [
+          "mySshAuthorizedKey"
+        ]
+      }
+    ]
+  }
+}
```
abhinavmpandey08, line 35: nit: add new line at the end of the files to avoid the red circle

pkg/dependencies/factory.go
thunderboltsid: Nothing in particular, can be moved into the provider.

pkg/executables/config/clusterctl.yaml
thunderboltsid: Sure; We haven't added the KubeRBACProxy to the nutanix bundle yet. If needed, we can take care of that in a future PR.

pkg/providers/nutanix/provider.go
```diff
@@ -0,0 +1,614 @@
...
+	return AnyImmutableFieldChanged(oldNmc, newNmc)
+}
+
+func nutanixIdentifierChanged(old, new v1alpha1.NutanixResourceIdentifier) bool {
+	if old.Type == v1alpha1.NutanixIdentifierName && old.Name != nil && new.Name != nil && *old.Name != *new.Name {
+		return true
+	}
+	if old.Type == v1alpha1.NutanixIdentifierUUID && old.UUID != nil && new.UUID != nil && *old.UUID != *new.UUID {
+		return true
+	}
```
abhinavmpandey08, line 244: don't we also wanna check if the new type changed from `NutanixIdentifierName` type to `NutanixIdentifierUUID` or vice-versa?

chrisdoherty4 (commented):
@abhinavmpandey08 will give the final approval.

pkg/providers/nutanix/provider.go
```diff
@@ -0,0 +1,614 @@
...
+	if oldNmc.Spec.VCPUSockets != newNmc.Spec.VCPUSockets {
+		return true
+	}
+	if oldNmc.Spec.VCPUsPerSocket != newNmc.Spec.VCPUsPerSocket {
+		return true
+	}
+	if oldNmc.Spec.OSFamily != newNmc.Spec.OSFamily {
+		return true
+	}
```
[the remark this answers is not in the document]
thunderboltsid: These refer to the fields that we can't swap out in-place from a CAPX standpoint.

pkg/providers/nutanix/provider.go
thunderboltsid: Good catch; updated the logic with a identifier type comparison in the beginning of the function

abhinavmpandey08 (approved):
Amazing work @thunderboltsid!
PR looks good to me
/lgtm

PR #3292, opened 2022-09-07
Add NutanixBundle type to VersionsBundle

thunderboltsid:
NutanixBundle contains the cluster API controller image and version.
Also updated the corresponding bundles CRD and testdata files.


thunderboltsid:
I guess we need to check-in add the cluster-api-provider for nutanix here https://github.com/aws/eks-anywhere-build-tooling

abhinavmpandey08:
> I guess we need to check-in add the cluster-api-provider for nutanix here https://github.com/aws/eks-anywhere-build-tooling

yep, we generally build our own images using our base image. Here's an example of how we do it for `cluster-api-provider-vsphere` https://github.com/aws/eks-anywhere-build-tooling/tree/main/projects/kubernetes-sigs/cluster-api-provider-vsphere
You should be able to re-use most of it for CAPX

thunderboltsid:
> > I guess we need to check-in add the cluster-api-provider for nutanix here https://github.com/aws/eks-anywhere-build-tooling
> 
> yep, we generally build our own images using our base image. Here's an example of how we do it for `cluster-api-provider-vsphere` https://github.com/aws/eks-anywhere-build-tooling/tree/main/projects/kubernetes-sigs/cluster-api-provider-vsphere You should be able to re-use most of it for CAPX

@abhinavmpandey08 We have a PR for the build tooling and corresponding Prow job:
- https://github.com/aws/eks-anywhere-build-tooling/pull/1313
- https://github.com/aws/eks-anywhere-prow-jobs/pull/199

release/pkg/test/testdata/main-bundle-release.yaml
```diff
@@ -1166,6 +1175,15 @@ spec:
...
     kubeVersion: "1.21"
+    nutanix:
+      clusterAPIController: {}
+      clusterTemplate:
+        uri: https://release-bucket/artifacts/v0.0.0-dev-build.0/cluster-api-provider-nutanix/manifests/infrastructure-nutanix/v0.5.1/cluster-template.yaml
+      components:
+        uri: https://release-bucket/artifacts/v0.0.0-dev-build.0/cluster-api-provider-nutanix/manifests/infrastructure-nutanix/v0.5.1/infrastructure-components.yaml
+      metadata:
+        uri: https://release-bucket/artifacts/v0.0.0-dev-build.0/cluster-api-provider-nutanix/manifests/infrastructure-nutanix/v0.5.1/metadata.yaml
+      version: v0.5.1+abcdef1
```
deepakm-ntnx, line 1186: @thunderboltsid where is +abcdedf1 coming from?

thunderboltsid: 🤷‍♂️

thunderboltsid: It's an autogenerated file.

release/pkg/bundles/bundles.go
```diff
@@ -175,6 +175,14 @@ func GetVersionsBundles(r *releasetypes.ReleaseConfig, imageDigests map[string]s
 			return nil, errors.Wrapf(err, "Error getting bundle for vSphere infrastructure provider")
 		}
 
+		var nutanixBundle anywherev1alpha1.NutanixBundle
```
abhay-krishna, line 178: this check can be moved out of this `for` loop since Nutanix bundle does not have separate artifacts for each Kubernetes version. There is a similar check for Snow bundle above on line 140, you can add it to that block.

release/pkg/assets/config/bundle_release.go
```diff
@@ -813,6 +815,32 @@ var bundleReleaseAssetsConfigMap = []assettypes.AssetConfig{
...
+				{
+					Name:          "infrastructure-nutanix",
+					ManifestFiles: []string{"infrastructure-components.yaml", "cluster-template.yaml", "metadata.yaml"},
+				},
+			},
+			OnlyForDevRelease: true,
+		}
+		conf = append(conf, nutanixConfig)
+	}
+	return conf
```
abhinavmpandey08, line 845: we don't need a special check here since `OnlyForDevRelease` already handles this.
Can you move this section inside the `bundleReleaseAssetsConfigMap` array? 
It should be similar to this https://github.com/aws/eks-anywhere/blob/main/release/pkg/assets/config/bundle_release.go#L179-L200

release/pkg/assets/config/bundle_release.go
```diff
@@ -813,6 +815,32 @@ var bundleReleaseAssetsConfigMap = []assettypes.AssetConfig{
 	},
 }
 
-func GetBundleReleaseAssetsConfigMap() []assettypes.AssetConfig {
-	return bundleReleaseAssetsConfigMap
+func GetBundleReleaseAssetsConfigMap(rc *releasetypes.ReleaseConfig) []assettypes.AssetConfig {
+	conf := bundleReleaseAssetsConfigMap
+	if rc.DevRelease && rc.BuildRepoBranchName == "main" {
```
abhay-krishna, line 820: This check is performed in assets/assets.go, and since we set `OnlyForDevRelease` to `true`, this AssetConfig can just be added to the above list.

release/pkg/test/testdata/main-bundle-release.yaml
abhay-krishna: It comes from constants.go, it's a dummy value. We want to make the testdata file deterministic and reproducible.

release/pkg/bundles/nutanix.go
```diff
@@ -0,0 +1,105 @@
...
+				manifestHash, err := version.GenerateManifestHash(r, manifestArtifact)
+				if err != nil {
+					return anywherev1alpha1.NutanixBundle{}, err
+				}
+				artifactHashes = append(artifactHashes, manifestHash)
+			}
+		}
+	}
+
+	nutanixCloudProviderArtifacts := r.BundleArtifactsTable[fmt.Sprintf("cloud-provider-nutanix-%s", eksDReleaseChannel)]
```
abhay-krishna, line 66: why do we need this? there's no key for `cloud-provider-nutanix`, also all CAPX artifacts have been covered above.

release/pkg/bundles/nutanix.go
```diff
@@ -0,0 +1,105 @@
...
+		version.NewVersionerWithGITTAG(r.BuildRepoSource, constants.CapxProjectPath, sourceBranch, r),
+		componentChecksum,
+	)
+	if err != nil {
+		return anywherev1alpha1.NutanixBundle{}, errors.Wrapf(err, "Error getting version for cluster-api-provider-nutanix")
+	}
+
+	bundle := anywherev1alpha1.NutanixBundle{
+		Version:              version,
+		ClusterAPIController: bundleImageArtifacts["cluster-api-nutanix-controller"],
```
abhay-krishna, line 98: This should match the `imageArtifact.AssetName` set in config/bundle_release.go since that's what is being used as the map key above.
```suggestion
		ClusterAPIController: bundleImageArtifacts["cluster-api-provider-nutanix"],
```

release/pkg/bundles/nutanix.go
```diff
@@ -0,0 +1,105 @@
...
+
+	anywherev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+	"github.com/aws/eks-anywhere/release/pkg/constants"
+	releasetypes "github.com/aws/eks-anywhere/release/pkg/types"
+	bundleutils "github.com/aws/eks-anywhere/release/pkg/util/bundles"
+	"github.com/aws/eks-anywhere/release/pkg/version"
+)
+
+// GetNutanixBundle returns the bundle for Nutanix
+func GetNutanixBundle(r *releasetypes.ReleaseConfig, eksDReleaseChannel string, imageDigests map[string]string) (anywherev1alpha1.NutanixBundle, error) {
```
abhay-krishna, line 16: we don't need to pass eksDReleaseChannel here as there's nothing EKS-D specific here.

release/pkg/assets/config/bundle_release.go
thunderboltsid: If you take a look at the test, the `release-0.11` is also used in the test as a "dev" release which is why I added the check for branch.

release/pkg/assets/config/bundle_release.go
thunderboltsid: If you take a look at `TestGenerateBundleManifest`,[ the release config](https://github.com/aws/eks-anywhere/blob/15a1cd2a1229abece283f24a2b499ee577b96006/release/pkg/operations/bundle_release_test.go#L59) has `DevRelease` set to true for both `main` and `release-0.11` branch. As such, it tries to fetch a GIT tag for a repo that doesn't exist and fails.

release/pkg/assets/config/bundle_release.go
thunderboltsid: If you take a look at TestGenerateBundleManifest,[ the release config](https://github.com/aws/eks-anywhere/blob/15a1cd2a1229abece283f24a2b499ee577b96006/release/pkg/operations/bundle_release_test.go#L59) has DevRelease set to true for both main and release-0.11 branch. As such, it tries to fetch a git tag for a project that doesn't exist in that branch and it fails.

release/pkg/assets/config/bundle_release.go
```diff
@@ -811,8 +813,30 @@ var bundleReleaseAssetsConfigMap = []assettypes.AssetConfig{
...
+				Name:          "infrastructure-nutanix",
+				ManifestFiles: []string{"infrastructure-components.yaml", "cluster-template.yaml", "metadata.yaml"},
+			},
+		},
+		OnlyForDevRelease: true,
+	},
 }
 
-func GetBundleReleaseAssetsConfigMap() []assettypes.AssetConfig {
+func GetBundleReleaseAssetsConfigMap(rc *releasetypes.ReleaseConfig) []assettypes.AssetConfig {
```
abhay-krishna, line 840: We no longer need to pass in ReleaseConfig as a parameter.

release/pkg/assets/assets.go
```diff
@@ -99,7 +99,7 @@ func getAssetsFromConfig(ac *assettypes.AssetConfig, rc *releasetypes.ReleaseCon
 
 func GetBundleReleaseAssets(supportedK8sVersions []string, eksDReleaseMap *filereader.EksDLatestReleases, rc *releasetypes.ReleaseConfig) (map[string][]releasetypes.Artifact, error) {
 	artifactsTable := map[string][]releasetypes.Artifact{}
-	assetConfigs := assetconfig.GetBundleReleaseAssetsConfigMap()
+	assetConfigs := assetconfig.GetBundleReleaseAssetsConfigMap(rc)
```
abhay-krishna, line 102: We no longer need to pass in ReleaseConfig as a parameter.

release/pkg/assets/config/bundle_release.go
thunderboltsid: Commenting out `release-0.11` test case for now.

PR #3309, opened 2022-09-08
feat(CLI): Add ValidateFileOrEnv function in kubeconfig package

ewollesen:
ValidateFileOrEnv validates a given kubeconfig filename, or, if and only if that filename is an empty string, falls back to a kubeconfig filename from the environment, and validates it.

This should be useful for CLI commands that want to find the first specified kubeconfig file that's fit to use.

While this isn't used yet, my next PR will use it. :)

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -48,17 +48,51 @@ func ValidateFile(filename string) error {
 		return fmt.Errorf("validating kubeconfig %q: %w", filename, err)
 	}
 
-	if !validations.FileExists(filename) {
+	// Trim whitespace from the beginning and end of the filename. While these
+	// could technically be valid filenames, it's far more likely a typo or
+	// shell-parsing bug.
+	trimmed := strings.TrimSpace(filename)
+
+	if !validations.FileExists(trimmed) {
 		return wrapError(fs.ErrNotExist)
 	}
 
-	if !validations.FileExistsAndIsNotEmpty(filename) {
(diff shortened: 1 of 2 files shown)


pkg/kubeconfig/kubeconfig.go
```diff
@@ -48,17 +48,51 @@ func ValidateFile(filename string) error {
...
 
-	if _, err := clientcmd.LoadFromFile(filename); err != nil {
+	if _, err := clientcmd.LoadFromFile(trimmed); err != nil {
 		return wrapError(err)
 	}
 
 	return nil
 }
+
+// ValidateFileOrEnv validates the given filename if it's not empty. If it's
```
a-cool-train, line 71: I don't think the comment is necessary here. The method is pretty self explantory.

ewollesen: I agree with @chrisdoherty4 . It's not something our teams have been particularly good about in the past, but it is a good practice. 

Sometimes it's difficult to come up with a good, useful comment about some of the more obvious functions. Sometimes it's hard because you change the function several times while you're editing the code, and so it feels like a waste of time. But in a few months, when you haven't seen that code for a while (or when it's your first time reading someone else's code), you'll be very glad for a thoughtful well-written comment that gives you a heads up into the _why_ behind the function, not just the what.

Not saying this is a particularly good or thoughtful comment. I'm still working on providing more insight with my comments too, but it's one of the little things I try to improve at.

PR #3312, opened 2022-09-08
Fix workernode inconsistency upgrade

maxdrib:
*Issue #, if available:*

*Description of changes:*
After the diskOffering field was made optional in https://github.com/aws/eks-anywhere/pull/3181, it resulted in some changes to the way we serialized/deserialized the field when mapping from the CloudStackMachineTemplate to the CloudStackMachineConfig. When the EKS-A controller was unpaused on cluster creation, it resulted in the following inconsistency

```
[DEBUG] Inequal DiskOffering between oldCSMC (&{{ } %!s(int64=0)    }) and newCSMC (%!s(*v1alpha1.CloudStackResourceDiskOffering=<nil>)) for csmc
```

resulting in the creation of a new CloudStackMachineTemplate for worker nodes. This PR modifies the way we map the CloudStackMachineTemplate so that if the DiskOffering isn't present on the CloudStackMachineTemplate, we do not set it on the CloudStackMachineConfig

*Testing (if applicable):*
Created cluster with overridden EKS-A controller (built from local) and ran TestCloudStackKubernetes120RedhatTo121Upgrade. The issue was no longer observed

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -34,10 +34,11 @@ spec:
           httpGet:
             path: /healthz
             port: healthz
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
       terminationGracePeriodSeconds: 10
       tolerations:
         - effect: NoSchedule
           key: node-role.kubernetes.io/master
       serviceAccountName: manager
-      securityContext:
-        runAsNonRoot: true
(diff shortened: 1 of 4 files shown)


g-gaston (approved):

jaxesn (approved):
/approve

PR #3314, opened 2022-09-09
Add validations with snowballdevice sdk

### LINKED ISSUE #3158: Run snow validation with Snow device SDK

jiayiwang7:
*Issue #, if available:*

Resolves #3158 

*Description of changes:*

Add snow preflight validations in CLI w snowballdevice sdk:
- ValidateDeviceIsUnlocked
- ValidateDeviceSoftware

Remove ValidateMachineDeviceIPs which is already covered by any other snow validation.

*Testing (if applicable):*

Unit tests


controllers/snow_machineconfig_controller.go
```diff
@@ -81,9 +80,6 @@ func (r *SnowMachineConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 func (r *SnowMachineConfigReconciler) reconcile(ctx context.Context, snowMachineConfig *anywherev1.SnowMachineConfig) (_ ctrl.Result, reterr error) {
 	var allErrs []error
-	if err := r.validator.ValidateMachineDeviceIPs(ctx, snowMachineConfig); err != nil {
```
vivek-koppuru, line 84: Are we doing this because now we check for this in the other two validate functions?

pkg/aws/client.go
```diff
@@ -63,3 +65,9 @@ func NewClientFromEC2(ec2 EC2Client) *Client {
 		ec2: ec2,
 	}
 }
+
+func NewClientFromSnowball(snowballdevice SnowballDeviceClient) *Client {
+	return &Client{
```
vivek-koppuru, line 70: If we are doing something like this instead, is it possible to separate the structs for `Client` and do the following?

```
type Client struct {
   *EC2Client
   *SnowballDeviceClient
}
```
That way, we can have snowball and ec2 related member methods separately, and still utilize a common Client if needed by the consumers. However, I am just throwing an idea and not sure if it works for this case.

pkg/aws/client.go
```diff
@@ -63,3 +65,9 @@ func NewClientFromEC2(ec2 EC2Client) *Client {
 		ec2: ec2,
 	}
 }
+
+func NewClientFromSnowball(snowballdevice SnowballDeviceClient) *Client {
```
g-gaston, line 69: What's the purpose of this constructor? it seems to build an incomplete `Client`

pkg/aws/snowballdevice.go
```diff
@@ -0,0 +1,40 @@
...
+type SnowballDeviceClient interface {
+	DescribeDevice(ctx context.Context, params *snowballdevice.DescribeDeviceInput, optFns ...func(*snowballdevice.Options)) (*snowballdevice.DescribeDeviceOutput, error)
+	DescribeDeviceSoftware(ctx context.Context, params *snowballdevice.DescribeDeviceSoftwareInput, optFns ...func(*snowballdevice.Options)) (*snowballdevice.DescribeDeviceSoftwareOutput, error)
+}
+
+func NewSnowballClient(config aws.Config) *snowballdevice.Client {
+	return snowballdevice.NewFromConfig(config)
+}
+
+func (c *Client) SnowballDeviceUnlocked(ctx context.Context) (bool, error) {
```
g-gaston, line 26: nit:
```suggestion
func (c *Client) IsSnowballDeviceUnlocked(ctx context.Context) (bool, error) {
```

controllers/snow_machineconfig_controller.go
jiayiwang7: yes, also the new ValidateDeviceIsUnlocked should already checked this too.

pkg/providers/snow/validator.go
```diff
@@ -9,6 +9,7 @@ import (
 
 const (
 	defaultAwsSshKeyName = "eksa-default"
+	minSoftwareVersion   = "102"
```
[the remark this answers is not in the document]
jiayiwang7: yes, snowballdevice software version

pkg/aws/client.go
jiayiwang7: for unit tests

pkg/providers/snow/validator.go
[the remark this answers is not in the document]
vivek-koppuru: Since this file is not only for snowball client, can we add that to the name?

vivek-koppuru (approved):

g-gaston (approved):

vivek-koppuru (approved):

PR #3324, opened 2022-09-09
clusterctl move retry

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Changes to enable retry of *clusterctl move* on networking-related failures.
- Wrap clusterctl move in exponential backoff retrier with 30m timeout.
- Unit test of retry on network-related failure.
- Factor pseudo-consts declared in kubectl.go to constants.go when retries were implemented there, for use in other places (like clusterctl.go)

Near identical approach to retries of *kubectl wait*, recently added.

*Testing (if applicable):*
Unit test of retry functionality added.  
Expected retry behavior also confirmed via debugger.


pkg/executables/clusterctl.go
```diff
@@ -26,6 +29,7 @@ const (
 	etcdadmBootstrapProviderName  = "etcdadm-bootstrap"
 	etcdadmControllerProviderName = "etcdadm-controller"
 	kubeadmBootstrapProviderName  = "kubeadm"
+	clusterctlMoveTimeout         = 30 * time.Minute
```
maxdrib, line 32: could you add a comment as to where this came from?

pkg/executables/clusterctl.go
```diff
@@ -148,12 +152,39 @@ func writeInfrastructureBundle(clusterSpec *cluster.Spec, rootFolder string, bun
...
+	}
+	const networkFaultBaseRetryTime = 10 * time.Second
+	const backoffFactor = 1.5
+	waitTime := time.Duration(float64(networkFaultBaseRetryTime) * math.Pow(backoffFactor, float64(totalRetries-1)))
+
+	if match := connectionRefusedRegex.MatchString(err.Error()); match {
+		return true, waitTime
+	}
+	if match := ioTimeoutRegex.MatchString(err.Error()); match {
+		return true, waitTime
```
maxdrib, line 170: I think illustrating some example scenarios where these errors might occur would help justify this change, i.e. "Cluster is unreachable due to X" or "Process y is not running because Z". In general, there is value to fast failure of operations and if we're retrying them it helps to explain the value added, either in comments or in the PR description.

pkg/executables/clusterctl.go
jweite-amazon: Added the requested comment.  This timeout is arbitrary.  I picked the value established by EKS-A for kubectl waits.  This duration really defines how long EKS-A will wait for a transient network disruption that occurs during a move to abate.  I'm open to suggestion about other durations.

pkg/executables/clusterctl.go
jweite-amazon: Added code comments describing the scenarios being protected against, primarily network disruptions between whatever's hosting clusterctl, and the source and target clusters it's moving between.  This also protects against situations where apiserver on these is temporarily not running, or becomes unresponsive during a move.

pkg/executables/clusterctl.go
```diff
@@ -26,11 +30,14 @@ const (
 	etcdadmBootstrapProviderName  = "etcdadm-bootstrap"
 	etcdadmControllerProviderName = "etcdadm-controller"
 	kubeadmBootstrapProviderName  = "kubeadm"
+	clusterctlMoveTimeout         = 30 * time.Minute // Arbitrarily established.  Equal to kubectl wait default timeouts.
```
vivek-koppuru, line 33: Did we check if `clusterctl` has some default timeouts on their end?

pkg/executables/clusterctl.go
```diff
@@ -148,12 +155,43 @@ func writeInfrastructureBundle(clusterSpec *cluster.Spec, rootFolder string, bun
 	return nil
 }
 
+func ClusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait time.Duration) {
```
chrisdoherty4, line 158: Is there a need to export this?

chrisdoherty4: I'm also wondering if this could be a generalized exponential backoff and live in the `retrier` package?

pkg/executables/clusterctl.go
jweite-amazon: clusterctl establishes timeouts for the individual API operations it performs, but I don't see any overarching timeout for the entire move operation, which this timeout is the equivalent of.

pkg/executables/clusterctl.go
vivek-koppuru: Yea this looks generic enough to maybe warrant a generic retry policy, and we can pass in the expected error too?

jweite-amazon: Agreed that this is possible, and would be DRYer.  This policy, and the one I created for kubectl wait, are near identical.  

The technique they use, of regex matching the the error.Error() string is distinctive to use cases where there's no more reliable way to identify the kind of error, and TBH kind of icky, so I'm hesitant to factor them out, as it implies that this is the a recommended technique for how retry policies should work.  I'd want the recommended technique uses distinctly enumerated error codes.  Keeping these "aberrations" sequestered in the modules that use them hints that they're very module-specific.

I revisited kubectl wait, and am reminded of the technique I used for aliasing an unexported method for testing (export_test.go).  So I can make this more private if still desired.

chrisdoherty4: I empathize with the error handling issue, we've not been great at managing errors. If we could unexport it we can refactor when we have better error handling. I'm not sure what technique you're referencing, but if you need any input let me know - I have several tricks.

g-gaston: I would probably just move this method to `pkg/executables/internal` so we can keep the unit test as is

g-gaston (commented):
one minor suggestion, rest looks great

pkg/executables/clusterctl.go
jweite-amazon: Thanks @chrisdoherty4 and @g-gaston.  I chose to follow the same approach I'd used recently kubectl wait (but apperently forgot about!): export clusterctlMoveRetryPolicy as ClusterctlMoveRetryPolicy in export_test.go.  Since this is a *test.go file, it's not included in non-test builds, so the name remains unexported there.  (Following from https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd)

chrisdoherty4 (approved):
lgtm

PR #3326, opened 2022-09-10
Enable node autoscaling option with Cluster Autoscaler

### LINKED ISSUE #3188: Add autoscaling option in cluster config

jiayiwang7:
*Issue #, if available:*

Resolves #3188 

*Description of changes:*

Introduce `AutoScalingConfiguration` field in `WorkerNodeGroupConfiguration`. When specified, we add the Cluster Autoscaler annotation to the CAPI `MachineDeployment` resource.

API
```go
type AutoScalingConfiguration struct {
	// MinCount defines the minimum number of nodes for the associated resource group.
	// +optional
	MinCount int `json:"minCount,omitempty"`

	// MaxCount defines the maximum number of nodes for the associated resource group.
	// +optional
	MaxCount int `json:"maxCount,omitempty"`
}
```

Also add validations for the field.

Something we need to follow up:
1. In the original design: https://github.com/aws/eks-anywhere/blob/main/designs/autoscaler.md#api-design, we plan to change `WorkerNodeGroupConfiguration.Count` to be optional. This change may introduce issues in provisioning workflow since we rely on the count to calculate the total nodes and compare it with the ready node. We will need to do more investigation on the topic.
2. This PR only covers the CAPI change in `apibuilder` and only Snow provider uses it. We need to have separate PRs to update the other CAPI machine deployment templates located in other provider code.

*Testing (if applicable):*

Unit tests

### DIFF
--- a/config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
+++ b/config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
@@ -263,6 +263,19 @@ spec:
               workerNodeGroupConfigurations:
                 items:
                   properties:
+                    autoscalingConfiguration:
+                      description: AutoScalingConfiguration defines the auto scaling
+                        configuration
+                      properties:
+                        maxCount:
+                          description: MaxCount defines the maximum number of nodes
+                            for the associated resource group.
+                          type: integer
+                        minCount:
+                          description: MinCount defines the minimum number of nodes
+                            for the associated resource group.
+                          type: integer
+                      type: object
                     count:
                       description: Count defines the number of desired worker nodes.
                         Defaults to 1.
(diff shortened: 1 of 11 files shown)


jonathanmeier5:
> 2\. This PR only covers the CAPI change in `apibuilder` and only Snow provider uses it. We need to have separate PRs to update the other CAPI machine deployment templates located in other provider code.

Are these additional provider PRs part of our current scope of work?

jonathanmeier5 (approved):
/lgtm

TerryHowe (approved):

jiayiwang7:
> > 2. This PR only covers the CAPI change in `apibuilder` and only Snow provider uses it. We need to have separate PRs to update the other CAPI machine deployment templates located in other provider code.
> 
> Are these additional provider PRs part of our current scope of work?

Yes, autoscaling should work cross providers.

PR #3334, opened 2022-09-12
Redact snow creds in verbose logs

taneyland:
*Issue #, if available:*

*Description of changes:*
Add snow creds & certs to redacted env vars

*Testing (if applicable):*
`eksctl-anywhere create cluster` with Snow & vSphere and viewed logs with redacted values

### DIFF
--- a/pkg/executables/executables.go
+++ b/pkg/executables/executables.go
@@ -5,11 +5,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"os/exec"
 	"strings"
 
 	"github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 )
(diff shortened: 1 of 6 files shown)


pkg/providers/snow/creds.go
```diff
@@ -2,27 +2,27 @@ package snow
...
-	snowCertsB64 string
-}
-
 func (p *SnowProvider) setupBootstrapCreds() error {
 	creds, err := aws.EncodeFileFromEnv(eksaSnowCredentialsFileKey)
 	if err != nil {
 		return fmt.Errorf("failed to set up snow credentials: %v", err)
 	}
-	p.bootstrapCreds.snowCredsB64 = creds
+	if err := os.Setenv(snowCredentialsKey, creds); err != nil {
```
jiayiwang7, line 15: we deliberately avoided setting the envs in the code and instead, used a struct to pass the secrets around so it can be used in different cases (both cli and controllers)

pkg/executables/executables.go
```diff
@@ -87,10 +91,10 @@ func redactCreds(cmd string) string {
 	return cmd
 }
 
-func execute(ctx context.Context, cli string, in []byte, args ...string) (stdout bytes.Buffer, err error) {
+func execute(ctx context.Context, cli string, in []byte, envVars map[string]string, args ...string) (stdout bytes.Buffer, err error) {
```
jiayiwang7, line 94: do we need to os.Setenv here?

taneyland: I thought we only called os.Setenv in order to lookup envs to redact, but setting the env was in place before we started redacting so I''m adding in back in the `Run()` method where it was originally

jiayiwang7 (approved):

PR #3337, opened 2022-09-12
Add snow identityRef support

### LINKED ISSUE #2723: Snow - use identityRef to manage custom bootstrap secret
**What would you like to be added**:

Wait for snow team to finish implementation on CAPAS first.

**Why is this needed**:

jiayiwang7:
*Issue #, if available:*

Resolves #2723 

*Description of changes:*

Use Snow identityRef similar to vSphere provider. This PR shall not break the controller workflow.

API changes

```go
type SnowDatacenterConfigSpec struct {
	IdentityRef Ref `json:"identityRef,omitempty"`
}
```

```yaml
kind: SnowDatacenterConfig
metadata:
  name: test-snow
spec:
  identityRef:
    name: test-snow-snow-credentials
    kind: Secret
```

User can specify the identityRef in snowdatacenter object and manage an EKS-A snow credentials secret in user defined namespace. EKS-A maintains an internal snow credentials secret in eksa-system namespace which acts as an "read-only" object from user point of view.

The internal secret is referred by CAPAS `AwsSnowCluster` `identityRef` and will be moved with `clusterctl move` operation during provisioning. EKS-A CLI makes sure the internal secret is always in sync with the user defined eks-a snow secret.

A couple of follow up PRs are needed:
- identityRef preflight validation: make the secret itself immutable, so that whenever user needs to update the credentials, they change the snowdatacenter identityRef which will trigger the controller reconciliation. 
- controller side of changes

*Testing (if applicable):*

Unit tests


pkg/providers/snow/creds.go
```diff
@@ -6,23 +6,29 @@ import (
...
-type bootstrapCreds struct {
-	snowCredsB64 string
-	snowCertsB64 string
+type BootstrapCreds struct {
+	credsB64 string
+	certsB64 string
+}
+
+// Set is only used for unit test purpose.
+func (b *BootstrapCreds) Set(credsB64, certsB64 string) {
```
vivek-koppuru, line 15: It looks like the struct is exported too, did you want to make the fields exported as well. Or we can just not export the struct and only export this method if you are only using a constructor to get the secret object.

pkg/providers/snow/objects.go
```diff
@@ -156,3 +169,26 @@ func recreateKubeadmConfigTemplateNeeded(new, old *bootstrapv1.KubeadmConfigTemp
...
+	if old != nil {
+		return nil, nil
+	}
+
+	// cli create case: we create the secret from credentials parsed from envs
+	if credentials != nil {
+		return SnowCredentialsSecret(clusterSpec, credentials), nil
+	}
+
+	// controller create case: `{cluster-name}-snow-credentials` secret needs to exist or be manually created first
```
vivek-koppuru, line 192: Is this an error then?

pkg/providers/snow/fetch.go
```diff
@@ -50,3 +51,15 @@ func oldWorkerMachineTemplate(ctx context.Context, kubeclient kubernetes.Client,
 	}
 	return mt, nil
 }
+
+func oldCredentialsSecret(ctx context.Context, kubeClient kubernetes.Client, clusterSpec *cluster.Spec) (*v1.Secret, error) {
+	secret := &v1.Secret{}
+	err := kubeClient.Get(ctx, CredentialsSecretName(clusterSpec), constants.EksaSystemNamespace, secret)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
```
vivek-koppuru, line 59: This is a bit confusing, is it better to handle this not found case in the caller of this function?

pkg/providers/snow/snow.go
```diff
@@ -57,6 +57,7 @@ func NewProvider(kubeUnAuthClient KubeUnAuthClient, configManager *ConfigManager
 		retrier:          retrier,
 		configManager:    configManager,
 		skipIpCheck:      skipIpCheck,
+		bootstrapCreds:   &BootstrapCreds{},
```
vivek-koppuru, line 60: Yea based on the above comment, if we have a constructor here instead, would that help being able to avoid a combination of doing this and having a `Set` method?

pkg/providers/snow/fetch.go
jiayiwang7: the caller function intentionally handles not found error  separately so that this method can treat not found differently than other errors.

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -44,6 +57,22 @@ func (s *SnowDatacenterConfig) ClearPauseAnnotation() {
...
+	if s.Spec.IdentityRef == nil {
+		s.Spec.IdentityRef = &Ref{
+			Kind: SnowIdentityKind,
+			Name: fmt.Sprintf("%s-snow-credentials", s.GetName()),
+		}
+	}
+}
+
+func (s *SnowDatacenterConfig) Validate() error {
+	if s.Spec.IdentityRef != nil && s.Spec.IdentityRef.Kind != SnowIdentityKind {
```
g-gaston, line 70: Ahhh this is interesting
We are allowing nil `IdentityRef` in this validation. Which makes sense for the CLI (assuming we run this before setting defaults, which maybe we shouldn't).

But it doesn't work for the webhook. So we probably need to add a second "layer" of validations that are run after or the defaults.

Ideally, all defaults, including the ones in the api package, the ones in the `cluster` package and the ones in the `snow` provider package would run before we do any validations. However, that will require some refactoring to the CLI flow. That's why I think the second layer of validations might be the best solution here, as long as we document the methods well.

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -44,6 +57,22 @@ func (s *SnowDatacenterConfig) ClearPauseAnnotation() {
 	}
 }
 
+func (s *SnowDatacenterConfig) SetDefaults() {
```
g-gaston, line 60: - [ ] Add these defaults to webhook

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -44,6 +57,22 @@ func (s *SnowDatacenterConfig) ClearPauseAnnotation() {
...
+	if s.Spec.IdentityRef == nil {
+		s.Spec.IdentityRef = &Ref{
+			Kind: SnowIdentityKind,
+			Name: fmt.Sprintf("%s-snow-credentials", s.GetName()),
+		}
+	}
+}
+
+func (s *SnowDatacenterConfig) Validate() error {
```
g-gaston, line 69: - [ ] Add these validations to webhook

pkg/cluster/config.go
```diff
@@ -20,6 +22,7 @@ type Config struct {
 	AWSIAMConfigs            map[string]*anywherev1.AWSIamConfig
 	GitOpsConfig             *anywherev1.GitOpsConfig
 	FluxConfig               *anywherev1.FluxConfig
+	SnowCredentialsSecret    *v1.Secret
```
g-gaston, line 25: - [ ] Add a client config builder for this member, so we read from the cluster when building the `cluster.Config` in the controller

pkg/providers/snow/apibuilder.go
```diff
@@ -135,11 +137,46 @@ func SnowCluster(clusterSpec *cluster.Spec) *snowv1.AWSSnowCluster {
 				Host: clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host,
 				Port: 6443,
 			},
+			IdentityRef: &snowv1.AWSSnowIdentityReference{
+				Name: CredentialsSecretName(clusterSpec),
+				Kind: snowv1.SecretKind,
```
g-gaston, line 142: I don't love this. If in the future we want to support more types, we need to change two places. If we want to hardcode this, then I would vote to remove the `kind` field from our spec and just take a name.

I would expect this to pass-through whatever is set in the eks-a spec, since we already validate that it can only be secrets.

pkg/providers/snow/apibuilder.go
```diff
@@ -135,11 +137,46 @@ func SnowCluster(clusterSpec *cluster.Spec) *snowv1.AWSSnowCluster {
 				Host: clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host,
 				Port: 6443,
 			},
+			IdentityRef: &snowv1.AWSSnowIdentityReference{
+				Name: CredentialsSecretName(clusterSpec),
```
g-gaston, line 141: Instead of this, could we pass the actual secret as an input to this method?

pkg/providers/snow/apibuilder.go
```diff
@@ -135,11 +137,46 @@ func SnowCluster(clusterSpec *cluster.Spec) *snowv1.AWSSnowCluster {
...
+func SnowCredentialsSecret(clusterSpec *cluster.Spec, credsB64, certsB64 string) *v1.Secret {
+	s := CredentialsSecret(CredentialsSecretName(clusterSpec), constants.EksaSystemNamespace, credsB64, certsB64)
+	label := map[string]string{
+		clusterctlv1.ClusterctlMoveLabelName: "true",
+	}
+	s.SetLabels(label)
+	return s
+}
+
+func EksaCredentialsSecret(clusterSpec *cluster.Spec, credsB64, certsB64 string) *v1.Secret {
```
g-gaston, line 176: nit: for consistency, can we just pass credsB64 and certsB64 around as `[]byte`?

pkg/providers/snow/creds.go
```diff
@@ -4,25 +4,21 @@ import (
...
 	}
-	p.bootstrapCreds.snowCredsB64 = creds
 
 	certs, err := aws.EncodeFileFromEnv(eksaSnowCABundlesFileKey)
 	if err != nil {
 		return fmt.Errorf("failed to set up snow certificates: %v", err)
 	}
-	p.bootstrapCreds.snowCertsB64 = certs
+
+	clusterSpec.SnowCredentialsSecret = EksaCredentialsSecret(clusterSpec, creds, certs)
```
g-gaston, line 21: And what step of the CLI flow does this happen?

pkg/providers/snow/apibuilder.go
```diff
@@ -135,11 +137,46 @@ func SnowCluster(clusterSpec *cluster.Spec) *snowv1.AWSSnowCluster {
...
+		},
+		Data: map[string][]byte{
+			v1alpha1.SnowCredentialsKey:  []byte(credsB64),
+			v1alpha1.SnowCertificatesKey: []byte(certsB64),
+		},
+		Type: v1.SecretTypeOpaque,
+	}
+}
+
+func SnowCredentialsSecret(clusterSpec *cluster.Spec, credsB64, certsB64 string) *v1.Secret {
```
g-gaston, line 167: nit: to make the distinction clear, it's not for snow in general, it's for the snow capi provider

```suggestion
func CAPASCredentialsSecret(clusterSpec *cluster.Spec, credsB64, certsB64 string) *v1.Secret {
```

pkg/providers/snow/objects.go
```diff
@@ -156,3 +166,28 @@ func recreateKubeadmConfigTemplateNeeded(new, old *bootstrapv1.KubeadmConfigTemp
...
+	credsB64, ok := clusterSpec.SnowCredentialsSecret.Data["credentials"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve credentials from secret [%s]", clusterSpec.SnowCredentialsSecret.GetName())
+	}
+	certsB64, ok := clusterSpec.SnowCredentialsSecret.Data["ca-bundle"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve ca-bundle from secret [%s]", clusterSpec.SnowCredentialsSecret.GetName())
+	}
+
+	return []kubernetes.Object{clusterSpec.SnowCredentialsSecret, SnowCredentialsSecret(clusterSpec, string(credsB64), string(certsB64))}, nil
```
g-gaston, line 192: why do we return `clusterSpec.SnowCredentialsSecret`?
I would expect that to already exist

pkg/providers/snow/objects.go
```diff
@@ -156,3 +166,28 @@ func recreateKubeadmConfigTemplateNeeded(new, old *bootstrapv1.KubeadmConfigTemp
 	}
 	return !equality.Semantic.DeepDerivative(new.Spec, old.Spec)
 }
+
+// credentialsSecret generates the credentials secret(s) used for provisioning a snow cluster.
+// - eks-a credentials secret: user managed secret referred from snowdatacenterconfig identityRef
+// - snow credentials secret: eks-a creates, updates and deletes in eksa-system namespace. this secret is fully managed by eks-a. User shall treat it as a "read-only" object
+func credentialsSecrets(ctx context.Context, clusterSpec *cluster.Spec) ([]kubernetes.Object, error) {
```
g-gaston, line 173: Do we need the context?

pkg/providers/snow/creds.go
g-gaston: Ahhh I see it now
I would vote to make this line part of `configManager.SetDefaultsAndValidate` (the defaults part)

pkg/providers/snow/objects.go
jiayiwang7: this might or might not exist -- for cli create, we need to create this object and push to the cluster. Also during CLI upgrade, if user changes ENVs, we need to update this eksa obj as well.

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -44,6 +57,22 @@ func (s *SnowDatacenterConfig) ClearPauseAnnotation() {
 	}
 }
 
+func (s *SnowDatacenterConfig) SetDefaults() {
+	if s.Spec.IdentityRef == nil {
```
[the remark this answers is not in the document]
jiayiwang7: chatted offline, we will move the setting identityRef out of `func (s *SnowDatacenterConfig) SetDefaults()` and add it as a new cluster defaultler. We keep the `Validate()` location as it is.

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -1,11 +1,24 @@
...
+	SnowCredentialsKey  = "credentials"
+	SnowCertificatesKey = "ca-bundle"
+)
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // SnowDatacenterConfigSpec defines the desired state of SnowDatacenterConfig
 type SnowDatacenterConfigSpec struct { // Important: Run "make generate" to regenerate code after modifying this file
+
+	// IdentityRef is a reference to a identity to be used when reconciling this cluster
```
g-gaston, line 20: nit:
```suggestion
	// IdentityRef is a reference to an identity for the Snow API to be used when reconciling this cluster
```

pkg/cluster/snow.go
```diff
@@ -145,3 +164,13 @@ func getSnowMachineConfigs(ctx context.Context, client Client, c *Config) error
 
 	return nil
 }
+
+// SetSnowDatacenterConfigDefaults only runs in CLI. snowDatacenterConfig.SetDefaults() will run in both CLI and webhook
+func SetSnowDatacenterConfigDefaults(s *anywherev1.SnowDatacenterConfig) {
```
g-gaston, line 169:
```suggestion
// SetSnowDatacenterIndentityRefDefault sets a default secret as the identity reference
// The secret will need to be created by the CLI flow as it's not provided by the user
// This only runs in CLI. snowDatacenterConfig.SetDefaults() will run in both CLI and webhook
func SetSnowDatacenterIndentityRefDefault(s *anywherev1.SnowDatacenterConfig) {
```

pkg/providers/snow/objects.go
```diff
@@ -31,7 +40,7 @@ func ControlPlaneObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeCli
 	}
 	capiCluster := CAPICluster(clusterSpec, snowCluster, kubeadmControlPlane)
 
-	return []kubernetes.Object{capiCluster, snowCluster, kubeadmControlPlane, new}, nil
+	return []kubernetes.Object{capiCluster, snowCluster, kubeadmControlPlane, new, clusterSpec.SnowCredentialsSecret, capasCredentialsSecret}, nil
```
g-gaston, line 43: interesting, why do we return `clusterSpec.SnowCredentialsSecret` from here?
I would expect the CLI to handle this independently. For example, when applying the cluster spec (since this secret now is part of the snow eks-a cluster specification)

pkg/providers/snow/objects.go
```diff
@@ -156,3 +165,27 @@ func recreateKubeadmConfigTemplateNeeded(new, old *bootstrapv1.KubeadmConfigTemp
 	}
 	return !equality.Semantic.DeepDerivative(new.Spec, old.Spec)
 }
+
+// credentialsSecret generates the credentials secret(s) used for provisioning a snow cluster.
+// - eks-a credentials secret: user managed secret referred from snowdatacenterconfig identityRef
+// - snow credentials secret: eks-a creates, updates and deletes in eksa-system namespace. this secret is fully managed by eks-a. User shall treat it as a "read-only" object
+func capasCredentialsSecret(clusterSpec *cluster.Spec) (*v1.Secret, error) {
+	if clusterSpec.SnowCredentialsSecret == nil {
+		return nil, errors.New("cluster.Config.SnowCredentialsSecret shall not be nil")
```
g-gaston, line 174:
```suggestion
		return nil, errors.New("snowCredentialsSecret in clusterSpec shall not be nil")
```

pkg/providers/snow/reconciler/reconciler_test.go
```diff
@@ -33,28 +33,29 @@ const (
...
-	).Return(remoteClient, nil)
-	tt.cniReconciler.EXPECT().Reconcile(tt.ctx, logger, remoteClient, tt.buildSpec())
-
-	result, err := tt.reconciler().Reconcile(tt.ctx, logger, tt.cluster)
-
-	tt.Expect(err).NotTo(HaveOccurred())
-	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
-	tt.Expect(result).To(Equal(controller.Result{}))
-}
+// TODO: add this back when implementing controller changes for identityRef
```
g-gaston, line 36: nit
just add `t.Skip("Needs controller changes for identityRef)`

pkg/providers/snow/objects.go
g-gaston: Added a different comment talking about this same thing and proposing a different solution. Closing this one.

pkg/api/v1alpha1/snowdatacenterconfig_types.go
```diff
@@ -44,6 +57,23 @@ func (s *SnowDatacenterConfig) ClearPauseAnnotation() {
 	}
 }
 
+func (s *SnowDatacenterConfig) SetDefaults() {}
```
g-gaston, line 60: we can just remove this method

g-gaston (approved):

pkg/api/v1alpha1/snowdatacenterconfig_types.go
g-gaston: no need, method removed

pkg/api/v1alpha1/snowdatacenterconfig_types.go
g-gaston: We will do this sepparetly

PR #3338, opened 2022-09-13
Lower the secret name for backward compatibility

wongni:
*Issue #, if available:*

*Description of changes:* 
Historically "Global" has been used for the name of the INI section containing URL and secrets to the target CloudStack API endpoint. However, the k8s secret object only supports the lowercase name. To support the customer wanting to use the 0.8.3 release and the 0.11.x and higher releases, we need this PR.

*Testing (if applicable):*


wongni:
The decoder.go already lowers the section name

		cloudstackProfiles = append(cloudstackProfiles, CloudStackProfileConfig{
			Name:          strings.ToLower(section.Name()),

PR #3342, opened 2022-09-13
Fetch only repo tags that match semver

abhay-krishna:
This PR fixes the logic of fetching repo tags to list only those that match semver format.


abhinavmpandey08 (approved):
/approve

PR #3351, opened 2022-09-14
Improve logging in eks-a e2e test runner

g-gaston:
*Description of changes:*

* Don't print output from e2e tests from instance (this is already included in their own logstream)
* Pass a logger to all downstream entities
* Use logger with values in parallel runners
* Avoid noisy debug logs when running SSM commands by using a discard logger
* Unify instance test result log

*Testing (if applicable):*
In only built the binary, so this could break the tests

### DIFF
--- a/internal/pkg/ssm/command.go
+++ b/internal/pkg/ssm/command.go
@@ -8,8 +8,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/go-logr/logr"
 
-	"github.com/aws/eks-anywhere/pkg/logger"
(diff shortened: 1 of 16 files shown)


internal/pkg/ssm/command.go
```diff
@@ -19,7 +19,7 @@ var initE2EDirCommand = "mkdir -p /home/e2e/bin && cd /home/e2e"
 
 func WaitForSSMReady(session *session.Session, instanceId string) error {
 	err := retrier.Retry(10, 20*time.Second, func() error {
-		return Run(session, instanceId, "ls")
+		return Run(session, logr.Discard(), instanceId, "ls")
```
danbudris, line 22: would it be useful to be able to toggle the discard on/off? using an env var or flag perhaps

danbudris (approved):

danbudris (approved):
/lgtm

Issue #3238, opened 2022-09-01
Preflight to check port 80 availability on the host for bare-metal


abhinavmpandey08:
**What would you like to be added**:
Preflight validation to check if port 80 is available on the host for bare-metal cluster creation as it is needed by boots

**Why is this needed**:
`boots` container needs port 80 to be available to serve http. If this port is already in use, `boots` panics with this error 
`panic: listen and serve http: listen tcp 10.80.8.22:80: bind: address already in use`

Adding a preflight check to verify that the port is available or not should help us circumvent this issue.

This should also resolve https://github.com/aws/eks-anywhere/issues/3229

PR #3357, opened 2022-09-15
Add preflight to check port availability on the host for baremetal

ahreehong:
*Issue #, if available:*
#3238 

*Description of changes:*
Added a preflight to ensure ports 80, 42113, and 50061 are not in use for baremetal

*Testing (if applicable):*
Tried creating a cluster and verified:
`❌ Validation failed	{"validation": "tinkerbell Provider setup is valid", "error": "port 42113 of host is already in use, please clean up ports", "remediation": ""}`

### DIFF
--- a/pkg/networkutils/networkutils_test.go
+++ b/pkg/networkutils/networkutils_test.go
@@ -63,6 +63,33 @@ func TestIsIPInUseFail(t *testing.T) {
 	g.Expect(res).To(gomega.BeTrue())
 }
 
+func TestIsPortInUsePass(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	g := gomega.NewWithT(t)
+
+	client := mocks.NewMockNetClient(ctrl)
+	client.EXPECT().DialTimeout("tcp", "10.10.10.10:80", 500*time.Millisecond).
+		Return(nil, errors.New("no connection"))
+
+	res := networkutils.IsPortInUse(client, "10.10.10.10", "80")
+	g.Expect(res).To(gomega.BeFalse())
+}
+
+func TestIsPortInUseFail(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	g := gomega.NewWithT(t)
+
+	conn := NewMockConn(ctrl)
+	conn.EXPECT().Close().Return(nil)
+
+	client := mocks.NewMockNetClient(ctrl)
+	client.EXPECT().DialTimeout("tcp", "10.10.10.10:80", 500*time.Millisecond).
+		Return(conn, nil)
+
+	res := networkutils.IsPortInUse(client, "10.10.10.10", "80")
+	g.Expect(res).To(gomega.BeTrue())
+}
+
 func TestGetLocalIP(t *testing.T) {
 	_, err := networkutils.GetLocalIP()
 	if err != nil {
--- a/pkg/providers/tinkerbell/assert.go
+++ b/pkg/providers/tinkerbell/assert.go
@@ -119,6 +119,17 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
 	return nil
 }
 
+func AssertPort80IsNotInUse(client networkutils.NetClient) ClusterSpecAssertion {
+	return func(spec *ClusterSpec) error {
+		port := "80"
+		host := "0.0.0.0"
+		if networkutils.IsPortInUse(client, host, port) {
+			return fmt.Errorf("port 80 of host is already in use")
+		}
+		return nil
+	}
+}
+
 // HardwareSatisfiesOnlyOneSelectorAssertion ensures hardware in catalogue only satisfies 1
 // of the MachineConfig's HardwareSelector's from the spec.
 func HardwareSatisfiesOnlyOneSelectorAssertion(catalogue *hardware.Catalogue) ClusterSpecAssertion {
--- a/pkg/providers/tinkerbell/assert.go
+++ b/pkg/providers/tinkerbell/assert.go
@@ -119,12 +119,12 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
 	return nil
 }
 
-func AssertPort80IsNotInUse(client networkutils.NetClient) ClusterSpecAssertion {
+// AssertPortsNotInUse ensures that ports 80, 42113, and 50061 are available
+func AssertPortsNotInUse(client networkutils.NetClient) ClusterSpecAssertion {
 	return func(spec *ClusterSpec) error {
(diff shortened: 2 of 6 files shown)


pkg/networkutils/networkutils_test.go
```diff
@@ -63,6 +63,35 @@ func TestIsIPInUseFail(t *testing.T) {
 	g.Expect(res).To(gomega.BeTrue())
 }
 
+func TestIsPortInUsePass(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	g := gomega.NewWithT(t)
+
+	client := mocks.NewMockNetClient(ctrl)
+	client.EXPECT().DialTimeout(gomock.Any(), gomock.Any(), gomock.Any()).
```
abhinavmpandey08, line 71: We know what the expected values are supposed to be, so let's use them instead of `gomock.Any()`

Here, it should expect `"tcp", address, 500*time.Millisecond`

pkg/networkutils/networkutils_test.go
```diff
@@ -63,6 +63,35 @@ func TestIsIPInUseFail(t *testing.T) {
...
+		AnyTimes().
```
abhinavmpandey08, line 72: This shouldn't be needed as `DialTimeout` is going to be called exactly once

pkg/networkutils/networkutils_test.go
```diff
@@ -63,6 +63,35 @@ func TestIsIPInUseFail(t *testing.T) {
...
+
+func TestIsPortInUseFail(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	g := gomega.NewWithT(t)
+
+	conn := NewMockConn(ctrl)
+	conn.EXPECT().Close().Return(nil)
+
+	client := mocks.NewMockNetClient(ctrl)
+	client.EXPECT().DialTimeout(gomock.Any(), gomock.Any(), gomock.Any()).
```
abhinavmpandey08, line 87: same as above, let's not use `gomock.Any()` unless really necessary

pkg/providers/tinkerbell/assert.go
```diff
@@ -119,6 +119,20 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
...
+// AssertPortsNotInUse ensures that ports 80, 42113, and 50061
+func AssertPortsNotInUse(client networkutils.NetClient) ClusterSpecAssertion {
+	return func(spec *ClusterSpec) error {
+		host := "0.0.0.0"
+		ports := []string{"80", "42113", "50061"}
+		for _, port := range ports {
+			if networkutils.IsPortInUse(client, host, port) {
+				return fmt.Errorf("port %s of host is already in use, please clean up ports", port)
+			}
+		}
```
abhinavmpandey08, line 131: Instead of erroring out on the first unavailable port, can we go through the whole list and check all the ports first and then display the error with all unavailable ports?

abhinavmpandey08 (commented):
looks good! some minor comments

pkg/providers/tinkerbell/assert_test.go
```diff
@@ -229,6 +229,41 @@ func TestAssertTinkerbellIPAndControlPlaneIPNotSame_SameFails(t *testing.T) {
 	g.Expect(tinkerbell.AssertTinkerbellIPAndControlPlaneIPNotSame(clusterSpec)).ToNot(gomega.Succeed())
 }
 
+func TestAssertPortsNotInUse_Succeeds(t *testing.T) {
+	g := gomega.NewWithT(t)
+	ctrl := gomock.NewController(t)
+
+	netClient := mocks.NewMockNetClient(ctrl)
+	netClient.EXPECT().
+		DialTimeout(gomock.Any(), gomock.Any(), gomock.Any()).
```
abhinavmpandey08, line 238: Only use `gomock.Any()` if you can't determine the value the mock client is expecting

pkg/providers/tinkerbell/assert_test.go
```diff
@@ -229,6 +229,41 @@ func TestAssertTinkerbellIPAndControlPlaneIPNotSame_SameFails(t *testing.T) {
...
+		AnyTimes().
```
abhinavmpandey08, line 239: this isn't needed as `DialTimeout` will only be called once

pkg/providers/tinkerbell/assert_test.go
```diff
@@ -229,6 +229,41 @@ func TestAssertTinkerbellIPAndControlPlaneIPNotSame_SameFails(t *testing.T) {
...
+	g := gomega.NewWithT(t)
+	ctrl := gomock.NewController(t)
+
+	server, client := net.Pipe()
+	defer server.Close()
+
+	netClient := mocks.NewMockNetClient(ctrl)
+	netClient.EXPECT().
+		DialTimeout(gomock.Any(), gomock.Any(), gomock.Any()).
+		AnyTimes().
```
abhinavmpandey08, line 258: same as above

pkg/providers/tinkerbell/validate.go
```diff
@@ -125,6 +125,26 @@ func validateIPUnused(client networkutils.NetClient, ip string) error {
 	return nil
 }
 
+func validatePortsAvailable(client networkutils.NetClient, host string) error {
+	unavailablePorts := getPortsUnavailable(client, host)
+
+	if len(unavailablePorts) != 0 {
+		return fmt.Errorf("ports %v of host is already in use, please clean up ports", strings.Join(unavailablePorts, ", "))
```
abhinavmpandey08, line 132:
```suggestion
		return fmt.Errorf("localhost ports [%v] are already in use, please ensure these ports are available", strings.Join(unavailablePorts, ", "))
```

pkg/providers/tinkerbell/assert_test.go
```diff
@@ -229,6 +229,41 @@ func TestAssertTinkerbellIPAndControlPlaneIPNotSame_SameFails(t *testing.T) {
...
 
+func TestAssertPortsNotInUse_Succeeds(t *testing.T) {
+	g := gomega.NewWithT(t)
+	ctrl := gomock.NewController(t)
+
+	netClient := mocks.NewMockNetClient(ctrl)
+	netClient.EXPECT().
+		DialTimeout(gomock.Any(), gomock.Any(), gomock.Any()).
+		AnyTimes().
+		Return(nil, errors.New("failed to connect"))
```
abhinavmpandey08, line 240: actually, this will be called 3 times exactly so you can do something like this instead
```suggestion
		Return(nil, errors.New("failed to connect")).Times(3)
```

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #3321, opened 2022-09-09
Allow workerNodeGroupConfigs to be empty

d8660091:
*Issue #, if available:*

*Description of changes:* 
For creating single-node cluster
1. Allow workerNodeGroupConfigs to be empty
2. Check if control plane taints is empty when there is no worker node
3. Warn when there is no worker node

*Testing (if applicable):*
1. unit-test
2. Invoke eksa create cluster without workerNodeGroups. Validation pass
```
2022-09-09T11:26:48.464-0400    V0      Private key saved to single-node/eks-a-id_rsa. Use 'ssh -i single-node/eks-a-id_rsa <username>@<Node-IP-Address>' to login to your cluster node
2022-09-09T11:26:50.968-0400    V0      ✅ Tinkerbell Provider setup is valid
2022-09-09T11:26:50.968-0400    V0      ✅ Validate certificate for registry mirror
2022-09-09T11:26:50.968-0400    V0      ✅ Validate authentication for git provider
2022-09-09T11:26:50.968-0400    V0      ✅ Create preflight validations pass
```
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3358, opened 2022-09-15
Pass empty taints to cluster-api when there is no worker node

d8660091:
*Issue #, if available:*

## Description of changes
Pass taints = [] to cluster-api when workerNodeGroupConfigurations is empty

Related PR (allow workerNodeGroupConfigurations = nil): https://github.com/aws/eks-anywhere/pull/3321/files

## Testing (if applicable) 
- Unit test
- Manually created a tinkerbel single-node cluster with this change

### DIFF
--- a/pkg/providers/tinkerbell/config/template-cp.yaml
+++ b/pkg/providers/tinkerbell/config/template-cp.yaml
@@ -92,7 +92,10 @@ spec:
 {{- if .kubeletExtraArgs }}
 {{ .kubeletExtraArgs.ToYaml | indent 10 }}
 {{- end }}
-{{- if .controlPlaneTaints }}
+{{- if not .workerNodeGroupConfigurations }}
+        taints: []
+{{- end }}
+{{- if and .workerNodeGroupConfigurations .controlPlaneTaints }}
         taints:
 {{- range .controlPlaneTaints}}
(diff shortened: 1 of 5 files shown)


pkg/providers/tinkerbell/config/template-cp.yaml
```diff
@@ -130,7 +133,10 @@ spec:
 {{- if .kubeletExtraArgs }}
 {{ .kubeletExtraArgs.ToYaml | indent 10 }}
 {{- end }}
-{{- if .controlPlaneTaints }}
+{{- if not .workerNodeGroupConfigurations }}
+        taints: []
+{{- end }}
+{{- if and .workerNodeGroupConfigurations .controlPlaneTaints }}
```
chrisdoherty4, line 139: Do we explicitly prevent users from specifying control plane taints but no worker node groups? As is, the taints will be ignored. Have we made a not to document this somewhere?

d8660091: Right, I have made another CR to prevent users from specifying control plane taints: https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster.go#L442-L444

chrisdoherty4 (approved):

PR #3362, opened 2022-09-15
Remove CoreDNS role patch

g-gaston:
*Description of changes:*

This was necessary with older versions of eks-d because of the CoreDNS version that was bundled with kubernetes 1.20 and 1.21, and those kubernetes versions missing some permissions for the CoreDNS role. This has been patched in new versions of eks-d by default, so we don't need it anymore.

*Testing (if applicable):*

Created a docker cluster manually. CoreDNS is up and running


vivek-koppuru:
I see that EKS-D 1.20 still uses `v1.8.3` of coredns, so are we sure we don't need this anymore?

chrisdoherty4:
@vivek-koppuru 

> I see that EKS-D 1.20 still uses v1.8.3 of coredns, so are we sure we don't need this anymore?

I queried with @g-gaston and he said we've released a version of EKS-D 1.20 with the patch. I suppose we could link it for clarity in this PR?

vivek-koppuru:
> @vivek-koppuru
> 
> > I see that EKS-D 1.20 still uses v1.8.3 of coredns, so are we sure we don't need this anymore?
> 
> I queried with @g-gaston and he said we've released a version of EKS-D 1.20 with the patch. I suppose we could link it for clarity in this PR?

Got it, I was expecting the patch to be located here but I didn't see anything so I was wondering where else it might be: https://github.com/aws/eks-distro/tree/main/projects/coredns/coredns/1-20/patches

g-gaston:
@vivek-koppuru 
We don't patch coredns to use need less permissions but kubeadm to assign it more:
https://github.com/aws/eks-distro/blob/main/projects/kubernetes/kubernetes/1-20/patches/0012-PATCH-kubeadm-CoreDNS-permissions-for-endpointslices.patch

That said, I did test this manually before creating the PR, just running a `create cluster` with docker.

PR #3369, opened 2022-09-16
Apply eksd manifest for each kubernetes version to cluster

taneyland:
*Issue #, if available:*

*Description of changes:*
When upgrading a new cluster using the full lifecycle api, we can't upgrade the kubernetes version if it differs from the mgmt cluster version because only the eksd manifest for that version is applied to the cluster. To avoid needing a network connection in the controller, we will apply eks-d versions for all kubernetes versions in the bundle to the mgmt cluster.

*Testing (if applicable):*
Create docker cluster
*Testing with new api now...

### DIFF
--- a/pkg/eksd/installer.go
+++ b/pkg/eksd/installer.go
@@ -41,22 +41,28 @@ func NewEksdInstaller(client EksdInstallerClient, reader Reader) *Installer {
 
 func (i *Installer) InstallEksdCRDs(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error {
 	var eksdCRDs []byte
-	if err := i.retrier.Retry(
-		func() error {
-			var readerErr error
-			eksdCRDs, readerErr = i.reader.ReadFile(clusterSpec.VersionsBundle.EksD.Components)
-			return readerErr
-		},
-	); err != nil {
-		return fmt.Errorf("loading manifest for eksd components: %v", err)
+	eksdCrds := map[string]struct{}{}
+	for _, vb := range clusterSpec.Bundles.Spec.VersionsBundles {
+		eksdCrds[vb.EksD.Components] = struct{}{}
 	}
+	for crd := range eksdCrds {
+		if err := i.retrier.Retry(
+			func() error {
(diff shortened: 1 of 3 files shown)


pkg/eksd/installer.go
```diff
@@ -41,22 +41,24 @@ func NewEksdInstaller(client EksdInstallerClient, reader Reader) *Installer {
...
-			return readerErr
-		},
-	); err != nil {
-		return fmt.Errorf("loading manifest for eksd components: %v", err)
-	}
+	for _, vb := range clusterSpec.Bundles.Spec.VersionsBundles {
+		if err := i.retrier.Retry(
+			func() error {
+				var readerErr error
+				eksdCRDs, readerErr = i.reader.ReadFile(vb.EksD.Components)
```
g-gaston, line 48: `EksD.Components` are probably the same for all version bundles. I would suggest creating a set with the urls and avoid duplicating calls

pkg/eksd/installer_test.go
```diff
@@ -91,10 +94,23 @@ func TestInstallEksdManifestSuccess(t *testing.T) {
...
+	tt.reader.EXPECT().ReadFile(tt.clusterSpec.Bundles.Spec.VersionsBundles[0].EksD.EksDReleaseUrl).Return([]byte(""), fmt.Errorf("error"))
 	if err := tt.eksdInstaller.InstallEksdManifest(tt.ctx, tt.clusterSpec, tt.cluster); err == nil {
 		t.Error("Eksd.InstallEksdManifest() error = nil, wantErr not nil")
 	}
 }
+
+func bundle() *v1alpha1.Bundles {
+	return &v1alpha1.Bundles{
+		Spec: v1alpha1.BundlesSpec{
+			VersionsBundles: []v1alpha1.VersionsBundle{
```
g-gaston, line 109: can we add at least two version bundles? just we test the iteration, etc.

pkg/eksd/installer.go
taneyland: Very good point. Updated

g-gaston (approved):

Issue #3381, opened 2022-09-17
Unable to create a cluster in vSphere if the dc name consists of only numbers


adonskoy:
**What happened**:
If the vSphere datacenter name specified in VSphereDatacenterConfig contains only numbers, then during generation of the VSphereMachineTemplate an integer value is substituted in spec.template.spec.datacenter (even if the value is a string in VSphereDatacenterConfig). This results in an error while applying the new resources of this kind. 
```
2022-09-17T12:57:07.798+0300    V6      Executing command       {"cmd": "/usr/local/bin/docker exec -i eksa_1663408305631319000 kubectl apply -f - --namespace eksa-system --kubeconfig eks-system/generated/eks-system.kind.kubeconfig"}
2022-09-17T12:57:08.194+0300    V5      Error happened during retry     {"error": "executing apply: Error from server (Invalid): error when creating \"STDIN\": VSphereMachineTemplate.infrastructure.cluster.x-k8s.io \"eks-system-control-plane-template-1663408479593\" is invalid: spec.template.spec.datacenter: Invalid value: \"integer\": spec.template.spec.datacenter in body must be of type string: \"integer\"\nError from server (Invalid): error when creating \"STDIN\": VSphereMachineTemplate.infrastructure.cluster.x-k8s.io \"eks-system-etcd-template-1663408479593\" is invalid: spec.template.spec.datacenter: Invalid value: \"integer\": spec.template.spec.datacenter in body must be of type string: \"integer\"\nError from server (Invalid): error when creating \"STDIN\": VSphereMachineTemplate.infrastructure.cluster.x-k8s.io \"eks-system-md-0-1663408479597\" is invalid: spec.template.spec.datacenter: Invalid value: \"integer\": spec.template.spec.datacenter in body must be of type string: \"integer\"\n", "retries": 28}
```

**What you expected to happen**:
When generating a resource manifest of kind VSphereMachineTemplate, always convert the spec.template.spec.datacenter value to a string.

**How to reproduce it (as minimally and precisely as possible)**:
Name the datacenter in vSphere like 1234 and try to create a new initial-cluster. 

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: v0.11.2
- EKS Distro Release:

vivek-koppuru:
So are you mentioning that it is possible to name a datacenter with all integers?

vivek-koppuru:
Closing in favor of #3382 getting merged

PR #3382, opened 2022-09-17
Adding quotes for datacenter field

adonskoy:
*Issue #, if available:* [#3381](https://github.com/aws/eks-anywhere/issues/3381)

*Description of changes:*
Added quotes to the `spec.template.spec.datacenter` value in VSphereMachineTemplate to ensure that the value type is always a string.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/vsphere/config/template-cp.yaml
+++ b/pkg/providers/vsphere/config/template-cp.yaml
@@ -50,7 +50,7 @@ spec:
   template:
     spec:
       cloneMode: linkedClone
-      datacenter: {{.vsphereDatacenter}}
+      datacenter: '{{.vsphereDatacenter}}'
       datastore: {{.controlPlaneVsphereDatastore}}
       diskGiB: {{.controlPlaneDiskGiB}}
       folder: '{{.controlPlaneVsphereFolder}}'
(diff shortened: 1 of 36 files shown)


TerryHowe (commented):
/ok-to-test

TerryHowe:
uugn, I could see one or two test failures, but way too many dependencies on that.

adonskoy:
> This is a great find, thank you for sharing this! Now my question is, should we do this for the other string values in this file as well if they can support only integers as well?

As far as I can see, other values cannot be integers in this file, because they will either initially be a string, or are the result of concatenating other strings (for example, datastore)

vivek-koppuru:
> > This is a great find, thank you for sharing this! Now my question is, should we do this for the other string values in this file as well if they can support only integers as well?
> 
> As far as I can see, other values cannot be integers in this file, because they will either initially be a string, or are the result of concatenating other strings (for example, datastore)

Good point, I think these changes are good then!

vivek-koppuru (approved):
/approve

vivek-koppuru:
@adonskoy Can you rebase the changes in your branch and add this change to the `expected_results_minimal_autoscaling_md.yaml` file in the vsphere package as well? Seems to be a new test file

vivek-koppuru (approved):

vivek-koppuru:
@adonskoy Thank you so much for your contribution!

PR #2938, opened 2022-08-15
Moving CloudStack to use v2 controller

maxdrib:
*Issue #, if available:*
* https://github.com/aws/eks-anywhere/issues/2929
* https://github.com/aws/eks-anywhere/issues/2260

*Description of changes:*
In order to address the issue, it seemed like migrating the Cloudstack provider off the legacy controller was the best way to enable the proper creation of the ClusterSpec. I have done so, based off the VSphere provider's example and would like some early feedback on my approach, and to make sure I haven't missed anything obvious.

*Testing (if applicable):*
TBD

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #2260, opened 2022-05-25
Migrate Cloudstack provider to use new cluster controller


maxdrib:
**What would you like to be added**:
I would like to move the eks-a controller logic for cloudstack away from the legacy controller and to use the new broken-out cluster controller instead

**Why is this needed**:
This code is on a deprecation path and I would like to use the newer code path to make sure it is properly tested and maintained

PR #3386, opened 2022-09-19
Adding api builder for cloudstack machine templates and machine deployments

maxdrib:
Issue #, if available:

Description of changes:
First PR of many covering changes required to support full lifecycle API controller. This moves the cloudstackmachinetemplate generation for control planes to use the apibuilder instead of templating.

Testing (if applicable):
* unit tests
* successfully ran 1.20->1.21 E2E test locally
* Successfully scaled cluster with changes from https://github.com/aws/eks-anywhere/pull/2938


pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	anywhereTypes "github.com/aws/eks-anywhere/pkg/types"
+)
+
+const (
+	CloudStackMachineTemplateKind      = "CloudStackMachineTemplate"
+	CloudStackInfrastructureAPIVersion = "infrastructure.cluster.x-k8s.io/v1beta2"
```
vivek-koppuru, line 22: Can we use the `InfrastructureAPIVersion` here from `cloudstackv1` or is there no method that gets this specific version?

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,119 @@
...
+
+func machineDeployment(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1.WorkerNodeGroupConfiguration, kubeadmConfigTemplate *bootstrapv1.KubeadmConfigTemplate, cloudstackMachineTemplate *cloudstackv1.CloudStackMachineTemplate) clusterv1.MachineDeployment {
+	return clusterapi.MachineDeployment(clusterSpec, workerNodeGroupConfig, kubeadmConfigTemplate, cloudstackMachineTemplate)
+}
+
+func MachineDeployments(clusterSpec *cluster.Spec, kubeadmConfigTemplates map[string]*bootstrapv1.KubeadmConfigTemplate, machineTemplates map[string]*cloudstackv1.CloudStackMachineTemplate) map[string]*clusterv1.MachineDeployment {
+	m := make(map[string]*clusterv1.MachineDeployment, len(clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations))
+
+	for _, workerNodeGroupConfig := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		deployment := machineDeployment(clusterSpec, workerNodeGroupConfig,
```
vivek-koppuru, line 33: This logic can be unit tested right?

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,119 @@
...
+		annotations[fmt.Sprintf("label.diskoffering.%s", constants.CloudstackAnnotationSuffix)] = machineConfig.DiskOffering.Label
+	}
+
+	if machineConfig.Symlinks != nil {
+		links := make([]string, 0)
+		for key := range machineConfig.Symlinks {
+			links = append(links, fmt.Sprintf("%s:%s", key, machineConfig.Symlinks[key]))
+		}
+		// sorting for unit test determinism
+		sort.Strings(links)
```
vivek-koppuru, line 57: Instead of needing to sort in the business logic, can we just change our verification logic in our unit test to use a hash set?

pkg/providers/cloudstack/apibuilder.go
maxdrib: Indeed we can, thanks

pkg/providers/cloudstack/apibuilder.go
maxdrib: The unit tests in provider/cloudstack.go are strictly validating against yaml "expected" testdata at the moment which don't have this capability :(

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -657,10 +659,34 @@ func (cs *CloudStackTemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *c
 		buildOption(values)
 	}
 
+	cpCsMtTemplateName, ok := values[cpTemplateNameKey]
```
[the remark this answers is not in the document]
maxdrib: You are correct, thanks for the note

pkg/providers/cloudstack/apibuilder.go
maxdrib: Required some additional setup but it's possible!

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		CustomSize: machineConfig.DiskOffering.CustomSize,
+		MountPath:  machineConfig.DiskOffering.MountPath,
+		Device:     machineConfig.DiskOffering.Device,
+		Filesystem: machineConfig.DiskOffering.Filesystem,
+		Label:      machineConfig.DiskOffering.Label,
+	}
+}
+
+func CloudStackMachineTemplate(name string, machineConfig *v1alpha1.CloudStackMachineConfigSpec, now anywhereTypes.NowFunc) *cloudstackv1.CloudStackMachineTemplate {
```
g-gaston, line 81: Maybe we should have the name being generated in this method?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -657,10 +659,34 @@ func (cs *CloudStackTemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *c
 		buildOption(values)
 	}
 
+	cpCsMtName, ok := values[cpTemplateNameKey]
+	if !ok {
+		return nil, fmt.Errorf("unable to determine control plane template name")
+	}
+	cpMachineTemplate := CloudStackMachineTemplate(fmt.Sprintf("%v", cpCsMtName), cs.controlPlaneMachineSpec, cs.now)
```
g-gaston, line 666: I would like the machine template name to come from a method, so it's easy to replicate

g-gaston, line 666: Just so I can see how this will fit the big picture, how are you planning on using this from the controller?

Are the code paths gonna be the same for cli and controller (minus the marshaling) or are they going to be composed differently

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,118 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   constants.EksaSystemNamespace,
+			Annotations: fillMachineTemplateAnnotations(machineConfig),
+		},
+		Spec: cloudstackv1.CloudStackMachineTemplateSpec{
+			Spec: cloudstackv1.CloudStackMachineTemplateResource{
+				ObjectMeta: metav1.ObjectMeta{
+					CreationTimestamp: metav1.Time{
+						Time: now(),
```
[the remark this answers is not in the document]
maxdrib: I tried not including this field but CAPC appeared to reject it with some validation webhook error saying `CreationTimestamp cannot be null` or something along those lines. I don't believe it's actually used anywhere in CAPC, but it has to be set.

pkg/providers/cloudstack/apibuilder.go
maxdrib: Name generation is complex, and depends on the usecase. It varies between the CLI and controller, and between old controller and new controller. I think it would be cleaner to leave that logic upstream and keep this method simple.

pkg/providers/cloudstack/cloudstack.go
maxdrib: all the immutable objects are going to be generated from the apibuilder (machinetemplates and machinedeployments) in both the CLI and the controller. All the mutable objects will be generated from the templates (kubeadmconfig, cloudstackcluster, capi cluster) in both the CLI and controller. In the CLI, the objects will be applied as yaml. In the controller, they will be compared and then applied as K8s objects.

maxdrib: currently, machine template name is only available from the `values` map. All this code is doing is reading from that map and then converting to a string. Are you suggesting factoring out that logic? Or are you proposing we centralize all the ways that the name is generated?

pkg/providers/cloudstack/apibuilder.go
vivek-koppuru: Can we create an action item to address this? I would really like to avoid doing something for a unit test in our business logic. Not sure if we do this today already, but ideally the symlinks passed into this method would already be sorted for that case. Are we able to control that?

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
vivek-koppuru: If it can't be null, then can we set it to an empty struct and have the system fill it in still? Or is that not possible?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -657,10 +659,34 @@ func (cs *CloudStackTemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *c
 		buildOption(values)
 	}
 
+	cpCsMtName, ok := values[cpTemplateNameKey]
```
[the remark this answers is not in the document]
vivek-koppuru: I think `cpTemplateName` or `cpMachineTemplateName` is fine to say here, no need to say `Cs`, unless we have a reference to the regular machine template name here that is different than this.

pkg/providers/cloudstack/apibuilder.go
maxdrib: The symlinks are passed in as a standard unsorted hashmap and they get sorted arbitrarily when you iterate over them. The other option would be only testing for a single symlink but then we do not get validation on the multi-symlink formatting, which feels like missed coverage and risky. I agree about adding logic for unit tests to business logic is messy, but it seems harmless in any real scenario in my opinion so I'm inclined to keep it for now.

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
g-gaston: @maxdrib do you think we can track where that error comes from? I'm ok moving forward with this if it's a blocker but it sounds weird that webhook will reject this. Looking upstream, it says its set by the system:
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L179-L188

pkg/providers/cloudstack/apibuilder.go
g-gaston: I would expect the CLI and the new controller to use the same name generation logic. How do those two differ right now?

pkg/providers/cloudstack/cloudstack.go
g-gaston: I see. But we are in `GenerateCAPISpecControlPlane` method, which returns `[]byte`
Then does this mean we are going to need to duplicate the logic that calls "go template + api structs" in the controller? Just to be able to access the api structs before they get marshaled to yaml

g-gaston: I see. But we are in `GenerateCAPISpecControlPlane` method, which returns `[]byte`
Then does this mean we are going to need to duplicate the logic that calls "go template + api structs" in the controller? Just to be able to access the api structs before they get marshaled to yaml

g-gaston: Ah I see
Is the machine template name being used in the go template? If not, no need to pass it as part of the map, we can make it a explicit argument to this/other methods

As for conversion, does `fmt.Sprintf("%s", cpCsMtName)` work?

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
chrisdoherty4: Is it getting rejected by a CAPC webhook?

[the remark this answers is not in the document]
maxdrib: Correct @chrisdoherty4, it is being rejected by a CAPC webhook when it's not set. I don't remember the exact logs. I could set this to an empty Time object but that'll be misleading since it'll be epoch time I imagine.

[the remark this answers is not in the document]
maxdrib:
```
2022-09-26T14:59:10.137-0400	V6	Executing command	{"cmd": "/usr/local/bin/docker exec -i eksa_1664218435037679000 kubectl apply -f - --namespace eksa-system --kubeconfig eksa-drib-b77a195/generated/eksa-drib-b77a195.kind.kubeconfig"}
2022-09-26T14:59:10.832-0400	V9	docker	{"stderr": "error: error validating \"STDIN\": error validating data: unknown object type \"nil\" in CloudStackMachineTemplate.spec.template.metadata.creationTimestamp; if you choose to ignore these errors, turn validation off with --validate=false\n"}
```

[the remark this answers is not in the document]
g-gaston: that doesn't seem to come from the webhook necessarily, it seems more like the api server
Is this a new machine template or are you updating an existing one?
If it's an update and you set the date during creation, that date would be part of the last-applied annotation, hence kubectl would try to remove the date that was already set.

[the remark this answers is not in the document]
maxdrib: It's a new one - this error appears on cluster creation when I exclude the ObjectMeta entirely. I think previously I had seen some actual webhook error from CAPC but I can't find the log now, it's been a few days.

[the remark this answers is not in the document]
maxdrib: This seems to be a serialization issue. For some reason, no matter what I try, it's being serialized as 
```
    metadata:
      creationTimestamp: null
```
which the api server does not like for some reason. This is also how the creationTimestamp appears in the CSMT's top level ObjectMeta (`creationTimestamp: null`) even when it's not explicitly specified. 

When we were using template-cp, the resulting spec did not have ObjectMeta specified here, and CAPC was happy with that. With that being said, I think we have three options to proceed here

1. Remove this ObjectMeta attribute from the CloudStackMachineTemplate.Spec.Spec in CAPC. This will require an API bump in CAPC and does not feel worth it at this time
2. Modify the serialization logic in the yaml package to remove this field, which feels hacky
3. Set the field to a semi-realistic value, acknowledging that it's unused and proceed with the current implementation.

With those options laid out, I'm inclined to proceed with option 3, and also creating a ticket under the CAPC repo to remove this unused ObjectMeta resource.

However, one thing I still don't understand is why the K8s server accepts the top-level creationTimestamp as null but not this lower-level one. The types are defined the same way in https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/main/api/v1beta2/cloudstackmachinetemplate_types.go

pkg/providers/cloudstack/apibuilder.go
maxdrib: When we generate the template name for cluster creation, we use `common.CPMachineTemplateName`. 

When we generate the template name for cluster Upgrade, we call `provider.getControlPlaneNameForCAPISpecUpgrade` which  checks if a new control plane template is required. If a new one is required by validating the previous EKS-A cluster spec, previous bundles, or AnyImmutableFieldChanged. If a new template is needed, it uses the same function as above, otherwise we use the existing one from the cp CloudStackMachineTemplate. 

The controller currently checks if any immutable field was changed on the CloudStackDatacenterConfig or CloudStackMachineConfigs and generates a new name if things have changed - otherwise it uses the one from the KCP in the cluster, at `cp.Spec.MachineTemplate.InfrastructureRef.Name`.

These are all sightly different ways to generate the name which make me nervous to just generate the name inside this function. It would also require feeding in a lot of context about the rest of the system to be able to check if a new one is needed or not.

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
chrisdoherty4: `kubectl` always outputs with a null creation timestamp. I've never checked if it was intentional or not but I'd imagine there's something akin to a missing `"omitempty"` somewhere.

```
~/Workspace/github.com/aws/eks-anywhere > kubectl create --dry-run=client -o yaml deployment --image=nginx nginx
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: nginx
  name: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
        resources: {}
status: {}
```

pkg/providers/cloudstack/apibuilder.go
chrisdoherty4, line 81: Can we drop the `now` parameter? I understand the testability aspect but it puts a burden on real consumers to need to care about a time function.

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,118 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	cloudstackv1 "sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	anywhereTypes "github.com/aws/eks-anywhere/pkg/types"
```
chrisdoherty4, line 17: I think this is only used for the time func: in any case, import alias' are subject to the same rules of package names, they should be all lower case (for future reference).

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
g-gaston: @maxdrib could we add a default for this field in the mutation webhook?

pkg/providers/cloudstack/apibuilder.go
g-gaston: I understand that concern. But even if we don't generate the name in this method, is possible to have a shared codepath generate the name for both controller and CLI?

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
maxdrib: I tried adding a defaulter to the CAPC webhook but it didn't even invoke the webhook - I think the request gets rejected before it reaches the webhooks by the K8s server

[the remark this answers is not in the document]
maxdrib: I think I figured it out. I found the exact same issue described in https://github.com/bitnami-labs/sealed-secrets/issues/971, and saw the fix in https://github.com/bitnami-labs/sealed-secrets/pull/979/files, which was to make the field nullable. This will require a change in CAPC

pkg/providers/cloudstack/apibuilder.go
maxdrib: I think sharing name generation logic makes sense, but I don't think it could be handled in this PR, which is intended to help make migration to cluster lifecycle api easier. I will create an issue for this

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
maxdrib: I think I figured it out. It was the same issue as https://github.com/bitnami-labs/sealed-secrets/issues/971. I've created a PR in CAPC to make the field nullable: https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/176, so i think we can proceed without the objectmeta being set here.

pkg/providers/cloudstack/cloudstack.go
maxdrib: Looks like it does, thanks for the tip

pkg/providers/cloudstack/apibuilder.go
maxdrib: done

pkg/providers/cloudstack/apibuilder.go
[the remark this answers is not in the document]
maxdrib: Fixed in the latest CAPC bundle

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,113 @@
...
+		deployment := machineDeployment(clusterSpec, workerNodeGroupConfig,
+			kubeadmConfigTemplates[workerNodeGroupConfig.Name],
+			machineTemplates[workerNodeGroupConfig.Name],
+		)
+		m[workerNodeGroupConfig.Name] = &deployment
+	}
+	return m
+}
+
+func fillMachineTemplateAnnotations(machineConfig *v1alpha1.CloudStackMachineConfigSpec) map[string]string {
```
g-gaston, line 40: nit:
```suggestion
func generateMachineTemplateAnnotations(machineConfig *v1alpha1.CloudStackMachineConfigSpec) map[string]string {
```

pkg/providers/cloudstack/apibuilder.go
```diff
@@ -0,0 +1,113 @@
...
+		CustomSize: machineConfig.DiskOffering.CustomSize,
+		MountPath:  machineConfig.DiskOffering.MountPath,
+		Device:     machineConfig.DiskOffering.Device,
+		Filesystem: machineConfig.DiskOffering.Filesystem,
+		Label:      machineConfig.DiskOffering.Label,
+	}
+}
+
+// MachineTemplate returns a generated CloudStackMachineTemplate object for a given EKS-A CloudStackMachineConfig.
+func MachineTemplate(name string, machineConfig *v1alpha1.CloudStackMachineConfigSpec) *cloudstackv1.CloudStackMachineTemplate {
```
g-gaston, line 81: curious why pass the `CloudStackMachineConfigSpec` instead of the `CloudStackMachineConfig`?

maxdrib: The main reason was encapsulation. There's no need for this method to have access to the whole CloudStackMachineConfig - it only needs the spec

g-gaston: I see
I prefer to pass the machine templates for consistency. The spec itself misses some context, it doesn't really represent anything by itself. I understand not passing big things like the cluster.Spec, but in this case it seems we lose more than we win.

That said, I don't have big concerns, I'll leave it up to you, lgtm

vivek-koppuru (approved):
/approve

PR #3388, opened 2022-09-19
Add docker provider autoscaling config

jonathanmeier5:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/3387

*Description of changes:*
Adds appropriate annotations to machine deployment for autoscaling.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/docker/config/template-md.yaml
+++ b/pkg/providers/docker/config/template-md.yaml
@@ -33,6 +33,11 @@ kind: MachineDeployment
 metadata:
   name: {{.workerNodeGroupName}}
   namespace: {{.eksaSystemNamespace}}
+{{- if .autoscalingConfig }}
+  annotations:
+    cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "{{ printf "%#v" .autoscalingConfig.MinCount }}"
+    cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "{{ printf "%#v" .autoscalingConfig.MaxCount }}"
+{{- end }}
 spec:
   clusterName: {{.clusterName}}
   replicas: {{.workerReplicas}}
--- a/pkg/providers/docker/config/template-md.yaml
+++ b/pkg/providers/docker/config/template-md.yaml
@@ -35,8 +35,8 @@ metadata:
   namespace: {{.eksaSystemNamespace}}
 {{- if .autoscalingConfig }}
   annotations:
-    cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "{{ printf "%#v" .autoscalingConfig.MinCount }}"
-    cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "{{ printf "%#v" .autoscalingConfig.MaxCount }}"
+    cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "{{ .autoscalingConfig.MinCount }}"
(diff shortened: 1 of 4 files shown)


pkg/providers/docker/config/template-md.yaml
```diff
@@ -33,6 +33,11 @@ kind: MachineDeployment
 metadata:
   name: {{.workerNodeGroupName}}
   namespace: {{.eksaSystemNamespace}}
+{{- if .autoscalingConfig }}
+  annotations:
+    cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "{{ printf "%#v" .autoscalingConfig.MinCount }}"
```
jiayiwang7, line 38: I don't think you need to convert to string here
```suggestion
    cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: {{ .autoscalingConfig.MinCount }}
```

pkg/providers/docker/docker_test.go
```diff
@@ -288,6 +288,21 @@ func TestProviderGenerateDeploymentFileSuccessUpdateMachineTemplate(t *testing.T
...
+				s.Cluster.Spec.KubernetesVersion = "1.19"
+				s.Cluster.Spec.ClusterNetwork.Pods.CidrBlocks = []string{"192.168.0.0/16"}
+				s.Cluster.Spec.ClusterNetwork.Services.CidrBlocks = []string{"10.128.0.0/12"}
+				s.Cluster.Spec.ControlPlaneConfiguration.Count = 3
+				s.VersionsBundle = versionsBundle
+				s.Cluster.Spec.ExternalEtcdConfiguration = &v1alpha1.ExternalEtcdConfiguration{Count: 3}
+				s.Cluster.Spec.WorkerNodeGroupConfigurations = []v1alpha1.WorkerNodeGroupConfiguration{{Count: 3, MachineGroupRef: &v1alpha1.Ref{Name: "test-cluster"}, Name: "md-0", AutoScalingConfiguration: &v1alpha1.AutoScalingConfiguration{MinCount: 3, MaxCount: 5}}}
+			}),
+			wantCPFile: "testdata/valid_deployment_cp_expected.yaml",
+			wantMDFile: "testdata/valid_autoscaler_deployment_md_expected.yaml",
```
jiayiwang7, line 304: where is the testdata/valid_autoscaler_deployment_md_expected.yaml ?

pkg/providers/docker/config/template-md.yaml
jonathanmeier5: I may be missing something, but from what I can tell keys/values of annotations [need to be strings](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/): 
```
"Note: The keys and the values in the map must be strings. In other words, you cannot use numeric, boolean, list or other types for either the keys or the values."
```

That's why the quotes are there. But maybe I've missed something?

jonathanmeier5: I removed the printf's. Somehow I couldn't get it to work without them earlier, but now the tests pass without a problem. Not sure what the problem was.

jiayiwang7: Yea Kubernetes annotation needs string. But we build the spec from template: byte array -> string yaml. I think it should not matter if you convert to string in the template or not. In the end it is going to be string format.

jiayiwang7 (approved):

PR #3392, opened 2022-09-19
Adding tools to fetch external etcd cluster from EKS-A Controller

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR adds the functions for the EKS-A controller to reconcile upgrades and consider an external etcd cluster.

*Testing (if applicable):*
Successfully scaled cluster with changes from https://github.com/aws/eks-anywhere/pull/2938, unit tests

### DIFF
--- a/pkg/controller/clusterapi_test.go
+++ b/pkg/controller/clusterapi_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


pkg/controller/clusters/clusterapi.go
```diff
@@ -37,3 +39,31 @@ func CheckControlPlaneReady(ctx context.Context, client client.Client, log logr.
 	log.Info("CAPI control plane is ready")
 	return controller.Result{}, nil
 }
+
+// CheckEtcdReady is a controller helper to check whether an external etcdadm cluster for
+// an eks-a cluster is ready or not. This is intended to be used from cluster reconcilers
+// due its signature and that it returns controller results with appropriate wait times whenever
+// the cluster is not ready
+func CheckEtcdReady(ctx context.Context, client client.Client, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
```
g-gaston, line 47: What is this going to be used for?
Same got the `Get` method

maxdrib: This would add on to the model introduced in the snow controller. Currently, every reconcile runs the following phases: 

		r.ValidateMachineConfigs,
		r.ReconcileControlPlane,
		r.CheckControlPlaneReady,
		r.ReconcileCNI,
		r.ReconcileWorkers,

For supporting external etcd, I believe we need to add another phase between `ReconcileControlPlane` and `CheckControlPlaneReady` to check the etcd is ready. This method defines that phase.

vivek-koppuru: Are you suggesting we make this change for Snow also or keep it independent from it?

maxdrib: I was under the impression that Snow has some fundamental limitations on supporting external etcd since that requires machine objects to have an IP which may not always be the case with the snow provider. So I'm inclined to keep it independent for now, based on that assumption

g-gaston: Ahhh I see now. However, I'm not completely sure we need this

There is no action between `CheckEtcdReady` and `CheckControlPlaneReady`. So, why would we have both as reconciliation phases?

Maybe this could become necessary in the future if we need to perform some actions in between or we want to bubble up these "phases" in the cluster status. I would suggest to keep it simple for now, get it to work and add more stuff as we see need based on our experience using it.

Opinions?

maxdrib: In the case of an upgrade with the CLI, we explicitly check for etcd upgrade being ready before proceeding to wait for the control plane upgrade to be completed. Why should the controller behave differently and jump straight to waiting for the control plane to be ready? Does the control plane become NotReady during an unstacked etcd upgrade?

pkg/controller/clusterapi_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
```
gwesterfieldjr, line 7: Can you not reference github.com/aws/etcdadm-controller instead of mrajashree fork?

pkg/controller/clusters/clusterapi.go
g-gaston: As far a the difference between controller and CLI, the CLi does a lot of things the controller doesn't need for two main reasons: provide more user feedback and avoid transient failures (since it doesn't have a retry mechanism, if it encounters an issue, it will fail).

That said, you make a good point, the control plane might not become not-ready while etcd is upgraded. That said, I don't think we need wait until it does before moving on with the next phase, as long as the api server is available we can continue.

However, I would like to understand a bit better how this upgrade process works (interaction between the different capi entities and their status). @taneyland is currently doing a spike on this topic in regards with implementing it in the controller. I would suggest keeping this PR open until we finish that spike and have a better idea of what's needed. Is that fair @maxdrib?

maxdrib: Sure, sounds good.

/hold until @taneyland completes her design spike to understand relationship between external etcd upgrade and KCP upgrade

pkg/controller/clusterapi_test.go
maxdrib: There's a dependency in go.mod on the mrajashree fork, so changing to the aws fork will require a bigger change which I'm not inclined to do here.

jonathanmeier5:
As far as I'm aware this work is now being captured in the lifecycle controller project for CloudStack, @ddjjia can correct me if I'm wrong. Closing this PR out.

PR #3405, opened 2022-09-21
Add os family validation when performing upgrade for vsphere

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
When looking for whether we validated `OSFamily` changes during upgrade, I noticed that we had it defined in our webhook here but not through the CLI.

https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/vspheremachineconfig_webhook.go#L67

When I tested if we can upgrade a cluster from Ubuntu to Bottlerocket, it failed due to some of the values of the mounts that CAPI expects for Bottlerocket being immutable. We don't support this at this moment then, so adding the validation. But this is something that we should consider in the future on how to get users from Ubuntu clusters to Bottlerocket if they prefer.

*Testing (if applicable):*
unit testing


TerryHowe (approved):
/lgtm

PR #3409, opened 2022-09-21
Move docker validation to api level

taneyland:
*Issue #, if available:*

*Description of changes:*
Moves docker validation to api level to use both with eksa cli commands & in the controller

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -168,7 +168,7 @@ var clusterConfigValidations = []func(*Cluster) error{
 	validateClusterConfigName,
 	// TODO(chrisdoherty) Uncomment and fix unit tests. This broke _lots_ of unit tests so will
 	// return shortly.
-	// validateControlPlaneEndpoint,
+	validateControlPlaneEndpoint,
 	// validateControlPlaneMachineRef,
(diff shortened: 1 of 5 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -389,6 +389,13 @@ func validateControlPlaneLabels(clusterConfig *Cluster) error {
 	return nil
 }
 
+func validateControlPlaneEndpoint(clusterConfig *Cluster) error {
+	if clusterConfig.Spec.DatacenterRef.Kind == DockerDatacenterKind && clusterConfig.Spec.ControlPlaneConfiguration.Endpoint != nil && clusterConfig.Spec.ControlPlaneConfiguration.Endpoint.Host != "" {
```
g-gaston, line 393: If the validation only applies to docker (unfortunate) maybe we should find a different place this. It's an edge case, we never had to run validations at the cluster level that were provider specific 

It's only one for now so I'm ok with this, but can we create a task in the project to come back to this?

taneyland: Yes I'll add a follow up task to the project!

mitalipaygude (approved):

PR #3413, opened 2022-09-21
Pause cluster and datacenter webhook validation during upgrade

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

During CLI upgrade w controller v2, the `Applying eksa yaml resources to cluster` step failed due to the eksa cluster webhook validation error: `admission.anywhere.amazonaws.com\" denied the request: upgrading self managed clusters is not supported`

We are adding the paused annotation to cluster/datacenterconfig during CLI upgrade, similar to create.

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -910,6 +910,9 @@ func (c *ClusterManager) CreateEKSANamespace(ctx context.Context, cluster *types
 	return c.clusterClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace)
 }
 
+// CreateEKSAResources applies the eks-a cluster specs (cluster, datacenterconfig, machine configs, etc.), as well as the
+// release bundle to the cluster. Before applying the spec, we pause eksa controller cluster and datacenter webhook validation
+// so that the cluster spec can be created or updated in the cluster without webhook validation error.
 func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec,
 	datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig,
 ) error {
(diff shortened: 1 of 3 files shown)


pkg/clustermanager/cluster_manager.go
```diff
@@ -918,6 +918,11 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
 			return err
 		}
 	}
+
+	// pause eksa controller cluster and datacenter webhook validation during apply
+	clusterSpec.Cluster.PauseReconcile()
```
g-gaston, line 923: Can we document this in the method doc comment? It's not obvious that this method will do this or why

g-gaston (approved):
/lgtm
/hold until tested

pkg/workflows/create.go
```diff
@@ -366,10 +366,6 @@ func (s *InstallEksaComponentsTask) Run(ctx context.Context, commandContext *tas
 	datacenterConfig := commandContext.Provider.DatacenterConfig(commandContext.ClusterSpec)
 	machineConfigs := commandContext.Provider.MachineConfigs(commandContext.ClusterSpec)
 
-	// this disables create-webhook validation during create
-	commandContext.ClusterSpec.Cluster.PauseReconcile()
```
vivek-koppuru, line 370: We did not support this for upgrade before? I expected us to have had this already.

jiayiwang7: we do not have this upgrade validation in cluster webhook before
https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_webhook.go#L87

vivek-koppuru: That was because we only supported a subset of changes right? For some reason I thought we always paused the controller for upgrade too because we don't want the controller to mess the changes that the CLI is making.

jiayiwang7:
upgrade test passed on snow admin instance.

/unhold
/approve

pkg/workflows/create.go
jiayiwang7: > we always paused the controller for upgrade too because we don't want the controller to mess the changes that the CLI is making

we already do that. 

this logic is to add paused annotation in the cluster spec that will be applied. The existing object in cluster already has paused annotation, but we need to keep that annotation in order to pass the validation.

Issue #3420, opened 2022-09-22
Control Plane Node missing no schedule taint


junshun:
**What happened**:
Created cluster with provided dev-cluster.yaml (Default generated by CLI with only change of kubernetesVersion from 1.23 -> 1.22). After cluster creation checking the taints on the nodes 
```
/local/home/junshun/DockerCluster kubectl get nodes -o custom-columns=NAME:.metadata.name,TAINTS:.spec.taints           
NAME                                TAINTS
dev-cluster-md-0-6dd8979879-f5d6j   <none>
dev-cluster-ml9r9                   <none>
```
Built eks-a binary out of main as of commit (5efbd99bfc2cfa8dcdeeb40172bd12b33ad65754)

**What you expected to happen**:
With 11.3 release release, control plane nodes are tainted with no schedule as expected with same cluster spec.
```
/local/home/junshun/DockerCluster kubectl get nodes -o custom-columns=NAME:.metadata.name,TAINTS:.spec.taints
NAME                                TAINTS
dev-cluster-757w5                   [map[effect:NoSchedule key:node-role.kubernetes.io/master]]
dev-cluster-md-0-7dbf5658dc-cw4cw   <none>
```
**How to reproduce it (as minimally and precisely as possible)**:
Build main and create cluster with dev-cluster.yaml
**Anything else we need to know?**:
Issue first seen with vsphere and reproduced with docker.

[dev-cluster.yaml.txt](https://github.com/aws/eks-anywhere/files/9625861/dev-cluster.yaml.txt)

Main branch build create cluster log with v6
[main-dev-cluster.log](https://github.com/aws/eks-anywhere/files/9625866/main-dev-cluster.log)

11.3 release create cluster log with v6
[11-3-dev-cluster.log](https://github.com/aws/eks-anywhere/files/9625867/11-3-dev-cluster.log)

chrisdoherty4:
@d8660091 Could this be related to the work you're doing for single node clusters?

d8660091:
Yes, I think it has some relations. I'm investigating it with Steven

drewvanstone:
@d8660091 has this been resolved?

d8660091:
I think so, I'll close this issue after double checking with junshun

PR #3422, opened 2022-09-22
Fix empty taints override issue

d8660091:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/3420

*Description of changes:* When taints is nil or [] in cluster.yaml, pass nil instead of [] to cluster-api, so that
control-plane node is not untainted (previously, [] is treated as nil by cluster-api)

*Testing (if applicable):* Unit testing

### DIFF
--- a/pkg/providers/cloudstack/config/template-cp.yaml
+++ b/pkg/providers/cloudstack/config/template-cp.yaml
@@ -342,8 +342,6 @@ spec:
           timeAdded: {{ .TimeAdded }}
 {{- end }}
 {{- end }}
-{{- else}}
-        taints: []
 {{- end }}
     joinConfiguration:
       nodeRegistration:
@@ -366,8 +364,6 @@ spec:
           timeAdded: {{ .TimeAdded }}
 {{- end }}
 {{- end }}
-{{- else}}
-        taints: []
 {{- end }}
     preKubeadmCommands:
     - swapoff -a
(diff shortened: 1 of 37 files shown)


junshun:
I believe you also need to update clusterapi/apibuilder.go in the KubeadmControlPlane function

d8660091:
> I believe you also need to update clusterapi/apibuilder.go in the KubeadmControlPlane function

Perhaps not. 

The problem is that template-cp set the taints to empty when cluster.yaml have a nil taints field, i.e.:
`ClusterAPITaints = []` when `clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Taints = nil`

However, at the KubeadmControlPlane function, the taints is set directly from the field in cluster.yml
`ClusterAPITaints = clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Taints`, in this case:
`ClusterAPITaints = nil` when `clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Taints = nil`.

The fact that KubeadmControlPlane function is working as expected can be also proved by that the generated cluster-api  yaml file:
https://github.com/d8660091/eks-anywhere/blob/single-node-tinkerbell/pkg/providers/snow/testdata/expected_results_main_cp.yaml#L88
We can see that nodeRegistration.taints is nil, which should be [] if it has the same issue as other providers.

Issue #3404, opened 2022-09-20
vsphere CSI nodes in crash loop backoff because `csi-vsphere-config` secret is missing vSphere creds


abhinavmpandey08:
**What happened**:
`vsphere-csi-controller` and `vsphere-csi-node` pods are in `CrashLoopBackoff` on workload cluster after performing upgrade.

Checking the logs shows 
```
{"level":"error","time":"2022-09-19T20:15:16.271474904Z","caller":"config/config.go:272","msg":"vcConfig.User is empty for vc vsphere.testlab.local"
```

Finally, describing the `csi-vsphere-config` secret shows that the `user` and `password` fields are set to empty values while all the other fields are set properly
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: csi-vsphere-config
  namespace: kube-system
stringData:
  csi-vsphere.conf: |+
    [Global]
    cluster-id = "default/abhinav-workload"
    thumbprint = ""

    [VirtualCenter "vsphere.testlab.local"]
    user = ""
    password = ""
    datacenters = "Datacenter"
    insecure-flag = "false"

    [Network]
    public-network = "/Datacenter/network/network-1"
type: Opaque
```

**How to reproduce it (as minimally and precisely as possible)**:
Create a workload cluster from an existing management cluster and upgrade the workload cluster.

**Environment**:
- EKS Anywhere Release: `v0.11.1`
- EKS Distro Release: `v1.23`
- OS: `BottleRocket`

jonathanmeier5:
I wasn't able to reproduce the issue with missing secrets building an artifact from source tagged at `v0.11.1`. 

I was able to reproduce the `vsphere-csi-controller` and `vsphere-csi-node` pods are in `CrashLoopBackoff` though, both with the following log message:
```
W0921 13:45:36.467668       1 connection.go:173] Still connecting to unix:///csi/csi.sock
```

Even before upgrade, I see warnings that seem significant in the `vsphere-csi-node` pod descriptions:
```
Events:
  Type     Reason            Age                    From               Message
  ----     ------            ----                   ----               -------
  Normal   Scheduled         35m                    default-scheduler  Successfully assigned kube-system/vsphere-csi-node-fn8gx to 10.61.250.110
  Warning  NetworkNotReady   35m (x3 over 35m)      kubelet            network is not ready: container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
[...]

PR #3424, opened 2022-09-22
Inject vSphere Creds in Controller

jonathanmeier5:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/3404

*Description of changes:*
The EKSA Controller didn't have necessary vsphere credentials. As a result, when it reconciled eksa-system:csi-vsphere-config it deleted the username and password.

*Testing (if applicable):*
How I tested:

* create:
    * create cluster on main
* upgrade:
    * create cluster on v0.11.1
    * upgrade on release-0.11 with my changes on top of it. Secret `eksa-system:vsphere-credentials` was correctly updated.
    * ran old controller while upgrading. `eksa-system:vsphere-credentials` was unaffected but controller continued to did still overwrite csi-vsphere-config secret
    * ran new controller after upgrade and csi-vsphere-config reconciled correctly with appropriately populated config information derived from `eksa-system:vsphere-credentials` 

### DIFF
--- a/controllers/resource/internal/vsphere.go
+++ b/controllers/resource/internal/vsphere.go
@@ -1,0 +1,43 @@
+package internal
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GetVSphereCredValues(credSecret *corev1.Secret) (map[string]string, error) {
+	usernameBytes, ok := credSecret.Data["username"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve username from secret")
+	}
+	passwordBytes, ok := credSecret.Data["password"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve password from secret")
+	}
+	usernameCSIBytes, ok := credSecret.Data["usernameCSI"]
+	if !ok {
+		usernameCSIBytes = usernameBytes
+	}
+	passwordCSIBytes, ok := credSecret.Data["passwordCSI"]
+	if !ok {
+		passwordCSIBytes = passwordBytes
+	}
+	usernameCPBytes, ok := credSecret.Data["usernameCP"]
+	if !ok {
+		usernameCPBytes = usernameBytes
+	}
+	passwordCPBytes, ok := credSecret.Data["passwordCP"]
+	if !ok {
+		passwordCPBytes = passwordBytes
+	}
+
+	values := map[string]string{}
+	values["eksaVsphereUsername"] = string(usernameBytes)
+	values["eksaVspherePassword"] = string(passwordBytes)
+	values["eksaCSIUsername"] = string(usernameCSIBytes)
+	values["eksaCSIPassword"] = string(passwordCSIBytes)
+	values["eksaCloudProviderUsername"] = string(usernameCPBytes)
+	values["eksaCloudProviderPassword"] = string(passwordCPBytes)
+	return values, nil
+}
--- a/controllers/resource/reconciler_test.go
+++ b/controllers/resource/reconciler_test.go
@@ -109,6 +109,12 @@ var nutanixWorkerMachineConfigSpec string
 //go:embed testdata/nutanix/etcdMachineConfig.yaml
 var nutanixEtcdMachineConfigSpec string
 
+func getSecret() *corev1.Secret {
+	return &corev1.Secret{
+		Data: map[string][]byte{"username": []byte("username"), "password": []byte("password"), "usernameCSI": []byte("usernameCSI"), "passwordCSI": []byte("passwordCSI"), "usernameCP": []byte("usernameCP"), "passwordCP": []byte("passwordCP")},
+	}
+}
+
 func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 	type args struct {
 		objectKey types.NamespacedName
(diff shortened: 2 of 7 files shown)


controllers/resource/template.go
```diff
@@ -163,6 +163,22 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
...
+		return nil, fmt.Errorf("unable to retrieve passwordCSI from secret")
+	}
+	usernameCPBytes, ok := credSecret.Data["usernameCP"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve usernameCP from secret")
+	}
+	passwordCPBytes, ok := credSecret.Data["passwordCP"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieve passwordCP from secret")
+	}
```
jonathanmeier5, line 181: I tried moving the secret objects we discussed from `template-cp.yaml` into `secrets.yaml` but cluster creation failed. They need to be there together with the cluster config object it seems.

I did find thoug that we are already using the 2nd approach we discussed of storing credentials in a secret and then reading them from the controller when reconciling.

Unfortunately they are reading a global secret file.

I need to wrap this up to get back to cluster autoscaler work, but I think we should raise that the vSphere provider currently only really supports a single set of credentials.

I can open a ticket to audit the codebase a fix it up to support multiple credentials by cluster in future if we want.

pkg/providers/vsphere/config/secret.yaml
```diff
@@ -7,6 +7,10 @@ type: kubernetes.io/basic-auth
 data:
   username: {{.vsphereUsername | b64enc}}
   password: {{.vspherePassword | b64enc}}
+  usernameCSI: "{{.eksaCSIUsername | b64enc}}"
+  passwordCSI: "{{.eksaCSIPassword | b64enc}}"
+  usernameCP: "{{.eksaCloudProviderUsername | b64enc}}"
+  passwordCP: "{{.eksaCloudProviderPassword | b64enc}}"
```
g-gaston, line 13: what are these for?

controllers/resource/template.go
```diff
@@ -61,6 +62,42 @@ type NutanixTemplate struct {
 	now anywhereTypes.NowFunc
 }
 
+func GetVSphereCredValues(credSecret *corev1.Secret) (map[string]string, error) {
```
g-gaston, line 65: does this need to be public?

g-gaston (commented):
Looks good!

controllers/resource/template.go
jonathanmeier5: It doesn't...but the existing public methods on this package are very difficult to test. The [test cases](https://github.com/aws/eks-anywhere/blob/808b159fa444a67fdab675e90d55ea4c93cb196b/controllers/resource/reconciler_test.go#L112) involve ~100 lines of setup each. Figuring all of that out to test ~6 lines of code in order to make this method private (which it probably should be) just didn't feel worth it.

pkg/providers/vsphere/config/secret.yaml
jonathanmeier5: This seems to be a little known [feature](https://anywhere.eks.amazonaws.com/docs/reference/clusterspec/vsphere/#optional-vsphere-credentials) of ours that allows developers to provide different sets of vSphere creds for cluster management, the CSI driver, and the vSphere Cloud Provider.

controllers/resource/template.go
g-gaston: Got you
then I would suggest moving this method to `controllers/resource/internal/vsphere.go`
That way we can test it individually but it's not exposed to other modules/packages

pkg/providers/vsphere/config/secret.yaml
g-gaston: Oh sorry, I meant why are we adding them to this secret :)

controllers/resource/template.go
jonathanmeier5: That's perfect - thanks!

pkg/providers/vsphere/config/secret.yaml
jonathanmeier5: Haha, fair.

The controller reads these secret values and injects them into the CP template when doing its reconciliation. This [cpOpt](https://github.com/aws/eks-anywhere/pull/3424/files#diff-42413ee5a1c110d438032dcbc28aa33f9af26f3235f367868883b4b5b3447579R200) string map is eventually injected into `template-cp.yaml` by the controller.

jonathanmeier5: The request against the secret happens here: https://github.com/aws/eks-anywhere/pull/3424/files#diff-42413ee5a1c110d438032dcbc28aa33f9af26f3235f367868883b4b5b3447579R190

controllers/resource/internal/test/util.go
```diff
@@ -0,0 +1,11 @@
+package util
```
chrisdoherty4, line 1: I think you can get away with duplicating this where its needed - seems to be test only. 

On a side note, package names should be the same as their directory names and `util` packages are strongly discouraged. Instead, prefer duplication until you can see a bundle of related utility functions and then consider creating a fosuced utility package like `secretutil`.

controllers/resource/internal/vsphere.go
```diff
@@ -0,0 +1,43 @@
+package internal
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GetVSphereCredValues(credSecret *corev1.Secret) (map[string]string, error) {
```
chrisdoherty4, line 9: Is it possible, given the simplicity of this func, to encorporate this in the `reconciler_test.go` somehow so we can do away with the `test` pkg?

chrisdoherty4 (changes requested):

controllers/resource/reconciler_test.go
```diff
@@ -109,6 +110,12 @@ var nutanixWorkerMachineConfigSpec string
 //go:embed testdata/nutanix/etcdMachineConfig.yaml
 var nutanixEtcdMachineConfigSpec string
 
+func getSecret() *corev1.Secret {
```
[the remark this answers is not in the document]
jonathanmeier5: In an offline convo with @g-gaston we decided to duplicate the test function across two test packages rather than use the `internal/test` package I had up.

controllers/resource/internal/vsphere.go
jonathanmeier5: This isn't a test function, it's used [here](https://github.com/aws/eks-anywhere/pull/3424/files#diff-42413ee5a1c110d438032dcbc28aa33f9af26f3235f367868883b4b5b3447579R158).

controllers/resource/internal/test/util.go
jonathanmeier5: This is really helpful, thanks 👍 Came to the same place after an offline conversation with @g-gaston .

pkg/providers/vsphere/config/secret.yaml
g-gaston: I see. The problem is existing clusters won't have these values in their secret, so the controller will fail reconciling. I fear we might need to read to other secrets for backwards compatibility.

jonathanmeier5: Right, of course. I'll default to the `username` and `password` values if the others aren't present.

controllers/resource/internal/vsphere.go
chrisdoherty4: Also, I see the confusion with my original comment 🤦🏻 . The above comment elaborates on my train of thought.

jonathanmeier5: Yeah, fair question. @g-gaston and I landed on this approach based on [this conversation](https://github.com/aws/eks-anywhere/pull/3424#discussion_r978872979).

chrisdoherty4 (approved):

PR #3432, opened 2022-09-22
Provider CAPI yaml to api structs spike

g-gaston:
*Description of changes:*
> **IMPORTANT**: this is a spike, it's not intended for production code. It just tries to prove the viability of an idea and it doesn't represent a proposal for mergeable code.

As mentioned in the [guide](https://github.com/aws/eks-anywhere/blob/main/docs/developer/cluster-lifecycle-api-provider-integration.md#immutable-objects), in order to integrate with the new cluster lifecycle controller, providers need to be able to generate CAPI specs without relying on old and new versions of the eks-a objects (as they currently do in the CLI). The proposed approach is to compare outputs (CAPI objects) instead of inputs (eks-a objects).

This presents a challenge with the current go template generation for yaml specs. However, re-implementing all the spec generation [using go API structs](https://github.com/aws/eks-anywhere/blob/main/docs/developer/cluster-lifecycle-api-provider-integration.md#templating-vs-api-structs) could take considerable amount of effort, specially for complex providers like vSphere.

In order to speed up the full cluster lifecycle project, reusing the yaml templates could be an option. This PR proves that is actually possible to do that and presents an implementation that can be reused across providers.

Only the CP spec has been implemented and only for docker. The one worker nodes should be similar. The vSphere implementation will require more work than the docker one since the CP spec is considerable bigger, including things like CRS.


pkg/clusterapi/controlplane.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return objs
+}
+
+// UpdateImmutableObjectNames checks if any control plane immutable objects have changed y comparing the new definition
+// with the current state of the cluster. If they had, it generates a new name for them by increasing a monotonic number
+// at the end of the name
+// This is applied to all provider machine templates
+// machineTemplateComparator receives the new definition of a machine template with the name of the current one
```
jiayiwang7, line 60: this is awesome!
wonder if we want to be more generic like `capiObjectComparator` since there are cases we need new `KubeadmConfigTemplate` or potentially more objects need comparison

pkg/clusterapi/yaml/parse.go
```diff
@@ -0,0 +1,136 @@
...
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// ProcessCluster finds the CAPI cluster in the parsed objects and sets it in ControlPlane
+func ProcessCluster[C, M clusterapi.Object, P clusterapi.ProviderControlPlane](cp *clusterapi.ControlPlane[C, M, P], lookup yamlutil.ObjectLookup) {
+	for _, obj := range lookup {
+		if obj.GetObjectKind().GroupVersionKind().Kind == "Cluster" {
```
maxdrib, line 17: what if the Cluster isn't found? same for the other Process* methods

Issue #3412, opened 2022-09-21
Add validations for tinkerbell `upgrade` to make unsupported fields immutable


pokearu:
## Description
Currently we support k8s version upgrade and scale up/down for `upgrade` with Tinkerbell provider.
We require validations to make the unsupported objects from the clusterconfig to be immutable.

```yaml
---
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: TinkerbellDatacenterConfig
...

---
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: TinkerbellMachineConfig
...

---
apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: TinkerbellTemplateConfig
...
```

This issue also helps prevent #3341

PR #3434, opened 2022-09-23
Bare metal upgrade: Add immutability checks...

ptrivedi:
...for fields not currently supported by bare metal upgrade

*Issue #3412*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/upgrade.go
+++ b/pkg/providers/tinkerbell/upgrade.go
@@ -3,6 +3,7 @@ package tinkerbell
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
@@ -25,8 +26,7 @@ func NeedsNewControlPlaneTemplate(oldSpec, newSpec *cluster.Spec, oldVdc, newVdc
 		return true
 	}
 
-	fieldsChanged := AnyImmutableFieldChanged(oldVdc, newVdc, oldTmc, newTmc)
-	return fieldsChanged
+	return AnyImmutableFieldChanged(oldVdc, newVdc, oldTmc, newTmc)
 }
 
 func NeedsNewWorkloadTemplate(oldSpec, newSpec *cluster.Spec, oldVdc, newVdc *v1alpha1.TinkerbellDatacenterConfig, oldTmc, newTmc *v1alpha1.TinkerbellMachineConfig) bool {
@@ -58,7 +58,6 @@ func NeedsNewEtcdTemplate(oldSpec, newSpec *cluster.Spec, oldVdc, newVdc *v1alph
 }
 
 func AnyImmutableFieldChanged(oldVdc, newVdc *v1alpha1.TinkerbellDatacenterConfig, oldTmc, newTmc *v1alpha1.TinkerbellMachineConfig) bool {
-	// @TODO: Add immutable fields check here
 	return false
 }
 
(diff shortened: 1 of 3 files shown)


pkg/providers/tinkerbell/upgrade.go
```diff
@@ -238,3 +275,33 @@ func (p *Provider) isScaleUpDown(currentSpec *cluster.Spec, newSpec *cluster.Spe
...
+	prevMachineConfig, err := p.providerKubectlClient.GetEksaTinkerbellMachineConfig(ctx, newConfig.Name, cluster.KubeconfigFile, clusterSpec.Cluster.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if newConfig.Spec.OSFamily != prevMachineConfig.Spec.OSFamily {
+		return fmt.Errorf("spec.osFamily os immutable. Previous value %v,   New value %v", prevMachineConfig.Spec.OSFamily, newConfig.Spec.OSFamily)
+	}
+
+	if newConfig.Spec.Users[0].SshAuthorizedKeys[0] != prevMachineConfig.Spec.Users[0].SshAuthorizedKeys[0] {
```
pokearu, line 289: any reason we are checking only the first key and first user? Dont we want to check them all? 
There should be a function for it 
https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L423

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -209,6 +207,45 @@ func (p *Provider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpe
...
+
+		if _, ok = prevMachineConfigRefs[machineConfig.Name]; ok {
+			err = p.validateMachineConfigImmutability(ctx, cluster, machineConfig, clusterSpec)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if nSpec.TinkerbellIP != oSpec.TinkerbellIP {
```
pokearu, line 242: Can we also check the optional `OSImageURL` and `HookImagesURLPath`? Those should be changed as well.

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -209,6 +207,45 @@ func (p *Provider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpe
...
+
+	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
+
+	for _, machineConfigRef := range clusterSpec.Cluster.MachineConfigRefs() {
+		machineConfig, ok := p.machineConfigs[machineConfigRef.Name]
+		if !ok {
+			return fmt.Errorf("cannot find machine config %s in tinkerbell provider machine configs", machineConfigRef.Name)
+		}
+
+		if _, ok = prevMachineConfigRefs[machineConfig.Name]; ok {
```
pokearu, line 234: If this returns false, isn't that also an error? We cant add/remove `machineconfigs` right?

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -209,6 +207,45 @@ func (p *Provider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpe
...
+	datacenterConfig := p.datacenterConfig
+
+	oSpec := prevDatacenterConfig.Spec
+	nSpec := datacenterConfig.Spec
+
+	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
+
+	for _, machineConfigRef := range clusterSpec.Cluster.MachineConfigRefs() {
+		machineConfig, ok := p.machineConfigs[machineConfigRef.Name]
+		if !ok {
```
pokearu, line 230: nit: Can we make this a single line?

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -209,6 +207,45 @@ func (p *Provider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpe
...
+
+	prevDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, prevSpec.Spec.DatacenterRef.Name, cluster.KubeconfigFile, prevSpec.Namespace)
+	if err != nil {
+		return err
+	}
+
+	datacenterConfig := p.datacenterConfig
+
+	oSpec := prevDatacenterConfig.Spec
+	nSpec := datacenterConfig.Spec
```
pokearu, line 224: nit: Can we directly do `p.datacenterConfig.Spec`? I see `datacenter` is not used anywhere else.

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -238,3 +275,33 @@ func (p *Provider) isScaleUpDown(currentSpec *cluster.Spec, newSpec *cluster.Spe
 
 	return false
 }
+
+func (p *Provider) validateMachineConfigImmutability(ctx context.Context, cluster *types.Cluster, newConfig *v1alpha1.TinkerbellMachineConfig, clusterSpec *cluster.Spec) error {
```
pokearu, line 279: Can we also check the `HardwareSelector`? I think that would also cause a rolling update because its part of the template-cp.

pokearu (commented):
Looks good!
Tagging issue #3412

pkg/providers/tinkerbell/upgrade.go
ptrivedi: > EKS-A has only ever supported one User and one sshAuthorizedKey. You will see everywhere in the code that we only deal with the first element [0] of those slices. Looks like there used to be a pre-flight validation to ensure that only one user/key was specified in that slice in the cluster config, but has now disappeared. So now, although multiple users/keys would be accepted by EKS-A, it will silently only use one. That preflight validation needs to come back until multiple users/keys support is implemented and tested.

pkg/providers/tinkerbell/upgrade.go
ptrivedi: > OSImageURL will be and must be changed for rolling upgrades. 

Changing these by themselves would not trigger a rollout. Thinking about the cases where these are changed in conjunction with the bare metal upgrade scenarios of k8s version upgrades or scale ups/downs:
- changing hookImageURL should not matter during rolling upgrades, beneficial to allow that even
- changing osImageURL needs to happen during rolling upgrades (validations for this have been discussed before and need to be dealt with separately)
- for scale up/down specifically, both these fields need to be plugged

pokearu (approved):

PR #3436, opened 2022-09-23
Remove redundant copyright notices

chrisdoherty4:
Remove redundant copyright notices in cmd files. These notices keep getting edited by IDEs as they have unnecessary newline characters but the notices are redundant anyway.


TerryHowe (approved):
/lgtm

PR #3441, opened 2022-09-23
Relocate tasks under /pkg/workflow/task with distinct packages

chrisdoherty4:
Stacks on https://github.com/aws/eks-anywhere/pull/3441/files

Relocate tasks to task specific packages under `/pkg/workflow/task` to enable better reusability among workflows.


pkg/workflow/task.go
```diff
@@ -2,15 +2,18 @@ package workflow
...
-	// GeGetNametID returns a unique identifier for the task.
-	GetName() TaskName
-
 	// RunTask executes the task. Tasks may return a context that should be used in subsequent task
 	// execution.
 	RunTask(context.Context) (context.Context, error)
 }
+
+// namedTask associates a name with a Task in the context of a Workflow to enable hook lookup.
+type namedTask struct {
```
vincentni, line 16: Shall we expose this struct as providers need to access it when creating their hooks?

g-gaston (approved):

pkg/workflow/task.go
chrisdoherty4: What makes you think providers will need this?

vincentni: When a provider sets up its HookRegistrars, it needs to create tasks and pass them to functions like BindPreWorkflowHook, BindPreTaskHook, etc. Do I misunderstand something here?

chrisdoherty4: That sounds correct. A data structure wanting to bind hooks would only need to register a task. That task doesn't need an explicit name though, at least we don't need it right now. I could see us needing to add names if we want to have some logging indicating what's executing; that sounds sensible for debugability. Even then though, I don't think an implementer of hooks would need a `namedTask` because we'd add the name parameter to the `Bind` funcs.

chrisdoherty4: To give an example.

```go
// We currently have no way of specifying names with hooks having removed it from tasks.
// This is how we originally designed hooks anyway and the name was a side effect
// of doubling down on the Task interface for hooks as well.
func (Foo) BindCreateManagementClusterHooks(binder workflow.HookBinder) {
    binder.BindPreTask(management.CreateBootstrapCluster, workflow.HookFunc(func(...) error {
        // do something
    })
}
```

```go
// If we want to identify hooks by a name we can do something like the following
func (Foo) BindCreateManagementClusterHooks(binder workflow.HookBinder) {
    binder.BindPreTask(management.CreateBootstrapCluster, "MySpecialHook", workflow.HookFunc(func(...) error {
        // do something
    })
}
```

vincentni: Sounds good to me. Thanks!

vincentni (approved):

PR #3450, opened 2022-09-26
e2e tools for packages

lewisdiamond:
*Description of changes:*
Extract some tooling from metallb PR as this is needed by Jonah. Metallb pr will be updated to reflect the newer metallb version.

### DIFF
--- a/test/e2e/curated_packages_test.go
+++ b/test/e2e/curated_packages_test.go
@@ -4,32 +4,177 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/aws/eks-anywhere/internal/pkg/api"
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
 const (
-	eksaPackageControllerHelmChartName = "eks-anywhere-packages"
-	eksaPackageControllerHelmURI       = "oci://public.ecr.aws/eks-anywhere/eks-anywhere-packages"
-	eksaPackageControllerHelmVersion   = "0.1.10-eks-a-10"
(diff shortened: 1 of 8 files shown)


test/e2e/curated_packages_test.go
```diff
@@ -4,32 +4,177 @@
...
+		"-o", "jsonpath='{.items[0].metadata.name}'",
+	)
+	if err != nil {
+		return "", err
+	}
+	bundle := bundleBytes.String()
+	return strings.Trim(bundle, "'"), nil
+}
+
+func TestPackagesInstallSimpleFlow(t *testing.T) {
```
ewollesen, line 164: What is this testing? It doesn't fit with the established naming pattern of the other hello-eks-anywhere tests. If it's testing something that isn't already covered, even just a comment explaining what it's testing would be good.

Either way, if this is a test we want to keep, could this please be renamed "TestCPackagesInstallSimpleFlow"? That way it can get picked up by the "all" target in [the Makefile](https://github.com/aws/eks-anywhere/blob/16f4ebe91cc683f70d0470957cb900e93c88e6f5/Makefile#L427).

test/framework/cluster.go
```diff
@@ -830,22 +831,58 @@ func (e *ClusterE2ETest) InstallHelmChart() {
...
+}
+
+func (e *ClusterE2ETest) DeleteNamespace(namespace string) {
+	kubeconfig := e.kubeconfigFilePath()
+	err := e.KubectlClient.DeleteNamespace(context.Background(), kubeconfig, namespace)
+	if err != nil {
+		e.T.Fatalf("Namespace deletion failed for %s", namespace)
+	}
+}
+
```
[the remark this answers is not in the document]
lewisdiamond: They're used in the Metallb tests. This is extracted from the other PR so it will be used.

test/e2e/curated_packages_test.go
lewisdiamond: Seems like it was resurrected from a rebase.

PR #3464, opened 2022-09-27
Refactor bootstrap tasks

chrisdoherty4:
Depends on https://github.com/aws/eks-anywhere/pull/3472

To be written

### DIFF
--- a/pkg/workflow/task/bootstrap/bootstrap.go
+++ b/pkg/workflow/task/bootstrap/bootstrap.go
@@ -2,82 +2,168 @@ package bootstrap
 
 import (
 	"context"
+	"fmt"
 
-	"github.com/aws/eks-anywhere/pkg/bootstrapper"
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime"
+	apimachinerytypes "k8s.io/apimachinery/pkg/types"
+
 	"github.com/aws/eks-anywhere/pkg/cluster"
-	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/workflow/contextutil"
 )
 
-// OptionsRetriever supplies bootstrap cluster options. This is typically satisfied
-// by a provider.
-type OptionsRetriever interface {
-	BootstrapClusterOpts(*cluster.Spec) ([]bootstrapper.BootstrapClusterOption, error)
-}
-
-// Bootstrapper creates and destroys bootstrap clusters. It is satisfied by the bootstrap package
-// and exists predominently for testability.
-type Bootstrapper interface {
-	// CreateCluster creates a new local cluster. It does not contain any EKS-A components.
-	CreateBootstrapCluster(
-		context.Context,
-		*cluster.Spec,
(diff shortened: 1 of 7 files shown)


pkg/workflow/task/bootstrap/bootstrap.go
```diff
@@ -2,75 +2,146 @@ package bootstrap
...
+	})
+
+	return ctx, nil
 }
 
-// DeleteCluster deletes a bootstrap cluster. It expects the bootstrap cluster to be
-// populated in the context using workflow.WithBootstrapCluster.
+// KubernetesClient is used to interact with a Kubernetes cluster.
+type KubernetesClient interface {
+	Get(ctx context.Context, kubeconfig string, key corev1.ObjectReference, obj runtime.Object) error
```
g-gaston, line 74: Although it makes sense from a concept perspective, I fear `ObjectReference` is not the best input for this kind of method. It contains things like UID, ResourceVersion and even FieldPath. We would need to document that those fields are not needed/ignored but still, it complicates the interface.

I would suggest to either have our own type with the minimum set of fields or maybe use a signature similar to https://github.com/aws/eks-anywhere/blob/main/pkg/clients/kubernetes/unauth.go#L41, which only requires name and namespace.

pkg/workflow/task/bootstrap/client.go
```diff
@@ -0,0 +1,46 @@
...
+	// GetKubeconfig retrieves an in-memory kubeconfig for the bootstrap cluster identified by
+	// name.
+	GetKubeconfig(ctx context.Context, name string) ([]byte, error)
+
+	// ClusterExists determines if a bootstrap cluster identified by name exists.
+	ClusterExists(ctx context.Context, name string) (bool, error)
+}
+
+// CreateBootstrapClusterOptions defines the options available for bootstrap cluster creation.
+type CreateBootstrapClusterOptions struct {
```
g-gaston, line 32: Curious why you would define this here as opposed to in `pkg/bootstrap` or similar.

This creates dependencies "retriever" -> `pkg/workflow/task/bootstrap` and "client" -> `pkg/workflow/task/bootstrap` and I'm not sure that's a good thing.

Although there is no other usecase for the bootstrap client/retriever than to be used from these tasks, it seems a bit odd than they depend on this package, which is just an orchestrator.

pkg/workflow/task/bootstrap/bootstrap.go
chrisdoherty4: Yea, it makes sense to descope it. I think there's a `NamespacedName` somewhere in apimachinery or some other Kubernetes lib?

pkg/workflow/task/bootstrap/client.go
chrisdoherty4: I'm not sure what we want yet. With the other changes in this PR, the `bootstrap.Bootstrapper` is redundant and disappears. If we put the options struct in `boostrapper` it only defines that type making that package feel rather unnecessary.

This would be a tradeoff that yields colocated code at the expense of an odd looking import statement in things populating the options.

chrisdoherty4: Another thought. It might be worth simplifying the options first to extract as much provider agnostic configuration out of providers and see whats left.

chrisdoherty4:
We've moved a different direction. We may revisit some of this in the near future.

PR #3471, opened 2022-09-28
Add vsphere client processors

taneyland:
*Issue #, if available:*

*Description of changes:*
Add client processors for vsphere reconciler

*Testing (if applicable):*


controllers/cluster_controller_test.go
```diff
@@ -289,6 +289,13 @@ func createWNMachineConfig() *anywherev1.VSphereMachineConfig {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name + "-wn",
 			Namespace: namespace,
+			OwnerReferences: []metav1.OwnerReference{
```
g-gaston, line 292: what are these for?

g-gaston (commented):
Looks great!

controllers/cluster_controller_test.go
taneyland: Since we [ensure owner reference](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L153) is set in the reconciler, this field is expected so I'm setting it here

g-gaston: But why are these tests being affected by adding processors to the spec build?

taneyland: We call the [DefaultConfigClientBuilder](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L203) in the method where we set the owner refs. And we added the new processors to the DefaultConfigClientBuilder. So it now adds the owner ref to those objects, which in turn affects the unit tests because it we expect the objects in the test to match what is actually passed in

g-gaston: Ahhhh understood, thanks @taneyland!

PR #3472, opened 2022-09-28
Remove contextutil package

chrisdoherty4:
The `contextutil` package was written as an experiment to contain context instrumentation behavior. Having refactored the code to have task specific packages it doesn't make sense to maintain for now.

This change moves the bootstrap cluster context instrumentation funcs to the `bootstrap` task package and removes `contextutil`.


pkg/workflow/task/bootstrap/context.go
```diff
@@ -0,0 +1,27 @@
...
+const clusterKey contextKey = "bootstrap-cluster"
+
+// WithCluster returns a context with a bootstrap cluster set.
+func WithCluster(ctx context.Context, cluster *types.Cluster) context.Context {
+	return context.WithValue(ctx, clusterKey, cluster)
+}
+
+// FromContext retrieves the bootstrap cluster from the context that was previously set
+// with WithCluster. It returns nil if no cluster was configured in ctx.
+func FromContext(ctx context.Context) *types.Cluster {
```
g-gaston, line 25: Does that mean other task packages need to import `bootstrap`?
Interesting, what do gain by doing that?

chrisdoherty4: I'm not sure we "gain" anything. Its more about clarity. `contextutil.BootstrapCluster` vs `bootstrap.FromContext`. 

Admittedly I'm still somewhat split having now implemented a hook that needed the spec which I played with retrieving from the context. There was no logical place to have a `SpecFromContext` so I, for now, created a `clusterspec` pkg but it seems excessive for context instrumentation and will likely promote an antipattern.

chrisdoherty4: I've toyed more and I think this is a bad idea. I'm going to close for now.

PR #3474, opened 2022-09-28
Fixing unintended rollingupgrade from older clusters related to diskoffering

maxdrib:
*Issue #, if available:*

*Description of changes:*
As described in the comment below, clusters created with older versions of eks-a will have an empty diskoffering included on the CloudStackMachineConfig, like

```
  diskOffering:
    mountPath: ""
    filesystem: ""
    label: ""
    device: ""
```

However, we have since made this field into a pointer and optional, so new versions of eks-a expect an unset diskoffering to have 
```
  diskOffering: nil
```

This resulted in unintended rollingupgrades taking place where the eks-a controller would think there's a [difference between the diskofferings in old and new](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/cloudstack.go#L608), and create a new CloudStackMachineTemplate. This change serves to align the two definitions to make them equivalent and prevent the rollingupgrade from taking place.

*Testing (if applicable):*
Unit test added, and e2e test succeeds where cluster is created with EKS-A 11.1 and empty diskoffering, then upgraded to EKS-A 11.3, still with empty diskoffering.

### DIFF
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_types_test.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_types_test.go
@@ -32,7 +32,7 @@ func TestCloudStackMachineConfigDiskOfferingEqual(t *testing.T) {
 	g.Expect(diskOffering1.Equal(diskOffering2)).To(BeTrue())
 }
 
-func TestCloudStackMachineConfigEmptyDiskOfferingEqual(t *testing.T) {
+func TestCloudStackMachineConfigNilDiskOfferingEqual(t *testing.T) {
 	var nilDiskOffering *v1alpha1.CloudStackResourceDiskOffering
 	emptyDiskOffering := &v1alpha1.CloudStackResourceDiskOffering{
 		MountPath:  "",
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/cloudstackmachineconfig_types_test.go
```diff
@@ -32,6 +32,18 @@ func TestCloudStackMachineConfigDiskOfferingEqual(t *testing.T) {
 	g.Expect(diskOffering1.Equal(diskOffering2)).To(BeTrue())
 }
 
+func TestCloudStackMachineConfigEmptyDiskOfferingEqual(t *testing.T) {
+	var nilDiskOffering v1alpha1.CloudStackResourceDiskOffering
```
wongni, line 36:
```suggestion
	var nilDiskOffering *v1alpha1.CloudStackResourceDiskOffering
```
I think * is needed to make this variable `nil`

wongni (changes requested):

wongni (approved):

jiayiwang7 (approved):

PR #3475, opened 2022-09-28
Split E2E tests by provider

abhay-krishna:
Currently we run all E2E tests in a single job but this PR refactors them to be run in separate jobs per provider, to provide quicker feedback and separation of concerns/ownership.

/hold


abhay-krishna:
This PR organizes the tests into `flavors`, which represent usecases of the CLI and contain reusable methods for different providers, and `providers`, which represent the various infrastructure providers that we support and run tests for. The providers folder contain individual tests that are run in the pipeline, these test methods use the methods exposed by the flavors packages.

test/e2e/flavors/awsiamauth/awsiamauth.go
```diff
@@ -0,0 +1,39 @@
+//go:build e2e
```
abhay-krishna, line 1: Moved over from `test/e2e/awsiam_test.go`.

test/e2e/flavors/curatedpackages/adot.go
```diff
@@ -0,0 +1,21 @@
+//go:build e2e
```
abhay-krishna, line 1: Moved over from `test/e2e/adot_test.go`

test/e2e/flavors/curatedpackages/curatedpackages.go
```diff
@@ -0,0 +1,54 @@
+//go:build e2e
```
abhay-krishna, line 1: Moved over from `test/e2e/curated_packages_test.go`.

g-gaston (commented):
Could you elaborate on why we are separating the tests in different packages? It doesn't seem like it's necessary to achieve the goal of this project, which was to run the tests separately by provider.

Doing that has other implications, like like all the `flavors` packages, that are only needed because they need to be imported by the different provider packages. In other words, it adds complexity. If we want to do that, we should justify why we need that complexity.

abhay-krishna:
> Could you elaborate on why we are separating the tests in different packages? It doesn't seem like it's necessary to achieve the goal of this project, which was to run the tests separately by provider.
> 
> Doing that has other implications, like like all the `flavors` packages, that are only needed because they need to be imported by the different provider packages. In other words, it adds complexity. If we want to do that, we should justify why we need that complexity.

The separation was done for better organization of the e2e folder. Earlier there was no specific place where we were putting tests, so finding tests or making a change to specific tests was more effort. Now we can easily identify where the tests go since they're scoped down by provider/usecase.

abhay-krishna:
@g-gaston  If you feel like the separate packages is out of scope and adds too much complexity, I can bring it back to a common folder.

TerryHowe:
Needs a rebase. Lots of merge conflicts.

abhay-krishna:
@g-gaston The e2e folder is now grouped based on providers as
```bash
test/e2e
├── cloudstack_test.go
├── constants.go
├── docker_test.go
├── nutanix_test.go
├── snow_test.go
├── tinkerbell_test.go
└── vsphere_test.go
```

g-gaston (commented):
* Do we still need the `flavors` package? I'm not against splitting the functionality into different packages. However, simply moving methods from one folder to separate ones doesn't necessarily give coherence to that package. Example: `taints.RunTaintsUpgradeFlow`. Also, if we want to do that, we should probably better define the difference between `flavors` and `framework`: what goes where? I would vote to keep everything under the e2e package for now and approach this package split as a separate piece of work.
* Should we split the provider tests into different files? Those single provider files can become quite big (if they are not already) over time. I'm ok doing this after merging this PR since it doesn't affect the test infra and how tests are run.

abhay-krishna:
> * Do we still need the `flavors` package? I'm not against splitting the functionality into different packages. However, simply moving methods from one folder to separate ones doesn't necessarily give coherence to that package. Example: `taints.RunTaintsUpgradeFlow`. Also, if we want to do that, we should probably better define the difference between `flavors` and `framework`: what goes where? I would vote to keep everything under the e2e package for now and approach this package split as a separate piece of work.
> * Should we split the provider tests into different files? Those single provider files can become quite big (if they are not already) over time. I'm ok doing this after merging this PR since it doesn't affect the test infra and how tests are run.

* Fair point, I will revert the changes to the flavors package and have them remain in the e2e folder.
* I want to split the provider files in a separate PR. Condensing the test files into one file per provider naturally makes each one large, proportional to the amount of test coverage we've allocated to each provider. In the future, we can have files for each provider pertaining to the usecase like `vsphere_curated_packages_test.go` or `vsphere_oidc_test.go`, which should be an easy change.

cmd/integration_test/build/buildspecs/build-eks-a-cli.yml
```diff
@@ -4,7 +4,23 @@ phases:
   build:
     commands:
     - make e2e
+    - |
+      for provider_test_file in test/e2e/*_test.go; do
+        provider=$(echo $(basename $provider_test_file) | cut -d_ -f1)
+        make e2e-tests-binary E2E_TAGS="e2e $provider" E2E_OUTPUT_FILE=bin/$provider/e2e.test
+      done
```
g-gaston, line 11: nit: move to script? (can be a follow up)

cmd/integration_test/build/buildspecs/cloudstack-test-eks-a-cli.yml
```diff
@@ -0,0 +1,86 @@
...
+  build:
+    commands:
+      - export JOB_ID=$CODEBUILD_BUILD_ID
+      - BUNDLES_OVERRIDE=false
+      - |
+        if [ -f ./bin/local-bundle-release.yaml ]; then
+          BUNDLES_OVERRIDE=true
+        fi
+      - SKIPPED_TESTS=$(yq e ".skipped_tests | @csv" ${CODEBUILD_SRC_DIR}/test/e2e/SKIPPED_TESTS.yaml)
+      - mv bin/cloudstack/e2e.test bin/e2e.test
```
g-gaston, line 66: nit: move this to scipt/make? and we can reuse it in all provider buildspecs

test/e2e/flavors/curatedpackages/utils.go
```diff
@@ -0,0 +1,200 @@
+//go:build e2e
+// +build e2e
+
+package curatedpackages
```
g-gaston, line 4: seems like this is a leftover? seem like a duplicate with test/e2e/curatedpackages.go

Makefile
```diff
@@ -589,7 +593,6 @@ e2e-cross-platform-%:
 
 .PHONY: e2e
 e2e: eks-a-e2e integration-test-binary ## Build integration tests
-	$(MAKE) e2e-tests-binary E2E_TAGS=e2e
```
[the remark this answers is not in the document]
abhay-krishna: Done. I will save the refactoring or renaming targets to a future PR, but I have added the option to use the `all_providers` tag to compile the test binary with all the tests allowing developers to use the same binary for all providers when testing locally.

test/e2e/flavors/curatedpackages/utils.go
abhay-krishna: Good catch, removed it.

cmd/integration_test/build/buildspecs/build-eks-a-cli.yml
abhay-krishna: I'll do this in a follow-up PR along with other cleanups.

cmd/integration_test/build/buildspecs/cloudstack-test-eks-a-cli.yml
abhay-krishna: Noted, I shall do this in a follow-up PR along with other cleanups.

g-gaston (approved):
Awesome!

jaxesn (approved):

PR #3476, opened 2022-09-28
[1/2] Add AWS IAM Authenticator hooks

chrisdoherty4:
### Description

This change consolidates AWS IAM Authentication installation to the `awsiamauth` package enabling new workflow hooks to be written and registered with the new create management cluster workflow.

The existing workflow code paths are honored as they call the same APIs created for new workflow hooks.

The AWS IAM Auth implementation hasn't been touched so should be functionally equivalent. A separate PR will address exported APIs and unit tests against the existing code in the `awsiamauth` package.

### Testing

The `local-e2e` make target was run with a Docker AWS IAM Auth test. Additional manual verification was performed by using the AWS Kubeconfig to retrieve pods from the Docker cluster.


pkg/awsiamauth/hook.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/workflow"
+	"github.com/aws/eks-anywhere/pkg/workflow/management"
+	"github.com/aws/eks-anywhere/pkg/workflow/workflowcontext"
+)
+
+type HookRegistrar struct {
+	*AwsIamAuth
+	Spec *cluster.Spec
```
g-gaston, line 15: For structs with tons of members I can buy that it makes the code a bit cleaner, but for this I would vote to have a constructor and make the members private.

It makes clear what is mandatory and what's not. Also, it stops anyone from tinkering with the inner members directly once the struct is constructed.

pkg/awsiamauth/hook.go
```diff
@@ -0,0 +1,51 @@
+package awsiamauth
```
g-gaston, line 1: What do you think about moving this to `pkg/awsiamauth/hook/`?
It avoids importing the workload packages from the main business logic packages.

It's not necessary with this one in particular but I feel it sets a good precedence for the future.

pkg/dependencies/factory.go
```diff
@@ -667,7 +667,7 @@ func (f *Factory) WithAwsIamAuth() *Factory {
 		}
 		certgen := crypto.NewCertificateGenerator()
 		clusterId := uuid.New()
-		f.dependencies.AwsIamAuth = awsiamauth.NewAwsIamAuth(certgen, clusterId)
```
g-gaston, line 670: I think we need to add this at the top of `WithAwsIamAuth`

```go
f.WithKubectl().WithWriter()
```

g-gaston (commented):
This is great

pkg/awsiamauth/hook.go
chrisdoherty4: I don't think its necessary in this case. Equally, I could see it being appropriate in other cases.

In general, I think its good to stick to the rule "package A provides everything to do with A" and only break it down if needed.

pkg/awsiamauth/hook.go
chrisdoherty4: For `Spec` in particular I'm not sure why I did it this way - will adjust. 

The `AwsIamAuth` is embedded so will have to remain exposed. That's done intentionally for convenience so consumers can use the registrar as anything satisfying existing AWS IAM interfaces.

chrisdoherty4: Done

pkg/dependencies/factory.go
chrisdoherty4: Done

pkg/awsiamauth/hook.go
chrisdoherty4: We agreed on my response above.

PR #3487, opened 2022-09-28
Fixing nil pointer error

jonahjon:

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/release/pkg/operations/upload.go
+++ b/release/pkg/operations/upload.go
@@ -80,12 +80,11 @@ func UploadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]re
 				}
 			}
 
-			// If the artifact is a helm chart, skip the skopeo copy as it's handled separately.
-			if artifact.Image.AssetName != "" && strings.HasSuffix(artifact.Image.AssetName, "helm") {
-				continue
-			}
-
 			if artifact.Image != nil {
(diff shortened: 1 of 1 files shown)


release/pkg/operations/upload.go
```diff
@@ -81,7 +81,7 @@ func UploadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]re
 			}
 
 			// If the artifact is a helm chart, skip the skopeo copy as it's handled separately.
-			if strings.HasSuffix(artifact.Image.AssetName, "helm") {
+			if artifact.Image.AssetName != "" && strings.HasSuffix(artifact.Image.AssetName, "helm") {
```
abhay-krishna, line 84: Can you move this into the if block on line 88?

abhay-krishna: And you can remove the empty string check.

TerryHowe: Is AssetName nil, because I think len(Assetname) == 0 is true for both "" and nil

jonahjon: updated!

jonahjon: I can remove the empty string check, but lemme doublecheck that works still.

abhay-krishna (approved):
/lgtm
/approve

PR #3508, opened 2022-09-29
Revert "Fixing nil pointer error"

jonahjon:
Reverts aws/eks-anywhere#3487

### DIFF
--- a/release/pkg/operations/upload.go
+++ b/release/pkg/operations/upload.go
@@ -80,11 +80,12 @@ func UploadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]re
 				}
 			}
 
+			// If the artifact is a helm chart, skip the skopeo copy as it's handled separately.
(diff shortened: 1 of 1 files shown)


release/pkg/operations/upload.go
```diff
@@ -80,11 +80,12 @@ func UploadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]re
...
+			// If the artifact is a helm chart, skip the skopeo copy as it's handled separately.
+			if strings.HasSuffix(artifact.Image.AssetName, "helm") {
+				continue
+			}
+
 			if artifact.Image != nil {
-				// If the artifact is a helm chart, skip the skopeo copy as it's handled separately.
-				if strings.HasSuffix(artifact.Image.AssetName, "helm") {
-					continue
-				}
```
TerryHowe, line 87: I'm struggling to see how this makes things better. Just the opposite it looks like.

TerryHowe (approved):
/lgtm

I assume you are reverting this, so you can revert the other

PR #3510, opened 2022-09-29
feat(CLI): error getting a bundle without providing a name

ewollesen:
Previously, this didn't error out, which led to other processes failing, and often with much harder to debug error messages.

The most likely reason this would be called with an empty bundle name is the case where a packages controller has come online, but has not yet activated a bundle. In that case, it's Spec.ActiveBundle value will be the empty string.

More quick PRs in this series will follow. Be on the lookout!


TerryHowe (approved):
/lgtm

PR #3513, opened 2022-09-29
refactor(CLI): GetResource => HasResource for clarity

ewollesen:
The functionality of GetResource is more accurately defined as HasResource,
since the resource itself was never returned to the caller anyway. It is also
refactored to use the existing get method that it had duplicated, almost
entirely.

Depends on #3511


TerryHowe (approved):

pkg/executables/kubectl.go
```diff
@@ -1627,17 +1628,17 @@ func (k *Kubectl) ApplyTolerationsFromTaints(ctx context.Context, oldTaints []co
 }
 
 func (k *Kubectl) KubeconfigSecretAvailable(ctx context.Context, kubeconfig string, clusterName string, namespace string) (bool, error) {
-	return k.GetResource(ctx, "secret", fmt.Sprintf("%s-kubeconfig", clusterName), kubeconfig, namespace)
+	return k.HasResource(ctx, "secret", fmt.Sprintf("%s-kubeconfig", clusterName), kubeconfig, namespace)
 }
 
-func (k *Kubectl) GetResource(ctx context.Context, resourceType string, name string, kubeconfig string, namespace string) (bool, error) {
```
a-cool-train, line 1633: I thought this method was used in other places apart from curated packages. Can you check on this if possible?

ewollesen: Yes, it was, I searched for and found them and have adapted each of them in this PR.

If you know of other places it was used, please let me know, but the code compiles, and runs, and the tests run, and my search for any instances of GetResources turns up nothing, so I don't have any reason to believe I've missed any.

PR #3514, opened 2022-09-29
feat(CLI): wait for bundle activation at package controller install

ewollesen:
This adds to InstallController to have it wait until a bundle has been
activated, before returning, and thus indicating that the controller is
"Installed". This help prevent subsequent activity (like package installation)
from failing because the controller isn't yet ready.

Depends on #3513

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -5,7 +5,9 @@ import (
 	_ "embed"
 	"fmt"
 	"strings"
+	"time"
 
+	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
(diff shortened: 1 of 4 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -89,12 +101,73 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
+				return
+			}
+
+			if pbc.Spec.ActiveBundle != "" {
+				logger.V(6).Info("found packages bundle controller active bundle",
+					"name", pbc.Spec.ActiveBundle)
+				return
+			}
+
+			logger.V(6).Info("waiting for package bundle controller to activate a bundle",
```
junshun, line 148: Do we want to add a retry counter that way its easier to follow in the log to figure out how much time has passed?

ewollesen: Could... But log messages are timestamped, no?

PR #3521, opened 2022-09-30
Refactor oras

a-cool-train:
*Description of changes:*
1. Refactor Oras to prevent live connection during unit tests
2. Refactor Docker TagImage and PushImage so that there is no logic and the logic is moved to Registry.go
3. Create New Objects in factory instead of import or download images cmd/ package.

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/import_images.go
+++ b/cmd/eksctl-anywhere/cmd/import_images.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"oras.land/oras-go/pkg/content"
 	"path/filepath"
@@ -69,6 +70,7 @@ func (c ImportImagesCommand) Call(ctx context.Context) error {
 	factory := dependencies.NewFactory()
 	deps, err := factory.
 		WithManifestReader().
+		WithBundleReader(c.includePackages).
 		Build(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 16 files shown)


pkg/curatedpackages/oras/pusher.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/logger"
+)
+
+type Pusher struct {
+	registry *content.Registry
+	store    *content.Memory
+}
+
+func NewPusher(registry *content.Registry, store *content.Memory) *Pusher {
```
chrisdoherty4, line 18: I noticed we're injecting structs instead of interfaces. In the case of the `content.Memory` type it doesn't have any options and we inject a raw `content.Memory` from `import_images.go` so could this just be constructed in-line here? 

The registry does take options but we don't configure them in any capacity. Do we reasonably expect to need to customize the registry at all?

cmd/eksctl-anywhere/cmd/import_images.go
```diff
@@ -135,15 +144,22 @@ func (c ImportImagesCommand) Call(ctx context.Context) error {
...
+		registry, err := content.NewRegistry(content.RegistryOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("creating registry: %w", err)
+		}
+		memoryStore := content.NewMemory()
+		bundlePusher := oras.NewPusher(registry, memoryStore)
+		return oras.NewFileRegistryImporter(registryEndpoint, username, password, artifactsFolder, bundlePusher), nil
 	}
-	return &artifacts.Noop{}
+	return &artifacts.Noop{}, nil
```
chrisdoherty4, line 164: Can we invert this logic? It normally reads better and helps focus on the important bits when you keep the 'core logic' as unindented as possible (this relates to 'returning early instead of nesting deeply').

```go
	if !includePackages {
		return &artifacts.Noop{}, nil
	}

	registry, err := content.NewRegistry(content.RegistryOptions{})
	if err != nil {
		return nil, fmt.Errorf("creating registry: %w", err)
	}
	memoryStore := content.NewMemory()
	bundlePusher := oras.NewPusher(registry, memoryStore)
	return oras.NewFileRegistryImporter(registryEndpoint, username, password, artifactsFolder, bundlePusher), nil
```

chrisdoherty4 (changes requested):
There's a lack of API documentation so its hard to decipher what behaviors satisfy which interfaces, a simple `// Foo satisfies bar.Bar` can be helpful here. The interfaces aren't documented so behavioral expectations are a little ambiguous. Some of the APIs have ambiguous parameters also making it had to appreciate what to pass in. In some instances, the parameters are abbreviations and it'd be useful to not use abbreviated terms (e.g. `art`).

Please could we document behavioral expectations and the purpose of structs etc.

TerryHowe:
Reopen if anyone thinks this should be revived.

PR #3525, opened 2022-09-30
Make audit policy apiVersion dynamic k8s version

g-gaston:
*Description of changes:*
This should avoid unnecessary cp rolling upgrades in workload clusters when updating the management cluster.

*Testing (if applicable):*


jiayiwang7 (approved):

pkg/providers/common/auditpolicy.go
```diff
@@ -0,0 +1,252 @@
...
+	return yaml.Marshal(GetAuditPolicy(kubeVersion))
+}
+
+func GetAuditPolicy(kubeVersion anywherev1.KubernetesVersion) *auditv1.Policy {
+	a := baseAuditPolicy()
+	if kubeVersion == anywherev1.Kube124 {
+		a.APIVersion = "audit.k8s.io/v1"
+	} else {
+		a.APIVersion = "audit.k8s.io/v1beta1"
+	}
```
maxdrib, line 20: if we'll be using v1 as the version going forward it seems like we should default to v1 instead of defaulting to v1beta1
```suggestion
	if kubeVersion in [anywherev1.Kube120, anywherev1.Kube121, anywherev1.Kube122, ..., anywherev1.Kube123] {
		a.APIVersion = "audit.k8s.io/v1beta1"
	} else {
		a.APIVersion = "audit.k8s.io/v1"
	}
```

pkg/providers/common/common.go
```diff
@@ -26,10 +26,6 @@ const (
 	publicKeyFileName  = "eks-a-id_rsa.pub"
 )
 
-func GetAuditPolicy() string {
```
maxdrib, line 29: we should remove config/audit-policy.yaml right?

pkg/providers/common/auditpolicy.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return a
+}
+
+func baseAuditPolicy() *auditv1.Policy {
```
maxdrib, line 25: when this gets serialized, do we have any way to validate that it won't have any deltas compared to what was being generated with the yaml spec? Otherwise we'll still have a rolling upgrade, but just due to serialization differences and the way the yaml is being generated.

pkg/providers/common/auditpolicy.go
```diff
@@ -0,0 +1,252 @@
...
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+func GetAuditPolicyYaml(kubeVersion anywherev1.KubernetesVersion) ([]byte, error) {
+	return yaml.Marshal(GetAuditPolicy(kubeVersion))
+}
+
+func GetAuditPolicy(kubeVersion anywherev1.KubernetesVersion) *auditv1.Policy {
+	a := baseAuditPolicy()
+	if kubeVersion == anywherev1.Kube124 {
```
abhay-krishna, line 16: Since v1beta1 audit API version has been removed in 1.24, can we generalize this to anything >= 1.24? Maybe using a custom comparator on KubeVersion?

maxdrib:
I made a few changes from unit testing to more closely align this policy with what the CloudStack provider (at least) had before, but there were still some minor differences due to serialization of empty objects, for example. However, the KCP object is very touchy and if anything gets changed in it, a rolling upgrade will be triggered. In the end, when I tried upgrading, I still observed the rolling upgrade on the workload cluster(s), so I do not believe this PR achieves its goal of preventing the rollingupgrade in the workload clusters.

pkg/providers/common/common.go
mitalipaygude: yes

g-gaston:
@mitalipaygude already addressed this in a different PR

PR #3528, opened 2022-09-30
attempt fix for devrelease helm move

jonahjon:

Currently if we are doing a dev release we skip helm chart.yaml fix. 

We should only do the packages chart upload skip if it's not dev release too, otherwise it will fail on dev releases.


abhay-krishna (approved):
/lgtm
/approve

Issue #3523, opened 2022-09-30
Improve EKS-A controller logging to understand cause of rollingupgrade


maxdrib:
**What would you like to be added**:
Throughout development of the cloudstack provider, we have observed countless RollingUpgrades happen when the EKS-A controller is unpaused, and had no clear way to understand what triggered this. Most often, there is some equality check which is not met and results in the unintended creation of a new CloudStackMachineTemplate. However, identify the failing equality check is extremely tedious because there is no logger available where these checks happen ([AnyImmutableFieldChanged](https://github.com/aws/eks-anywhere/blob/246dd770e40ac5a310213c444149c0eff470463b/pkg/providers/cloudstack/cloudstack.go#L582)).

Troubleshooting these issues requires either dumping the CAPI resources before/after the EKS-A controller is unpaused, or building a custom controller with `fmt.Println` commands scattered throughout this section of code. 

**Why is this needed**:
Troubleshooting unintended rollingupgrades from the EKS-A controller is extremely tedious and additional logging would help save time when developing and troubleshooting.

maxdrib:
I ran some experiments with a custom controller build using logger.Info() in the code linked above but it was not logging to the controller logs. So far, fmt.Println() is the only way I've found to log, other than plumbing the logger in to this function all the way from the controller.

PR #3529, opened 2022-09-30
Plumbing external loggers into cloudstack provider for improved debugging

maxdrib:
*Issue #, if available:*
resolves https://github.com/aws/eks-anywhere/issues/3523

*Description of changes:*
Recently with the old EKS-A controller, there have been many occasions where a RollingUpgrade is triggered and it is unclear why. Usually a new CloudStackMachineTemplate name was generated because some equality condition was not met. This PR plumbs the controller's logger into the code performing the equality checks to clearly indicate which objects are not equal, at least for the provider-specific components.

*Testing (if applicable):*
Unit tests pass. 

In existing cluster with custom EKS-A controller build, controller logs properly when an object is changed:
```
1.6645658719716415e+09	INFO	controllers.Cluster	Old and new CloudStackMachineConfig UserCustomDetails does not match	{"machineConfig": "", "oldUserCustomDetails": {"foo":"bar"}, "newUserCustomDetails": {"foo":"bar2"}}
```

CLI properly logs during cluster creation:
```
2022-09-30T15:29:24.934-0400	V0	Installing secrets on bootstrap cluster
```


controllers/resource/template.go
```diff
@@ -222,9 +225,10 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster
 
 func (r *CloudStackTemplate) getControlPlaneTemplateName(ctx context.Context, eksaCluster *anywherev1.Cluster, oldCsdc *anywherev1.CloudStackDatacenterConfig, csdc anywherev1.CloudStackDatacenterConfig, oldCpCsmc *anywherev1.CloudStackMachineConfig, cpCsmc anywherev1.CloudStackMachineConfig, clusterName string) (string, error) {
 	var controlPlaneTemplateName string
-	updateControlPlaneTemplate := cloudstack.AnyImmutableFieldChanged(oldCsdc, &csdc, oldCpCsmc, &cpCsmc)
+	updateControlPlaneTemplate := cloudstack.AnyImmutableFieldChanged(oldCsdc, &csdc, oldCpCsmc, &cpCsmc, r.log)
```
vivek-koppuru, line 228: I didn't realize this wasn't a member method, which if it was we could pass in the logger in the constructor instead of here like the other methods.

vivek-koppuru (approved):

controllers/resource/template.go
```diff
@@ -222,9 +225,10 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster
...
 	if updateControlPlaneTemplate {
 		controlPlaneTemplateName = common.CPMachineTemplateName(clusterName, r.now)
+		r.log.Info(fmt.Sprintf("control plane machine template updated to new name %s", controlPlaneTemplateName))
```
g-gaston, line 231:
```suggestion
		r.log.Info("Control plane machine template updated", "new name", controlPlaneTemplateName)
```

controllers/resource/template.go
```diff
@@ -253,6 +257,7 @@ func (r *CloudStackTemplate) getEtcdTemplateName(ctx context.Context, eksaCluste
 				return "", err
 			}
 			etcdTemplateName = common.EtcdMachineTemplateName(clusterName, r.now)
+			r.log.Info(fmt.Sprintf("etcd template updated to new name %s", etcdTemplateName))
```
g-gaston, line 260:
```suggestion
			r.log.Info("Etcd template updated", "new name", etcdTemplateName)
```

Same comment for other logs using string interpolation :)

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -255,6 +257,7 @@ func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, machineC
 			etcdMachineSpec:             etcdMachineSpec,
 			now:                         now,
 		},
+		log:       logger.Get(),
```
g-gaston, line 260: This is great. Could we inject this as an argument in the constructor?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -590,32 +595,48 @@ func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterCon
...
 			oldAz.Domain == newAz.Domain) {
 
+			log.Info("Old and new CloudStackDatacenterConfigs do not match", "oldAvailabilityZone", oldAz,
+				"newAvailabilityZone", newAz)
 			return true
 		}
 	}
 
 	if !oldCsmc.Spec.Template.Equal(&newCsmc.Spec.Template) {
+		log.Info("Old and new CloudStackMachineConfig Templates do not match", "machineConfig", oldCsmc.Name, "oldTemplate", oldCsmc.Spec.Template,
```
g-gaston, line 605:
```suggestion
		log.V(4).Info("Old and new CloudStackMachineConfig Templates do not match", "machineConfig", oldCsmc.Name, "oldTemplate", oldCsmc.Spec.Template,
```

Same for the other new logs similar to this one
For reference: https://github.com/aws/eks-anywhere/blob/main/pkg/logger/doc.go#L17

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -574,8 +577,10 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
 // AnyImmutableFieldChanged Used by EKS-A controller and CLI upgrade workflow to compare generated CSDC/CSMC's from
 // CAPC resources in fetcher.go with those already on the cluster when deciding whether or not to generate and apply
 // new CloudStackMachineTemplates
-func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig) bool {
+func AnyImmutableFieldChanged(oldCsdc, newCsdc *v1alpha1.CloudStackDatacenterConfig, oldCsmc, newCsmc *v1alpha1.CloudStackMachineConfig, log logr.Logger) bool {
```
g-gaston, line 580: When looking at the code, passing the logger around just to get here and log any possible difference, it makes me think this might be information we want to return from this method. Instead of being just a boolean, return something like `EksaCloudstackDiff` and then let the caller decide what to do with that info (log it, pass it around, transform it, ignore it...).

No need to address as part of this PR since I understand we want to have this logs as soon as possible. But we should probably move into that direction.

controllers/resource/template.go
maxdrib: Will fix the logging formatting

pkg/providers/cloudstack/cloudstack.go
maxdrib: I created an experiment for testing this logging configuration. I added a couple of test logs in this function:

```
	log.V(4).Info("[DEBUG] Testing logging capabilities (v4)")
	log.Info("[DEBUG] Testing logging capabilities (no verbosity)")
```

and then replaced the eks-a controller in my cluster. I observed the no verbosity logs many times, but didn't see any instances of V4 logs

```
1.664898488914078e+09	INFO	controllers.Cluster	[DEBUG] Testing logging capabilities (no verbosity)
```

pkg/providers/cloudstack/cloudstack.go
jiayiwang7: ideally we want to generate the logger instance in factory as dependency setup and inject it to create the provider instance. we do not have that in any of the providers right now. I think it's ok, we can follow up w a separate PR for the change.

jiayiwang7 (approved):

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -1664,7 +1665,7 @@ func TestAnyImmutableFieldChangedDiskOfferingNoChange(t *testing.T) {
 	newDcConfig := givenDatacenterConfig(t, testClusterConfigMainFilename)
 	newMachineConfigsMap := givenMachineConfigs(t, testClusterConfigMainFilename)
 
-	assert.False(t, AnyImmutableFieldChanged(dcConfig, newDcConfig, machineConfigsMap["test"], newMachineConfigsMap["test"]), "Should not have any immutable fields changes")
+	assert.False(t, AnyImmutableFieldChanged(dcConfig, newDcConfig, machineConfigsMap["test"], newMachineConfigsMap["test"], logger.Get()), "Should not have any immutable fields changes")
```
g-gaston, line 1668: You can use:
```suggestion
	assert.False(t, AnyImmutableFieldChanged(dcConfig, newDcConfig, machineConfigsMap["test"], newMachineConfigsMap["test"], test.NewNullLogger()), "Should not have any immutable fields changes")
```

pkg/providers/cloudstack/cloudstack.go
g-gaston: We can call `logger.Get()` in the factory, that's not an issue. If the need to use custom loggers in the factory come up, we just need to update the factory.

IMO we should try to inject all dependencies in the constructor. We haven't always done that in the past, specially with the logger. But since you are already doing the hardest part (change all log entries to use the field logger), I would vote to inject it from the constructor and make it complete.

https://github.com/aws/eks-anywhere/blob/main/docs/developer/best-practice.md#-dependencies
> Avoid constructing dependencies in-line. If you do construct a dependency in-line because it compliments the most common usage, provide a way to override it such as with a setter.

https://github.com/aws/eks-anywhere/blob/main/docs/developer/best-practice.md#loggers

pkg/providers/cloudstack/cloudstack.go
g-gaston: That's probably because the logger in the controller is not initialized to output v4 logs
We can change that if we want these logs in prod by default or just ask whoever is trying to debug an issue to enable them in their cluster.

maxdrib: Fixed

pkg/providers/cloudstack/cloudstack_test.go
maxdrib: done

g-gaston (approved):

PR #3538, opened 2022-10-03
Add provider CAPI control plane

g-gaston:
*Description of changes:*

This allows to represent a provider specific CAPI control plane for both stacked and unstacked etcd. It implements logic to generate new names for immutable objects when these change. It does this by comparing the new spec to the current state of the objects in the cluster.

This is part of a bigger effort, allowing a transition from yaml go templates to go structs. For context: https://github.com/aws/eks-anywhere/pull/3432

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/clusterapi/controlplane.go
+++ b/pkg/clusterapi/controlplane.go
@@ -1,0 +1,142 @@
+package clusterapi
+
+import (
+	"context"
+
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+)
+
+// Object represents a kubernetes API object
+type Object interface {
+	kubernetes.Object
+}
+
+// ObjectComparator returns true only if only both kubernetes Object's are identical
+// Most of the time, this only requires comparing the Spec field, but that can variate
+// from object to object
+type ObjectComparator[O Object] func(current, new O) bool
+
+// ObjectRetriever gets a kubernetes API object using the provided client
+// If the object doesn't exist, it returns a NotFound error
+type ObjectRetriever[O Object] func(ctx context.Context, client kubernetes.Client, name, namespace string) (O, error)
+
+// ControlPlane represents the provider-specific spec for a CAPI control plane using the kubeadm CP provider
+type ControlPlane[C, M Object] struct {
+	Cluster *clusterv1.Cluster
+
+	// ProviderCluster is the provider-specific resource that holds the details
+	// for provisioning the infrastructure, referenced in Cluster.Spec.InfrastructureRef
+	ProviderCluster C
+
+	KubeadmControlPlane *controlplanev1.KubeadmControlPlane
+
+	// ControlPlaneMachineTemplate is the provider-specific machine template referenced
+	// in KubeadmControlPlane.Spec.MachineTemplate.InfrastructureRef
+	ControlPlaneMachineTemplate M
+
+	EtcdCluster *etcdv1.EtcdadmCluster
+
+	// EtcdMachineTemplate is the provider-specific machine template referenced
+	// in EtcdCluster.Spec.InfrastructureTemplate
+	EtcdMachineTemplate M
+}
+
+// Objects returns all API objects that form a concrete provider-specific control plane
+func (cp *ControlPlane[C, M]) Objects() []kubernetes.Object {
+	objs := make([]kubernetes.Object, 0, 4)
+	objs = append(objs, cp.Cluster, cp.KubeadmControlPlane, cp.ProviderCluster, cp.ControlPlaneMachineTemplate)
+	if cp.EtcdCluster != nil {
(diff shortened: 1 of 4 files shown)


pkg/clusterapi/controlplane_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+func TestControlPlaneUpdateImmutableObjectNamesNoEtcdCluster(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	cp := controlPlaneStackedEtcd()
+	originalCPMachineTemplateName := "my-machine-template-1"
+	cp.ControlPlaneMachineTemplate.Name = originalCPMachineTemplateName
+	cp.KubeadmControlPlane.Spec.MachineTemplate.InfrastructureRef.Name = originalCPMachineTemplateName
+	client := test.NewFakeKubeClient(clientutil.ObjectsToClientObjects(cp.Objects())...)
+	cp.EtcdCluster = etcdCluster()
```
taneyland, line 201: Can you explain what this test is representing? Are we mimicking a case where the etcdCluster was not originally a part of the eksa cluster so it ignores it since we don't allow configuring etcd during upgrade?

pkg/clusterapi/controlplane_test.go
```diff
@@ -0,0 +1,322 @@
...
+	originalName := "my-machine-template-1"
+	mt := dockerMachineTemplate()
+	mt.Name = originalName
+	client := test.NewFakeKubeClient()
+
+	g.Expect(clusterapi.EnsureNewNameIfChanged(ctx, client, dummyRetriever, noChangesCompare, mt)).To(Succeed())
+	g.Expect(mt.Name).To(Equal(originalName))
+}
+
+func TestControlPlaneUpdateImmutableObjectNamesNoKubeadmControlPlane(t *testing.T) {
```
taneyland, line 140: Since we aren't building any objects in the client, this test is basically a create cluster case, right? If KCP doesn't exists, `UpdateImmutableObjectNames` will return nil & everything should default to the default names

pkg/clusterapi/controlplane_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	g.Expect(cp.UpdateImmutableObjectNames(ctx, client, dummyRetriever, noChangesCompare)).To(Succeed())
+	g.Expect(cp.ControlPlaneMachineTemplate.Name).To(Equal(originalCPMachineTemplateName))
+}
+
+func TestControlPlaneUpdateImmutableObjectNamesErrorReadingControlPlane(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	cp := controlPlaneStackedEtcd()
+	client := test.NewFakeKubeClientAlwaysError()
```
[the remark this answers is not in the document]
g-gaston: Unfortunately that wouldn't work. This is testing the action of reading the kubeadmcontrolplane, which is not read using the retriever but using the client directly. The retriever is only used to read the provider machine templates.

pkg/clusterapi/controlplane_test.go
g-gaston: This is dealing with the scenario where an etcdCluster is defined in the CP spec but that etcdCluster doesn't exist yet "in" the api server.

That scenario might it might no be possible depending on preflight validations, but this code makes no assumption about those. The logic followed is: if there is not an EtcdCluster object yet, that means there won't be a name collision, hence we don't need to generate a new name for the machine template.

pkg/clusterapi/controlplane.go
```diff
@@ -0,0 +1,142 @@
...
+// ObjectComparator returns true only if only both kubernetes Object's are identical
+// Most of the time, this only requires comparing the Spec field, but that can variate
+// from object to object
+type ObjectComparator[O Object] func(current, new O) bool
+
+// ObjectRetriever gets a kubernetes API object using the provided client
+// If the object doesn't exist, it returns a NotFound error
+type ObjectRetriever[O Object] func(ctx context.Context, client kubernetes.Client, name, namespace string) (O, error)
+
+// ControlPlane represents the provider-specific spec for a CAPI control plane using the kubeadm CP provider
```
jiayiwang7, line 29: We also have other provider-specific objects consider as CP objects (resourceset, secrets, configmap, etc.). Where will they be if not here?

pkg/clusterapi/controlplane.go
```diff
@@ -0,0 +1,142 @@
...
+	cp.EtcdMachineTemplate.SetName(currentEtcdCluster.Spec.InfrastructureTemplate.Name)
+	if err = EnsureNewNameIfChanged(ctx, client, machineTemplateRetriever, machineTemplateComparator, cp.EtcdMachineTemplate); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// EnsureNewNameIfChanged updates an object's name if such object is different from its current state in the cluster
+func EnsureNewNameIfChanged[M Object](ctx context.Context,
```
jiayiwang7, line 109: we might want move this to a separate file since it can be used in worker node generation

pkg/clusterapi/controlplane.go
g-gaston: Great question! We will compose around this struct, building on top. Example: https://github.com/aws/eks-anywhere/pull/3432/files#diff-44aae5984bb86210e8481b396a2b8270c54ca9f136e264b0095e89ce141f765cR37

pkg/clusterapi/controlplane.go
g-gaston: Fine if I do that when the time comes?

jiayiwang7: fine fine

jiayiwang7 (approved):

PR #3540, opened 2022-10-03
feat(Makefile): adds a couple of small coverage based targets

ewollesen:
New (or modified) Makefile variables:

COVER_PROFILE - has its default renamed from coverage.html to coverage.out, as it's not HTML, but rather a coverage profile in plain-text.

COVER_HTML - is introduced, which defaults to coverage.html, and will contain the HTML output from the coverage-html target.

New Makefile targets:

coverage - is an alias for coverage-unit-test

coverage-html - will generate HTML output you can view in your browser

coverage-view - will open your browser to view the coverage info

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


TerryHowe:
I was wondering what was available there myself, but never looked. This change lgtm, but I should let someone else look at this.

PR #3550, opened 2022-10-04
Move vsphere machine config validations to api

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR moves some vsphere machine config validations to the api level. This is especially important for the full lifecycle api to validate the fields during both create and upgrade through the controller

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/vspheremachineconfig.go
+++ b/pkg/api/v1alpha1/vspheremachineconfig.go
@@ -123,3 +123,20 @@ func setVSphereMachineConfigDefaults(machineConfig *VSphereMachineConfig) {
 		logger.V(1).Info("SSHUsername is not set or is empty for VSphereMachineConfig, using default", "machineConfig", machineConfig.Name, "user", machineConfig.Spec.Users[0].Name)
 	}
 }
+
+func validateVSphereMachineConfig(config *VSphereMachineConfig) error {
+	if len(config.Spec.Datastore) <= 0 {
+		return fmt.Errorf("VSphereMachineConfig %s datastore is not set or is empty", config.Name)
+	}
+	if len(config.Spec.ResourcePool) <= 0 {
+		return fmt.Errorf("VSphereMachineConfig %s VM resourcePool is not set or is empty", config.Name)
+	}
+	if config.Spec.OSFamily != Bottlerocket && config.Spec.OSFamily != Ubuntu && config.Spec.OSFamily != RedHat {
+		return fmt.Errorf("VSphereMachineConfig %s osFamily: %s is not supported, please use one of the following: %s, %s, %s", config.Name, config.Spec.OSFamily, Bottlerocket, Ubuntu, RedHat)
+	}
+	if config.Spec.OSFamily == Bottlerocket && config.Spec.Users[0].Name != bottlerocketDefaultUser {
+		return fmt.Errorf("SSHUsername %s is invalid. Please use 'ec2-user' for Bottlerocket", config.Spec.Users[0].Name)
+	}
+
+	return nil
+}
--- a/pkg/providers/vsphere/validator.go
+++ b/pkg/providers/vsphere/validator.go
@@ -107,22 +107,6 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, vsphereCl
 	if controlPlaneMachineConfig == nil {
 		return fmt.Errorf("cannot find VSphereMachineConfig %v for control plane", vsphereClusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name)
 	}
-	if len(controlPlaneMachineConfig.Spec.Datastore) <= 0 {
-		return errors.New("VSphereMachineConfig datastore for control plane is not set or is empty")
-	}
-	if len(controlPlaneMachineConfig.Spec.ResourcePool) <= 0 {
-		return errors.New("VSphereMachineConfig VM resourcePool for control plane is not set or is empty")
-	}
-	if controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Bottlerocket && controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Ubuntu && controlPlaneMachineConfig.Spec.OSFamily != anywherev1.RedHat {
-		return fmt.Errorf("control plane osFamily: %s is not supported, please use one of the following: %s, %s, %s", controlPlaneMachineConfig.Spec.OSFamily, anywherev1.Bottlerocket, anywherev1.Ubuntu, anywherev1.RedHat)
-	}
-
-	workerNodeGroupConfigs := vsphereClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations
-	for _, workerNodeGroupConfig := range workerNodeGroupConfigs {
-		if workerNodeGroupConfig.Name == "" {
-			return errors.New("must specify name for worker nodes")
-		}
-	}
 
 	var workerNodeGroupMachineConfigs []*anywherev1.VSphereMachineConfig
 	for _, workerNodeGroupConfiguration := range vsphereClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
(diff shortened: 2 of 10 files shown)


pkg/api/v1alpha1/vspheremachineconfig.go
```diff
@@ -76,3 +78,23 @@ func GetVSphereMachineConfigs(fileName string) (map[string]*VSphereMachineConfig
 	}
 	return configs, nil
 }
+
+func validateVSphereMachineConfig(config *VSphereMachineConfig) error {
+	if len(config.Spec.Datastore) <= 0 {
+		return fmt.Errorf("VSphereMachineConfig %s datastore is not set or is empty", config.Name)
+	}
+	if len(config.Spec.Folder) <= 0 {
+		logger.Info("VSphereMachineConfig folder is not set or is empty. Will default to root vSphere folder.")
```
g-gaston, line 87: I'm not sure this log belongs here, could we find a better place? For starters this is talking about some defaulting logic, while this method runs validations

pkg/providers/vsphere/validator.go
```diff
@@ -107,17 +107,8 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, vsphereCl
...
-	}
-	if len(controlPlaneMachineConfig.Spec.Folder) <= 0 {
-		logger.Info("VSphereMachineConfig folder for control plane is not set or is empty. Will default to root vSphere folder.")
-	}
-	if len(controlPlaneMachineConfig.Spec.ResourcePool) <= 0 {
-		return errors.New("VSphereMachineConfig VM resourcePool for control plane is not set or is empty")
-	}
-	if controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Bottlerocket && controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Ubuntu {
-		return fmt.Errorf("control plane osFamily: %s is not supported, please use one of the following: %s, %s", controlPlaneMachineConfig.Spec.OSFamily, anywherev1.Bottlerocket, anywherev1.Ubuntu)
+	if err := controlPlaneMachineConfig.Validate(); err != nil {
```
g-gaston, line 110: Do we need to run this here? Shouldn't this already be run when we validate the `cluster.Config`?

pkg/providers/vsphere/validator.go
```diff
@@ -694,3 +635,31 @@ func (v *Validator) getMissingPrivs(ctx context.Context, vsc govmomi.VSphereClie
 
 	return missingPrivs, nil
 }
+
+func (v *Validator) sameOSFamily(configs map[string]*anywherev1.VSphereMachineConfig) bool {
+	var osFamily anywherev1.OSFamily
+	for _, machineConfig := range configs {
+		if osFamily == "" {
+			osFamily = machineConfig.Spec.OSFamily
```
g-gaston, line 643: Could we do this before the loop by assigning it to the first value?

pkg/providers/vsphere/validator.go
```diff
@@ -694,3 +635,31 @@ func (v *Validator) getMissingPrivs(ctx context.Context, vsc govmomi.VSphereClie
...
+		}
+	}
+	return true
+}
+
+func (v *Validator) sameTemplate(configs map[string]*anywherev1.VSphereMachineConfig) bool {
+	var template string
+	for _, machineConfig := range configs {
+		if template == "" {
+			template = machineConfig.Spec.Template
```
g-gaston, line 657: Same comment here about doing this out the loop

g-gaston (commented):
Looks good!

pkg/api/v1alpha1/vspheremachineconfig.go
taneyland: I'll move it to the method where we set defaults. That will be in a separate PR coming soon

pkg/providers/vsphere/validator.go
taneyland: True, we don't need it here. Removing

pkg/providers/vsphere/validator.go
taneyland: I thought the same thing, but the object that we're traversing through is a map so I don't think there's a good way to access only the first object since we don't know what the first key will be. We could pass in the `OSFamily` of the control plane machine config to use as a source of truth for this, but I think this method would be nice to reuse in our `VSphereReconciler` when we validate the machine configs so I think this is the best way to satisfy both cases

pkg/providers/vsphere/validator.go
```diff
@@ -188,51 +158,12 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, vsphereCl
...
-				return fmt.Errorf("validating SSHUsername for etcd VSphereMachineConfig %v: %v", etcdMachineConfig.Name, err)
-			}
-		}
-	} else {
-		return fmt.Errorf("validating SSHUsername for control plane VSphereMachineConfig %v: %v", controlPlaneMachineConfig.Name, err)
-	}
-
-	for _, machineConfig := range vsphereClusterSpec.machineConfigsLookup {
-		if machineConfig.Namespace != vsphereClusterSpec.Cluster.Namespace {
-			return errors.New("VSphereMachineConfig and Cluster objects must have the same namespace specified")
```
taneyland, line 216: Removing because this is validated [here](https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/vsphere.go#L56)

pkg/providers/vsphere/validator.go
```diff
@@ -107,25 +107,9 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, vsphereCl
...
-		return errors.New("VSphereMachineConfig VM resourcePool for control plane is not set or is empty")
-	}
-	if controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Bottlerocket && controlPlaneMachineConfig.Spec.OSFamily != anywherev1.Ubuntu {
-		return fmt.Errorf("control plane osFamily: %s is not supported, please use one of the following: %s, %s", controlPlaneMachineConfig.Spec.OSFamily, anywherev1.Bottlerocket, anywherev1.Ubuntu)
-	}
-
-	workerNodeGroupConfigs := vsphereClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations
-	for _, workerNodeGroupConfig := range workerNodeGroupConfigs {
-		if workerNodeGroupConfig.Name == "" {
-			return errors.New("must specify name for worker nodes")
```
taneyland, line 126: Removing because it is already validated [here](https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster.go#L409)

mitalipaygude (approved):

pkg/providers/vsphere/validator.go
```diff
@@ -179,51 +148,12 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, vsphereCl
...
 
-	if etcdMachineConfig != nil && etcdMachineConfig.Spec.OSFamily != anywherev1.Bottlerocket && etcdMachineConfig.Spec.OSFamily != anywherev1.Ubuntu {
-		return fmt.Errorf("etcd node osFamily: %s is not supported, please use one of the following: %s, %s", etcdMachineConfig.Spec.OSFamily, anywherev1.Bottlerocket, anywherev1.Ubuntu)
-	}
-
-	if etcdMachineConfig != nil && controlPlaneMachineConfig.Spec.OSFamily != etcdMachineConfig.Spec.OSFamily {
-		return errors.New("control plane and etcd machines must have the same osFamily specified")
-	}
-
-	if err := v.validateSSHUsername(controlPlaneMachineConfig); err == nil {
```
mitalipaygude, line 190: Were these validations moved too in a different PR?

pkg/providers/vsphere/validator.go
mitalipaygude: is this link the right one?

taneyland: Right file, wrong line number. Updated it!

pkg/providers/vsphere/validator.go
taneyland: No, I moved it [here](https://github.com/aws/eks-anywhere/blob/9192960b13e7a0fe95f40abfb2e43cac89e023be/pkg/api/v1alpha1/vspheremachineconfig.go#L137) in this same PR. I just moved it out of the method since it is just one validation that can stand alone like the others

pkg/providers/vsphere/validator.go
g-gaston:
```go
for _, machineConfig := range configs {
    osFamily = machineConfig.Spec.OSFamily
    break
}
```

taneyland: Yeah I considered doing this before. I personally don't love the idea of adding this extra loop but will add this for efficiency's sake

g-gaston (commented):
/lgtm

PR #3552, opened 2022-10-04
Move some cluster validations to api

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR moves some cluster validations from the provider to the Cluster object API. This removes the need to redefine the same validations across providers because they will now happen at he cluster level. These changes also allow us to validate these fields in the controller during cluster reconciliation. These changes are particularly important for the new full lifecycle API that will allow us to also create additional workload clusters managed by an initial EKS-A cluster

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -166,10 +166,8 @@ func NewCluster(clusterName string) *Cluster {
 
 var clusterConfigValidations = []func(*Cluster) error{
 	validateClusterConfigName,
-	// TODO(chrisdoherty) Uncomment and fix unit tests. This broke _lots_ of unit tests so will
-	// return shortly.
 	validateControlPlaneEndpoint,
-	// validateControlPlaneMachineRef,
+	validateMachineGroupRefs,
 	validateControlPlaneReplicas,
 	validateWorkerNodeGroups,
 	validateNetworking,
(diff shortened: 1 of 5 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -393,6 +390,17 @@ func validateControlPlaneEndpoint(clusterConfig *Cluster) error {
 	if clusterConfig.Spec.DatacenterRef.Kind == DockerDatacenterKind && clusterConfig.Spec.ControlPlaneConfiguration.Endpoint != nil && clusterConfig.Spec.ControlPlaneConfiguration.Endpoint.Host != "" {
 		return fmt.Errorf("specifying endpoint host configuration in Cluster is not supported")
 	}
+
+	if clusterConfig.Spec.ControlPlaneConfiguration.Endpoint != nil && len(clusterConfig.Spec.ControlPlaneConfiguration.Endpoint.Host) <= 0 {
```
g-gaston, line 394:
```suggestion
	if clusterConfig.Spec.ControlPlaneConfiguration.Endpoint == nil || len(clusterConfig.Spec.ControlPlaneConfiguration.Endpoint.Host) <= 0 {
```

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -878,6 +878,12 @@ func TestGetAndValidateClusterConfig(t *testing.T) {
 			wantCluster: nil,
 			wantErr:     true,
 		},
+		{
+			testName:    "no control plane endpoint",
+			fileName:    "testdata/cluster_invalid_no_cp_endpoint.yaml",
```
g-gaston, line 883: Could we test it without going through yaml? I understand other tests do it that way, but it's unnecessary complexity

g-gaston (commented):
There a lot of changes in this PR and it's a bit difficult to follow without context. Could you complete the PR description with the set of changes you are doing here, why you are making those changes and how does this affect the controller/CLI behavior? 🙏🏻

PR #3554, opened 2022-10-04
Add namespace to controller install

TerryHowe:
Add namespace options to controller helm install


TerryHowe:
/hold 

Wait for https://github.com/aws/eks-anywhere-packages/pull/506

jonahjon:
What happens when you uninstall package controller, and try to re-install here?

TerryHowe:
> What happens when you uninstall package controller, and try to re-install here?


It should work, but we need to use these options on install now.

PR #3517, opened 2022-09-30
Add FluxConfig validation for workload cluster

jiayiwang7:
*Issue #, if available:*

Relates to #3468 

*Description of changes:*

We have workload cluster GitOpsConfig validation but hasn't added FluxConfig validation since its introduced.
Rewrite unit tests

*Testing (if applicable):*

Unit tests

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3556, opened 2022-10-04
Fix gitopsConfig validation for workload cluster

jiayiwang7:
*Issue #, if available:*

Fix a validation bug introduced in #3517 

*Description of changes:*

When using the deprecated `gitopsConfig`, CLI converts the object to `fluxConfig` and adds it in Spec. We need to skip the fluxConfig validation for this case since it's actually a gitOpsConfig validation that is covered in different method.

*Testing (if applicable):*

Unit tests


jonathanmeier5 (approved):
/approve
/lgtm

PR #3557, opened 2022-10-04
Add namespace in uri

vincentni:
*Issue #, if available:*
Harbor defines a project that features as the root namespace in URLs when interacting with OCI images. The PR proposes a CLI parameter to facilitate instrumenting that root namespace in conjunction with registry mirror configuration in the YAML.
This would patch URLs from `https://registry-mirror/my/image:tag` to `https://registry-mirror/namespace/my/image:tag`. `namespace` defaults to empty so it is backward compatible.

*Description of changes:*
The proposed changes are to introduce namespace (a.k.a projects in Harbor registry) to mirror registries in the cluster spec. The same concept is also introduced to `import images` command by a new flag `--namespace`. By default, namespace defaults to empty for backward compatibility. With such a concept in place, Harbor registry can parse out correct repository part from image/chart URIs and fetch them from remote ecr repositories in the case of Harbor being used as proxy cache.

Later, another curated-package specific namespace will be introduced to the `registryMirrorConfiguration` as in Harbor, one project can only proxy one remote registry but curated package images reside in a different place from other eksa images and requires credentials for access.

*Testing (if applicable):*


chrisdoherty4:
@vincentni I'm not gathering from the description what problem this solves, would you mind elaborating some?

PR #3566, opened 2022-10-05
Use CAPI machine name for node name and add validation on failuredomain name

wanyufe:
*Issue #, if available:*
Since failuredomain name can be used as label value for node, an validation is added to confirm failuredomain name is a valid label value:

-     must be 63 characters or less (can be empty),
-     unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
-     could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

Use CAPI machine name as node name so that they matches. 
*Description of changes:*
Original cluster yaml definition allows node label for control plane and worker node, to support label with a dynamic value (failure domain), we use `ds.meta_data.failuredomain` as value in eksa cluster yaml. Note: we have to remove {{ }} since '{' is not an allowed char in label value.
Specify CAPI machine name as node name when kubelet registers node.
*Testing (if applicable):*
Unit testing, manual testing.

```
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/os: linux
    cluster.x-k8s.io/failure-domain: az1
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: wanyufe-hkwh4
    kubernetes.io/os: linux
    node-role.kubernetes.io/control-plane: ""
    node-role.kubernetes.io/master: ""
    node.kubernetes.io/exclude-from-external-load-balancers: ""
  name: wanyufe-hkwh4

```

```
➜  eks-main git:(main) ✗ k get machines -A
NAMESPACE     NAME                            CLUSTER   NODENAME                        PROVIDERID                                           PHASE     AGE    VERSION
eksa-system   wanyufe-etcd-lkbj8              wanyufe                                   cloudstack:///eb75d397-00cc-44b2-a47b-681f58e86bd2   Running   3m3s   
eksa-system   wanyufe-hkwh4                   wanyufe   wanyufe-hkwh4                   cloudstack:///c53402a6-57ff-43a8-8dd7-e8292295814f   Running   3m3s   v1.21.13-eks-1-21-17
eksa-system   wanyufe-md-0-784765b4f6-s2tg5   wanyufe   wanyufe-md-0-784765b4f6-s2tg5   cloudstack:///9a53e6c5-6b23-4706-8dda-3e975a1a4ef7   Running   3m3s   v1.21.13-eks-1-21-17
```


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -40,6 +40,8 @@ import (
 const (
 	eksaLicense                = "EKSA_LICENSE"
 	controlEndpointDefaultPort = "6443"
+	nodeFailuredomainLabel     = "cluster.x-k8s.io/failure-domain"
+	nodeFailuredomainLabelValue = "'{{ds.meta_data.availability_zone}}'"
```
maxdrib, line 44: where does `availability_zone` come from on the host metadata?

pkg/providers/cloudstack/testdata/expected_results_main_md.yaml
```diff
@@ -14,6 +14,7 @@ spec:
             provider-id: cloudstack:///'{{ ds.meta_data.instance_id }}'
             read-only-port: "0"
             anonymous-auth: "false"
+            node-labels: cluster.x-k8s.io/failure-domain='{{ds.meta_data.availability_zone}}'
```
maxdrib, line 17: I think this will trigger a rollingupgrade on all of a management cluster's workload clusters. We should try to avoid that from happening. I'm not sure what the best way of doing that is.

controllers/resource/testdata/cloudstackKubeadmconfigTemplateSpec.yaml
```diff
@@ -14,6 +14,7 @@ spec:
             anonymous-auth: "false"
             provider-id: cloudstack:///'{{ ds.meta_data.instance_id }}'
             read-only-port: "0"
+            node-labels: cluster.x-k8s.io/failure-domain='{{ds.meta_data.availability_zone}}'
```
maxdrib, line 17: should we put a hold on this PR, if an updated CAPC version is a prerequisite?

wanyufe: This one does not depends on CAPC since `ds.meta_data.availability_zone` is a known/existed cloudstack metadata

pkg/providers/cloudstack/testdata/expected_results_main_md.yaml
wanyufe: We are trying to add label for nodes that is provisioned using eks-a, should this be considered a valid reason for upgrade?

pkg/providers/cloudstack/cloudstack.go
wanyufe: This is a built-in meta_data provided by cloudstack platform. Cloud-init jinja's template engine has the value to replace it with machine's zone name.

wanyufe: One concern is that the value of `ds.meta_data.availability_zone` is controlled by cloudstack platform and currently its value is the zone name where node/machine is deployed.

This availability_zone name maybe different from a higher level abstract failure_domain name customer defined in cluster.yaml file.

for example: 
`apiVersion: anywhere.eks.amazonaws.com/v1alpha1
kind: CloudStackDatacenterConfig
metadata:
  name: wanyufe
spec:
  availabilityZones:
  - name: az1
    domain: ROOT
    account: admin
    zone:
      name: zone1
      network:
        name: GuestNet1
    credentialsRef: global
    managementApiEndpoint: http://192.168.1.130:8080/client/api
---
`

the `ds.meta_data.availability_zone` value is zone1, but customer may want to see az1 as failure_domain. 

If so, we can provide a different name like `ds.meta_data.failure_domain`, and replace it with failure domain name in CAPC code. This needs to be discussed.

jweite-amazon: Agreed that Zone alone is no longer accurate.  Let's open an issue for this.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -1313,4 +1313,4 @@ func (p *cloudstackProvider) InstallCustomProviderComponents(ctx context.Context
 
 func machineDeploymentName(clusterName, nodeGroupName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, nodeGroupName)
-}
+}
```
maxdrib, line 1316: could you readd this newline?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -197,6 +199,11 @@ func (v *CloudStackDatacenterConfig) Validate() error {
 	}
 	azSet := make(map[string]bool)
 	for _, az := range v.Spec.AvailabilityZones {
+		errorMessages := validation.IsValidLabelValue(az.Name)
+		if len(errorMessages) > 0 {
```
maxdrib, line 203: Why are we collecting a list of errorMessages here? Then we check for multiple errors when there can only be one a single one. It seems like it would make more sense to have

```
if azNameInvalid {
    return fmt.Errorf("invalid name)
}
```

pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -333,7 +333,7 @@ spec:
 {{- if .kubeletExtraArgs }}
 {{ .kubeletExtraArgs.ToYaml | indent 10 }}
 {{- end }}
-        name: '{{`{{ ds.meta_data.local_hostname }}`}}'
```
maxdrib, line 336: why are we changing these templates? If we're passing the labels using the WorkerNodeLabels mechanism, why should these change?

maxdrib (changes requested):

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -197,6 +199,11 @@ func (v *CloudStackDatacenterConfig) Validate() error {
 	}
 	azSet := make(map[string]bool)
 	for _, az := range v.Spec.AvailabilityZones {
+		errorMessages := validation.IsValidLabelValue(az.Name)
```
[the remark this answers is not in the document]
wanyufe: yes, if we don't use it as label values, we don't need the validation.

Do you think we limit the chars domain unnecessary?

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
wanyufe: The error message from `errorMessages` can be multiple, here we collapse them into one. 
If we only use `invalid name`, customer misses the details about how to fix the error.

pkg/providers/cloudstack/config/template-cp.yaml
wanyufe: This is the fix for hostname mismatch between node and CAPI machine.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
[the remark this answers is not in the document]
maxdrib: I'm glad you're adding validation on the names. I think the Label definition might be too loose. It allows empty strings and periods. I think we should try to make it a valid K8s object name (alphanumeric with - or _). I'm not sure if there's an existing validation for that

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
maxdrib: could you change this for all the errors being returned by this function then? Otherwise it's only being used in one place.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -165,6 +165,7 @@ type ControlPlaneConfiguration struct {
 	// Taints define the set of taints to be applied on control plane nodes
 	Taints []corev1.Taint `json:"taints,omitempty"`
 	// Labels define the labels to assign to the node
+	// NOTE: CAPC provider supports using 'ds.meta_data.failuredomain' as label value and replaces it with proper failuredomain value.
```
maxdrib, line 168: Can you add an e2e test case for this contract change?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -239,6 +240,7 @@ type WorkerNodeGroupConfiguration struct {
 	// Taints define the set of taints to be applied on worker nodes
 	Taints []corev1.Taint `json:"taints,omitempty"`
 	// Labels define the labels to assign to the node
+	// NOTE: CAPC provider supports using 'ds.meta_data.failuredomain' as label value and replaces it with proper failuredomain value.
```
jiayiwang7, line 243: Same as above.

pkg/api/v1alpha1/cluster_types.go
wanyufe: I agree, do we have a EKA-A doc somewhere I can update.

pkg/api/v1alpha1/cloudstackdatacenterconfig_types.go
```diff
@@ -16,6 +16,8 @@ package v1alpha1
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/validation"
```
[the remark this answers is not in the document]
wanyufe: moved.

wongni (approved):

maxdrib (approved):

pkg/api/v1alpha1/cluster_types.go
wanyufe: Since eks-a doc is being worked on by some other resource, create an issue and notify :
https://github.com/aws/eks-anywhere/issues/3735

PR #3569, opened 2022-10-05
Add mutation webhook for vsphere machine configs

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR adds a mutation webhook to set defaults on vsphere machine configs. This is especially useful when creating clusters using the full lifecycle API via tools like `kubectl apply` or `Gitops`.

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/vspheremachineconfig.go
+++ b/pkg/api/v1alpha1/vspheremachineconfig.go
@@ -7,6 +7,8 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
+
+	"github.com/aws/eks-anywhere/pkg/logger"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


pkg/api/v1alpha1/vspheremachineconfig.go
```diff
@@ -76,3 +80,46 @@ func GetVSphereMachineConfigs(fileName string) (map[string]*VSphereMachineConfig
 	}
 	return configs, nil
 }
+
+func setVSphereMachineConfigDefaults(machineConfig *VSphereMachineConfig) {
+	if len(machineConfig.Spec.Folder) <= 0 {
+		logger.Info("VSphereMachineConfig Folder is not set or is empty. Defaulting to root vSphere folder.")
```
d8660091, line 86: Missing defaulting value assignment?

d8660091: Got it, so it's just for logging purpose

mitalipaygude (approved):

PR #3578, opened 2022-10-05
Fix trimming of Helm chart asset name

abhay-krishna:
`strings.Trim`'s second argument is a cutset, so it trims any character in the set if found at the start or end. The method that we want is `strings.TrimSuffix`.


release/pkg/operations/upload.go
```diff
@@ -86,7 +86,7 @@ func UploadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]re
 				// and then use Helm package and push commands to upload chart to ECR Public
 				if !r.DryRun && (strings.HasSuffix(artifact.Image.AssetName, "helm") || strings.HasSuffix(artifact.Image.AssetName, "chart")) {
 					// Trim -helm on the packages helm chart, but don't need to trim tinkerbell chart since the AssetName is the same as the repoName
-					trimmedAsset := strings.Trim(artifact.Image.AssetName, "-helm")
+					trimmedAsset := strings.TrimSuffix(artifact.Image.AssetName, "-helm")
```
TerryHowe, line 89: What about the "chart" case? I didn't know we had this dependency here.

abhay-krishna: For the Tinkerbell chart, we don't need this since the asset name `tinkerbell-chart` is the same as the ECR repo name.  For packages, the asset name is `eks-anywhere-packages-helm`, so we do this.

PR #3580, opened 2022-10-05
Fix URI argument to helm push command

abhay-krishna:
Helm push command takes in a tarball and a remote URI as arguments. If the destination repo name has many parts like `foo/bar/baz`, the tarball should be baz.tar.gz while the remote URI would be `oci://registry/foo/bar`. Currently, the release code does not append the prefix to the registry URI, which causes errors.


abhay-krishna:
Overriding since this fix doesn't affect these tests
/override eks-anywhere-presubmit

Issue #3248, opened 2022-09-02
Setup vSphere User Command


jonathanmeier5:
As outlined in [our design](https://github.com/aws/eks-anywhere/blob/main/designs/vsphere-user-setup.md), this task involves building the following command:

```
eksctl anywhere exp vsphere setup user -f config.yaml
```

PR #3589, opened 2022-10-06
Setup vSphere User Business Logic

jonathanmeier5:
*Issue #, if available:*

*Description of changes:*
Add business logic for setting up a vsphere user.

*Testing (if applicable):*
Manual testing running CLI to setup user.


pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+func SetupGOVCEnv(ctx context.Context, vsuc *VSphereSetupUserConfig) error {
+	os.Setenv("GOVC_URL", vsuc.Spec.Connection.Server)
+	os.Setenv("GOVC_INSECURE", strconv.FormatBool(vsuc.Spec.Connection.Insecure))
+	os.Setenv("GOVC_DATACENTER", vsuc.Spec.Datacenter)
+	return nil
+}
+
+func Run(ctx context.Context, vsuc *VSphereSetupUserConfig, govc vsphere.ProviderGovcClient) error {
```
chrisdoherty4, line 26: I would avoid depending on the `ProviderGovcClient` (bad naming too) and define the interface you need in this package.

chrisdoherty4, line 26: Missing func doc. Elaborate on what this does/is.

cmd/eksctl-anywhere/cmd/vspheresetupuser.go
```diff
@@ -0,0 +1,86 @@
...
+	if vsuo.force && vsuo.password != "" {
+		return fmt.Errorf("--password and --force are mutually exclusive. --force may only be run on an existing user")
+	}
+
+	c, err := setupuser.GenerateConfig(ctx, vsuo.fileName)
+	if err != nil {
+		return err
+	}
+
+	// hacky
```
chrisdoherty4, line 54: Is this still needed? An explanation on the func doc would be more effective.

cmd/eksctl-anywhere/cmd/vspheresetupuser.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return err
+	}
+	deps, err := dependencies.NewFactory().WithGovc().Build(ctx)
+	if err != nil {
+		return err
+	}
+	defer close(ctx, deps)
+
+	// if force flag not used, we should create a user
```
chrisdoherty4, line 65: We we be able to convert this into a "why" explanation? Its not clear why we need to create the user if force isn't specified or why we don't if it is specified.

cmd/eksctl-anywhere/cmd/vspheresetupuser.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (vsuo *vSphereSetupUserOptions) setupUser(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+
+	if vsuo.force && vsuo.password != "" {
+		return fmt.Errorf("--password and --force are mutually exclusive. --force may only be run on an existing user")
+	}
+
+	c, err := setupuser.GenerateConfig(ctx, vsuo.fileName)
```
chrisdoherty4, line 49: Given `c` is used 31 lines after its used, would we be able to give it a slightly more elaborate name, e.g. `cfg`.

cmd/eksctl-anywhere/cmd/vspheresetupuser.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/aws/eks-anywhere/pkg/providers/vsphere/setupuser"
+)
+
+type vSphereSetupUserOptions struct {
+	fileName string
+	force    bool
+	password string
+}
+
+var vsuo = &vSphereSetupUserOptions{}
```
chrisdoherty4, line 19: Given this is package state (which I hope we remove across the board for commands in the future) do you think we could expand its name even if that means repeating the type? Its similar to my other comment about usage far from the site of declaration.

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,187 @@
...
+
+	err = associateRolesToObjects(ctx, vsuc, govc)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func CreateUser(ctx context.Context, vsuc *VSphereSetupUserConfig, govc vsphere.ProviderGovcClient, password string) error {
```
chrisdoherty4, line 50: This function seems redundant. We don't handle the error at all so it can be abbreviated to:

```go
func CreateUser(ctx context.Context, vsuc *VSphereSetupUserConfig, govc vsphere.ProviderGovcClient, password string) error {
	return govc.CreateUser(ctx, ...)
}
```

The caller can just call `govc` directly.

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createRoles(ctx context.Context, vsuc *VSphereSetupUserConfig, govc vsphere.ProviderGovcClient) error {
+	// create roles
```
chrisdoherty4, line 78: Looks like this can be removed.

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	return nil
+}
+
+func associateRolesToObjects(ctx context.Context, vsuc *VSphereSetupUserConfig, govc vsphere.ProviderGovcClient) error {
+	// global on root
+	// admin to template and vm folders
+	// user on network, datastore, and resourcepool
```
chrisdoherty4, line 103: Some more removable developer notes?

chrisdoherty4 (commented):
Generally looks good, some clarifying questions and tidy up comments.

pkg/providers/vsphere/setupuser/setupuser.go
jonathanmeier5: Thought I got this, cleaned up now.

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,185 @@
...
+
+func createGroup(ctx context.Context, vsuc *VSphereSetupUserConfig, govc GovcClient) error {
+	exists, err := govc.GroupExists(ctx, vsuc.Spec.GroupName)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		err = govc.CreateGroup(ctx, vsuc.Spec.GroupName)
+	} else {
+		logger.V(0).Info(fmt.Sprintf("Skipping creating %s because it already exists", vsuc.Spec.GroupName))
```
ptrivedi, line 71:
```suggestion
		logger.V(0).Info(fmt.Sprintf("Skipping creating group %s because it already exists", vsuc.Spec.GroupName))
```

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func addUserToGroup(ctx context.Context, vsuc *VSphereSetupUserConfig, govc GovcClient) error {
+	// associate user to group
+	err := govc.AddUserToGroup(ctx, vsuc.Spec.GroupName, vsuc.Spec.Username)
```
ptrivedi, line 125: just like for the Role, have we tested to see that if the user is not found (if someone used the force flag accidentally with a non-existent user), err will truly be non-nil?

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,185 @@
...
+	GroupExists(ctx context.Context, name string) (bool, error)
+	AddUserToGroup(ctx context.Context, name string, username string) error
+	RoleExists(ctx context.Context, name string) (bool, error)
+	CreateRole(ctx context.Context, name string, privileges []string) error
+	SetGroupRoleOnObject(ctx context.Context, principal string, role string, object string, domain string) error
+}
+
+// SetupGOVCEnv creates appropriate govc environment variables to build govc client.
+func SetupGOVCEnv(ctx context.Context, vsuc *VSphereSetupUserConfig) error {
+	os.Setenv("GOVC_URL", vsuc.Spec.Connection.Server)
```
ptrivedi, line 32: good to check for err return from Setenv

pkg/providers/vsphere/setupuser/setupuser.go
jonathanmeier5: Great question. Yes, error is non-nil when the user does not exist:
```
Error: govc returned error govc: user "jwmeier-22-10-17-1" not found
```

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,185 @@
...
+	return requiredPrivs, nil
+}
+
+type vsphereRole struct {
+	name  string
+	privs []string
+}
+
+func getRoles(vsuc *VSphereSetupUserConfig) []vsphereRole {
+	globalPrivs, _ := getPrivsFromFile(config.VSphereGlobalPrivsFile)
```
ptrivedi, line 162: check for err returns

pkg/providers/vsphere/setupuser/setupuser.go
```diff
@@ -0,0 +1,185 @@
...
+	AddUserToGroup(ctx context.Context, name string, username string) error
+	RoleExists(ctx context.Context, name string) (bool, error)
+	CreateRole(ctx context.Context, name string, privileges []string) error
+	SetGroupRoleOnObject(ctx context.Context, principal string, role string, object string, domain string) error
+}
+
+// SetupGOVCEnv creates appropriate govc environment variables to build govc client.
+func SetupGOVCEnv(ctx context.Context, vsuc *VSphereSetupUserConfig) error {
+	os.Setenv("GOVC_URL", vsuc.Spec.Connection.Server)
+	os.Setenv("GOVC_INSECURE", strconv.FormatBool(vsuc.Spec.Connection.Insecure))
```
ptrivedi, line 33: Let's test and double-check if we can avoid using GOVC_INSECURE by specifying thumbprint for self-signed certs and also see if it can be avoided in trusted cert cases.

jonathanmeier5: Breaking out this work into a future ticket based on offline conversation.
Ticket: https://github.com/aws/eks-anywhere/issues/3708

Issue #2814, opened 2022-07-28
Workload cluster deployment with `AWSIamConfig` fails if the management cluster doesn't have `AWSIamConfig` 


abhinavmpandey08:
**What happened**:
Configuring `aws-iam-authenticator` on workload cluster results in a failed deployment if the management cluster didn't have `aws-iam-authenticator` configured. 
 The CLI will timeout at "Creating workload cluster" stage after the external ETCD nodes are created.
If you describe the control plane machine, you will see the following status:
```yaml
  status:
    bootstrapReady: false
    conditions:
    - lastTransitionTime: "2022-07-26T23:50:02Z"
      message: 0 of 2 completed
      reason: DataSecretGenerationFailed
      severity: Warning
      status: "False"
      type: Ready
    - lastTransitionTime: "2022-07-26T23:50:02Z"
      message: 'failed to resolve file source: secret not found: eksa-system/aws-iam-authenticator-ca:
        secrets "aws-iam-authenticator-ca" not found'
      reason: DataSecretGenerationFailed
      severity: Warning
      status: "False"
      type: BootstrapReady
    - lastTransitionTime: "2022-07-26T23:50:02Z"
      reason: WaitingForBootstrapData
      severity: Info
      status: "False"
      type: InfrastructureReady
```

**What you expected to happen**:
Workload cluster with `AWSIamConfig` to be created successfully

**How to reproduce it (as minimally and precisely as possible)**:
Create management cluster without `AWSIamConfig` and then try to create a workload cluster from the management cluster with `AWSIamConfig`

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

PR #3594, opened 2022-10-06
IAM authenticator only on workload support

pokearu:
*Issue #2814*

*Description of changes:*
Changes enable setting up `aws-iam-authenticator` on a workload cluster when its not configured on the management cluster.
We now prepend cluster name to the `aws-iam-authenticator-ca` that is needed on the bootstrap/management cluster before adding it on a workload cluster.

*Testing:*
```bash
eksctl-anywhere create cluster -f br-cluster-workload.yaml --kubeconfig ./br/br-eks-a-cluster.kubeconfig
```

### DIFF
--- a/pkg/awsiamauth/awsiamauth.go
+++ b/pkg/awsiamauth/awsiamauth.go
@@ -92,12 +92,13 @@ func (a *AwsIamAuth) GenerateManifestForUpgrade(clusterSpec *cluster.Spec) ([]by
 	return a.templateBuilder.GenerateManifest(clusterSpec, uuid.Nil)
 }
 
-func (a *AwsIamAuth) GenerateCertKeyPairSecret() ([]byte, error) {
+func (a *AwsIamAuth) GenerateCertKeyPairSecret(bootstrapClusterName string) ([]byte, error) {
 	certPemBytes, keyPemBytes, err := a.certgen.GenerateIamAuthSelfSignCertKeyPair()
 	if err != nil {
 		return nil, fmt.Errorf("generating aws-iam-authenticator cert key pair secret: %v", err)
 	}
 	data := map[string]string{
+		"clusterName":  bootstrapClusterName,
 		"namespace":    constants.EksaSystemNamespace,
 		"certPemBytes": base64.StdEncoding.EncodeToString(certPemBytes),
 		"keyPemBytes":  base64.StdEncoding.EncodeToString(keyPemBytes),
--- a/pkg/awsiamauth/awsiamauth.go
+++ b/pkg/awsiamauth/awsiamauth.go
@@ -92,13 +92,13 @@ func (a *AwsIamAuth) GenerateManifestForUpgrade(clusterSpec *cluster.Spec) ([]by
 	return a.templateBuilder.GenerateManifest(clusterSpec, uuid.Nil)
 }
 
-func (a *AwsIamAuth) GenerateCertKeyPairSecret(bootstrapClusterName string) ([]byte, error) {
+func (a *AwsIamAuth) GenerateCertKeyPairSecret(managementClusterName string) ([]byte, error) {
 	certPemBytes, keyPemBytes, err := a.certgen.GenerateIamAuthSelfSignCertKeyPair()
 	if err != nil {
 		return nil, fmt.Errorf("generating aws-iam-authenticator cert key pair secret: %v", err)
 	}
 	data := map[string]string{
-		"clusterName":  bootstrapClusterName,
+		"clusterName":  managementClusterName,
 		"namespace":    constants.EksaSystemNamespace,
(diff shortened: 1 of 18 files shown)


pkg/clustermanager/cluster_manager.go
```diff
@@ -123,7 +123,7 @@ type Networking interface {
 type AwsIamAuth interface {
 	GenerateManifest(clusterSpec *cluster.Spec) ([]byte, error)
 	GenerateManifestForUpgrade(clusterSpec *cluster.Spec) ([]byte, error)
-	GenerateCertKeyPairSecret() ([]byte, error)
+	GenerateCertKeyPairSecret(bootstrapClusterName string) ([]byte, error)
```
ptrivedi, line 126: let's call this variable name 'clusterName' for clarity. It could either take workload or bootstrap cluster names

pkg/awsiamauth/awsiamauth.go
```diff
@@ -92,12 +92,13 @@ func (a *AwsIamAuth) GenerateManifestForUpgrade(clusterSpec *cluster.Spec) ([]by
 	return a.templateBuilder.GenerateManifest(clusterSpec, uuid.Nil)
 }
 
-func (a *AwsIamAuth) GenerateCertKeyPairSecret() ([]byte, error) {
+func (a *AwsIamAuth) GenerateCertKeyPairSecret(bootstrapClusterName string) ([]byte, error) {
```
ptrivedi, line 95: call this 'clusterName' because it could be either bootstrap or workload cluster?

pkg/workflows/create.go
```diff
@@ -99,6 +99,14 @@ type InstallCuratedPackagesTask struct{}
 
 func (s *CreateBootStrapClusterTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
 	if commandContext.BootstrapCluster != nil {
+		if commandContext.ClusterSpec.AWSIamConfig != nil {
+			logger.Info("Creating aws-iam-authenticator certificate and key pair secret on bootstrap cluster")
+			if err := commandContext.ClusterManager.CreateAwsIamAuthCaSecret(ctx, commandContext.BootstrapCluster, commandContext.ClusterSpec.Cluster.Name); err != nil {
+				commandContext.SetError(err)
```
ptrivedi, line 105: are you able to cover these via a test?

pkg/clustermanager/cluster_manager.go
pokearu: Actually I named it this because we should always pass in the bootstrap cluster name here, either KinD or the management cluster.
I could change it and add a comment instead.

pkg/awsiamauth/awsiamauth.go
pokearu: It should always be the bootstrap cluster (so KinD or management). I can add a comment 👍

abhinavmpandey08: non-blocking but I think [bootstrap cluster](https://cluster-api.sigs.k8s.io/reference/glossary.html#bootstrap-cluster) refers to a temporary cluster that is used to provision a long lived cluster. In our case, this would be the kind cluster. The permanent management cluster is never a bootstrap cluster.
`managementClusterName` would be more appropriate since the bootstrap cluster is also technically a management cluster for most of its duration.
I don't really have a strong opinion on this, so I will leave it up to you guys to decide.

pkg/workflows/create.go
```diff
@@ -99,6 +99,14 @@ type InstallCuratedPackagesTask struct{}
...
+				return &CollectMgmtClusterDiagnosticsTask{}
+			}
```
abhinavmpandey08, line 107: This is being called here as well as on line 141 below. Can you move this before `if commandContext.BootstrapCluster != nil {` check on line 101 and remove it on line 141?

abhinavmpandey08 (commented):
looks good!
small question: with this change, will all the clusters have a separate set of certs?
Also, how will that affect IAM auth on already existing clusters where all the workloads are using the same certs.

pokearu:
> looks good! small question: with this change, will all the clusters have a separate set of certs? Also, how will that affect IAM auth on already existing clusters where all the workloads are using the same certs.

So basically the secret was only used when writing the cert files, so technically even though we referenced the same secret before, each cluster had separate files on their CP nodes host.
Now if a user upgrades a cluster with the old IAM, the KCP would get updated to reference the new secrets and roll the nodes. The new node would have an iam-auth that reads from the new certs.

pkg/workflows/create.go
pokearu: Actually I think that might not work? The call on line 141 actually uses the newly created KinD bootstrap cluster, and that reference is only available on 120. That is why I had to repeat the calls.

PR #3595, opened 2022-10-06
Remove extra replica of cilium operator on single-node

d8660091:
*Issue #, if available:* On single-node cluster, an extra pod for cilium-operator is scheduled, but it will never run since the cluster has only 1 host.

*Description of changes:* Add replicas = 1 [parameter](https://github.com/cilium/cilium/blob/v1.11.8/install/kubernetes/cilium/values.yaml#L1362) to cilium chart.

*Testing (if applicable):*
1. Unit test
2. Manual e2e test on Dell single-node cluster

### DIFF
--- a/pkg/networking/cilium/templater_test.go
+++ b/pkg/networking/cilium/templater_test.go
@@ -385,3 +385,18 @@ func TestGenerateNetworkPolicyManifestInvalidKubeVersion(t *testing.T) {
 	tt.Expect(err).To(HaveOccurred(), "templater.GenerateUpgradeManifest() should fail")
 	tt.Expect(err).To(MatchError(ContainSubstring("invalid major version in semver")))
 }
+
+func TestTemplaterGenerateManifestForSingleNodeCluster(t *testing.T) {
+	tt := newtemplaterTest(t)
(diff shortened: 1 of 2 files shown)


pkg/networking/cilium/templater_test.go
```diff
@@ -385,3 +385,20 @@ func TestGenerateNetworkPolicyManifestInvalidKubeVersion(t *testing.T) {
...
+	tt.h.EXPECT().
+		Template(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+		Do(func(_ interface{}, _ interface{}, _ interface{}, _ interface{}, values map[string]interface{}, _ interface{}) {
+			tt.Expect(reflect.ValueOf(values["operator"]).MapIndex(reflect.ValueOf("replicas")).Interface().(int)).To(Equal(1))
+		})
+
+	_, err := tt.t.GenerateManifest(tt.ctx, tt.spec)
+	if err != nil {
+		t.Fatalf("failed to generate manifest: %v", err)
+	}
```
ptrivedi, line 403: do you want to add an assertion for the contents of the manifest?

d8660091: Good idea!

PR #3597, opened 2022-10-06
Add support for VLAN ID in Tinkerbell provider:

jacobweinstock:
*Issue #, if available:*
#3596

*Description of changes:*
This enables trunked port environments to work with EKSA provider Tinkerbell. Official docs to follow. This functionality will only work properly with updated Tinkerbell stack services. The updated stack versions PR in `eks-anywhere-build-tooling` is incoming.

*Testing (if applicable):*
Manual tests were run to validate that vlan ids provided in the `hardware.csv` propagated properly into the `hardware.tinkerbell.org` crd.


pkg/providers/tinkerbell/hardware/validator.go
```diff
@@ -134,6 +136,17 @@ func StaticMachineAssertions() MachineAssertion {
 			}
 		}
 
+		if m.VLANID != "" {
+			i, err := strconv.Atoi(m.VLANID)
+			if err != nil {
+				return errors.New("VLANID: must be an integer")
+			}
+
+			if i < 1 || i > 4094 {
```
chrisdoherty4, line 145: Could we throw down a constant name and/or a link to something like https://en.m.wikipedia.org/wiki/VLAN#IEEE_802.1Q so the 4094 is more understandable?

I double took at this and googled it.

pkg/providers/tinkerbell/hardware/validator_test.go
```diff
@@ -206,6 +206,12 @@ func TestStaticMachineAssertions_InvalidMachines(t *testing.T) {
 		"InvalidWithJustDev": func(h *hardware.Machine) {
 			h.Disk = "/dev/"
 		},
+		"InvalidVLAN": func(h *hardware.Machine) {
+			h.VLANID = "0"
+		},
+		"NonIntVLAN": func(h *hardware.Machine) {
```
chrisdoherty4, line 212: Do we want a maximum bounds rest too?

pkg/providers/tinkerbell/hardware/validator.go
jacobweinstock: for sure, will add.

pkg/providers/tinkerbell/hardware/validator_test.go
jacobweinstock: sounds good, will add.

pkg/providers/tinkerbell/hardware/validator.go
jacobweinstock: added

pkg/providers/tinkerbell/hardware/validator_test.go
jacobweinstock: added

chrisdoherty4 (approved):

PR #3601, opened 2022-10-07
go-native client preparations (1 of 6)

ewollesen:
These two small commits are a precursor to a series of PRs I'll be opening to bring a go-native k8s client to the curated packages CLI commands.

PR #3602, opened 2022-10-07
feat(CLI): go-native client for CLI commands (2 of 6)

ewollesen:
This adds functions for creating a go-native kubernetes client to be used by
CLI commands.

The client is not yet used, that will be coming in follow-up commits/PRs. I've
split them up because of their size. This will work as an expand-contract
process[1], where new features like this kubernetes client are added, then
existing code will be modified to use these new features, then the old
solutions for these operations will be removed. Hopefully this will keep the
PR sizes manageable, without leaving the main branch in a broken state.

[1]: https://martinfowler.com/bliki/ParallelChange.html

Depends on #3601

### DIFF
--- a/pkg/curatedpackages/kubeclient.go
+++ b/pkg/curatedpackages/kubeclient.go
@@ -1,0 +1,66 @@
+package curatedpackages
+
+import (
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+// NewKubeClientFromFilename creates a controller-runtime k8s client for use
+// by CLI commands.
+func NewKubeClientFromFilename(kubeConfigFilename string) (client.Client, error) {
+	return newKubeClient(kubeConfigFilename, nil)
+}
+
+// newKubeClient creates a controller-runtime k8s client for use by CLI commands.
+//
+// If the RESTConfigurator is nil, a default configurator will be used.
+//
+// This isn't exported because it shouldn't be necessary in normal usage. It
+// exists because we wanted to be able to inject a RESTConfigurator for tests,
+// to bypass the network calls that the k8s client makes at
+// initialization. There's no reason why the RESTConfigurator would need to be
+// modified otherwise. If that condition changes, then so too should the
+// constructor story here.
+func newKubeClient(kubeConfigFilename string, rc restConfigurator) (client.Client, error) {
+	data, err := os.ReadFile(kubeConfigFilename)
+	if err != nil {
+		return nil, err
+	}
+	if rc == nil {
+		rc = restConfigurator(clientcmd.RESTConfigFromKubeConfig)
+	}
+	restConfig, err := rc.Config(data)
+	if err != nil {
+		return nil, err
+	}
+	scheme := runtime.NewScheme()
+	if err := packagesv1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+	return client.New(restConfig, client.Options{Scheme: scheme})
+}
+
+// restConfigurator abstracts the creation of a controller-runtime *rest.Config.
+//
+// This abstraction improves testing, as all known methods of instantiating a
+// *rest.Config try to make network calls, and that's something we'd like to
+// keep out of our unit tests as much as possible. In addition, where we do
+// use them in unit tests, we need to be prepared with a controller-runtime
+// EnvTest environment.
+//
+// For normal, non-test use, this can safely be ignored.
+type restConfigurator func([]byte) (*rest.Config, error)
+
+func (c restConfigurator) Config(data []byte) (*rest.Config, error) {
+	return c(data)
+}
--- a/pkg/curatedpackages/kubeclient_test.go
+++ b/pkg/curatedpackages/kubeclient_test.go
@@ -1,0 +1,24 @@
+package curatedpackages
+
+import (
+	"bytes"
+	"testing"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/aws/eks-anywhere/internal/test"
+)
+
+func TestNewKubeClient(t *testing.T) {
+	t.Parallel()
+	cfg := test.UseEnvTest(t)
+	rc := restConfigurator(func(_ []byte) (*rest.Config, error) { return cfg, nil })
+	f := test.WithFakeFileContents(t, bytes.NewBufferString(""))
+
+	t.Run("golden path", func(t *testing.T) {
+		_, err := newKubeClient(f.Name(), rc)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+	})
+}
--- a/internal/test/helpers.go
+++ b/internal/test/helpers.go
@@ -1,0 +1,87 @@
+package test
+
+import (
+	"errors"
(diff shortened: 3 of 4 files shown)


pkg/curatedpackages/kubeclient.go
```diff
@@ -0,0 +1,57 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
+)
+
+// NewKubeClientFromFilename creates a controller-runtime k8s client for use
+// by CLI commands.
+//
+// If the RESTConfigurator is nil, a default configurator will be used.
+func NewKubeClientFromFilename(kubeConfigFilename string, rc RESTConfigurator) (client.Client, error) {
```
a-cool-train, line 20: I assume this would be used from factory.go? Currently any bootstrapping is performed from factory.go.

ewollesen: I'm not sure that I understand your question. The function is exported, so you can call it from anywhere you like (so long as you import the curatedpackages package). I suspect there might be a better place for this to live long term, but for now, this seems to be fine.

pkg/curatedpackages/kubeclient.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if rc == nil {
+		rc = restConfigurator(clientcmd.RESTConfigFromKubeConfig)
+	}
+	restConfig, err := rc.Config(data)
+	if err != nil {
+		return nil, err
+	}
+	scheme := runtime.NewScheme()
+	utilruntime.Must(packagesv1.AddToScheme(scheme))
```
g-gaston, line 33: nit: I wonder if we should just return an error instead of panic here
Although I could buy the argument this should only fail in case of a programing error

pkg/curatedpackages/kubeclient.go
g-gaston, line 20: I would love to avoid passing nil, we are not really representing the abscence of data but the intent to not alter the default config. Also, I think it makes the caller code more difficult to read. 

We could use a builder pattern (maybe with closures as options). Or alternatively have a `NewKubeClientFromFilenameWithDefaultConfigurator(kubeConfigFilename string)` that just calls `NewKubeClientFromFilename`.

test/helpers.go
```diff
@@ -0,0 +1,89 @@
+package test
```
g-gaston, line 1: We already have a package for this kind of thing. It's in `internal/test`.
I would suggest putting this file there until we make the decision to move everything here/somewhere else.

test/helpers.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// EnvTest provides fake k8s control plane components for testing
+// purposes. The process of bringing up and tearing down the EnvTest framework
+// involves running a few binaries, and is not integrated into a vanilla "go
+// test" run, but rather can be run via the unit-test target in Makefile. See
+// https://book.kubebuilder.io/reference/envtest.html for details.
+//
+// TODO: What could be done to integrate EnvTest with go test, so that "go
+// test" would work?
+func UseEnvTest(t *testing.T) *rest.Config {
```
g-gaston, line 69: Couldn't we use the `envtest` package we have in our repo?
And call it liken this
https://github.com/aws/eks-anywhere/blob/119aab6f6bceed10028b391a5d722828f8c2348d/controllers/main_test.go#L12-L14

pkg/curatedpackages/kubeclient.go
ewollesen: I have no strong opinion, I was just copying what I've seen (specifically other places in the codebase manager/main.go). But I suppose since I can easily return an error, I should. I'll tweak.

pkg/curatedpackages/kubeclient.go
ewollesen: Yeah, I guess I'm just used to passing nil in constructors for things you don't need/want. You've brought this up before, so clearly you're not. :)

I'd argue that, especially for internal-facing code, where the only intended consumers are ourselves, this is fine. Clear documentation makes up for it. It's certainly simpler and easier to grok than closure-based initializers, with their multiple hops of indirection and abstraction.

So especially given that we've only got a single thing to pass or not right now, I'll elect to go with NewKubeClientFromFilenameWithOptions(kubeConfigFilename string, rc RESTConfigurator) and NewKubeClientFromFilename(kubeConfigFilename string) as the two choices. Hopefully that strikes a nice compromise between our views on the matter?

test/helpers.go
ewollesen: I think the two could be merged, but we'd probably need to have a good chat about that. Issues that come to mind immediately upon looking at newEnvironment:

1. It isn't clear if the client returned by environment can have its scheme modified, which is a requirement.
2. I have a hunch that we'd really like to be able to get our filthy paws on the *rest.Config that's returned by testenv, and that's not possible at present.
3. We don't need the manager at this time, it'd be preferable to make that optional. I already hate that I have to use testenv at all, and that it prevents me from being able to simply "go test" things. I'd prefer not to add controller startup to the things slowing down our tests.

Are you OK with it as a follow-up?

pkg/curatedpackages/kubeclient.go
ewollesen: Having now made these changes, I wonder if we're really doing the consumer any favors. They now have two different constructors to choose between. We haven't really made anything simpler for them.

All the "simpler" of the two constructors does is immediately call the other constructor, which I guess is a strong implication that the RESTConfigurator argument can be nil, which ... I guess is good, but I don't know that it's better than the simplicity of having one constructor with a decent comment explaining its usage. I don't feel like having to read a godoc to use some facility is a terrible thing.

All that said, I'm fine with either approach if you are.

test/helpers.go
g-gaston: Follow up sounds good!

pkg/curatedpackages/kubeclient.go
g-gaston: I prefer the two methods than passing a nil argument :)
But I do understand your concern about having two constructors to choose from

A third alternative would be to have a public setter to update the configurator after initialization. That would mean we would need to separate construction from initialization, like `NewClient` + `client.Init`

ewollesen: We had a bit of a discussion in Slack and settled on a private constructor that allows for the restConfigurator to be specified, but a single exported constructor that doesn't. This satisfies both of our desires for a simple clean interface, while still giving us testable code. Win-win I think!

pkg/curatedpackages/kubeclient_test.go
```diff
@@ -0,0 +1,24 @@
+package curatedpackages
```
vivek-koppuru, line 1: So we don't want to make this a test package, like `curatedpackages_test`? I understand that we are making using of `restConfigurator`, but if this abstraction improves testing, we can still have that comment and make it exported. Just curious if there is any other reason we aren't exporting that function. 

I see that we are calling `newKubeClient` as well. Not sure what you and @g-gaston discussed about this, but this can be a situation where we set the restConfigurator after calling the initial constructor.

ewollesen: I don't export restConfigurator because no one else needs it. Each name that's exported is something that could break someone else, often in surprising ways. I try not to export something without a good reason, and even then, something this small, and laser focused—I'd be just as likely to copy it as I would be to export it. I find its preferable to reduce my dependency surface vs remove 5 lines of duplicated code. There's also something to be said for having the definition close at hand, as opposed to in a file X packages away.

As for newKubeClient, I think you were included in the Slack thread, so you can see what we discussed. Yes, we could set restConfigurator after the fact, but felt that this approach melds the best of both worlds. The only place that has any reason to set a restConfigurator is a test, and so we elected to keep any mention of restConfigurator or anything to manipulate it, out of the exported interface. Normal users should not have any reason to modify the restConfigurator, and they shouldn't even have to know it exists.

vivek-koppuru (approved):

PR #3603, opened 2022-10-07
feat(CLI): BundleClient: go-native client for bundle operations (3 of 6)

ewollesen:
The bundle client uses a go-native kubernetes client to communicate with a
cluster to perform operations relating to curated packages bundles.

The client is not used yet, that will be coming in a separate commit/PR. I've
split them up because of their size. This will work as a sort of
expand-contract process[1], where new features like the bundle client are
added, then existing code will be modified to use these new features, then the
old solutions for these operations will be removed. Hopefully this will keep
the PR sizes manageable, without leaving the main branch in a broken state.

[1]: https://martinfowler.com/bliki/ParallelChange.html

Depends on #3602


ewollesen:
The test failed with an error opening pkg/bundle in eks-anywhere-packages. I checked the version in use v0.2.0-alpha.6, and as far as I can tell, pkg/bundle is definitely there. Going to ask for a retest and see what happens.

/retest

ewollesen:
OK, the problem is apparently when generating mocks. Interesting that I was able to run that locally... Let me try to reproduce.

ewollesen:
OK, the problem was an added -source flag in Makefile. Must have worked its way in via a rebase or something. A quick fix incoming.

pkg/curatedpackages/bundleclient.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/aws/eks-anywhere-packages/pkg/bundle"
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+// BundleClient abstracts the source of package bundle information.
+//
+// Whether the source is a registry, a cluster, the BundleClient should
+// abstract the details of retrieving a particular curated packages bundle.
+type BundleClient interface {
+	ActiveOrLatest(ctx context.Context) (*packagesv1.PackageBundle, error)
```
g-gaston, line 21: nit: 
```suggestion
	GetActiveOrLatestBundle(ctx context.Context) (*packagesv1.PackageBundle, error)
```

pkg/curatedpackages/bundleclient.go
```diff
@@ -0,0 +1,172 @@
...
+	// when specifying a custom registry mirror that requires authentication.
+	RegistryClient bundle.RegistryClient
+	// restConfigurator is used to inject testing constructs.
+	//
+	// It shouldn't be required for normal use.
+	restConfigurator restConfigurator
+}
+
+// NewBundleClient constructs a BundleClient based on the value of source.
+func NewBundleClient(source BundleSource, opts BundleClientOptions) (BundleClient, error) {
```
g-gaston, line 55: Having that `source` argument makes me think this should probably be two constructors and the caller should decide which one to call.

pkg/curatedpackages/bundleclient.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+}
+
+// clusterBundleClient implements BundleClient by querying an existing cluster.
+type clusterBundleClient struct {
+	kubeClient  client.Client
+	clusterName string
+}
+
+var _ BundleClient = (*clusterBundleClient)(nil)
```
g-gaston, line 88: what is this for?

pkg/curatedpackages/bundleclient.go
```diff
@@ -0,0 +1,172 @@
...
+
+// NewClusterBundleClient retrieves package bundle info from an existing cluster.
+func NewClusterBundleClient(kubeClient client.Client, clusterName string) *clusterBundleClient {
```
g-gaston, line 91: Can we just make `clusterBundleClient`? or make the constructor private if it's not used from the outside

pkg/curatedpackages/bundleclient.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		return NewClusterBundleClient(kubeClient, opts.ClusterName), nil
+	default:
+		return nil, fmt.Errorf("unhandled package bundle source %q", source)
+	}
+}
+
+// clusterBundleClient implements BundleClient by querying an existing cluster.
+type clusterBundleClient struct {
+	kubeClient  client.Client
```
g-gaston, line 84: ahhh interesting
Did you decide to use the controller runtime client instead of clientgo? curious about the rationale/experience

g-gaston, line 84: If you are interested, we have an interface than can satisfied by both kubectl and the controller runtime client. Not sure if that would simplify your transition or not.

We use it in a few places of the codebase where the CLI and the controller share code.

pkg/curatedpackages/bundleclient.go
ewollesen: Bundle is repetitive. The package is bundleclient, so bundleclient.ActiveOrLatestBundle is a form of stuttering. I'd prefer to avoid it. Same for Get, it's implied. If you need an authority: https://go.dev/doc/effective_go#names

pkg/curatedpackages/bundleclient.go
ewollesen: https://go.dev/doc/effective_go#blank_implements

It's a common thing you'll see a lot. In English, it says "an object of type pointer to clusterBundleClient implements the BundleClient interface". This bit of code will throw an error if the compiler determines that the previous statement is no longer true.

pkg/curatedpackages/bundleclient.go
ewollesen: No. If you could, I wouldn't have NewClusterBundleClient. :) The reason I say you can't, is because if you tried to instantiate your own clusterBundleClient, you wouldn't have a guarantee that you had kubeClient, or a clusterName, so every method would have to check for the existence of those things and throw a run-time error if they weren't defined. By making the struct not-exported, and the constructor exported, I let them be created only via the constructor, which requires the cluster name and the kubeClient.

I could make the constructor private I guess... But it doesn't buy me anything, since NewBundleClient calls it anyway, and that has to be exported... 

NewBundleClient is just a convenience method for lifting some common CLI code out of the separate CLI commands, and putting it here in this ... well, I guess it's factory method. But the point is there are several CLI commands that need to have a bundle client, for either a source argument, or a cluster argument, and NewBundleClient moves that repetitive logic out of the CLI commands to a single place that really knows more about it anyway.

pkg/curatedpackages/bundleclient.go
ewollesen: The experience with both has been pretty horrible so far. I haven't found good docs, so it's all just reading the code, and the testability of said code has been low and frustrating.

I went with this because as far as I could tell, clientgo wasn't extendable to general resources. That is, if you wanted to use it for something other than core kubernetes resources (and we definitely did), you were kinda out of luck. It's entirely possible I missed how to do that, but the controller-runtime version works great, once you get it set up, so...

ewollesen: I'm not _not_ interested. I know I had seen something with a client.Client elsewhere, but when I looked at it (admittedly briefly), it felt like it was in controller code, and was able to make assumptions about being in a cluster, which dramatically changed how it was instantiated (and consequently wouldn't work for my purposes, being that all of this is run from outside of a cluster). I definitely didn't do super thorough checking though.

ewollesen: I'd be interested as more of a follow-up item though, too, admittedly.

pkg/curatedpackages/bundleclient.go
ewollesen: If this truly is a nit, I think I'll stand by my choices. I prefer the lack of repetition and generally find "Get*" to be implied. https://go.dev/doc/effective_go#Getters

https://go.dev/doc/effective_go#package-names refers more to package names, but I think the same case can be made, ie that bundleclient.ActiveOrLatestBundle is a form of repetition that doesn't add any value over bundleclient.ActiveOrLatest. (Plus, if you're in doubt about what type it returns, since Go is a typed language, your editor or the compiler can easily tell you that information with greater certainty, and it's never out of date.)

pkg/curatedpackages/bundleclient.go
ewollesen: I could, but I think this is a better solution because there are a few different CLI commands that have to instantiate a bundle client with these same options. NewBundleClient lifts that common logic out of the individual CLI commands to a single, more authoritative place.

pkg/curatedpackages/bundleclient.go
g-gaston: It's a nit, so you can stand by your choice :)

That said, my suggestion to add `Get` is not because this is a getter (as in an accessor to a field), I would agree with you, those don't need the word `Get`. But this method is actually going and getting something for somewhere else. And I believe representing that in the method name gives context to the caller about the possible cost of calling this multiple times (vs calling an accessor multiple times, which is mostly free).

About the `Bundles` part, I see what you are saying. However, when I read `bundlesclient` I'm not completely sure if that client can only return a bundle or it might be able to return other things. That in addition with the absence of the word `Get` (or any other synonym) create ambiguity. `bundleclient.ActiveOrLatest` could very much return the active or latest client.

I'm all in favor of not using redundant names and try to keep them short, that makes the code easier to read. But in this case, IMO, we are taking it a bit to far and actually making the code more difficult to read.

pkg/curatedpackages/bundleclient.go
g-gaston: > I could make the constructor private I guess... But it doesn't buy me anything, since NewBundleClient calls it anyway, and that has to be exported...
There is a mayor difference: `NewBundleClient` return a public interface. So from there you don't return a private type, you return a public interface implemented by a private struct.

I just don't think it's a good idea to return private types from public methods in general. You leave the caller without information about what they can do with that return value or even what that value actually is. You would need to go a read the actual implementation of that type to know what public methods has and what they do.

pkg/curatedpackages/bundleclient.go
g-gaston: Maybe I need more context to see how this is going to be used by consumers.
Should we table this conversation until we move to those PRs? As long as you are OK revisiting this discussion when the time comes.

pkg/curatedpackages/bundleclient.go
g-gaston: Cool. Valuable info in case we ever want to move more parts of the CLI to a golang client.

pkg/curatedpackages/bundleclient.go
ewollesen: I'm willing to re-open it later, but the context you need is in the code today. Look at the commands for list packages, generate package, or install package.

pkg/curatedpackages/bundleclient.go
ewollesen: I'll make it private. If you get a struct that you don't know what it is, or how to use it, I consider that a strong hint that you shouldn't use that function is all. *shrug*

pkg/curatedpackages/bundleclient.go
ewollesen: No, in this case we don't need it. As I said, it helps me communicate intent, and identify problems faster. I'll remove it though.

PR #3607, opened 2022-10-07
feat(Makefile): run unit tests for only modified packages

ewollesen:
Two new Makefile targets:

    1. unit-test-patch

        Does a little git magic to figure out from the files that've changed
	from upstream which go packages have had changes, and then runs only
	those unit tests. This is convenient when running the tests
	frequently, as in red-green development or for careful refactoring.

    2. coverage-view-patch

        Like unit-test-patch, but generates and opens in your browser a
	coverage report for your changes. This is different from "make
	coverage-view" in that it only generates reports for files in the
	packages that are modified from upstream. What that means is that you
	have far fewer files to sort through to find useful test coverage
	information in the report, as it's likely to include only a handful of
	files, as opposed to every source file in the repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -391,7 +391,7 @@ test: unit-test capd-test  ## Run unit and capd tests
 .PHONY: unit-test
 unit-test: ## Run unit tests
 unit-test: $(SETUP_ENVTEST) 
-unit-test: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path --arch amd64 $(KUBEBUILDER_ENVTEST_KUBERNETES_VERSION))
+unit-test: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path --arch $(GO_ARCH) $(KUBEBUILDER_ENVTEST_KUBERNETES_VERSION))
 unit-test:
 	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" $(GO_TEST) $$($(GO) list ./... | grep -vE "$(UNIT_TEST_PACKAGE_EXCLUSION_REGEX)") -cover -tags "$(BUILD_TAGS)" $(GO_TEST_FLAGS)
 
--- a/Makefile
+++ b/Makefile
@@ -395,6 +395,22 @@ unit-test: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path 
 unit-test:
 	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" $(GO_TEST) $$($(GO) list ./... | grep -vE "$(UNIT_TEST_PACKAGE_EXCLUSION_REGEX)") -cover -tags "$(BUILD_TAGS)" $(GO_TEST_FLAGS)
 
+
+# unit-test-patch is a convenience target that restricts test runs to modified
+# packages' tests. This is not a replacement for running the unit-test target,
+# but can be useful for red-green development or targeted refactoring.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -416,6 +425,15 @@ coverage-html: coverage-unit-test
 coverage-view: coverage-unit-test
 	$(GO) tool cover -html=$(COVER_PROFILE)
 
+# coverage-view-patch generates an HTML file limited to changes and opens it in your browser.
+.PHONY: coverage-view-patch
+coverage-view-patch: GO_PKGS ?= $(shell ./scripts/go-packages-in-patch.sh)
+coverage-view-patch: $(SETUP_ENVTEST)
+coverage-view-patch: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path --arch amd64 $(KUBEBUILDER_ENVTEST_KUBERNETES_VERSION))
```
TerryHowe, line 432: Aren't new people getting M1s? Is there anything here that detects arch?

chrisdoherty4: We might be able to get away with `GOARCH`, arch naming is pretty universal.

chrisdoherty4:
This is really awesome, great utility for a quick sanity check. I wonder if we might find somewhere to note (preferably as close to the target as possible) the trade-offs. Namely the fact that dependencies of the changed code _could_ be fail if behavioral expectations have been tweaked. In the same spot, we might want to establish the exepcted use also that you detailed in the description. 

> This is convenient when running the tests frequently, as in red-green development or for careful refactoring.

Makefile
ewollesen: Good suggestions both, I hadn't noticed that, it was a cut and paste error. I'll address it.

ewollesen:
> This is really awesome, great utility for a quick sanity check. I wonder if we might find somewhere to note (preferably as close to the target as possible) the trade-offs. Namely the fact that dependencies of the changed code _could_ have their tests fail if behavioral expectations have been tweaked.
> 
> Ideally that won't happen if unit tests on the changed code are sufficient, but we live in the real world. Hopefully a note will help make clear this is an _additional_ tool to `make test`, not a replacement.
> 
> In the same spot, we might want to establish the exepcted use also that you detailed in the description.
> 
> > This is convenient when running the tests frequently, as in red-green development or for careful refactoring.

Great thoughts, I'll add something.

Makefile
```diff
@@ -417,6 +433,21 @@ coverage-html: coverage-unit-test
...
+# coverage-view-patch opens a Go HTML coverage report limited to modified
+# packages. It builds on the unit-test-patch target. Because it covers only
+# modified packages, the list of files in the report is more manageable.
+# BE WARNED: a passing coverage-view-patch does not imply that unit-test will
+# pass. Unmodified packages that depend on changed behaviors will not have
+# their tests run by this target.
+.PHONY: coverage-view-patch
+coverage-view-patch: GO_PKGS ?= $(shell ./scripts/go-packages-in-patch.sh)
+coverage-view-patch: $(SETUP_ENVTEST)
+coverage-view-patch: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path --arch $(GO_VIEW) $(KUBEBUILDER_ENVTEST_KUBERNETES_VERSION))
```
TerryHowe, line 445:
```suggestion
coverage-view-patch: KUBEBUILDER_ASSETS ?= $(shell $(SETUP_ENVTEST) use --use-env -p path --arch $(GO_ARCH) $(KUBEBUILDER_ENVTEST_KUBERNETES_VERSION))
```

chrisdoherty4 (approved):

Makefile
TerryHowe: not sure why this was GO_VIEW

PR #3608, opened 2022-10-07
Allow users to skip deploying `kube-vip` load balancer for tinkerbell stack

abhinavmpandey08:
*Description of changes:*
Currently, EKS-A for bare-metal deploys `kube-vip` loadBalancer on the target cluster to expose the Tinkerbell services externally.
This change allows users to skip deploying `kube-vip` by default as they may want to use a different loadBalancer post cluster creation to expose tinkerbell stack. This is something users will have to do post cluster creation to ensure they have an HA cluster.

A new field `skipLoadBalancerDeployment` is added to the `TinkerbellDatacenterConfig` to allow users to achieve this
```yaml
kind: TinkerbellDatacenterConfig
metadata:
  name: abhnvp
spec:
  tinkerbellIP: "1.2.3.4"
  skipLoadBalancerDeployment: true
```

The default behavior is still unchanged. When you run the generate command, this field won't show up in the clusterconfig. Instead, users will have to manually add it as a precaution so they don't accidentally set this field unless they absolutely need to.


abhinavmpandey08:
/hold
this PR depends on https://github.com/aws/eks-anywhere-build-tooling/pull/1209
The above PRs to be merged and the chart needs to be updated before merging this

pkg/providers/tinkerbell/create.go
```diff
@@ -80,15 +80,20 @@ func (p *Provider) PostBootstrapSetup(ctx context.Context, clusterConfig *v1alph
...
 		ctx,
 		clusterSpec.VersionsBundle.Tinkerbell,
 		p.templateBuilder.datacenterSpec.TinkerbellIP,
 		cluster.KubeconfigFile,
 		p.datacenterConfig.Spec.HookImagesURLPath,
 		stack.WithBootsOnKubernetes(),
-		stack.WithHostPortEnabled(false),
-		stack.WithLoadBalancer(),
+		stack.WithHostPortEnabled(false), // disable host port on workload cluster
+		stack.WithEnvoyEnabled(true), // use envoy on workload cluster
```
ptrivedi, line 95: so envoy installation is not optional, only kube-vip is? that should be okay for now... not sure if for single node resource conservation concerns, at some point we will need to make envoy optional

config/crd/bases/anywhere.eks.amazonaws.com_tinkerbelldatacenterconfigs.yaml
```diff
@@ -35,21 +35,25 @@ spec:
...
                 description: HookImagesURLPath can be used to override the default
-                  Hook images path to pull from a local server
+                  Hook images path to pull from a local server.
                 type: string
               osImageURL:
                 description: OSImageURL can be used to override the default OS image
-                  path to pull from a local server
+                  path to pull from a local server.
                 type: string
+              skipLBSetup:
```
ptrivedi, line 49: instead of acronym 'LB', 'LoadBalancer' would be clearer when looking at cluster config

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -184,15 +191,13 @@ func (s *Installer) Install(ctx context.Context, bundle releasev1alpha1.Tinkerbe
 			deploy: true,
 			image:  bundle.TinkerbellStack.Rufio.URI,
 		},
-		loadBalancer: map[string]interface{}{
-			"enabled": s.loadBalancer,
-			"ip":      tinkerbellIP,
-		},
 		kubevip: map[string]interface{}{
```
ptrivedi, line 194: why were loadBalancer and kubevip separate over here before?

ptrivedi:
linter issues and nits, looks good otherwise

pkg/providers/tinkerbell/create.go
abhinavmpandey08: I would recommend keeping envoy even for single node. It has a really small footprint and prevents us from having to expose two load balancer services on the same IP, which some LBs might not support

config/crd/bases/anywhere.eks.amazonaws.com_tinkerbelldatacenterconfigs.yaml
abhinavmpandey08: done

pkg/providers/tinkerbell/stack/stack.go
abhinavmpandey08: previously, the `loadBalancer` field was being used to manage both `kube-vip` and `envoy` deployments.
Now that we have separate flags to control them separately, the `loadBalancer` field isn't needed anymore

pkg/providers/tinkerbell/create.go
```diff
@@ -80,15 +80,20 @@ func (p *Provider) PostBootstrapSetup(ctx context.Context, clusterConfig *v1alph
 func (p *Provider) PostWorkloadInit(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
 	logger.V(4).Info("Installing Tinkerbell stack on workload cluster")
 
+	if p.datacenterConfig.Spec.SkipLoadBalancerDeployment {
+		logger.Info("Warning: Skipping load balancer deployment. Please install and configure a load balancer once the cluster is created.")
+	}
+
```
ptrivedi, line 86: Can you cover these with a unit test?

pkg/providers/tinkerbell/stack/stack.go
ptrivedi: tinkerbellIP should be made optional then

PR #3615, opened 2022-10-08
Upgrade github.com/tinkerbell/{tink,rufio} dependencies:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This is needed for VLAN ID support as the latest upstream CAPT uses these same versions.

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/collectors.go
+++ b/pkg/diagnostics/collectors.go
@@ -317,9 +317,9 @@ func (c *collectorFactory) snowCrdCollectors() []*Collect {
 
 func (c *collectorFactory) tinkerbellCrdCollectors() []*Collect {
 	captCrds := []string{
-		"baseboardmanagements.bmc.tinkerbell.org",
-		"bmcjobs.bmc.tinkerbell.org",
-		"bmctasks.bmc.tinkerbell.org",
+		"machines.bmc.tinkerbell.org",
+		"jobs.bmc.tinkerbell.org",
+		"tasks.bmc.tinkerbell.org",
 		"hardware.tinkerbell.org",
 		"templates.tinkerbell.org",
 		"tinkerbellclusters.infrastructure.cluster.x-k8s.io",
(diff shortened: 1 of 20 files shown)


pkg/diagnostics/collectors.go
```diff
@@ -317,9 +317,9 @@ func (c *collectorFactory) snowCrdCollectors() []*Collect {
 
 func (c *collectorFactory) tinkerbellCrdCollectors() []*Collect {
 	captCrds := []string{
-		"baseboardmanagements.bmc.tinkerbell.org",
-		"bmcjobs.bmc.tinkerbell.org",
-		"bmctasks.bmc.tinkerbell.org",
+		"machines.bmc.tinkerbell.org",
```
abhinavmpandey08, line 320: Did we rename this CRD to `machines` upstream?
CAPI already has a CRD named `machines` so this is gonna be ambiguous

abhinavmpandey08 (commented):
Will this break compatibility with older EKS-A versions since we are re-naming the CRDs?

Also, is build-tooling updated to accommodate these changes?

pkg/diagnostics/collectors.go
jacobweinstock: we did. https://github.com/tinkerbell/rufio/blob/main/config/crd/bases/bmc.tinkerbell.org_machines.yaml
It's namespaced though.

ptrivedi: I agree that a CRD named `machines` is going to be confusing and ambiguous

chrisdoherty4: Custom resources are namespaced so there aren't any collisions from a technical point of view; I can appreciate humans may do a double take. There are numerous projects that define the same kind across different groups (namespaces). For example, [argo-cd](https://doc.crds.dev/github.com/argoproj/argo-cd) and [litmus](https://doc.crds.dev/github.com/litmuschaos/litmus) define `Workflow` and `CronWorkflow` types. [Crossplane](https://doc.crds.dev/github.com/crossplane/crossplane) defines 2 `Configuration` resources in the same project but under different namespaces. There are many other examples.

Its infeasible, and defeats the point of namespacing, to expect projects to need to create unique names - not to mention this is upstream so can't be changed here.

chrisdoherty4:
Completed as a separate piece of work.

PR #3616, opened 2022-10-09
Add CAPI provider Workers

g-gaston:
*Description of changes:*
This allows to represent a provider specific CAPI workers for both. It implements logic to generate new names for immutable objects when these change. It does this by comparing the new spec to the current state of the objects in the cluster.

This is part of a bigger effort, allowing a transition from yaml go templates to go structs. For context: https://github.com/aws/eks-anywhere/pull/3432

*Testing (if applicable):*
Unit tests

### DIFF
--- a/pkg/clusterapi/controlplane.go
+++ b/pkg/clusterapi/controlplane.go
@@ -12,20 +12,6 @@ import (
 	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
 )
 
-// Object represents a kubernetes API object
-type Object interface {
-	kubernetes.Object
-}
-
-// ObjectComparator returns true only if only both kubernetes Object's are identical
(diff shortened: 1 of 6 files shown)


pkg/clusterapi/controlplane.go
```diff
@@ -104,39 +90,3 @@ func (cp *ControlPlane[C, M]) UpdateImmutableObjectNames(
 
 	return nil
 }
-
-// EnsureNewNameIfChanged updates an object's name if such object is different from its current state in the cluster
-func EnsureNewNameIfChanged[M Object](ctx context.Context,
```
g-gaston, line 109: This hasn't changed at all, I'm just moving the code to a different file
Same with the tests

taneyland (approved):
awesome!

PR #3619, opened 2022-10-10
Implement storage class installation hook

chrisdoherty4:
### Description

Dependencies: https://github.com/aws/eks-anywhere/pull/3476

This change removes the `GenerateStorageClass()` from the provider interface and shifts the responsibility of storage class installation to providers. The only provider performing storage class installation is vSphere. A new workflow hook has been written using the same install method that the cluster manager now tests for and calls ensuring existing code paths are unaffected.

#### Rationale for anonymous interfaces

The implementation in the cluster manager uses an anonymous interface type assertion to determine if the provider has the necessary behavior for storage class installation. It was written this way acknowledging that the cluster managers storage class installation orchestration will be deleted when we switch to new workflows as the storage class will be opt-in from the vSphere provider only via a workflow hook.

To ensure the vSphere provider still honors existing behavior a unit test verifying the storage class installation behavior is still available with commentary on the cluster managers current usage.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -167,6 +167,13 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 			CreateBootstrapClusterOptions: deps.Provider,
 		}
 		wflw.WithHookRegistrar(awsiamauth.NewHookRegistrar(deps.AwsIamAuth, clusterSpec))
+
+		// Not all provider implementations want to bind hooks so we explicitly check if they
+		// want to bind hooks before registering it.
+		if registrar, ok := deps.Provider.(management.CreateClusterHookRegistrar); ok {
+			wflw.WithHookRegistrar(registrar)
+		}
+
 		err = wflw.Run(ctx)
 	} else {
 		err = createCluster.Run(ctx, clusterSpec, createValidations, cc.forceClean)
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -593,15 +593,21 @@ func getProviderNamespaces(providerDeployments map[string][]string) []string {
 }
 
 func (c *ClusterManager) InstallStorageClass(ctx context.Context, cluster *types.Cluster, provider providers.Provider) error {
-	storageClass := provider.GenerateStorageClass()
-	if storageClass == nil {
-		return nil
-	}
-
-	logger.Info("Installing storage class on cluster")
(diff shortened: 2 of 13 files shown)


cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -167,6 +167,13 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 			CreateBootstrapClusterOptions: deps.Provider,
 		}
 		wflw.WithHookRegistrar(awsiamauth.NewHookRegistrar(deps.AwsIamAuth, clusterSpec))
+
+		// Not all provider implementations want to bind hooks so we explicitly check if they
+		// want to bind hooks before registering it.
+		if registrar, ok := deps.Provider.(management.CreateClusterHookRegistrar); ok {
```
g-gaston, line 173: Is this to avoid the no-op methods in providers?

pkg/clustermanager/cluster_manager.go
```diff
@@ -592,15 +592,21 @@ func getProviderNamespaces(providerDeployments map[string][]string) []string {
...
-		return fmt.Errorf("applying storage class manifest: %v", err)
+	// Historically, vSphere has been the only provider wanting to install a storage class. The new
+	// workflow hook capability enables inverting the provider relationship so only the vSphere
+	// provider contains storage class installation code.
+	//
+	// To maintain backward compatibility, we're checking for an anonymous interface implemented
+	// on the vSphere provider only to determine if the provider wants to install a storage class.
+	// This code should be deleted when we convert completely to new workflows.
+	installer, ok := provider.(interface {
+		InstallStorageClass(context.Context, *types.Cluster) error
```
g-gaston, line 603: nit:
I would vote to avoid doing this and leave the provider cleanup for when we completely transition to the new workflows

g-gaston (approved):

pkg/clustermanager/cluster_manager.go
chrisdoherty4: Have you any particular rationale? 

Currently, this is wrapped in tests and I think its easier to see "whats left" if we gradually reduce the interface where it makes sense.

cmd/eksctl-anywhere/cmd/createcluster.go
chrisdoherty4: Yes.

pkg/clustermanager/cluster_manager.go
chrisdoherty4: We talked. There's no compelling reasoning not to do this - its well protected with tests and drives towards our end goal.

PR #3641, opened 2022-10-11
Refactor bootstrap cluster tasks

chrisdoherty4:


pkg/workflow/task/bootstrap/bootstrap.go
```diff
@@ -2,37 +2,31 @@ package bootstrap
...
 )
 
-// OptionsRetriever supplies bootstrap cluster options. This is typically satisfied
-// by a provider.
-type OptionsRetriever interface {
-	BootstrapClusterOpts(*cluster.Spec) ([]bootstrapper.BootstrapClusterOption, error)
+// BootstrapCluster is an instance of a bootstrap cluster.
+type BootstrapCluster interface {
+	// CreateCluster creates the bootstrap cluster instance.
+	CreateCluster(context.Context) error
```
g-gaston, line 17: nit: maybe just `Create`? Or something like `cluster.SpinUp`?
Same for delete. The cluster part seems a bit redundant given this is a cluster.

g-gaston (commented):
lgtm

pkg/workflow/task/bootstrap/bootstrap.go
chrisdoherty4: We could do create. It felt like a bit of a weak interface doing that though. That said, I'm not sure `*Cluster` strengthens the interface 🤦🏻

PR #3644, opened 2022-10-11
Feature remote management

a-cool-train:
*Description of changes:*
Please reference this documentation for the changes in this PR: https://github.com/aws/eks-anywhere-packages/blob/main/docs/design/remote-management.md. 
This PR does the following:
1. During controller installation, check whether the current cluster is workload cluster and if so, only install controller resources instead of the controller itself.
2. Remove cert manager verification since the controller would not be installed on a workload cluster. Please reference the documentation above.
3. When installing curated packages on workload cluster, use management cluster kubeconfig instead of current cluster's kubeconfig.
*Testing (if applicable):*


TerryHowe (approved):
/lgtm

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -111,6 +120,30 @@ func (pc *PackageControllerClient) InstallController(ctx context.Context) error
...
+// InstallPBCResources installs Curated Packages Bundle Controller Custom Resource
+// This method is used only for Workload clusters
+// Please refer to this documentation: https://github.com/aws/eks-anywhere-packages/blob/main/docs/design/remote-management.md
+func (pc *PackageControllerClient) InstallPBCResources(ctx context.Context) error {
+	templateValues := map[string]string{
+		"clusterName": pc.clusterName,
+		"namespace":   constants.EksaPackagesName,
+	}
+
+	result, err := templater.Execute(packageBundleContrllerYaml, templateValues)
```
g-gaston, line 132: Instead of using go templates, can we build the go struct and marshal it into yaml?
Makes the code easier to read and to extend :)

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -74,6 +77,12 @@ func NewPackageControllerClient(chartInstaller ChartInstaller, kubectl KubectlRu
 //   - credentials refreshing cron job creation
 //   - activation of a curated packages bundle
 func (pc *PackageControllerClient) InstallController(ctx context.Context) error {
```
g-gaston, line 79: We need to update the comment above
Also should we rename this method? Like `EnableCuratedPackages` (avoiding using `install` to remove confusion with installing the actual packages). It doesn't just install the controller, now it depends on what type of cluster

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: I think that would be an interesting thought but in this case both formats should be readable and I could argue that the go templates in this case is more applicable and makes it easy to read. For extendability, if there is a need then yes we can make the change but in this case I think the templating should work well.

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: Yes I like EnableCuratedPackages

TerryHowe (approved):

pkg/curatedpackages/packagecontrollerclient.go
a-cool-train: Ok if we are trying to get away from that, we also have awssecret.yaml which is in the same line as this file. Let me revisit both in a later PR.

PR #3648, opened 2022-10-11
Add CNI install to create workload task

g-gaston:
## Description of changes

CNI installers are now stateful, being constructed with all configuration injected at build time. Only dynamic objects like the context or the cluster authentication are provided as runtime method argument. They all build on top of a stateless version, intended for other usecases where the cluster spec doesn't exist at runtime.

Part of the refactor to move to the new workflow paradigm.

## Testing

Unit tests


internal/test/files.go
```diff
@@ -37,29 +38,50 @@ func AssertFilesEquals(t *testing.T, gotPath, wantPath string) {
 
 func AssertContentToFile(t *testing.T, gotContent, wantFile string) {
```
g-gaston, line 39: This should be mostly the same, I just extracted the `diff` part to a separate method to reuse it

pkg/dependencies/factory.go
```diff
@@ -636,8 +642,39 @@ func (f *Factory) WithNetworking(clusterConfig *v1alpha1.Cluster) *Factory {
...
+
+		return nil
+	})
+
+	return f
+}
+
 func (f *Factory) WithCiliumTemplater() *Factory {
-	f.WithHelm()
+	f.WithHelm(executables.WithInsecure())
```
g-gaston, line 677: I'm not sure what `f.WithHelm(executables.WithInsecure())` does, but it was being used before to build the the helm passed to `Cilium`

pkg/dependencies/factory.go
```diff
@@ -636,8 +642,39 @@ func (f *Factory) WithNetworking(clusterConfig *v1alpha1.Cluster) *Factory {
 	return f
 }
 
+func (f *Factory) WithCNIInstaller(spec *cluster.Spec, provider providers.Provider) *Factory {
```
g-gaston, line 645: It's unfortunate that we need to pass the provider, which makes dep construction a two step process.

We only need the namespaces from it. My suggestion would be to decouple this data retrieval from the provider (it's static and known at build time). This would mean having another switch for the datacenter type, but I think it's worth it. The questions would be, where do we put it. The `providers` package would be ideal, but that would create a circular dependency if we try to import the provider specific packages from there.

pkg/networking/cilium/cilium.go
```diff
@@ -1,24 +1,26 @@
...
-func NewCilium(client Client, helm Helm) *Cilium {
+func NewCilium(client KubernetesClient, templater InstallUpgradeTemplater) *Cilium {
 	return &Cilium{
-		Upgrader: NewUpgrader(client, helm),
+		Installer: NewInstaller(client, templater),
+		Upgrader:  NewUpgrader(client, templater),
 	}
 }
-
-func (c *Cilium) GenerateManifest(ctx context.Context, clusterSpec *cluster.Spec, providerNamespaces []string) ([]byte, error) {
```
g-gaston, line 22: Moved this to the upgrader, where it makes more sense

pkg/networking/cilium/client.go
```diff
@@ -25,43 +25,44 @@ type Client interface {
 	RolloutRestartDaemonSet(ctx context.Context, name, namespace, kubeconfig string) error
 }
 
-type retrierClient struct {
+// RetrierClient wraps basic kubernetes API operations around a retrier
+type RetrierClient struct {
```
g-gaston, line 29: I just made this public to not have tests in the `cilium` package. There were circular imports because of the mocks package

pkg/providers/vsphere/reconciler/reconciler.go
```diff
@@ -204,42 +202,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *an
 		}}, err
 	}
 
-	if result, err := r.reconcileCNI(ctx, cluster, capiCluster, specWithBundles, log); err != nil {
```
g-gaston, line 207: Deleted all this code because it's not used and it was building cilium in line

pkg/workflow/task/workload/create.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/workflow/workflowcontext"
+)
+
+type CNIInstaller interface {
+	// Install configures a CNI for the first time in a kubernetes cluster
+	Install(ctx context.Context, cluster *types.Cluster) error
+}
+
+// Create spins up a conformant cluster based on a EKS-A spec using CAPI
```
chrisdoherty4, line 16:
```suggestion
// Create creates a Kubernetes conformant cluster.
```

pkg/networking/cilium/installer.go
```diff
@@ -0,0 +1,63 @@
...
+
+func NewInstaller(client KubernetesClient, templater InstallTemplater) *Installer {
+	return &Installer{
+		templater: templater,
+		k8s:       client,
+	}
+}
+
+// Installs configures Cilium in an EKS-A cluster
+func (i *Installer) Install(ctx context.Context, cluster *types.Cluster, spec *cluster.Spec, allowedNamespaces []string) error {
```
chrisdoherty4, line 49: Is this behavior needed by anything? If not would it make sense to have a single `Installer` type exporting the `Install` behavior expected by `workflow` constructs and an unexported `install` (this behavior) reducing the API surface area?

E.g.
```
type Installerr struct {
	Templater
	allowedNamespaces []string
	spec              *cluster.Spec
	k8s               KubernetesClient
}

func (i Installerr) Install(ctx context.Context, cluster *types.Cluster) error {
	// Similar impl
}
```

chrisdoherty4 (commented):
The integration with the workflows looks clean. I'm just curious if we can simplify the CNI implementations given there seems to be degrees of data structure nesting. I don't think the PR needs holding but it might be useful spending a small amount of time on it.

pkg/networking/cilium/installer.go
g-gaston: The current workflow uses this (through cluster manager)

chrisdoherty4 (approved):

chrisdoherty4 (approved):

PR #3650, opened 2022-10-11
add vpshere tags simple one-pager design

gwesterfieldjr:
add vpshere tags simple one-pager design

### DIFF
--- a/designs/vcenter-tags.md
+++ b/designs/vcenter-tags.md
@@ -1,0 +1,50 @@
+# Support Tagging VM's in a vSphere cluster
+
+## Introduction
+
+**Problem:** User's have no way to Tag VM's provisioned by EKS-A in vSphere. Tags are needed by customers for various reasons.
+A specific use-case is for NSX. NSX polices & rules can be configured based on tags on a VM. Customers may control network traffic based on these tags. Therefore, we need to support adding tags at deployment time.
+
+### Goals and Objectives
+
+As an EKS Anywhere user:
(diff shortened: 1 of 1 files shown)


designs/vcenter-tags.md
```diff
@@ -0,0 +1,50 @@
...
+kind: VSphereMachineConfig
+metadata:
+   name: my-cluster-machines
+spec:
+  tagIDs:
+  - urn:vmomi:InventoryServiceTag:8e0ce079-0675-47d6-8665-16ada4e6dabd:GLOBAL
+  diskGiB: 25
+  datastore: "Datastore"
+  folder: "Folder"
+...
```
TerryHowe, line 45: I think what you propose here is probably the best solution for consistency, but I think `tags` would have been nicer than `tagIDs`

gwesterfieldjr: I like this. I will make it tags.

gwesterfieldjr: Done.

ndeksa:
how about the upgrade/migration, will that covered with the ? also, am assuming this field is optional?

vignesh-goutham (approved):
/approve

vignesh-goutham:
> how about the upgrade/migration, will that covered with the ? also, am assuming this field is optional?

IIRC, if not called out/coded explicitly, the field is immutable. So any upgrades from /to using tags will be covered.

PR #3654, opened 2022-10-12
Bugfix: Ensure `nutanixInsecure` is always set

thunderboltsid:
This is to ensure that the underlying yaml manifests get populated with the correct true/false value regardless of whether a trust bundle is present or not. Earlier the manifest would show up as `no data` in case of the trust bundle was not provided.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #3656, opened 2022-10-12
Enable linters: cyclo comp, docs and naked returns

g-gaston:
### Rules being enforced
* All exported package members need a doc comment
* Doc comments should start with capital letter and end with a period
* Minimum permissions for new created directories and files
* No naked returns
* Avoid method with high cyclomatic complexity. We are starting with a limit of 10, which is the most strict config in the recommended range 10-20. We’ll start here and we’ll make it less restrictive as we go if we see need.

### Linters
* `revive`: replacement for the deprecated golint. Enabled 3 issues that are disabled by default for exported member comments and file/dir permissions. We used to run `golint`, but then it got deprecated and dropped from golangci-lint. This means we have a bunch of issues in the code we wrote in between then and now. We should try to fix those over time.
* godot: checks that exported member comments start with upper case and end with a period (".").
* `nakedret`: forbids naked returns.
* `gocyclo`: measures cyclomatic complexity of functions. Restricted to a max of 10. Recommended is 10-20. This requisite can be lowered down if we find it to restrictive.

Now the linters only run in the diff between the current code and main. This means it will only trigger issues for infringent new code. We should strive to fix all old ocurrences if possible and add exceptions for the rest. so we can renable a full repo scan.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,6 +10,10 @@ linters:
   enable:
     - gofumpt
     - gci
+    - godot
+    - nakedret
+    - gocyclo
+    - revive
 linters-settings:
   gci:
     sections:
@@ -18,6 +22,22 @@ linters-settings:
       - prefix(github.com/aws/eks-anywhere)
     custom-order: true
     skip-generated: false
+  godot:
+    exclude:
+      # Exclude todo comments.
+      - "(?i)^\\s*todo"
+    period: true
+    capital: true
+  nakedret:
+    # never allow naked returns
+    max-func-lines: 0
+  gocyclo:
+    # Minimal code complexity to report.
+    min-complexity: 10
 issues:
   max-same-issues: 0
   max-issues-per-linter: 0
+  include:
+    - EXC0012  # EXC0012 revive: exported (.+) should have comment( \(or a comment on this block\))? or be unexported
+    - EXC0014  # EXC0014 revive: comment on exported (.+) should be of the form "(.+)..."
+    - EXC0009  # EXC0009 revive: (Expect directory permissions to be 0750 or less|Expect file permissions to be 0600 or less)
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -18,6 +22,22 @@ linters-settings:
...
+      # Exclude todo comments.
+      - "^todo:"
+    period: true
+    capital: true
+  nakedret:
+    # never allow naked returns
+    max-func-lines: 0
+  gocyclo:
+    # Minimal code complexity to report.
+    min-complexity: 10
```
chrisdoherty4, line 36: Curious if you did any testing against this? I haven't evaluated the algorithm for gocyclo but Go typically suffers a little because of how expressive it is and consequently you often need to bump by 1 or 2 points.

g-gaston: Nop, I didn't. 10-20 is the recommended value by gocyclo, so I opted for started with the lowest limit and go from there. Making it less restrictive is easier than making it more restrictive.

.golangci.yml
```diff
@@ -18,6 +22,22 @@ linters-settings:
       - prefix(github.com/aws/eks-anywhere)
     custom-order: true
     skip-generated: false
+  godot:
+    exclude:
+      # Exclude todo comments.
+      - "^todo:"
```
chrisdoherty4, line 28: Is this regex? Wanna make it case insensitive?

g-gaston: Good idea, done

.golangci.yml
```diff
@@ -18,6 +22,22 @@ linters-settings:
       - prefix(github.com/aws/eks-anywhere)
     custom-order: true
     skip-generated: false
+  godot:
+    exclude:
+      # Exclude todo comments.
+      - "(?i)^\\s*todo:"
```
chrisdoherty4, line 28: We're still in draft sooo, perhaps drop the `:`. I usually add a BOM to my todo comments so I can `grep` easily in the future.

e.g.

`TODO(chrisdoherty)`

chrisdoherty4 (approved):

PR #3662, opened 2022-10-12
fixed kube-vip yaml for nutanix provider

deepakm-ntnx:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


abhay-krishna:
@deepakm-ntnx The release tooling presubmit is failing because the bundle generation code for Nutanix ([bundles/nutanix.go](https://github.com/aws/eks-anywhere/blob/main/release/pkg/bundles/nutanix.go)) doesn't have the Kubevip field in the Nutanix struct, but that section is there in the golden file. So there's a mismatch when regenerating the bundle and checking it against the golden file. You just need to add the kube-vip field in nutanix.go like we do for vsphere(Lines [33](https://github.com/aws/eks-anywhere/blob/main/release/pkg/bundles/vsphere.go#L33) and [116](https://github.com/aws/eks-anywhere/blob/main/release/pkg/bundles/vsphere.go#L116))

abhinavmpandey08 (approved):
/approve
/woof

PR #3663, opened 2022-10-12
Make audit policy apiVersion dynamic based on k8s version

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Make audit policy apiVersion dynamic based on k8s version

*Testing (if applicable):*


pkg/providers/common/auditpolicy.go
```diff
@@ -0,0 +1,300 @@
+package common
```
chrisdoherty4, line 1: Could we pull this out into its own package? Perhaps something already exists that's appropriate?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -804,6 +803,11 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec, controlPlaneMachineSpec, etcd
 		values["awsIamAuth"] = true
 	}
 
+	auditPolicy, err := common.GetAuditPolicyYaml(clusterSpec.Cluster.Spec.KubernetesVersion)
```
maxdrib, line 806: There will be some unit test failures here, since we assert on yaml files which are likely to change if we move to a serialized go struct instead of the previous template. Also, I suspect that change itself will trigger a rolling upgrade on the workload clusters during a management cluster upgrade still, which was what I thought we were trying to avoid with this PR.

mitalipaygude: Yes, we are planning on now keeping the same yaml for versions before 1.23

mitalipaygude: Check it out here https://github.com/aws/eks-anywhere/pull/3672

mitalipaygude:
Follow up https://github.com/aws/eks-anywhere/pull/3672

PR #3667, opened 2022-10-13
Handle kube-vip deployment in single-node cluster

d8660091:
*Issue #, if available:* Services can be exposed as LoadBalancer type supported by kube-vip. But in single-node cluster, kube-vip daemonset cannot be deployed correctly as it has to run in controlplane nodes. So we need this adjustment, to let the kube-vip daemonset of cp nodes to handle LoadBalancer services.

*Description of changes:*

*Testing (if applicable):* 
1. Unit test
2. Manual tested SkipLoadBalancerDeployment on single-node cluster, when the field is set to false/true

### DIFF
--- a/pkg/providers/tinkerbell/create.go
+++ b/pkg/providers/tinkerbell/create.go
@@ -95,7 +95,9 @@ func (p *Provider) PostWorkloadInit(ctx context.Context, cluster *types.Cluster,
 		stack.WithBootsOnKubernetes(),
 		stack.WithHostPortEnabled(false), // disable host port on workload cluster
 		stack.WithEnvoyEnabled(true),     // use envoy on workload cluster
-		stack.WithLoadBalancerEnabled(!p.datacenterConfig.Spec.SkipLoadBalancerDeployment), // configure load balancer based on datacenterConfig.Spec.SkipLoadBalancerDeployment
+		stack.WithLoadBalancerEnabled(
+			len(clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations) != 0 && // load balancer is handled by kube-vip in control plane nodes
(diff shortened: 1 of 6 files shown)


pkg/providers/tinkerbell/create.go
```diff
@@ -95,7 +95,9 @@ func (p *Provider) PostWorkloadInit(ctx context.Context, cluster *types.Cluster,
 		stack.WithBootsOnKubernetes(),
 		stack.WithHostPortEnabled(false), // disable host port on workload cluster
 		stack.WithEnvoyEnabled(true),     // use envoy on workload cluster
-		stack.WithLoadBalancerEnabled(!p.datacenterConfig.Spec.SkipLoadBalancerDeployment), // configure load balancer based on datacenterConfig.Spec.SkipLoadBalancerDeployment
+		stack.WithLoadBalancerEnabled(
+			len(clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations) != 0 && // load balancer is handled by kube-vip in control plane nodes
+				!p.datacenterConfig.Spec.SkipLoadBalancerDeployment), // configure load balancer based on datacenterConfig.Spec.SkipLoadBalancerDeployment
```
cxbrowne1207, line 100: Can we assign this to a variable?

cxbrowne1207 (commented):
One code refactor suggestion

pkg/providers/tinkerbell/create.go
d8660091: We can, but it's a little subjective. My opinion is that there is not much need for creating a separate variable if the variable would be used only once.

cxbrowne1207: Okay, that's good

PR #3668, opened 2022-10-13
Invalid hardware.csv message

tatlat:
*Issue #, if available:* 

*Description of changes:*
When running `eksctl anywhere create cluster --hardware-csv hardware.csv -f eksa-single-node-cluster.yaml -v` and there is an issue with hardware.csv, it will now tell you that "hardware.csv is not correct" along with the specific problem whereas before it would just say the problem.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/hardware/translate.go
+++ b/pkg/providers/tinkerbell/hardware/translate.go
@@ -45,7 +45,7 @@ func Translate(reader MachineReader, writer MachineWriter, validator MachineVali
 	}
 
 	if err != nil {
-		return fmt.Errorf("read: %v", err)
+		return fmt.Errorf("hardware.csv is not correct: read: %v", err)
 	}
 
 	if err := validator.Validate(machine); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/providers/tinkerbell/hardware/translate.go
```diff
@@ -45,7 +45,7 @@ func Translate(reader MachineReader, writer MachineWriter, validator MachineVali
 	}
 
 	if err != nil {
-		return fmt.Errorf("read: %v", err)
+		return fmt.Errorf("hardware.csv is not correct: read: %v", err)
```
chrisdoherty4, line 48: Its not appropriate to talk about a CSV file at this point as this particular algorithm is agnostic to the input source.

```suggestion
		return fmt.Errorf("read: invalid hardware: %v", err)
```

chrisdoherty4 (changes requested):
See inline comments.

chrisdoherty4 (approved):

PR #3670, opened 2022-10-13
Construct git tag path properly when falling back to artifacts from main

abhay-krishna:
If there is no build corresponding to release branches for some project, the release process falls back to sourcing artifacts built from `main`. At that time, it reads the git tag for the project from main, in order to construct the appropriate tag. But this fallback doesn't consider release-branched projects which may have their git tag at a nested folder. This PR fixes that.


pokearu (approved):

abhinavmpandey08 (approved):
/approve
/lgtm

PR #3673, opened 2022-10-13
Add CNI reconciler for VSphere

taneyland:
*Issue #, if available:*

*Description of changes:*
Reconcile CNI in vsphere cluster reconciler

*Testing (if applicable):*


controllers/cluster_controller_test.go
```diff
@@ -52,11 +53,13 @@ func newVsphereClusterReconcilerTest(t *testing.T, objs ...runtime.Object) *vsph
 
 	validator := vsphere.NewValidator(govcClient, &networkutils.DefaultNetClient{}, vcb)
 	defaulter := vsphere.NewDefaulter(govcClient)
+	cniReconciler := vspherereconcilermocks.NewMockCNIReconciler(ctrl)
```
g-gaston, line 56: Can we create a task to remove/move these tests? If they are still valid, they should be in `reconciler_test.go` :)

g-gaston (approved):
this is beautiful

controllers/cluster_controller_test.go
taneyland: Yes, will do!

PR #3679, opened 2022-10-14
Generate package with kube-version and cluster name

TerryHowe:


cmd/eksctl-anywhere/cmd/generatepackage.go
```diff
@@ -27,6 +29,9 @@ func init() {
 	generatePackageCommand.Flags().StringVar(&gpOptions.registry, "registry", "", "Used to specify an alternative registry for package generation")
 	generatePackageCommand.Flags().StringVar(&gpOptions.kubeConfig, "kubeconfig", "",
 		"Path to an optional kubeconfig file to use.")
+	if err := generatePackageCommand.MarkFlagRequired("cluster"); err != nil {
+		log.Fatalf("marking cluster flag as required: %s", err)
```
a-cool-train, line 33: Now I am getting concerned with requiring a cluster flag in case we are trying to generate before a cluster is created

TerryHowe: before a cluster created is registry discovery 🤷

TerryHowe: Anyway, I'm not dealing with this today. Tests fail for me on main, so I didn't notice these borked tests.

PR #3680, opened 2022-10-14
Added note on RHEL to bare metal docs

chrisnegus:
*Description of changes:*

Using RHEL images for EKS-A bare metal clusters requires that legacy bios be enabled on those machines. This PR adds a note about that to the docs.


TerryHowe (approved):
/lgtm
/approve

Issue #3349, opened 2022-09-14
Support for authenticated local/private registry 


ptrivedi:
**What would you like to be added**:
Allow for authenticated pulls from local/private registry using credentials.  If the private registry has password-based authentication credentials set up, EKS Anywhere should allow configuration of those credentials and be able to pull from it.

**Why is this needed**:
EKS Anywhere currently supports local/private registry configuration, but only allows for unauthenticated pulls.

Paulius0112:
Hi. Is it currently possible to pass a self-signed registry certificate in bare metal bottlerocket? We need it for pulling images from private registry, since providing certificate under `registryMirrorConfiguration.caCertContent` doesn't solve this problem.

PR #3684, opened 2022-10-15
Add local registry auth support to ubuntu on vsphere

ahreehong:
*Issue #, if available:*
[#3349]

*Description of changes:*
Added support for authenticated local registry for ubuntu on vsphere

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/aws/eks-anywhere/pkg/config"
 	"io/ioutil"
 	"net"
 	"net/url"
@@ -685,7 +684,7 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 	}
 
 	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate && clusterConfig.Spec.DatacenterRef.Kind != VSphereDatacenterKind {
-		return errors.New("authenticated local registry is only supported for vsphere provider currently")
+		return errors.New("authenticated registry mirror is only supported for vSphere provider currently")
 	}
 
 	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/aws/eks-anywhere/pkg/config"
 	"io/ioutil"
 	"net"
 	"net/url"
-	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -293,11 +293,12 @@ func (c *Cluster) RegistryMirror() string {
 	return net.JoinHostPort(c.Spec.RegistryMirrorConfiguration.Endpoint, c.Spec.RegistryMirrorConfiguration.Port)
 }
 
+// RegistryAuth returns whether or not
 func (c *Cluster) RegistryAuth() bool {
-	if c.Spec.RegistryMirrorConfiguration.Authenticate {
-		return true
+	if c.Spec.RegistryMirrorConfiguration == nil {
+		return false
 	}
-	return false
+	return c.Spec.RegistryMirrorConfiguration.Authenticate
 }
 
 func (c *Cluster) ProxyConfiguration() map[string]string {
@@ -684,10 +685,9 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 	}
 
 	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
-		username := os.Getenv("REGISTRY_USERNAME")
-		password := os.Getenv("REGISTRY_PASSWORD")
-		if username == "" || password == "" {
-			return fmt.Errorf("username or password not set, Provide REGISTRY_USERNAME and REGISTRY_PASSWORD to use authenticated registry mirror")
+		_, _, err := config.ReadCredentials()
+		if err != nil {
+			return err
 		}
 	}
 	return nil
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/yaml"
 
+	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 )
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -674,6 +674,10 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 	if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify && clusterConfig.Spec.DatacenterRef.Kind != SnowDatacenterKind {
 		return errors.New("insecureSkipVerify is only supported for snow provider")
 	}
+
+	//if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
+	//	if
+	//}
 	return nil
 }
 
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/url"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -290,6 +291,13 @@ func (c *Cluster) RegistryMirror() string {
 	}
 
 	return net.JoinHostPort(c.Spec.RegistryMirrorConfiguration.Endpoint, c.Spec.RegistryMirrorConfiguration.Port)
+}
+
+func (c *Cluster) RegistryAuth() bool {
+	if c.Spec.RegistryMirrorConfiguration.Authenticate {
+		return true
+	}
+	return false
 }
 
 func (c *Cluster) ProxyConfiguration() map[string]string {
@@ -675,9 +683,13 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 		return errors.New("insecureSkipVerify is only supported for snow provider")
 	}
 
-	//if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
-	//	if
-	//}
+	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
+		username := os.Getenv("REGISTRY_USERNAME")
+		password := os.Getenv("REGISTRY_PASSWORD")
+		if username == "" || password == "" {
+			return fmt.Errorf("username or password not set, Provide REGISTRY_USERNAME and REGISTRY_PASSWORD to use authenticated registry mirror")
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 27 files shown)


pkg/executables/kind.go
```diff
@@ -165,7 +169,7 @@ func (k *Kind) WithEnv(env map[string]string) bootstrapper.BootstrapClusterClien
 	}
 }
 
-func (k *Kind) WithRegistryMirror(endpoint string, caCertFile string) bootstrapper.BootstrapClusterClientOption {
+func (k *Kind) WithRegistryMirror(endpoint string, caCertFile string, env map[string]string) bootstrapper.BootstrapClusterClientOption {
```
chrisdoherty4, line 172: Can we avoid using `map[string]string` to pass credentials and toggle registry auth configuration? It makes for a hard to understand API.

We might want to shift to a `RegistryMirror` configuration object and describe what's optional etc making expectations very clear.

E.g.

```go
type RegistryMirrorAuth struct {
  Username string
  Password string
}

type RegistryMirrorConfig struct {
  Endpoint   string
  CACertPath string
  *RegistryMirrorAuth
}

func (k *Kind) WithRegistryMirror(RegistryMirrorConfig) bootstrapper.BootstrapClusterClientOption {
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -674,6 +682,14 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
...
 		return errors.New("insecureSkipVerify is only supported for snow provider")
 	}
+
+	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
+		username := os.Getenv("REGISTRY_USERNAME")
+		password := os.Getenv("REGISTRY_PASSWORD")
+		if username == "" || password == "" {
+			return fmt.Errorf("username or password not set, Provide REGISTRY_USERNAME and REGISTRY_PASSWORD to use authenticated registry mirror")
+		}
+	}
```
chrisdoherty4, line 692: Ideally we should separate reading configuration from validating it. This enables changing the source of configuration without creating a bunch of touch points. 

On a side note, we're duplicating environment variables across the code base as `REGISTRY_USERNAME` and `REGISTRY_PASSWORD` are specified elsewhere also. Are we confident their semantics are the same in all places? If so, we should be using constants rather than re-specifying.

pkg/dependencies/factory.go
```diff
@@ -116,6 +117,7 @@ func ForSpec(ctx context.Context, clusterSpec *cluster.Spec) *Factory {
 type Factory struct {
 	executablesConfig        *executablesConfig
 	registryMirror           string
+	registryAuth             bool
```
chrisdoherty4, line 120: I think we probably want to consolidate these under a single configuration object.

pkg/dependencies/factory.go
```diff
@@ -250,6 +257,16 @@ func (f *Factory) WithExecutableBuilder() *Factory {
...
 		if f.executablesConfig.useDockerContainer {
+
+			if f.registryAuth {
+				username := os.Getenv("REGISTRY_USERNAME")
+				password := os.Getenv("REGISTRY_PASSWORD")
+				err := f.executablesConfig.dockerClient.Login(context.Background(), f.registryMirror, username, password)
+				if err != nil {
+					return err
+				}
+			}
```
chrisdoherty4, line 268: This doesn't seem like the right place to do a docker login. I wonder if we can do this earlier? Perhaps it even requires an explicit step high up in the call path?

pkg/executables/config/kind.yaml
```diff
@@ -25,6 +25,11 @@ containerdConfigPatches:
 {{- else }}
         ca_file = "/etc/containerd/certs.d/{{.RegistryMirrorEndpoint}}/ca.crt"
 {{- end }}
+{{- if .RegistryAuth }}
+      [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.RegistryMirrorEndpoint}}".auth]
+        username = "{{.RegistryUsername}}"
+        password = "{{.RegistryPassword}}"
```
chrisdoherty4, line 31: How are we handling passwords containing quotes? Or usernames for that matter (albeit that's odd).

It looks like we could do with some more testing around this.

pkg/executables/dockercontainer.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 type DockerClient interface {
+	Login(ctx context.Context, endpoint, username, password string) error
```
chrisdoherty4, line 17: What's the consumer of this interface?

pkg/providers/vsphere/template.go
```diff
@@ -287,6 +293,13 @@ func buildTemplateMapMD(
 		if len(clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.CACertContent) > 0 {
 			values["registryCACert"] = clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.CACertContent
 		}
+
+		if clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.Authenticate {
```
chrisdoherty4, line 297: Is this going to need populating on every provider? It seems like an extremely hard to maintain approach if so - did we brainstorm any alternatives by chance?

pkg/providers/vsphere/config/template-md.yaml
```diff
@@ -83,6 +83,11 @@ spec:
             [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.registryMirrorConfiguration}}".tls]
               ca_file = "/etc/containerd/certs.d/{{.registryMirrorConfiguration}}/ca.crt"
             {{- end }}
+            {{- if .registryAuth }}
+            [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.registryMirrorConfiguration}}".auth]
+              username = "{{.registryUsername}}"
+              password = "{{.registryPassword}}"
```
chrisdoherty4, line 89: Similar to my other comment on quotes, we'll need additional test case(s).

pkg/providers/vsphere/testdata/expected_results_mirror_with_auth_config_md.yaml
```diff
@@ -0,0 +1,121 @@
...
+        path: "/etc/containerd/certs.d/1.2.3.4:1234/ca.crt"
+      - content: |
+          [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
+            [plugins."io.containerd.grpc.v1.cri".registry.mirrors."public.ecr.aws"]
+              endpoint = ["https://1.2.3.4:1234"]
+            [plugins."io.containerd.grpc.v1.cri".registry.configs."1.2.3.4:1234".tls]
+              ca_file = "/etc/containerd/certs.d/1.2.3.4:1234/ca.crt"
+            [plugins."io.containerd.grpc.v1.cri".registry.configs."1.2.3.4:1234".auth]
+              username = "username"
+              password = "password"
```
chrisdoherty4, line 48: @vivek-koppuru @vignesh-goutham Is there somewhere we need to call out the use of plain-text secrets?

Even better would be a different way of achieving this, but I don't think there is.

chrisdoherty4 (changes requested):

pkg/api/v1alpha1/cluster.go
```diff
@@ -674,6 +682,14 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 	if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify && clusterConfig.Spec.DatacenterRef.Kind != SnowDatacenterKind {
 		return errors.New("insecureSkipVerify is only supported for snow provider")
 	}
+
+	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate {
+		username := os.Getenv("REGISTRY_USERNAME")
+		password := os.Getenv("REGISTRY_PASSWORD")
+		if username == "" || password == "" {
+			return fmt.Errorf("username or password not set, Provide REGISTRY_USERNAME and REGISTRY_PASSWORD to use authenticated registry mirror")
+		}
```
abhinavmpandey08, line 691: you can just use this predefined function https://github.com/aws/eks-anywhere/blob/main/pkg/config/registry.go#L8

pkg/api/v1alpha1/cluster.go
```diff
@@ -292,6 +293,13 @@ func (c *Cluster) RegistryMirror() string {
 	return net.JoinHostPort(c.Spec.RegistryMirrorConfiguration.Endpoint, c.Spec.RegistryMirrorConfiguration.Port)
 }
 
+func (c *Cluster) RegistryAuth() bool {
```
abhinavmpandey08, line 296: Can you add a comment briefly describing what this func is for? Looks like the linter is complaining about it

pkg/dependencies/factory.go
```diff
@@ -250,6 +257,16 @@ func (f *Factory) WithExecutableBuilder() *Factory {
 		}
 
 		if f.executablesConfig.useDockerContainer {
+
+			if f.registryAuth {
+				username := os.Getenv("REGISTRY_USERNAME")
+				password := os.Getenv("REGISTRY_PASSWORD")
+				err := f.executablesConfig.dockerClient.Login(context.Background(), f.registryMirror, username, password)
+				if err != nil {
+					return err
```
abhinavmpandey08, line 266: you can re-use https://github.com/aws/eks-anywhere/blob/main/pkg/config/registry.go#L8

pkg/providers/vsphere/vsphere_test.go
```diff
@@ -1160,6 +1160,35 @@ func TestProviderGenerateDeploymentFileWithMirrorAndCertConfig(t *testing.T) {
 	test.AssertContentToFile(t, string(md), "testdata/expected_results_mirror_config_with_cert_md.yaml")
 }
 
+func TestProviderGenerateDeploymentFileWithMirrorAuth(t *testing.T) {
+	clusterSpecManifest := "cluster_mirror_with_auth_config.yaml"
+	mockCtrl := gomock.NewController(t)
+	setupContext(t)
+	os.Setenv("REGISTRY_USERNAME", "username")
+	os.Setenv("REGISTRY_PASSWORD", "password")
```
abhinavmpandey08, line 1168: can you check the return value of `os.Setenv`? It returns an err I think

pkg/executables/dockercontainer.go
ahreehong: it's used when we do the docker login in factory.go:

`err := f.executablesConfig.dockerClient.Login(context.Background(), f.registryMirror, username, password)`

pkg/executables/config/kind.yaml
ahreehong: I can do more testing around this and follow up in a different pr if any changes are necessary

pkg/providers/vsphere/config/template-md.yaml
ahreehong: https://github.com/aws/eks-anywhere/pull/3684#discussion_r997547774

pkg/api/v1alpha1/cluster.go
```diff
@@ -674,6 +683,17 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 	if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify && clusterConfig.Spec.DatacenterRef.Kind != SnowDatacenterKind {
 		return errors.New("insecureSkipVerify is only supported for snow provider")
 	}
+
+	if clusterConfig.Spec.RegistryMirrorConfiguration.Authenticate && clusterConfig.Spec.DatacenterRef.Kind != VSphereDatacenterKind {
+		return errors.New("authenticated local registry is only supported for vsphere provider currently")
```
abhinavmpandey08, line 688:
```suggestion
		return errors.New("authenticated registry mirror is only supported for vSphere provider currently")
```

abhinavmpandey08 (approved):
/approve
/lgtm

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #3686, opened 2022-10-16
Simplify vsphere templater and ssh key defaults

g-gaston:
## Description of changes

While integrating the template builder with the full cluster lifecycle controller, I faced some issues. I started unraveling the code and I ended up with this. It's not perfect by any means but it does simplify the code a little bit and reduces coupling, which makes reusing the template builder easier. 

* This avoids modifying user's API input by moving SSH key normalization to the template builder level as opposed to defaulting. It's a good practice and makes the template builder more reusable, specially from controllers. This is the main behavioral change: if users include the `user@host` comment in the ssh key, this will be kept in the eks-a vsphere machine config and just removed from the vsphere machine template.

* It simplifies template builder by making use of the API structs in the cluster.Config instead of maintaining its own specs. This also reduces the coupling between provider and template builder.

* It also reduces the use of internal provider pointers to API structs. It doesn't completely remove them, but at least now the tests don't depend on the internal private members of the provider.

* Finally, it removes the provider members to maintain SSH keys. The purpose wasn't clear, it coupled some public methods together and it didn't seem necessary.

## Testing

* Unit tests
* Run create/upgrade e2e test manually

### DIFF
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -81,6 +81,7 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 	switch cs.Spec.DatacenterRef.Kind {
 	case anywherev1.VSphereDatacenterKind:
 		vdc := &anywherev1.VSphereDatacenterConfig{}
+		spec.VSphereMachineConfigs = make(map[string]*anywherev1.VSphereMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations)+2)
 		cpVmc := &anywherev1.VSphereMachineConfig{}
 		etcdVmc := &anywherev1.VSphereMachineConfig{}
 		workerVmc := &anywherev1.VSphereMachineConfig{}
@@ -89,22 +90,26 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 		if err != nil {
 			return err
 		}
+		spec.VSphereDatacenter = vdc
 		err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.ControlPlaneConfiguration.MachineGroupRef.Name}, cpVmc)
 		if err != nil {
 			return err
 		}
+		spec.VSphereMachineConfigs[cs.Spec.ControlPlaneConfiguration.MachineGroupRef.Name] = cpVmc
 		for _, workerNodeGroupConfiguration := range cs.Spec.WorkerNodeGroupConfigurations {
 			err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: workerNodeGroupConfiguration.MachineGroupRef.Name}, workerVmc)
 			if err != nil {
 				return err
 			}
 			workerVmcs[workerNodeGroupConfiguration.MachineGroupRef.Name] = *workerVmc
+			spec.VSphereMachineConfigs[workerNodeGroupConfiguration.MachineGroupRef.Name] = workerVmc
 		}
 		if cs.Spec.ExternalEtcdConfiguration != nil {
 			err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name}, etcdVmc)
 			if err != nil {
 				return err
 			}
+			spec.VSphereMachineConfigs[cs.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name] = etcdVmc
 		}
 		r, err := cor.vsphereTemplate.TemplateResources(ctx, cs, spec, *vdc, *cpVmc, *etcdVmc, workerVmcs)
 		if err != nil {
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -81,6 +81,7 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 	switch cs.Spec.DatacenterRef.Kind {
 	case anywherev1.VSphereDatacenterKind:
 		vdc := &anywherev1.VSphereDatacenterConfig{}
+		// max len = len(workers) + CP + etcd
 		spec.VSphereMachineConfigs = make(map[string]*anywherev1.VSphereMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations)+2)
 		cpVmc := &anywherev1.VSphereMachineConfig{}
 		etcdVmc := &anywherev1.VSphereMachineConfig{}
--- a/controllers/resource/reconciler_test.go
+++ b/controllers/resource/reconciler_test.go
@@ -125,7 +125,7 @@ func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 		args    args
 		want    controllerruntime.Result
 		wantErr bool
-		prepare func(context.Context, *mocks.MockResourceFetcher, *mocks.MockResourceUpdater, string, string)
+		prepare func(*testing.T, context.Context, *mocks.MockResourceFetcher, *mocks.MockResourceUpdater, string, string)
 	}{
 		{
 			name: "worker node reconcile (Vsphere provider) - worker nodes has changes",
@@ -138,7 +138,7 @@ func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 				},
 			},
 			want: controllerruntime.Result{},
-			prepare: func(ctx context.Context, fetcher *mocks.MockResourceFetcher, resourceUpdater *mocks.MockResourceUpdater, name string, namespace string) {
+			prepare: func(t *testing.T, ctx context.Context, fetcher *mocks.MockResourceFetcher, resourceUpdater *mocks.MockResourceUpdater, name string, namespace string) {
 				cluster := &anywherev1.Cluster{}
 				cluster.SetName(name)
 				cluster.SetNamespace(namespace)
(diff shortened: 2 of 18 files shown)


controllers/resource/reconciler.go
```diff
@@ -81,6 +81,7 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 	switch cs.Spec.DatacenterRef.Kind {
 	case anywherev1.VSphereDatacenterKind:
 		vdc := &anywherev1.VSphereDatacenterConfig{}
+		spec.VSphereMachineConfigs = make(map[string]*anywherev1.VSphereMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations)+2)
```
g-gaston, line 84: Adding all vsphere structs to the cluster spec, since it's initially built without these and the template builder makes use of them

controllers/resource/reconciler_test.go
```diff
@@ -127,7 +127,7 @@ func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 		args    args
 		want    controllerruntime.Result
 		wantErr bool
-		prepare func(context.Context, *mocks.MockResourceFetcher, *mocks.MockResourceUpdater, string, string)
+		prepare func(*testing.T, context.Context, *mocks.MockResourceFetcher, *mocks.MockResourceUpdater, string, string)
```
g-gaston, line 130: passing `t` here so we can actually distinguish failures between the subtests

controllers/resource/reconciler_test.go
```diff
@@ -216,7 +216,7 @@ func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 						Users: []anywherev1.UserConfiguration{
 							{
 								Name:              "capv",
-								SshAuthorizedKeys: []string{"ssh-rsa ssh_key_value"},
+								SshAuthorizedKeys: []string{"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC8ZEibIrz1AUBKDvmDiWLs9f5DnOerC4qPITiDtSOuPAsxgZbRMavBfVTxodMdAkYRYlXxK6PqNo0ve0qcOV2yvpxH1OogasMMetck6BlM/dIoo3vEY4ZoG9DuVRIf9Iry5gJKbpMDYWpx1IGZrDMOFcIM20ii2qLQQk5hfq9OqdqhToEJFixdgJt/y/zt6Koy3kix+XsnrVdAHgWAq4CZuwt1G6JUAqrpob3H8vPmL7aS+35ktf0pHBm6nYoxRhslnWMUb/7vpzWiq+fUBIm2LYqvrnm7t3fRqFx7p2sZqAm2jDNivyYXwRXkoQPR96zvGeMtuQ5BVGPpsDfVudSW21+pEXHI0GINtTbua7Ogz7wtpVywSvHraRgdFOeY9mkXPzvm2IhoqNrteck2GErwqSqb19mPz6LnHueK0u7i6WuQWJn0CUoCtyMGIrowXSviK8qgHXKrmfTWATmCkbtosnLskNdYuOw8bKxq5S4WgdQVhPps2TiMSZndjX5NTr8= ubuntu@ip-10-2-0-6"},
```
g-gaston, line 219: We know need to pass a valid ssh key here

controllers/resource/testdata/eksa-cluster.yaml
```diff
@@ -51,7 +51,7 @@ spec:
   users:
     - name: capv
       sshAuthorizedKeys:
-        - "ssh-rsa ssh_key_value"
+        - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC8ZEibIrz1AUBKDvmDiWLs9f5DnOerC4qPITiDtSOuPAsxgZbRMavBfVTxodMdAkYRYlXxK6PqNo0ve0qcOV2yvpxH1OogasMMetck6BlM/dIoo3vEY4ZoG9DuVRIf9Iry5gJKbpMDYWpx1IGZrDMOFcIM20ii2qLQQk5hfq9OqdqhToEJFixdgJt/y/zt6Koy3kix+XsnrVdAHgWAq4CZuwt1G6JUAqrpob3H8vPmL7aS+35ktf0pHBm6nYoxRhslnWMUb/7vpzWiq+fUBIm2LYqvrnm7t3fRqFx7p2sZqAm2jDNivyYXwRXkoQPR96zvGeMtuQ5BVGPpsDfVudSW21+pEXHI0GINtTbua7Ogz7wtpVywSvHraRgdFOeY9mkXPzvm2IhoqNrteck2GErwqSqb19mPz6LnHueK0u7i6WuQWJn0CUoCtyMGIrowXSviK8qgHXKrmfTWATmCkbtosnLskNdYuOw8bKxq5S4WgdQVhPps2TiMSZndjX5NTr8= ubuntu@ip-10-2-0-6
```
g-gaston, line 54: Same here, we know need to pass a valid ssh key here

controllers/resource/template.go
```diff
@@ -105,7 +105,7 @@ func (r *VsphereTemplate) TemplateResources(ctx context.Context, eksaCluster *an
 		if err != nil {
 			return nil, err
 		}
-		if vsphere.NeedsNewKubeadmConfigTemplate(&workerNodeGroupConfiguration, oldWn, oldVmc, &vmc) {
+		if needsVSphereNewKubeadmConfigTemplate(&workerNodeGroupConfiguration, oldWn, oldVmc, &vmc) {
```
g-gaston, line 108: using a different algorithm to compare SSH keys
Since we try to rebuild the vsphere machine config from the vsphere maachine template by reversing the implicit conversion logic of the template builder, we need to be smart when comparing these. Such conversion logic is a one way street where data is lost, so a perfect reverse conversion is impossible. In particular, the `user@host` part is lost
This new method takes that into account

pkg/providers/vsphere/spec.go
```diff
@@ -7,45 +7,33 @@ import (
...
-		}
-		machineConfigsInCluster[m.Name] = machineConfig
-	}
-
+// NewSpec constructs a new vSphere cluster Spec.
+func NewSpec(clusterSpec *cluster.Spec) *Spec {
 	return &Spec{
-		Spec:                 clusterSpec,
-		datacenterConfig:     datacenterConfig,
-		machineConfigsLookup: machineConfigsInCluster,
```
g-gaston, line 27: This is not needed anymore since this data is already in the cluster.Spec

pkg/providers/vsphere/template.go
```diff
@@ -18,29 +18,18 @@ import (
...
-	controlPlaneMachineSpec, etcdMachineSpec *anywherev1.VSphereMachineConfigSpec,
-	workerNodeGroupMachineSpecs map[string]anywherev1.VSphereMachineConfigSpec,
 	now types.NowFunc,
 	fromController bool,
 ) *VsphereTemplateBuilder {
 	return &VsphereTemplateBuilder{
-		datacenterSpec:              datacenterSpec,
-		controlPlaneMachineSpec:     controlPlaneMachineSpec,
-		WorkerNodeGroupMachineSpecs: workerNodeGroupMachineSpecs,
-		etcdMachineSpec:             etcdMachineSpec,
```
g-gaston, line 31: No need to maintain all of these in here, it's in the cluster spec
Also, this complicated a lot the code, having to keep the template in builder in sync with the defaulting/changes done to the cluster spec (this doesn't use pointers)

pkg/providers/vsphere/vsphere_test.go
```diff
@@ -1681,28 +1643,26 @@ func TestSetupAndValidateCreateClusterNoServer(t *testing.T) {
...
 	var tctx testContext
 	tctx.SaveContext()
 
 	err := provider.SetupAndValidateCreateCluster(ctx, clusterSpec)
 	if err != nil {
 		t.Fatalf("Unexpected error <%v>", err)
 	}
-	if provider.datacenterConfig.Spec.Thumbprint != "" {
-		t.Fatalf("Expected=<> actual=<%s>", provider.datacenterConfig.Spec.Thumbprint)
+	if clusterSpec.VSphereDatacenter.Spec.Thumbprint != "" {
```
g-gaston, line 1656: Avoid having to inspect unexported members: defaulting is reflected in the cluster Spec

pkg/providers/vsphere/vsphere.go
```diff
@@ -245,109 +225,25 @@ func (p *vsphereProvider) MachineResourceType() string {
...
-				p.etcdSshAuthKey = p.controlPlaneSshAuthKey
-			} else if useKeyGeneratedForWorker {
-				p.etcdSshAuthKey = p.workerSshAuthKey // if cp key was provided by user, check if worker key was generated by cli and use that
+func (p *vsphereProvider) generateSSHKeysIfNotSet() error {
+	var generatedKey string
+	for _, machineConfig := range p.machineConfigs {
+		user := machineConfig.Spec.Users[0]
+		if user.SshAuthorizedKeys[0] == "" {
+			if generatedKey != "" { // use the same key
+				user.SshAuthorizedKeys[0] = generatedKey
```
abhinavmpandey08, line 234: `generatedKey` is always gonna be empty at this point right?

pkg/providers/vsphere/vsphere.go
```diff
@@ -493,11 +389,7 @@ func (p *vsphereProvider) SetupAndValidateUpgradeCluster(ctx context.Context, cl
 		return err
 	}
 
-	err := p.setupSSHAuthKeysForUpgrade()
-	if err != nil {
-		return fmt.Errorf("failed setup and validations: %v", err)
-	}
-	err = p.validateMachineConfigsNameUniqueness(ctx, cluster, clusterSpec)
```
abhinavmpandey08, line 500: Shouldn't we also generate ssh keys during upgrade if it's set to empty? Otherwise, there will be a way to upgrade a cluster to not have SSH keys at all, when that's not allowed during create.

abhinavmpandey08 (commented):
looks good! Just some minor comments/nits

pkg/providers/vsphere/vsphere.go
g-gaston: No necessarily. It can have been generated in prior iterations of the loop. The first machine config that is found without an ssh key will encounter an empty `generatedKey` which will trigger an SSH key generation. The next machine config (if there is any) without a key, will reuse `generatedKey`

pkg/providers/vsphere/vsphere.go
g-gaston: If you look at what `setupSSHAuthKeysForUpgrade` was doing, it never generated new keys for empty fields. It only stripped the comment from them. That's a difference in behavior between create and upgrade. Not sure why that was done in the first place, but with this refactor I tried to keep as much behavior as possible intact.

pkg/providers/vsphere/template.go
```diff
@@ -319,5 +345,17 @@ func buildTemplateMapMD(
 		values["bottlerocketBootstrapVersion"] = bundle.BottleRocketBootstrap.Bootstrap.Tag()
 	}
 
-	return values
+	return values, nil
+}
+
+func etcdMachineConfig(spec *cluster.Spec) *anywherev1.VSphereMachineConfig {
```
taneyland, line 351: Instead of defining a new method here, can we re-use this one? https://github.com/aws/eks-anywhere/blob/15bb295f98245fb418fcb2b086ca9733bf83e62d/pkg/providers/vsphere/spec.go#L27

Same for CP and worker node configs

pkg/providers/vsphere/vsphere.go
abhinavmpandey08: oh totally missed the for loop

pkg/providers/vsphere/vsphere.go
abhinavmpandey08: hmm interesting, we can track this in a separate issue

abhinavmpandey08 (approved):
looks good!

taneyland:
/lgtm 
🙌

taneyland (approved):

PR #3694, opened 2022-10-17
Retrieve Nutanix CAPX image only if it is in the bundle

abhinavmpandey08:
*Description of changes:*
Nutanix is not added to the release bundle so retrieve it only if it's in the bundle.
This resolves issue with `import-images` where it fails trying to import an empty image


abhay-krishna (approved):
/approve

PR #3698, opened 2022-10-17
Surface errors emitted by Helm push operation

abhay-krishna:
We were dropping the errors from the Helm pusher's `Run` command, which shouldn't have been the case.


abhay-krishna:
/override eks-anywhere-release-tooling-presubmit
/override eks-anywhere-release-tooling-test-presubmit

This method is not tested in any of these presubmits.

PR #3712, opened 2022-10-18
Fixing CloudStack e2e test issues

maxdrib:
*Issue #, if available:*

*Description of changes:*
First, CLOUDSTACK_PROVIDER=true is being set in the integration tests build spec for running older releases of EKS-A with CloudStack, but it's not being used by the e2e test framework, since that doesn't automatically set up all environment variables. Instead, it looks at the list of required env vars for the e2e tests and only passes those to the test environment, as shown [here](https://github.com/aws/eks-anywhere/blob/9bdf8dd8037ea4a079bbaa55c2c7ab0a3e9538fe/internal/test/e2e/cloudstack.go#L14)

Second, we often observe control plane IP address collisions when e2e tests are running which appear as the error `x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kubernetes")` from the K8s api server on a cluster. Therefore, we are increasing the CIDR range from which these control plane IP's are being generated eight-fold, from /24 to /21

*Testing (if applicable):*


jaxesn (approved):
/approve

PR #3724, opened 2022-10-18
Coverage view fix

ewollesen:
One quick fix and one quick "feature" for the patch-based Makefile targets I recently added.

The fix is for changes that modify package source files without modifying a _test.go file in the same package---those packages would have been ignored.

The new "feature" is to continue running the Make recipe when there's a test failure, so that the coverage report will still open.


TerryHowe (approved):
/approve

PR #3747, opened 2022-10-19
Expose maxSurge and maxUnavailable knobs for rolling update strategy

ptrivedi:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -95,6 +95,13 @@ func WithClusterEndpoint() ClusterGenerateOpt {
 	}
 }
 
+// WithCPUpgradeRolloutStrategy allows add UpgradeRolloutStrategy option to cluster config under ControlPlaneConfiguration.
+func WithCPUpgradeRolloutStrategy(maxSurge int, maxUnavailable int) ClusterGenerateOpt {
+	return func(c *ClusterGenerate) {
+		c.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy = &ControlPlaneUpgradeRolloutStrategy{Type: "RollingUpdate", RollingUpdate: ControlPlaneRollingUpdateParams{MaxSurge: maxSurge}}
+	}
+}
+
 func WithDatacenterRef(ref ProviderRefAccessor) ClusterGenerateOpt {
 	return func(c *ClusterGenerate) {
 		c.Spec.DatacenterRef = Ref{
@@ -131,6 +138,16 @@ func WithWorkerMachineGroupRef(ref ProviderRefAccessor) ClusterGenerateOpt {
 		c.Spec.WorkerNodeGroupConfigurations[0].MachineGroupRef = &Ref{
 			Kind: ref.Kind(),
 			Name: ref.Name(),
+		}
+	}
+}
+
+// WithWorkerMachineUpgradeRolloutStrategy allows add UpgradeRolloutStrategy option to cluster config under WorkerNodeGroupConfiguration.
+func WithWorkerMachineUpgradeRolloutStrategy(maxSurge int, maxUnavailable int) ClusterGenerateOpt {
+	return func(c *ClusterGenerate) {
+		c.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy = &WorkerNodesUpgradeRolloutStrategy{
+			Type:          "RollingUpdate",
+			RollingUpdate: WorkerNodesRollingUpdateParams{MaxSurge: maxSurge, MaxUnavailable: maxUnavailable},
 		}
 	}
 }
@@ -178,6 +195,7 @@ var clusterConfigValidations = []func(*Cluster) error{
 	validateProxyConfig,
 	validateMirrorConfig,
 	validatePodIAMConfig,
+	validateCPUpgradeRolloutStrategy,
 	validateControlPlaneLabels,
 }
 
@@ -437,6 +455,10 @@ func validateWorkerNodeGroups(clusterConfig *Cluster) error {
 
 		if err := validateAutoscalingConfig(&workerNodeGroupConfig); err != nil {
 			return fmt.Errorf("validating autoscaling configuration: %v", err)
+		}
+
+		if err := validateMDUpgradeRolloutStrategy(&workerNodeGroupConfig); err != nil {
+			return fmt.Errorf("validating upgrade rollout strategy configuration: %v", err)
 		}
 
 		if workerNodeGroupNames[workerNodeGroupConfig.Name] {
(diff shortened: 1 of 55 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
...
+	}
+
+	if clusterConfig.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy.RollingUpdate.MaxSurge > 1 {
+		return fmt.Errorf("ControlPlaneConfiguration: maxSurge for control plane must be 0 or 1")
+	}
+
+	return nil
+}
+
+func validateMDUpgradeRolloutStrategy(clusterConfig *Cluster) error {
```
maxdrib, line 780: don't we need to check for provider support here, if only Tinkerbell supports it

ptrivedi: This would be provider-wide. It is generic capi functionality. I only have unit tests for tinkerbell right now, but I am going to add similar unit test in other providers. Currently these knobs default to the values that CAPI already has.

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -194,6 +194,7 @@ func generateClusterConfig(clusterName string) error {
 			v1alpha1.WorkerNodeConfigCount(1),
 			v1alpha1.WorkerNodeConfigName(constants.DefaultWorkerNodeGroupName),
 		)
+		clusterConfigOpts = append(clusterConfigOpts, v1alpha1.WithCPUpgradeRolloutStrategy(1, 0), v1alpha1.WithWorkerMachineUpgradeRolloutStrategy(1, 0))
```
vivek-koppuru, line 197: I thought this was going to be optional and not a default?

vivek-koppuru (commented):
We should gate this feature specifically for Tinkerbell and fail for other providers too right? That would also help us be able to see the scope for implementing this for external etcd for other providers.

ptrivedi:
> We should gate this feature specifically for Tinkerbell and fail for other providers too right? That would also help us be able to see the scope for implementing this for external etcd for other providers.

We are not exposing these for external etcd at all. So, technically, this feature is currently supported for all providers, but only for CP and MD. Rolling upgrade for external etcd anyway requires the surge to be equal to the number of nodes, so it is a whole different thing where surge/unavailable don't apply.

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
ptrivedi: It is optional, but the default values do show up in cluster config. If it is specifically removed, a warning to the effect of 
`ControlPlaneConfiguration: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1`
or 
`WorkerNodeGroupConfigurations: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1 and maxUnavailable=0`
is printed and cli moves on.

pkg/api/v1alpha1/cluster.go
```diff
@@ -95,6 +95,13 @@ func WithClusterEndpoint() ClusterGenerateOpt {
 	}
 }
 
+// WithCPUpgradeRolloutStrategy allows add UpgradeRolloutStrategy option to cluster config under ControlPlaneConfiguration.
+func WithCPUpgradeRolloutStrategy(maxSurge int, maxUnavailable int) ClusterGenerateOpt {
+	return func(c *ClusterGenerate) {
+		c.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy = &ControlPlaneUpgradeRolloutStrategy{Type: "RollingUpdate", RollingUpdate: ControlPlaneRollingUpdateParams{MaxSurge: maxSurge}}
+	}
```
abhinavmpandey08, line 102: nit: match the indentation in `WithWorkerMachineUpgradeRolloutStrategy()`
```suggestion
		c.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy = &ControlPlaneUpgradeRolloutStrategy{
		    Type: "RollingUpdate",
		    RollingUpdate: ControlPlaneRollingUpdateParams{MaxSurge: maxSurge},
		}
	}
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
 	}
 	return nil
 }
+
+func validateCPUpgradeRolloutStrategy(clusterConfig *Cluster) error {
+	if clusterConfig.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy == nil {
+		fmt.Println("ControlPlaneConfiguration: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1")
```
abhinavmpandey08, line 765: Use logger instead of `fmt.Println`
```suggestion
		logger.Info("ControlPlaneConfiguration: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1")
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
...
+
+	return nil
+}
+
+func validateMDUpgradeRolloutStrategy(clusterConfig *Cluster) error {
+	if len(clusterConfig.Spec.WorkerNodeGroupConfigurations) <= 0 {
+		return nil
+	}
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy == nil {
+		fmt.Println("WorkerNodeGroupConfigurations: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1 and maxUnavailable=0")
```
abhinavmpandey08, line 785: same as above: use logger

pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
...
+	}
+
+	return nil
+}
+
+func validateMDUpgradeRolloutStrategy(clusterConfig *Cluster) error {
+	if len(clusterConfig.Spec.WorkerNodeGroupConfigurations) <= 0 {
+		return nil
+	}
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy == nil {
```
abhinavmpandey08, line 784: we should validate all the `WorkerNodeGroupConfigurations[*].UpgradeRolloutStrategy` not just the first one

pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
...
+		return nil
+	}
+
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.Type != "RollingUpdate" {
+		return fmt.Errorf("WorkerNodeGroupConfiguration: only 'RollingUpdate' supported for upgrade rollout strategy type")
+	}
+
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxSurge == 0 && clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxUnavailable == 0 {
+		return fmt.Errorf("WorkerNodeGroupConfiguration: maxSurge and maxAvailable not specified or are 0. maxSurge and maxUnavailable cannot both be 0")
+	}
```
abhinavmpandey08, line 795: same as above: validate all WorkerNodeGroupConfigurations

pkg/api/v1alpha1/cluster.go
```diff
@@ -740,3 +759,39 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
...
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy == nil {
+		fmt.Println("WorkerNodeGroupConfigurations: UpgradeRolloutStrategy not specified in cluster config. CAPI will default to 'RollingUpdate' with maxSurge=1 and maxUnavailable=0")
+		return nil
+	}
+
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.Type != "RollingUpdate" {
+		return fmt.Errorf("WorkerNodeGroupConfiguration: only 'RollingUpdate' supported for upgrade rollout strategy type")
+	}
+
+	if clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxSurge == 0 && clusterConfig.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxUnavailable == 0 {
```
abhinavmpandey08, line 793: can these 2  values be negative? If not, let's also add checks for that

pkg/providers/tinkerbell/template.go
```diff
@@ -467,6 +472,12 @@ func buildTemplateMapMD(clusterSpec *cluster.Spec, workerNodeGroupMachineSpec v1
 		"workerNodeGroupTaints":  workerNodeGroupConfiguration.Taints,
 	}
 
+	if clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy != nil {
+		values["upgradeRolloutStrategy"] = true
+		values["maxSurge"] = clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxSurge
+		values["maxUnavailable"] = clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].UpgradeRolloutStrategy.RollingUpdate.MaxUnavailable
+	}
```
abhinavmpandey08, line 479: each worker node group can have different `UpgradeRolloutStrategy`
Let's move this to here https://github.com/aws/eks-anywhere/blob/3fc030e5cf750f5eb7a9f3b4ca647eb4313cfabf/pkg/providers/tinkerbell/template.go#L147-L150
so we can pass `UpgradeRolloutStrategy`  by worker node group

pkg/providers/tinkerbell/testdata/cluster_tinkerbell_multiple_node_groups.yaml
```diff
@@ -30,11 +35,21 @@ spec:
...
       name: md-0
     - count: 1
       machineGroupRef:
         name: test-md
         kind: TinkerbellMachineConfig
+      upgradeRolloutStrategy:
+        type: "RollingUpdate"
+        rollingUpdate:
+          maxSurge: 1
+          maxUnavailable: 0
```
abhinavmpandey08, line 52: can you give the second worker node group a different `maxSurge` to make sure it's not using the same values for all
```suggestion
      upgradeRolloutStrategy:
        type: "RollingUpdate"
        rollingUpdate:
          maxSurge: 2
          maxUnavailable: 0
```

cmd/eksctl-anywhere/cmd/generateclusterconfig.go
vivek-koppuru: Can we not default to it showing up? This would then truly be an optional field, unless we want it to be specified?

ptrivedi: I was only showing it for Tinkerbell provider because someone would highly likely want to change the default values that match with CAPI. But it might not be a bad idea to make it consistent with other providers and hide it, then communicate to users via docs. Will change for now to not show, we can discuss further

PR #3765, opened 2022-10-19
Remove cgroup driver arg for docker 1.24 onwards

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Remove cgroup driver arg for docker 1.24 onwards

[Extra reference
](https://github.com/kubernetes-sigs/cluster-api/blob/69ab7870322d4e69eeb5c616299564ba78655ae0/test/e2e/data/infrastructure-docker/v1beta1/v1.2/clusterclass-quick-start.yaml#L92)*Testing (if applicable):*

### DIFF
--- a/pkg/clusterapi/extraargs.go
+++ b/pkg/clusterapi/extraargs.go
@@ -102,6 +102,13 @@ func ControlPlaneNodeLabelsExtraArgs(cpc v1alpha1.ControlPlaneConfiguration) Ext
 	return nodeLabelsExtraArgs(cpc.Labels)
 }
 
+// CgroupDriverExtraArgs args added for kube versions below 1.24.
+func CgroupDriverExtraArgs() ExtraArgs {
+	args := ExtraArgs{}
+	args.AddIfNotEmpty("cgroup-driver", "cgroupfs")
+	return args
+}
+
 func nodeLabelsExtraArgs(labels map[string]string) ExtraArgs {
 	args := ExtraArgs{}
 	args.AddIfNotEmpty("node-labels", labelsMapToArg(labels))
--- a/pkg/providers/docker/docker.go
+++ b/pkg/providers/docker/docker.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers"
 	"github.com/aws/eks-anywhere/pkg/providers/common"
+	"github.com/aws/eks-anywhere/pkg/semver"
 	"github.com/aws/eks-anywhere/pkg/templater"
 	"github.com/aws/eks-anywhere/pkg/types"
 	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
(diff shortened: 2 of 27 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -569,6 +570,12 @@ const (
 	Kube124 KubernetesVersion = "1.24"
 )
 
+// KubeVersionToValidSemver converts kube version to semver for comparisons.
+func KubeVersionToValidSemver(kubeVersion KubernetesVersion) (*semver.Version, error) {
```
vivek-koppuru, line 574: This shouldn't be in the cluster_types.go file I think. Sounds like a utility function based on when we need to use, because this gives the impression that `.0` is the version used in the cluster.

pkg/clusterapi/extraargs.go
```diff
@@ -102,6 +102,13 @@ func ControlPlaneNodeLabelsExtraArgs(cpc v1alpha1.ControlPlaneConfiguration) Ext
 	return nodeLabelsExtraArgs(cpc.Labels)
 }
 
+// CgroupDriverExtraArgs args added for kube versions below 1.24.
+func CgroupDriverExtraArgs() ExtraArgs {
+	args := ExtraArgs{}
+	args.AddIfNotEmpty("cgroup-driver", "cgroupfs")
```
vivek-koppuru, line 108: Did we check if the default is systemd for all kube versions we support? Wondering if we want to choose between cgroupfs and systemd instead of omitting

pkg/providers/docker/docker.go
```diff
@@ -184,13 +187,36 @@ func (d *DockerTemplateBuilder) GenerateCAPISpecWorkers(clusterSpec *cluster.Spe
...
 	sharedExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs()
 	kubeletExtraArgs := clusterapi.SecureTlsCipherSuitesExtraArgs().
 		Append(clusterapi.ResolvConfExtraArgs(clusterSpec.Cluster.Spec.ClusterNetwork.DNS.ResolvConf)).
 		Append(clusterapi.ControlPlaneNodeLabelsExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration))
+
+	cgroupDriverArgs, err := kubeletCgroupDriverExtraArgs(clusterSpec.Cluster.Spec.KubernetesVersion)
+	if err != nil {
+		return nil, err
+	}
+	kubeletExtraArgs.Append(cgroupDriverArgs)
```
vivek-koppuru, line 218: This would append an extra empty struct if nothing exists right? Should we make this a conditional or avoid it differently? Wonder if we have other examples of this

pkg/clusterapi/extraargs.go
mitalipaygude: yes from 1.24 onwards it defaults to systemd, confirmed that with testing.

pkg/api/v1alpha1/cluster_types.go
mitalipaygude: moved it to semver utils pkg

vivek-koppuru (approved):

PR #3771, opened 2022-10-20
Ensure Additional Trust bundle is propagated to the Nutanix Client

thunderboltsid:
The additional trust bundle was not being injected into the Nutanix Prism client. This PR ensures the certificate provided in the additional trust bundle is tacked on as a trusted certificate when constructing the Nutanix client.


pkg/dependencies/testdata/cluster_nutanix_with_empty_trust_bundle.yaml
```diff
@@ -0,0 +1,75 @@
...
+---
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: NutanixDatacenterConfig
+metadata:
+  name: eksa-unit-test
+  namespace: default
+spec:
+  endpoint: "prism.nutanix.com"
+  port: 9440
+  additionalTrustBundle: |
```
deepakm-ntnx, line 48: can we remove this completely in order to mimic real empty or nil

thunderboltsid: no

abhinavmpandey08 (commented):
Can you add a preflight to validate the cert using https://github.com/aws/eks-anywhere/blob/main/pkg/crypto/validator.go#L39

abhinavmpandey08 (approved):
/approve
/lgtm

PR #3784, opened 2022-10-20
fix orphaned machines left after timeout of a tinkerbell upgrade test

gwesterfieldjr:
Add cp timeout on all tinkerbell tests and ensure all tests power off hardware prior to test. This guards against orphaned machines from previous test run that timeout and were left in provisioned state.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #3786, opened 2022-10-20
Support tagging VMs

gwesterfieldjr:
Add tags option to vSphereMachineConfig so users can assign vSphere tags to their VM's during the EKS-A deployment.


gwesterfieldjr:
All the lint errors are not on my changes...Looks like they were introduced after I rebased. Is something up with the linter?

codecov.yml
```diff
@@ -11,6 +11,7 @@ coverage:
 
 ignore:
   - internal/test/e2e
+  - internal/pkg/api
```
gwesterfieldjr, line 14: Ignoring this as its is all used by internal/test/e2e which is already ignored

pkg/executables/govc.go
```diff
@@ -854,13 +854,15 @@ func (g *Govc) GetTags(ctx context.Context, path string) ([]string, error) {
 	return tags, nil
 }
 
-type tag struct {
+// Tag struct to represent a vSphere Tag.
+type Tag struct {
```
TerryHowe, line 858: why is this public now? It would be nice if this stayed private

pkg/providers/vsphere/internal/tags/factory.go
```diff
@@ -13,7 +14,7 @@ type Factory struct {
 }
 
 type GovcClient interface {
-	ListTags(ctx context.Context) ([]string, error)
+	ListTags(ctx context.Context) ([]executables.Tag, error)
```
TerryHowe, line 17: Oh, this is probably why Tag is public

pkg/providers/vsphere/validator.go
```diff
@@ -90,7 +90,31 @@ func (v *Validator) ValidateVCenterConfig(ctx context.Context, datacenterConfig
...
-// TODO: dry out machine configs validations.
+func (v *Validator) validateMachineConfigTagsExist(ctx context.Context, machineConfigs []*anywherev1.VSphereMachineConfig) error {
+	tags, err := v.govc.ListTags(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to check if tags exists in vSphere: %v", err)
+	}
+
+	tagIds := make([]string, 0, len(tags))
+	for _, t := range tags {
+		tagIds = append(tagIds, t.Id)
```
TerryHowe, line 101: I mean if this was it for tags, the strings would have worked, no?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -12,23 +12,22 @@ import (
...
 
 	// ControlPlaneAnnotation is an annotation that can be applied to EKS-A machineconfig
-	// object to prevent a controller from making changes to that resource.
+	// Object to prevent a controller from making changes to that resource.
 	controlPlaneAnnotation = "anywhere.eks.amazonaws.com/control-plane"
 
 	clusterResourceType = "clusters.anywhere.eks.amazonaws.com"
 
-	// etcdAnnotation can be applied to EKS-A machineconfig CR for etcd, to prevent controller from making changes to it.
+	// EtcdAnnotation Can be applied to EKS-A machineconfig CR for etcd, to prevent controller from making changes to it.
```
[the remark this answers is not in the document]
gwesterfieldjr: Yea the stupid linter kept failing unless I made all comments start with a capitol letter. so annoying lol

pkg/providers/vsphere/validator.go
gwesterfieldjr: No because there are other areas looping on tag name whereas I need Id so I exposed whole tag struct so each case can get the value it needs.

pkg/providers/vsphere/internal/tags/factory.go
gwesterfieldjr: Yup!

TerryHowe:
Anything worth saving here?

TerryHowe:
Ah, I see Abhinav finished this work.

PR #159, opened 2021-09-09
Shuffle around doc content

TerryHowe:
* The ovas.md is supposed to be a generated list of OVAs.
* Moved the contents from that file into the artifacts file
* Added link for EKS-A ECR repo
* Added link for EKS-D ECR repo

PR #3799, opened 2022-10-21
Add E2E test case for CloudStack node failuredomain label and node name matche

wanyufe:
*Issue #, if available:*
Need E2E test cases to test node labels in control plane and worker node group for CloudStack cluster.
Need E2E test cases to test node name matches for CAPI machine name.
*Description of changes:*
Add E2E test cases to test node labels in control plane and worker node group for CloudStack cluster.
Add E2E test cases to test node name matches for CAPI machine name.
*Testing (if applicable):*
```
2022-10-20T22:40:17.692-0400	V3	Cleaning up long running container	{"name": "eksa_1666319446710418000"}
2022-10-20T22:40:18.880-0400	V4	e2e	Validating control plane failuredomain labels
2022-10-20T22:40:18.880-0400	V4	e2e	node label: 	{"cluster.x-k8s.io/failure-domain": "az-zone1"}
2022-10-20T22:40:19.988-0400	V4	e2e	Validating worker node failuredomain labels	{"worker node group": "md-0"}
2022-10-20T22:40:19.988-0400	V4	e2e	node label: 	{"cluster.x-k8s.io/failure-domain": "az-zone1"}
    cluster.go:516: Validating cluster node status
    cluster.go:528: Validating cluster node version
    cluster.go:535: Validating cluster node and CAPI machine name match
2022-10-20T22:40:21.413-0400	V4	e2e	check node and machine name match: 	{"node name": "eksa-test-3fbdd0f-4gzlg", "CAPI machine name": "eksa-test-3fbdd0f-4gzlg"}
2022-10-20T22:40:21.414-0400	V4	e2e	check node and machine name match: 	{"node name": "eksa-test-3fbdd0f-md-0-74b8f9f8cd-cc5xb", "CAPI machine name": "eksa-test-3fbdd0f-md-0-74b8f9f8cd-cc5xb"}
```


pkg/executables/kubectl.go
```diff
@@ -1498,6 +1498,28 @@ func (k *Kubectl) ValidateNodesVersion(ctx context.Context, kubeconfig string, k
 	return nil
 }
 
+func (k *Kubectl) ValidateMachineAndNodeNamesMatch(ctx context.Context, kubeconfig string, namespace string) error {
```
maxdrib, line 1501: we should get some unit tests on here, and check the linting

test/e2e/labels_test.go
```diff
@@ -4,6 +4,7 @@
 package e2e
 
 import (
+	"github.com/aws/eks-anywhere/pkg/constants"
```
maxdrib, line 7: I think this import should be lower, with the other github imports

test/e2e/cloudstack_test.go
```diff
@@ -0,0 +1,35 @@
+//go:build e2e
+// +build e2e
+
```
maxdrib, line 3: I'm not a fan of this file name. It's too generic. Why can't we put this test in the labels test file like it was before? That seemed fitting. We could just add a comment indicating it's a specific feature for the CloudStack provider

pkg/constants/constants.go
```diff
@@ -43,6 +43,11 @@ const (
 	DefaultRegistry            = "public.ecr.aws"
 	CloudstackAnnotationSuffix = "cloudstack.anywhere.eks.amazonaws.com/v1alpha1"
 
+	FailuredomainLabelName = "cluster.x-k8s.io/failure-domain"
```
maxdrib, line 46:
```suggestion
	FailureDomainLabelName = "cluster.x-k8s.io/failure-domain"
```

pkg/constants/constants.go
```diff
@@ -43,6 +43,11 @@ const (
...
+
+	// provider specific keywork placeholder
+	CloudstackFailuredomainPlaceholder = "ds.meta_data.failuredomain"
```
maxdrib, line 49:
```suggestion
	CloudstackFailureDomainPlaceholder = "ds.meta_data.failuredomain"
```

test/framework/labels.go
```diff
@@ -18,11 +19,31 @@ func ValidateControlPlaneLabels(controlPlane v1alpha1.ControlPlaneConfiguration,
 	return validateLabels(controlPlane.Labels, node)
 }
 
+// ValidateControlPlaneFailureDomainLabels validate if Cloudstack provider replaces ds.meta_data.failuredomain with proper failuredomain name
+// in control plane node label 'cluster.x-k8s.io/failure-domain'
+func ValidateControlPlaneFailureDomainLabels(controlPlane v1alpha1.ControlPlaneConfiguration, node corev1.Node) error {
+	if controlPlane.MachineGroupRef.Kind == "CloudStackMachineConfig" {
```
maxdrib, line 25: why do we need these checks for MachineConfig kind?

test/framework/labels.go
```diff
@@ -44,3 +65,20 @@ func retrieveTestNodeLabels(nodeLabels map[string]string) map[string]string {
 	}
 	return labels
 }
+
+func validateFailuredomainLabel(expectedLabels map[string]string, node corev1.Node) error {
```
maxdrib, line 69:
```suggestion
func validateFailureDomainLabel(expectedLabels map[string]string, node corev1.Node) error {
```

test/framework/labels.go
wanyufe: ds.meta_data.failuredomain is keyword only for cloudstack provider. For other providers like vsphere, it will not be replaced with failuredomain.

test/e2e/cloudstack_test.go
wanyufe: I changed back and forth in my local for the location of that piece of code. 

1. node name matching capi machine name is a requirements for cloudstack, other providers may not follow this rule (vsphere did follow, not sure other providers). Since it is provider specific, I choose to have a cloudstack file name, just like tinkerbell_test.go
2. label_test.go is code related to labels, so I believe the node name matching validation logic does not belong there.

failuredomain node label is label related and cloudstack specific, so I am open to put it inside cloudstack_test.go or label_test.go

Instead of cteating 2 e2e to test node label and node name separately, I combined them together into cloudstack_test.go: 
```
test.ValidateControlPlaneNodes(framework.ValidateControlPlaneFailureDomainLabels, framework.ValidateControlPlaneNodeNameMatchCAPIMachineName)
	test.ValidateWorkerNodes(framework.ValidateWorkerNodeFailureDomainLabels, framework.ValidateWorkerNodeNameMatchCAPIMachineName)
	```

File: pkg/executables/kubectl.go (hunk 1 above)
Comment (wanyufe) [reply pkg/executables/kubectl.go]: refactored code in test/framework

File: test/e2e/labels_test.go (hunk 2 above)
Comment (wanyufe) [reply test/e2e/labels_test.go]: code refactored to other places.

File: cmd/eksctl-anywhere/cmd/generateclusterconfig.go [hunk 8]
```diff
@@ -156,6 +156,7 @@ func generateClusterConfig(clusterName string) error {
 			v1alpha1.ControlPlaneConfigCount(2),
 			v1alpha1.ExternalETCDConfigCount(3),
 			v1alpha1.WorkerNodeConfigCount(2),
+			v1alpha1.WorkerNodeConfigName(constants.DefaultWorkerNodeGroupName),
```
g-gaston, line 159: I would expect all commands to remain untouched in a PR that only adds e2e tests
just for my understanding, why was this needed?
Also, aren't worker node config names set by default already?

wanyufe: Without this added line, a worker node group will not have name, and there is no way to add a label to a worker node group without name. 

I do notice that there is default logic to set a `md-0` if no name is specified, and it is invoked when running `create/upgrade cluster`. But that default logic is not invoked before e2e trying to set up the config (adding label). 

If you check the same file of other providers (docker/vsphere/snow/tinkerbell/nutanix, they all have a line like #159

test/e2e/cloudstack_test.go
maxdrib: Even though this is CloudStack specific, a lot of tests are cloudstack specific. These files are intentionally organized by feature rather than provider. Since we include the provider name in the test, that itself should be sufficiently distinguishable, together with a comment, to indicate that this feature is Cloudstack-specific.

test/framework/labels.go
maxdrib: That makes sense, but it seems like other providers wouldn't call this function in the first place, so the extra check is unnecessary.

maxdrib: In that case, we should handle the precondition failure case, and return some form of error saying it's only for CloudStack machines

test/e2e/cloudstack_test.go
maxdrib: Alright, I'll defer to @g-gaston on this one - I'm fine either way.

g-gaston: what about `cloudstack_taints_test.go`?
If the "flow" for cloudstack and other providers for this tested scenario is not the same, I'm no against having a different file. However, `cloudstack_test` seems very broad and risks becoming a huge dumping ground for all cloudstack tests. Which makes reading the tests quite difficult.

wanyufe: It is good. I can change it to cloudstack_labels_test.go - we are not testing cloudstack taint here.

maxdrib (approved):

test/framework/labels.go
```diff
@@ -18,11 +19,31 @@ func ValidateControlPlaneLabels(controlPlane v1alpha1.ControlPlaneConfiguration,
...
+
 func ValidateWorkerNodeLabels(w v1alpha1.WorkerNodeGroupConfiguration, node corev1.Node) error {
 	logger.V(4).Info("Validating worker node labels", "worker node group", w.Name)
 	return validateLabels(w.Labels, node)
 }
 
+// ValidateWorkerNodeFailureDomainLabels validate if Cloudstack provider replaces ds.meta_data.failuredomain with proper failuredomain name
+// in worker group node label 'cluster.x-k8s.io/failure-domain'.
+func ValidateWorkerNodeFailureDomainLabels(w v1alpha1.WorkerNodeGroupConfiguration, node corev1.Node) error {
+	if w.MachineGroupRef.Kind == "CloudStackMachineConfig" {
```
g-gaston, line 40: I think we have a constant for this, could we use it here?

test/framework/nodename.go
```diff
@@ -0,0 +1,40 @@
...
+		logger.V(4).Info("Validating worker node matches CAPI machine name")
+		return validateNodeNameMatchCAPIMachineName(node)
+	}
+	return nil
+}
+
+func validateNodeNameMatchCAPIMachineName(node corev1.Node) error {
+	capiMachineName, ok := node.Annotations["cluster.x-k8s.io/machine"]
+	if ok {
+		if node.Name != capiMachineName {
```
g-gaston, line 33:
```suggestion
	if ok && node.Name != capiMachineName {
```

g-gaston (approved):

test/framework/labels.go
wanyufe: updated to use constant.

test/framework/labels.go
wanyufe: add error handling for non cloudstackmachineConfig.

PR #3808, opened 2022-10-21
Remove snow machine objects exist check in controller

jiayiwang7:
*Issue #, if available:*

Change #3795 

*Description of changes:*

We already check the machine config existence in controller here: https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/snow.go#L157-L160.

*Testing (if applicable):*


taneyland:
Do we still need it here? https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/snow.go#L74

jiayiwang7:
> Do we still need it here? https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/snow.go#L74

yea we need it for cli validation since cli does not build spec from cluster

PR #3823, opened 2022-10-23
Remove vsphere provider dependency on api structs

g-gaston:
## Description of changes

It makes the code easier to follow, given the cluster spec is passed to most methods as an argument. This reduces api structs copies to one, removing ambiguity around which one contains the most updated data. It also avoids bugs where updates are only performed in one place but not the other.

If needed, a stateful version of the provider can be implemented by composing on top of this one.

This also fixes a bug where the controller rolled out new machines immediately after the cluster was created. It was detecting a change in machine configs due to a default not being applied in the cluster spec machine configs. I could have patched that alone and the change would have been smaller, but I decided to do this and avoid these kinds of bugs all together.

## Testing
* Unit tests
* Upgrade e2e test run manually

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -68,14 +68,13 @@ var (
 var requiredEnvs = []string{vSphereUsernameKey, vSpherePasswordKey, expClusterResourceSetKey}
 
 type vsphereProvider struct {
-	datacenterConfig      *v1alpha1.VSphereDatacenterConfig
-	machineConfigs        map[string]*v1alpha1.VSphereMachineConfig
 	clusterConfig         *v1alpha1.Cluster
 	providerGovcClient    ProviderGovcClient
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -695,7 +695,7 @@ func (p *vsphereProvider) GenerateCAPISpecForCreate(ctx context.Context, _ *type
 }
 
 func (p *vsphereProvider) GenerateStorageClass() []byte {
-	if p.datacenterConfig.Spec.DisableCSI {
+	if !p.csiEnabled {
```
g-gaston, line 698: This is not ideal, we should probably pass the cluster spec to this method
But that would require wider changes, affecting all providers. At least this removed the dependency and extra copies of the vsphere datacenter

vivek-koppuru (approved):

taneyland (approved):

mitalipaygude (approved):

PR #3824, opened 2022-10-24
Build Vsphere CP spec to directly access CP objects

taneyland:
*Issue #, if available:*

*Description of changes:*
Builds VSphere control plane spec by generating the control plane yaml and parsing that yaml to obtain the CP objects

*Testing (if applicable):*

### DIFF
--- a/pkg/clusterapi/name.go
+++ b/pkg/clusterapi/name.go
@@ -90,8 +90,14 @@ func clusterWorkerNodeGroupName(clusterSpec *cluster.Spec, workerNodeGroupConfig
 	return fmt.Sprintf("%s-%s", clusterSpec.Cluster.Name, workerNodeGroupConfig.Name)
 }
 
-func ControlPlaneMachineTemplateName(clusterSpec *cluster.Spec) string {
-	return DefaultObjectName(fmt.Sprintf("%s-control-plane", clusterSpec.Cluster.Name))
+// ControlPlaneMachineTemplateName sets the default object name on the control plane machine template.
+func ControlPlaneMachineTemplateName(cluster *v1alpha1.Cluster) string {
+	return DefaultObjectName(fmt.Sprintf("%s-control-plane", cluster.Name))
+}
+
+// EtcdMachineTemplateName sets the default object name on the etcd machine template.
+func EtcdMachineTemplateName(cluster *v1alpha1.Cluster) string {
+	return DefaultObjectName(fmt.Sprintf("%s-etcd", cluster.Name))
 }
 
 func WorkerMachineTemplateName(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1.WorkerNodeGroupConfiguration) string {
--- a/pkg/providers/vsphere/controlplane.go
+++ b/pkg/providers/vsphere/controlplane.go
@@ -1,10 +1,17 @@
 package vsphere
 
 import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	vspherev1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1beta1"
+	addonsv1 "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
 
 	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/clusterapi"
 	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
 	"github.com/aws/eks-anywhere/pkg/constants"
@@ -19,7 +26,7 @@ type ControlPlane struct {
 	BaseControlPlane
 	Secrets            []*corev1.Secret
 	ConfigMaps         []*corev1.ConfigMap
-	ClusterResourceSet *clusterapi.ClusterResourceSet
+	ClusterResourceSet *addonsv1.ClusterResourceSet
 }
 
 // Objects returns the control plane objects associated with the VSphere cluster.
@@ -27,7 +34,7 @@ func (p ControlPlane) Objects() []kubernetes.Object {
 	o := p.BaseControlPlane.Objects()
 	o = getSecrets(o, p.Secrets)
 	o = getConfigMaps(o, p.ConfigMaps)
-	// TODO: Get ClusterResourceSet
+	o = append(o, p.ClusterResourceSet)
 
 	return o
 }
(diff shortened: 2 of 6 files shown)


pkg/providers/vsphere/controlplane.go
```diff
@@ -52,6 +58,77 @@ func (b *ControlPlaneBuilder) BuildFromParsed(lookup yamlutil.ObjectLookup) erro
 	return nil
 }
 
+// ControlPlaneSpec builds a vsphere ControlPlane definition based on an eks-a cluster spec.
+func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes.Client, spec *cluster.Spec) (*ControlPlane, error) {
+	templateBuilder := NewVsphereTemplateBuilder(time.Now, true)
```
taneyland, line 63: I set the second parameter to `true` since this will be called from the controller, although we might be able to remove this soon since this new controller upgrades in order (control plane, etcd upgrade before worker nodes), right? @g-gaston

pkg/clusterapi/name.go
```diff
@@ -93,6 +93,11 @@ func ControlPlaneMachineTemplateName(clusterSpec *cluster.Spec) string {
 	return DefaultObjectName(fmt.Sprintf("%s-control-plane", clusterSpec.Cluster.Name))
 }
 
+// EtcdMachineTemplateName sets the default object name on the etcd machine template.
+func EtcdMachineTemplateName(clusterSpec *cluster.Spec) string {
```
g-gaston, line 97: can we take only the eks-a cluster as input? it makes the dependency more clear and it's more reusable since you don't need the whole spec

pkg/providers/vsphere/controlplane.go
g-gaston: It will follow that order indeed. However, do we need any check in between?
Also, I would like to understand why we made that cgroup part conditional in CLI. Even if it wasn't needed, did it cause any issues? Ideally the spec generated by the CLI and the controller should be the same.

pkg/providers/vsphere/controlplane.go
```diff
@@ -52,6 +58,77 @@ func (b *ControlPlaneBuilder) BuildFromParsed(lookup yamlutil.ObjectLookup) erro
...
+
+	controlPlaneYaml, err := templateBuilder.GenerateCAPISpecControlPlane(
+		spec,
+		func(values map[string]interface{}) {
+			values["controlPlaneTemplateName"] = clusterapi.ControlPlaneMachineTemplateName(spec)
+			values["etcdTemplateName"] = clusterapi.EtcdMachineTemplateName(spec)
```
g-gaston, line 69: Can we avoid passing overrides? I would like for the template builder to generate everything that is needed so we have the logic only in one place.

pkg/providers/vsphere/controlplane.go
```diff
@@ -52,6 +58,77 @@ func (b *ControlPlaneBuilder) BuildFromParsed(lookup yamlutil.ObjectLookup) erro
...
+		return nil, nil, errors.Wrap(err, "building vsphere control plane parser")
+	}
+
+	err = parser.RegisterMappings(
+		yamlutil.NewMapping(constants.SecretKind, func() yamlutil.APIObject {
+			return &corev1.Secret{}
+		}),
+		yamlutil.NewMapping(constants.ConfigMapKind, func() yamlutil.APIObject {
+			return &corev1.ConfigMap{}
+		}),
```
g-gaston, line 117: do we need mappings for CRS as well?

pkg/providers/vsphere/controlplane.go
```diff
@@ -52,6 +58,77 @@ func (b *ControlPlaneBuilder) BuildFromParsed(lookup yamlutil.ObjectLookup) erro
...
+	}
+
+	err = parser.Parse(controlPlaneYaml, builder)
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing vsphere control plane yaml")
+	}
+
+	cp := builder.ControlPlane
+
+	return cp, nil
```
g-gaston, line 88:
```suggestion
	return builder.ControlPlane, nil
```

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -105,47 +117,56 @@ func TestControlPlaneObjects(t *testing.T) {
...
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(cp).NotTo(BeNil())
+	g.Expect(cp.Cluster).To(Equal(capiCluster()))
+	g.Expect(cp.KubeadmControlPlane).NotTo(BeNil())
+	g.Expect(cp.EtcdCluster).NotTo(BeNil())
+	g.Expect(cp.ProviderCluster).NotTo(BeNil())
+	g.Expect(cp.ControlPlaneMachineTemplate.Name).To(Equal("test-control-plane-1"))
+	g.Expect(cp.EtcdMachineTemplate.Name).To(Equal("test-etcd-1"))
+	g.Expect(cp.Secrets).NotTo(BeNil())
+	g.Expect(cp.ConfigMaps).NotTo(BeNil())
```
g-gaston, line 137: Can we add `Equal` expectations for all objects in the same way we have it for `Cluster`?

pkg/providers/vsphere/controlplane.go
taneyland: Ok so this conditional case was added pre-GA, specifically for 1.20 --> 1.21 upgrades: https://github.com/aws/eks-anywhere/pull/16. Soon after this, we implemented in-order upgrades by separating the cp & md capi specs & applying the cp spec first.
So it looks like we can remove it now since we upgrade in order for both CLI and the controller.

g-gaston: Sweet
Do you know the implications of always setting the cgroup? for both CLI and controller

pkg/providers/vsphere/controlplane.go
g-gaston: I see I see. That seems quite a bit of work for little benefit. Let's chat about this offline.

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -105,47 +129,54 @@ func TestControlPlaneObjects(t *testing.T) {
...
-			},
-			ControlPlaneRef: &corev1.ObjectReference{
-				Name:       "cp",
-				Namespace:  constants.EksaSystemNamespace,
-				Kind:       "KubeadmControlPlane",
-				APIVersion: "controlplane.clusterapi.k8s/v1beta1",
-			},
-		},
-	}
+func TestControlPlaneSpecNewCluster(t *testing.T) {
```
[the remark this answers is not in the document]
taneyland: Yes, I separated that part into a separate PR. I have it ready to open this as soon as this PR merges. Includes this test case

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -10,14 +12,23 @@ import (
...
 )
 
+const (
+	testClusterConfigMainFilename = "cluster_main.yaml"
+	testDataDir                   = "testdata"
+)
+
 type baseControlPlane = clusterapi.ControlPlane[*v1beta1.VSphereCluster, *v1beta1.VSphereMachineTemplate]
 
 func TestBuildFromParsed(t *testing.T) {
```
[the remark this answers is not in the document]
taneyland: No you're right! Removed it

pkg/providers/vsphere/controlplane.go
taneyland: I'm not completely sure but from my current understanding, it seems that as long as we have in order upgrades (which is true for both the CLI and new controller), we wouldn't have to set the cgroup at all because it will auto default to the desired cgroup driver, which is systemd for k8s 1.21+.
https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/template.go#L82-L84

I can add details in the ticket to verify that with both CLI and controller.

g-gaston: Chat offline, @mitalipaygude will address this separately

g-gaston (approved):

PR #3828, opened 2022-10-24
Use dynamic affinity group and remove symlinks from test-etcd CloudStackMachineConfig

wongni:
*Issue #, if available:*

*Description of changes:*
- affinityGroupIds isn't working with multi-endpoint feature.  
- symlinks doesn't allow a trailing / and having one for a system folder such as /var/lib is dangerous

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/cloudstack/testdata/cluster_main.yaml
+++ b/pkg/providers/cloudstack/testdata/cluster_main.yaml
@@ -74,8 +74,8 @@ spec:
     label: "data_disk"
   symlinks:
     /var/log/kubernetes: /data-small/var/log/kubernetes
-  affinityGroupIds:
-  - control-plane-anti-affinity
+  affinity:
+  - pro
 ---
 apiVersion: anywhere.eks.amazonaws.com/v1alpha1
 kind: CloudStackMachineConfig
@@ -100,8 +100,8 @@ spec:
   symlinks:
     /var/log/pods: /data-small/var/log/pods
     /var/log/containers: /data-small/var/log/containers
-  affinityGroupIds:
-  - worker-affinity
+  affinity:
+  - anti
   userCustomDetails:
     foo: bar
 ---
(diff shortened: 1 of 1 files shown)


pkg/providers/cloudstack/testdata/cluster_main.yaml
```diff
@@ -74,8 +74,8 @@ spec:
     label: "data_disk"
```
maxdrib, line 74: why are we messing with this file? It's for unit testing purposes, not documentation purposes

pkg/providers/cloudstack/testdata/cluster_main.yaml
```diff
@@ -125,9 +125,5 @@ spec:
     device: "/dev/vdb"
     filesystem: "ext4"
     label: "data_disk"
-  symlinks:
-    /var/lib/: /data-small/var/lib
```
wanyufe, line 129: can we leave this symlink for etcd, and fix the trailing `/` ?

wongni: I think having a symlink for this kind of system folder is risky.

wanyufe: the idea is trying to redirect the etcd files into an attched disk offering.
/var/lib is the only settings that we tested working. If we put /var/lib/etcd , etcd will complain and refuse to start.

also, we have scripts that works with it:
```
{{- range $dir, $target := .cloudstackEtcdSymlinks}}
    - >-
      if [ ! -L {{$dir}} ] ;
        then
          mv {{$dir}} {{$dir}}-$(tr -dc A-Za-z0-9 < /dev/urandom | head -c 10) ;
          mkdir -p {{$target}} && ln -s {{$target}} {{$dir}} ;
        else
          echo "{{$dir}} already symlnk" ;
      fi
{{- end}}
```

PR #3837, opened 2022-10-25
Modify empty WorkerNodeGroupConfigs validations

d8660091:
*Issue #, if available:* Single-node feature will be GA for only Tinkerbell. We need to prevent user from running into issue of other providers

*Description of changes:* Modify the validation process to allow empty WorkerNodeGroupConfigs for only Tinkerbell provider.

*Testing (if applicable):* 
1. Unit testing
2. Manual testing that the validation fails for vSphere datacenter when WorkerNodeGroupConfigs is empty.


cxbrowne1207 (approved):
/lgtm

Issue #1571, opened 2022-03-22
Validate size of BR disk before continuing


vivek-koppuru:
We should fail earlier if the BR disk is not big enough instead of waiting until create cluster gets stuck in provisioning the nodes due to a lack of space.

vignesh-goutham:
There is a govc command `datastore.disk.info` that returns the metadata about the disk stored on vSphere datastore. This can be used to validate the disk size. Please refer here for usage - https://github.com/vmware/govmomi/blob/master/govc/USAGE.md#datastorediskinfo

rahulbabu95:
Added a validation for BR disk size with PR #5000 which will be available in release .15.0.

PR #3838, opened 2022-10-25
Validate size of BR disk before continuing

alikhan1621:
*Issue #, if available:*
#1571 
*Description of changes:*
Validating BR disk size
*Testing (if applicable):*
Tested on Vsphere client 

### DIFF
--- a/docs/content/en/docs/tasks/cluster/cluster-scale/baremetal-scale.md
+++ b/docs/content/en/docs/tasks/cluster/cluster-scale/baremetal-scale.md
@@ -30,7 +30,7 @@ spec:
 ```
 
 Next, you must ensure you have enough available hardware for the scale-up operation to function. Available hardware could have been fed to the cluster as extra hardware during a prior create command, or could be fed to the cluster during the scale-up process by providing the hardware CSV file to the upgrade cluster command (explained in detail below).
-For scale-down operation, you can skip directly to the upgrade cluster command.
+For scale-down operation, you can skip directly to the [upgrade cluster command]({{< relref "baremetal-scale/#upgrade-cluster-command-for-scale-updown" >}}).
 
 To check if you have enough available hardware for scale up, you can use the `kubectl` command below to check if there are hardware with the selector labels corresponding to the controlplane/worker node group and without the `ownerName` label. 
 
@@ -50,13 +50,7 @@ eksa-worker1                    type=worker-group-1,v1alpha1.tinkerbell.org/owne
 eksa-worker2                    type=worker-group-1
 ```
 
-If you don't have any available hardware that match this requirement in the cluster, you can [setup a new hardware CSV]({{< relref "../../../reference/baremetal/bare-preparation/#prepare-hardware-inventory" >}}). You can feed this hardware inventory file during the upgrade cluster command as shown below.
-
-#### Upgrade Cluster Command for Scale Up/Down
-
-```bash
-eksctl anywhere upgrade cluster -f cluster.yaml --hardware-csv <hardware.csv>
-```
+If you don't have any available hardware that match this requirement in the cluster, you can [setup a new hardware CSV]({{< relref "../../../reference/baremetal/bare-preparation/#prepare-hardware-inventory" >}}). You can feed this hardware inventory file during the [upgrade cluster command]({{< relref "baremetal-scale/#upgrade-cluster-command-for-scale-updown" >}}).
 
 or 
 
@@ -64,6 +58,20 @@ Prior to running the upgrade cluster command shown above, you can run the follow
 
 ```bash
 eksctl anywhere generate hardware -z <hardware.csv> | kubectl apply -f -
+```
+
+#### Upgrade Cluster Command for Scale Up/Down
+
+##### With Hardware CSV File
+
+```bash
+eksctl anywhere upgrade cluster -f cluster.yaml --hardware-csv <hardware.csv>
+```
+
+##### Without Hardware CSV File
+
+```bash
+eksctl anywhere upgrade cluster -f cluster.yaml
 ```
 
 ### Autoscaling


pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
-	logger.V(4).Info("Marking vm as template", "templateName", templateFullPath)
-	if err := g.markVMAsTemplate(ctx, datacenter, templateFullPath); err != nil {
-		return err
+func parseDiskSize(str string) int {
+	var nonAlphanumericRegex = regexp.MustCompile(`[^0-9]+`)
+	replacedString := nonAlphanumericRegex.ReplaceAllString(str, "")
+
+	i, err := strconv.Atoi(replacedString)
+	if err != nil {
+		// ... handle error
```
vignesh-goutham, line 403: please remove this comment

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+
+				if parsedDiskSize == -1 {
+					logger.V(2).Info("Unable to get converted KB to GB int")
+					return fmt.Errorf("unable to get converted KB to GB int: %d", parsedDiskSize)
+				}
+				if disk2 != "" {
+					//check if disk1 is 20gb
+					if parsedDiskSize != 20 {
+						logger.V(2).Info("Unable to resize disk1 to 20 GB")
+						return fmt.Errorf("error resizing disk1 to 20 GB - new size is: %d", parsedDiskSize)
```
vignesh-goutham, line 380: As a general standard, we avoid having the word "error" in fmt.Errorfs. This would anyways print "Error: <message>". The fmt error above this is a good example. This could be `resizing disk1 to 20 GB - actual size is: %d`

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+				parsedDiskSize := parseDiskSize(deviceDiskSize)
+
+				if parsedDiskSize == -1 {
+					logger.V(2).Info("Unable to get converted KB to GB int")
+					return fmt.Errorf("unable to get converted KB to GB int: %d", parsedDiskSize)
+				}
+				if disk2 != "" {
+					//check if disk1 is 20gb
+					if parsedDiskSize != 20 {
+						logger.V(2).Info("Unable to resize disk1 to 20 GB")
```
vignesh-goutham, line 379: We can skip redundant log statements, when we are going to error anyways.

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+						return fmt.Errorf("error resizing disk1 to 20 GB - new size is: %d", parsedDiskSize)
+					}
+				} else if disk1 != "" {
+					//check if disk2 is 22 gb
+					if parsedDiskSize != 23 {
```
vignesh-goutham, line 384:
```suggestion
					if parsedDiskSize != 22 {
```

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+						logger.V(2).Info("Unable to resize disk2 to 22 GB")
+						return fmt.Errorf("error resizing disk2 to 22 GB - new size is: %d", parsedDiskSize)
+					}
+				} else {
+					return fmt.Errorf("template %v is not valid as there are no associated disks", templateName)
```
vignesh-goutham, line 389:
```suggestion
					return fmt.Errorf("invalid template %v, no associated disks found", templateName)
```

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+
+	i, err := strconv.Atoi(replacedString)
+	if err != nil {
+		// ... handle error
+		fmt.Errorf("error converting string to int: %v", err)
+		return -1
 	}
 
-	return nil
+	return i / 1000000
```
vignesh-goutham, line 408: Correct me if I'm wrong here. Shouldn't this be 1024 * 1024?

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -66,15 +73,27 @@ func (f *Factory) CreateIfMissing(ctx context.Context, datacenter string, machin
...
+
+		templateName := filepath.Base(templateFullPath)
+		if strings.EqualFold(string(osFamily), string(v1alpha1.Bottlerocket)) {
+			resizeBRDisk = true
+		}
+		if resizeBRDisk {
+			if err1 := f.resizeDisk(ctx, templateFullPath, templateName, f.datacenter); err1 != nil {
+				return fmt.Errorf("error resizing disk with template: %v", err1)
+			}
+		}
```
vignesh-goutham, line 91: So, I think we should not resize the disk just if the template is Bottlerocket and especially if the user had imported it. We actually have proper documentation that asks users to issue resize commands while manually importing. We should just check if the template in input has the right size. This is primarily because the user imported it, and EKS-A editing the template that user imported would not be right.

vignesh-goutham: You can probably make a function out the validate BR template disks, and re-use it.

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -148,3 +167,143 @@ func (f *Factory) importOVAIfMissing(ctx context.Context, templateName, ovaURL s
 
 	return nil
 }
+
+func (f *Factory) resizeDisk(ctx context.Context, templateFullPath, templateName, datacenter string) error {
```
vignesh-goutham, line 171: From the above comment, lets change this to just have a function to validate and not resize user imported templates. For the validation function too, lets try and have one function which we can re-use.

vignesh-goutham (changes requested):
Good start, have few comments below!

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
 	}
+	return nil
+}
 
-	logger.V(4).Info("Marking vm as template", "templateName", templateFullPath)
-	if err := g.markVMAsTemplate(ctx, datacenter, templateFullPath); err != nil {
-		return err
+func parseDiskSize(str string) int {
+	var nonAlphanumericRegex = regexp.MustCompile(`[^0-9]+`)
+	replacedString := nonAlphanumericRegex.ReplaceAllString(str, "")
```
g-gaston, line 399: Can you use capturing group for this?

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -66,15 +73,27 @@ func (f *Factory) CreateIfMissing(ctx context.Context, datacenter string, machin
...
 	}
+	osFamily := machineConfig.Spec.OSFamily
+	resizeBRDisk := false
+
 	if err == nil && len(templateFullPath) > 0 {
 		machineConfig.Spec.Template = templateFullPath // TODO: move this out of the factory into the defaulter, it's a side effect
 		logger.V(2).Info("Template already exists. Skipping creation", "template", machineConfig.Spec.Template)
+
+		templateName := filepath.Base(templateFullPath)
+		if strings.EqualFold(string(osFamily), string(v1alpha1.Bottlerocket)) {
```
g-gaston, line 84: you can compare osFamily and `v1alpha1.Bottlerocket` directly

pkg/executables/govc.go
```diff
@@ -326,36 +337,76 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
 
-	logger.V(4).Info("Taking template snapshot", "templateName", templateFullPath)
-	if err := g.createVMSnapshot(ctx, datacenter, templateFullPath); err != nil {
-		return err
+		for _, deviceInfo := range devicesInfo2.([]interface{}) {
+			deviceMetadata := deviceInfo.(map[string]interface{})["DeviceInfo"]
+			deviceLabel := deviceMetadata.(map[string]interface{})["Label"].(string)
+			deviceDiskSize := deviceMetadata.(map[string]interface{})["Summary"].(string)
+
+			if strings.EqualFold(deviceLabel, "Hard disk 1") || strings.EqualFold(deviceLabel, "Hard disk 2") {
```
g-gaston, line 369: is there a real reason to use `EqualFold` all the time or is just in case?

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -66,15 +73,27 @@ func (f *Factory) CreateIfMissing(ctx context.Context, datacenter string, machin
 	if err != nil {
 		return fmt.Errorf("checking for template: %v", err)
 	}
+	osFamily := machineConfig.Spec.OSFamily
+	resizeBRDisk := false
```
g-gaston, line 77: why define a flag that is only used in one place instead of just checking the condition when you do the branching?

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -148,3 +167,143 @@ func (f *Factory) importOVAIfMissing(ctx context.Context, templateName, ovaURL s
...
+		diskName = disk1
+		diskSizeInGB = 22
+	} else {
+		return fmt.Errorf("template %v is not valid as there are no associated disks", templateName)
+	}
+
+	//logger.V(2).Info("VM Info. ", "vm info", response.String())
+
+	logger.V(2).Info("Template Name. ", "Template name: ", templateName)
+	logger.V(2).Info("Printing diskSize for resize: ", "Disk Size: ", diskSizeInGB)
```
g-gaston, line 242: spaces are not needed at the end of the keys
I would suggest checking other logs in the codebase :)

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -148,3 +167,143 @@ func (f *Factory) importOVAIfMissing(ctx context.Context, templateName, ovaURL s
...
+		logger.V(2).Info("Resizing disk 2 of template to 20G")
+		diskName = disk2
+		diskSizeInGB = 20
+	} else if disk1 != "" {
+		logger.V(2).Info("Resizing disk 1 of template to 22G")
+		diskName = disk1
+		diskSizeInGB = 22
+	} else {
+		return fmt.Errorf("template %v is not valid as there are no associated disks", templateName)
+	}
```
g-gaston, line 237: all this logic seems very similar to the one in govc.go
Assuming this is necessary, could we get a small writeup in the PR description about what all of this is doing and why is necessary? It might help understand the implementation :)

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -148,3 +167,143 @@ func (f *Factory) importOVAIfMissing(ctx context.Context, templateName, ovaURL s
...
+		return err
+	}
+
+	logger.V(4).Info("Marking vm as template", "templateName", templateFullPath)
+	if err := f.client.MarkVMAsTemplate(ctx, datacenter, templateFullPath); err != nil {
+		return err
+	}
+
+	return nil
+}
```
g-gaston, line 296: I would suggest revisiting this whole method. It's quite lengthy and has a lot of nested branches, it could use a small refactor

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -39,6 +41,11 @@ type GovcClient interface {
 	AddTag(ctx context.Context, path, tag string) error
 	ListCategories(ctx context.Context) ([]string, error)
 	CreateCategoryForVM(ctx context.Context, name string) error
+	DevicesInfo(ctx context.Context, datacenter, templateName string) (interface{}, error)
+	ResizeDisk(ctx context.Context, datacenter, template, diskName string, diskSizeInGB int) error
+	Execute(ctx context.Context, command, datacenter, templateName string) error
```
g-gaston, line 46: is this been used?

g-gaston (commented):
Thanks for doing this! A few suggestions

pkg/providers/vsphere/vsphere.go
```diff
@@ -105,6 +105,12 @@ type ProviderGovcClient interface {
 	AddTag(ctx context.Context, path, tag string) error
 	ListCategories(ctx context.Context) ([]string, error)
 	CreateCategoryForVM(ctx context.Context, name string) error
+	DevicesInfo(ctx context.Context, datacenter, templateName string) (interface{}, error)
+	ValidateSize(ctx context.Context, datacenter, templateName string) error
```
chrisdoherty4, line 109: This doesn't strike me as the right place to do validation.

I'd think the only thing the vSphere client is used for is retrieving data and formatting it in a model of some kind. That model is then validated by logic residing in the vSphere package.

pkg/providers/vsphere/vsphere.go
```diff
@@ -694,12 +700,11 @@ func (p *vsphereProvider) GenerateCAPISpecForCreate(ctx context.Context, _ *type
...
 }
 
-func (p *vsphereProvider) InstallStorageClass(ctx context.Context, cluster *types.Cluster) error {
+func (p *vsphereProvider) GenerateStorageClass() []byte {
 	if !p.csiEnabled {
 		return nil
 	}
-
-	return p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, defaultStorageClass)
+	return defaultStorageClass
```
chrisdoherty4, line 707: This looks like an incorrectly resolved diff. `GenerateStorageClass` has gone.

pkg/providers/vsphere/internal/templates/factory.go
```diff
@@ -66,15 +68,29 @@ func (f *Factory) CreateIfMissing(ctx context.Context, datacenter string, machin
 	if err != nil {
```
chrisdoherty4, line 68: I followed the code paths through and it seems rather odd to be validating if an existing template has appropriately sized disks in a `CreateIfMissing` function. I don't think this really has any business validating existing templates. 

`CreateIfMissing` is called in [1 place only](https://github.com/aws/eks-anywhere/blob/24b3b939f727753d23f81f28cbbd7f7361f2a710/pkg/providers/vsphere/defaults.go#L95). I wonder if, right after calling `CreateIfMissing` we should validate the disk. 

We can optimize further by changing `CreateIfMissing` to just `Create` and return an error that can be inspected. E.g. `ErrTemplateExists` and `errors.Is(err, templates.ErrTemplateExists)`. With that error, we can choose to ignore it and then validate the 'existing' template. 

These changes would make the code far more explicit in its intent.

PR #3848, opened 2022-10-28
Add chart and version options to package install

TerryHowe:


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -77,14 +77,19 @@ func NewPackageControllerClient(chartInstaller ChartInstaller, kubectl KubectlRu
 //
 // In case the cluster is a workload cluster, it performs the following actions:
 //   - Creation of package bundle controller (PBC) custom resource in management cluster
-func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) error {
```
lewisdiamond, line 80: We should keep the previous version of this function and add another one with the two additional parameters instead to avoid having to pass `"", ""` most of the times.

TerryHowe: There are two places where this method is called that are not tests and I've changed one of them in this PR. If this flies I was going to talk about changing the other, but it requires a cluster spec change, so I did not include it in this PR.

TerryHowe:
Going with some other approach

PR #3852, opened 2022-10-30
Worker count pointer dereference fix

pokearu:
*Description of changes:*
The `workerNodeGroupConfiguration.Count` was made an Integer pointer #3624. Adding a fix for places where the pointer was not dereferenced when the value needed to be compared.


abhay-krishna (approved):
/approve

PR #3860, opened 2022-10-31
Fix bug in controller vsphere template generator

g-gaston:
## Description of changes

Avoid overwriting ssh keys for CP and etcd machines since the vsphere template builder is already capable of doing it and runs more robust logic. This was causing new machine rollouts when the specified ssh key had a comment due a difference in kubeadm control plane, etcdadm cluster and vsphere machine templates. The CLI strips them out before creating the vsphere capi objects because it uses the template builder logic.

This was causing some e2e tests to fail due to cluster instability: the controller kept rolling out new machines so the cluster didn't become ready fast enough and the delete operation timed out waiting for it. 

## Testing
Unit tests and manual e2e tests

### DIFF
--- a/controllers/resource/reconciler_test.go
+++ b/controllers/resource/reconciler_test.go
@@ -14,9 +14,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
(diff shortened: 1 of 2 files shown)


taneyland (approved):

controllers/resource/reconciler_test.go
```diff
@@ -248,6 +252,21 @@ func TestClusterReconcilerReconcileVSphere(t *testing.T) {
 							t.Errorf("unmarshal failed: %v", err)
 						}
 						assert.Equal(t, expectedMCDeployment, template, "values", expectedMCDeployment, template)
+					case "KubeadmControlPlane":
+						kubeadmControlPlane := &controlplanev1.KubeadmControlPlane{}
+						assert.NoError(t, runtime.DefaultUnstructuredConverter.FromUnstructured(template.Object, kubeadmControlPlane))
```
ewollesen, line 257: You might consider using `require.NoError` here, as that will end the test, preventing the user from having a slew of errors that aren't really errors, they just all failed because this failed.

ewollesen: Alternatively, one can do:

```go
if assert.NoError(t, blah) {
  assert.Equal(t, strippedSSHKey, ...)
}
```

But a lot of people understandably don't like having branching logic (even simple branching logic) in their test checks.

PR #3866, opened 2022-10-31
Docker single node

tatlat:
*Issue #, if available:*

*Description of changes:*
Pass empty taints (taints = []) to cluster-api when workerNodeGroupConfigurations is empty

*Testing (if applicable):*
Unit test and manually creating a docker single-node cluster.


pkg/providers/docker/config/template-cp.yaml
```diff
@@ -201,7 +204,10 @@ spec:
 {{- if .kubeletExtraArgs }}
 {{ .kubeletExtraArgs.ToYaml | indent 10 }}
 {{- end }}
```
TerryHowe, line 206: This is it? Seems like an else would be nice here, but not a deal breaker for me

TerryHowe:
Fix these tests maybe?

TerryHowe (approved):
/lgtm

PR #3870, opened 2022-10-31
Refactoring cloudmonkey executable & validator to support running from EKS-A controller

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR refactors the cloudmonkey executable to support initialization in the EKS-A controller, where it needs to use the K8s secrets on the cluster, instead of the local environment variable passed in. This is a prerequisite change to support the cluster lifecycle api feature in the EKS-A controller.

*Testing (if applicable):*
* Unit tests pass
* e2e tests TestCloudStackKubernetes120RedhatTo121Upgrade and TestCloudStackKubernetes121WorkloadCluster pass


pkg/dependencies/factory.go
```diff
@@ -504,20 +504,47 @@ func (f *Factory) WithGovc() *Factory {
...
+
+		return nil
+	})
 
-		f.dependencies.Cmk = f.executablesConfig.builder.BuildCmkExecutable(f.dependencies.Writer, execConfig.Profiles)
+	return f
+}
+
+// WithCmkFromSecrets initializes an empty Cmk object to make it available in the controller, which will set the profiles from K8s secrets.
+func (f *Factory) WithCmkFromSecrets() *Factory {
```
g-gaston, line 533: How is the controller going to set the config in the cmk executable?
Would that happen on every reconciliation request? That wouldn't be thread safe

If the cmk executable is designed to only operate on one set of profiles, then I think we should build a new cmk executable on every reconciliation request. In order to do that, we can have a `CmkFactory` that implements a `CloudstackClientRegistry`. From the controller perspective it would be something like `cloudstackClientRegistry.Get(cloudStackExecConfigFromSecret)`. And such factory would just know how to create the cmk executable. Which means it probably needs the executable builder and the writer.

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -16,10 +17,53 @@ const (
...
+	SecretKeyKey = "secret-key"
+	APIUrlKey    = "api-url"
+	VerifySslKey = "verify-ssl"
 )
 
-// ParseCloudStackSecret parses the input b64 string into the ini object to extract out the api key, secret key, and url.
-func ParseCloudStackSecret() (*CloudStackExecConfig, error) {
+// ParseCloudStackCredsFromSecrets parses a list of secrets to extract out the api keys, secret keys, and urls.
+func ParseCloudStackCredsFromSecrets(secrets []apiv1.Secret) (*CloudStackExecConfig, error) {
+	var cloudstackProfiles []CloudStackProfileConfig
```
g-gaston, line 29: can we set the size here if we know it?

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -16,10 +17,53 @@ const (
...
+		cloudstackProfiles = append(cloudstackProfiles, CloudStackProfileConfig{
+			Name:          secret.Name,
+			ApiKey:        string(apiKey),
+			SecretKey:     string(secretKey),
+			ManagementUrl: string(apiURL),
+			VerifySsl:     string(verifySsl),
+		})
+	}
+
+	if len(cloudstackProfiles) == 0 {
```
g-gaston, line 56: could make this check at the beginning by checking the secret slice len?

controllers/factory.go
```diff
@@ -73,7 +73,7 @@ func (f *Factory) Build(ctx context.Context) (*Reconcilers, error) {
 }
 
 func (f *Factory) WithClusterReconciler(capiProviders []clusterctlv1.Provider) *Factory {
-	f.dependencyFactory.WithGovc()
+	f.dependencyFactory.WithGovc().WithCmkFromSecrets()
```
g-gaston, line 76: I think this would go in `withCloudstackClusterReconciler` since it's a cloudstack only dependency

pkg/providers/cloudstack/decoder/decoder.go
maxdrib: that will mean we can no longer use append, right? Instead we have to set using slices? What's the purpose? It seems like that introduces more risk of some off-by-one out of bounds error in the future

controllers/factory.go
maxdrib: Sounds good, will remove. I was going off the Govc example.

pkg/providers/cloudstack/decoder/decoder.go
maxdrib: Sure

pkg/providers/cloudstack/validator_registry.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/networkutils"
+	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
+)
+
+// ValidatorRegistry exposes a single method for retrieving the CloudStack validator, and abstracts away how they are injected.
+type ValidatorRegistry interface {
+	Get(execConfig *decoder.CloudStackExecConfig) (*Validator, error)
```
maxdrib, line 14: I was unable to define a generic Validator interface for mocking because it resulted in an import cycle:

```
# github.com/aws/eks-anywhere/pkg/providers/cloudstack
package github.com/aws/eks-anywhere/pkg/providers/cloudstack
	imports github.com/aws/eks-anywhere/pkg/providers/cloudstack/mocks: import cycle not allowed in test
FAIL	github.com/aws/eks-anywhere/pkg/providers/cloudstack [setup failed]
```

pkg/providers/cloudstack/decoder/decoder.go
```diff
@@ -16,10 +17,52 @@ const (
...
+			verifySsl = []byte(defaultVerifySslValue)
+		}
+		cloudstackProfiles[idx] = CloudStackProfileConfig{
+			Name:          secret.Name,
+			ApiKey:        string(apiKey),
+			SecretKey:     string(secretKey),
+			ManagementUrl: string(apiURL),
+			VerifySsl:     string(verifySsl),
+		}
+	}
```
g-gaston, line 57: If you want to use append
```suggestion
	cloudstackProfiles := make([]CloudStackProfileConfig, 0, len(secrets))
	for idx, secret := range secrets {
		apiKey, ok := secret.Data[APIKeyKey]
		if !ok {
			return nil, fmt.Errorf("secret %s is missing required key %s", secret.Name, APIKeyKey)
		}
		secretKey, ok := secret.Data[SecretKeyKey]
		if !ok {
			return nil, fmt.Errorf("secret %s is missing required key %s", secret.Name, SecretKeyKey)
		}
		apiURL, ok := secret.Data[APIUrlKey]
		if !ok {
			return nil, fmt.Errorf("secret %s is missing required key %s", secret.Name, APIUrlKey)
		}
		verifySsl, ok := secret.Data[VerifySslKey]
		if !ok {
			verifySsl = []byte(defaultVerifySslValue)
		}
		cloudstackProfiles = append(
		    cloudstackProfiles,
		    CloudStackProfileConfig{
			   Name:          secret.Name,
			   ApiKey:        string(apiKey),
			   SecretKey:     string(secretKey),
			   ManagementUrl: string(apiURL),
			   VerifySsl:     string(verifySsl),
		   },
		)
	}
```

pkg/providers/cloudstack/validator_registry.go
g-gaston: Yeah you would need to define the validator interface in this package

pkg/providers/cloudstack/validator_registry.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// ValidatorFactory implements the ValidatorRegistry interface and holds the necessary structs for building fresh Validator objects.
+type ValidatorFactory struct {
+	builder     *executables.ExecutablesBuilder
```
g-gaston, line 19: can we make this an interface so we don't depend on the executables package?

pkg/providers/cloudstack/validator_registry_test.go
```diff
@@ -0,0 +1,36 @@
...
+	g := NewGomegaWithT(t)
+	registry := cloudstack.NewValidatorFactory(executables.NewLocalExecutablesBuilder(), nil, false)
+	_, err := getFromRegistry(registry, nil)
+	g.Expect(err).NotTo(BeNil())
+}
+
+func TestRegistryGetSuccess(t *testing.T) {
+	g := NewGomegaWithT(t)
+	registry := cloudstack.NewValidatorFactory(executables.NewLocalExecutablesBuilder(), nil, false)
+	_, err := getFromRegistry(registry, &decoder.CloudStackExecConfig{Profiles: []decoder.CloudStackProfileConfig{
```
g-gaston, line 27: can we check the returned value it not nil and it's a valid validator struct?

pkg/providers/cloudstack/validator.go
```diff
@@ -45,7 +45,9 @@ type localAvailabilityZone struct {
 	DomainId string
 }
 
+// ProviderCmkClient defines the methods used by Cmk as a separate interface to be mockable when injected into other objects.
 type ProviderCmkClient interface {
+	Close(ctx context.Context) error
```
g-gaston, line 50: what is this needed for?

maxdrib: This was added when I was working on mocking the Cmk client. Will remove.

pkg/providers/cloudstack/validator_registry_test.go
maxdrib: I'm not sure how we could check that it's a valid validator struct, but we can definitely check not nil

pkg/providers/cloudstack/validator_registry.go
maxdrib: I'm not sure I understand your suggestion. How do we remove the dependency on the executables package?

pkg/providers/cloudstack/decoder/decoder.go
maxdrib: Sounds good, thanks. Could you explain the value of using this approach over the original one?

pkg/providers/cloudstack/validator_registry.go
g-gaston: Ahh I see, yeah, that's another issue of having the tests in the same package as the code, we should move away from that.

In the meantime, you can create the mocks in the cloudstack package, that should bypass the circular dependency.

pkg/providers/cloudstack/decoder/decoder.go
g-gaston: If with the original one you mean the one you had originally, the advantage is you avoid multiple array resizing (which involves memory allocation and memory copy) and your slice only takes the memory it needs (the auto sizing algorithm doesn't know how many elements you need so it might overshoot the size.

If you are asking compared to your current code, using `slice[position` assignment, I think what I propose is simpler (since you avoid caring about positioning), less prone to errors and completely removes the possibility of having an out of bounds panic bug.

pkg/providers/cloudstack/validator_registry.go
g-gaston: ahh yeah
```go

type ClientBuilder interface {
    BuildCloudtsackClient(writer filewriter.FileWriter, skipIPCheck bool) Client
}

type ValidatorFactory struct {
	builder     ClientBuilder
	writer      filewriter.FileWriter
	skipIPCheck bool
}
```

pkg/providers/cloudstack/validator_registry_test.go
g-gaston: If this was returning an interface, which I think it should, just making a type assertion

pkg/providers/cloudstack/validator_registry.go
maxdrib: Created https://github.com/aws/eks-anywhere/issues/3921 to address the unit test refactoring work. I don't actually need mocks for this validator currently, so maybe that would be something that could be done when the v2 controller is ready to be reviewed.

pkg/providers/cloudstack/validator_registry.go
maxdrib: I was not able to get rid of the executables dependency. When I tried using the ProviderCmk interface (where you refer to `Client`), the ClientBuilder did not implement the interface properly

```
Cannot use 'builder' (type *executables.ExecutablesBuilder) 
as the type ClientBuilder Type does not implement 'ClientBuilder' 
need the method: BuildCmkExecutable(writer filewriter.FileWriter, config *decoder.CloudStackExecConfig) (ProviderCmkClient, error) 
have the method: BuildCmkExecutable(writer filewriter.FileWriter, config *decoder.CloudStackExecConfig) (*Cmk, error)
```

Using the original `Cmk` interface retains the dependency on the executables package, and plumbing down the ProviderCmk interface into the executables package doesn't make sense either.

g-gaston: > plumbing down the ProviderCmk interface into the executables package doesn't make sense either

Ideally we shouldn't need to do that but IMO it's the lesser of two evils. I prefer to import an interface defined in the cloudstack package from the executables builder than using an executables concrete implementation from the cloudstack package. This follows inwards pointing dependency rule (your details [client implementation] can depend on your business logic but not the other way around).

If we want to avoid this it would require deeply refactoring the executables package (which I think it's too big), but that's another topic :)

Btw I would create a new method, keep `BuildCmkExecutable` as is and have a `BuildCloustackClient` that returns the interface.

pkg/providers/cloudstack/validator_registry_test.go
g-gaston: Yeah, I think it should return an interface. If not, you can't mock it in the controller and you are back to mocking the underlying client.

maxdrib: Done. But the linter complained when I added the type assertion, saying 
```
pkg/providers/cloudstack/validator_registry_test.go:36:11: S1040: type assertion to the same type: validator already has type cloudstack.ProviderValidator (gosimple)
	_, ok := validator.(cloudstack.ProviderValidator)
	         ^
```
so I removed it.

pkg/executables/cmk/cmk_builder.go
```diff
@@ -0,0 +1,23 @@
+package cmk
```
g-gaston, line 1: nit: rename file to `builder.go`

g-gaston (approved):

jaxesn (approved):
/approve
/lgtm

Issue #3856, opened 2022-10-31
Remove template autoimport from Defaulter or make it optional


jonathanmeier5:
We want to avoid the controller to perform this action. It can take a long time and sometimes requires manual intervention to cleanup. We want to ask users to provide templates in the machineconfigs. Make sure we validate this in the validation webhook.

PR #3879, opened 2022-11-01
Fail VSphereMachineConfig Create Webhook on Missing Template

jonathanmeier5:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/3856

*Description of changes:*
The current controller implementation already skips the codepath that auto-imports a template if it is not present and throws an error if the template cannot be found:
https://github.com/aws/eks-anywhere/blob/6e032751ae1a7b04e94de0c41b34ab275050e053/pkg/providers/vsphere/validator.go#L171-L171

This PR just fails the create webhook if the template field is empty so that the user does not end up in a confusing state.

*Testing (if applicable):*
Manually tested creating a workload cluster on a management cluster by kubectl apply-ing a workload cluster definition file.

### DIFF
--- a/pkg/api/v1alpha1/vspheremachineconfig.go
+++ b/pkg/api/v1alpha1/vspheremachineconfig.go
@@ -140,3 +140,11 @@ func validateVSphereMachineConfig(config *VSphereMachineConfig) error {
 
 	return nil
 }
+
+func validateVSphereMachineConfigHasTemplate(config *VSphereMachineConfig) error {
+	if config.Spec.Template == "" {
+		return fmt.Errorf("template field is required")
+	}
+
+	return nil
+}
--- a/pkg/api/v1alpha1/vspheremachineconfig_types.go
+++ b/pkg/api/v1alpha1/vspheremachineconfig_types.go
@@ -123,6 +123,13 @@ func (c *VSphereMachineConfig) Validate() error {
 	return validateVSphereMachineConfig(c)
 }
 
+// ValidateHasTemplate verifies that a VSphereMachineConfig object has a template.
+// Specifying a template is required when submitting an object via webhook,
+// as we only support auto-importing templates when creating a cluster via CLI.
+func (c *VSphereMachineConfig) ValidateHasTemplate() error {
+	return validateVSphereMachineConfigHasTemplate(c)
+}
+
 // +kubebuilder:object:generate=false
 
 // Same as VSphereMachineConfig except stripped down for generation of yaml file during generate clusterconfig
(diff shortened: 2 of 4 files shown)


pkg/api/v1alpha1/vspheremachineconfig.go
```diff
@@ -140,3 +140,11 @@ func validateVSphereMachineConfig(config *VSphereMachineConfig) error {
 
 	return nil
 }
+
+func validateVSphereMachineConfigHasTemplate(config *VSphereMachineConfig) error {
```
jonathanmeier5, line 144: Is there a more appropriate place to put this?

pkg/api/v1alpha1/vspheremachineconfig_types.go
```diff
@@ -123,6 +123,13 @@ func (c *VSphereMachineConfig) Validate() error {
 	return validateVSphereMachineConfig(c)
 }
 
+// ValidateHasTemplate verifies that a VSphereMachineConfig object has a template.
+// Specifying a template is required when submitting an object via webhook,
+// as we only support auto-importing templates when creating a cluster via CLI.
+func (c *VSphereMachineConfig) ValidateHasTemplate() error {
+	return validateVSphereMachineConfigHasTemplate(c)
```
jonathanmeier5, line 130: I really wanted to add this to `VSphereMachineConfig.Validate`, but that is not possible because we use that validate method in the CLI to semantically validate cluster config in the CLI flow.

Writing the following more for myself than anyone else:
My intuition is that we are close on this, but for future direction we may want to make an effort to clearly delineate:
- CLI cluster config semantic validation
- CLI defaulter logic
- Webhook cluster config semantic validation
- Webhook defaulter logic

A lot of things will be shared across those four bullet points.

pkg/api/v1alpha1/vspheremachineconfig_webhook.go
```diff
@@ -41,8 +41,11 @@ var _ webhook.Validator = &VSphereMachineConfig{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *VSphereMachineConfig) ValidateCreate() error {
```
jonathanmeier5, line 42: I only added this on the ValidateCreate webhook because the `Template` field is immutable on update.

pkg/api/v1alpha1/vspheremachineconfig.go
taneyland: I think it makes sense here since it's where we have the other validations

taneyland (approved):
:raised_hands:

PR #3898, opened 2022-11-01
Add vSphere workers CAPI spec generator

g-gaston:
## Description of changes
Part of the cluster lifecycle project. Generate CAPI spec in yaml with the vsphere template builder and parse it into api structs. Generate new immutable changes if a difference is detected (upgrades).

### DIFF
--- a/pkg/clusterapi/controlplane.go
+++ b/pkg/clusterapi/controlplane.go
@@ -13,7 +13,7 @@ import (
 )
 
 // ControlPlane represents the provider-specific spec for a CAPI control plane using the kubeadm CP provider.
-type ControlPlane[C, M Object] struct {
+type ControlPlane[C Object[C], M Object[M]] struct {
 	Cluster *clusterv1.Cluster
 
 	// ProviderCluster is the provider-specific resource that holds the details
--- a/pkg/clusterapi/workers.go
+++ b/pkg/clusterapi/workers.go
@@ -13,7 +13,8 @@ import (
 	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
 )
 
-type Workers[M Object] struct {
+// Workers represents the provider specific CAPI spec for an eks-a cluster's workers.
+type Workers[M Object[M]] struct {
 	Groups []WorkerGroup[M]
 }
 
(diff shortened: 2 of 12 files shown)


pkg/clusterapi/controlplane.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // ControlPlane represents the provider-specific spec for a CAPI control plane using the kubeadm CP provider
-type ControlPlane[C, M Object] struct {
+type ControlPlane[C Object[C], M Object[M]] struct {
```
jonathanmeier5, line 16: Checking my understanding here: `Object` itself is now a generic, and we pass `[C]` and `[M]` respectively to specify Object's type?

pkg/clusterapi/workers.go
```diff
@@ -45,7 +47,7 @@ type WorkerGroup[M Object] struct {
 // with the current state of the cluster. If they had, it generates a new name for them by increasing a monotonic number
 // at the end of the name.
 // This process is performed to the provider machine template and the kubeadmconfigtemplate.
-// The kubeadmconfigtemplate is not immutable at the API level but we treat is a such for consistency
+// The kubeadmconfigtemplate is not immutable at the API level but we treat is a such for consistency.
```
jonathanmeier5, line 50:
```suggestion
// The kubeadmconfigtemplate is not immutable at the API level but we treat it as such for consistency.
```

jonathanmeier5:
My understanding of this PR is that the new `WorkersSpec` function uses the old yaml-driven vsphere worker template  builder to generate a new set of CAPI objects based on an EKS-A cluster spec in the management cluster.

After generating the CAPI objects as yaml documents, it parses them into structs and diffs the generated CAPI objects against those in the cluster. If a diff is found, a new name is generated via a monotonically increasing integer appended to the name. The new object is applied, causing an "update".

This all sounds reasonable to me. Naturally I defer to others on the team who have been on the project longer.

/lgtm

pkg/clusterapi/controlplane.go
g-gaston: addressed offline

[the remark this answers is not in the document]
g-gaston: Yeah
Now the `Object` interface is typed, so I can define a `DeepCopy` object, that returns the type `C`

PR #3914, opened 2022-11-02
Fix godot linting errors

chrisdoherty4:
We recently introduced the `godot` linter that barks when comments don't end in a period. This fixes all comments in the code base to end in a period.


abhinavmpandey08 (approved):
/lgtm

PR #3919, opened 2022-11-03
docs: add us-west-2 region to the "aws ecr get-login-password" command in curated packages docs

Cajga:
*Description of changes:*
Following the example on how to login to ECR with docker does not work:
```
# aws ecr get-login-password

You must specify a region. You can also configure your region by running "aws configure".
```

It is also important to use the proper region:
```
# aws ecr get-login-password --region eu-central-1 |docker login --username AWS --password-stdin 783794618700.dkr.ecr.us-west-2.amazonaws.com
Error response from daemon: login attempt to https://783794618700.dkr.ecr.us-west-2.amazonaws.com/v2/ failed with status: 400 Bad Request
```

The solution is to use the same region where the registry resides:
```
# aws ecr get-login-password --region us-west-2 |docker login --username AWS --password-stdin 783794618700.dkr.ecr.us-west-2.amazonaws.com
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
```

The proposed PR adds this to the documentation.

### DIFF
--- a/docs/content/en/docs/tasks/packages/_index.md
+++ b/docs/content/en/docs/tasks/packages/_index.md
@@ -58,7 +58,7 @@ aws sts get-caller-identity
 Login to docker
 
 ```bash
-aws ecr get-login-password |docker login --username AWS --password-stdin 783794618700.dkr.ecr.us-west-2.amazonaws.com
+aws ecr get-login-password --region us-west-2 |docker login --username AWS --password-stdin 783794618700.dkr.ecr.us-west-2.amazonaws.com
 ```
 
 Verify you can pull an image


TerryHowe (approved):
/ok-to-test
/lgtm
/approve

Thanks!

Cajga:
@TerryHowe I've found two more issues on the same page which is coming from the fact that eks anywhere v0.12.0 does not have some flags anymore:
```
# eksctl anywhere list packages --source registry --kube-version 1.23
Error: unknown flag: --source
# eksctl anywhere generate package harbor --kube-version 1.23 > packages.yaml
Error: required flag(s) "cluster" not set
```

I've already figured out the correct commands so I could include the fix into this PR... Or do you want me to create a separate PR for these?

TerryHowe:
sorry, they'll have to go in a new PR

PR #3920, opened 2022-11-03
Update CP/etcd template names if there are changes

taneyland:
*Issue #, if available:*

*Description of changes:*
Adds the step to update CP and etcd template names if there are changes

*Testing (if applicable):*
Unit test

### DIFF
--- a/pkg/providers/vsphere/controlplane_test.go
+++ b/pkg/providers/vsphere/controlplane_test.go
@@ -2,7 +2,6 @@ package vsphere_test
 
 import (
 	"context"
-	"path"
 	"testing"
 
 	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
@@ -17,15 +16,13 @@ import (
 
 	"github.com/aws/eks-anywhere/internal/test"
 	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
-	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/clusterapi"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/providers/vsphere"
 )
 
 const (
-	testClusterConfigMainFilename = "cluster_main.yaml"
-	testDataDir                   = "testdata"
+	testClusterConfigMainFilename = "testdata/cluster_main.yaml"
 )
 
 type baseControlPlane = clusterapi.ControlPlane[*v1beta1.VSphereCluster, *v1beta1.VSphereMachineTemplate]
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/controlplane_test.go
```diff
@@ -125,6 +125,38 @@ func TestControlPlaneSpecNoKubeVersion(t *testing.T) {
...
+		vsphereMachineTemplate("test-etcd-2"),
+	)
+
+	kcp := wantKCP()
+	kcp.Spec.MachineTemplate.InfrastructureRef.Name = "test-control-plane-2"
+
+	etcd := wantEtcdCluster()
+	etcd.Spec.InfrastructureTemplate.Name = "test-etcd-3"
+
+	cp, err := vsphere.ControlPlaneSpec(ctx, logger, client, spec)
```
g-gaston, line 149: why is the name update being triggered if we are not modifying the spec?

pkg/providers/vsphere/workers.go
```diff
@@ -71,16 +70,3 @@ func machineTemplateMapping() yamlutil.Mapping[*vspherev1.VSphereMachineTemplate
 		},
 	)
 }
-
-func getMachineTemplate(ctx context.Context, client kubernetes.Client, name, namespace string) (*vspherev1.VSphereMachineTemplate, error) {
```
g-gaston, line 75: nit: Why remove them from here just to copy them to the other file?
If you don't want them to be in this file because they are shared, move them to a new third file, maybe? Not sure what we gain by just having them in the controlplane file instead of here.

pkg/providers/vsphere/controlplane_test.go
g-gaston: I would like to see the spec being modified and those changes being reflected in the returned objects.

pkg/providers/vsphere/workers.go
taneyland: It just makes more sense to me to have them defined in the CP file vs. the workers file, but from a development perspective, yeah it doesn't change anything. I added it to a new, separate file.

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -125,6 +125,46 @@ func TestControlPlaneSpecNoKubeVersion(t *testing.T) {
...
+	)
+	spec.VSphereMachineConfigs["test-etcd"].Spec.Datastore = "new-datastore"
+
+	kcp := wantKCP()
+	kcp.Spec.MachineTemplate.InfrastructureRef.Name = "test-control-plane-2"
+
+	etcd := wantEtcdCluster()
+	etcd.Spec.InfrastructureTemplate.Name = "test-etcd-3"
+
+	wantCPtemplate := vsphereMachineTemplate("test-control-plane-2")
```
g-gaston, line 150: could we do a deep copy of the original one instead of generating it again?
same for the etcd one

I think it helps understand your intent: you want the same object and modify the expected changes on top

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -125,6 +125,46 @@ func TestControlPlaneSpecNoKubeVersion(t *testing.T) {
...
+		originalEtcdCluster,
+		vsphereMachineTemplate("test-control-plane-1"),
+		vsphereMachineTemplate("test-etcd-2"),
+	)
+	spec.VSphereMachineConfigs["test-etcd"].Spec.Datastore = "new-datastore"
+
+	kcp := wantKCP()
+	kcp.Spec.MachineTemplate.InfrastructureRef.Name = "test-control-plane-2"
+
+	etcd := wantEtcdCluster()
```
g-gaston, line 147:
```suggestion
	wantEtcd := originalEtcdCluster.DeepCopy()
```

pkg/providers/vsphere/controlplane_test.go
```diff
@@ -125,6 +125,46 @@ func TestControlPlaneSpecNoKubeVersion(t *testing.T) {
...
+	originalEtcdCluster.Spec.InfrastructureTemplate.Name = "test-etcd-2"
+	client := test.NewFakeKubeClient(
+		originalKubeadmControlPlane,
+		originalEtcdCluster,
+		vsphereMachineTemplate("test-control-plane-1"),
+		vsphereMachineTemplate("test-etcd-2"),
+	)
+	spec.VSphereMachineConfigs["test-etcd"].Spec.Datastore = "new-datastore"
+
+	kcp := wantKCP()
```
g-gaston, line 144:
```suggestion
	wantKCP := originalKubeadmControlPlane.DeepCopy()
```

you probably want to do this before passing the original ones to the client

PR #3940, opened 2022-11-04
Remove controller check for vsphere template md setting cgroup driver

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
- The check was added when the controller was not performing inorder upgrades. Now that we have those, we are enabling setting the cgroup driver for both cli and controller.
- From version 1.22 kubeadm defaults the cgroup driver to systemd, so the versions check only checks if the version is 1.21 now.

Reference: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/#configuring-the-kubelet-cgroup-driver

*Testing (if applicable):*
- unit tests
- vsphere cluster creation for different versions.


taneyland (approved):

pkg/providers/vsphere/template.go
```diff
@@ -68,7 +68,7 @@ func (vs *VsphereTemplateBuilder) isCgroupDriverSystemd(clusterSpec *cluster.Spe
 	if err != nil {
 		return false, fmt.Errorf("parsing kubernetes version %v: %v", bundle.KubeDistro.Kubernetes.Tag, err)
 	}
-	if vs.fromController && k8sVersion.Major == 1 && k8sVersion.Minor >= 21 {
```
g-gaston, line 71: can we remove the `fromController` field from the template builder? :)

g-gaston (commented):
> The check was added when the controller was not performing in order upgrades. Now that we have those, this check can be removed.

This is a bit misleading. We might not need it now, we are not removing it, we are enabling it by default for both CLI and controller. very `nit`, but just don't want to leave the wrong info here in case someone comes back to revisit this PR to see what the change was about :)

pkg/providers/vsphere/template.go
```diff
@@ -23,7 +23,6 @@ func NewVsphereTemplateBuilder(
 ) *VsphereTemplateBuilder {
 	return &VsphereTemplateBuilder{
 		now:            now,
-		fromController: fromController,
```
g-gaston, line 26: sorry, to be annoying
can we remove this from the template builder constructor?

g-gaston: Also, I believe there is a test specifically for the `fromController: true` case
Can we remove that (and the expected file for it)?

g-gaston (approved):

PR #3948, opened 2022-11-04
Update reference/config/baremetal.md

d8660091:
Add instruction for setting workerNodeGroupConfigurations to nil


cxbrowne1207 (commented):
Have we made a note somewhere that single node is not supported with Kubernetes version 1.21? Additionally, maybe it would be good to add a version check and error message in the CLI for it as well.

d8660091:
> Have we made a note somewhere that single node is not supported with Kubernetes version 1.21? Additionally, maybe it would be good to add a version check and error message in the CLI for it as well.

Good point. I have added notes for not supporting Kube version <= 1.21 and created a ticket to handle version check.

Issue #3917, opened 2022-11-02
Implement the ProviderClusterReconciler interface: ControlPlane reconciliation


jonathanmeier5:
Based on the findings from the spike, we should not need the `upgrading` annotation but we need to be sure to apply the Etcd objects first during reconciliation.

jonathanmeier5:
High level design doc outlining project goals. I'm not sure how up to date this is, but it seems to contain the gist of what we are trying to do: https://github.com/aws/eks-anywhere/blob/b1fe20de2b0b24d82c63de57d4d7dddebfac5e6a/designs/full-cluster-lifecycle-api.md

Implementing this ticket means filling in this method: https://github.com/aws/eks-anywhere/blob/b1fe20de2b0b24d82c63de57d4d7dddebfac5e6a/pkg/providers/vsphere/reconciler/reconciler.go#L144-L144

Implementing the sister ticket for worker reconciliation involves filling in this method: https://github.com/aws/eks-anywhere/blob/b1fe20de2b0b24d82c63de57d4d7dddebfac5e6a/pkg/providers/vsphere/reconciler/reconciler.go#L163-L163

See the Snow reconciler for an example of what this should look like for Control Plane: https://github.com/aws/eks-anywhere/blob/1f98ebf5ff0b13a68f73b776315ff8dfc530e655/pkg/providers/snow/reconciler/reconciler.go#L78

Or for workers: https://github.com/aws/eks-anywhere/blob/1f98ebf5ff0b13a68f73b776315ff8dfc530e655/pkg/providers/snow/reconciler/reconciler.go#L103


PR with worker spec generation business logic: https://github.com/aws/eks-anywhere/pull/3898/files#diff-4fb00197119ed701845ee17f209df000d9c3099b5ee3ced0d04d7ab59e250082R28


### Reconcile Flow
- [this function](https://github.com/aws/eks-anywhere/pull/3824/files#diff-44aae5984bb86210e8481b396a2b8270c54ca9f136e264b0095e89ce141f765cR61) generates a CAPI control plane spec
[...]

PR #3950, opened 2022-11-04
Reconcile control plane in vsphere

ddjjia:
*Issue #, if available:*
Issue #3917 

*Description of changes:*
Implement the ProviderClusterReconciler interface: ControlPlane reconciliation

*Testing (if applicable):*
unit test

### DIFF
--- a/pkg/providers/vsphere/reconciler/reconciler_test.go
+++ b/pkg/providers/vsphere/reconciler/reconciler_test.go
@@ -256,8 +256,10 @@ func TestReconcilerReconcileControlPlaneSuccess(t *testing.T) {
 func TestReconcilerReconcileControlPlaneFailure(t *testing.T) {
 	tt := newReconcilerTest(t)
 	tt.createAllObjs()
+	spec := tt.buildSpec()
+	spec.Cluster.Spec.KubernetesVersion = ""
 
-	_, err := tt.reconciler().ReconcileControlPlane(tt.ctx, test.NewNullLogger(), nil)
+	_, err := tt.reconciler().ReconcileControlPlane(tt.ctx, test.NewNullLogger(), spec)
 
 	tt.Expect(err).To(HaveOccurred())
 }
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/reconciler/reconciler_test.go
```diff
@@ -149,6 +149,17 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
 
+func TestReconcilerReconcileControlPlaneSuccess(t *testing.T) {
+	tt := newReconcilerTest(t)
+	tt.createAllObjs()
+
+	result, err := tt.reconciler().ReconcileControlPlane(tt.ctx, test.NewNullLogger(), tt.buildSpec())
+
+	tt.Expect(err).NotTo(HaveOccurred())
+	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
+	tt.Expect(result).To(Equal(controller.Result{}))
```
g-gaston, line 160: I would be great if we could check that the capi objects for workers have been created :)

ddjjia: You mean check if the control plane objects are created during the test right? How do you capture a variable created in runtime without return it

g-gaston: Yeah exactly that
Ahh great question
You can use the client.Client to read it from the cluster :)

pkg/providers/vsphere/reconciler/reconciler_test.go
```diff
@@ -149,6 +150,26 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
+
+	obj := &addonsv1.ClusterResourceSet{ObjectMeta: metav1.ObjectMeta{
+		Name:      "workload-cluster-crs-0",
+		Namespace: "eksa-system",
+	}}
+	key := client.ObjectKeyFromObject(obj)
+	tt.Eventually(func() error {
+		return tt.client.Get(tt.ctx, key, obj)
+	}).Should(Succeed(), "\"object %s should eventually exist", obj.GetName())
```
g-gaston, line 170: Can we do the same thing for the other CP objects? at least the obvious ones: capi cluster, kcp and vsphere machine template
I'm approving the PR, I think you can add it as a follow up PR :)

g-gaston (approved):

taneyland (approved):

PR #3951, opened 2022-11-04
Check CP ready phase for VSphere reconciler

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/vsphere/reconciler/reconciler_test.go
```diff
@@ -86,6 +86,21 @@ func TestReconcilerFailToSetUpMachineConfigCP(t *testing.T) {
...
+		c.Name = tt.cluster.Name
+	})
+	tt.eksaSupportObjs = append(tt.eksaSupportObjs, capiCluster)
+	tt.withFakeClient()
+
+	result, err := tt.reconciler().CheckControlPlaneReady(tt.ctx, test.NewNullLogger(), tt.buildSpec())
+
+	tt.Expect(err).NotTo(HaveOccurred())
+	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
+	tt.Expect(result).To(Equal(controller.Result{}))
```
g-gaston, line 101: shouldn't this return a non 0 wait?

taneyland: No we are [setting the condition to true](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/reconciler/reconciler_test.go#L541-L543) in the capi cluster in this test file so it shouldn't return a wait time here

g-gaston: Ahh understood. I would suggest reverting the condition of this test then. I find more valuable to check that we don't continue the reconcile if the CP is not ready than the other way around. Maybe you can make that a full `Reconcile` method test instead?

g-gaston (approved):

PR #3963, opened 2022-11-07
Validate kube version when creating cluster without worker nodes

d8660091:
*Issue #, if available:* Empty worker nodes doesn't work with kube version <= 1.21, due to the limitation of cluster-api

*Description of changes:* Add kube version check when worker nodes is empty

*Testing (if applicable):* unit test


cxbrowne1207 (approved):
/lgtm

PR #3993, opened 2022-11-08
Add custom logic to reconcile a CAPI control plane

g-gaston:
## Description of changes

When using a control plane with external etcd, upgrades can't be done by just applying al changes, and depending on the changes, it needs to be orchestrated through annotations.

Marking the etcdadm cluster as upgrading, signals the kcp to not react
to changes in the etcd endpoints until the annotation is removed. This
will be done by the etcdadm cluster controller once the rolling upgrade
is finished.

Pausing the kcp reconciliation avoids having a unnecessary rolling
upgrade of the control plane when there are both cp and etcd changes
(for example, a k8s version upgrade). The CAPI cluster controller will
remove the pause annotation once the etcdadm cluster becomes ready.

### DIFF
--- a/internal/test/envtest/client.go
+++ b/internal/test/envtest/client.go
@@ -150,6 +150,22 @@ func (a *APIExpecter) ShouldEventuallyExist(ctx context.Context, obj client.Obje
 	}, a.timeout).Should(gomega.Succeed(), "object %s should eventually exist", obj.GetName())
 }
 
+// ShouldEventuallyMatch defines an eventual expectation that succeeds if the provided object
+// becomes readable by the client and matches the provider expectation before the timeout expires.
+func (a *APIExpecter) ShouldEventuallyMatch(ctx context.Context, obj client.Object, match func(g gomega.Gomega)) {
+	a.t.Helper()
+	key := client.ObjectKeyFromObject(obj)
+	a.g.Eventually(func(g gomega.Gomega) error {
+		if err := a.client.Get(ctx, key, obj); err != nil {
+			return err
+		}
+
+		match(g)
+
+		return nil
+	}, a.timeout).Should(gomega.Succeed(), "object %s should eventually match", obj.GetName())
+}
+
 // ShouldEventuallyNotExist defines an eventual expectation that succeeds if the provided object
 // becomes not found by the client before the timeout expires.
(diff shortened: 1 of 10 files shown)


internal/test/envtest/environment.go
```diff
@@ -36,8 +36,10 @@ import (
 )
 
 const (
-	capiPackage = "sigs.k8s.io/cluster-api"
-	capvPackage = "sigs.k8s.io/cluster-api-provider-vsphere"
+	capiPackage         = "sigs.k8s.io/cluster-api"
+	capdPackage         = "sigs.k8s.io/cluster-api/test"
+	capvPackage         = "sigs.k8s.io/cluster-api-provider-vsphere"
+	etcdProviderPackage = "github.com/mrajashree/etcdadm-controller"
```
taneyland, line 42: Should we point this to our build of this controller?

g-gaston: we should
but our go.mod, and all our imports still point to this personal one
So, I would vote to make that change separately

pkg/controller/clusters/controlplane.go
```diff
@@ -0,0 +1,100 @@
...
+	key := client.ObjectKey{
+		Name:      cluster.Spec.ManagedExternalEtcdRef.Name,
+		Namespace: cluster.Spec.ManagedExternalEtcdRef.Namespace,
+	}
+	if err = c.Get(ctx, key, etcdadmCluster); err != nil {
+		return controller.Result{}, errors.Wrap(err, "reading etcdadm cluster")
+	}
+
+	if !equality.Semantic.DeepDerivative(cp.EtcdCluster.Spec, etcdadmCluster.Spec) {
+		// If the etcdadm cluster has changes, this will require an rolling upgrade
```
taneyland, line 83:
```suggestion
		// If the etcdadm cluster has changes, this will require a rolling upgrade
```

internal/test/envtest/client.go
```diff
@@ -150,6 +150,22 @@ func (a *APIExpecter) ShouldEventuallyExist(ctx context.Context, obj client.Obje
 	}, a.timeout).Should(gomega.Succeed(), "object %s should eventually exist", obj.GetName())
 }
 
+// ShouldEventuallyMatch defines an eventual expectation that succeeds if the provided object
```
taneyland, line 153: Super useful testing tool!

taneyland (approved):

PR #3995, opened 2022-11-08
Adding Packages E2E test for k8s 1.23

junshun:
Adding the following test for packages E2E

1. TestCPackagesDockerUbuntuKubernetes123SimpleFlow
2. TestCPackagesVSphereKubernetes123SimpleFlow
3. TestCPackagesVSphereKubernetes123BottleRocketSimpleFlow
4. TestCPackagesVSphereKubernetes123UbuntuWorkloadCluster
5. TestCPackagesVSphereKubernetes123BottleRocketWorkloadCluster


jonahjon:
Can we also turn these on in the same PR?

jonahjon:
Otherwise these don't get run right.

junshun:
They are on by default. I believe only skipped test are called out.

PR #4007, opened 2022-11-09
Increase lint timeout from 5m to 10m

chrisdoherty4:
The golangci-lint-action has a history of being slow. We've observed the tool timeout on several occasions now and its possibly due to a defecft in the action.

https://github.com/golangci/golangci-lint-action/issues/297


g-gaston (approved):

TerryHowe (approved):
/lgtm

PR #4014, opened 2022-11-10
Add Docker ControlPlaneSpec to generate capi objects

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Added functionality to build Docker Control Plane(CP) Spec by generating the CP yaml and parsing that to obtain the corresponding objects

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/docker/controlplane.go
+++ b/pkg/providers/docker/controlplane.go
@@ -1,0 +1,77 @@
+package docker
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	dockerv1 "sigs.k8s.io/cluster-api/test/infrastructure/docker/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// ControlPlane represents a CAPI Docker control plane.
+type ControlPlane = clusterapi.ControlPlane[*dockerv1.DockerCluster, *dockerv1.DockerMachineTemplate]
+
+type controlPlaneBuilder = yamlcapi.ControlPlaneBuilder[*dockerv1.DockerCluster, *dockerv1.DockerMachineTemplate]
+
+// ControlPlaneSpec builds a docker ControlPlane definition based on an eks-a cluster spec.
+func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes.Client, spec *cluster.Spec) (*ControlPlane, error) {
+	templateBuilder := NewDockerTemplateBuilder(time.Now)
+
+	controlPlaneYaml, err := templateBuilder.GenerateCAPISpecControlPlane(
+		spec,
+		func(values map[string]interface{}) {
+			values["controlPlaneTemplateName"] = clusterapi.ControlPlaneMachineTemplateName(spec.Cluster)
+			values["etcdTemplateName"] = clusterapi.EtcdMachineTemplateName(spec.Cluster)
+		},
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "generating docker control plane yaml spec")
+	}
+
+	parser, builder, err := newControlPlaneParser(logger)
+	if err != nil {
+		return nil, err
+	}
+
+	err = parser.Parse(controlPlaneYaml, builder)
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing docker control plane yaml")
+	}
+
+	cp := builder.ControlPlane
+	if err = cp.UpdateImmutableObjectNames(ctx, client, GetMachineTemplate, MachineTemplateEqual); err != nil {
+		return nil, errors.Wrap(err, "updating docker immutable object names")
+	}
+
+	return cp, nil
+}
+
+func newControlPlaneParser(logger logr.Logger) (*yamlutil.Parser, *controlPlaneBuilder, error) {
+	parser, builder, err := yamlcapi.NewControlPlaneParserAndBuilder(
+		logger,
+		yamlutil.NewMapping(
+			"DockerCluster",
+			func() *dockerv1.DockerCluster {
+				return &dockerv1.DockerCluster{}
+			},
+		),
+		yamlutil.NewMapping(
+			"DockerMachineTemplate",
+			func() *dockerv1.DockerMachineTemplate {
+				return &dockerv1.DockerMachineTemplate{}
+			},
+		),
+	)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "building docker control plane parser")
+	}
+
+	return parser, builder, nil
+}
--- a/pkg/providers/docker/controlplane_test.go
+++ b/pkg/providers/docker/controlplane_test.go
@@ -1,0 +1,692 @@
+package docker_test
+
+import (
+	"context"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/mrajashree/etcdadm-bootstrap-provider/api/v1beta1"
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	dockerv1 "sigs.k8s.io/cluster-api/test/infrastructure/docker/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/internal/test"
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/providers/docker"
+	"github.com/aws/eks-anywhere/pkg/utils/ptr"
+	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+func TestControlPlaneObjects(t *testing.T) {
+	tests := []struct {
+		name         string
+		controlPlane *docker.ControlPlane
+		want         []kubernetes.Object
+	}{
+		{
+			name: "stacked etcd",
+			controlPlane: &docker.ControlPlane{
+				Cluster:                     capiCluster(),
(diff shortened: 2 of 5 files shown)


pkg/clusterapi/controlplane.go
```diff
@@ -56,6 +56,7 @@ func (cp *ControlPlane[C, M]) UpdateImmutableObjectNames(
 ) error {
 	currentKCP := &controlplanev1.KubeadmControlPlane{}
 	err := client.Get(ctx, cp.KubeadmControlPlane.Name, cp.KubeadmControlPlane.Namespace, currentKCP)
+
```
mitalipaygude, line 59: nit: new line?

pkg/providers/docker/controlplane_test.go
```diff
@@ -0,0 +1,527 @@
...
+					Bootstrapped: false,
+				},
+			},
+		},
+	}
+}
+
+func kubeadmControlPlane() *controlplanev1.KubeadmControlPlane {
+	var kcp *controlplanev1.KubeadmControlPlane
+	b := []byte(`apiVersion: controlplane.cluster.x-k8s.io/v1beta1
```
mitalipaygude, line 265: can we add this as an expected results cp yaml in the testdata folder? in my opinion its easier to maintain the yaml and use for other tests.

pkg/providers/docker/testdata/cluster_main.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: test
+  namespace: test-namespace
+spec:
+  controlPlaneConfiguration:
+    count: 3
+  kubernetesVersion: "1.19"
```
mitalipaygude, line 9: 1.19 has been deprecated at the moment, can we use 1.22/1.23 instead?

pkg/providers/docker/controlplane_test.go
```diff
@@ -0,0 +1,527 @@
...
+	cp, err := docker.ControlPlaneSpec(ctx, logger, client, spec)
+
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(cp).NotTo(BeNil())
+	g.Expect(cp.Cluster).To(Equal(capiCluster()))
+	g.Expect(cp.KubeadmControlPlane).To(Equal(kubeadmControlPlane()))
+	g.Expect(cp.EtcdCluster).To(Equal(etcdCluster()))
+	g.Expect(cp.ProviderCluster).To(Equal(dockerCluster()))
+	g.Expect(cp.ControlPlaneMachineTemplate.Name).To(Equal("test-control-plane-1"))
+	g.Expect(cp.EtcdMachineTemplate.Name).To(Equal("test-etcd-1"))
```
g-gaston, line 99: Can we assert the content of the machine templates?

pkg/providers/docker/controlplane.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// BaseControlPlane represents a CAPI Docker control plane.
+type BaseControlPlane = clusterapi.ControlPlane[*dockerv1.DockerCluster, *dockerv1.DockerMachineTemplate]
+
+// ControlPlane holds the Docker specific objects for a CAPI Docker control plane.
+type ControlPlane struct {
```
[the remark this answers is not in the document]
cxbrowne1207: Alright. I've made the necessary changes to remove the extra struct.

pkg/providers/docker/controlplane.go
```diff
@@ -0,0 +1,106 @@
...
+	BaseControlPlane
+}
+
+// Objects returns the control plane objects associated with the Docker cluster.
+func (p ControlPlane) Objects() []kubernetes.Object {
```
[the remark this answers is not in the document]
cxbrowne1207: It has been removed.

pkg/providers/docker/controlplane.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// BaseControlPlane represents a CAPI Docker control plane.
+type BaseControlPlane = clusterapi.ControlPlane[*dockerv1.DockerCluster, *dockerv1.DockerMachineTemplate]
```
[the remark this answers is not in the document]
cxbrowne1207: Thanks. This has been updated.

pkg/clusterapi/controlplane.go
cxbrowne1207: Ah, this was a line I probably touched while debugging but never discarded the changes properly. Will change it back.

pkg/providers/docker/machinetemplate.go
```diff
@@ -0,0 +1,24 @@
...
+	m := &dockerv1.DockerMachineTemplate{}
+	if err := client.Get(ctx, name, namespace, m); err != nil {
+		return nil, errors.Wrap(err, "reading dockerMachineTemplate")
+	}
+
+	return m, nil
+}
+
+func machineTemplateEqual(new, old *dockerv1.DockerMachineTemplate) bool {
+	return equality.Semantic.DeepDerivative(new, old)
```
mitalipaygude, line 23:
```suggestion
	return equality.Semantic.DeepDerivative(new.Spec, old.Spec)
```

pkg/providers/docker/controlplane_test.go
```diff
@@ -0,0 +1,527 @@
...
+		})
+	}
+}
+
+func TestControlPlaneSpecNewCluster(t *testing.T) {
+	g := NewWithT(t)
+	logger := test.NewNullLogger()
+	ctx := context.Background()
+	client := test.NewFakeKubeClient()
+	spec := test.NewFullClusterSpec(t, testClusterConfigMainFilename)
```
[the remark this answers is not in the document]
cxbrowne1207: Sure thing. It's been adjusted!

pkg/clusterapi/controlplane.go
cxbrowne1207: Fixed

pkg/providers/docker/testdata/cluster_main.yaml
cxbrowne1207: Fixed. Using 1.23 insead

pkg/providers/docker/controlplane_test.go
cxbrowne1207: Yes., I've made the changes

pkg/providers/docker/controlplane.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// ControlPlane represents a CAPI Docker control plane.
+type ControlPlane = clusterapi.ControlPlane[*dockerv1.DockerCluster, *dockerv1.DockerMachineTemplate]
+
+// ControlPlaneBuilder defines the builder for all objects in the CAPI Docker control plane.
+type ControlPlaneBuilder struct {
```
taneyland, line 22: Do we need this struct? I think we can also remove this since the docker provider has no extra cp objects. The builder that it creates should suffice

pkg/providers/docker/controlplane_test.go
taneyland: If we take a route like this, I would suggest maybe trying to leverage one of the existing expected results test files?

But I would vote to define the objects directly in the file as it is now. Imo, it's better for debugging & verifying the contents of each individual cp object in the test

pkg/providers/docker/controlplane.go
cxbrowne1207: Agreed. Tis cleaner

pkg/providers/docker/controlplane_test.go
cxbrowne1207: Is there's an existing way to parse a yaml into an object? It may be possible to cover both options.

I do think the yaml file is easier to do deal with in general, but it may be more beneficial for debugging and verification as is now like @taneyland mentioned.

cxbrowne1207: Also, since I used Kube123 for the tests, I defined and used a more up to date VersionsBundle object than the existing one in `docker_test.go` , so the result would never match one the existing files.

Not sure if the VersionsBundle matters as much when using a different Kube version for the tests though. Let me know what you guys think.

taneyland: Yeah good point, the versions bundle should match so using the existing files wouldn't work since we updated the version. I think we should keep it as is.

mitalipaygude: Personally i think this makes the test files look more cluttered, and would be inconsistent compared to all the other tests we have written. I am fine with it however if the team decides to use objects from here on out.

taneyland (approved):

pkg/providers/docker/controlplane_test.go
```diff
@@ -0,0 +1,566 @@
...
+	ctx := context.Background()
+	client := test.NewFakeKubeClient()
+	spec := testClusterSpec()
+	spec.Cluster.Spec.KubernetesVersion = ""
+
+	_, err := docker.ControlPlaneSpec(ctx, logger, client, spec)
+	g.Expect(err).To(MatchError(ContainSubstring("generating docker control plane yaml spec")))
+}
+
+func TestControlPlaneSpecUpdateMachineTemplates(t *testing.T) {
```
g-gaston, line 110: Could we have a test for the scenario of an upgrade without changes in the machine templates?
We recently added one of those to sphere if you want to check it out :)

I think it's important, we missed it for vsphere and that's why we didn't the bug earlier.

pkg/providers/docker/controlplane_test.go
g-gaston: I probably missed part of the convo, but any reason not to use go structs directly?

pkg/providers/docker/machinetemplate_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestMachineTemplateEqualDifferentCustomImages(t *testing.T) {
+	g := NewWithT(t)
+
+	machineTemplate := dockerMachineTemplate("test-machine-1")
+	otherMachineTemplate := machineTemplate.DeepCopy()
+	otherMachineTemplate.Spec.Template.Spec.CustomImage = "other-custom-image"
+
+	isEqual := docker.MachineTemplateEqual(machineTemplate, otherMachineTemplate)
+	g.Expect(isEqual).To(BeFalse())
```
g-gaston, line 35: nit:
```suggestion
	g.Expect(docker.MachineTemplateEqual(machineTemplate, otherMachineTemplate)).To(BeFalse())
```

pkg/providers/docker/controlplane_test.go
cxbrowne1207: Added this. `DockerMachineTemplate` does not have a Snapshot field, but I used the `ProviderID` field instead to test defaults.

pkg/providers/docker/machinetemplate_test.go
cxbrowne1207: Fixed

g-gaston (approved):
This is awesome!

PR #4018, opened 2022-11-10
Split instances in chunks for termination

g-gaston:
*Description of changes:*
The AWS API has a limit of a 1000 instances per request. If the cleanup job fails to run a couple times, the accumulated instances number can exceed this limit. The API docs recommend to split this into even smaller chunks, so going with 500 for now.


jaxesn (approved):
Thanks!

PR #4019, opened 2022-11-10
moved validation to docker validator

tatlat:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/pull/3409#discussion_r976956447

*Description of changes:*
Moved provider specific docker endpoint validation from cluster level to provider level.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/docker/validator.go
+++ b/pkg/providers/docker/validator.go
@@ -6,6 +6,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/cluster"
 )
 
+// ValidateControlPlaneEnpoint - checks to see if endpoint host configuration is specified for docker cluster and returns an error if true.
 func ValidateControlPlaneEndpoint(clusterSpec *cluster.Spec) error {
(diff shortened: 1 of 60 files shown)


pkg/providers/docker/validator.go
```diff
@@ -0,0 +1,14 @@
+package docker
+
+import (
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+)
+
+func validateControlPlaneEndpoint(clusterSpec *cluster.Spec) error {
```
taneyland, line 9: For other reviewers: the idea is to call this from the controller when validating the docker clusters during create

taneyland (approved):

pkg/providers/docker/validator_test.go
```diff
@@ -0,0 +1,30 @@
+package docker
```
taneyland, line 1: Can we put this file in the `docker_test` package? We can make the new method public so that we can call it from here

mitalipaygude (approved):

PR #4020, opened 2022-11-10
minor rework for eks-a-test-tool

ewollesen:
- Don't make output directories until a file is written

  This prevents useless and empty directories from being created when there's a usage error or something.

- Change colons to underscores in directory names

  Just a small quality of life improvement. Most terminals, by default, will recognize a colon as a word-stop, and so double-clicking will select to the colon but no further. Changing to underscore lets a nice double-click select the whole directory.

https://user-images.githubusercontent.com/169516/201174235-0a39a057-6c8a-43c6-ba3a-89d74fc174f8.mp4


TerryHowe (approved):
/lgtm
/approve

TerryHowe (approved):
/lgtm

PR #4027, opened 2022-11-10
Fix vsphere machine template equal compare

g-gaston:
## Description of changes
When looking for changes in immutable objects, we only care about the spec field. Any other change is allowed and doesn't require a new name. If we don't do it, we create a new machine template on every single reconciliation loop, which trigger an infinite rollout of new machines.

## Testing
Unit tests and manual testing with tilt by @jonathanmeier5 and @mitalipaygude 


jonathanmeier5 (approved):
/lgtm
/approve

mitalipaygude (approved):

PR #4041, opened 2022-11-11
Improve E2E vSphere template selection logic

g-gaston:
## Description of changes

Due to the the speed at which we bump eks-d versions, template config gets out of sync frequently. This makes the E2E tests fail. The solution is easy, updating the config with the new templates and importing if they are haven't been yet. However, due to the lenght of a full suite run, it takes up to a full day to get feedback on the fix.

This PRs introduces extra logic to try to select the right template even if the configuration is missing while still accepting explicit config to override the defaults. It follows these steps:

1. Look for explicit configuration through an env var: "T_VSPHERE_TEMPLATE_{osFamily}_{eks-d version}".
This should be used for explicit configuration, mostly in local development for overrides.
2. If not present, look for a template if the default templates folder: "/SDDC-Datacenter/vm/Templates/{eks-d version}-{osFamily}" This is what should be used most of the time in CI, the explicit configuration is not present but the right template has already been imported to vSphere.
3. If the template doesn't exist, default to the value of the default template env vars: eg. "T_VSPHERE_TEMPLATE_UBUNTU_1_20". This is a catch all condition. Mostly for edge cases where the bundle has been updated with a new eks-d version, but the the new template hasn't been imported yet. It also preserves backwards compatibility.

### DIFF
--- a/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
+++ b/cmd/integration_test/build/buildspecs/test-eks-a-cli.yml
@@ -15,6 +15,7 @@ env:
     TEST_RUNNER_GOVC_LIBRARY: "eks-a-templates"
     TEST_RUNNER_GOVC_TEMPLATE: "eks-a-admin-ci"
     INTEGRATION_TEST_INFRA_CONFIG: "/tmp/test-infra.yml"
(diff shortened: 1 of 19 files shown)


jaxesn:
Merging to test in e2e tests.

PR #4042, opened 2022-11-12
Modifying instance tagging and upload artifacts

pokearu:
*Description of changes:*
Uploading artifacts for tinkerbell tests was failing because we were trying to upload artifacts for passed tests.
Also the VM tagging was tagging all instances as successful runs.


abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #4045, opened 2022-11-14
Move tagging of images to registry.go

a-cool-train:
*Description of changes:*
- Currently Replacement of default registry and curated packages prod/dev registries occur in docker.go. Ideally we should move this replacement to the registry.go.

*Testing (if applicable):*

### DIFF
--- a/pkg/docker/registry.go
+++ b/pkg/docker/registry.go
@@ -107,6 +107,9 @@ func getUpdatedEndpoint(originalEndpoint, image string) string {
 
 // Replace original registries with the new endpoint from the registry mirror configuration.
 func getLocalImage(endpoint string, image string) string {
+	// Performing a string replacement instead of url replacement because the url parsing
+	// has an issue distinguishing between / and %2f when parsing the url
+	// Issue can be found here: https://github.com/golang/go/issues/10887.
 	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdDomain, endpoint, packageDevDomain, endpoint)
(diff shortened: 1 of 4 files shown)


pkg/docker/registry.go
```diff
@@ -104,6 +105,13 @@ func getUpdatedEndpoint(originalEndpoint, image string) string {
 	return originalEndpoint
 }
 
+// Replace original registries with the new endpoint from the registry mirror configuration.
+func getLocalImage(endpoint string, image string) string {
+	replacer := strings.NewReplacer(defaultRegistry, endpoint, packageProdDomain, endpoint, packageDevDomain, endpoint)
```
g-gaston, line 110: curious, why not just always replace the host, irregardless of which one it is? That way we make this agnostic of the different ecr domains, it just knows that it need to replace the domain with `endpoint`

a-cool-train: Not sure if there was an implementation detail here. I just added extra endpoints to an already existing functionality. I can try urls.replacehost and see if that would work the same mechanism.

a-cool-train: Looks like there is an issue with url parsing that causes an issue. I put a comment on the code to indicate the issue.

jonahjon:
Looks like a failing test, they might need to be tweaked.

TerryHowe:
The copy packages command should make this unnecessary. There is some clean-up here that should be done, but that looks a lot different.

PR #4010, opened 2022-11-09
Add create worklow task

chrisdoherty4:
Add "Create workflow" task to new workflows. The new task should result in a functional, permanent, workload cluster created out of the CAPI process.

PR #4049, opened 2022-11-14
Refactor workload cluster creation

chrisdoherty4:
Depends on #4010 

This PR introduces a shim that facilitates cluster creation using new workflows. 

In #4010 we introduced the Workload Cluster creation task that results in a functional workload cluster. It models a single interface that represents our "to be created" cluster. The "shim" in this PR satisfies the `Cluster` interface required by the task. The task itself should be considered the top level orchestrator. 

To implement the shim it was necessary to refactor the cluster manager to exercise similar code paths to old workflows. The refactor took into consideration a bug found by the CloudStack team that results in nodes being remediated incorrectly. 

Bug TL;DR our machine health checks require nodes to be in a ready state. For a node to enter a ready state the CNI must be installed; given we blocked until all control plane nodes were ready it meant worker nodes could be remediated accidentally. Reach out to @mrog for more details.

This PR circumvents the problem by waiting for the control plane to be initialized which can be defined as "the api server is available on a single control plane node". Doing so lets us move to install the CNI minimizing the chance of a health check identifying nodes as unhealthy.


mrog:
I don't know EKS-A well enough to sign off on most of these changes, but the new WaitForFirstControlPlaneReady function looks good.  Are you planning on installing CNI and adding the MHCs right after this wait is complete?

pkg/executables/kubectl.go
```diff
@@ -295,6 +295,11 @@ func (k *Kubectl) WaitForControlPlaneReady(ctx context.Context, cluster *types.C
 	return k.Wait(ctx, cluster.KubeconfigFile, timeout, "ControlPlaneReady", fmt.Sprintf("%s/%s", capiClustersResourceType, newClusterName), constants.EksaSystemNamespace)
 }
 
+// WaitForFirstControlPlaneReady blocks until the first control plane node is ready.
+func (k *Kubectl) WaitForFirstControlPlaneReady(ctx context.Context, cluster *types.Cluster, timeout string, newClusterName string) error {
+	return k.Wait(ctx, cluster.KubeconfigFile, timeout, "ControlPlaneInitialized", fmt.Sprintf("%s/%s", capiClustersResourceType, newClusterName), constants.EksaSystemNamespace)
```
g-gaston, line 300: What's the rationale between using `ControlPlaneInitialized` instead of `ControlPlaneReady`?

`ControlPlaneReady` comes after `ControlPlaneInitialized` and ensures the CP is "contactable", which I think suits better our usecase. `ControlPlaneInitialized` only ensures that the kubeconfig exists.

If we have a good reason to change it, I would suggest documenting it at least with a comment, since it's a change in behavior in one of core pieces of logic.

chrisdoherty4: That's not quite accurate.

> // ControlPlaneInitializedCondition reports if the cluster's control plane has been initialized such that the
> // cluster's apiserver is reachable and at least one control plane Machine has a node reference.

The change is necessary because we need to install the CNI as quickly as possible to facilitate our `MachineHealthChecks` that expect the `Ready` condition (nodes won't be ready until they have a CNI installed).

pkg/clustermanager/cluster_manager.go
```diff
@@ -247,72 +251,134 @@ func (c *ClusterManager) writeCAPISpecFile(clusterName string, content []byte) e
...
-	}
-
-	logger.V(3).Info("Waiting for workload kubeconfig generation", "cluster", workloadCluster.Name)
-	err = c.Retrier.Retry(
-		func() error {
-			workloadCluster.KubeconfigFile, err = c.generateWorkloadKubeconfig(ctx, workloadCluster.Name, managementCluster, provider)
-			return err
-		},
+	logger.V(3).Info("Waiting for first control plane to be ready")
+	err := c.clusterClient.WaitForFirstControlPlaneReady(
```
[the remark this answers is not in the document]
chrisdoherty4: I'm indifferent. When the first control plane is ready, its available; equally when its available, the first control plane must be ready. 

Though I'm not sure I agree its leaking detail. The devil is in the detail and the detail should probably be in a comment in this case.

chrisdoherty4:
> Are you planning on installing CNI and adding the MHCs right after this wait is complete?

Yes. See the implementation at https://github.com/aws/eks-anywhere/pull/4010/files#diff-75cc19e2353a432ad163dd2ab002569f81dfb5032dbf400c38eafe681a24f758R52

jacobweinstock (approved):
All seems good to me, thanks.

PR #4059, opened 2022-11-15
Added workload cluster upgrade functionality for tinkerbell

panktishah26:
*Description of changes:*
Added workload cluster upgrade functionality for Tinkerbell. Added e2e tests for workload cluster create and upgrade functionality.

*Testing (if applicable):*

--- PASS: TestTinkerbellKubernetes122BottlerocketWorkloadCluster
--- PASS: TestTinkerbellKubernetes122UbuntuWorkloadCluster
--- PASS: TestTinkerbellUpgradeMulticlusterWorkloadClusterWorkerScaleup
--- PASS: TestTinkerbellUpgradeMulticlusterWorkloadClusterK8sUpgrade

### DIFF
--- a/pkg/providers/tinkerbell/upgrade.go
+++ b/pkg/providers/tinkerbell/upgrade.go
@@ -134,7 +134,23 @@ func (p *Provider) SetupAndValidateUpgradeCluster(ctx context.Context, cluster *
 		return err
 	}
 
-	return p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec)
+	if err := p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec); err != nil {
+		return err
+	}
+
+	if p.clusterConfig.IsManaged() {
+		if err := p.applyHardwareUpgrade(ctx, clusterSpec.ManagementCluster); err != nil {
+			return err
+		}
+		if len(p.catalogue.AllBMCs()) > 0 {
+			err = p.providerKubectlClient.WaitForBaseboardManagements(ctx, cluster, "5m", "Contactable", constants.EksaSystemNamespace)
+			if err != nil {
+				return fmt.Errorf("waiting for baseboard management to be contactable: %v", err)
+			}
+		}
+	}
+
+	return nil
 }
 
 func (p *Provider) validateAvailableHardwareForUpgrade(ctx context.Context, currentSpec, newClusterSpec *cluster.Spec) (err error) {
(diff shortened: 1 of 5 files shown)


pkg/providers/tinkerbell/upgrade.go
```diff
@@ -134,7 +134,23 @@ func (p *Provider) SetupAndValidateUpgradeCluster(ctx context.Context, cluster *
 		return err
 	}
 
-	return p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec)
+	if err := p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec); err != nil {
```
pokearu, line 137: Should this be done after `if p.clusterConfig.IsManaged()` block? Once the new hardware is applied on to existing management cluster, we would want to validate if we have enough hardware for upgrade right?

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -134,7 +134,23 @@ func (p *Provider) SetupAndValidateUpgradeCluster(ctx context.Context, cluster *
...
+		return err
+	}
+
+	if p.clusterConfig.IsManaged() {
+		if err := p.applyHardwareUpgrade(ctx, clusterSpec.ManagementCluster); err != nil {
+			return err
+		}
+		if len(p.catalogue.AllBMCs()) > 0 {
```
pokearu, line 145: This check might not work here. During upgrade we do not add the existing BMCs back into the catalogue. This means `if len(p.catalogue.AllBMCs()) > 0` would be `true` only if a new hardware CSV with BMCs are fed during workload cluster upgrades.

We could do a check like https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/upgrade.go#L190

pkg/providers/tinkerbell/upgrade.go
panktishah26: We validate all the hardwares (hardware object) before applying it to the cluster https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/upgrade.go#L75. I also felt the same but during normal create/upgrade too, we apply all the hardwares during PostBootstrapSetupCreate or PostBootstrapSetupUpgrade step after validating it.

pokearu: Oh i see. So the validate only runs against the catalogue object and not the cluster itself. So doing it before would be fine 👍

pokearu (approved):

PR #4077, opened 2022-11-15
Add ControlPlane reconciler for docker

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Implementing the control plane reconciler for docker

*Testing (if applicable):*

### DIFF
--- a/internal/test/testdata.go
+++ b/internal/test/testdata.go
@@ -1,10 +1,16 @@
 package test
 
 import (
+	"time"
+
 	eksdv1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
@@ -123,4 +129,34 @@ func Bundle() *releasev1.Bundles {
 			},
 		},
 	}
+}
+
+type capiClusterOpt func(*clusterv1.Cluster)
+
+// CAPICluster returns a capi cluster which can be configured by passing in opts arguments.
+func CAPICluster(opts ...capiClusterOpt) *clusterv1.Cluster {
+	c := &clusterv1.Cluster{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       anywherev1.ClusterKind,
+			APIVersion: clusterv1.GroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: constants.EksaSystemNamespace,
+		},
+		Status: clusterv1.ClusterStatus{
+			Conditions: clusterv1.Conditions{
+				{
+					Type:               clusterapi.ControlPlaneReadyCondition,
+					Status:             corev1.ConditionTrue,
+					LastTransitionTime: metav1.NewTime(time.Now()),
+				},
+			},
+		},
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	return c
 }
--- a/pkg/providers/docker/reconciler/reconciler_test.go
+++ b/pkg/providers/docker/reconciler/reconciler_test.go
@@ -3,11 +3,16 @@ package reconciler_test
 import (
 	"context"
 	"errors"
+	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	etcdv1 "github.com/mrajashree/etcdadm-controller/api/v1beta1"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	dockerv1 "sigs.k8s.io/cluster-api/test/infrastructure/docker/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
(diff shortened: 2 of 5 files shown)


pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -246,3 +303,32 @@ func dataCenter() *anywherev1.DockerDatacenterConfig {
 		},
 	}
 }
+
+type capiClusterOpt func(*clusterv1.Cluster)
+
+func capiCluster(opts ...capiClusterOpt) *clusterv1.Cluster {
```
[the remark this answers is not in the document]
cxbrowne1207: Great idea. I've made the changes and refactored some.

vivek-koppuru (approved):

internal/test/testdata.go
```diff
@@ -124,3 +130,33 @@ func Bundle() *releasev1.Bundles {
 		},
 	}
 }
+
+type capiClusterOpt func(*clusterv1.Cluster)
+
+// CapiCluster returns a capi cluster which can be configured by passing in opts arguments.
+func CapiCluster(opts ...capiClusterOpt) *clusterv1.Cluster {
```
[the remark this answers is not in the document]
cxbrowne1207: Fixed

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -89,6 +92,52 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
+
+func TestReconcilerReconcileControlPlaneFailure(t *testing.T) {
+	tt := newReconcilerTest(t)
+	tt.createAllObjs()
+	spec := tt.buildSpec()
+	spec.Cluster.Spec.KubernetesVersion = ""
+
+	_, err := tt.reconciler().ReconcileControlPlane(tt.ctx, test.NewNullLogger(), spec)
+
+	tt.Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
cxbrowne1207: Fixed.

cxbrowne1207:
/hold

working on adding some extra tests

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -203,6 +278,49 @@ func (tt *reconcilerTest) allObjs() []client.Object {
 	return objs
 }
 
+func (tt *reconcilerTest) runReconcileControlPlaneTestFlow() {
```
g-gaston, line 281: I understand the idea here, however I would vote against this. It does avoid repeated code which is good, however, drying out tests too much can make them more difficult to read. This almost creates its own framework for tests.

It you still think this is a good idea, I would suggest splitting this into smaller chunks so the reader has more information about what's going on when reading the tests body.

For example:

```go
func TestMyTest(t *testing.T) {
    tt := newReconcilerTest(t)
    tt.createAllObjs()
    
    logger := test.NewNullLogger()
	result, err := tt.reconciler().ReconcileControlPlane(tt.ctx, logger, tt.buildSpec())

	tt.Expect(err).NotTo(HaveOccurred())

    tt.checkCAPIClusterExists()
	tt.checkKubeadmControlPlaneExists()	
	tt.checkDockerMachineTemplateExists()
	tt.checkDockerClusterExists()
}
```

IMO this is almost as short as your tests, however, by adding a few repeated lines, this gives a way better idea of what the test is doing:
* I can see what steps the setup is doing, although abstracted in a couple methods
* I can clearly see the method that is being tested: `ReconcileControlPlane`
* I can clearly see what the expected returned values are
* I can see the additional expectations over the objects states and relate them to what the tested method is supposed to do

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -103,7 +167,7 @@ type reconcilerTest struct {
 	datacenterConfig     *anywherev1.DockerDatacenterConfig
 }
 
-func newReconcilerTest(t testing.TB) *reconcilerTest {
+func newReconcilerTest(t testing.TB, opts ...clusterOpt) *reconcilerTest {
```
g-gaston, line 170: I would vote against this
I get that it makes the setup code flow a bit better, but it only works for cluster modification, and it opens the door to want to do the same with other objects, which is not possible because go only accepts one set of variadic arguments. In that case, we would need two different patterns to update two different objects and that, imo, makes the tests more confusing.

g-gaston (commented):
looking good!

pkg/providers/docker/reconciler/reconciler_test.go
cxbrowne1207: Okay, I see your point. I've decided to leave the assertions in the test itself.

pkg/providers/docker/reconciler/reconciler_test.go
cxbrowne1207: Noted, that makes sense. I've updated it taken the feedback into consideration.

internal/test/testdata.go
```diff
@@ -124,3 +130,33 @@ func Bundle() *releasev1.Bundles {
 		},
 	}
 }
+
+type capiClusterOpt func(*clusterv1.Cluster)
```
g-gaston, line 134: since the method using this as arg is exported, we should make it exported

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -89,6 +94,132 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      tt.cluster.Name,
+				Namespace: constants.EksaSystemNamespace,
+			},
+		},
+	)
+	tt.ShouldEventuallyNotExist(tt.ctx,
+		&dockerv1.DockerMachineTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "workload-cluster-etcd-1",
```
g-gaston, line 139:
```suggestion
				Name:       tt.cluster.Name + "-etcd-1",
```

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -89,6 +94,132 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
+				Namespace: constants.EksaSystemNamespace,
+			},
+		},
+	)
+	tt.ShouldEventuallyNotExist(tt.ctx,
+		&etcdv1.EtcdadmCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "workload-cluster-etcd",
```
g-gaston, line 147:
```suggestion
				Name:      tt.cluster.Name + "-etcd",
```

g-gaston (approved):

PR #4082, opened 2022-11-15
Stop reconciliation if child objects are invalid

g-gaston:
## Description of changes

If datacenter or machine configs are invalid for a cluster, reconciliation should stop to prevent any further changes.

When the failure message is not set, it's safe for us to assume the object hasn't been validated yet but still might be valid. In this case, we shouldn't bubble an error to the cluster status and just return and wait for the object to updated, either to be valid or to have a failure message.

### DIFF
--- a/pkg/providers/snow/reconciler/reconciler.go
+++ b/pkg/providers/snow/reconciler/reconciler.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/reconciler/reconciler_test.go
```diff
@@ -136,9 +137,9 @@ func TestReconcilerFailToSetUpMachineConfigCP(t *testing.T) {
 
 	result, err := tt.reconciler().ValidateMachineConfigs(tt.ctx, logger, tt.buildSpec())
 	g := NewWithT(t)
-	g.Expect(err).To(MatchError(ContainSubstring("validating vCenter setup for VSphereMachineConfig")))
-	g.Expect(*tt.cluster.Status.FailureMessage).To(ContainSubstring("validating vCenter setup for VSphereMachineConfig"))
-	g.Expect(result).To(Equal(controller.Result{}))
+	tt.Expect(err).To(BeNil(), "error should be nil to prevent requeue")
+	tt.Expect(result).To(Equal(controller.Result{Result: &reconcile.Result{}}), "result should stop reconciliation")
+	g.Expect(tt.cluster.Status.FailureMessage).To(HaveValue(ContainSubstring("validating vCenter setup for VSphereMachineConfig")))
```
cxbrowne1207, line 142: Is using `g` here instead of `tt` intentional?

cxbrowne1207 (commented):
Left a comment to confirm something that looks weird to me with one of the assertions.

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -76,14 +78,15 @@ func (r *Reconciler) ValidateMachineConfigs(ctx context.Context, log logr.Logger
 	log = log.WithValues("phase", "validateMachineConfigs")
 	for _, machineConfig := range clusterSpec.SnowMachineConfigs {
 		if !machineConfig.Status.SpecValid {
-			failureMessage := fmt.Sprintf("SnowMachineConfig %s is invalid", machineConfig.Name)
 			if machineConfig.Status.FailureMessage != nil {
```
taneyland, line 81: combine with the prev `if` statement?

pkg/providers/vsphere/reconciler/reconciler.go
```diff
@@ -128,14 +128,15 @@ func (r *Reconciler) ValidateDatacenterConfig(ctx context.Context, log logr.Logg
 	dataCenterConfig := clusterSpec.VSphereDatacenter
 
 	if !dataCenterConfig.Status.SpecValid {
```
taneyland, line 130: same comment to combine the  `if` statements here

pkg/providers/vsphere/reconciler/reconciler_test.go
g-gaston: Fixed

pkg/providers/snow/reconciler/reconciler.go
g-gaston: I think that wouldn't have the effect you expect it to have. If I combine those, I can't handle the case where `SpecValid` is false but the failure message is nil. In that case we also want to stop the reconciliation.

taneyland: Ah yes I see. Ok disregard then!

Issue #3491, opened 2022-09-29
Support for Kubernetes 1.24


mitalipaygude:
- [x] Create 1.24 resources in build CDK
- [x] Update EKSA build tooling with dev support for 1.24
- [x] Run vsphere ubuntu 1.24 cluster and conformance tests
- [x] Update ci secret with template
- [x] Add feature flag for 1.24
- [x] Add e2e tests for Ubuntu
- [x] Import BR OVAs and create BR cluster
- [x] Add BR 1.24 tests and appropriate resources
- [x] Update EKSD to prod release
- [x] Support for CloudStack
- [x] Support for baremetal
- [x] Remove feature flag
- [x] Add build statuses
- [x] Update docs

PR #4084, opened 2022-11-15
Remove feature flag for 1.24 

mitalipaygude:
*Issue #, if available:*
#3491 
*Description of changes:*
Remove feature flag for 1.24 
*Testing (if applicable):*


docs/content/en/docs/reference/changelog.md
```diff
@@ -4,6 +4,15 @@ linkTitle: "What's New?"
 weight: 35
 ---
 
+## Unreleased
```
vivek-koppuru, line 7: We don't want this to be showed up as "Unreleased" in this branch though because it is already public facing right?

vivek-koppuru (approved):

PR #4100, opened 2022-11-16
Add docker workers CAPI spec generator

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Add docker workers CAPI spec generator
*Testing (if applicable):*


g-gaston:
/override "codecov/patch"

The lines not covered are actually very difficult to test and doing so wouldn't provide much value. They will never fail unless there is a programming error (the code changes to something invalid), since they just deal with object init and don't depend on anything external. If the code were to change to something invalid, the other tests that assert for no error will fail, preventing us from shipping a bug.

Given that and since coverage is 75%, overriding.

cxbrowne1207 (approved):
/lgtm

PR #4105, opened 2022-11-16
Add stale issue manager to GH actions

markapruett:
*Issue #, if available:*

*Description of changes:*
Add GH workflow to mark and auto close stale issues. Can easily be extended to apply to PRs as well

*Testing (if applicable):*
Tested on my fork and works as expected 

### DIFF
--- a/.github/workflows/stale-issue-mgr.yaml
+++ b/.github/workflows/stale-issue-mgr.yaml
@@ -3,7 +3,7 @@
 # This action will: 
 # * Add a label "stale" on issues after 60 days of inactivity and comment on them if they also have the "external" label
 # * Close the stale issues and pull requests after 7 days of inactivity
-# * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restartif they have the "external" label
+# * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart if they have the "external" label
 
 # Runs every Monday morning at 1AM.
 # Docs: https://github.com/actions/stale
@@ -27,5 +27,5 @@ jobs:
           stale-issue-label: 'stale'
           any-of-issue-labels: 'external'
 env:
-  ISSUE_STALE_DAYS: 60
+  ISSUE_STALE_DAYS: 0
   ISSUE_CLOSE_AFTER_DAYS: 7
--- a/.github/workflows/stale-issue-mgr.yaml
+++ b/.github/workflows/stale-issue-mgr.yaml
@@ -5,7 +5,7 @@
 # * Close the stale issues and pull requests after 7 days of inactivity
 # * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart if they have the "external" label
 
-# Runs every Monday morning at 1AM.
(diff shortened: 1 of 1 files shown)


vivek-koppuru (approved):

.github/workflows/stale-issue-mgr.yaml
```diff
@@ -0,0 +1,31 @@
+# Stale issue handler (could also be extended for stale PRs in future)
+
+# This action will: 
+# * Add a label "stale" on issues after 60 days of inactivity and comment on them if they also have the "external" label
+# * Close the stale issues and pull requests after 7 days of inactivity
+# * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restartif they have the "external" label
```
drewvanstone, line 6:
```suggestion
# * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart if they have the "external" label
```

drewvanstone (commented):
I think we may need `remove-issue-stale-when-updated` to remove the stale tag if there is a comment or update to the issue

.github/workflows/stale-issue-mgr.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Runs every Monday morning at 1AM.
+# Docs: https://github.com/actions/stale
+
+name: 'Label and close stale issues'
+on:
+  schedule:
+    - cron: '0 1 * * MON'
+  workflow_dispatch:
```
chrisdoherty4, line 15: What do we need the dispatch event for?

markapruett: I used it for testing and left it there in case we want to trigger manually.

markapruett:
> I think we may need `remove-issue-stale-when-updated` to remove the stale tag if there is a comment or update to the issue:
> 
> https://github.com/actions/stale#remove-issue-stale-when-updated

Thanks @drewvanstone. I don't think we need it. That's the override for issues only vs PRs & issues, presumably to set it to `false`, since [the default](https://github.com/actions/stale#remove-stale-when-updated) is already set to `true`. 

When I tested the current configuation on my fork it removed the `stale` label if there was a comment. See [example here](https://github.com/markapruett/eks-anywhere/issues/6)

markapruett:
I changed the execution to every day, to have more frequent un-stales if people comment, instead of waiting a week and pinging us asking to remove the label.

.github/workflows/stale-issue-mgr.yaml
```diff
@@ -0,0 +1,31 @@
...
+# * Close the stale issues and pull requests after 7 days of inactivity
+# * If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart if they have the "external" label
+
+# Runs every morning at 1AM: We don't want people having to wait a whole week to get the stale label removed 
+# Docs: https://github.com/actions/stale
+
+name: 'Label and close stale issues'
+on:
+  schedule:
+    - cron: '0 1 * * *'
```
markapruett, line 14: I changed the execution to every day, to have more frequent un-stales if people comment, instead of waiting a week and pinging us asking to remove the label.

chrisdoherty4 (approved):

PR #4115, opened 2022-11-16
Improve controller manager logs

g-gaston:
## Description of changes
This now makes use of the tooling provided by upstream controller-runtime and component-base/logs. As a consequence, the manager now takes a standard set of flags to configure logging.

The default logging format is configured to json, it's the direction upstream is going due to its tooling support. For tilt, using text for now until we provide better tooling to analyze the logs.

Instead of injecting a logger in the reconciler on construction, we now use contextual logging. The benefit here is controller-runtime will inject a default set of values that uniquely identify the reconcile request.

More info
* https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/1602-structured-logging/README.md
* https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/3077-contextual-logging/README.md
* https://github.com/kubernetes-sigs/controller-runtime/pull/1827


go.mod
```diff
@@ -42,25 +42,25 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.24.2
-	k8s.io/apimachinery v0.24.3
-	k8s.io/apiserver v0.24.2
-	k8s.io/client-go v0.24.2
+	k8s.io/api v0.25.4
```
jonathanmeier5, line 45: Why all the go mod changes?

manager/main.go
```diff
@@ -62,34 +57,62 @@ func init() {
...
+
+	if err := logsv1.ValidateAndApply(config.logging, nil); err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	// klog.Background will automatically use the right logger.
+	ctrl.SetLogger(klog.Background())
+	// Once controller-runtime logger has been setup correctly, retrieve again
+	setupLog = ctrl.Log.WithName("setup")
```
vivek-koppuru, line 106: What's the purpose of defining this again? Not aware if there is some caching logic here that pulls the same logger with that name, which I see is passed onto all the reconcilers too.

manager/main.go
```diff
@@ -62,34 +57,62 @@ func init() {
...
-	initFlags(pflag.CommandLine)
+	initFlags(pflag.CommandLine, config)
 	pflag.Parse()
-	ctrl.SetLogger(kzap.New(kzap.UseFlagOptions(&opts)))
-	features.FeedGates(gates)
+
+	// Temporary logger for initialization
+	setupLog := ctrl.Log.WithName("setup")
+
+	if err := logsv1.ValidateAndApply(config.logging, nil); err != nil {
```
vivek-koppuru, line 98: Is this how the logging gets assigned to all the reconcilers?

controllers/cluster_controller.go
```diff
@@ -104,10 +103,10 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // +kubebuilder:rbac:groups=distro.eks.amazonaws.com,resources=releases,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awssnowclusters;awssnowmachinetemplates;vsphereclusters;vspheremachinetemplates,verbs=get;list;watch;create;update;patch;delete
 func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
-	log := r.log.WithValues("cluster", req.NamespacedName)
+	log := ctrl.LoggerFrom(ctx)
```
vivek-koppuru, line 106: When does the association happen from logger to the ctx?

vivek-koppuru (commented):
Just some questions that I am not sure of

go.mod
g-gaston: I bumped `controller-runtime` to use their latest logger attributes (`reconcileID` for example).

Also, go upgraded some modules when adding `component-base`.

manager/main.go
g-gaston: I'm not sure I understand the question about the cache
The purpose is more or less what the comment says: at the beginning of the execution, there is a default logger. Once the configuration has been read and applied, a new logger with the proper setup is created and set with `ctrl.SetLogger`. This line just retrieves that new configured logger to use for the rest of the setup execution.

manager/main.go
g-gaston: This reads the logging config (that can be overridden with manager flags) and inits (if necessary) a logger configured accordingly.

The reconciler don't get a logger assigned. `controller-runtime` injects a logger in the context on each reconciliation loop.

controllers/cluster_controller.go
g-gaston: `controller-runtime` does that for us.

manager/main.go
vivek-koppuru: I see the same line on line 96 above so I was wondering how we were modifying `ctrl.Log` and retrieving a different instance of `setupLog`, so I was wondering what might have changed and whether this log is being stored somewhere.

g-gaston: `ctrl.Log` gets modified when we call `ctrl.SetLogger`
Which we set with the klog logger. Which is configured when we call `VerifyAndApply`

vivek-koppuru (approved):

vivek-koppuru (approved):

Issue #4089, opened 2022-11-15
Reconcile CNI for vSphere


jonathanmeier5:
We should be able to update the CNI policy via the cluster lifecycle controller.

Right now, the webhook fails updates on the `ClusterNetwork` key. We should allow `ClusterNetwork.cniConfig` to be updated.

PR #4122, opened 2022-11-16
Reconcile Cilium Enforcement Policy

jonathanmeier5:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/4089

*Description of changes:*
Before this, the reconciler only reacted to changes in the Cilium
version specified in the Bundles. Now it will detect changes in the
Cilium config as well and reapply the manifest without going through the
version upgrade process.

*Testing (if applicable):*
Manual and unit test

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -148,10 +148,28 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 			field.Forbidden(specPath.Child("datacenterRef"), fmt.Sprintf("field is immutable %v", new.Spec.DatacenterRef)))
 	}
 
-	if !new.Spec.ClusterNetwork.Equal(&old.Spec.ClusterNetwork) {
+	if !new.Spec.ClusterNetwork.Pods.Equal(&old.Spec.ClusterNetwork.Pods) {
 		allErrs = append(
 			allErrs,
-			field.Forbidden(specPath.Child("ClusterNetwork"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
+			field.Forbidden(specPath.Child("clusterNetwork", "pods"), "field is immutable"))
+	}
+
+	if !new.Spec.ClusterNetwork.Services.Equal(&old.Spec.ClusterNetwork.Services) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("clusterNetwork", "services"), "field is immutable"))
+	}
+
+	if !new.Spec.ClusterNetwork.DNS.Equal(&old.Spec.ClusterNetwork.DNS) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("clusterNetwork", "dns"), "field is immutable"))
+	}
+
+	if !new.Spec.ClusterNetwork.Nodes.Equal(old.Spec.ClusterNetwork.Nodes) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("clusterNetwork", "nodes"), "field is immutable"))
 	}
 
 	if !new.Spec.ProxyConfiguration.Equal(old.Spec.ProxyConfiguration) {
--- a/pkg/networking/cilium/reconciler/reconciler.go
+++ b/pkg/networking/cilium/reconciler/reconciler.go
@@ -19,7 +19,9 @@ import (
 	"github.com/aws/eks-anywhere/pkg/utils/oci"
 )
 
-const defaultRequeueTime = time.Second * 10
+const (
+	defaultRequeueTime = time.Second * 10
+)
 
 type Templater interface {
 	GenerateUpgradePreflightManifest(ctx context.Context, spec *cluster.Spec) ([]byte, error)
@@ -52,9 +54,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
 
 	logger.Info("Cilium is already installed, checking if it needs upgrade")
 	upgradeInfo := cilium.BuildUpgradePlan(installation, spec)
-	if !upgradeInfo.Needed() {
-		logger.Info("Cilium is already in desired version")
-	} else {
+
+	if upgradeInfo.VersionUpgradeNeeded() {
 		logger.Info("Cilium upgrade needed", "reason", upgradeInfo.Reason())
 
 		if result, err := r.upgrade(ctx, logger, client, installation, spec); err != nil {
@@ -62,6 +63,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
 		} else if result.Return() {
 			return result, nil
 		}
+	} else if upgradeInfo.ConfigUpdateNeeded() {
+		logger.Info("Cilium config update needed", "reason", upgradeInfo.Reason())
+		if err := r.updateConfig(ctx, client, spec); err != nil {
+			return controller.Result{}, err
+		}
+	} else {
+		logger.Info("Cilium is already up to date")
 	}
 
 	return r.deletePreflightIfExists(ctx, client, spec)
@@ -69,12 +77,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
 
 func (r *Reconciler) install(ctx context.Context, log logr.Logger, client client.Client, spec *cluster.Spec) (controller.Result, error) {
 	log.Info("Installing Cilium")
-	ciliumSpec, err := r.templater.GenerateManifest(ctx, spec)
-	if err != nil {
-		return controller.Result{}, err
-	}
-	if err := serverside.ReconcileYaml(ctx, client, ciliumSpec); err != nil {
-		return controller.Result{}, err
+	if err := r.applyFullManifest(ctx, client, spec); err != nil {
+		return controller.Result{}, errors.Wrap(err, "installing Cilium")
 	}
 
 	return controller.Result{}, nil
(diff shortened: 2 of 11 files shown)


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -148,10 +148,16 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
...
 		allErrs = append(
 			allErrs,
-			field.Forbidden(specPath.Child("ClusterNetwork"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
+			field.Forbidden(specPath.Child("ClusterNetwork.Services"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
+	}
+
+	if !new.Spec.ClusterNetwork.Pods.Equal(&old.Spec.ClusterNetwork.Pods) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(specPath.Child("ClusterNetwork.Pods"), fmt.Sprintf("field is immutable %v", new.Spec.ClusterNetwork)))
```
jonathanmeier5, line 160: I need to verify which other fields are immutable.

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -45,23 +50,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
 	if err != nil {
 		return controller.Result{}, err
 	}
+	// desired state
+	// read current state (find config map with cilium config)
+	// think about splitting out upgrade into upgradeVersion and upgradeConfig
+	// extend UpgradePlan to detect configuration changes as well
```
jonathanmeier5, line 56: Remove

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -45,23 +50,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
...
 		} else if result.Return() {
 			return result, nil
 		}
+	} else if upgradeInfo.ConfigUpgradeNeeded() {
+		logger.Info("Cilium config upgrade needed", "reason", upgradeInfo.Reason())
+		if result, err := r.upgradeConfig(ctx, logger, client, installation, spec); err != nil {
+			return controller.Result{}, err
+		} else if result.Return() {
+			return result, nil
+		}
```
jonathanmeier5, line 82: I don't like how crowded this code reads. Will try to find a way to make it easier to read.

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -148,6 +166,21 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
 	return controller.Result{}, nil
 }
 
+func (r *Reconciler) upgradeConfig(ctx context.Context, logger logr.Logger, client client.Client, installation *cilium.Installation, spec *cluster.Spec) (controller.Result, error) {
```
g-gaston, line 169: this doesn't need to return a result
It will also simplify the code in the main reconcile method

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -148,6 +166,21 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
...
+
+	upgradeManifest, err := r.templater.GenerateManifest(ctx, spec)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	if err := serverside.ReconcileYaml(ctx, client, upgradeManifest); err != nil {
+		return controller.Result{}, err
```
g-gaston, line 177: probably worth a wrap to add context to the error

pkg/networking/cilium/upgrade_plan.go
```diff
@@ -116,3 +134,31 @@ func operatorUpgradePlan(operator *appsv1.Deployment, clusterSpec *cluster.Spec)
...
+	} else {
+		newEnforcementPolicy = string(clusterSpec.Cluster.Spec.ClusterNetwork.CNIConfig.Cilium.PolicyEnforcementMode)
+	}
+
+	info := ComponentUpgradePlan{
+		NewEnforcementPolicy: newEnforcementPolicy,
+	}
+
+	if configMap == nil {
+		info.UpgradeReason = "Cilium config doesn't exist"
```
g-gaston, line 152:
```suggestion
		info.UpgradeReason = "Cilium config config doesn't exist"
```

pkg/networking/cilium/upgrade_plan.go
```diff
@@ -116,3 +134,31 @@ func operatorUpgradePlan(operator *appsv1.Deployment, clusterSpec *cluster.Spec)
...
+		return info
+	}
+
+	if val, ok := configMap.Data["enable-policy"]; ok {
```
g-gaston, line 156: if `enable-policy` is not present, we probably need to run the upgrade right?

pkg/networking/cilium/reconciler/reconciler.go
g-gaston, line 169: nit:
```suggestion
func (r *Reconciler) updateConfig(ctx context.Context, logger logr.Logger, client client.Client, installation *cilium.Installation, spec *cluster.Spec) (controller.Result, error) {
```

g-gaston: again, a nit, so no big deal
In general, I think upgrade indicates more a "destructive" change (like changing the version that requires rolling out new node or pod) and update for changes that don't have such drastic side effects.

That said, one could argue that updating the config might also end up rolling new DS (or maybe restarting them) and we have no control over that.

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -19,7 +19,11 @@ import (
 	"github.com/aws/eks-anywhere/pkg/utils/oci"
 )
 
-const defaultRequeueTime = time.Second * 10
+const (
+	ciliumConfigMapName   = "cilium-config"
+	ciliumConfigNamespace = "kube-system"
```
g-gaston, line 24: probably worth moving these two wherever they are being used

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -52,16 +56,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
 
 	logger.Info("Cilium is already installed, checking if it needs upgrade")
 	upgradeInfo := cilium.BuildUpgradePlan(installation, spec)
+
+	logger.Info("Attempting to retrieve ConfigMap")
```
g-gaston, line 60: is this just for testing?

pkg/networking/cilium/upgrade_plan.go
```diff
@@ -46,6 +61,8 @@ type ComponentUpgradePlan struct {
 	UpgradeReason string
 	OldImage      string
 	NewImage      string
+	OldEnforcementPolicy string
+	NewEnforcementPolicy string
```
g-gaston, line 65: these fields might not make a lot of sense anymore since the config doesn't deal with versions and CM and deployments don't deal with policy enforcement. The abstraction has broken a bit.

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -52,16 +56,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
...
 
 		if result, err := r.upgrade(ctx, logger, client, installation, spec); err != nil {
 			return controller.Result{}, err
 		} else if result.Return() {
 			return result, nil
 		}
+	} else if upgradeInfo.ConfigUpgradeNeeded() {
+		logger.Info("Cilium config upgrade needed", "reason", upgradeInfo.Reason())
+		err := r.updateConfig(ctx, logger, client, installation, spec)
+		return controller.Result{}, err
```
g-gaston, line 74:
```suggestion
		if err := r.updateConfig(ctx, logger, client, installation, spec); err != nil {
		    return controller.Result{}, err
		}
```

pkg/api/v1alpha1/cluster_webhook.go
jonathanmeier5: `Nodes` is also immutable per @mitalipaygude. I didn't have a chance to deep dive into `ClusterNetwork.DNS` and the person who wrote that component is no longer on the team, so I was planning to make it immutable until we heard otherwise. There might be more info if you deep dive the code but I haven't had a chance to.

pkg/networking/cilium/reconciler/reconciler.go
jonathanmeier5: yes this was just testing garbage.

pkg/networking/cilium/upgrade_plan.go
jonathanmeier5: Yeah agreed. I just went with the simplest solution first. We still do need some notion of "this is what it is now" and "this is what we want it to be" though.

tatlat (commented):
Looks good. Not really a complaint since there are comments and different signatures but it was slightly confusing to see methods named Reason and reason in upgrade_plan.go. I can't  think of a better name, so it's fine to ignore this comment.

pkg/api/v1alpha1/cluster_webhook.go
taneyland: I would have to test to confirm the upgrade behavior since I added this feature a while back, but I think we should be safe to say that `ClusterNetwork.DNS` is mutable. If the user doesn't specify it, it will just use the default `resolv.conf` file.

g-gaston: Made `Nodes` and `DNS` immutable. I don't see any reason why they should be, but let's start with this and we can remove the protection as needed.

cxbrowne1207 (approved):

g-gaston:
no changes, just rebase to updated main for codecov report

mitalipaygude (approved):

Issue #3918, opened 2022-11-02
MachineHealthChecks cause endless loop during upgrade in slow environment


maxdrib:
**What happened**:
User was attempting to upgrade a large management cluster with many workload clusters. System slowed down significantly. new machines were being created as part of the RollingUpgrade, but taking a long time to get the Bootstrap Secret from CAPI. As a result, the MachineHealthCheck kept recycling those CAPI machines. 

**What you expected to happen**:
I was expecting CAPI to be running quickly and for the MachineHealthChecks to not take down the machine so quickly. Disabling the machinehealthcheck allowed the upgrade to proceed.

I think a potential solution here would be to make the machinehealthcheck timeout configurable

**How to reproduce it (as minimally and precisely as possible)**:
Potentially:
1. Create a management cluster with 8 workload clusters on it. Each having 3 etcd - 2 CP - 3 worker nodes
2. Trigger a rolling upgrade on the eks-a cluster such as changing the Offering or image

**Anything else we need to know?**:

**Environment**: Cloudstack 4.14
- EKS Anywhere Release: v0.8.x
- EKS Distro Release:

PR #4123, opened 2022-11-16
Add flag to set machine health check timeout (#3918)

msanjaq:
Before this change, an unhealthy machine health check would timeout after five minutes. This leads to an endless loop for a RolingUpgrade on a slow system. Disabling the health checks circumvents this issue, but is not ideal. This change adds a flag `--unhealthy-machine-timeout` which customers can use to make the upgrade process more reliable without sacrificing health checks.

*Issue #, if available:*  Resolves https://github.com/aws/eks-anywhere/issues/3918

*Description of changes:* Allow users to override the default unhealthy machine condition timeout using an `--unhealthy-machine-timeout` flag

*Testing (if applicable):* 

Created unit tests for the scenarios:
1. No environmental variable sets the timeout to the default value
2. Invalid environmental variable (e.g negatives, NaNs, large values which cause int overflows) sets the timeout to 5m
3. Valid environmental variable (e.g 20), sets the timeout to 20m

I performed a manual test by running the command
```eksctl anywhere create cluster --unhealthy-machine-timeout 20m -f mgmt_cluster.yaml```
and confirmed the timeout was set correctly with
```kubectl get -o yaml machinehealthcheck -A --kubeconfig sanjaqmo-healthcheck-test-mgmt/sanjaqmo-healthcheck-test-mgmt-eks-a-cluster.kubeconfig```

### DIFF
--- a/pkg/clustermanager/cluster_manager_test.go
+++ b/pkg/clustermanager/cluster_manager_test.go
@@ -1306,7 +1306,8 @@ func TestClusterManagerCreateEKSAResourcesFailure(t *testing.T) {
 	tt.Expect(c.CreateEKSAResources(ctx, tt.cluster, tt.clusterSpec, datacenterConfig, machineConfigs)).NotTo(Succeed())
 }
 
-var wantMHC = []byte(`apiVersion: cluster.x-k8s.io/v1beta1
+func expectedMachineHealthCheck(timeout time.Duration) []byte {
+	healthCheck := fmt.Sprintf(`apiVersion: cluster.x-k8s.io/v1beta1
 kind: MachineHealthCheck
 metadata:
   creationTimestamp: null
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by maxdrib — APPROVED** [review]

**Review by vivek-koppuru — APPROVED** [review]
/approve

File: pkg/clustermanager/cluster_manager_test.go [hunk 1]
```diff
@@ -1306,24 +1306,25 @@ func TestClusterManagerCreateEKSAResourcesFailure(t *testing.T) {
...
+	healthCheck := fmt.Sprintf(`apiVersion: cluster.x-k8s.io/v1beta1
 kind: MachineHealthCheck
 metadata:
   creationTimestamp: null
   name: fluxTestCluster-worker-1-worker-unhealthy
   namespace: eksa-system
 spec:
   clusterName: fluxTestCluster
-  maxUnhealthy: 40%
+  maxUnhealthy: 40%%
```
vivek-koppuru, line 1318: Is this a mistake?

vivek-koppuru (approved):
/approve

PR #4126, opened 2022-11-16
Default to Reconcile mode for CRS

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/vsphere/config/template-cp.yaml
```diff
@@ -452,6 +452,7 @@ spec:
     name: cloud-provider-vsphere-credentials
   - kind: ConfigMap
     name: cpi-manifests
+  strategy: Reconcile
```
g-gaston, line 455: CRS.spec.strategy is immutable, so this is not gonna work
We need to either delete the CRS before we apply the spec or just use a different name for the reconcile CRS so there is no collision. If we decide to use a new name maybe we should add something to clean up the old one. Although we can do as a follow up.

pkg/providers/vsphere/vsphere.go
```diff
@@ -967,20 +960,6 @@ func (p *vsphereProvider) ChangeDiff(currentSpec, newSpec *cluster.Spec) *types.
 }
 
 func (p *vsphereProvider) RunPostControlPlaneUpgrade(ctx context.Context, oldClusterSpec *cluster.Spec, clusterSpec *cluster.Spec, workloadCluster *types.Cluster, managementCluster *types.Cluster) error {
```
g-gaston, line 962:
```suggestion
func (p *vsphereProvider) RunPostControlPlaneUpgrade(_ context.Context, _ *cluster.Spec, _ *cluster.Spec, _ *types.Cluster, _ *types.Cluster) error {
```

pkg/providers/vsphere/config/template-cp.yaml
taneyland: Updated the name to be `<clusterName>-crs` instead of `<clusterName>-crs-0`. Will do the clean up in a follow up

g-gaston: Cool
Now that I think about it, what do you think about `<clusterName>-cpi-csi`?
Since that's what this crs manages

taneyland: yes, i like that! updated it

PR #4128, opened 2022-11-17
Add namespace support

vincentni:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/utils/urls/replace.go
```diff
@@ -21,5 +21,25 @@ func ReplaceHost(orgURL, host string) string {
 		u.Scheme = ""
 	}
 	u.Host = host
-	return strings.TrimPrefix(u.String(), "//")
+	return strings.ReplaceAll(strings.TrimPrefix(u.String(), "//"), "%2F", "/")
+}
+
+// ToAPIEndpoint turns URL to a valid API endpoint used in
```
chrisdoherty4, line 27: I think we need more detail on the allowed formats of `url` beacuse, as is, I don't think I'd know how or what this function can be used for. I'm also not sure it belongs in the `urls` package because its focused heavily on registries rather than generic URL manipulation.

pkg/utils/urls/replace.go
```diff
@@ -21,5 +21,25 @@ func ReplaceHost(orgURL, host string) string {
...
+	index := strings.Index(URL, "/")
+	if index == -1 {
+		return URL
+	}
+	return URL[:index] + "/v2" + URL[index:]
+}
+
+// ToAPIEndpoints utilizes ToAPIEndpoint to turn all URLs from a
+// map to valid API endpoints for a local registry.
+func ToAPIEndpoints(URLs map[string]string) map[string]string {
```
chrisdoherty4, line 39: Why is this a map and not a slice?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+		registryMirrorMappings[constants.DefaultRegistry] = base
+	}
+	if _, ok := registryMirrorMappings[constants.DefaultPackageRegistryRegex]; !ok {
+		registryMirrorMappings[constants.DefaultPackageRegistryRegex] = base
+	}
+	return registryMirrorMappings
+}
+
+// OCINamespacesSliceEqual is used to check equality of the OCINamespaces fields of two RegistryMirrorConfgiuration.
+func OCINamespacesSliceEqual(a, b []OCINamespace) bool {
```
chrisdoherty4, line 204: Do we have an existing pattern of putting utility functions in the `api` package? Seems peculiar to have this live here.

pkg/clusterapi/registry_mirror.go
```diff
@@ -47,7 +48,7 @@ func registryMirrorConfig(registryMirrorConfig *v1alpha1.RegistryMirrorConfigura
 
 	if registryMirrorConfig.CACertContent != "" {
 		files = append(files, bootstrapv1.File{
-			Path:    fmt.Sprintf("/etc/containerd/certs.d/%s/ca.crt", registryAddress),
+			Path:    fmt.Sprintf("/etc/containerd/certs.d/%s/ca.crt", registryAddressMappings[constants.DefaultRegistryMirrorKey]),
```
chrisdoherty4, line 51: I think the function in the api package that creates this mapping needs to return a specialized type with a function for retrieving the default. I mention this because currently, using the builtin map, you're requiring all consumers to embed the key knowledge of `constants.DefaultRegistryMirrorKey`. 

E.g.

```go
type OCINamespaceMap map[string]string

// define GetDefault func
```

pkg/constants/constants.go
```diff
@@ -40,8 +40,10 @@ const (
 	EksaLicenseName        = "eksa-license"
 	EksaPackagesName       = "eksa-packages"
 
-	DefaultRegistry            = "public.ecr.aws"
-	CloudstackAnnotationSuffix = "cloudstack.anywhere.eks.amazonaws.com/v1alpha1"
+	DefaultRegistryMirrorKey    = "base"
+	DefaultRegistry             = "public.ecr.aws"
+	DefaultPackageRegistryRegex = "783794618700.dkr.ecr.*.amazonaws.com"
+	CloudstackAnnotationSuffix  = "cloudstack.anywhere.eks.amazonaws.com/v1alpha1"
```
chrisdoherty4, line 46: Whats the rationale for housing these constants here? The `constants` package is turning into a dumping ground for totally unrelated things beyond the fact their definitions are typed with `const`.

chrisdoherty4 (commented):
Did an initial run through, will finish reviewing later today.

pkg/api/v1alpha1/cluster.go
```diff
@@ -303,11 +304,21 @@ func (c *Cluster) ClearPauseAnnotation() {
...
+	if mirrors == nil {
 		return ""
 	}
 
-	return net.JoinHostPort(c.Spec.RegistryMirrorConfiguration.Endpoint, c.Spec.RegistryMirrorConfiguration.Port)
+	return mirrors[constants.DefaultRegistry]
+}
+
+// RegistryMirrors returns registry mirror mappings of all upstream registries.
+func (c *Cluster) RegistryMirrors() map[string]string {
```
chrisdoherty4, line 316: Do we need to bubble this up to the `Cluster` object? Having 2 methods doing the same thing with very different names seems convoluted.

pkg/dependencies/factory.go
```diff
@@ -953,6 +958,10 @@ func (f *Factory) WithPackageControllerClient(spec *cluster.Spec, kubeConfig str
 
 		httpProxy, httpsProxy, noProxy := getProxyConfiguration(spec)
 		eksaAccessKeyId, eksaSecretKey, eksaRegion := os.Getenv(config.EksaAccessKeyIdEnv), os.Getenv(config.EksaSecretAccessKeyEnv), os.Getenv(config.EksaRegionEnv)
+		endpoints := (map[string]string)(nil)
```
chrisdoherty4, line 961:
```suggestion
		var endpoints map[string]string
```

chrisdoherty4: We're also passing this on to other things. This seems a little dangerous also. We're passing this map onward to other data structures. If those data structures try to append something the program will panic. It seems like this would be safer if it were initialized with `make(map[string]string)`.

pkg/dependencies/factory_test.go
```diff
@@ -276,7 +280,11 @@ func TestFactoryBuildWithRegistryMirrorAuth(t *testing.T) {
 	tt := newTest(t, vsphere)
 	deps, err := dependencies.NewFactory().
 		WithLocalExecutables().
-		WithRegistryMirror("1.2.3.4:443", true).
+		WithRegistryMirror(
+			map[string]string{
+				"base":           "1.2.3.4:443",
```
chrisdoherty4, line 285: Magic strings.

pkg/executables/config/kind.yaml
```diff
@@ -13,20 +13,20 @@ kubeadmConfigPatches:
...
-{{- if (ne .RegistryMirrorEndpoint "") }}
+{{- if .RegistryMirrorConfiguration }}
 containerdConfigPatches:
   - |
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
       [plugins."io.containerd.grpc.v1.cri".registry.mirrors."public.ecr.aws"]
-        endpoint = ["https://{{.RegistryMirrorEndpoint}}"]
-      [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.RegistryMirrorEndpoint}}".tls]
+        endpoint = ["https://{{index .RegistryMirrorConfiguration "public.ecr.aws"}}"]
+      [plugins."io.containerd.grpc.v1.cri".registry.configs."{{index .RegistryMirrorConfiguration "base"}}".tls]
```
chrisdoherty4, line 22: This feels a little error prone and flaky because it pertains to a string defined in `constants` that itself living far from the core code.

We could create some functions we inject into the templater which would allow us to define the functions in Go code and clearly document them. Its similar to my other comments mentioning the use of `map[string]string` and how that's super difficult to appreciate semantics for as a maintainer.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+			registryMirrorMappings[constants.DefaultPackageRegistryRegex] = registry
+		} else {
+			registryMirrorMappings[ociNamespace.Registry] = registry
+		}
+	}
+	if _, ok := registryMirrorMappings[constants.DefaultRegistry]; !ok {
+		registryMirrorMappings[constants.DefaultRegistry] = base
+	}
+	if _, ok := registryMirrorMappings[constants.DefaultPackageRegistryRegex]; !ok {
+		registryMirrorMappings[constants.DefaultPackageRegistryRegex] = base
```
g-gaston, line 198: same comment here, I think we should ask users to do this explicitly.

pkg/clusterapi/config/containerd_config_append.toml
```diff
@@ -1,10 +1,12 @@
 [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors."public.ecr.aws"]
-    endpoint = ["https://{{.registryMirrorAddress}}"]
+    endpoint = ["https://{{index .registryMirrorAddressMappings "public.ecr.aws"}}"]
```
g-gaston, line 3: why is `public.ecr.aws` hardcoded here? shouldn't it be part of the passed mappings?

pkg/clusterapi/config/containerd_config_append.toml
```diff
@@ -1,10 +1,12 @@
...
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."783794618700.dkr.ecr.*.amazonaws.com"]
+    endpoint = ["https://{{index .registryMirrorAddressMappings "783794618700.dkr.ecr.*.amazonaws.com"}}"]
```
g-gaston, line 5: same comment here hardcoding the `783794618700` account

pkg/clusterapi/registry_mirror.go
g-gaston: I was gonna comment something similar. Can we make that `GetRegistryMirrorAddressMappings` (or a different one) to return a struct that has this info more normalized?
I would expect the caller to not have to care about the map content except to just iterate over it to read the duples.

pkg/dependencies/factory.go
```diff
@@ -284,7 +289,7 @@ func (f *Factory) WithExecutableBuilder() *Factory {
 		if f.executablesConfig.useDockerContainer {
 			image := f.executablesConfig.image
 			if f.registryMirror != nil {
-				image = urls.ReplaceHost(f.executablesConfig.image, f.registryMirror.endpoint)
+				image = urls.ReplaceHost(f.executablesConfig.image, f.registryMirror.endpoints[constants.DefaultRegistry])
```
g-gaston, line 292: I think we should check the mappings and only do the replace if there is one than matches the image.

I think we could benefit from a service object that does this for us that we can inject in other places, instead of keep using `ReplaceHost` everywhere.

pkg/dependencies/factory.go
```diff
@@ -659,7 +664,7 @@ func (f *Factory) WithHelm(opts ...executables.HelmOpt) *Factory {
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		if f.registryMirror != nil {
-			opts = append(opts, executables.WithRegistryMirror(f.registryMirror.endpoint))
+			opts = append(opts, executables.WithRegistryMirror(f.registryMirror.endpoints[constants.DefaultRegistry]))
```
g-gaston, line 667: we do this `endpoints[constants.DefaultRegistry]` in multiple places. It's probably worth to have another field in `f.registryMirror` for this

That said, shouldn't helm also support multiple registry mappings now? I think everywhere that used to understand the concept of registry mirror for public ecr should now understand the concept of multiple mappings

pkg/executables/config/kind.yaml
```diff
@@ -13,20 +13,20 @@ kubeadmConfigPatches:
...
 {{- if (eq .RegistryCACertPath "") }}
         insecure_skip_verify = true
 {{- else }}
-        ca_file = "/etc/containerd/certs.d/{{.RegistryMirrorEndpoint}}/ca.crt"
+        ca_file = "/etc/containerd/certs.d/{{index .RegistryMirrorConfiguration "base"}}/ca.crt"
```
g-gaston, line 26: can't we pass here the folder name for the cert?

pkg/executables/config/kind.yaml
```diff
@@ -13,20 +13,20 @@ kubeadmConfigPatches:
...
 {{- end }}
 {{- if .RegistryAuth }}
-      [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.RegistryMirrorEndpoint}}".auth]
+      [plugins."io.containerd.grpc.v1.cri".registry.configs."{{index .RegistryMirrorConfiguration "base"}}".auth]
```
g-gaston, line 29: same here passing the name

pkg/executables/kind.go
```diff
@@ -181,20 +181,21 @@ func (k *Kind) DeleteBootstrapCluster(ctx context.Context, cluster *types.Cluste
 
 func (k *Kind) setupExecConfig(clusterSpec *cluster.Spec) error {
 	bundle := clusterSpec.VersionsBundle
+	registry := clusterSpec.Cluster.RegistryMirror()
 	k.execConfig = &kindExecConfig{
-		KindImage:            urls.ReplaceHost(bundle.EksD.KindNode.VersionedImage(), clusterSpec.Cluster.RegistryMirror()),
-		KubernetesRepository: bundle.KubeDistro.Kubernetes.Repository,
+		KindImage:            urls.ReplaceHost(bundle.EksD.KindNode.VersionedImage(), registry),
+		KubernetesRepository: urls.ReplaceHost(bundle.KubeDistro.Kubernetes.Repository, registry),
```
g-gaston, line 187: this another place that can benefit from a `ImageRegistryRepository` (whatever other name) that takes an image url and return another one with any necessary replacement done based on the registry mirror mapping configs

pkg/utils/urls/replace.go
```diff
@@ -21,5 +21,25 @@ func ReplaceHost(orgURL, host string) string {
 		u.Scheme = ""
 	}
 	u.Host = host
-	return strings.TrimPrefix(u.String(), "//")
+	return strings.ReplaceAll(strings.TrimPrefix(u.String(), "//"), "%2F", "/")
```
g-gaston, line 24: this seems hacky. We should change the implementation of this method so `%2F` is not added

pkg/utils/urls/replace.go
```diff
@@ -21,5 +21,25 @@ func ReplaceHost(orgURL, host string) string {
...
+}
+
+// ToAPIEndpoint turns URL to a valid API endpoint used in
+// a containerd config file for a local registry.
+func ToAPIEndpoint(URL string) string {
```
g-gaston, line 29: this probably fits better in the utils/oci package

pkg/utils/urls/replace.go
```diff
@@ -21,5 +21,25 @@ func ReplaceHost(orgURL, host string) string {
...
+	index := strings.Index(URL, "/")
```
g-gaston, line 30: is this safe? what about `oci://bbalba.com:123/balbal`

pkg/providers/vsphere/config/template-md.yaml
```diff
@@ -72,19 +72,19 @@ spec:
...
         owner: root:root
-        path: "/etc/containerd/certs.d/{{.registryMirrorConfiguration}}/ca.crt"
+        path: "/etc/containerd/certs.d/{{index .registryMirrorConfiguration "base"}}/ca.crt"
 {{- end }}
 {{- if .registryMirrorConfiguration }}
       - content: |
           [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
             [plugins."io.containerd.grpc.v1.cri".registry.mirrors."public.ecr.aws"]
-              endpoint = ["https://{{.registryMirrorConfiguration}}"]
+              endpoint = ["https://{{index .registryMirrorConfiguration "public.ecr.aws"}}"]
```
g-gaston, line 81: why don't just iterate over the mappings and them without checking for specific keys?

pkg/providers/vsphere/config/template-cp.yaml
```diff
@@ -368,7 +368,7 @@ spec:
 {{- end }}
 {{- if and .registryMirrorConfiguration (eq .format "bottlerocket") }}
       registryMirror:
-        endpoint: {{.registryMirrorConfiguration}}
+        endpoint: {{index .registryMirrorConfiguration "public.ecr.aws"}}
```
g-gaston, line 371: i would vote to pass an specific value for this to the template, this hides too much logic in the template

pkg/providers/vsphere/config/template-cp.yaml
```diff
@@ -271,19 +271,19 @@ spec:
...
-          [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.registryMirrorConfiguration}}".tls]
-            ca_file = "/etc/containerd/certs.d/{{.registryMirrorConfiguration}}/ca.crt"
+          [plugins."io.containerd.grpc.v1.cri".registry.configs."{{index .registryMirrorConfiguration "base"}}".tls]
+            ca_file = "/etc/containerd/certs.d/{{index .registryMirrorConfiguration "base"}}/ca.crt"
           {{- end }}
           {{- if .registryAuth }}
-          [plugins."io.containerd.grpc.v1.cri".registry.configs."{{.registryMirrorConfiguration}}".auth]
+          [plugins."io.containerd.grpc.v1.cri".registry.configs."{{index .registryMirrorConfiguration "base"}}".auth]
             username = "{{.registryUsername}}"
             password = "{{.registryPassword}}"
```
g-gaston, line 288: We can probably benefit from extracting this whole containrd config to a separate template so we can reuse it in different providers and also test it independently.

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -2437,6 +2463,60 @@ func TestClusterRegistryMirror(t *testing.T) {
...
+			want:    nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			result := tt.cluster.RegistryMirrors()
+			g.Expect(len(result)).To(Equal(len(tt.want)))
+			for k, v := range tt.want {
+				if regexp.MustCompile(constants.DefaultPackageRegistryRegex).MatchString(k) {
```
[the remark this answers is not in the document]
vincentni: The key for curated packages in the map is a regex ("783794618700.dkr.ecr.*.amazonaws.com") to handle multiple regions, but the registry for curated packages from the cluster spec is the one with a concrete region (e.g., 783794618700.dkr.ecr.us-west-2.amazonaws.com).

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+	return n.Endpoint == o.Endpoint && n.Port == o.Port && n.CACertContent == o.CACertContent && n.InsecureSkipVerify == o.InsecureSkipVerify && n.Authenticate == o.Authenticate && OCINamespacesSliceEqual(n.OCINamespaces, o.OCINamespaces)
+}
+
+// GetRegistryMirrorAddressMappings reads OCINamespaces field of RegistryMirrorConfiguration, and returns
+// a mapping from original registry URLs to registry mirror URLs.
+func (n *RegistryMirrorConfiguration) GetRegistryMirrorAddressMappings() map[string]string {
+	registryMirrorMappings := make(map[string]string)
+	base := net.JoinHostPort(n.Endpoint, n.Port)
+	// add registry mirror base address
+	registryMirrorMappings[constants.DefaultRegistryMirrorKey] = base
```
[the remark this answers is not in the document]
vincentni: This is the key to retrieve the base registry address without any namespace. It is needed for tls and auth sections in containred configuration, because all namespaces under the same registry share the same tls and auth config (https://github.com/containerd/containerd/blob/main/docs/cri/registry.md).

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+	re := regexp.MustCompile(constants.DefaultPackageRegistryRegex)
+	// for each namespace, add corresponding endpoint
+	for _, ociNamespace := range n.OCINamespaces {
+		registry := base + "/" + ociNamespace.Namespace
+		if re.MatchString(ociNamespace.Registry) {
```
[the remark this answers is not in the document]
vincentni: This condition is to test if the incoming registry is for curated packages. As mentioned above, customers specify a registry with concrete region ("783794618700.dkr.ecr.us-west-2.amazonaws.com") in the cluster spec for curated packages, but the key in the map is a regex ("783794618700.dkr.ecr.*.amazonaws.com").

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+			// handle curated packages in all regions
+			registryMirrorMappings[constants.DefaultPackageRegistryRegex] = registry
+		} else {
+			registryMirrorMappings[ociNamespace.Registry] = registry
+		}
+	}
+	if _, ok := registryMirrorMappings[constants.DefaultRegistry]; !ok {
+		registryMirrorMappings[constants.DefaultRegistry] = base
```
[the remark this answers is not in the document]
vincentni: Well, this is for such a use case where customers specify the endpoint and port for the registry mirror but leave ociNamespaces unspecified. In this case, the code will set registry mirror mapping for both eksa artifacts and curated packages to the same as the base registry mirror where `base := net.JoinHostPort(n.Endpoint, n.Port)`.

pkg/api/v1alpha1/cluster_types.go
vincentni: see the reply above.

pkg/clusterapi/config/containerd_config_append.toml
vincentni: this is just used to retrieve the value for that particular key from the mapping. Also, `public.ecr.aws` and `783794618700.dkr.ecr.*.amazonaws.com` are production registry for eksa artifacts and curated packages respectively. They are extremely unlikely to change.

pkg/clusterapi/config/containerd_config_append.toml
vincentni: same as above

pkg/utils/urls/replace.go
vincentni: If we make it as a slice, the relationship between which position in the slice and which registry that position maps to is implicit, thus resulting in some maintenance issue. Whereas, map turns that relationship to be explicit.

pkg/api/v1alpha1/cluster_types.go
vincentni: Although it is a utility function to facilitate the equality check of `OCINamespaces` field, this is very likely the only place for its usage.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -148,14 +155,75 @@ type RegistryMirrorConfiguration struct {
...
+}
+
+// GetRegistryMirrorAddressMappings reads OCINamespaces field of RegistryMirrorConfiguration, and returns
+// a mapping from original registry URLs to registry mirror URLs.
+func (n *RegistryMirrorConfiguration) GetRegistryMirrorAddressMappings() map[string]string {
+	registryMirrorMappings := make(map[string]string)
+	base := net.JoinHostPort(n.Endpoint, n.Port)
+	// add registry mirror base address
+	registryMirrorMappings[constants.DefaultRegistryMirrorKey] = base
+	re := regexp.MustCompile(constants.DefaultPackageRegistryRegex)
```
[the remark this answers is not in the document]
vincentni: Valid point. But I'd like to say `constants.DefaultPackageRegistryRegex` is a constant defined by us, so it is not some value that may vary and fail the regular expression compilation in the runtime. Failure on that constant always gets caught by unit tests before the code release.

pkg/dependencies/factory.go
vincentni: The map sets to `nil` in case of no registry mirror specified in the cluster spec, so making it as a empty map and leaving possibility of adding elements to it later is not logically intended.

pkg/registrymirror/containerd/utils.go
```diff
@@ -0,0 +1,39 @@
...
+			index = strings.Index(URL, u.Path)
+		}
+	} else {
+		index = strings.Index(URL, "/")
+	}
+	if index == -1 {
+		return URL
+	}
+	return URL[:index] + "/v2" + URL[index:]
+}
```
g-gaston, line 29: Maybe we could do something different to avoid relying on index based string manipulation, which in my experience can lead to bugs, specially with URLs due to their complexity
For example

```suggestion
func ToAPIEndpoint(url string) string {
	u, err := parseURL(url)
	if err != nil {
		return url
	}

	if u.Path != "" {
		u.Path = filepath.Join("v2", u.Path)
	}

	return strings.TrimPrefix(u.String(), "//")
}

func parseURL(in string) (*url.URL, error) {
	urlIn := in
	if !strings.Contains(in, "//") {
		urlIn = "//" + in
	}

	return url.Parse(urlIn)
}
```

We should also document what happens when the provided url is malformed (it returns the original input?)

pkg/providers/cloudstack/testdata/expected_results_mirror_config_with_cert_md.yaml
```diff
@@ -36,13 +36,13 @@ spec:
...
-        path: "/etc/containerd/certs.d/1.2.3.4/ca.crt"
+        path: "/etc/containerd/certs.d/1.2.3.4:443/ca.crt"
       - content: |
           [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
             [plugins."io.containerd.grpc.v1.cri".registry.mirrors."public.ecr.aws"]
-              endpoint = ["https://1.2.3.4"]
-            [plugins."io.containerd.grpc.v1.cri".registry.configs."1.2.3.4".tls]
-              ca_file = "/etc/containerd/certs.d/1.2.3.4/ca.crt"
+              endpoint = ["https://1.2.3.4:443"]
+            [plugins."io.containerd.grpc.v1.cri".registry.configs."1.2.3.4:443".tls]
```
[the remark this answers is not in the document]
vincentni: The correct format should be [ip address:port] and this one is not consistent with others (e.g., https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/testdata/expected_results_mirror_config_md.yaml#L24). The default port is 443.

pkg/clusterapi/registry_mirror.go
```diff
@@ -54,18 +55,19 @@ func SetRegistryMirrorInKubeadmConfigTemplateForUbuntu(kct *bootstrapv1.KubeadmC
...
 
-func registryMirrorConfigContent(registryAddress, registryCert string, insecureSkip bool) (string, error) {
+func registryMirrorConfigContent(registryMirror *registrymirror.RegistryMirror, registryCert string, insecureSkip bool) (string, error) {
 	val := values{
-		"registryMirrorAddress": registryAddress,
-		"registryCACert":        registryCert,
-		"insecureSkip":          insecureSkip,
+		"registryMirrorMap": containerd.ToAPIEndpoints(registryMirror.NamespacedRegistryMap),
+		"mirrorBase":        registryMirror.BaseRegistry,
+		"registryCACert":    registryCert,
```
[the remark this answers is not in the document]
vincentni: I have the same thoughts. We can gradually move all registry mirror related stuff to this struct, e.g., username and password. Since this PR is already big enough, those improvements can be made in separate smaller PRs in the future.

pkg/executables/kind.go
```diff
@@ -181,20 +181,22 @@ func (k *Kind) DeleteBootstrapCluster(ctx context.Context, cluster *types.Cluste
...
-		KindImage:            urls.ReplaceHost(bundle.EksD.KindNode.VersionedImage(), clusterSpec.Cluster.RegistryMirror()),
-		KubernetesRepository: bundle.KubeDistro.Kubernetes.Repository,
+		KindImage:            registryMirror.ReplaceRegistry(bundle.EksD.KindNode.VersionedImage()),
+		KubernetesRepository: registryMirror.ReplaceRegistry(bundle.KubeDistro.Kubernetes.Repository),
 		KubernetesVersion:    bundle.KubeDistro.Kubernetes.Tag,
-		EtcdRepository:       bundle.KubeDistro.Etcd.Repository,
+		EtcdRepository:       registryMirror.ReplaceRegistry(bundle.KubeDistro.Etcd.Repository),
 		EtcdVersion:          bundle.KubeDistro.Etcd.Tag,
-		CorednsRepository:    bundle.KubeDistro.CoreDNS.Repository,
+		CorednsRepository:    registryMirror.ReplaceRegistry(bundle.KubeDistro.CoreDNS.Repository),
```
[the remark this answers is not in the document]
vincentni: Without such changes, from the testing, those images are not pulling from local registry during `bootstrap-cluster-init` phase.

pkg/executables/kind.go
```diff
@@ -181,20 +181,22 @@ func (k *Kind) DeleteBootstrapCluster(ctx context.Context, cluster *types.Cluste
...
 		CorednsVersion:       bundle.KubeDistro.CoreDNS.Tag,
 		env:                  make(map[string]string),
 	}
 	if clusterSpec.Cluster.Spec.RegistryMirrorConfiguration != nil {
-		k.execConfig.RegistryMirrorEndpoint = net.JoinHostPort(clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.Endpoint, clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.Port)
+		k.execConfig.MirrorBase = registryMirror.BaseRegistry
+		k.execConfig.RegistryMirrorMap = containerd.ToAPIEndpoints(registryMirror.NamespacedRegistryMap)
```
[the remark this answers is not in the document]
g-gaston: Got you

- [ ] Standalone template for registry mirror and mirror map cleanup

pkg/providers/cloudstack/testdata/expected_results_mirror_config_with_cert_md.yaml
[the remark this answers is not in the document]
g-gaston: is `:port` required by containerd? even for the default port?

pkg/clusterapi/registry_mirror.go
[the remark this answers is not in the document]
g-gaston: Sounds good. Could you create a ticket to track this follow up items, please?

- [ ] Add registry mirror cert to `registrymirror.RegistryMirror` and possibly credentials as well

pkg/providers/cloudstack/testdata/expected_results_mirror_config_with_cert_md.yaml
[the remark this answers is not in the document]
vincentni: If `http` or `https` is used, then default port can be left off, otherwise has to be specified.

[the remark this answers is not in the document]
g-gaston: If we list the port, do we need to include that when tagging and pushing the image?

pkg/api/v1alpha1/cluster.go
```diff
@@ -698,6 +693,28 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
...
+
+	mirrorCount := 0
+	ociNamespaces := clusterConfig.Spec.RegistryMirrorConfiguration.OCINamespaces
+	for _, ociNamespace := range ociNamespaces {
+		if ociNamespace.Registry == "" {
+			return errors.New("registry can't be set to empty in OCINamespaces")
+		}
+		if re.MatchString(ociNamespace.Registry) {
+			mirrorCount++
+			// More than one curated package mirrors would confuse package controller where to pull packages
```
g-gaston, line 705:
```suggestion
			// More than one mirror for curated package would introduce ambiguity in the package controller
			// when deciding where to pull packages from, since it only allows for one source registry
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -84,12 +85,29 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
-	ociUri := fmt.Sprintf("%s%s", "oci://", pc.uri)
-	registry := GetRegistry(pc.uri)
-
-	sourceRegistry := fmt.Sprintf("sourceRegistry=%s", registry)
+	ociURI := fmt.Sprintf("%s%s", "oci://", pc.registryMirror.ReplaceRegistry(pc.chart.Image()))
+	var values []string
 	clusterName := fmt.Sprintf("clusterName=%s", pc.clusterName)
-	values := []string{sourceRegistry, clusterName}
+	if pc.registryMirror != nil {
+		// account is added as part of registry name in package controller chart
```
g-gaston, line 92: sorry I'm still missing to understand why this is necessary
why are we adding that to the `sourceRegistry` and `defaultRegistry`?

pkg/dependencies/factory.go
```diff
@@ -259,7 +255,11 @@ func (f *Factory) WithDockerLogin() *Factory {
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
 		username, password, _ := config.ReadCredentials()
-		err := f.executablesConfig.dockerClient.Login(context.Background(), f.registryMirror.endpoint, username, password)
+		endpoint := ""
+		if f.registryMirror != nil {
+			endpoint = f.registryMirror.BaseRegistry
+		}
+		err := f.executablesConfig.dockerClient.Login(context.Background(), endpoint, username, password)
```
g-gaston, line 262: do we need this login if registry mirror is not set?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -137,18 +138,11 @@ func NewProvider(
...
-	} else {
-		registryMirror = nil
-	}
-
 	return &Provider{
 		clusterConfig:         clusterConfig,
 		datacenterConfig:      datacenterConfig,
 		machineConfigs:        machineConfigs,
-		stackInstaller:        stack.NewInstaller(docker, writer, helm, constants.EksaSystemNamespace, clusterConfig.Spec.ClusterNetwork.Pods.CidrBlocks[0], registryMirror),
+		stackInstaller:        stack.NewInstaller(docker, writer, helm, constants.EksaSystemNamespace, clusterConfig.Spec.ClusterNetwork.Pods.CidrBlocks[0], registrymirror.FromClusterRegistryMirrorConfiguration(clusterConfig.Spec.RegistryMirrorConfiguration)),
```
g-gaston, line 145: nit, probably simpler?:
```suggestion
		stackInstaller:        stack.NewInstaller(docker, writer, helm, constants.EksaSystemNamespace, clusterConfig.Spec.ClusterNetwork.Pods.CidrBlocks[0], registrymirror.FromCluster(clusterConfig)),
```

pkg/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+// RegistryMirror configures mirror mappings for artifact registries.
+type RegistryMirror struct {
+	// the address of registry mirror without namespace
+	BaseRegistry string
```
g-gaston, line 17:
```suggestion
	// BaseRegistry is the address of the registry mirror without namespace. Just the host and the port.
	BaseRegistry string
```

pkg/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,102 @@
...
+	// it stores mirror mappings for artifact registries
```
g-gaston, line 18:
```suggestion
	// NamespacedRegistryMap stores mirror mappings for artifact registries
```

pkg/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,102 @@
...
+	// TODO: handle this case after BottleRocket supports multiple registry mirros
+	return ""
+}
+
+// CuratedPackagesMirror returns the mirror for curated packages.
+func (r *RegistryMirror) CuratedPackagesMirror() string {
+	if v, ok := r.NamespacedRegistryMap[constants.DefaultCuratedPackagesRegistryRegex]; ok {
+		return v
+	}
+	return ""
```
g-gaston, line 76:
```suggestion
	return r.NamespacedRegistryMap[constants.DefaultCuratedPackagesRegistryRegex]
```

pkg/registrymirror/registrymirror.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+// CoreEKSAMirror returns the configured mirror for public.ecr.aws.
+func (r *RegistryMirror) CoreEKSAMirror() string {
+	if v, ok := r.NamespacedRegistryMap[constants.DefaultCoreEKSARegistry]; ok {
+		return v
+	}
+	// TODO: handle this case after BottleRocket supports multiple registry mirros
+	return ""
```
g-gaston, line 68:
```suggestion
	return r.NamespacedRegistryMap[constants.DefaultCoreEKSARegistry]
```

g-gaston (commented):
looks good, minor suggestions

pkg/curatedpackages/packagecontrollerclient.go
vincentni: https://github.com/aws/eks-anywhere-packages/blob/main/charts/eks-anywhere-packages/values.yaml#L15-L18
they incorporate account into the registry address

g-gaston: Got you, can you add that info to the comment? I'm sure I'll read this code again in the future and forgot why this was done and that will help me :)

PR #4132, opened 2022-11-17
Delete unwanted MachineDeployments in reconcile

g-gaston:
*Description of changes:*
Fixes a bug where machines didn't get deleted when removing worker node groups from the eks-a cluster object.

*Testing (if applicable):*
Unite tests and manual e2e with tilt

### DIFF
--- a/pkg/controller/clusters/workers.go
+++ b/pkg/controller/clusters/workers.go
@@ -49,7 +49,7 @@ func (g *WorkerGroup) objects() []client.Object {
 }
 
 // ToWorkers converts the generic clusterapi Workers definition to the concrete one defined
-// here. It's just a helper for callers generating workers spec using the clusterpai package.
+// here. It's just a helper for callers generating workers spec using the clusterapi package.
 func ToWorkers[M clusterapi.Object[M]](capiWorkers *clusterapi.Workers[M]) *Workers {
 	w := &Workers{
 		Groups: make([]WorkerGroup, 0, len(capiWorkers.Groups)),
@@ -76,7 +76,7 @@ func ReconcileWorkersForEKSA(ctx context.Context, log logr.Logger, c client.Clie
 	}
 
 	if capiCluster == nil {
-		// cluster doesn't exit, this might be transient, requeuing
+		// cluster doesn't exist, this might be transient, requeuing
 		log.Info("CAPI cluster doesn't exist yet, this might be transient if the CP have just been created, requeueing")
 		return controller.ResultWithRequeue(5 * time.Second), nil
 	}
--- a/pkg/controller/clusters/workers.go
+++ b/pkg/controller/clusters/workers.go
@@ -1,0 +1,122 @@
+package clusters
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	kubeadmv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/collection"
+	"github.com/aws/eks-anywhere/pkg/controller"
+	"github.com/aws/eks-anywhere/pkg/controller/serverside"
+)
+
+// Workers represents the CAPI spec for an eks-a cluster's workers.
+type Workers struct {
+	Groups []WorkerGroup
+}
+
+// objects returns a list of API objects for a collection of worker groups.
+func (w *Workers) objects() []client.Object {
+	objs := make([]client.Object, 0, len(w.Groups)*3)
+	for _, g := range w.Groups {
+		objs = append(objs, g.objects()...)
+	}
+
+	return objs
+}
+
+// WorkerGroup represents the CAPI spec for an eks-a worker group.
+type WorkerGroup struct {
+	KubeadmConfigTemplate   *kubeadmv1.KubeadmConfigTemplate
+	MachineDeployment       *clusterv1.MachineDeployment
+	ProviderMachineTemplate client.Object
+}
+
+func (g *WorkerGroup) objects() []client.Object {
+	return []client.Object{
+		g.KubeadmConfigTemplate,
+		g.MachineDeployment,
+		g.ProviderMachineTemplate,
+	}
+}
+
+// ToWorkers converts the generic clusterapi Workers definition to the concrete one defined
+// here. It's just a helper for callers generating workers spec using the clusterpai package.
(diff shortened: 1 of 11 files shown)


pkg/controller/clusters/workers.go
```diff
@@ -0,0 +1,122 @@
...
+func (g *WorkerGroup) objects() []client.Object {
+	return []client.Object{
+		g.KubeadmConfigTemplate,
+		g.MachineDeployment,
+		g.ProviderMachineTemplate,
+	}
+}
+
+// ToWorkers converts the generic clusterapi Workers definition to the concrete one defined
+// here. It's just a helper for callers generating workers spec using the clusterpai package.
```
mitalipaygude, line 52:
```suggestion
// here. It's just a helper for callers generating workers spec using the clusterapi package.
```

pkg/providers/vsphere/reconciler/reconciler_test.go
```diff
@@ -172,6 +172,10 @@ func TestReconcilerControlPlaneIsNotReady(t *testing.T) {
 
 func TestReconcilerReconcileWorkersSuccess(t *testing.T) {
 	tt := newReconcilerTest(t)
+	capiCluster := capiCluster(func(c *clusterv1.Cluster) {
```
cxbrowne1207, line 175:
```suggestion
	capiCluster := test.CAPICluster(func(c *clusterv1.Cluster) {
```

pkg/providers/snow/reconciler/reconciler_test.go
```diff
@@ -148,6 +148,10 @@ func TestReconcilerValidateMachineConfigsMachineConfigNotValidated(t *testing.T)
 
 func TestReconcilerReconcileWorkers(t *testing.T) {
 	tt := newReconcilerTest(t)
+	capiCluster := capiCluster(func(c *clusterv1.Cluster) {
```
cxbrowne1207, line 151:
```suggestion
	capiCluster := test.CAPICluster(func(c *clusterv1.Cluster) {
```

pkg/controller/clusters/workers.go
```diff
@@ -0,0 +1,122 @@
...
+// It takes care of applying all desired objects in the Workers spec and deleting the
+// old MachineDeployments that are not in it.
+func ReconcileWorkersForEKSA(ctx context.Context, log logr.Logger, c client.Client, cluster *anywherev1.Cluster, w *Workers) (controller.Result, error) {
+	capiCluster, err := controller.GetCAPICluster(ctx, c, cluster)
+	if err != nil {
+		return controller.Result{}, errors.Wrap(err, "reconciling workers for EKS-A cluster")
+	}
+
+	if capiCluster == nil {
+		// cluster doesn't exit, this might be transient, requeuing
```
mitalipaygude, line 79:
```suggestion
		// cluster doesn't exist, this might be transient, requeuing
```

cxbrowne1207 (commented):
Looking good! Left a few suggested changes.

pkg/controller/clusters/workers_test.go
```diff
@@ -0,0 +1,271 @@
...
+	api.ShouldEventuallyNotExist(ctx, existingMachineDeployment2)
+	api.ShouldEventuallyExist(ctx, existingMachineDeployment3)
+}
+
+func TestReconcileWorkersErrorApplyingObjects(t *testing.T) {
+	g := NewWithT(t)
+	c := env.Client()
+	ctx := context.Background()
+	ns := "fake-ns"
+	// ns doesn't exit, it will fail
```
mitalipaygude, line 73:
```suggestion
	// ns doesn't exist, it will fail
```

pkg/controller/clusters/workers.go
```diff
@@ -0,0 +1,122 @@
...
+			MachineDeployment:       g.MachineDeployment,
+			KubeadmConfigTemplate:   g.KubeadmConfigTemplate,
+			ProviderMachineTemplate: g.ProviderMachineTemplate,
+		})
+	}
+
+	return w
+}
+
+// ReconcileWorkersForEKSA orchestrates the worker node reconciliation logic for a particular EKS-A cluster.
```
mitalipaygude, line 69: Do you mind explaining why we need reconcile workers for EKSA differently?

pkg/controller/clusters/workers.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	machineDeployments := &clusterv1.MachineDeploymentList{}
+	if err := c.List(ctx, machineDeployments,
+		client.MatchingLabels{clusterv1.ClusterLabelName: cluster.Name},
+		client.InNamespace(cluster.Namespace)); err != nil {
+		return controller.Result{}, errors.Wrap(err, "listing current machine deployments")
+	}
+
+	desiredMachineDeploymentNames := collection.MapSet(w.Groups, func(g WorkerGroup) string {
```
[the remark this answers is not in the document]
g-gaston: It's creating a set. It maps a slice to a set using a closure accessor. I added some docs to the `collection` package methods.

so at the end, `desiredMachineDeploymentNames` contains a list of unique machine deployment names.

pkg/controller/clusters/workers.go
g-gaston: What do you mean by differently?
It's not different than any one capi cluster, it actually uses the `ReconcileWorkers` implementation. But `ReconcileWorkersForEKSA` takes an eksa cluster as input, which it uses to derive which one is the right capi cluster.

I think I might not be fully getting the question 😓

cxbrowne1207 (approved):
/lgtm

pkg/controller/clusters/workers.go
```diff
@@ -0,0 +1,122 @@
...
+		})
+	}
+
+	return w
+}
+
+// ReconcileWorkersForEKSA orchestrates the worker node reconciliation logic for a particular EKS-A cluster.
+// It takes care of applying all desired objects in the Workers spec and deleting the
+// old MachineDeployments that are not in it.
+func ReconcileWorkersForEKSA(ctx context.Context, log logr.Logger, c client.Client, cluster *anywherev1.Cluster, w *Workers) (controller.Result, error) {
```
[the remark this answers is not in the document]
g-gaston: The generic version of `Workers` introduces complexity (bc it uses generics). That complexity has some benefits for spec generation, since it allows to have a shared implementation for all providers that still return concrete api types. However, the code in this method doesn't need to know about concrete provider types and using an interface simplifies it. I would admit I don't like having the same object defined twice, but I find this tradeoff better than dragging the generic complexity over to this package.

I could define the non generic version of Workers in the clusterapi package but I decided to start here and move it there if we see the need to use it somewhere else.

As far as moving this whole method to `clusterapi`, i don't think it's the right place for it. `clusterapi` was supposed to deal with capi object generation and api utils in general. It's not for controller specific code. We could expand the scope, but I don't see any benefit, given that `controller/clusters` already existed and its current scope is exactly that: reusable cluster controller specific code.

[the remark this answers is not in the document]
taneyland: Ok yeah, that's fair! I also don't love the duplication but you made some valid points. Thanks for the in-depth explanation!

tatlat (approved):
Looks good!

pkg/controller/clusters/workers.go
mitalipaygude: I was essentially asking for the difference between `ReconcileWorkers` and `ReconcileWorkersForEKSA`

PR #4138, opened 2022-11-17
Wait until status is refreshed in cache in envtest

g-gaston:
## Description of changes
If we don't wait for the subresource to be updated, there can be race conditions where the status still has the default value when reading from the cache. If the tests rely on the status to be updated to spec, this could cause flakiness.


cxbrowne1207 (approved):
Looks good!

PR #4141, opened 2022-11-18
Reconcile all workers

tatlat:
*Issue #, if available:*

*Description of changes:*
Implementing the workers reconciler for docker

*Testing (if applicable):*


pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -54,6 +63,24 @@ func TestReconcilerReconcileWorkerNodesSuccess(t *testing.T) {
 	tt.Expect(err).NotTo(HaveOccurred())
 	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
 	tt.Expect(result).To(Equal(controller.Result{}))
+
+	tt.ShouldEventuallyExist(tt.ctx,
```
g-gaston, line 67: we should do this for the docker machine template too :)

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -94,6 +121,36 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
 
+func TestReconcilerReconcileWorkersSuccess(t *testing.T) {
+	tt := newReconcilerTest(t)
+	tt.createAllObjs()
+
+	result, err := tt.reconciler().ReconcileWorkers(tt.ctx, test.NewNullLogger(), tt.buildSpec())
+
+	tt.Expect(err).NotTo(HaveOccurred())
+	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
+	tt.Expect(result).To(Equal(controller.Result{}))
```
g-gaston, line 132: can we add here the same expectation as in `TestReconcilerReconcileWorkerNodesSuccess`?

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -94,6 +121,36 @@ func TestReconcileCNIErrorClientRegistry(t *testing.T) {
...
+	})
+	tt.cluster.Spec.KubernetesVersion = ""
+	tt.eksaSupportObjs = append(tt.eksaSupportObjs, capiCluster)
+	tt.createAllObjs()
+
+	logger := test.NewNullLogger()
+
+	result, err := tt.reconciler().ReconcileWorkerNodes(tt.ctx, logger, tt.cluster)
+
+	tt.Expect(err).To(HaveOccurred())
```
g-gaston, line 150: can we assert the content of the error? if you add some context to the error with a wrap, it becomes easier to assert that part

pkg/providers/docker/reconciler/reconciler.go
```diff
@@ -7,7 +7,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
 	"github.com/aws/eks-anywhere/pkg/cluster"
+	c "github.com/aws/eks-anywhere/pkg/cluster"
```
g-gaston, line 12: could we not alias this? if there is a conflict with a `cluster` variable, just rename that variable to `c` (or something else)

pkg/providers/docker/reconciler/reconciler_test.go
```diff
@@ -148,6 +184,7 @@ func TestReconcilerReconcileWorkerNodesFail(t *testing.T) {
 	result, err := tt.reconciler().ReconcileWorkerNodes(tt.ctx, logger, tt.cluster)
 
 	tt.Expect(err).To(HaveOccurred())
```
cxbrowne1207, line 186: We probably don't need this assertion since we have the one after this line?

pkg/providers/docker/reconciler/reconciler_test.go
tatlat: not sure if I did it correctly, but the test now expects to match a substring from the error

tatlat: not sure if I did it correctly, but the test now expects to match a substring from the error

pkg/providers/docker/reconciler/reconciler_test.go
tatlat: I thought it might be needed if err is nil, but it looks like you're right. Removing it doesn't make a difference.

pkg/providers/docker/reconciler/reconciler.go
```diff
@@ -61,8 +63,29 @@ func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, clusterS
 
 // ReconcileWorkerNodes validates the cluster definition and reconciles the worker nodes
 // to the desired state.
-func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
-	return controller.Result{}, nil
+func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger, c *anywherev1.Cluster) (controller.Result, error) {
+	log = log.WithValues("provider", "docker", "reconcile type", "workers")
+	clusterSpec, err := cluster.BuildSpec(ctx, clientutil.NewKubeClient(r.client), c)
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("Failed to construct cluster Spec: %v", err)
```
mitalipaygude, line 70:
```suggestion
		return controller.Result{}, fmt.Errorf("failed to construct cluster Spec: %v", err)
```
The errors can usually get added on to previous error statements.

mitalipaygude:
Looks good!
/lgtm

g-gaston (approved):

PR #3964, opened 2022-11-07
Add workers reconcile for self managed clusters

g-gaston:
## Description of changes
Until now, the new full lifecycle controller completely ignored self managed (management and standalone) clusters. This PR adds partial reconcile support by reacting to changes in the worker part of the spec.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4174, opened 2022-11-21
Remove old check to prevent self-managed cluster upgrades with controller

taneyland:
*Issue #, if available:*

*Description of changes:*
We now allow users to update worker node configs for the self-managed clusters through the controller, so we can remove this check

*Testing (if applicable):*
Unit tests + upgraded self-managed Docker cluster (scaled nodes up/down  + added a new wn group) using `kubectl apply` 

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook_test.go
+++ b/pkg/api/v1alpha1/cluster_webhook_test.go
@@ -1715,7 +1715,7 @@ func TestClusterValidateUpdateInvalidRequest(t *testing.T) {
 	cNew.Spec.ControlPlaneConfiguration.Count = cNew.Spec.ControlPlaneConfiguration.Count + 1
 	g := NewWithT(t)
 	err := cNew.ValidateUpdate(cOld)
-	g.Expect(err).To(MatchError(ContainSubstring("upgrading self managed clusters is not supported")))
+	g.Expect(err).To(MatchError(ContainSubstring("spec.ControlPlaneConfiguration: Forbidden: field is immutable")))
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1715,7 +1715,7 @@ func TestClusterValidateUpdateInvalidRequest(t *testing.T) {
 	cNew.Spec.ControlPlaneConfiguration.Count = cNew.Spec.ControlPlaneConfiguration.Count + 1
 	g := NewWithT(t)
 	err := cNew.ValidateUpdate(cOld)
-	g.Expect(err).To(MatchError(ContainSubstring("upgrading self managed clusters is not supported")))
+	g.Expect(err).To(MatchError(ContainSubstring("spec.ControlPlaneConfiguration: Forbidden: field is immutable")))
```
cxbrowne1207, line 1718: So, if I'm understanding correctly, the worker node configs can be updated through the controller but the ControlPlaneConfiguration is still immutable, is that correct?

taneyland: Yes, exactly! Initially, the new full lifecycle controller completely ignored self managed clusters. But we recently added partial reconcile support by reacting to changes in the worker part of the spec in this [PR](https://github.com/aws/eks-anywhere/pull/3964)

cxbrowne1207: Alright, thanks for clarifying!

cxbrowne1207 (approved):

mitalipaygude (approved):

PR #4180, opened 2022-11-21
Cleanup cluster status failure after validations

g-gaston:


mitalipaygude:
If we error out in validations, in which situation would we continue with cluster creation and have the status contain errors for validations?

Issue #4050, opened 2022-11-14
Snow BottleRocket OS support


jiayiwang7:
After the initial research and POC, we need to make certain accommodations to support BottleRocket (BR) OS as the node image for Snow provider. 

Currently EKS-A uses a BR host container to bootstrap the cluster with `kubeadm`: we pass the BR and cluster settings through user-data, and the host container parses the data and executes `kubeadm` steps to bootstrap the nodes for EKS-A cluster. This host container runs after `kubelet` starts.

Snow requires custom settings on root volume mount, hostname and networking (DNI) to bootstrap the node itself and need to be set up before `kubelet` starts. Thus we need to build and run a standalone BR bootstrap container for Snow before running the existing EKS-A host container. The workflow should be:

1. EKS-A CAPI patch generates the BR format user-data, including the new snow BR bootstrap container setting
2. CAPAS parses the user-data to add snow bootstrap container settings (hostname, etc.) (TBD: may move to snow bootstrap container)
3. CAPAS starts a BR instance with the new user-data
4. CAPAS creates DNIs for the instance
5. CAPAS creates a VNI for the instance 
6. BR instance pulls the snow bootstrap container image, and starts bootstrap container
7. Snow bootstrap container configures network, volume and reboot BR instance
9. BR instance pulls and starts the EKS-A host-container
10. EKS-A host container writes user-data files and runs `kubeadm` operations
11. Once the machine is up running, CAPAS detaches and deletes the VNI

High level checklist:
- [x] Build a standalone Snow BR bootstrap container for node bootstrapping
- [x] Add BR custom bootstrap containers config option in CAPI
- [x] Add BR admin container override image option in CAPI
- [x] Add Snow bootstrap container image to EKS-A bundle
- [x] Add control container image to EKS-A bundle
- [x] Update EKS-A to support new BR Snow configurations

PR #4183, opened 2022-11-21
Support bottlerocket OS for snow provider

jiayiwang7:
*Issue #, if available:*

#4050 

*Description of changes:*

Add bottlerocket support for Snow
- move kube-vip pod manifest to user data
- update proxy, registry mirror config option for BR

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/clusterapi/proxy.go
+++ b/pkg/clusterapi/proxy.go
@@ -14,6 +14,50 @@ import (
 //go:embed config/http-proxy.conf
 var proxyConfig string
 
+func proxy(cluster *v1alpha1.Cluster) bootstrapv1.ProxyConfiguration {
+	return bootstrapv1.ProxyConfiguration{
+		HTTPSProxy: cluster.Spec.ProxyConfiguration.HttpsProxy,
+		NoProxy:    noProxyList(cluster),
+	}
+}
+
+// SetProxyConfigInKubeadmControlPlaneForBottlerocket sets up proxy configuration in kubeadmControlPlane for bottlerocket.
+func SetProxyConfigInKubeadmControlPlaneForBottlerocket(kcp *controlplanev1.KubeadmControlPlane, cluster *v1alpha1.Cluster) {
+	if cluster.Spec.ProxyConfiguration == nil {
+		return
+	}
+
+	kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.Proxy = proxy(cluster)
+	kcp.Spec.KubeadmConfigSpec.JoinConfiguration.Proxy = proxy(cluster)
+}
+
+// SetProxyConfigInKubeadmControlPlaneForUbuntu sets up proxy configuration in kubeadmControlPlane for ubuntu.
+func SetProxyConfigInKubeadmControlPlaneForUbuntu(kcp *controlplanev1.KubeadmControlPlane, cluster *v1alpha1.Cluster) error {
+	if cluster.Spec.ProxyConfiguration == nil {
+		return nil
+	}
+
+	return addProxyConfigInKubeadmConfigSpecFiles(&kcp.Spec.KubeadmConfigSpec, cluster)
+}
+
+// SetProxyConfigInKubeadmConfigTemplateForBottlerocket sets up proxy configuration in kubeadmConfigTemplate for bottlerocket.
+func SetProxyConfigInKubeadmConfigTemplateForBottlerocket(kct *bootstrapv1.KubeadmConfigTemplate, cluster *v1alpha1.Cluster) {
+	if cluster.Spec.ProxyConfiguration == nil {
+		return
+	}
+
+	kct.Spec.Template.Spec.JoinConfiguration.Proxy = proxy(cluster)
+}
+
+// SetProxyConfigInKubeadmConfigTemplateForUbuntu sets up proxy configuration in kubeadmConfigTemplate for ubuntu.
+func SetProxyConfigInKubeadmConfigTemplateForUbuntu(kct *bootstrapv1.KubeadmConfigTemplate, cluster *v1alpha1.Cluster) error {
+	if cluster.Spec.ProxyConfiguration == nil {
+		return nil
+	}
+
+	return addProxyConfigInKubeadmConfigSpecFiles(&kct.Spec.Template.Spec, cluster)
+}
+
 func NoProxyDefaults() []string {
 	return []string{
 		"localhost",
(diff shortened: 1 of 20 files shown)


taneyland:
These changes look good to me & they seem very well tested with the unit tests! I haven't worked closely with registry mirror or proxy configs, so I will let someone else with more context add the label

pkg/clusterapi/proxy.go
```diff
@@ -8,12 +8,64 @@ import (
...
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/templater"
 )
 
 //go:embed config/http-proxy.conf
 var proxyConfig string
 
+func proxy(clusterSpec *cluster.Spec) bootstrapv1.ProxyConfiguration {
```
g-gaston, line 19: can just take only the cluster?

pkg/clusterapi/proxy.go
```diff
@@ -8,12 +8,64 @@ import (
...
+}
+
+// SetProxyConfigInKubeadmControlPlane sets up proxy configuration in kubeadmControlPlane.
+// Currently only support proxy setting for bottlerocket and ubuntu. No-op for any other OS.
+func SetProxyConfigInKubeadmControlPlane(kcp *controlplanev1.KubeadmControlPlane, clusterSpec *cluster.Spec, os v1alpha1.OSFamily) error {
+	if clusterSpec.Cluster.Spec.ProxyConfiguration == nil {
+		return nil
+	}
+
+	switch os {
```
g-gaston, line 33: I would prefer is this check was made at the provider level and we had two methods: SetProxyConfigInKubeadmControlPlaneForBottlerocket` and `SetProxyConfigInKubeadmControlPlaneForUbuntu`.

The os support is still a provider concern IMO, at least from the eks-a perspective.

pkg/clusterapi/proxy.go
```diff
@@ -8,12 +8,64 @@ import (
...
+	case v1alpha1.Bottlerocket:
+		kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.Proxy = proxy(clusterSpec)
+		kcp.Spec.KubeadmConfigSpec.JoinConfiguration.Proxy = proxy(clusterSpec)
+
+	case v1alpha1.Ubuntu:
+		return addProxyConfigInKubeadmConfigSpecFiles(&kcp.Spec.KubeadmConfigSpec, clusterSpec.Cluster.Spec)
+
+	default:
+		logger.MarkWarning(fmt.Sprintf("  Warning: unsupported OS family %s for proxy config. Please use %s or %s", os, v1alpha1.Bottlerocket, v1alpha1.Ubuntu))
```
g-gaston, line 42: can we pass the logger as an input?
It makes code more reusable from the controller.
Also those `MarkWarning` methods only make sense in the CLI. And it proxy is not supported with a particular osFamily I would expect a validation error before we get here.

pkg/clusterapi/proxy.go
```diff
@@ -8,12 +8,64 @@ import (
...
+}
+
+// SetProxyConfigInKubeadmConfigTemplate sets up proxy configuration in kubeadmConfigTemplate.
+// Currently only support proxy setting for bottlerocket and ubuntu. No-op for any other OS.
+func SetProxyConfigInKubeadmConfigTemplate(kct *bootstrapv1.KubeadmConfigTemplate, clusterSpec *cluster.Spec, os v1alpha1.OSFamily) error {
+	if clusterSpec.Cluster.Spec.ProxyConfiguration == nil {
+		return nil
+	}
+
+	switch os {
```
g-gaston, line 55: same comment here about the two methods

g-gaston (commented):
I have one file left, will continue shortly

pkg/providers/snow/apibuilder.go
```diff
@@ -57,16 +61,22 @@ func KubeadmControlPlane(clusterSpec *cluster.Spec, snowMachineTemplate *snowv1.
...
 	}
 
-	if err := clusterapi.SetProxyConfigInKubeadmControlPlane(kcp, clusterSpec.Cluster.Spec); err != nil {
+	if err := clusterapi.SetProxyConfigInKubeadmControlPlane(kcp, clusterSpec, osFamily); err != nil {
 		return nil, err
 	}
 
-	clusterapi.CreateContainerdConfigFileInKubeadmControlPlane(kcp, clusterSpec.Cluster.Spec)
-	clusterapi.RestartContainerdInKubeadmControlPlane(kcp, clusterSpec.Cluster.Spec)
+	if osFamily == v1alpha1.Bottlerocket {
```
g-gaston, line 74: ahh we already have this check here
This is where I would put all the other methods I was suggesting:
`SetRegistryMirrorInKubeadmControlPlaneForBottlerocket`...

pkg/clusterapi/proxy.go
```diff
@@ -14,6 +14,50 @@ import (
...
 
+func proxy(cluster v1alpha1.ClusterSpec) bootstrapv1.ProxyConfiguration {
+	return bootstrapv1.ProxyConfiguration{
+		HTTPSProxy: cluster.ProxyConfiguration.HttpsProxy,
+		NoProxy:    noProxyList(cluster),
+	}
+}
+
+// SetProxyConfigInKubeadmControlPlaneForBottlerocket sets up proxy configuration in kubeadmControlPlane for bottlerocket.
+func SetProxyConfigInKubeadmControlPlaneForBottlerocket(kcp *controlplanev1.KubeadmControlPlane, cluster v1alpha1.ClusterSpec) {
```
g-gaston, line 25: for all this methods I would be in favor of passing the *Cluster instead of the ClusterSpec, for these reasons:
* We never have a ClusterSpec without a Cluster object so it doesn't make it more reusable or easier to read
* ClusterSpec is passed by value, so it copies the whole thing every time you pass it to a method (not the biggest performance hit, but something)
* ClusterSpec and cluster.Spec can be confusing

pkg/providers/snow/apibuilder.go
```diff
@@ -26,12 +27,16 @@ func CAPICluster(clusterSpec *cluster.Spec, snowCluster *snowv1.AWSSnowCluster,
 	return clusterapi.Cluster(clusterSpec, snowCluster, kubeadmControlPlane)
 }
 
-func KubeadmControlPlane(clusterSpec *cluster.Spec, snowMachineTemplate *snowv1.AWSSnowMachineTemplate) (*controlplanev1.KubeadmControlPlane, error) {
+func KubeadmControlPlane(clusterSpec *cluster.Spec, snowMachineTemplate *snowv1.AWSSnowMachineTemplate, log logr.Logger) (*controlplanev1.KubeadmControlPlane, error) {
```
g-gaston, line 30: nit: 
```suggestion
func KubeadmControlPlane(log logr.Logger, clusterSpec *cluster.Spec, snowMachineTemplate *snowv1.AWSSnowMachineTemplate) (*controlplanev1.KubeadmControlPlane, error) {
```
For consistency, we usually pass the logger before the other arguments

pkg/providers/snow/objects.go
```diff
@@ -17,7 +18,7 @@ import (
 	snowv1 "github.com/aws/eks-anywhere/pkg/providers/snow/api/v1beta1"
 )
 
-func ControlPlaneObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]kubernetes.Object, error) {
+func ControlPlaneObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client, log logr.Logger) ([]kubernetes.Object, error) {
```
g-gaston, line 21:
```suggestion
func ControlPlaneObjects(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]kubernetes.Object, error) {
```

pkg/providers/snow/snow.go
```diff
@@ -99,8 +102,8 @@ func (p *SnowProvider) UpdateSecrets(ctx context.Context, cluster *types.Cluster
 	return nil
 }
 
-func CAPIObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) (controlPlaneSpec, workersSpec []byte, err error) {
-	controlPlaneObjs, err := ControlPlaneObjects(ctx, clusterSpec, kubeClient)
+func CAPIObjects(ctx context.Context, clusterSpec *cluster.Spec, kubeClient kubernetes.Client, log logr.Logger) (controlPlaneSpec, workersSpec []byte, err error) {
```
g-gaston, line 105:
```suggestion
func CAPIObjects(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) (controlPlaneSpec, workersSpec []byte, err error) {
```

g-gaston (approved):

PR #4186, opened 2022-11-21
removed feature is active checks for nutanix provider

deepakm-ntnx:
*Issue #, if available:*

*Description of changes:*
This is given as we want to go GA without setting the feature flag

*Testing (if applicable):*
TODO

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -139,10 +139,6 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		return fmt.Errorf("provider snow is not supported in this release")
 	}
 
-	if !features.IsActive(features.NutanixProvider()) && deps.Provider.Name() == constants.NutanixProviderName {
-		return fmt.Errorf("provider nutanix is not supported in this release")
-	}
-
 	createCluster := workflows.NewCreate(
 		deps.Bootstrapper,
 		deps.Provider,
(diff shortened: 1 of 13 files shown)


pokearu:
@deepakm-ntnx Looks like we might need to remove the `"github.com/aws/eks-anywhere/pkg/features"` import from some of the files under `test/e2e`. 
Could we try running `make e2e` it should display the errors locally.

deepakm-ntnx:
> @deepakm-ntnx Looks like we might need to remove the `"github.com/aws/eks-anywhere/pkg/features"` import from some of the files under `test/e2e`. Could we try running `make e2e` it should display the errors locally.

Thanks, this helped. pushed the fix. was looking at the wrong log in prow.

pokearu:
Could we also remove the feature gate function and the variable
https://github.com/aws/eks-anywhere/blob/main/pkg/features/features.go#L59

If we do not plan on using it again?

deepakm-ntnx:
> Could we also remove the feature gate function and the variable https://github.com/aws/eks-anywhere/blob/main/pkg/features/features.go#L59
> 
> If we do not plan on using it again?

would you recommend using the same or a specific feature name based such function for say a feature which we don't want to enable yet.

chrisdoherty4:
> > Could we also remove the feature gate function and the variable [`main`/pkg/features/features.go#L59](https://github.com/aws/eks-anywhere/blob/main/pkg/features/features.go?rgh-link-date=2022-11-23T23%3A41%3A14Z#L59)
> > If we do not plan on using it again?
> 
> would you recommend using the same or a specific feature name based such function for say a feature which we don't want to enable yet.

Each feature should really have its own flag so they can be released independently. Re-using this one could be misleading.

abhinavmpandey08:
/hold
Let's hold this until the official release

abhinavmpandey08 (approved):
looks good! We will merge this once we are closer to Nutanix GA

abhinavmpandey08 (approved):
/approve
/lgtm

PR #4199, opened 2022-11-22
Don't assume env vars when running tests

ewollesen:
Some tests are assuming the values of environment variables when they're run, instead of setting them to known values first. This can cause test failures when running in environments where those variables are already set to different values than expected.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


TerryHowe (approved):
/lgtm

PR #4204, opened 2022-11-23
Package controller specification design

TerryHowe:
Proposed design for cluster specification change to configure package controller installation.

Partially fixes: https://github.com/aws/eks-anywhere-packages/issues/635

### DIFF
--- a/designs/package-spec.md
+++ b/designs/package-spec.md
@@ -4,16 +4,17 @@
 
 **Problem:** currently the ability to configure the package controller is very limited and requires code changes
 
-Code changes currently have to be made to configure the package controller during cluster creation or using the install command. Changes need to be coordinated between the CLI and the package controller helm chart to alter configuration. This makes it more difficult to support things like various private registry configurations.
+The preference in the CLI is to validate the values passed to the package controller, so the EKS Anywhere cluster specification will include all the necessary configuration values for the controller.
 
 ### Tenets
 
-* ***Flexible:*** reduce dependency of the CLI on helm chart.
+* ***Validation:*** allow the customer the flexibility to configure the package controller and validate those values.
 
 ### Goals and Objectives
 
 As an EKS-A cluster administrator I want to configure the package controller for:
 
+* Allow packages only to be configured from a private registry
 * Harbor using proxy cache
 * Harbor using replication
 * Harbor as a private registry using the import images command
@@ -40,7 +41,7 @@ Add a section to the cluster specification to enable/disable the package control
 
 ## Solution Details
 
-Add a packageController section to the cluster specification for example to use development/staging builds:
+A sample package controller section in the cluster specification. The values specified here:
 ```yaml
 apiVersion: anywhere.eks.amazonaws.com/v1alpha1
 kind: Cluster
@@ -49,29 +50,49 @@ metadata:
 spec:
   packageController:
     enable: true
-    values: |-
-      defaultRegistry: public.ecr.aws/l0g8r8j6
-      defaultImageRegistry: 857151390494.dkr.ecr.us-west-2.amazonaws.com
+    # -- defaultRegistry for all package helm charts.
+    defaultRegistry: public.ecr.aws/eks-anywhere
+    # -- defaultImageRegistry for all package images.
+    defaultImageRegistry: 783794618700.dkr.ecr.us-west-2.amazonaws.com
+    controller:
+      # -- Controller repository name.
+      repository: "eks-anywhere-packages"
+      # -- Controller image tag
+      tag: "{{eks-anywhere-packages-tag}}"
+      # -- Controller image digest
+      digest: "{{eks-anywhere-packages}}"
+      # -- Whether to turn on Webhooks for the controller image
+      enableWebhooks: "true"
+      # -- Additional environment variables for the controller pod.
+      # - name: EKSA_PUBLIC_KEY
+      #   value: ""
+      env:
+        - name: EKSA_PUBLIC_KEY
+          value: "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEnP0Yo+ZxzPUEfohcG3bbJ8987UT4f0tj+XVBjS/s35wkfjrxTKrVZQpz3ta3zi5ZlgXzd7a20B1U1Py/TtPsxw=="
+      resources:
+        requests:
+          cpu: 100m
+          memory: 50Mi
+        limits:
+          cpu: 750m
+          memory: 450Mi
+    # Additional Variables to configure cronjob
+    cronjob:
+      name: ecr-refresher
+      # -- ECR refresher repository name.
+      repository: "ecr-token-refresher"
+      # -- ECR refresher tag
+      tag: "{{ecr-token-refresher-tag}}"
+      # -- ECR refresher digest
+      digest: "{{ecr-token-refresher}}"
+      suspend: false
+    registryMirrorSecret:
+      endpoint: ""
+      cacertcontent: ""
+      insecure: "ZmFsc2UK"
 ```
 
-These values will be passed to the helm chart during installation just like `--set defaultRegistry=public.ecr.aws/l0g8r8j6` for example.
-
-## Implementation
-
-The proposed design:
-```
-// PackageController defines configuration of the package controller.
-type PackageController struct {
-    // +kubebuilder:validation:Optional
-    // +kubebuilder:default:="true"
-    // Enable installation of the package controller.
-    Enable bool `json:"enable,omitempty"`
-
-    // +kubebuilder:validation:Optional
-    // Values to pass into package controller installation.
-    Values string `json:"values,omitempty"`
-}
-```
+These values will be passed to the helm chart during installation using a values.yaml file to protect the secrets.
 
 ## Testing
 
--- a/designs/package-spec.md
+++ b/designs/package-spec.md
@@ -48,7 +48,7 @@ metadata:
   name: billy
 spec:
   packageController:
-    enable: true
+    disable: false
     # -- defaultRegistry for all package helm charts.
     defaultRegistry: public.ecr.aws/eks-anywhere
     # -- defaultImageRegistry for all package images.
--- a/designs/package-spec.md
+++ b/designs/package-spec.md
@@ -1,0 +1,78 @@
+# Package Controller Configuration
+
+## Introduction
+
+**Problem:** currently the ability to configure the package controller is very limited and requires code changes
+
+Code changes currently have to be made to configure the package controller during cluster creation or using the install command. Changes need to be coordinated between the CLI and the package controller helm chart to alter configuration. This makes it more difficult to support things like various private registry configurations.
+
+### Tenets
+
+* ***Flexible:*** reduce dependency of the CLI on helm chart.
+
+### Goals and Objectives
+
+As an EKS-A cluster administrator I want to configure the package controller for:
+
+* Harbor using proxy cache
+* Harbor using replication
+* Harbor as a private registry using the import images command
+* Configure ECR token refresher
+* Configure for kubelet private registry authentication
+* Support proxy configuration
+* Support package controller installation from alternate registries
+* Disable installation of package controller
+* Allow for future configuration options
+
+### Statement of Scope
+
+**In scope**
+* All of the above
+
+## Current state
+The package controller can be used currently with:
+1. Proxy configurations
+1. ECR token refresher enabled/disabled
+1. Installation of the package controller from alternate registries
+
+## Overview of Solution
+Add a section to the cluster specification to enable/disable the package controller installation and pass configuration values.
+
+## Solution Details
+
+Add a packageController section to the cluster specification for example to use development/staging builds:
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: billy
+spec:
+  packageController:
+    enable: true
+    values: |-
+      defaultRegistry: public.ecr.aws/l0g8r8j6
+      defaultImageRegistry: 857151390494.dkr.ecr.us-west-2.amazonaws.com
+```
+
+These values will be passed to the helm chart during installation just like `--set defaultRegistry=public.ecr.aws/l0g8r8j6` for example.
+
+## Implementation
+
+The proposed design:
+```
+// PackageController defines configuration of the package controller.
+type PackageController struct {
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:default:="true"
+    // Enable installation of the package controller.
+    Enable bool `json:"enable,omitempty"`
+
+    // +kubebuilder:validation:Optional
+    // Values to pass into package controller installation.
+    Values string `json:"values,omitempty"`
+}
+```
+
+## Testing
+
+We should add at least one E2E to test disabled package controller, configuration of values will be used to test staging builds.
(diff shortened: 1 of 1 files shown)


designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: billy
+spec:
+  packageController:
+    enable: true
+    values: |-
+      defaultRegistry: public.ecr.aws/l0g8r8j6
```
ivyostosh, line 53:
```suggestion
      defaultRegistry: public.ecr.aws/eks-anywhere
```

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+      defaultImageRegistry: 857151390494.dkr.ecr.us-west-2.amazonaws.com
```
ivyostosh, line 54:
```suggestion
      defaultImageRegistry: 783794618700.dkr.ecr.us-west-2.amazonaws.com
```

designs/package-spec.md
ivyostosh: Although this is just an example in the design doc, considering the doc is public on Github, we should stay away from referencing the l0g8r8j6 account.

designs/package-spec.md
ivyostosh: Same comment here.

designs/package-spec.md
TerryHowe: * The values you suggest *are* the default values for these two settings.
* This example shows how to configure the controller for staging/dev env
* While I wouldn't recommend customers use these repos, no harm in people knowing about it. It is referenced all over already.

ivyostosh: Fair point on what we want to illustrate here. I will vote for a dummy value like what you used for metadata.name than a real account, but I can also be ok with this set up as what it is.

a-cool-train:
/hold
Might be a good idea to get a review from cli team.

chrisdoherty4:
If I understand correctly, you just want to be able to specify values YAML overrides via the EKS-A API specifically for the package controller?

Assuming that's accurate:

1. Is there any existing configuration specified elsewhere in the EKS-A API that results in a value override? If so, is it possible to harmonize and how do we intend on harmonizing that (i.e. is there a deprecation roadmap)? If its not possible to harmonize, what does the configuration landscape look like with seemingly splintered configuration given it might be possible to also configure certain bits via values override?
2. What does finding out what values I can specify look like? What workflow do I need to follow?
3. Would this ever be needed for individual packages (I suspect this is already handled)?

chrisdoherty4 (commented):
If I understand correctly, you just want to be able to specify values YAML overrides via the EKS-A API specifically for the package controller?

Assuming that's accurate:

1. Is there any existing configuration specified elsewhere in the EKS-A API that results in a value override? If so, is it possible to harmonize and how do we intend on harmonizing that (i.e. is there a deprecation roadmap)? If its not possible to harmonize, what does the configuration landscape look like with seemingly splintered configuration given it might be possible to also configure certain bits via values override?
2. What does finding out what values I can specify look like? What workflow do I need to follow?
3. Would this ever be needed for individual packages (I suspect this is already handled)?

TerryHowe:
> 3. Would this ever be needed for individual packages (I suspect this is already handled)?

We use this approach for packages, a yaml string in yaml, but for packages we validate the string with the schema.json  from the helm chart in the webhook. We could in theory validate these values that way here, but not sure if it is worth it. This will mostly be used for development and test.

I was thinking we'd document a very small subset of all the possible values.

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+Add a packageController section to the cluster specification for example to use development/staging builds:
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: billy
+spec:
+  packageController:
+    enable: true
+    values: |-
```
g-gaston, line 52: I vote to structurize this field
I would start by adding just the fields we know users need right now and add more as need arises

That would provide better validation out of the box, reduces the testing surface and makes sure we expose only the config that makes sense for eks-a.

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+## Implementation
+
+The proposed design:
+```
+// PackageController defines configuration of the package controller.
+type PackageController struct {
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:default:="true"
+    // Enable installation of the package controller.
+    Enable bool `json:"enable,omitempty"`
```
g-gaston, line 68: We can leave this discussion for the implementation PR, but since you added the go struct here: this is going to be difficult to default to true. After parsing the yaml file, if not set, this field will be false. But we won't know if it wasn't set (hence, should be defaulted to to true) or it was explicitly set to false.

One alternative would be to have it as `*bool`, although that makes accessing it in the code a bit annoying. Given this is probably going to be read only once in the codepath, it's probably an ok tradeoff.

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+## Solution Details
+
+Add a packageController section to the cluster specification for example to use development/staging builds:
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+metadata:
+  name: billy
+spec:
+  packageController:
```
g-gaston, line 50: We probably need a validation to make sure this is only set for management clusters? Or would we allow to install it in workload clusters as well?

If it's allowed in workload clusters, we would need to document what happens when this is specified directly through the API, since we don't have a "packages controller operator" that would install/update the packages controller.

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
+# Package Controller Configuration
+
+## Introduction
+
+**Problem:** currently the ability to configure the package controller is very limited and requires code changes
+
+Code changes currently have to be made to configure the package controller during cluster creation or using the install command. Changes need to be coordinated between the CLI and the package controller helm chart to alter configuration. This makes it more difficult to support things like various private registry configurations.
```
g-gaston, line 7: Could you explain a bit more what are the "code changes" and the coordination between CLI and helm chart needed?

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Harbor using proxy cache
+* Harbor using replication
+* Harbor as a private registry using the import images command
+* Configure ECR token refresher
+* Configure for kubelet private registry authentication
+* Support proxy configuration
+* Support package controller installation from alternate registries
+* Disable installation of package controller
+* Allow for future configuration options
```
g-gaston, line 25: nit: this sounds more like a requirement from a eks-a developer?

designs/package-spec.md
TerryHowe: Currently, when the CLI installs the package controller, we have to change code to pass those values to the helm install command(e.g.: `--set defaultRegistry=public.ecr.aws/213456789`). If we change the names of the values or add ones we have to change the CLI.

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+### Statement of Scope
+
+**In scope**
+* All of the above
+
+## Current state
+The package controller can be used currently with:
+1. Proxy configurations
+1. ECR token refresher enabled/disabled
+1. Installation of the package controller from alternate registries
```
jonahjon, line 36: Does this work for alternate ECR registries? I thought the token refresher only worked with 857 and 7837.

designs/package-spec.md
TerryHowe: Removed

designs/package-spec.md
TerryHowe: Removed

designs/package-spec.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Overview of Solution
+Add a section to the cluster specification to enable/disable the package controller installation and pass configuration values.
+
+## Solution Details
+
+Add a packageController section to the cluster specification for example to use development/staging builds:
```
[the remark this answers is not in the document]
TerryHowe: I would hope all these values would be optional and default to not set. This is for power users.

As for registry mirror configuration, I think it makes sense to possibly pull packages from a registry mirror while EKS-A is sourced from another mirror or ECR. I think if a registry mirror is set here, it would override the mirror from EKS-A spec.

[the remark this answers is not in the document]
TerryHowe: As to preflight, including these values in the cluster spec will do some validation preflight.

TerryHowe:
> Is there any existing configuration specified elsewhere in the EKS-A API that results in a value override?

You guys could answer this question better than I could. I look at this configuration more like overriding the defaults than overriding other values in the spec. How this would work as far as things that actually get overridden:

1. If there is no registry mirror, ECR
2. If there is a registry mirror use that
3. If defaultImageRegistry or defaultRegistry is set, use that

designs/package-spec.md
TerryHowe: The CLI only works on those accounts as well. That is all that is supported right now.

TerryHowe:
> If that's the case, do we need something on the public API? As much as I hate hard to discover environment configuration, would it make sense as a hidden environment variable?

There are a lot of other things a customer might want to set here for the package controller, so environment variables will not cut it. The registries are just what *I* want to use it for.

designs/package-spec.md
TerryHowe: This example is for testing, but basically most fields in our values.yaml, we'd want to have here. I've added them.

designs/package-spec.md
TerryHowe: These values potentially will be used for workload clusters as well. We currently just create some custom resources and secrets for workload clusters, but Vincent is working on a helm chart for them that would use this.

designs/package-spec.md
TerryHowe: I removed that.

designs/package-spec.md
TerryHowe: Done.

designs/package-spec.md
```diff
@@ -0,0 +1,98 @@
...
+      repository: "ecr-token-refresher"
+      # -- ECR refresher tag
+      tag: "{{ecr-token-refresher-tag}}"
+      # -- ECR refresher digest
+      digest: "{{ecr-token-refresher}}"
+      suspend: false
+    registryMirrorSecret:
+      endpoint: ""
+      cacertcontent: ""
+      insecure: "ZmFsc2UK"
```
g-gaston, line 91: Is this supposed to be a secret?

designs/package-spec.md
```diff
@@ -0,0 +1,98 @@
...
+kind: Cluster
+metadata:
+  name: billy
+spec:
+  packageController:
+    disable: false
+    # -- defaultRegistry for all package helm charts.
+    defaultRegistry: public.ecr.aws/eks-anywhere
+    # -- defaultImageRegistry for all package images.
+    defaultImageRegistry: 783794618700.dkr.ecr.us-west-2.amazonaws.com
```
g-gaston, line 55: Just for my understanding, what would be the difference between using these two and using the cluster top level registry mirror?

designs/package-spec.md
```diff
@@ -0,0 +1,98 @@
...
+    cronjob:
+      name: ecr-refresher
+      # -- ECR refresher repository name.
+      repository: "ecr-token-refresher"
+      # -- ECR refresher tag
+      tag: "{{ecr-token-refresher-tag}}"
+      # -- ECR refresher digest
+      digest: "{{ecr-token-refresher}}"
+      suspend: false
+    registryMirrorSecret:
```
g-gaston, line 88: what is this for? isn't this equivalent to the top level registry mirror?

designs/package-spec.md
g-gaston: Ah ok, that's where my packages knowledge falls short. Does that mean we are going to allow users to install the package controller in workload clusters as well?

I'm just trying to figure out what should be the behavior of this new field for workload clusters created through the controller and not through the CLI.

g-gaston (commented):
Looks good, just minor questions

designs/package-spec.md
TerryHowe: This would override the settings of the top level registry mirror if the user wanted to pull packages from a different mirror or *only* wanted packages from a mirror.

designs/package-spec.md
TerryHowe: The package controller will not be installed on workload clusters. A helm install will be done on the management cluster for each workload cluster so we can manage those configuration resources easier.

designs/package-spec.md
TerryHowe: We store these values in a secret with username and password. The username and password will come from the environment variables and the rest will come from here.

designs/package-spec.md
TerryHowe: This will allow users to pull *just* packages from a registry mirror.

g-gaston: Closing this since I asked the same question twice

designs/package-spec.md
g-gaston: Ahh I see. I thought @vincentni's most recent changes to the top level registry mirror already allowed to do that (having separate registry mirrors for packages/eks-a or only one for packages).

Would this replace that functionality? Or am I missing something and it does something different?

designs/package-spec.md
g-gaston: Sorry, the question wasn't clear. What is `insecure` supposed to represent? is it just a flag like "skip tls verification"? or is it some kind of authentication token?

TerryHowe: This is just for insecure TLS on the registry mirror

designs/package-spec.md
g-gaston: Ah cool. So since this filed configures the package controller and the package controller is never installed in workload clusters, should we add a validation that checks that is always disabled and no extra configuration is provided for workload clusters?

designs/package-spec.md
TerryHowe: I'm not sure what change you are talking about, but as far as I know nothing else allows separate registries for EKS-A and packages now. Vincent has/had a PR out to have different registries for management vs. workload clusters recently. That is different.

designs/package-spec.md
TerryHowe: This configuration is used to configure the workload cluster *on* the management cluster for a workload cluster.

g-gaston: Ok

Are all fields relevant for both management and workload cluster? When I look at the `controller` subfield, that seems like it only affects the controller deployment running in the management cluster?

What should the eks-a cluster controller do when a workload cluster is created through the API? does it need to install/update a helm chart?

designs/package-spec.md
TerryHowe: PR #4128 allows the user to map our registry to their registry, but this allows the user to specify a registry mirror for packages when they are NOT using a registry mirror for EKS-A

TerryHowe: okay, vincent tried it out and he was able to use that feature to replicate this behavior I'm looking for

TerryHowe: okay, removed the default registry settings

designs/package-spec.md
TerryHowe: Yes, the controller section would only be relevant for a management cluster.

The cluster controller takes the same action for management and workload clusters, it runs the same helm install.

PR #4207, opened 2022-11-23
Enable full cluster lifecycle controller support

g-gaston:
## Description of changes
The controller manager flags will be dynamically set based on the cluster infra provider and the env var feature flags. This should be enventually moved to use CRDs for manager config, but that will require to update the manager setup code as well.

## Testing
Unit tests and tested manually with a docker cluster


mitalipaygude:
nit: The commit message has a typo in lifecycle.

pkg/clustermanager/cluster_manager.go
```diff
@@ -934,19 +939,19 @@ func (c *ClusterManager) removeOldWorkerNodeGroups(ctx context.Context, workload
...
-			return err
-		}
-	}
+
+	// TODO(g-gaston): should this be moved inside the components installer?
 	return provider.InstallCustomProviderComponents(ctx, cluster.KubeconfigFile)
 }
 
+// Upgrade updates the eksa components in a cluster according to a Spec.
+func (c *ClusterManager) Upgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error) {
```
vivek-koppuru, line 951: Should this be called `UpgradeCustomComponents`? I see that you removed `upgrader.go`, so if this is replacing that, can this also have other logic in the future if we have other upgrades we need to do within ClusterManager?

pkg/clustermanager/eksa_installer.go
```diff
@@ -0,0 +1,217 @@
...
+	changeDiff := EksaChangeDiff(currentSpec, newSpec)
+	if changeDiff == nil {
+		log.V(1).Info("Nothing to upgrade for controller and CRDs")
+		return nil, nil
+	}
+	log.V(1).Info("Starting EKS-A components upgrade")
+	oldVersion := currentSpec.VersionsBundle.Eksa.Version
+	newVersion := newSpec.VersionsBundle.Eksa.Version
+	if err := i.Install(ctx, log, cluster, newSpec); err != nil {
+		return nil, fmt.Errorf("failed upgrading EKS-A components from version %v to version %v: %v", oldVersion, newVersion, err)
```
vivek-koppuru, line 80:
```suggestion
		return nil, fmt.Errorf("upgrading EKS-A components from version %v to version %v: %v", oldVersion, newVersion, err)
```

pkg/clustermanager/cluster_manager.go
g-gaston: It probably should yeah
However, this satisfies an interface defined in the workflows, so I kept the name to avoid extending changes elsewhere.
Ideally this operation should go directly to the installer from the workflow and not through cluster manager, but again, that expanded the changes a bit too far.

vivek-koppuru (approved):

PR #4208, opened 2022-11-23
Add Bottlerocket control container image in bundle

jiayiwang7:
*Issue #, if available:*

#4050 

*Description of changes:*

The BR AWS Variant used by Snow will run the control container during instance boot by default. The default control container is stored in private ECR registry which is only accessible with credentials and internet. This is not an option for Snow air-gap users. We already have the control container image override option in CAPI patch and now we just need to add the public ECR repo in our bundle to override the default private image.

Change the version bundle BR API by moving `bottlerocketBootstrap` to `bottlerocketHostContainers`
```
bottlerocketHostContainers:
  admin: ...
  control: ...
  kubeadmBootstrap (this is the former "bottlerocketBootstrap"): ...
```
*Testing (if applicable):*

Unit tests


abhay-krishna (approved):
/approve

PR #4212, opened 2022-11-23
Add snow specific Bottlerocket settings: admin, control, bootstrap snow containers

jiayiwang7:
*Issue #, if available:*

#4050 

Base on #4208 

*Description of changes:*

- Override default BR admin, control container images URL
- Add bottlerocket-bootstrap-snow to custom bootstrap container image

Hold till #4208 is merged first.

*Testing (if applicable):*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -209,5 +209,5 @@ replace (
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2
 
 	// need the modifications eksa made to the capi api structs
-	sigs.k8s.io/cluster-api => github.com/abhay-krishna/cluster-api v1.2.0-custom.rc2
+	sigs.k8s.io/cluster-api => github.com/abhay-krishna/cluster-api v1.2.0-custom.rc3
 )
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -209,5 +209,5 @@ replace (
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2
 
 	// need the modifications eksa made to the capi api structs
-	sigs.k8s.io/cluster-api => github.com/abhay-krishna/cluster-api v1.2.0-custom.rc2
+	sigs.k8s.io/cluster-api => github.com/jiayiwang7/cluster-api v0.0.0-20221123142628-4b8c7b3f61eb
```
jiayiwang7, line 212: discard this change when https://github.com/abhay-krishna/cluster-api/pull/3 is merged

config/crd/bases/anywhere.eks.amazonaws.com_bundles.yaml
```diff
@@ -216,7 +216,7 @@ spec:
                       - metadata
                       - version
                       type: object
-                    bottlerocketAdmin:
+                    bottlerocket:
```
vivek-koppuru, line 219: If we upgrade an older cluster to a newer one with these changes and the older bundle is referenced, would this be a problem? Do we need to phase it out instead?

pkg/providers/snow/apibuilder.go
```diff
@@ -235,3 +242,24 @@ func SnowMachineTemplate(name string, machineConfig *v1alpha1.SnowMachineConfig)
 		},
 	}
 }
+
+func bottlerocketBootstrapSnow(image releasev1.Image) bootstrapv1.BottlerocketBootstrapContainer {
```
vivek-koppuru, line 246: Why is this here and not in the clusterapi package? Do we want a bootstrap file for this instead?

jiayiwang7: bc it's a snow specific bootstrap container including the settings like name and mode. I can move it to another file

config/crd/bases/anywhere.eks.amazonaws.com_bundles.yaml
jiayiwang7: I don't think it would be an issue: we have the core components - eksa components upgrade in the upgrade workflow. This will update the eksa CRD (including the bundle CRD) and apply the new bundle spec, before the actual cluster upgrade.

vivek-koppuru (approved):

PR #4213, opened 2022-11-23
Validate control plane IP uniqueness in cluster webhook

taneyland:
*Issue #, if available:*

*Description of changes:*
Validate control plane IP uniqueness at the API level so that we can validate that when creating clusters with the new API 

*Testing (if applicable):*


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -65,6 +66,10 @@ func (r *Cluster) ValidateCreate() error {
 		}
 	}
 
+	if err := ValidateControlPlaneIPUniqueness(r, &networkutils.DefaultNetClient{}); err != nil {
```
chrisdoherty4, line 69: This will result in at least a 2.5s lag period for every submission; that's rather significant. Are we content with that? (Webhooks default timeout is 10s).

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -118,6 +123,18 @@ func validateBundlesRefCluster(new, old *Cluster) field.ErrorList {
 	return allErrs
 }
 
+// ValidateControlPlaneIPUniqueness checks whether or not the control plane endpoint defined
+// in the cluster spec is available.
+func ValidateControlPlaneIPUniqueness(cluster *Cluster, netClient networkutils.NetClient) error {
```
chrisdoherty4, line 128: Does this need to be exported?

chrisdoherty4: It looks like it does which makes me think this is the wrong package to contain this function. It could reside in a `validate` package (or whatever already exists with this little funcs if we have one).

pkg/api/v1alpha1/cluster_webhook.go
taneyland: Hmm I'm not sure of how we can avoid that, but I'm open to ideas for sure.

Without having the validation here, creating a cluster using the new API with a control plane IP that's already in use will result in it continuing on with cluster creation and getting stuck instead of validating beforehand that the ip wont be assigned

pkg/api/v1alpha1/cluster_webhook.go
chrisdoherty4: > I think we might run into an import cycle since the first parameter requires the v1alpha1 package. Because if we move it elsewhere, we would still need to call that method here in the v1alpha1 package since we want to validate things at the API level for the new controller.

Good catch, that's right. 

My gripe is with the snow package calling a validate function in the API package - that just seems odd to me but I suppose framed the right way it could make sense. Thanks for humoring me.

mitalipaygude:
Were we able to decide if this validation needs to be in the webhook or the reconciler?

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -75,6 +80,19 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
 	).Run(ctx, log, clusterSpec)
 }
 
+// ValidateControlPlaneIP validates that the control plane IP is unique.
+func (r *Reconciler) ValidateControlPlaneIP(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "validateControlPlaneIP")
+
+	if err := r.ipValidator.ValidateControlPlaneIPUniqueness(clusterSpec.Cluster.Spec); err != nil {
+		failureMessage := fmt.Sprintf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, please provide a unique IP", clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host)
```
g-gaston, line 88: why do we need to generate an error message, can't we use the returned error?

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -31,14 +32,17 @@ type Reconciler struct {
 	client               client.Client
 	cniReconciler        CNIReconciler
 	remoteClientRegistry RemoteClientRegistry
+	ipValidator          validator.IPValidator
```
g-gaston, line 35: can we define this interface here, where we consume it?

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1483,7 +1483,7 @@ func TestClusterValidateUpdateInvalidWorkloadCluster(t *testing.T) {
 			// Invalid control plane configuration
 			tt.clusterNew.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{
 				Endpoint: &v1alpha1.Endpoint{
-					Host: "1.1.1.1",
+					Host: "1.2.3.4",
```
g-gaston, line 1486: is this needed?

pkg/providers/validator/validate.go
```diff
@@ -10,9 +10,24 @@ import (
...
+}
+
+// NewIPValidator initializes a new IPValidator object.
+func NewIPValidator() IPValidator {
+	return DefaultIPValidator{}
+}
+
+// ValidateControlPlaneIPUniqueness checks whether or not the control plane endpoint defined
+// in the cluster spec is available.
+func (v DefaultIPValidator) ValidateControlPlaneIPUniqueness(spec v1alpha1.ClusterSpec) error {
```
g-gaston, line 28: can we pass the cluster instead if the spec?

pkg/providers/validator/validate.go
```diff
@@ -10,9 +10,24 @@ import (
...
+	netClient := &networkutils.DefaultNetClient{}
```
g-gaston, line 29: This makes this method untestable. I think the network client should be a dependency injected on construction or at least a dependency that can updated from the tests, even if you default to the DefaultNetClient on construction

pkg/providers/validator/validate.go
```diff
@@ -10,9 +10,24 @@ import (
...
+	ip := spec.ControlPlaneConfiguration.Endpoint.Host
 	if networkutils.IsIPInUse(netClient, ip) {
 		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, please provide a unique IP", ip)
```
g-gaston, line 32:
```suggestion
		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, control plane IP must be unique", ip)
```

pkg/providers/validator/validate.go
```diff
@@ -10,9 +10,24 @@ import (
...
+// DefaultIPValidator defines the struct for the default IPValidator.
+type DefaultIPValidator struct{}
+
+// IPValidator defines an interface for the methods to validate the control plane IP.
+type IPValidator interface {
+	ValidateControlPlaneIPUniqueness(cluster v1alpha1.ClusterSpec) error
+}
+
+// NewIPValidator initializes a new IPValidator object.
+func NewIPValidator() IPValidator {
```
g-gaston, line 22: I think this should return the struct

pkg/providers/validator/validate_test.go
```diff
@@ -6,35 +6,13 @@ import (
...
 
-	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/features"
-	"github.com/aws/eks-anywhere/pkg/networkutils/mocks"
 	mockprovider "github.com/aws/eks-anywhere/pkg/providers/mocks"
 	"github.com/aws/eks-anywhere/pkg/providers/validator"
 )
 
-func TestValidateControlPlaneIpUniqueness(t *testing.T) {
```
g-gaston, line 19: why are we deleting a test? or are we replacing this with something else?

pkg/providers/vsphere/reconciler/reconciler.go
```diff
@@ -37,17 +38,19 @@ type Reconciler struct {
 	defaulter            *vsphere.Defaulter
 	cniReconciler        CNIReconciler
 	remoteClientRegistry RemoteClientRegistry
+	ipValidator          validator.IPValidator
```
g-gaston, line 41: same comment here, I would define IPValidator interface here

pkg/providers/vsphere/reconciler/reconciler.go
```diff
@@ -122,6 +126,19 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
 	).Run(ctx, log, clusterSpec)
 }
 
+// ValidateControlPlaneIP validates that the control plane IP is unique.
+func (r *Reconciler) ValidateControlPlaneIP(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "validateControlPlaneIP")
+
+	if err := r.ipValidator.ValidateControlPlaneIPUniqueness(clusterSpec.Cluster.Spec); err != nil {
+		failureMessage := fmt.Sprintf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, please provide a unique IP", clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host)
```
g-gaston, line 134: same comment about reusing the error message from `err`

ddjjia:
Shall we add another test scenario to vsphere and docker after merge?

taneyland:
> Shall we add another test scenario to vsphere and docker after merge?

We have test scenarios for snow & vsphere, but if there are any gaps, we can definitely add it in a follow-up. & the Docker provider doesnt use control plane IPs, so we're good there

pkg/providers/validator/validate.go
```diff
@@ -10,11 +10,43 @@ import (
...
+func (v *DefaultIPValidator) ValidateControlPlaneIPUniqueness(cluster *v1alpha1.Cluster, opts ...IPValidatorOpt) error {
+	for _, opt := range opts {
+		opt(v)
+	}
+
 	ip := cluster.Spec.ControlPlaneConfiguration.Endpoint.Host
-	if networkutils.IsIPInUse(netClient, ip) {
-		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, please provide a unique IP", ip)
+	if networkutils.IsIPInUse(v.netClient, ip) {
+		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, control plane IP must be uniqu", ip)
```
mitalipaygude, line 49:
```suggestion
		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, control plane IP must be unique ", ip)
```

pkg/providers/validator/validate.go
```diff
@@ -10,11 +10,43 @@ import (
...
+// NewIPValidator initializes a new IPValidator object.
+func NewIPValidator() *DefaultIPValidator {
+	return &DefaultIPValidator{
+		netClient: &networkutils.DefaultNetClient{},
+	}
+}
+
+// ValidateControlPlaneIPUniqueness checks whether or not the control plane endpoint defined
+// in the cluster spec is available.
+func (v *DefaultIPValidator) ValidateControlPlaneIPUniqueness(cluster *v1alpha1.Cluster, opts ...IPValidatorOpt) error {
```
g-gaston, line 42: I would suggest passing these options to the constructor, not to this method.
It keeps the interface cleaner, since it won't depend on types from the package. And it's also thread safe since you can't modify `DefaultIPValidator` once it has been constructed.

pkg/providers/validator/validate.go
```diff
@@ -10,11 +10,43 @@ import (
...
+	for _, opt := range opts {
+		opt(v)
+	}
+
 	ip := cluster.Spec.ControlPlaneConfiguration.Endpoint.Host
-	if networkutils.IsIPInUse(netClient, ip) {
-		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, please provide a unique IP", ip)
+	if networkutils.IsIPInUse(v.netClient, ip) {
+		return fmt.Errorf("cluster controlPlaneConfiguration.Endpoint.Host <%s> is already in use, control plane IP must be unique", ip)
```
g-gaston, line 49: nit:
I would use `errors.Errorf` to get the stack

pkg/providers/validator/validate.go
```diff
@@ -10,11 +10,43 @@ import (
 	"github.com/aws/eks-anywhere/pkg/providers"
 )
 
-// TODO: update vsphere, cloudstack, tinkerbell validators to use this common validation method.
-func ValidateControlPlaneIpUniqueness(cluster *v1alpha1.Cluster, netClient networkutils.NetClient) error {
+// DefaultIPValidator defines the struct for the default IPValidator.
+type DefaultIPValidator struct {
```
g-gaston, line 14:
```suggestion
type IPValidator struct {
```

pkg/providers/vsphere/reconciler/reconciler.go
```diff
@@ -122,6 +132,18 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
 	).Run(ctx, log, clusterSpec)
 }
 
+// ValidateControlPlaneIP validates that the control plane IP is unique.
+func (r *Reconciler) ValidateControlPlaneIP(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "validateControlPlaneIP")
+
+	if err := r.ipValidator.ValidateControlPlaneIPUniqueness(clusterSpec.Cluster); err != nil {
+		clusterSpec.Cluster.Status.FailureMessage = ptr.String(err.Error())
+		log.Error(errors.New(*clusterSpec.Cluster.Status.FailureMessage), "Unavailable control plane IP")
```
g-gaston, line 141:
```suggestion
		log.Error(err, "Unavailable control plane IP")
```

g-gaston (approved):
/lgtm

PR #4217, opened 2022-11-28
Remove labels for worker node groups

taneyland:
*Issue #, if available:*

*Description of changes:*
We need to directly compare the labels maps to ensure a new name for the immutable kubeadm config template object when removing worker node labels

*Testing (if applicable):*
Unit test. Upgrade with new API on Docker cluster

### DIFF
--- a/pkg/providers/docker/workers_test.go
+++ b/pkg/providers/docker/workers_test.go
@@ -131,6 +131,69 @@ func TestWorkersSpecUpgradeCluster(t *testing.T) {
 	g.Expect(workers.Groups).To(ConsistOf(*expectedGroup1, *expectedGroup2))
 }
 
+func TestWorkersSpecUpgradeClusterRemoveLabels(t *testing.T) {
+	g := NewWithT(t)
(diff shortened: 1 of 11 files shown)


pkg/providers/docker/workers_test.go
```diff
@@ -131,6 +131,69 @@ func TestWorkersSpecUpgradeCluster(t *testing.T) {
 	g.Expect(workers.Groups).To(ConsistOf(*expectedGroup1, *expectedGroup2))
 }
 
+func TestWorkersSpecUpgradeClusterRemoveLabels(t *testing.T) {
```
g-gaston, line 134: can we move this method to `clusterapi/workers_test.go`?

taneyland: I already added a test case in `clusterapi/workers_test.go` to make sure that the kubeadmConfigTemplate name changes if the labels are removed. Do you think there should be another test in that file?

mitalipaygude (approved):

PR #4234, opened 2022-11-30
Fix delete workload clusters from CLI

g-gaston:
## Description of changes
The new full lifecycle controller adds a finalizer to eks-a Clusters to be able to orchestrate its deletion. This interferes with the delete CLI operation, since it doesn't yet make use of the controller and it implements its own logic. The CLI tried to delete the eks-a cluster and get blocked forever since the controller wouldn't remove the finalizer for a paused cluster.

To solve this, we introduce a new annotation that signals the controller that the cluster is being managed the CLI and hence it should allow to be deleted without going through its normal deletion process.

There were also other hidden problems in the way cluster manager implemented the deletion, by wrapping a bunch of API operations in one big retry block. Some of these are not idempotent in combination with the others (for example, try to update annotations after deleting a resource). In addition, some of these operations had their own retry mechanisms internally, causing the total timeout to increase significantly. Now each individual operation has its own retry mechanism.

## Testing
Unit tests and manually ran `TestVSphereKubernetes121MulticlusterWorkloadCluster`

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
@@ -128,19 +129,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ 
 		}
 	}()
 
+	if !cluster.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, log, cluster)
+	}
+
 	// If the cluster is paused, return without any further processing.
 	if cluster.IsReconcilePaused() {
 		log.Info("Cluster reconciliation is paused")
 		return ctrl.Result{}, nil
 	}
 
-	if cluster.DeletionTimestamp.IsZero() {
-		if !controllerutil.ContainsFinalizer(cluster, ClusterFinalizerName) {
-			controllerutil.AddFinalizer(cluster, ClusterFinalizerName)
-		}
-	} else {
-		return r.reconcileDelete(ctx, log, cluster)
-	}
+	// AddFinalizer	is idempotent
+	controllerutil.AddFinalizer(cluster, ClusterFinalizerName)
 
 	if cluster.Spec.BundlesRef == nil {
 		if err = r.setBundlesRef(ctx, cluster); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -15,6 +15,11 @@ const (
 	// object to prevent a controller from processing a resource.
 	pausedAnnotation = "anywhere.eks.amazonaws.com/paused"
 
+	// ManagedByCLIAnnotation can be applied to an EKS-A Cluster to signal when the CLI is currently
+	// performing an operation so the controller should not take any action. When marked for deletion,
+	// the controller will remove the finalizer and let the cluster be deleted.
+	ManagedByCLIAnnotation = "anywhere.eks.amazonaws.com/managed-by-cni"
```
jiayiwang7, line 21:
```suggestion
	ManagedByCLIAnnotation = "anywhere.eks.amazonaws.com/managed-by-cli"
```

pkg/clustermanager/cluster_manager.go
```diff
@@ -399,46 +399,61 @@ func (c *ClusterManager) RunPostCreateWorkloadCluster(ctx context.Context, manag
...
+		return err
+	}
 
-			return provider.PostClusterDeleteValidate(ctx, managementCluster)
-		},
-	)
+	return provider.PostClusterDeleteValidate(ctx, managementCluster)
+}
+
+func (c *ClusterManager) deleteEKSAObjects(ctx context.Context, managementCluster, clusterToDelete *types.Cluster, provider providers.Provider, clusterSpec *cluster.Spec) error {
```
jiayiwang7, line 416: What does the controller workload cluster deletion logic look like? I would imagine running `kubectl delete eksa-cluster` would trigger the chain of deletion from eks-a objects (including gitops, iam, etc.) all the way to the capi objects. If that is the case shouldn't we let controller clean up the resources instead of from CLI?

pkg/clustermanager/cluster_manager.go
```diff
@@ -399,46 +399,61 @@ func (c *ClusterManager) RunPostCreateWorkloadCluster(ctx context.Context, manag
...
+	}
+
+	if clusterSpec.AWSIamConfig != nil {
+		log.V(2).Info("Deleting AWSIamConfig", "name", clusterSpec.AWSIamConfig.Name)
+		if err := c.clusterClient.DeleteAWSIamConfig(ctx, managementCluster, clusterSpec.AWSIamConfig.Name, clusterSpec.AWSIamConfig.Namespace); err != nil {
+			return err
+		}
+	}
+
+	log.V(2).Info("Cleaning up provide specific resources")
```
taneyland, line 451:
```suggestion
	log.V(2).Info("Cleaning up provider specific resources")
```

pkg/clustermanager/cluster_manager.go
g-gaston: We should let the controller do the deletion yeah, but that's a bigger change. I'm not convinced the controller should delete the other objects but that's a different topic.

In fact we should use the controller for create, upgrade and delete workload clusters and have the logic in one single place. But that's out scope for this first implementation, hopefully a fast follower.

pkg/clustermanager/cluster_manager.go
```diff
@@ -1082,8 +1107,9 @@ func (c *ClusterManager) resumeReconcileForCluster(ctx context.Context, clusterC
 	pausedAnnotation := cluster.PausedAnnotation()
 	err := c.clusterClient.RemoveAnnotationInNamespace(ctx, provider.DatacenterResourceType(), cluster.Spec.DatacenterRef.Name, pausedAnnotation, clusterCreds, cluster.Namespace)
 	if err != nil {
-		return fmt.Errorf("updating annotation when unpausing datacenterconfig reconciliation: %v", err)
+		return fmt.Errorf("removing paused annotation when unpausing datacenterconfig reconciliation: %v", err)
```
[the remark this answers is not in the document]
g-gaston: done

taneyland (approved):

jiayiwang7 (approved):

controllers/cluster_controller.go
```diff
@@ -178,6 +178,17 @@ func (r *ClusterReconciler) reconcileDelete(ctx context.Context, log logr.Logger
...
+	if metav1.HasAnnotation(cluster.ObjectMeta, anywherev1.ManagedByCLIAnnotation) {
+		log.Info("Clusters is managed by CLI, removing finalizer")
+		controllerutil.RemoveFinalizer(cluster, ClusterFinalizerName)
+		return ctrl.Result{}, nil
+	}
+
+	if cluster.IsReconcilePaused() {
+		log.Info("Cluster reconciliation is paused, won't process cluster deletion")
+		return ctrl.Result{}, nil
+	}
```
cxbrowne1207, line 190: Why is this check needed again after adding and potentially removing the finalizer?

cxbrowne1207 (approved):
/lgtm

PR #4239, opened 2022-11-30
[WIP] fixed the loop of get cluster by name

deepakm-ntnx:
currently list cluster api returns all the cluster and ignores filter name=clustername
To fix this behavior, we are also now checking if the returned list entity has same name along with existing check if its a pc entity or not.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


abhinavmpandey08 (approved):
/approve
/lgtm

PR #4245, opened 2022-12-01
Add adot e2e test

ivyostosh:

### DIFF
--- a/test/e2e/adot_test.go
+++ b/test/e2e/adot_test.go
@@ -6,14 +6,17 @@ package e2e
 import (
 	"testing"
 
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
-func TestCPackagesAdotInstallSimpleFlow(t *testing.T) {
+func TestCPackagesAdotDockerUbuntuKubernetes121SimpleFlow(t *testing.T) {
 	framework.CheckCuratedPackagesCredentials(t)
 	test := framework.NewClusterE2ETest(t, framework.NewDocker(t),
-		framework.WithPackageConfig(t, EksaPackageBundleURI,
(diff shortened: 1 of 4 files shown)


test/e2e/adot_test.go
```diff
@@ -0,0 +1,33 @@
...
+	framework.CheckCuratedPackagesCredentials(t)
+	test := framework.NewClusterE2ETest(t, framework.NewDocker(t),
+		framework.WithPackageConfig(t, EksaPackageBundleURI,
+			EksaPackageControllerHelmChartName, EksaPackageControllerHelmURI,
+			EksaPackageControllerHelmVersion, EksaPackageControllerHelmValues),
+	)
+	runAdotInstallSimpleFlow(test) // other args as necessary
+}
+
+func runAdotInstallSimpleFlow(test *framework.ClusterE2ETest) {
```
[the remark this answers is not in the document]
ivyostosh: The main test here for adot is to check if metrics are available for scrapping at `8888/metrics`, and the main logic for that is covered under `test/framework/cluster.go`.

I will still call this SimpleFlow since there are more advanced cases I can think so, but I'm still assessing if they are worth adding.

test/e2e/adot_test.go
```diff
@@ -0,0 +1,33 @@
...
+package e2e
+
+import (
+	"testing"
+
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/test/framework"
+)
+
+func TestCPackagesAdotInstallSimpleFlow(t *testing.T) {
```
jonahjon, line 13: I totally understand the premise of starting small on the tests, but the naming convention here should include the K8s version number, and provider. I think at a minimum you should be running one test on each provider as well.

jonahjon: usually it's k8s version, Node OS, Provider in the name. It's a bit cumbersome, but it's nice to see what is failing just by looking at a test name, rather than digging into the code.
`TestCPackagesEmissaryDockerUbuntuKubernetes122SimpleFlow`

ivyostosh: Added!

PR #4261, opened 2022-12-01
TinkerbellIP of management cluster and workload cluster should be same

panktishah26:
*Description of changes:*
Override tinkerbellIP of workload cluster with the tinkerbellIP of management cluster while cluster creation. 

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/create.go
+++ b/pkg/providers/tinkerbell/create.go
@@ -170,6 +170,19 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		if err := p.getHardwareFromManagementCluster(ctx, clusterSpec.ManagementCluster); err != nil {
 			return err
 		}
+
+		// for workload cluster use tinkerbell IP of the management cluster
+		managementClusterSpec, err := p.providerKubectlClient.GetEksaCluster(ctx, clusterSpec.ManagementCluster, clusterSpec.ManagementCluster.Name)
+		if err != nil {
+			return err
+		}
+
+		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, clusterSpec.Cluster.Namespace)
+		if err != nil {
+			return fmt.Errorf("getting TinkerbellIP of management cluster: %s", err)
+		}
+
+		p.datacenterConfig.Spec.TinkerbellIP = managementDatacenterConfig.Spec.TinkerbellIP
 	}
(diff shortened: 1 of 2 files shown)


pkg/providers/tinkerbell/create.go
```diff
@@ -170,6 +170,19 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		if err := p.getHardwareFromManagementCluster(ctx, clusterSpec.ManagementCluster); err != nil {
 			return err
 		}
+
+		// for workload cluster use tinkerbell IP of the management cluster
+		managementClusterSpec, err := p.providerKubectlClient.GetEksaCluster(ctx, clusterSpec.ManagementCluster, clusterSpec.ManagementCluster.Name)
```
pokearu, line 175: Are we considering this a part of a preflight validations? Because I think we wanted to add two things

1. Default the TinkerbellIP for workload to that from management.
2. Validate to ensure that the tinkerbell IPs are match.

pkg/providers/tinkerbell/create.go
```diff
@@ -170,6 +170,19 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
...
+		if err != nil {
+			return err
+		}
+
+		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, clusterSpec.Cluster.Namespace)
+		if err != nil {
+			return fmt.Errorf("getting TinkerbellIP of management cluster: %s", err)
+		}
+
+		p.datacenterConfig.Spec.TinkerbellIP = managementDatacenterConfig.Spec.TinkerbellIP
```
pokearu, line 185: I am guessing `p.datacenterConfig` is the object that gets passed around but do we want to update the clusterSpec object as well?

pkg/providers/tinkerbell/create.go
```diff
@@ -170,6 +170,19 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
...
 	}
 	// TODO(chrisdoherty4) Look to inject the validator. Possibly look to use a builder for
 	// constructing the validations rather than injecting flags into the provider.
```
pokearu, line 188: Do we want to add it as a part of `clusterSpecValidator` below?

pkg/providers/tinkerbell/create.go
panktishah26: We are validating tinkerbellIP constraints after overriding it inside `clusterSpecValidator`.

pokearu:
Discussed comments offline
/lgtm

PR #4268, opened 2022-12-02
Refactor ip in use algorithm

chrisdoherty4:
Refactor the IsIPInUse algorithm to be faster and more reliable.

The IsIPInUse algorithm iterated over a series of ports testing if they were open at some target address via TCP. It failed to take into consideration devices that are serving the IP but have no services listening on the checked ports resulting in a false negative.

Instead of relying on ports we can focus on whether the device responded. That can be achieved by either validating a TCP connection was in-fact established, or simply that the connection was reset by the server. The timeout period is still our indicator that the IP is not in use.

This check is still best effort but fixes the false negative case where none of the checked ports are open while also being 2s faster making it more appropriate for use in webhooks if desired.

#### Additional testing

I manually verified the errors returned by running the algorithm as part of an independent binary.

### DIFF
--- a/pkg/networkutils/networkutils.go
+++ b/pkg/networkutils/networkutils.go
@@ -1,9 +1,11 @@
 package networkutils
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


jacobweinstock (approved):

g-gaston (approved):
/lgtm

PR #4273, opened 2022-12-02
Block upgrade rollout strategy configuration for vsphere

ptrivedi:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -271,6 +271,9 @@ func (p *vsphereProvider) PostMoveManagementToBootstrap(_ context.Context, _ *ty
 }
 
 func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpec *cluster.Spec) error {
+	if err := p.validator.validateUpgradeRolloutStrategy(clusterSpec); err != nil {
+		return fmt.Errorf("failed setup and validations: %v", err)
+	}
 	if err := SetupEnvVars(clusterSpec.VSphereDatacenter); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -271,6 +271,14 @@ func (p *vsphereProvider) PostMoveManagementToBootstrap(_ context.Context, _ *ty
...
 
 func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpec *cluster.Spec) error {
+	if clusterSpec.Cluster.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy != nil {
+		return fmt.Errorf("Upgrade rollout strategy customization is not supported for vSphere provider")
+	}
+	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		if workerNodeGroupConfiguration.UpgradeRolloutStrategy != nil {
+			return fmt.Errorf("Upgrade rollout strategy customization is not supported for vSphere provider")
+		}
+	}
```
abhinavmpandey08, line 281: Since this same validation is repeated for create, upgrade and delete, can we place this in a separate function. Something like `validateRolloutStrategy` and just call that function?

abhinavmpandey08 (commented):
looks good, just 1 minor comment

pkg/providers/vsphere/vsphere.go
ptrivedi: Yes

PR #4275, opened 2022-12-02
AWS IAM Auth lifecycle controller support

pokearu:
*Description of changes:*
Adding support for reconciling awsiamconfig identity provider. The changes adds steps in the cluster controller reconcile flow

- We create a CA secret on the management cluster before the controlplane comes up for the workload cluster.
- Once the new workload cluster is ready we apply the aws-iam-authenticator manifest.
- If changes are made to mutable fields on the awsiamconfig object, the controller reconciles and does necessary updates.

*Testing:*
```bash
kubectl apply -f cluster-with-iam.yaml
```


pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	err := client.Get(ctx, types.NamespacedName{Name: awsiamauth.GetAwsIamAuthCaSecretName(clusterName), Namespace: constants.EksaSystemNamespace}, s)
+	if apierrors.IsNotFound(err) {
+		logger.Info("Applying aws-iam-authenticator CA secret")
+		return r.applyCASecret(ctx, client, clusterName)
+	}
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("fetching secret %s: %v", awsiamauth.GetAwsIamAuthCaSecretName(clusterName), err)
+	}
+
+	// If Secret is present, no-op
```
g-gaston, line 67: same with this comment, is not really adding anything

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	logger.Info("aws-iam-authenticator CA secret found", "Name", awsiamauth.GetAwsIamAuthCaSecretName(clusterName))
```
g-gaston, line 68: What are you trying to communicate with this log?
If I didn't know where the log is in the code and I read this in the controller output, I wouldn't know at all what happened

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	return controller.Result{}, nil
+}
+
+func (r *Reconciler) applyCASecret(ctx context.Context, client client.Client, clusterName string) (controller.Result, error) {
```
g-gaston, line 72:
```suggestion
func (r *Reconciler) createCASecret(ctx context.Context, client client.Client, clusterName string) (controller.Result, error) {
```

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	// Generate k8s secret yaml for aws-iam-authenticator
```
g-gaston, line 73: same with this comment

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	yaml, err := r.templateBuilder.GenerateCertKeyPairSecret(r.certgen, clusterName)
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("generating aws-iam-authenticator ca secret: %v", err)
+	}
+
+	return controller.Result{}, serverside.ReconcileYaml(ctx, client, yaml)
```
g-gaston, line 79: Since we want to ensure this secret is only generated one, I would vote to have a method that just creates objects instead of apply. That ensures it will fail it we try to create it again.

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+		certgen:              certgen,
+		templateBuilder:      &awsiamauth.TemplateBuilder{},
+		clusterID:            clusterID,
+		remoteClientRegistry: remoteClientRegistry,
+	}
+}
+
+// ReconcileAWSIAMAuthCASecret ensures the AWS IAM Authenticator secret is present.
+// It uses a controller.Result to indicate when requeues are needed.
+func (r *Reconciler) ReconcileAWSIAMAuthCASecret(ctx context.Context, logger logr.Logger, client client.Client, clusterName string) (controller.Result, error) {
```
g-gaston, line 55: Why don't we have the client injected as a dependency in the constructor?

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	// Generate k8s secret yaml for aws-iam-authenticator
+	yaml, err := r.templateBuilder.GenerateCertKeyPairSecret(r.certgen, clusterName)
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("generating aws-iam-authenticator ca secret: %v", err)
+	}
+
+	return controller.Result{}, serverside.ReconcileYaml(ctx, client, yaml)
+}
+
+// Reconcile takes the AWS IAM Authenticator indentity provider to the desired state defined in AWSIAMConfig.
```
g-gaston, line 82:
```suggestion
// Reconcile takes the AWS IAM Authenticator installation to the desired state defined in AWSIAMConfig.
```

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+// It uses a controller.Result to indicate when requeues are needed.
+// Intended to be used in a kubernetes controller.
+func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client client.Client, cluster *anywherev1.Cluster) (controller.Result, error) {
+	// Build the cluster spec for the Cluster
```
g-gaston, line 86: same for this comment and all the other comments in the method

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+		// Use a new clusterID
+		clusterID = r.clusterID
+		return r.applyIAMAuthManifest(ctx, rClient, clusterSpec, clusterID)
+	}
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("fetching configmap %s: %v", awsiamauth.AwsIamAuthConfigMapName, err)
+	}
+
+	logger.Info("Applying aws-iam-authenticator manifest")
+	return r.applyIAMAuthManifest(ctx, rClient, clusterSpec, clusterID)
```
g-gaston, line 112: nit: can we avoid having two calls to `applyIAMAuthManifest`? it makes the code a bit more difficult to read

Also, a couple good examples where comments can help understand the code

```suggestion
	cm := &corev1.ConfigMap{}
	err = rClient.Get(ctx, types.NamespacedName{Name: awsiamauth.AwsIamAuthConfigMapName, Namespace: constants.KubeSystemNamespace}, cm)
	if apierrors.IsNotFound(err) {
		// If the ConfigMap doesn't exist, this is the first time we install IAM auth
		// in the cluster, so generate a new UUID
		clusterID = r.clusterID
	} else if err != nil {
		return controller.Result{}, fmt.Errorf("fetching %s configmap: %v", awsiamauth.AwsIamAuthConfigMapName, err)
	}

	logger.Info("Applying aws-iam-authenticator manifest")
	// If clusterID is not set, this assumes there is already one and won't overwrite it
	return r.applyIAMAuthManifest(ctx, rClient, clusterSpec, clusterID)
```

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (r *Reconciler) applyIAMAuthManifest(ctx context.Context, client client.Client, clusterSpec *anywhereCluster.Spec, clusterID uuid.UUID) (controller.Result, error) {
```
g-gaston, line 115: We should probably document that `clusterID` can be not set and this method will respect the current value. Also, should we make it a pointer so it can actually be nil?

pkg/awsiamauth/template.go
```diff
@@ -83,3 +96,37 @@ func (t *TemplateBuilder) mapUsersToYaml(m []v1alpha1.MapUsers) (string, error)
...
+		"keyPemBytes":  base64.StdEncoding.EncodeToString(keyPemBytes),
+	}
+	awsIamAuthCaSecret, err := templater.Execute(awsIamAuthCaSecretTemplate, data)
+	if err != nil {
+		return nil, fmt.Errorf("generating aws-iam-authenticator cert key pair secret: %v", err)
+	}
+	return awsIamAuthCaSecret, nil
+}
+
+// GenerateKubeconfig generates a Kubeconfig for access management with AWS IAM Authenticator.
```
g-gaston, line 119:
```suggestion
// GenerateKubeconfig generates a Kubeconfig in yaml format to authenticate with AWS IAM Authenticator.
```

g-gaston (commented):
Nice!

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+// Intended to be used in a kubernetes controller.
+func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client client.Client, cluster *anywherev1.Cluster) (controller.Result, error) {
+	// Build the cluster spec for the Cluster
+	clusterSpec, err := anywhereCluster.BuildSpec(ctx, clientutil.NewKubeClient(client), cluster)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	// Get the client object for the workload cluster
+	rClient, err := r.remoteClientRegistry.GetClient(ctx, controller.CapiClusterObjectKey(cluster))
```
g-gaston, line 93: Before doing this, I would check if the CP is ready. This line will fail if not, so the end result would be similar, but it's a better way of handling such scenario.

pkg/awsiamauth/reconciler/reconciler.go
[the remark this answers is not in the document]
pokearu: Sounds good 👍

pkg/awsiamauth/reconciler/reconciler.go
pokearu: Actually we check for UUID.Nil in the [GenerateManifest](https://github.com/aws/eks-anywhere/blob/main/pkg/awsiamauth/template.go#L22) function. Thats why i hadn't used pointers. I could probably look into refactoring that separately since it changes the CLI create as well.

pkg/awsiamauth/reconciler/reconciler.go
pokearu: So currently the point in the code where this is called is after the workload cluster has come up. Since we requeue from the cluster controller till that.
But I can add a check to be sure 👍

controllers/cluster_controller.go
```diff
@@ -162,6 +165,11 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, log logr.Logger, clus
 
 	var reconcileResult controller.Result
 	var err error
+
+	if _, err = r.preClusterProviderReconcile(ctx, log, cluster); err != nil {
```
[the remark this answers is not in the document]
pokearu: Good point. I think we can check it, I went with the assumption that the only step there returns no requeues. But that is not a valid assumption going forward.

pkg/awsiamauth/reconciler/reconciler.go
pokearu: Yup that seems better 👍 Initially I was not sure if I would need client or remoteClient more so i was just passing it in.

controllers/cluster_controller.go
```diff
@@ -172,9 +180,44 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, log logr.Logger, clus
...
+	}
+
 	return reconcileResult.ToCtrlResult(), nil
 }
 
+func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	// Reconcile Cluster IdentityProviders
+	for _, identityProvider := range cluster.Spec.IdentityProviderRefs {
+		switch identityProvider.Kind {
+		case anywherev1.AWSIamConfigKind:
```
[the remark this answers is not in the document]
pokearu: Added a HasIamConfig method to perform the checks 👍

controllers/factory.go
```diff
@@ -291,3 +296,26 @@ func (f *Factory) withCNIReconciler() *Factory {
...
+func (f *Factory) withAWSIamConfigReconciler() *Factory {
+	f.withTracker()
+
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.awsIamConfigReconciler != nil {
+			return nil
+		}
+
+		certgen := crypto.NewCertificateGenerator()
+		clusterID := uuid.New()
```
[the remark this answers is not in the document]
pokearu: A new UUID is used for every new install of aws-iam-auth. Modified to have a UUIDGenerator type.

controllers/cluster_controller.go
```diff
@@ -172,9 +190,34 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, log logr.Logger, clus
...
+	if reconcileResult, err = r.postClusterProviderReconcile(ctx, log, cluster); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	return reconcileResult.ToCtrlResult(), nil
 }
 
+func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	if cluster.HasAWSIamConfig() {
+		return r.awsIamAuth.EnsureCASecret(ctx, log, cluster)
```
g-gaston, line 207: nit, if you want to have a generic method like this, we shouldn't return directly
```suggestion
		if err := r.awsIamAuth.EnsureCASecret(ctx, log, cluster); err != nil {
		    return err
		}
```

controllers/cluster_controller.go
```diff
@@ -172,9 +190,34 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, log logr.Logger, clus
...
+	}
+
+	return controller.Result{}, nil
+}
+
+func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	if cluster.HasAWSIamConfig() {
+		return r.awsIamAuth.Reconcile(ctx, log, cluster)
```
g-gaston, line 215: same comment about returning here

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -58,6 +58,18 @@ type ClusterSpec struct {
 	BundlesRef *BundlesRef `json:"bundlesRef,omitempty"`
 }
 
+// HasAWSIamConfig checks if AWSIamConfig is configured for the cluster.
+func (c *Cluster) HasAWSIamConfig() bool {
+	for _, identityProvider := range c.Spec.IdentityProviderRefs {
+		switch identityProvider.Kind {
+		case AWSIamConfigKind:
+			return true
+		}
```
g-gaston, line 67: why a switch for only one case?
```suggestion
		if identityProvider.Kind == AWSIamConfigKind {
			return true
		}
```

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	err := r.client.Get(ctx, types.NamespacedName{Name: awsiamauth.CASecretName(clusterName), Namespace: constants.EksaSystemNamespace}, s)
+	if apierrors.IsNotFound(err) {
+		logger.Info("Creating aws-iam-authenticator CA secret")
+		return r.createCASecret(ctx, cluster)
+	}
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("fetching secret %s: %v", awsiamauth.CASecretName(clusterName), err)
+	}
+
+	logger.Info("aws-iam-authenticator CA secret found. Skipping secret create.", "Name", awsiamauth.CASecretName(clusterName))
```
g-gaston, line 69: nit
```suggestion
	logger.Info("aws-iam-authenticator CA secret found. Skipping secret create.", "name", awsiamauth.CASecretName(clusterName))
```

g-gaston: nit: can we store the secret name in a variable? so we don't need call this method twice

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	return controller.Result{}, nil
+}
+
+func (r *Reconciler) createCASecret(ctx context.Context, cluster *anywherev1.Cluster) (controller.Result, error) {
+	yaml, err := r.templateBuilder.GenerateCertKeyPairSecret(r.certgen, cluster.Name)
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("generating aws-iam-authenticator ca secret: %v", err)
+	}
+
+	return controller.Result{}, serverside.ReconcileYaml(ctx, r.client, yaml)
```
g-gaston, line 79: This is still using server side apply, we should change this to create

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+// Reconcile takes the AWS IAM Authenticator installation to the desired state defined in AWSIAMConfig.
+// It uses a controller.Result to indicate when requeues are needed.
+// Intended to be used in a kubernetes controller.
+func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, cluster *anywherev1.Cluster) (controller.Result, error) {
+	clusterSpec, err := anywhereCluster.BuildSpec(ctx, clientutil.NewKubeClient(r.client), cluster)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	result, err := clusters.CheckControlPlaneReady(ctx, r.client, logger, cluster)
```
g-gaston, line 91: this was intended to be used as a cluster reconciler phase. The interface (using Result) doesn't fit this code very well. We should probably some of move the logic inside `CheckControlPlaneReady` to a shared method that just returns true or false.

I'm ok doing this as a follow up.

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ensureCASecretOwnerRef ensures that the CAPI Cluster object is set as an ownerReference.
+// The ownerReference ensures the CA Secret is deleted when the cluster is deleted.
+func (r *Reconciler) ensureCASecretOwnerRef(ctx context.Context, logger logr.Logger, cluster *anywherev1.Cluster) error {
+	secret := &corev1.Secret{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: awsiamauth.CASecretName(cluster.Name), Namespace: constants.EksaSystemNamespace}, secret); err != nil {
+		return fmt.Errorf("fetching secret %s: %v", awsiamauth.CASecretName(cluster.Name), err)
+	}
+
+	if len(secret.GetOwnerReferences()) != 0 {
```
g-gaston, line 141: this is not safe, what if the secret already has a different owner ref?

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil
+	}
+
+	capiCluster := &clusterv1.Cluster{}
```
g-gaston, line 145: Why use the capi cluster? I think we should use the eks-a cluster for this.
And this can be done when the secret is created.

pkg/awsiamauth/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,314 @@
...
+	"github.com/aws/eks-anywhere/pkg/awsiamauth"
+	"github.com/aws/eks-anywhere/pkg/awsiamauth/reconciler"
+	reconcilermocks "github.com/aws/eks-anywhere/pkg/awsiamauth/reconciler/mocks"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/controller"
+	cryptomocks "github.com/aws/eks-anywhere/pkg/crypto/mocks"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+func TestEnsureCASecret_SecretFound(t *testing.T) {
```
g-gaston, line 34: Can we use the standard test naming?
`Test[Entity tested][Method tested][info about the scenario tested]`

pkg/awsiamauth/reconciler/reconciler.go
pokearu: I initially tired that. But that gave an error because we need the owner and the secret to be in the same namespace.

The EKSA cluster can be in Default but the Secret has to be in the `eksa-system`.

Thats why i used to CAPI cluster and set it later.

pkg/awsiamauth/reconciler/reconciler.go
pokearu: Fair point. Will do a previous count comparison.

pkg/awsiamauth/reconciler/reconciler.go
pokearu: Makes sense. 
I will strip out the logic and fetch the capi object. Since I am fetching it again for ownerRef setting, I can prevent that and fetch it once.

pkg/awsiamauth/reconciler/reconciler.go
g-gaston: Yeah, that makes sense. Still, using the capi cluster is a bit hacky. How is the capi cluster the owner of this secret, if the capi spec depends on this secret?

pokearu: Well the KCP object has a secretRef to the CA secret, the CAPI cluster itself does not depend on the Secret.
I was going with the idea that since the consumer of this secret is a CAPI API object, the CAPI cluster as an owner would make sense?

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+	secretName := awsiamauth.CASecretName(cluster.Name)
+	secret := &corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      secretName,
+			Namespace: constants.EksaSystemNamespace,
+		},
+	}
+
+	logger.Info("Deleting aws-iam-authenticator ca secret", "name", secretName, "namespace", constants.EksaSystemNamespace)
+	if err := r.client.Delete(ctx, secret); err != nil {
```
g-gaston, line 158: probably want to check if this is a not found and then not return an error

pkg/awsiamauth/name.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// CASecretName returns the name of AWS IAM Authenticator secret containing the CA for the cluster.
+func CASecretName(clusterName string) string {
+	return fmt.Sprintf("%s-%s", clusterName, awsIamAuthCaSecretSuffix)
+}
+
+// KubeconfigSecretName returns the name of the AWS IAM Authenticator kubeconfig secret for the cluster.
+func KubeconfigSecretName(clusterName string) string {
+	return fmt.Sprintf("%s-%s", clusterName, awsIamAuthKubeconfigSuffix)
```
g-gaston, line 23: Add test

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	// CheckControlPlaneReady was not meant to be used here.
+	// It was intended as a phase in cluster reconciliation.
+	// TODO (pokearu): Break down the function to better reuse it.
+	result, err := clusters.CheckControlPlaneReady(ctx, r.client, logger, cluster)
+	if err != nil {
+		return controller.Result{}, fmt.Errorf("checking controlplane ready: %v", err)
```
g-gaston, line 110: use errors package to get stack trace

pkg/awsiamauth/reconciler/reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+
+	clusterCaSecretName := clusterapi.ClusterCASecretName(cluster.Name)
+	clusterCaSecret := &corev1.Secret{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: clusterCaSecretName, Namespace: constants.EksaSystemNamespace}, clusterCaSecret); err != nil {
+		return fmt.Errorf("fetching cluster ca secret: %v", err)
+	}
+
+	clusterTLSCrt, ok := clusterCaSecret.Data["tls.crt"]
+	if !ok {
+		return fmt.Errorf("tls.crt key not found in cluster CA secret: %s", clusterCaSecret.Name)
```
g-gaston, line 165:
```suggestion
		return fmt.Errorf("tls.crt key not found in cluster CA secret %s", clusterCaSecret.Name)
```

pkg/clustermanager/eksa_installer_test.go
```diff
@@ -59,7 +59,7 @@ func TestEKSAInstallerInstallSuccessWithRealManifest(t *testing.T) {
 	tt := newInstallerTest(t)
 	tt.newSpec.VersionsBundle.Eksa.Components.URI = "../../config/manifest/eksa-components.yaml"
 	tt.client.EXPECT().Apply(tt.ctx, tt.cluster.KubeconfigFile, gomock.AssignableToTypeOf(&appsv1.Deployment{}))
-	tt.client.EXPECT().Apply(tt.ctx, tt.cluster.KubeconfigFile, gomock.Any()).Times(30) // there 30 other objects in the manifes
+	tt.client.EXPECT().Apply(tt.ctx, tt.cluster.KubeconfigFile, gomock.Any()).Times(32) // there 30 other objects in the manifes
```
vivek-koppuru, line 62: does the comment need to be updated too to 32?

vivek-koppuru (approved):

PR #4289, opened 2022-12-05
Save CLI log output to file included in support bundle

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
 This PR customizes the zap logger to log verbose logs to a file, and then include a copy of it in the support bundles when one is generated. This offers an easier way for the customer recover CLI run logs for debugging, when there are errors and interruptions of some kind.

*Testing (if applicable):*

### DIFF
--- a/pkg/diagnostics/collectors.go
+++ b/pkg/diagnostics/collectors.go
@@ -2,6 +2,7 @@ package diagnostics
 
 import (
 	"fmt"
+	"path/filepath"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -9,6 +10,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/files"
 	"github.com/aws/eks-anywhere/pkg/providers"
 )
 
@@ -168,6 +170,27 @@ func (c *collectorFactory) PackagesCollectors() []*Collect {
 	var collectors []*Collect
 	collectors = append(collectors, c.packagesCrdCollectors()...)
 	collectors = append(collectors, c.packagesLogCollectors()...)
+	return collectors
+}
+
+func (c *collectorFactory) FileCollectors(paths []string) []*Collect {
+	collectors := []*Collect{}
+
+	for _, path := range paths {
+		r := files.NewReader()
+		content, err := r.ReadFile(path)
+		if err != nil {
+			content = []byte(fmt.Sprintf("Failed to retrieve file %s for collection: %s", path, err))
+		}
+
+		collectors = append(collectors, &Collect{
+			Data: &data{
+				Data: string(content),
+				Name: filepath.Base(path),
+			},
+		})
+	}
+
 	return collectors
 }
 
--- a/pkg/diagnostics/diagnostic_bundle.go
+++ b/pkg/diagnostics/diagnostic_bundle.go
@@ -70,6 +70,7 @@ func newDiagnosticBundleManagementCluster(af AnalyzerFactory, cf CollectorFactor
 	}
 
 	b.WithDefaultCollectors().
+		WithFileCollectors([]string{logger.GetOutputFilePath()}).
 		WithDefaultAnalyzers().
 		WithManagementCluster(true).
 		WithDatacenterConfig(spec.Cluster.Spec.DatacenterRef, spec).
@@ -116,6 +117,7 @@ func newDiagnosticBundleFromSpec(af AnalyzerFactory, cf CollectorFactory, spec *
 		WithManagementCluster(spec.Cluster.IsSelfManaged()).
 		WithDefaultAnalyzers().
 		WithDefaultCollectors().
+		WithFileCollectors([]string{logger.GetOutputFilePath()}).
 		WithPackagesCollectors().
 		WithLogTextAnalyzers()
 
@@ -142,7 +144,9 @@ func newDiagnosticBundleDefault(af AnalyzerFactory, cf CollectorFactory) *EksaDi
 		analyzerFactory:  af,
 		collectorFactory: cf,
 	}
-	return b.WithDefaultAnalyzers().WithDefaultCollectors().WithManagementCluster(true)
+	return b.WithDefaultAnalyzers().
+		WithDefaultCollectors().
+		WithManagementCluster(true)
 }
 
 func newDiagnosticBundleCustom(af AnalyzerFactory, cf CollectorFactory, client BundleClient, kubectl *executables.Kubectl, bundlePath string, kubeconfig string, writer filewriter.FileWriter) *EksaDiagnosticBundle {
@@ -257,6 +261,12 @@ func (e *EksaDiagnosticBundle) WithManagementCluster(isSelfManaged bool) *EksaDi
 		e.bundle.Spec.Analyzers = append(e.bundle.Spec.Analyzers, e.analyzerFactory.ManagementClusterAnalyzers()...)
 		e.bundle.Spec.Collectors = append(e.bundle.Spec.Collectors, e.collectorFactory.ManagementClusterCollectors()...)
 	}
+	return e
+}
+
+// WithFileCollectors appends collectors that collect static data from the specified paths to the bundle.
+func (e *EksaDiagnosticBundle) WithFileCollectors(paths []string) *EksaDiagnosticBundle {
+	e.bundle.Spec.Collectors = append(e.bundle.Spec.Collectors, e.collectorFactory.FileCollectors(paths)...)
 	return e
 }
 
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -16,14 +16,22 @@ const (
 )
 
 var (
-	l    logr.Logger = logr.Discard()
-	once sync.Once
+	l              logr.Logger = logr.Discard()
+	once           sync.Once
+	outputFilePath string
 )
 
-func set(logger logr.Logger) {
+func set(logger logr.Logger, out string) {
 	once.Do(func() {
 		l = logger
+		outputFilePath = out
 	})
+}
+
+// GetOutputFilePath returns the path to the file where high verbosity logs are written to.
+// If the logger hasn't been configured to output to a file, it returns an empty string.
+func GetOutputFilePath() string {
+	return outputFilePath
 }
 
 // Get returns the logger instance that has been previously set.
@@ -82,12 +90,4 @@ func MarkFail(msg string, keysAndValues ...interface{}) {
 
 func MarkWarning(msg string, keysAndValues ...interface{}) {
 	l.V(0).Info(markWarning+msg, keysAndValues...)
-}
-
-type LoggerOpt func(logr *logr.Logger)
-
-func WithName(name string) LoggerOpt {
-	return func(logr *logr.Logger) {
-		*logr = (*logr).WithName(name)
-	}
 }
--- a/pkg/logger/zap.go
+++ b/pkg/logger/zap.go
@@ -2,45 +2,34 @@ package logger
 
 import (
 	"fmt"
+	"os"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
+
+// ZapOpts represents a set of arguments for initializing the zap logger.
+type ZapOpts struct {
+	Level          int      // indicates the log level of the logger.
+	OutputFilePath string   // if specified, the logger will output to file at this path.
+	WithNames      []string // specified name elements are added to the logger's name.
+}
 
 // InitZap creates a zap logger with the provided verbosity level
 // and sets it as the package logger.
 // 0 is the least verbose and 10 the most verbose.
 // The package logger can only be init once, so subsequent calls to this method
 // won't have any effect.
-func InitZap(level int, opts ...LoggerOpt) error {
-	cfg := zap.NewDevelopmentConfig()
-	cfg.Level = zap.NewAtomicLevelAt(zapcore.Level(-1 * level))
-	cfg.EncoderConfig.EncodeLevel = nil
-	cfg.EncoderConfig.EncodeTime = NullTimeEncoder
-	cfg.DisableCaller = true
-	cfg.DisableStacktrace = true
-
-	// Only enabling this at level 4 because that's when
-	// our debugging levels start. Ref: doc.go
-	if level >= 4 {
-		cfg.EncoderConfig.EncodeLevel = VLevelEncoder
-		cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
-	}
-
-	zapLog, err := cfg.Build()
+func InitZap(args ZapOpts) error {
+	logr, err := newZap(args)
 	if err != nil {
-		return fmt.Errorf("creating zap logger: %v", err)
+		return err
 	}
-
-	logr := zapr.NewLogger(zapLog)
-	for _, opt := range opts {
-		opt(&logr)
-	}
-
-	set(logr)
-	l.V(4).Info("Logger init completed", "vlevel", level)
+	set(logr, args.OutputFilePath)
+	l.V(4).Info("Logger init completed", "vlevel", args.Level)
 
 	return nil
 }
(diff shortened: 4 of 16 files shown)


cxbrowne1207:
\hold

cxbrowne1207:
\unhold

pkg/logger/logcacher.go
```diff
@@ -0,0 +1,149 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+)
+
+// Log represents a Loggern log entry.
+type Log struct {
```
g-gaston, line 13: Does this need to be exported?

pkg/logger/logcacher.go
```diff
@@ -0,0 +1,149 @@
...
+			filtered = append(filtered, log)
+		}
+	}
+
+	return filtered
+}
+
+// AddLog creates a new Log and stores it.
+func (c *LogCacher) AddLog(level int, msg string, keysAndValues ...interface{}) {
+	timestamp := time.Now().Format("2006-01-02T15:04:05.000-0600")
```
g-gaston, line 81: why do we need to handle this manually?

g-gaston (commented):
* Is there a requirement to store the logs in memory? Or is that just an implementation detail.
* Do we need this log dump always or should be just do it selectively where we need it (CLI commands for example)?
* How does this solve the interrupted SSH connection? It seems like we are only reading the in memory logs when the support bundle report is triggered. And that only happens if the CLI execution is maintained uninterrupted and only after a CLI error happens. If the SSH connection is dropped and the process is interrupted, the support bundle won't be created and there won't be any logs. TBH, the solution to this problem should be to use tmux or screen, SSH connections are out of the CLI's domain of responsibility, so any SSH associated issue should be solved with SSH tools. But you described that situation in your problem statement and I don't think this code solves that usecase.

Wouldn't this be simpler if we use [`config.OutputPaths`](https://pkg.go.dev/go.uber.org/zap#Config)?

cxbrowne1207:
So, if we want to always log verbosely to the file, using only the `OutputPath` would not work . Would have to customize the creation of the logger instead of using `Zap.Config.Build`. So, our options are

1. Settle for logging to the file with whatever level is specified by the customer.
2. Customize the step to build a logger where `zapcore.NewTee()` is used to duplicate the logs, to the file and to the console.
3. We may be able to instantiate, a secondary logger that only logs to a file but would have to customize that as well to have the same options and behavior as the original logger

danbudris:
> 2. Customize the step to build a logger where `zapcore.NewTee()` is used to duplicate the logs, to the file and to the console.

This seems reasonable to me; from looking [at the docs](https://pkg.go.dev/go.uber.org/zap#hdr-Configuring_Zap) and [some quick googling ](https://stackoverflow.com/a/72093594/4504862) this seems like the recommended approach.

g-gaston:
For the record, there was quorum since I voted option 2 offline as well :)

pkg/logger/config.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// Config helps to construct a customized zap logger.
+type Config struct {
```
g-gaston, line 12: does this need to be exported?

pkg/logger/config.go
```diff
@@ -0,0 +1,56 @@
...
+	OutputPaths   []string
+	Level         zap.AtomicLevel
+	EncoderConfig zapcore.EncoderConfig
+}
+
+// Build constructs a logger and returns a logger.
+func (cfg Config) Build() (*zap.Logger, error) {
```
g-gaston, line 19: exported?

pkg/logger/config.go
g-gaston, line 12: maybe name this file zapconfig? or just put it in the zap.go file

pkg/logger/config.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	logger := zap.New(cfg.buildCore(sink), cfg.buildOptions()...)
+	return logger, nil
+}
+
+func (cfg Config) buildOptions() []zap.Option {
+	opts := []zap.Option{}
+	// Enable development mode
+	opts = append(opts, zap.Development())
```
g-gaston, line 36: what does `zap.Development()` do and why are we setting it?

pkg/logger/config.go
```diff
@@ -0,0 +1,56 @@
...
+	return opts
```
g-gaston, line 37:
```suggestion
	// Enable development mode
	return []zap.Option{zap.Development()}
```

pkg/logger/config.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (cfg Config) openSinks() (zapcore.WriteSyncer, error) {
+	sink, _, err := zap.Open(cfg.OutputPaths...)
+	if err != nil {
+		return nil, err
+	}
+	return sink, nil
```
g-gaston, line 45:
```suggestion
	return sink, err
```

pkg/logger/zap.go
```diff
@@ -52,3 +55,8 @@ func VLevelEncoder(l zapcore.Level, enc zapcore.PrimitiveArrayEncoder) {
 
 // NullTimeEncoder skips time serialization.
 func NullTimeEncoder(t time.Time, enc zapcore.PrimitiveArrayEncoder) {}
+
+// NewAtomicLevelAt returns an appropriate zap.AtomicLevel given an integer representing the log level.
+func NewAtomicLevelAt(level int) zap.AtomicLevel {
```
g-gaston, line 60: does this need to be exported?

pkg/logger/zap.go
```diff
@@ -15,12 +15,15 @@ import (
 // The package logger can only be init once, so subsequent calls to this method
 // won't have any effect.
 func InitZap(level int, opts ...LoggerOpt) error {
-	cfg := zap.NewDevelopmentConfig()
-	cfg.Level = zap.NewAtomicLevelAt(zapcore.Level(-1 * level))
+	filename := fmt.Sprintf("./eksa-cli-%s.log", time.Now().Format("2006-01-02T15:04:05"))
+	cfg := Config{
+		Level:         NewAtomicLevelAt(level),
+		EncoderConfig: zap.NewDevelopmentEncoderConfig(),
+		OutputPaths:   []string{filename},
```
g-gaston, line 22: IMO this should be optional, not by default

pkg/logger/zap.go
```diff
@@ -15,12 +15,15 @@ import (
...
+	}
+
 	cfg.EncoderConfig.EncodeLevel = nil
 	cfg.EncoderConfig.EncodeTime = NullTimeEncoder
-	cfg.DisableCaller = true
-	cfg.DisableStacktrace = true
```
g-gaston, line 23: are we replacing these two with something else?

pkg/logger/logger.go
```diff
@@ -16,16 +16,23 @@ const (
...
 
-func set(logger logr.Logger) {
+func set(logger logr.Logger, out string) {
 	once.Do(func() {
 		l = logger
+		outputFilename = out
 	})
 }
 
+// GetOutputFileName returns the current outputFileName.
```
g-gaston, line 31:
```suggestion
// GetOutputFileName returns the path to the file where high verbosity logs are written to.
```

Or or something else, but this current comment doesn't tell me anything.

pkg/logger/logger.go
```diff
@@ -16,16 +16,23 @@ const (
 )
 
 var (
-	l    logr.Logger = logr.Discard()
-	once sync.Once
+	l              logr.Logger = logr.Discard()
+	once           sync.Once
+	outputFilename string
```
g-gaston, line 21: We should think about grouping these 3 (or at least the logger and output file) in a struct
Can be a follow up

pkg/diagnostics/collectors.go
```diff
@@ -171,6 +174,28 @@ func (c *collectorFactory) PackagesCollectors() []*Collect {
...
 
+func (c *collectorFactory) FileCollectors(paths []string) []*Collect {
+	collectors := []*Collect{}
+
+	for _, path := range paths {
+		r := files.NewReader()
+		content, err := r.ReadFile(path)
+
+		if err != nil {
+			logger.V(5).Info(fmt.Sprintf("skipping collection for file %s", path))
```
g-gaston, line 185: shouldn't we return an error?

pkg/diagnostics/diagnostic_bundle.go
```diff
@@ -142,7 +144,10 @@ func newDiagnosticBundleDefault(af AnalyzerFactory, cf CollectorFactory) *EksaDi
 		analyzerFactory:  af,
 		collectorFactory: cf,
 	}
-	return b.WithDefaultAnalyzers().WithDefaultCollectors().WithManagementCluster(true)
+	return b.WithDefaultAnalyzers().
+		WithDefaultCollectors().
+		WithFileCollectors([]string{logger.GetOutputFileName()}).
```
g-gaston, line 149: how does this work when the bundle is not generated from a `cluster` command? It could be generated directly with the `generate supportbundle` or whatever command. In that case, we probably don't need this.

pkg/logger/logger.go
cxbrowne1207: Agreed

pkg/logger/zap.go
cxbrowne1207: This was exported mainly so that I could access it from the tests

pkg/logger/zap.go
cxbrowne1207: no, there is not a need to I think. Under the hood of zap's Config in the buildOptions method [here](https://github.com/uber-go/zap/blob/master/config.go#L198), if this was true, it would skip a step. Since, we want to skip it anyway, i didn't include it in our custom config.

pkg/logger/config.go
cxbrowne1207: Development "changes the behavior of DPanicLevel and takes stack-traces more liberally." according the [documentation](https://pkg.go.dev/go.uber.org/zap#Config). It was enabled originally, although I am unsure if we need this seeing that we were disabling stack-traces.

pkg/logger/config.go
g-gaston: If it's not needed elsewhere, I would make it unexported and figure something else for the tests
Let's sync up offline

pkg/logger/config.go
g-gaston: Umm yeah I don't think we need it then

pkg/logger/zap.go
g-gaston: oh I'm not saying the CLI should be able to configure it (at least for now)
But that it should be optional when innit the zap logger

pkg/logger/zap.go
g-gaston: Does that mean that the new way of building a config will result in the same behavior as before?

pkg/diagnostics/collectors.go
g-gaston: Well that depends on how you handle the error up the stack
Either way, just logging this won't give any clue to the support engineer as of why the file is missing from the bundle, since this log won't be included in the bundle. Maybe we can make the content of the file the error message?

pkg/logger/zap.go
cxbrowne1207: Yes, that was the goal

cxbrowne1207: Yes, it should

pkg/diagnostics/collectors.go
cxbrowne1207: Good idea. I'll work that it

pkg/logger/config.go
cxbrowne1207: Removed

pkg/logger/zap.go
cxbrowne1207: Oh yeah, the logger is initialized in a few different places and they all don't need this. I've updated adding an output path to be optional.

pkg/logger/logger.go
```diff
@@ -16,16 +16,23 @@ const (
...
 
-func set(logger logr.Logger) {
+func set(logger logr.Logger, out string) {
 	once.Do(func() {
 		l = logger
+		outputFilePath = out
 	})
 }
 
+// GetOutputFilePath returns the path to the file where high verbosity logs are written to.
```
g-gaston, line 31:
```suggestion
// GetOutputFilePath returns the path to the file where high verbosity logs are written to.
// If the logger hasn't been configured to output to a file, it returns an empty string.
```

pkg/logger/zap.go
```diff
@@ -2,45 +2,62 @@ package logger
...
 	"time"
 
 	"github.com/go-logr/zapr"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+// InitZapArgs represents a set of arguments for initializing the zap logger.
+type InitZapArgs struct {
```
g-gaston, line 15:
```suggestion
type ZapOpts struct {
```

pkg/logger/zap.go
```diff
@@ -2,45 +2,62 @@ package logger
...
+	Level          int
+	OutputFilePath string
+	Opts           []LoggerOpt
```
g-gaston, line 18: This seems a bit too complex, nested options in options. I would just make this another normal field, like `WithNames []string`.

pkg/logger/zap.go
```diff
@@ -2,45 +2,62 @@ package logger
...
 // InitZap creates a zap logger with the provided verbosity level
 // and sets it as the package logger.
 // 0 is the least verbose and 10 the most verbose.
 // The package logger can only be init once, so subsequent calls to this method
 // won't have any effect.
-func InitZap(level int, opts ...LoggerOpt) error {
-	cfg := zap.NewDevelopmentConfig()
-	cfg.Level = zap.NewAtomicLevelAt(zapcore.Level(-1 * level))
+func InitZap(args InitZapArgs) error {
+	// filename := fmt.Sprintf("./eksa-cli-%s.log", time.Now().Format("2006-01-02T15:04:05"))
```
g-gaston, line 27:
```suggestion
```

pkg/logger/zap.go
```diff
@@ -52,3 +69,45 @@ func VLevelEncoder(l zapcore.Level, enc zapcore.PrimitiveArrayEncoder) {
...
+func (cfg config) openSinks() (zapcore.WriteSyncer, error) {
+	sink, _, err := zap.Open(cfg.OutputPaths...)
+	return sink, err
+}
+
+// config helps to construct a customized zap logger.
+type config struct {
+	OutputPaths   []string
+	Level         zap.AtomicLevel
+	EncoderConfig zapcore.EncoderConfig
```
g-gaston, line 102:
```suggestion
	outputPaths   []string
	level         zap.AtomicLevel
	encoderConfig zapcore.EncoderConfig
```

pkg/logger/zap.go
```diff
@@ -2,45 +2,62 @@ package logger
...
+
 // InitZap creates a zap logger with the provided verbosity level
 // and sets it as the package logger.
 // 0 is the least verbose and 10 the most verbose.
 // The package logger can only be init once, so subsequent calls to this method
 // won't have any effect.
-func InitZap(level int, opts ...LoggerOpt) error {
-	cfg := zap.NewDevelopmentConfig()
-	cfg.Level = zap.NewAtomicLevelAt(zapcore.Level(-1 * level))
+func InitZap(args InitZapArgs) error {
```
g-gaston, line 26: I feel like all of this would be easier to test if we split this method in
`InitZap` and `newZap`
`InitZap` would just call `newZap` and `set`.
And `newZap` would have the construction logic. That way you can test `newZap` directly and forget about `config`, `build`, etc. which are just implementation details.

pkg/logger/zap.go
g-gaston, line 18: I would add doc comments for the 3 fields to explain what they do

pkg/diagnostics/collectors.go
```diff
@@ -171,6 +173,27 @@ func (c *collectorFactory) PackagesCollectors() []*Collect {
...
 }
 
+func (c *collectorFactory) FileCollectors(paths []string) []*Collect {
+	collectors := []*Collect{}
+
+	for _, path := range paths {
+		r := files.NewReader()
+		content, err := r.ReadFile(path)
+		if err != nil {
+			content = []byte(fmt.Sprintf("Error retrieving file %s for collection", path))
```
g-gaston, line 183: can we include the error content?
```suggestion
			content = []byte(fmt.Sprintf("Failed retrieving file %s for collection: %s", path, err))
```

pkg/logger/zap.go
```diff
@@ -2,53 +2,116 @@ package logger
...
-		return fmt.Errorf("creating zap logger: %v", err)
+		return logr.Discard(), fmt.Errorf("creating zap logger: %v", err)
 	}
 
 	logr := zapr.NewLogger(zapLog)
-	for _, opt := range opts {
-		opt(&logr)
+
+	for _, name := range args.WithNames {
+		WithName(name)(&logr)
```
g-gaston, line 75: nit:
```suggestion
		logr = logr.WithName(name)
```
I think it's a bit more readable? Unless I'm missing something

g-gaston (approved):

PR #4305, opened 2022-12-07
Update WaitForPackagesInstalled command

ivyostosh:
Instead of using cluster as a parameter, we will use kubeconfig instead. This allows more versatile use cases as kubeconfig can not only be obtained by the cluster object, but also by the E2E environment.

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -379,8 +379,8 @@ func (k *Kubectl) WaitForJobCompleted(ctx context.Context, kubeconfig, timeout s
 }
 
 // WaitForPackagesInstalled waits for a package resource to reach installed state before returning.
-func (k *Kubectl) WaitForPackagesInstalled(ctx context.Context, cluster *types.Cluster, name string, timeout string, namespace string) error {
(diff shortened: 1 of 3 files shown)


pkg/executables/kubectl.go
```diff
@@ -379,8 +379,8 @@ func (k *Kubectl) WaitForJobCompleted(ctx context.Context, kubeconfig, timeout s
 }
 
 // WaitForPackagesInstalled waits for a package resource to reach installed state before returning.
-func (k *Kubectl) WaitForPackagesInstalled(ctx context.Context, cluster *types.Cluster, name string, timeout string, namespace string) error {
-	return k.WaitJSONPathLoop(ctx, cluster.KubeconfigFile, timeout, "status.state", "installed", fmt.Sprintf("%s/%s", eksaPackagesType, name), namespace)
+func (k *Kubectl) WaitForPackagesInstalled(ctx context.Context, kubeconfig, name string, timeout string, namespace string) error {
```
TerryHowe, line 382: Kind of odd that works, the type is string, but it uses the type from `name`

TerryHowe:
```suggestion
func (k *Kubectl) WaitForPackagesInstalled(ctx context.Context, kubeconfig string, name string, timeout string, namespace string) error {
```

PR #4307, opened 2022-12-07
Package the skipped tests file to be consumed in test stage

abhay-krishna:
The test stage does not have a source, so we need to get the file from the previous stage as an artifact.


pokearu (approved):
/woof

PR #4316, opened 2022-12-07
Refactor e2e framework to unify eks-a API updates

g-gaston:
## Description of changes

How we update the cluster config has changed over time, resulting in duplicated code by providers and some logic that is difficult to reason about. This doesn’t completely solve it but it’s a step in a better direction.

First, stop relying on storing the yaml content of the manifest and just use a cluster.Config. We will follow up with more changes, simplifying how tests update the cluster config, moving towards using `test.UpdateClusterConfig`. However, this requires changing the interface to the framework and hence the tests themselves. We will do that separately and piece-meal.

This allows to decouple updating the cluster config from running an 'cluster upgrade` command. This is extremely useful when testing the full lifecycle API from both kubectl and GitOps.

*Testing (if applicable):*
I ran a few vSphere tests manually, it's possible that this breaks other tests if they are making hidden assumptions about the e2e test object's internals.


tatlat:
https://github.com/aws/eks-anywhere/blob/7b473655d415d45f5376f77f08c7d0ffccc85141/internal/pkg/api/nutanix.go#L41-L42
Why does nutanix still need the parse from file method while other providers do not? Is this here to avoid breaking tests?

cxbrowne1207 (commented):
/lgtm

Nothing aside from what's already been said. How bad would it be if this is merged and  tests start to break after the cutover?

internal/pkg/api/tinkerbell.go
```diff
@@ -1,96 +1,56 @@
...
-	"sigs.k8s.io/yaml"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/providers"
-	"github.com/aws/eks-anywhere/pkg/templater"
 )
 
 type TinkerbellConfig struct {
-	clusterConfig    *anywherev1.Cluster
```
ddjjia, line 16: why we have this config change on tinkerbell?

g-gaston: This field wasn't needed and removing it simplified the code

g-gaston:
just squashing

g-gaston:
Fixing the lint for comments

PR #4333, opened 2022-12-08
Update ADOT e2e log comparison logic

ivyostosh:

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -1212,7 +1212,7 @@ func (e *ClusterE2ETest) VerifyAdotPackageDeploymentUpdated(packageName string, 
 
 	// Deploy ADOT as a deployment and scrape the apiservers
 	e.T.Log("Apply changes to package", packageName)
-	e.T.Log("This will update ", packageName, " to be a deployment, and scrape the apiservers")
+	e.T.Log("This will update", packageName, "to be a deployment, and scrape the apiservers")
 	err := e.KubectlClient.ApplyKubeSpecFromBytesWithNamespace(ctx, e.cluster(), adotPackageDeployment, packageMetadatNamespace)
 	if err != nil {
 		e.T.Fatalf("Error upgrading adot package: %s", err)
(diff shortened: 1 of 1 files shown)


test/framework/cluster.go
```diff
@@ -1244,10 +1244,11 @@ func (e *ClusterE2ETest) VerifyAdotPackageDeploymentUpdated(packageName string,
 		e.T.Fatalf("failure getting pod logs %s", err)
 	}
 	fmt.Printf("Logs from aws-otel-collector pod\n %s\n", logs)
-	expectedLogs := "MetricsExporter	{\"kind\": \"exporter\", \"data_type\": \"metrics\", \"name\": \"logging\", \"#metrics\": 146}"
-	ok := strings.Contains(logs, expectedLogs)
+	expectedLogs1 := "MetricsExporter	{\"kind\": \"exporter\", \"data_type\": \"metrics\", \"name\": \"logging\", \"#metrics\": 145}"
+	expectedLogs2 := "MetricsExporter	{\"kind\": \"exporter\", \"data_type\": \"metrics\", \"name\": \"logging\", \"#metrics\": 146}"
+	ok := strings.Contains(logs, expectedLogs1) || strings.Contains(logs, expectedLogs2)
```
junshun, line 1249: Where does the 145 or 146 come from? Is that something that can change in the future?

test/framework/cluster.go
```diff
@@ -1290,10 +1291,11 @@ func (e *ClusterE2ETest) VerifyAdotPackageDaemonSetUpdated(packageName string, t
 		e.T.Fatalf("failure getting pod logs %s", err)
 	}
 	fmt.Printf("Logs from aws-otel-collector pod\n %s\n", logs)
-	expectedLogs := "MetricsExporter	{\"kind\": \"exporter\", \"data_type\": \"metrics\", \"name\": \"logging\", \"#metrics\": 95}"
-	ok := strings.Contains(logs, expectedLogs)
+	expectedLogs1 := "MetricsExporter	{\"kind\": \"exporter\", \"data_type\": \"metrics\", \"name\": \"logging\", \"#metrics\": 94}"
```
junshun, line 1294: Same here why is it sometimes 95 vs 96?

test/framework/cluster.go
ivyostosh: Those are the expected metrics number for api-server. Although after further testing, I noted those numbers are not as predictable as what I expected them to be, and depends on the sequence of how metrics are scrapped, this could fail. So I updated the logic to not rely on this numeric number check anymore. Please refer to the new commit.

I'm still confident that with this new update (w/o specifying the metrics #), it still completes the functional test. That's because right before reading the pod log, we switch the deployment / daemonset mode, which forces the pods to re-load (and there are checks added to make sure this is happening), and flushes out the old logs.

test/framework/cluster.go
ivyostosh: Same comment applies here

PR #4347, opened 2022-12-09
Update tag to consume bug fix for clusterresourceset

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Consume the tag that points to https://github.com/abhay-krishna/cluster-api/pull/6

(This got resolved, was an upstream Go issue) In order to get `go mod tidy` to work, I also needed to set `GOPRIVATE=github.com/abhay-krishna/cluster-api`

*Testing (if applicable):*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -172,7 +172,7 @@ require (
 	github.com/vmware/govmomi v0.29.0
 	github.com/xanzy/ssh-agent v0.3.0 // indirect
 	go.uber.org/atomic v1.10.0 // indirect
-	go.uber.org/multierr v1.8.0
+	go.uber.org/multierr v1.8.0 // indirect
 	golang.org/x/net v0.0.0-20220812174116-3211cb980234 // indirect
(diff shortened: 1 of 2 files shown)


vivek-koppuru:
/approve
/release-0.13

vivek-koppuru:
Looks like the `GOPRIVATE` export I did locally might be needed for athens too, but not sure if there is just a propagation delay until the tag gets ready.

vivek-koppuru:
I see someone else has filed a ticket for a somewhat similar issue, curious what they are going to respond with: https://github.com/golang/go/issues/57185

g-gaston:
/lgtm
However, if the module is giving problems, we don't need this to be merged just yet, so no rush. The change was only in the crs code, which the cli doesn't depend on directly

vivek-koppuru:
Seems to have been an issue on the golang upstream servers side. Looks like the issue is resolved now

PR #4352, opened 2022-12-09
Fix ADOT e2e daemonset not found error

ivyostosh:
Occasionally, some tests will fail due to an error daemonsets.apps "generated-adot-aws-otel-collector-agent" not found. It appears that after the package changed to the "installed" state, some time buffer is still needed in between before we can check the daemonset status. The time needed may vary by machine. We will start with 15s.


TerryHowe (approved):
/lgtm
/approve

PR #4377, opened 2022-12-12
update defaultImageRegistry with the region specified through EKSA_AW…

vincentni:
Honor `EKSA_AWS_REGION` and update `dafaultImageRegistry` accordingly.

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -106,7 +106,8 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
 		}
 	} else {
 		sourceRegistry := fmt.Sprintf("sourceRegistry=%s", GetRegistry(pc.chart.Image()))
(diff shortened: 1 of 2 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -106,7 +106,8 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
 		}
 	} else {
 		sourceRegistry := fmt.Sprintf("sourceRegistry=%s", GetRegistry(pc.chart.Image()))
-		values = []string{sourceRegistry, clusterName}
+		defaultImageRegistry := fmt.Sprintf("defaultImageRegistry=%s", strings.ReplaceAll(constants.DefaultCuratedPackagesRegistryRegex, "*", pc.eksaRegion))
```
TerryHowe, line 109: Might want to only do this if the region is set

TerryHowe (commented):
Thanks doing this.

TerryHowe (approved):
/approve

Issue #4237, opened 2022-11-30
Allow users to configure VMX swap settings


abhinavmpandey08:
**What would you like to be added**:
Allow users to enable/disable memory swap settings on VM level `sched.swap.vmxSwapEnabled=FALSE`

Looks like this can be configured using the V clone spec object https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/969a4ae80bdac9e4636d5a4dcd5b44af5d2a3333/apis/v1beta1/types.go#L161

**Why is this needed**:
Users may want to control enabling/disabling swap memory as it has resource and performance implications.

PR #4383, opened 2022-12-13
Design: vSphere VMX Settings

abhinavmpandey08:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/4237

*Description of changes:*
Rendered doc: https://github.com/abhinavmpandey08/eks-anywhere/blob/vsphere-custom-vmx-design/designs/vsphere-advanced-vmx-settings.md

### DIFF
--- a/designs/vsphere-advanced-vmx-settings.md
+++ b/designs/vsphere-advanced-vmx-settings.md
@@ -1,0 +1,153 @@
+# Support advanced Virtual Machine Configuration (VMX) settings on vSphere node VMs 
+
+## Introduction
+
+**Problem:** 
+
+vSphere allows users to configure [advanced VMX settings](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenter.configuration.doc/GUID-62184858-32E6-4DAC-A700-2C80667C3558.html) for individual VMs. However, EKS Anywhere on vSphere currently doesn't provide a way for users to modify these settings.
+
+One of the use-cases for exposing this is to allow users to [disable swap space](https://github.com/aws/eks-anywhere/issues/4237) on VMs by adding this VMX config `sched.swap.vmxSwapEnabled=FALSE`.
+
+### Goals and Objectives
+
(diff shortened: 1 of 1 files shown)


designs/vsphere-advanced-vmx-settings.md
```diff
@@ -0,0 +1,153 @@
...
+
+Pros:
+
+* Users won't have to worry about knowing the exact VMX keys and values for a specific setting like `"sched.swap.vmxSwapEnabled": "FALSE"`. Instead, they can configure it using user-friendly fields like `enableSwap: false`.
+* It makes validations a lot easier since EKS-A will know what VMX options and values to expect.
+
+Cons: 
+
+* Limits the number of options users will be allowed to configure since 
+* Adding new options will require code changes.
```
markapruett, line 114: Seems like this should be one line?

drewvanstone:
I think Option 2 is the best option for the customer, but we may want to do whatever we can to make any untested configurations easier to debug.

designs/vsphere-advanced-vmx-settings.md
```diff
@@ -0,0 +1,153 @@
...
+```yaml
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: VSphereMachineConfig
+metadata:
+   name: my-cluster-machines
+spec:
+  diskGiB: 25
+  datastore: "Datastore"
+  folder: "Folder"
+  vmxOptions:
```
jacobweinstock, line 101: I would recommend we put this under a field name something like `advanced` or `experimental`, etc. This will allow communication in the code that these fields are non standard, or not validated, or etc. We can also toggle all fields under this to allow for advanced use-cases and the like.

abhinavmpandey08:
Based on the discussions, we have decided to go with `Option 1` where we only expose certain options that we have tested and validated

abhinavmpandey08:
Closing this design since we aren't planning on implementing this feature right now.
We can revisit it once it's prioritized again.

/close

PR #4387, opened 2022-12-13
Adjust ADOT e2e sleep time while waiting on daemonset

ivyostosh:
15s wait time appears to be not sufficient in some cases, mainly Nutanix and vSphere Bottle Rocket (about 75% pass rate). Increase sleep time to 30s.

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -1263,10 +1263,11 @@ func (e *ClusterE2ETest) VerifyAdotPackageDaemonSetUpdated(packageName string, t
 		e.T.Fatalf("waiting for adot package update timed out: %s", err)
 	}
 
-	time.Sleep(30 * time.Second) // Add sleep to allow daemonset to be created
 	e.T.Log("Waiting for package", packageName, "daemonset to be rolled out")
-	err = e.KubectlClient.WaitForDaemonsetRolledout(ctx,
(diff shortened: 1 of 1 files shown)


test/framework/cluster.go
```diff
@@ -1263,7 +1263,7 @@ func (e *ClusterE2ETest) VerifyAdotPackageDaemonSetUpdated(packageName string, t
 		e.T.Fatalf("waiting for adot package update timed out: %s", err)
 	}
 
-	time.Sleep(15 * time.Second) // Add sleep to allow daemonset to be created
+	time.Sleep(30 * time.Second) // Add sleep to allow daemonset to be created
```
TerryHowe, line 1266: Something similar to this maybe?
```
    e.T.Log("Waiting for control plane to be ready")
    err := retrier.New(5 * time.Minute).Retry(func() error {
        return e.KubectlClient.ValidateControlPlaneNodes(context.Background(), e.cluster(), e.ClusterName)
    })
    if err != nil {
        e.T.Fatal(err)
    }
```

ivyostosh: Thanks. I've made an update.

I tested this locally and noticed that w/o the sleep the downstream checking log step can have issues too, so added a retrier there too.

TerryHowe (approved):
/lgtm
/approve

PR #4390, opened 2022-12-13
Add run-gci target to make file

TerryHowe:
It can be convenient to have make targets to perform formatting fixes instead of relying on IDE configurations.


ewollesen (commented):
/lgtm

TerryHowe (commented):
/approve

PR #4414, opened 2022-12-14
Add ClusterValidator for Cluster Lifecycle API e2e tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*
<WIP>

### DIFF
--- a/test/framework/clustervalidator.go
+++ b/test/framework/clustervalidator.go
@@ -1,0 +1,105 @@
+package framework
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type clusterValidation = func() error
+type validationOpt struct {
+	timeout  time.Duration
+	validate clusterValidation
+}
+
+// ClusterValidator is responsible for validating a cluster.
+type ClusterValidator struct {
+	clusterConfig *cluster.Config
+	client        client.Client
+	validations   []validationOpt
+}
+
+// NewDefaultClusterValidator returns a cluster validator pre-loaded with validations by default.
+func NewDefaultClusterValidator(client client.Client, config *cluster.Config) ClusterValidator {
+	cv := ClusterValidator{
+		clusterConfig: config,
+		client:        client,
+	}
+
+	cv.withValidation(cv.validateObjects, 5*time.Minute)
+	cv.withValidation(func() error {
+		return cv.validateControlPlaneNodes()
+	}, 5*time.Minute)
+	return cv
+}
+
+func (c *ClusterValidator) validateObjects() error {
+	ctx := context.Background()
+
+	for _, obj := range c.clusterConfig.ChildObjects() {
+		u := &unstructured.Unstructured{}
+		err := c.client.Get(ctx, client.ObjectKeyFromObject(obj), u)
+
+		if err != nil {
+			return fmt.Errorf("cluster object does not exist %s", err)
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


test/framework/clustervalidator.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type clusterValidation = func(client client.Client, config *cluster.Config) error
```
g-gaston, line 16: You might want to consider passing the `cluster.Spec`
That's the one that will also include the `Bundles`

You'll find that to run certain validations, you will need the version of the components installed (eks-d version, cilium version, etc.). These versions are specified in the `Bundles`, so with `cluster.Config` you don't have them but with `cluster.Spec` you do. 

The other option would be to just pass both the `cluster.Config` and the `Bundles` (or probably better, `VersionsBundle`, since that once is already specific for one kubernetes version).

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,92 @@
...
+
+type validationOpt struct {
+	timeout  time.Duration
```
g-gaston, line 19: We might wanna also configure the wait time?

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,92 @@
...
+
+// NewDefaultClusterValidator returns a cluster validator pre-loaded with validations by default.
+func NewDefaultClusterValidator(client client.Client, config *cluster.Config) ClusterValidator {
+	cv := ClusterValidator{
+		clusterConfig: config,
+		client:        client,
+	}
+
+	cv.withValidation(validateObjects, 5*time.Minute)
+	cv.withValidation(validateControlPlaneNodes, 5*time.Minute)
```
g-gaston, line 61: what's the purpose of having a system to dynamically register validations if we hardcode the validations in the constructor?

I would vote to decouple both, having a constructor that just returns the validator and allowing to add validations later. And if you want, we can have a wrapper that init and register the default validations.

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (c *ClusterValidator) withValidation(validate clusterValidation, timeout time.Duration) {
+	c.validations = append(c.validations, validationOpt{
+		timeout:  timeout,
+		validate: validate,
+	})
+}
+
+// ValidateCluster runs through the set clusterValidations returns an error if any of them fail after a number of retries.
+func (c *ClusterValidator) ValidateCluster() error {
```
g-gaston, line 38:
```suggestion
func (c *ClusterValidator) Validate(ctx context.Context) error {
```

test/framework/clustervalidator.go
g-gaston, line 16:
```suggestion
type clusterValidation = func(ctx context.Context, client client.Client, config *cluster.Config) error
```

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type clusterValidation = func(client client.Client, config *cluster.Config) error
+
+type validationOpt struct {
```
g-gaston, line 18:
```suggestion
type retriableValidation struct {
```

test/framework/clustervalidator.go
cxbrowne1207: Ah, thanks for the heads up!

test/framework/clustervalidator.go
cxbrowne1207: My thought was that we could have a constructor that returns a validator without validations as well, in addition to one that's already pre-loaded. I can see the confusion though, I think having a wrapper that inits and registers defaults validations is where I'll take this.

Issue #4109, opened 2022-11-16
import images should push images in OCI format


TerryHowe:
Currently import images pushes image in docker v2s2 format and it should push in oci format. The change in format means there is a mismatch between ECR shasums and the private registry shasum

PR #4420, opened 2022-12-15
Add Copy packages command for registry to registry image and chart copy

TerryHowe:
New command that takes advantage of oras copy command to copy images from one registry to another. The primary use case for this would be to copy images from ECR to Harbor or other private registry.
* supports multiple architecture image copy duplicating arm and amd
* preserves shasum of images from ECR to harbor so we can use the shasum for installation


TerryHowe:
Closes: https://github.com/aws/eks-anywhere/issues/4109

pkg/registry/cache_test.go
```diff
@@ -0,0 +1,35 @@
+package registry
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
g-gaston, line 6: Could we use gomega? That's the preferred test library in this repo

pkg/registry/credentials.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// CredentialStore for registry credentials such as ~/.docker/config.json.
+type CredentialStore struct {
+	directory  string
+	configFile *configfile.ConfigFile
+}
+
+// NewCredentialStore create a create store.
+func NewCredentialStore(args ...string) CredentialStore {
```
g-gaston, line 17: nit: doc comment about how to use args to define a directory?

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+// OCIRegistryClient storage client for an OCI registry.
+type OCIRegistryClient struct {
+	Host        string
+	Project     string
+	CertFile    string
+	Insecure    bool
+	DryRun      bool
```
g-gaston, line 28: do these need to be exported if they can be set through the constructor?

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+		OI:       &OrasImplementation{},
+	}
+}
+
+// Init registry configuration.
+func (or *OCIRegistryClient) Init() error {
+	if or.initialized {
+		return nil
+	}
+	or.initialized = true
```
g-gaston, line 51: nit: set this at the end of the method? That way if there is an error, we don't mark it as initialized.

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
+package registry
```
g-gaston, line 1: maybe we can name this file `client.go`?

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return fmt.Errorf("registry copy destination: %v", err)
+	}
+
+	dstStorage, err := dstClient.GetStorage(ctx, image)
+	if err != nil {
+		return fmt.Errorf("registry copy destination: %v", err)
+	}
+
+	fmt.Println(dstClient.Destination(image))
```
g-gaston, line 154: could we use some kind of logger for this?

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+	dstRepo := or.Project + image.Repository()
+	repo, err = or.OI.Repository(ctx, or.registry, dstRepo)
+	if err != nil {
+		return nil, fmt.Errorf("error creating repository %s: %v", dstRepo, err)
+	}
+	return repo, nil
+}
+
+// Copy a image from a source to a destination.
+func (or *OCIRegistryClient) Copy(ctx context.Context, image releasev1.Image, dstClient StorageClient) (err error) {
```
g-gaston, line 125: could we make the exported interface less coupled to the release package?
Maybe with something like `type Artifact releasev1.Image`
And then we can later decide to define our own struct here if necessary.

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+	committed := &sync.Map{}
```
g-gaston, line 126: what is this being used for? it seems like we insert entries but never read it

pkg/registry/storage.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	orasregistry "oras.land/oras-go/v2/registry"
+
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+// StorageClient interface for general image storage client.
+type StorageClient interface {
```
g-gaston, line 12: could we just define this where it's used? or do you envision this being needed as a shared interface for something else?

release/api/v1alpha1/artifact_types.go
```diff
@@ -73,6 +73,38 @@ func (i Image) ChartName() string {
...
+		result = strings.Split(i.URI, ":")
+		if len(result) < 2 {
+			return ""
+		}
+		return strings.TrimPrefix(result[0], i.Registry()+"/")
+	}
+	return strings.TrimPrefix(result[0], i.Registry()+"/")
+}
+
+func (i *Image) Version() string {
```
g-gaston, line 96: tests for these 3?

pkg/registry/ociregistry.go
```diff
@@ -0,0 +1,164 @@
...
+
+var _ StorageClient = (*OCIRegistryClient)(nil)
+
+// NewOCIRegistry create an OCI registry client.
+func NewOCIRegistry(host string, certFile string, insecure bool) *OCIRegistryClient {
+	return &OCIRegistryClient{
+		Host:     host,
+		CertFile: certFile,
+		Insecure: insecure,
+		OI:       &OrasImplementation{},
```
g-gaston, line 42: why not inject this as a dependency?

pkg/registry/ociregistry_test.go
```diff
@@ -0,0 +1,183 @@
+package registry
```
g-gaston, line 1: could we move all tests to `registry_test`

pkg/registry/orasinterface.go
```diff
@@ -0,0 +1,39 @@
...
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content"
+	orasregistry "oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+// OrasInterface thin layer for oras.
+type OrasInterface interface {
```
g-gaston, line 14:
```suggestion
type Oras interface {
```

since the type already says it's an interface and this what we expose to the outside caller

pkg/registry/orasinterface.go
```diff
@@ -0,0 +1,39 @@
...
+
+var _ OrasInterface = (*OrasImplementation)(nil)
+
+// Repository for the given registry.
+func (oi *OrasImplementation) Repository(ctx context.Context, reg *remote.Registry, name string) (orasregistry.Repository, error) {
+	result, err := reg.Repository(ctx, name)
+	return result, err
+}
+
+// Resolve call oras Resolve.
```
g-gaston, line 31: could we explain a bit what this does?

pkg/registry/orasinterface.go
```diff
@@ -0,0 +1,39 @@
...
+
+// OrasInterface thin layer for oras.
+type OrasInterface interface {
+	Repository(ctx context.Context, reg *remote.Registry, name string) (orasregistry.Repository, error)
+	Resolve(ctx context.Context, srcStorage orasregistry.Repository, reference string) (ocispec.Descriptor, error)
+	CopyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, root ocispec.Descriptor, opts oras.CopyGraphOptions) error
+}
+
+// OrasImplementation thin wrapper on oras.
+type OrasImplementation struct{}
```
g-gaston, line 21: curious why you decided to create this layer on top of oras.
I was thinking instead we could just define the interface with same same signature as the oras original methods
And then compose those original oras objects into one to satisfy that interface

like
```go
type OrasInterface interface {
    Repository(ctx context.Context, name string) (orasregistry.Repository, error)
    Resolve(ctx context.Context, reference string) (ocispec.Descriptor, error)
    CopyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, root ocispec.Descriptor, opts oras.CopyGraphOptions) error
}
```

Although it's still coupled to the oras module, since it uses its structs as input and output, it still can be implemented in different ways since it's not dependent on the "service" objects (`remote.Registry`, `rasregistry.Repository`, etc.).

pkg/registry/cache_test.go
TerryHowe: I thought we had a discussion that everyone hated gomega

pkg/registry/orasinterface.go
TerryHowe: I only did this to get enough test coverage.

go.mod
```diff
@@ -153,7 +155,7 @@ require (
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
-	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799 // indirect
+	github.com/opencontainers/image-spec v1.1.0-rc2
```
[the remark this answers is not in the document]
TerryHowe: Done.

pkg/registry/storage.go
TerryHowe: I was planning on using this interface for a file storage client, not just a registry client. The file client would read/write from a directory or tarball.

TerryHowe: The file storage client already exists in oras, but it is not completely implemented. They are planning on finishing it up in January.

release/api/v1alpha1/artifact_types.go
TerryHowe: Good, call, done.

release/api/v1alpha1/artifact_types_test.go
```diff
@@ -121,3 +121,140 @@ func TestImageTag(t *testing.T) {
 		})
 	}
 }
+
```
TerryHowe, line 124: I loathe table tests, so I am willing to convert these, but this is the style used here.

pkg/registry/credentials.go
TerryHowe: I just added a SetDirectory method instead.

pkg/registry/ociregistry.go
TerryHowe: Done.

pkg/registry/ociregistry.go
TerryHowe: Made them private

pkg/registry/ociregistry.go
TerryHowe: That would ruin SoC

pkg/registry/ociregistry.go
TerryHowe: Done.

pkg/registry/ociregistry.go
TerryHowe: Removed.

pkg/registry/ociregistry.go
TerryHowe: Changed to use "log" 🤷

pkg/registry/ociregistry_test.go
TerryHowe: Done.

cmd/eksctl-anywhere/cmd/copypackages.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/manifests/bundles"
+	"github.com/aws/eks-anywhere/pkg/registry"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+// copyPackagesCmd is the context for the copy packages command.
+var copyPackagesCmd = &cobra.Command{
+	Use:          "packages",
+	Short:        "Copy package images and charts from a source to a destination",
```
ewollesen, line 20:
```suggestion
	Short:        "Copy curated package images and charts from a source to a destination",
```
I think the added "curated" might be worthwhile here, since the term "package" is so generic.

ewollesen: The same might go for the `Long` member. Up to you. 😄

cmd/eksctl-anywhere/cmd/copypackages.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func init() {
+	copyCmd.AddCommand(copyPackagesCmd)
+
+	copyPackagesCmd.Flags().StringVarP(&copyPackagesCommand.bundleFile, "bundle", "b", "", "EKS-A bundle file to read artifact dependencies from")
+	if err := copyPackagesCmd.MarkFlagRequired("bundle"); err != nil {
+		log.Fatalf("Cannot mark 'bundle' flag as required: %s", err)
+	}
+	copyPackagesCmd.Flags().StringVarP(&copyPackagesCommand.dstCert, "dst-cert", "", "", "Package bundles file to read artifact dependencies from")
```
ewollesen, line 39: Looks like some sort of cut and paste errors? Both here on line 39 and on line 40?

pkg/registry/cache.go
```diff
@@ -0,0 +1,29 @@
...
+
+import "fmt"
+
+// Cache storage client for an OCI registry.
+type Cache struct {
+	registries map[string]StorageClient
+}
+
+// NewCache creates an OCI registry client.
+func NewCache() *Cache {
```
ewollesen, line 11: How useful is this cache, given that this is used via the CLI and won't be long-lived?

I imagine we don't care, but this too isn't thread-safe... But unless we decide to parallelize the for loop that's reading from this cache, I suppose it's not important.

pkg/registry/cache_test.go
```diff
@@ -0,0 +1,37 @@
...
+	result, err := cache.Get("localhost", "", false)
+
+	assert.NoError(t, err)
+	ociRegistry, ok := result.(*registry.OCIRegistryClient)
+	assert.True(t, ok)
+	assert.Equal(t, "localhost", ociRegistry.GetHost())
+
+	result, err = cache.Get("localhost", "", false)
+
+	assert.NoError(t, err)
```
ewollesen, line 23:
```suggestion
	result, err = cache.Get("localhost", "", false)
	assert.NoError(t, err)	
```
For improved readability, assuming you agree.

pkg/registry/cache_test.go
```diff
@@ -0,0 +1,37 @@
...
+	sameOciRegistry, ok := result.(*registry.OCIRegistryClient)
+	assert.True(t, ok)
+	assert.Equal(t, "localhost", sameOciRegistry.GetHost())
+	assert.Equal(t, "", sameOciRegistry.GetCertFile())
+	assert.False(t, sameOciRegistry.IsInsecure())
+	assert.Equal(t, ociRegistry, sameOciRegistry)
+
+	result, err = cache.Get("example.com", "bogus.file", true)
+
+	assert.EqualError(t, err, "error with repository example.com: error reading certificate file <bogus.file>: open bogus.file: no such file or directory")
```
ewollesen, line 33:
```suggestion
	result, err = cache.Get("example.com", "bogus.file", true)
	assert.EqualError(t, err, "error with repository example.com: error reading certificate file <bogus.file>: open bogus.file: no such file or directory")
```

pkg/registry/client.go
```diff
@@ -0,0 +1,177 @@
...
+	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+// OCIRegistryClient storage client for an OCI registry.
+type OCIRegistryClient struct {
+	host        string
+	project     string
+	certFile    string
+	insecure    bool
+	dryRun      bool
```
ewollesen, line 26: Maybe an idea for the future, but if there were a separate implementation, `DryRunRegistryClient` or the like, then the code doesn't need to branch on the dry run flag (i.e. it's simpler), and is therefore easier to read, and less prone to someone adding a method implementation and forgetting to check the value of the dry run flag and abort before making changes.

pkg/registry/client.go
```diff
@@ -0,0 +1,177 @@
...
+		host:     host,
+		certFile: certFile,
+		insecure: insecure,
+		OI:       &OrasImplementation{},
+	}
+}
+
+// Init registry configuration.
+func (or *OCIRegistryClient) Init() error {
+	if or.initialized {
```
ewollesen, line 46: I guess there's not much to worry about, but this is a race condition... Consider a `sync.Mutex` or a `sync.Once` if it's really important that this code only be run once per instantiation. I don't see anything here that isn't safe to run more than once?

pkg/registry/client.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil
+	}
+
+	credentialStore := NewCredentialStore()
```
ewollesen, line 50: For the future, it's typically a better pattern to build these in your main, and pass them to a constructor. That makes your code more testable (you can easily pass in a test implementation), and it corrals the possible Init() errors to the main method where they can more easily be dealt with.

pkg/registry/client.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// SetDryRun dry run validates read, but not write.
+func (or *OCIRegistryClient) SetDryRun(value bool) {
+	or.dryRun = value
+}
+
+// Destination of this storage registry.
+func (or *OCIRegistryClient) Destination(image Artifact) string {
+	return strings.Replace(image.VersionedImage(), image.Registry+"/", or.host+"/"+or.project, 1)
```
ewollesen, line 119:
```suggestion
	return strings.Replace(image.VersionedImage(), url.JoinPath(url.PathEscape(image.Registry), url.PathEscape(or.host), url.PathEscape(or.project)), 1)
```
It seems a little like overkill, but IIRC, some of these values were supplied by the user, and so should be considered as possibly malicious.

cmd/eksctl-anywhere/cmd/copypackages.go
```diff
@@ -0,0 +1,120 @@
...
+	Use:          "packages",
+	Short:        "Copy package images and charts from a source to a destination",
+	Long:         `Copy all the EKS Anywhere package images and helm charts from a source to a destination.`,
+	SilenceUsage: true,
+	RunE:         runCopyPackages,
+	Args: func(cmd *cobra.Command, args []string) error {
+		if err := cobra.ExactArgs(1)(cmd, args); err == nil {
+			return nil
+		}
+		return fmt.Errorf("A destination must be specified as an argument")
```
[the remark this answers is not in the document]
TerryHowe: I think this was just copy/pasta

pkg/registry/cache.go
TerryHowe: There is no intention of parallelizing this for now, this is for convenience so we we can mindless loop over the images to copy and reuse a client. The normal case, there are two sources and one destination.

TerryHowe: oras paralielizes blob (aka image layer) copy under the covers

pkg/registry/cache_test.go
TerryHowe: Hmm, this was spaced this way for given/when/then, but this isn't a great example of that. Let me think about it.

pkg/registry/client.go
TerryHowe: That was my original intention, but it ended up having a lot of repeated code. It was a complete copy of the client minus the actual call to CopyGraph

pkg/registry/cache_test.go
TerryHowe: Yeh, I think no space makes sense since this test is covering 3 different test cases. Fixed.

pkg/registry/storage.go
g-gaston: I see. I don't fully understand it but I'm cool moving forward with this and we can revisit if necessary when we add the file storage client.

TerryHowe: The file storage client would be for download/import so we could have those in OCI format as well, but future thing.

pkg/registry/client.go
```diff
@@ -0,0 +1,146 @@
...
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	orasregistry "oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+// Context describes aspects of a registry.
+type Context struct {
```
g-gaston, line 20: I personally find a bit confusing to name this context given how common `context.Context` is in Go and that this one doesn't implement `context.Context`.

What do you think about about `RegistryConfig`?

pkg/registry/orasinterface.go
g-gaston: Ah got you. Then if it doesn't require too much work, I would to change this to something similar to what I suggested. I believe it should allow us to reach the same amount of coverage.

pkg/registry/client.go
TerryHowe: sounds good to me.

pkg/registry/orasinterface.go
TerryHowe: Yeh, compose isn't going to work, but I was going to rearrange the code a bit to get the coverage back up AND get rid of this OrasImplementation

pkg/registry/client.go
TerryHowe: lint doesn't like RegistryContext, so I'll go with something else like maybe StorageContext

g-gaston (approved):
/lgtm
/unapprove
Not sure if you were gonna change the tests to use gomega, so un-approving jic. Feel free to approve it again if not.

jonahjon:
/lgtm

Good luck with those tests

TerryHowe:
/override eks-anywhere-release-tooling-test-presubmit

this is failing because of the golden file being out of date again. I just merged a new one this morning and already it is out of date.

PR #4451, opened 2022-12-16
Enrich registry mirror struct with auth and cert

vincentni:
Add cert and auth info to registrymirror struct and create a registry mirror secret for curated package


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -123,26 +126,54 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	if err := pc.chartInstaller.InstallChart(ctx, pc.chart.Name, ociURI, pc.chart.Tag(), pc.kubeConfig, "", values); err != nil {
 		return err
 	}
 
 	// Customers are currently requesting to show a warning in case
 	// credentials are not provided for curated packages
-	if err = pc.CreateCredentials(ctx); err != nil {
+	if err := pc.CreateAWSCredentials(ctx); err != nil {
 		logger.MarkWarning("  Unable to create credentials for curated packages: ", "warning", err)
 	}
```
TerryHowe, line 138: For the AWS creds are we *not* creating them if the env vars are not set? It seems like it would be easier for the customer if an empty secret was created and they could configure it later if they wanted to. I think that idea was shot down before, but I still hold that opinion.

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -399,6 +405,10 @@ func TestEnableCuratedPackagesFailNoActiveBundle(t *testing.T) {
 }
 
 func TestEnableCuratedPackagesSuccessWhenApplySecretFails(t *testing.T) {
+	os.Setenv("REGISTRY_USERNAME", "admin")
+	os.Setenv("REGISTRY_PASSWORD", "Harbor12345")
+	defer os.Unsetenv("REGISTRY_USERNAME")
+	defer os.Unsetenv("REGISTRY_PASSWORD")
```
TerryHowe, line 411: There is a t.Setenv method as I recall that will only be used for the context of this test and no need to defer.

pkg/registrymirror/registrymirror.go
```diff
@@ -69,6 +78,15 @@ func (r *RegistryMirror) CuratedPackagesMirror() string {
 	return r.NamespacedRegistryMap[constants.DefaultCuratedPackagesRegistryRegex]
 }
 
+// Credentials returns the credential for the registry mirror.
+func (r *RegistryMirror) Credentials() (string, string, error) {
+	username, password, err := config.ReadCredentials()
```
abhinavmpandey08, line 83: Can you move the logic from `config.ReadCredentials()` to here and remove that function?

vincentni: What about other places that call `config.ReadCredentials()`? They now have to get notion of `RegistryMirror` in order to call `Credentials()`. Any thoughts?

pkg/curatedpackages/packagecontrollerclient.go
vincentni: quotes are added to the aws-secret template.

pkg/registrymirror/registrymirror.go
abhinavmpandey08: Can we just make this since it doesn't really need the RegistryMirror object?
It's just reading env and return the values
```
func Credentials() (string, string, error) {
```

pkg/dependencies/factory.go
```diff
@@ -253,7 +253,7 @@ func (f *Factory) UseExecutablesDockerClient(client executables.DockerClient) *F
 
 // dockerLogin performs a docker login with the ENV VARS.
 func dockerLogin(ctx context.Context, registry string, docker executables.DockerClient) error {
-	username, password, _ := config.ReadCredentials()
+	username, password, _ := registrymirror.Credentials()
```
chrisdoherty4, line 256: Any particular reason the error isn't being checked?

chrisdoherty4, line 256: Whats the rationale for using the `registrymirror` pkg for reading the credentials? The credentials read from that func seem like generic registry creds as opposed to being specific to a mirror.

vincentni: As per our doc says (https://anywhere.eks.amazonaws.com/docs/reference/clusterspec/optional/registrymirror/), `REGISTRY_USERNAME` and `REGISTRY_PASSWORD` are used for registry mirror authentication. Given that context, it makes more sense to me to move that to `registrymirror` pkg.

vincentni: If either of them is not set, the error gets caught immediately by the next line `err := docker.Login(ctx, registry, username, password)`.

pkg/registrymirror/registrymirror.go
```diff
@@ -19,6 +21,12 @@ type RegistryMirror struct {
 	NamespacedRegistryMap map[string]string
 	// Auth should be marked as true if authentication is required for the registry mirror
 	Auth bool
+	// CACertContent defines the contents registry mirror CA certificate
+	CACertContent string
+	// InsecureSkipVerify skips the registry certificate verification.
+	// Only use this solution for isolated testing or in a tightly controlled, air-gapped environment.
+	// Currently only supported for snow provider
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
```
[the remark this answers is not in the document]
vincentni: good catch

pkg/dependencies/factory.go
chrisdoherty4: Yea I don't see a reason not to check the error; it gives an opportunity to create a much clearer error.

chrisdoherty4: Thanks Vincent for the clarification. As Guillermo says though, its still good practice to isolate reading configuration because it means we can change that source without touching lots of packages.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -123,26 +126,54 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	if err != nil {
+		return err
+	}
+	templateValues := map[string]interface{}{
+		"mirrorUsername":      username,
+		"mirrorPassword":      password,
+		"mirrorCACertContent": strings.Split(pc.registryMirror.CACertContent, "\n"),
+	}
+	if err := pc.ApplySecret(ctx, registrySecretYaml, templateValues); err != nil {
+		return errors.New("unable to detect registry mirror credentials for curated packages. Artifacts pull may fail due to unauthorized access to the registry mirror")
```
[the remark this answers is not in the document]
chrisdoherty4: We should also avoid insinuating what may or may not happen in the future in error messages. Just let the program error naturally.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -123,26 +126,54 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	username, password, err := registrymirror.Credentials()
+	if err != nil {
+		return err
+	}
+	templateValues := map[string]interface{}{
+		"mirrorUsername":      username,
+		"mirrorPassword":      password,
+		"mirrorCACertContent": strings.Split(pc.registryMirror.CACertContent, "\n"),
+	}
+	if err := pc.ApplySecret(ctx, registrySecretYaml, templateValues); err != nil {
```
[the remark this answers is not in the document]
vincentni: it has been moved to the helm chart.

pkg/curatedpackages/packagecontrollerclient.go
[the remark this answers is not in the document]
vincentni: Since the secret creation has been moved to the curated packages helm chart, this issue gets resolved automatically.

pkg/executables/helm.go
```diff
@@ -123,14 +123,18 @@ func (h *Helm) InstallChartFromName(ctx context.Context, ociURI, kubeConfig, nam
...
+func (h *Helm) InstallChart(ctx context.Context, chart, ociURI, version, kubeconfigFilePath, namespace, valueFilePath string, values []string) error {
 	valueArgs := GetHelmValueArgs(values)
 	params := []string{"install", chart, ociURI, "--version", version}
 	params = append(params, valueArgs...)
 	params = append(params, "--kubeconfig", kubeconfigFilePath)
 	if len(namespace) > 0 {
 		params = append(params, "--create-namespace", "--namespace", namespace)
 	}
+	if valueFilePath != "" {
+		params = append(params, "-f", valueFilePath)
```
g-gaston, line 136: maybe add one simple test with `valueFilePath`?

pkg/providers/vsphere/template.go
```diff
@@ -216,13 +216,13 @@ func buildTemplateMapCP(
...
+		if len(registryMirror.CACertContent) > 0 {
+			values["registryCACert"] = registryMirror.CACertContent
 		}
 
-		if clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.Authenticate {
-			values["registryAuth"] = clusterSpec.Cluster.Spec.RegistryMirrorConfiguration.Authenticate
-			username, password, _ := config.ReadCredentials()
+		if registryMirror.Auth {
+			values["registryAuth"] = registryMirror.Auth
+			username, password, _ := registryMirror.Credentials()
```
g-gaston, line 225: why ignore the error?

pkg/registrymirror/registrymirror.go
```diff
@@ -69,6 +78,15 @@ func (r *RegistryMirror) CuratedPackagesMirror() string {
 	return r.NamespacedRegistryMap[constants.DefaultCuratedPackagesRegistryRegex]
 }
 
+// Credentials returns the credential for the registry mirror.
+func (r *RegistryMirror) Credentials() (string, string, error) {
```
g-gaston, line 82: I'm not sure why we want this method with a `RegistryMirror` receiver. It doesn't depend on the state/dependencies of `RegistryMirror`.

By adding this wrapper, now we have two ways of doing the same thing and it's not obvious they are the same.

Moreover, IMO it increases the scope of `RegistryMirror` to something that is not very well defined. Before, it just wrapped around a registry mirror config and offered some helpers to read such config. How does reading envvars fit in that definition? What if we injected the credentials info when we construct it?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -119,34 +121,66 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	filePath, err := writer.Write(valueFileName, content)
+	if err != nil {
+		return "", err
+	}
+	return filePath, nil
+}
 
-	return nil
+// GenerateHelmOverrideValues generates override values.
+func (pc *PackageControllerClient) GenerateHelmOverrideValues() ([]byte, error) {
```
g-gaston, line 159: Does this need to be exported?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -119,34 +121,66 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
 
-// CreateCredentials creates necessary credentials to enable the installation of curated packages.
-// In order to create the credentials, there are two necessary steps:
-//   - Creating a kubernetes secret
-//   - Creating a single run of the cron job to ensure the secret is consumed
-func (pc *PackageControllerClient) CreateCredentials(ctx context.Context) error {
-	if err := pc.ApplySecret(ctx); err != nil {
-		return errors.New("unable to detect AWS authentication credentials for curated packages. Skipping package installation. Follow documentation to set credentials after cluster creation completes")
+// CreateHelmOverrideValuesYaml creates a temp file to override certain values in package controller helm install.
+func (pc *PackageControllerClient) CreateHelmOverrideValuesYaml(writer filewriter.FileWriter) (string, error) {
```
g-gaston, line 146: Does this need to be exported?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -119,34 +121,66 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
 		noProxy := fmt.Sprintf("proxy.NO_PROXY=%s", strings.Join(pc.noProxy, "\\,"))
 		values = append(values, httpProxy, httpsProxy, noProxy)
 	}
-	if pc.eksaSecretAccessKey == "" || pc.eksaAccessKeyID == "" {
+	if (pc.eksaSecretAccessKey == "" || pc.eksaAccessKeyID == "") && pc.registryMirror == nil {
 		values = append(values, "cronjob.suspend=true")
 	}
 
-	err := pc.chartInstaller.InstallChart(ctx, pc.chart.Name, ociURI, pc.chart.Tag(), pc.kubeConfig, "", values)
+	writer, err := filewriter.NewWriter(pc.clusterName)
```
g-gaston, line 128: could we inject this as a dependency?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -119,34 +121,66 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	defer writer.CleanUpTemp()
```
g-gaston, line 129: I would avoid doing this. When invoked from the a `cluster` workflow, it will delete al the filed in the `generated` folder. We only do this after a successful completion of the whole workflow.

pkg/executables/helm.go
vincentni: added

pkg/curatedpackages/packagecontrollerclient.go
vincentni: By exporting it, it makes the testing of this core functionality easier.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -235,28 +269,6 @@ func (pc *PackageControllerClient) IsInstalled(ctx context.Context) bool {
 	return bool && err == nil
 }
 
-func (pc *PackageControllerClient) ApplySecret(ctx context.Context) error {
```
a-cool-train, line 238: Does this mean we plan on overriding the helm values rather than applying the secret to the cluster?

vincentni: Right, now the secrets are not created by CLI but become part of the helm chart.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -119,34 +123,64 @@ func (pc *PackageControllerClient) EnableCuratedPackages(ctx context.Context) er
...
+	return filePath, content, nil
+}
 
-	return nil
+func (pc *PackageControllerClient) generateHelmOverrideValues() ([]byte, error) {
+	var err error
+	endpoint, username, password, caCertContent := "", "", "", ""
+	if pc.registryMirror != nil {
+		endpoint = pc.registryMirror.BaseRegistry
+		username, password, err = config.ReadCredentials()
```
g-gaston, line 164: ideally we would inject the username/password to the constructor. I'm ok having that as a follow up.

g-gaston (approved):

PR #4476, opened 2022-12-19
Add infrastructure for Cluster Lifecycle API e2e tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR includes some infrastructure needed to write e2e tests for the Cluster Lifecycle API and a vSphere workload cluster e2e test that utilizes it.

*Testing (if applicable):*
Run `TestVSphereMulticlusterWorkloadClusterAPI` e2e test

### DIFF
--- a/pkg/clients/kubernetes/runtimeclient_wb_test.go
+++ b/pkg/clients/kubernetes/runtimeclient_wb_test.go
@@ -1,0 +1,5 @@
+package kubernetes
+
+var NewRuntimeClient = newRuntimeClient
+
+type RestConfigurator = restConfigurator
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -22,13 +22,16 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
 	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
 	"github.com/aws/eks-anywhere/internal/pkg/api"
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
 	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
 	"github.com/aws/eks-anywhere/pkg/git"
@@ -72,6 +75,7 @@ type ClusterE2ETest struct {
 	WithNoPowerActions     bool
 	ClusterName            string
 	ClusterConfig          *cluster.Config
+	clusterValidator       *ClusterValidator
 	Provider               Provider
 	clusterFillers         []api.ClusterFiller
 	KubectlClient          *executables.Kubectl
@@ -541,6 +545,21 @@ func (e *ClusterE2ETest) parseClusterConfigFromDisk() {
 	e.ClusterConfig = config
 }
 
+func (e *ClusterE2ETest) parseClusterConfigWithDefaultsFromDisk() (*v1alpha1.Cluster, error) {
+	fullClusterConfigLocation := filepath.Join(e.ClusterConfigFolder, e.ClusterName+"-eks-a-cluster.yaml")
+	content, err := os.ReadFile(fullClusterConfigLocation)
+	if err != nil {
+		return nil, fmt.Errorf("reading cluster config file: %v", err)
+	}
+
+	parsedCluster := &v1alpha1.Cluster{}
+	if err := yaml.Unmarshal(content, parsedCluster); err != nil {
+		return nil, fmt.Errorf("unable to marshal cluster config file contents %v", err)
+	}
+
+	return parsedCluster, err
+}
+
 // WithClusterConfig generates a base cluster config using the CLI `generate clusterconfig` command
 // and udpates them with the provided fillers. Helpful for defining the initial Cluster config
 // before running a create operation.
@@ -812,6 +831,13 @@ func (e *ClusterE2ETest) cluster() *types.Cluster {
 	return &types.Cluster{
 		Name:           e.ClusterName,
 		KubeconfigFile: e.kubeconfigFilePath(),
+	}
+}
+
+func (e *ClusterE2ETest) managementCluster() *types.Cluster {
+	return &types.Cluster{
+		Name:           e.ClusterConfig.Cluster.ManagedBy(),
+		KubeconfigFile: e.managementKubeconfigFilePath(),
 	}
 }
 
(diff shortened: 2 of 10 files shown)


test/framework/clustervalidator.go
```diff
@@ -0,0 +1,182 @@
...
+
+// WithValidation registers a validation to the ClusterValidator that will be run when Validate is called.
+func (c *ClusterValidator) WithValidation(validation clusterValidation, backoffPeriod time.Duration, maxRetries int) {
+	c.validations = append(c.validations, retriableValidation{validation, backoffPeriod, maxRetries})
+}
+
+// WithWorkloadClusterValidations registers a validation set for a management cluster.
+func (c *ClusterValidator) WithWorkloadClusterValidations() {
+	c.WithValidation(validateClusterReady, 5*time.Second, 60)
+	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 60)
```
taneyland, line 41: Are we going to add a check for worker nodes?

test/framework/workload.go
```diff
@@ -21,3 +32,71 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
 	w.deleteCluster(opts...)
 }
+
+// CreateClusterWithKubectl creates a cluster by applying a workload cluster config using the kubectl client.
+func (w *WorkloadCluster) CreateClusterWithKubectl() {
+	ctx := context.Background()
+	w.T.Logf("Applying workload cluster %s spec for creation located at %s", w.ClusterName, w.ClusterConfigLocation)
+	w.applyClusterManifest(ctx)
+	w.waitForKubeconfig(ctx)
+	time.Sleep(2 * time.Minute)
```
taneyland, line 42: Is this number (2 mins) derived from the average amount of time it takes for everything to be ready?

taneyland (commented):
Very nice! 

Do we plan to follow up in future PRs to expand & add more validations?

test/framework/cluster.go
```diff
@@ -41,7 +44,7 @@ import (
 const (
 	defaultClusterConfigFile         = "cluster.yaml"
 	defaultBundleReleaseManifestFile = "bin/local-bundle-release.yaml"
-	defaultEksaBinaryLocation        = "eksctl anywhere"
+	defaultEksaBinaryLocation        = "eksctl-anywhere"
```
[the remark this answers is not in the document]
cxbrowne1207: This is a mistake, will fix. This is an alias that points to my local binary in the repo.

test/framework/workload.go
cxbrowne1207: This sleep ideally shouldn't be here, but the test was throwing a `connect: network is unreachable` when trying to use it. Working on fixing this by implementing retries where necessary.

test/framework/clustervalidator.go
cxbrowne1207: Yes, we will! Currently working on checking for the existence of EKSA objects first, but that is the plan

cxbrowne1207:
/test all

> Very nice!
> 
> Do we plan to follow up in future PRs to expand & add more validations?

If we deem it necessary, yes. More validations can be added in the future

pkg/clients/kubernetes/runtimeclient_wb_test.go
```diff
@@ -0,0 +1,5 @@
+package kubernetes
+
+var NewRuntimeClient = newRuntimeClient
```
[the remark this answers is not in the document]
cxbrowne1207: newRuntimeClient is needed for the tests, but they it isn't exported (probably shouldn't export it just to test it). This is a way around it, and not having to put the tests in the same package.

tatlat (commented):
Nice! /lgtm

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,178 @@
...
+func validateControlPlaneNodes(ctx context.Context, validateOpts clusterOpts) error {
+	cpNodes := &corev1.NodeList{}
+	err := validateOpts.ClusterClient.List(ctx, cpNodes, client.MatchingLabels{"node-role.kubernetes.io/control-plane": ""})
+	if err != nil {
+		return fmt.Errorf("failed to list controlplane nodes %s", err)
+	}
+
+	cpConfig := validateOpts.ClusterSpec.Cluster.Spec.ControlPlaneConfiguration
+	if len(cpNodes.Items) != cpConfig.Count {
+		return fmt.Errorf("control plane node count does not match expected: %v of %v", len(cpNodes.Items), cpConfig.Count)
```
g-gaston, line 135: i think we need to use %d here?

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	for _, node := range cpNodes.Items {
+		err := validateNodeReady(node)
+		if err != nil {
```
g-gaston, line 140: one liner?

test/framework/clustervalidator.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("failed to list controlplane nodes %s", err)
+	}
+
+	cpConfig := validateOpts.ClusterSpec.Cluster.Spec.ControlPlaneConfiguration
+	if len(cpNodes.Items) != cpConfig.Count {
+		return fmt.Errorf("control plane node count does not match expected: %v of %v", len(cpNodes.Items), cpConfig.Count)
+	}
+
+	for _, node := range cpNodes.Items {
+		err := validateNodeReady(node)
```
g-gaston, line 139: We are gonna need to add something else here
this validation only checks there is a determined number of nodes and all nodes are ready. But it doesn't check if those nodes are the nodes we expect. For example, we might be expecting 3 nodes, but after an upgrade, this validation will succeed with two old and one new node.

And easy way would be to check the kubernetes version. There are other things we should check, but I think that's the bare minimum to ensure nodes have been rolled out.

test/framework/workload.go
```diff
@@ -21,3 +32,60 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
 	w.deleteCluster(opts...)
 }
+
+// ApplyClusterManifest uses client-side logic to create/update objects defined in a cluster yaml manifest.
+func (w *WorkloadCluster) ApplyClusterManifest() {
+	ctx := context.Background()
+	w.T.Logf("Applying workload cluster %s spec located at %s", w.ClusterName, w.ClusterConfigLocation)
+	w.applyClusterManifest(ctx)
+	w.waitForKubeconfig(ctx)
+	w.ValidateClusterState()
```
g-gaston, line 42: I would prefer if we called this separately. IMO this method should just call kubectl apply. If that succeeds, return. And leave the wait and validations for the test to decide.

In general, the framework should offer exported methods for individual operations and leave the orchestration of those operations to the tests.

test/framework/workload.go
```diff
@@ -21,3 +32,60 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
+}
+
+// DeleteClusterWithKubectl uses client-side logic to delete a cluster.
+func (w *WorkloadCluster) DeleteClusterWithKubectl() {
+	ctx := context.Background()
+	w.T.Logf("Deleting workload cluster %s with kubectl", w.ClusterName)
+	if err := w.KubectlClient.DeleteCluster(ctx, w.managementCluster(), w.cluster()); err != nil {
+		w.T.Fatalf("failed to delete workload cluster config: %s", err)
+	}
+	w.ValidateClusterDelete()
```
g-gaston, line 53: same comment about orchestration here

test/framework/workload.go
```diff
@@ -21,3 +32,60 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
+
+func (w *WorkloadCluster) applyClusterManifest(ctx context.Context) {
+	if err := w.KubectlClient.ApplyManifest(ctx, w.ManagementClusterKubeconfigFile(), w.ClusterConfigLocation); err != nil {
+		w.T.Fatalf("Failed to apply workload cluster config: %s", err)
+	}
+}
+
+func (w *WorkloadCluster) waitForKubeconfig(ctx context.Context) {
+	w.T.Logf("Waiting for workload cluster %s kubeconfig to be available", w.ClusterName)
+	err := retrier.Retry(12, 5*time.Second, func() error {
```
g-gaston, line 65: if it works this fast, cool
but this seems a bit too aggresive. I'm not sure if all providers will be able to spin up a control plane node in a minute.

test/framework/workload.go
```diff
@@ -21,3 +32,60 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
+		return w.writeKubeconfigToDisk(ctx)
+	})
+	if err != nil {
+		w.T.Fatalf("Failed waiting for cluster kubeconfig: %s", err)
+	}
+}
+
+func (w *WorkloadCluster) writeKubeconfigToDisk(ctx context.Context) error {
+	secret, err := w.KubectlClient.GetSecretFromNamespace(ctx, w.ManagementClusterKubeconfigFile(), fmt.Sprintf("%s-kubeconfig", w.ClusterName), constants.EksaSystemNamespace)
```
g-gaston, line 74: do we have a method somewhere to get the kubeconfig secret name? maybe even upstream capi has one we can use.

test/framework/workload.go
```diff
@@ -21,3 +32,60 @@ func (w *WorkloadCluster) DeleteCluster(opts ...CommandOpt) {
...
+		return fmt.Errorf("failed to write kubeconfig to disk: %v", err)
+	}
+
+	_, err = writer.Write(filepath.Base(w.kubeconfigFilePath()), kubeconfig, func(op *filewriter.FileOptions) {
+		op.IsTemp = false
+	})
+	if err != nil {
+		return fmt.Errorf("failed to write kubeconfig to disk: %v", err)
+	}
+	return err
```
g-gaston, line 90: return nil?

test/e2e/adot_test.go
```diff
@@ -124,11 +124,11 @@ func runCuratedPackagesAdotInstallWithUpdate(test *framework.ClusterE2ETest) {
 }
 
 func runCuratedPackagesAdotInstallSimpleFlow(test *framework.ClusterE2ETest) {
-	test.WithCluster(runCuratedPackagesAdotInstall)
+	test.RunWithCluster(runCuratedPackagesAdotInstall)
```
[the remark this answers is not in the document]
cxbrowne1207: Indeed. This was already fixed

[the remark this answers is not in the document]
cxbrowne1207: Indeed. This was already fixed

test/framework/cluster.go
```diff
@@ -1658,3 +1684,94 @@ func (e *ClusterE2ETest) CombinedAutoScalerMetricServerTest(autoscalerName strin
...
+
+// ValidateClusterDelete verifies the cluster has been deleted.
+func (e *ClusterE2ETest) ValidateClusterDelete() {
+	ctx := context.Background()
+	e.T.Logf("Validating cluster deletion %s", e.ClusterName)
+	if e.clusterValidator == nil {
+		return
+	}
+
+	e.clusterValidator.Reset()
```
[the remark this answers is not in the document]
cxbrowne1207: The reason for this was creating a new validator requires the a client, but after running the kubectl delete, it was not possible to instantiate a new client, which the code currently does every-time. I opted to re-use the validator, but I probably could have also re-used the client instead.

Issue #3282, opened 2022-09-07
Support static IP in Snow


jiayiwang7:
Checklist
- [x] Update snow bottlerocket bootstrap container network template to support static IP
- [x] Add static IP option in EKS-A snowmachineconfig

PR #4481, opened 2022-12-19
Support ip pools for snow worker groups

jiayiwang7:
*Issue #, if available:*

#3282 

Wait for #4455 to merge first.

*Description of changes:*

Add ipPools objects to snow worker groups.

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/providers/snow/objects.go
+++ b/pkg/providers/snow/objects.go
@@ -27,27 +27,56 @@ func ControlPlaneObjects(ctx context.Context, log logr.Logger, clusterSpec *clus
 	return cp.Objects(), nil
 }
 
+// WorkersObjects generates all the objects that compose a Snow specific CAPI spec for the worker nodes of an eks-a cluster.
+func WorkersObjects(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]kubernetes.Object, error) {
+	w, err := WorkersSpec(ctx, log, clusterSpec, kubeClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return w.Objects(), nil
+}
+
 type (
 	// Workers represents the Snow specific CAPI spec for worker nodes.
-	Workers     = clusterapi.Workers[*snowv1.AWSSnowMachineTemplate]
-	workerGroup = clusterapi.WorkerGroup[*snowv1.AWSSnowMachineTemplate]
+	BaseWorkers     = clusterapi.Workers[*snowv1.AWSSnowMachineTemplate]
+	baseWorkerGroup = clusterapi.WorkerGroup[*snowv1.AWSSnowMachineTemplate]
 )
+
+// Workers holds the Snow specific objects for CAPI snow worker groups.
+type Workers struct {
+	BaseWorkers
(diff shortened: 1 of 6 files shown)


pkg/providers/snow/objects.go
```diff
@@ -57,18 +86,17 @@ func WorkersSpec(ctx context.Context, log logr.Logger, spec *cluster.Spec, kubeC
...
 }
 
-// WorkersObjects generates all the objects that compose a Snow specific CAPI spec for the worker nodes of an eks-a cluster.
-func WorkersObjects(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec, kubeClient kubernetes.Client) ([]kubernetes.Object, error) {
-	w, err := WorkersSpec(ctx, log, clusterSpec, kubeClient)
-	if err != nil {
-		return nil, err
+func workersIPPools(clusterSpec *cluster.Spec) CAPASIPPools {
+	capasPools := CAPASIPPools{}
+	for _, workerNodeGroupConfig := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
```
g-gaston, line 91: you are already iterating over `WorkerNodeGroupConfigurations` in `WorkersSpec` to build the worker groups, could we take advantage of that loop to build the IPPools?

pkg/providers/snow/reconciler/reconciler.go
```diff
@@ -156,3 +159,25 @@ func toClientControlPlane(cp *snow.ControlPlane) *clusters.ControlPlane {
 		Other:                       other,
 	}
 }
+
+func toClientWorkers(workers *snow.Workers) *clusters.Workers {
+	other := make([]client.Object, 0, len(workers.CAPASIPPools))
+	for _, p := range workers.CAPASIPPools {
+		other = append(other, p)
+	}
```
g-gaston, line 167: can we add ip pools to a test? we don't need a new test, just add eks-a ip pools to one of the test's clusters input. And then just verify that the aws snow ip pools get created.

pkg/providers/snow/objects.go
jiayiwang7: The IPPools are needed to build workerMachineTemplates and kubeadmConfigTemplates. I don't think it will work to use the loop in the workersSpec. I can certainly restruct the WorkersSpec and  WorkersMachineAndConfigTemplate to maybe only have one loop but I don't think it's necessary.

g-gaston: Got you. Refactoring the existing code seems out of scope for this PR. I'll leave it up to you if you want to do that later or not.

I think eventually we should unify how all providers generate the capi spec. Maybe we should start thinking about how that is gonna look and then refactor this once we have that common vision defined.

jiayiwang7: I think we talked abt it offline, I agree we should have consistent logic. I will have followup PR to refactor how snow worker spec gets generated to align with the clusterapi API. It's in my TODO

pkg/providers/snow/reconciler/reconciler.go
jiayiwang7: updated reconciler test

g-gaston (approved):

PR #4485, opened 2022-12-20
docs: Add additional details for VMware docs

raykrueger:
In many cases we require a fully qualified path, rather than a name. For example, a network must be referenced with a fully qualified path, where as a datacenter is simply a name. I have tried to call out "path" vs "name" where applicable.

Additionally, I have added example govc commands to see what values exist.


TerryHowe (approved):
/lgtm
/approve

PR #4488, opened 2022-12-21
Update helm because CVE

TerryHowe:
update go version to 1.18 and update helm because of a CVE. Replaces https://github.com/aws/eks-anywhere/pull/4417


TerryHowe:
Looks like top level bad go.sum/go.mod was borking this, so updated and added them.

PR #4489, opened 2022-12-21
Refactor snow worker spec generation code

jiayiwang7:
*Issue #, if available:*

https://github.com/aws/eks-anywhere/pull/4481#discussion_r1053468444

*Description of changes:*

Refactor snow worker spec builder with the `clusterapi.Workers` API.

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/providers/snow/workers.go
+++ b/pkg/providers/snow/workers.go
@@ -1,0 +1,76 @@
+package snow
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	snowv1 "github.com/aws/eks-anywhere/pkg/providers/snow/api/v1beta1"
+)
+
+type (
+	// BaseWorkers represents the Snow specific CAPI spec for worker nodes.
+	BaseWorkers     = clusterapi.Workers[*snowv1.AWSSnowMachineTemplate]
+	baseWorkerGroup = clusterapi.WorkerGroup[*snowv1.AWSSnowMachineTemplate]
+)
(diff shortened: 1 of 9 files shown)


pkg/providers/snow/workers.go
```diff
@@ -0,0 +1,76 @@
...
+			KubeadmConfigTemplate:   kubeadmConfigTemplate,
+			ProviderMachineTemplate: machineTemplate,
+		})
+	}
+
+	if err := w.UpdateImmutableObjectNames(ctx, client, getMachineTemplate, MachineTemplateDeepDerivative); err != nil {
+		return nil, errors.Wrap(err, "updating snow worker immutable object names")
+	}
+
+	w.CAPASIPPools = capasPools
```
taneyland, line 73: Just to confirm, so the IP pools are meant to just be another part of the workers spec, not tied to specific worker node groups?

jiayiwang7: correct, not necessarily tied to worker node groups -- the same IPPool can be referenced by multiple worker groups.

pkg/providers/snow/workers.go
```diff
@@ -0,0 +1,76 @@
...
+			Groups: make([]baseWorkerGroup, 0, len(spec.Cluster.Spec.WorkerNodeGroupConfigurations)),
+		},
+	}
+
+	capasPools := CAPASIPPools{}
+
+	for _, wc := range spec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		machineConfig := spec.SnowMachineConfig(wc.MachineGroupRef.Name)
+
+		capasPools.addPools(machineConfig.Spec.Network.DirectNetworkInterfaces, spec.SnowIPPools)
```
g-gaston, line 51: do we need to accumulate all the ip pools for all the worker node groups or should we reset it on each loop iteration?

jiayiwang7: i chose accumulating because i can directly return capasPools in the end without another concat

PR #4490, opened 2022-12-21
Update CAPAS API

jiayiwang7:
*Issue #, if available:*

Copy from https://github.com/aws/cluster-api-provider-aws-snow latest

*Description of changes:*

CAPAS updates:
- add more comments for API fields
- make `Network` required field
- use `corev1.ObjectReference` for ipPool reference instead of capas specific one
- remove `DNS` field (which only applies to BottleRocket with static ip). CAPAS auto generates the DNS based on IP Pool and pass it into BottleRocket network config.

*Testing (if applicable):*

Unit tests


pkg/providers/snow/api/v1beta1/awssnowippool_types.go
```diff
@@ -29,17 +29,17 @@ type AWSSnowIPPoolSpec struct {
...
+	// IPPools defines a range of ip addresses for static IP configurations.
 	IPPools []IPPool `json:"pools,omitempty"`
 }
 
-// IPPool is the configuration of static ips used for machine's ip.
+// IPPool is the configuration of static ip, it provides a range of ip addresses
 type IPPool struct {
-	// IPStart is the start of an ip range
+	// IPStart is the start ip address of an ip range
 	IPStart *string `json:"ipStart,omitempty"`
```
g-gaston, line 39: are these optional?

jiayiwang7: from snow:
> If IPStart or IPEnd are nil, we will return an error: "Both IpStart or IpEnd is required". I changed them to pointers bc it looks more formal to check if they are nil rather than check if they are ""(empty string). Since I found metal3 also use pointers: https://github.com/metal3-io/ip-address-manager/blob/e323982264ebe518e3c5cd1272d297ffc44ab46c/api/v1alpha1/ippool_types.go#L31

taneyland (approved):

PR #4493, opened 2022-12-21
Add e2e test to upgrade management cluster using API 

taneyland:
*Issue #, if available:*

*Description of changes:*
We currently only support upgrading worker node config fields for mgmt clusters when using the API, so here's an e2e test for that.

I also refactored the cluster validator a bit to validate the cluster, eks-a objs, and nodes for self-managed clusters too.

*Testing (if applicable):*
```
--- PASS: TestVSphereKubernetes124UbuntuUpgradeAndRemoveWorkerNodeGroupsAPI (783.76s)
PASS
```


test/e2e/multiple_worker_node_groups_test.go
```diff
@@ -52,6 +52,46 @@ func TestVSphereKubernetes124BottlerocketAndAndRemoveWorkerNodeGroups(t *testing
 	)
 }
 
+func TestVSphereKubernetes124BottlerocketUpgradeAndRemoveWorkerNodeGroupsAPI(t *testing.T) {
```
cxbrowne1207, line 55: Does this test setup a self-managed cluster right?  I don't think we can use self-managed clusters for the API tests since it does not support it.

Can we use `MulticlusterE2ETest` to setup the test instead? Check the `TestVSphereMulticlusterWorkloadClusterAPI` for an example.

test/framework/cluster.go
```diff
@@ -661,12 +661,36 @@ func (e *ClusterE2ETest) GetCapiMachinesForCluster(clusterName string) map[strin
...
 }
 
+// UpgradeClusterWithKubectl uses client-side logic to upgrade a cluster.
+func (e *ClusterE2ETest) UpgradeClusterWithKubectl(clusterOpts []ClusterE2ETestOpt, commandOpts ...CommandOpt) {
+	for _, opt := range clusterOpts {
+		opt(e)
+	}
+	e.buildClusterConfigFile()
+	e.ApplyClusterManifest()
+}
```
cxbrowne1207, line 692: Can we add this method to `/test/framework/workload.go` and use the `ApplyClusterManifest` instead there as well?

test/framework/clustervalidator.go
```diff
@@ -35,11 +36,14 @@ func (c *ClusterValidator) WithValidation(validation clusterValidation, backoffP
...
-// WithWorkloadClusterValidations registers a validation set for a management cluster.
-func (c *ClusterValidator) WithWorkloadClusterValidations() {
+// WithWorkloadClusterValidations registers a validation set for a workload cluster.
+func (c *ClusterValidator) WithWorkloadClusterValidations() {}
+
+func (c *ClusterValidator) baseValidations() {
 	c.WithValidation(validateClusterReady, 5*time.Second, 60)
 	c.WithValidation(validateEKSAObjects, 5*time.Second, 60)
 	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 60)
+	c.WithValidation(validateWorkerNodes, 5*time.Second, 60)
```
cxbrowne1207, line 46: What was the reason for this change?

test/framework/clustervalidator.go
```diff
@@ -75,6 +79,7 @@ func NewClusterValidator(opts ...ClusterValidatorOpts) *ClusterValidator {
 		ClusterOpts: clusterOpts{},
 		validations: []retriableValidation{},
 	}
+	cv.baseValidations()
```
cxbrowne1207, line 82: Why do we need to have base validations in the constructor vs registering them when they're needed with `WithWorkloadClusterValidations`?

test/e2e/upgrade_test.go
```diff
@@ -37,6 +37,15 @@ func runSimpleUpgradeFlow(test *framework.ClusterE2ETest, updateVersion v1alpha1
 	test.DeleteCluster()
 }
 
+func runUpgradeFlowWithAPI(test *framework.ClusterE2ETest, updateVersion v1alpha1.KubernetesVersion, clusterOpts ...framework.ClusterE2ETestOpt) {
+	test.GenerateClusterConfig()
+	test.CreateCluster()
+	test.UpgradeClusterWithKubectl(clusterOpts)
+	test.ValidateCluster(updateVersion)
```
cxbrowne1207, line 44: Do we still need this after using `ValidateClusterState`?

taneyland: `ValidateCluster` also validates the node version. So I can add that to both of the node validation methods & remove this

test/e2e/vsphere_workload_api_test.go
```diff
@@ -22,7 +22,7 @@ func TestVSphereMulticlusterWorkloadClusterAPI(t *testing.T) {
 			api.WithWorkerNodeCount(1),
 			api.WithStackedEtcdTopology(),
 		),
-		vsphere.WithBottleRocket123(),
+		vsphere.WithUbuntu123(),
```
taneyland, line 25: Changing this to Ubuntu until we update the worker node validation to also handle BR nodes

test/e2e/upgrade_test.go
```diff
@@ -37,6 +37,15 @@ func runSimpleUpgradeFlow(test *framework.ClusterE2ETest, updateVersion v1alpha1
 	test.DeleteCluster()
 }
 
+func runUpgradeFlowWithAPI(test *framework.ClusterE2ETest, clusterOpts ...framework.ClusterE2ETestOpt) {
+	test.GenerateClusterConfig()
+	test.CreateCluster()
+	test.UpgradeClusterWithKubectl(clusterOpts)
+	test.ValidateClusterState()
```
cxbrowne1207, line 44: I think this should be inside `UpgradeClusterWithKubectl` command, since we'll always need to call it. It's basically apart of the process.

cxbrowne1207 (commented):
Looks great! Left one more comment, let me know what you think

PR #4498, opened 2022-12-22
Add infrastructure for Cluster Lifecycle API GitOps e2e tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR includes some infrastructure needed to write Gitops e2e tests for the Cluster Lifecycle API and vSphere workload cluster e2e tests that utilizes it.

*Testing (if applicable):*
```
--- PASS: TestVSphereMulticlusterWorkloadClusterAPIGitHubFlux (1312.20s)
PASS
```

### DIFF
--- a/pkg/filewriter/writer.go
+++ b/pkg/filewriter/writer.go
@@ -64,7 +64,7 @@ func (w *writer) CleanUpTemp() {
 	}
 }
 
-// Create creates a file with the given name rooted at w's base direcftion.
+// Create creates a file with the given name rooted at w's base directory.
 func (w *writer) Create(name string, opts ...FileOptionsFunc) (_ io.WriteCloser, path string, _ error) {
 	o := buildOptions(w, opts)
 
--- a/test/e2e/vsphere_workload_api_flux_test.go
+++ b/test/e2e/vsphere_workload_api_flux_test.go
@@ -10,42 +10,21 @@ import (
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
-func TestVSphereMulticlusterWorkloadClusterAPI(t *testing.T) {
+func TestVSphereMulticlusterWorkloadClusterAPIGitHubFlux(t *testing.T) {
 	vsphere := framework.NewVSphere(t)
 	managementCluster := framework.NewClusterE2ETest(
-		t, vsphere,
+		t, vsphere, framework.WithFluxGithubEnvVarCheck(), framework.WithFluxGithubCleanup(),
 	).WithClusterConfig(
 		api.ClusterToConfigFiller(
 			api.WithControlPlaneCount(1),
 			api.WithWorkerNodeCount(1),
 			api.WithStackedEtcdTopology(),
 		),
-		vsphere.WithUbuntu123(),
+		framework.WithFluxGithubConfig(),
+		vsphere.WithUbuntu124(),
 	)
 	test := framework.NewMulticlusterE2ETest(t, managementCluster)
 	test.WithWorkloadClusters(
-		framework.NewClusterE2ETest(
-			t, vsphere, framework.WithClusterName(test.NewWorkloadClusterName()),
-		).WithClusterConfig(
-			api.ClusterToConfigFiller(
-				api.WithManagementCluster(managementCluster.ClusterName),
-				api.WithControlPlaneCount(1),
-				api.WithWorkerNodeCount(1),
-				api.WithStackedEtcdTopology(),
-			),
-			vsphere.WithUbuntu121(),
-		),
-		framework.NewClusterE2ETest(
-			t, vsphere, framework.WithClusterName(test.NewWorkloadClusterName()),
-		).WithClusterConfig(
-			api.ClusterToConfigFiller(
-				api.WithManagementCluster(managementCluster.ClusterName),
-				api.WithControlPlaneCount(1),
-				api.WithWorkerNodeCount(1),
-				api.WithStackedEtcdTopology(),
-			),
-			vsphere.WithUbuntu122(),
-		),
 		framework.NewClusterE2ETest(
 			t, vsphere, framework.WithClusterName(test.NewWorkloadClusterName()),
 		).WithClusterConfig(
@@ -64,17 +43,18 @@ func TestVSphereMulticlusterWorkloadClusterAPI(t *testing.T) {
 				api.WithManagementCluster(managementCluster.ClusterName),
 				api.WithControlPlaneCount(1),
 				api.WithWorkerNodeCount(1),
-				api.WithStackedEtcdTopology(),
+				api.WithExternalEtcdTopology(1),
 			),
 			vsphere.WithUbuntu124(),
 		),
 	)
+
 	test.CreateManagementCluster()
-	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
-		wc.ApplyClusterManifest()
+	test.RunInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		test.PushWorkloadClusterToGit(wc)
 		wc.WaitForKubeconfig()
 		wc.ValidateClusterState()
-		wc.DeleteClusterWithKubectl()
+		test.DeleteWorkloadClusterFromGit(wc)
 		wc.ValidateClusterDelete()
 	})
 	test.ManagementCluster.StopIfFailed()
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -711,12 +711,23 @@ func (e *ClusterE2ETest) upgradeCluster(clusterOpts []ClusterE2ETestOpt, command
 func (e *ClusterE2ETest) generateClusterConfigYaml() []byte {
 	childObjs := e.ClusterConfig.ChildObjects()
 	yamlB := make([][]byte, 0, len(childObjs)+1)
+
+	// This is required because Flux requires a namespace be specified for objects
+	// to be able to reconcile right.
+	if e.ClusterConfig.Cluster.Namespace == "" {
+		e.ClusterConfig.Cluster.Namespace = "default"
+	}
 	clusterConfigB, err := yaml.Marshal(e.ClusterConfig.Cluster)
 	if err != nil {
 		e.T.Fatal(err)
 	}
 	yamlB = append(yamlB, clusterConfigB)
 	for _, o := range childObjs {
+		// This is required because Flux requires a namespace be specified for objects
+		// to be able to reconcile right.
+		if o.GetNamespace() == "" {
+			o.SetNamespace("default")
+		}
 		objB, err := yaml.Marshal(o)
 		if err != nil {
 			e.T.Fatalf("Failed marshalling %s config: %v", o.GetName(), err)
@@ -899,10 +910,6 @@ func (e *ClusterE2ETest) GetEksaVSphereMachineConfigs() []v1alpha1.VSphereMachin
 	}
 
 	return machineConfigs
-}
-
-func (e *ClusterE2ETest) getJobIdFromEnv() string {
-	return os.Getenv(JobIdVar)
 }
 
 func GetTestNameHash(name string) string {
--- a/test/framework/clustervalidator.go
+++ b/test/framework/clustervalidator.go
@@ -44,7 +44,7 @@ func (c *ClusterValidator) WithWorkloadClusterValidations() {}
 func (c *ClusterValidator) WithExpectedObjectsExist() {
 	c.WithValidation(validateClusterReady, 5*time.Second, 60)
 	c.WithValidation(validateEKSAObjects, 5*time.Second, 60)
-	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 60)
+	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 120)
 	c.WithValidation(validateWorkerNodes, 5*time.Second, 60)
 }
 
--- a/test/framework/flux.go
+++ b/test/framework/flux.go
@@ -60,17 +60,21 @@ var fluxGitCreateGenerateRepoEnvVars = []string{
 	GithubTokenVar,
 }
 
+func getJobIDFromEnv() string {
+	return os.Getenv(JobIdVar)
+}
+
 func WithFluxGit(opts ...api.FluxConfigOpt) ClusterE2ETestOpt {
 	return func(e *ClusterE2ETest) {
 		checkRequiredEnvVars(e.T, fluxGitRequiredEnvVars)
-		jobId := strings.Replace(e.getJobIdFromEnv(), ":", "-", -1)
+		jobID := strings.Replace(getJobIDFromEnv(), ":", "-", -1)
 		e.ClusterConfig.FluxConfig = api.NewFluxConfig(DefaultFluxConfigName,
 			api.WithGenericGitProvider(
 				api.WithStringFromEnvVarGenericGitProviderConfig(GitRepoSshUrl, api.WithGitRepositoryUrl),
 			),
 			api.WithSystemNamespace("default"),
-			api.WithClusterConfigPath(jobId),
-			api.WithBranch(jobId),
+			api.WithClusterConfigPath(jobID),
+			api.WithBranch(jobID),
 		)
 		e.clusterFillers = append(e.clusterFillers,
 			api.WithGitOpsRef(DefaultFluxConfigName, v1alpha1.FluxConfigKind),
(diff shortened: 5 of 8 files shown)


cxbrowne1207:
/hold

I need to run the tests for FluxLegacy and FluxGit once more to ensure they work

cxbrowne1207:
\unhold

cxbrowne1207:
\unhold

pkg/filewriter/writer.go
```diff
@@ -38,6 +38,18 @@ func (w *writer) Write(fileName string, content []byte, opts ...FileOptionsFunc)
 	return filePath, nil
 }
 
+// Remove removes the file with the given path rooted at w's base direcftion.
+func (w *writer) Remove(fileName string, opts ...FileOptionsFunc) error {
+	o := buildOptions(w, opts)
+
+	dir := filepath.Join(o.BasePath, fileName)
```
vivek-koppuru, line 45:
```suggestion
	absPath := filepath.Join(o.BasePath, fileName)
```

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+func TestVSphereMulticlusterWorkloadClusterAPIGitHubFlux(t *testing.T) {
+	vsphere := framework.NewVSphere(t)
+	managementCluster := framework.NewClusterE2ETest(
+		t, vsphere,
+	)
+	managementCluster.WithClusterConfig(
+		api.ClusterToConfigFiller(
+			api.WithKubernetesVersion(v1alpha1.Kube123),
```
vivek-koppuru, line 21: Why not use the latest Kubernetes version now which is 1.24?

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,228 @@
...
+		),
+	)
+
+	test.CreateManagementCluster()
+	test.RunInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.CreateClusterWithAPIGitOps()
+		wc.DeleteClusterWithAPIGitOps()
+	})
+	test.ManagementCluster.StopIfFailed()
+	test.DeleteManagementCluster()
```
vivek-koppuru, line 83: Abstract out this shared logic?

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"testing"
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/test/framework"
+)
+
+func TestVSphereMulticlusterWorkloadClusterAPIGitHubFlux(t *testing.T) {
+	vsphere := framework.NewVSphere(t)
+	managementCluster := framework.NewClusterE2ETest(
```
vivek-koppuru, line 16: Can you remind me what name is used here if we are specifying new names for the workload clusters below? Should we maybe reuse the name and append something so that we can associate the clusters better?

test/framework/cluster.go
```diff
@@ -688,12 +688,18 @@ func (e *ClusterE2ETest) upgradeCluster(clusterOpts []ClusterE2ETestOpt, command
 func (e *ClusterE2ETest) generateClusterConfigYaml() []byte {
 	childObjs := e.ClusterConfig.ChildObjects()
 	yamlB := make([][]byte, 0, len(childObjs)+1)
+	if e.ClusterConfig.Cluster.Namespace == "" {
+		e.ClusterConfig.Cluster.Namespace = "default"
```
vivek-koppuru, line 692: This is required because it's a requirement for Flux to be able to reconcile right? Add a comment maybe?

test/e2e/vsphere_workload_api_flux_test.go
cxbrowne1207: One of the workload clusters is using 1.24, so that gets tested as well. There's no problem adjusting this if there's a preference for 1.24 on the management cluster though.

test/e2e/vsphere_workload_api_flux_test.go
cxbrowne1207: `NewWorkloadClusterName` returns a new unique name for a workload cluster based on the management cluster name. So, I think this already does what you're saying.

test/e2e/vsphere_workload_api_flux_test.go
vivek-koppuru: Yea let's stick to the latest for the management cluster, we can do older ones for test cases that tests running a 1.24 workload cluster on a 1.22 management cluster for example, which is a test case we should cover.

pkg/filewriter/writer.go
```diff
@@ -38,6 +38,18 @@ func (w *writer) Write(fileName string, content []byte, opts ...FileOptionsFunc)
...
+// Remove removes the file with the given path rooted at w's base direcftion.
+func (w *writer) Remove(fileName string, opts ...FileOptionsFunc) error {
+	o := buildOptions(w, opts)
+
+	filePath := filepath.Join(o.BasePath, fileName)
+	if _, err := os.Stat(filePath); err == nil {
+		return os.Remove(filePath)
+	}
+
+	return nil
```
vivek-koppuru, line 50: This should be a quick case to cover right?

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,220 @@
...
+			api.WithControlPlaneCount(1),
+			api.WithWorkerNodeCount(1),
+			api.WithStackedEtcdTopology(),
+		),
+		managementCluster.WithFluxGithub(),
+		vsphere.WithBottleRocket123(),
+	)
+	test := framework.NewMulticlusterE2ETest(t, managementCluster)
+	test.WithWorkloadClusters(
+		framework.NewClusterE2ETest(
```
vivek-koppuru, line 41: Does the Kubernetes version not need to be set here for the cluster config?

cxbrowne1207: That is now being done through `WithClusterConfig`, which generates the base cluster config and then updates it with some fillers

test/framework/workload.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type WorkloadCluster struct {
 	*ClusterE2ETest
+	ManagementCluster               *ClusterE2ETest
```
g-gaston, line 16: Is this necessary? I would prefer if the workload cluster access to the management cluster tests is very limited. From here the workload cluster code could do anything to the management cluster and that seems dangerous. We already have a very blurred line dividing responsibilities, this creates another big cross dependency.

test/framework/workload.go
```diff
@@ -54,15 +61,96 @@ func (w *WorkloadCluster) DeleteClusterWithKubectl() {
...
+	if err := w.addConfigChangesInManagementClusterToGit(ctx); err != nil {
+		w.T.Errorf("Error pushing local changes to remote git repo: %v", err)
+	}
+}
+
+// DeleteClusterWithAPIGitOps uses gitops to delete a cluster with the API.
+func (w *WorkloadCluster) DeleteClusterWithAPIGitOps() {
+	ctx := context.Background()
+	w.deleteWithAPIGitOps(ctx)
+	w.ValidateClusterDelete()
```
g-gaston, line 126: Validations shouldn't be included in methods that denote operations IMO

test/framework/workload.go
```diff
@@ -54,15 +61,96 @@ func (w *WorkloadCluster) DeleteClusterWithKubectl() {
...
 
-func (w *WorkloadCluster) applyClusterManifest(ctx context.Context) {
-	if err := w.KubectlClient.ApplyManifest(ctx, w.ManagementClusterKubeconfigFile(), w.ClusterConfigLocation); err != nil {
-		w.T.Fatalf("Failed to apply workload cluster config: %s", err)
+// CreateClusterWithAPIGitOps uses gitops to create a cluster with the API.
+func (w *WorkloadCluster) CreateClusterWithAPIGitOps() {
+	ctx := context.Background()
+	w.createWithAPIGitOps(ctx)
+	w.waitForKubeconfig(ctx)
+	w.ValidateClusterState()
```
g-gaston, line 69: same comment about validations and waiting for kubeconfig

test/framework/workload.go
```diff
@@ -54,15 +61,96 @@ func (w *WorkloadCluster) DeleteClusterWithKubectl() {
...
+	w.T.Logf("Successfully created version controlled cluster configuration")
+}
+
+func (w *WorkloadCluster) createWithAPIGitOps(ctx context.Context) {
+	w.ManagementCluster.initGit(ctx)
```
g-gaston, line 74: this code is all very entangled
Plus it's not thread safe, you can be possibly modifying the management cluster test concurrently from the workload clusters

This goes back to my comment about giving the workload cluster test access to the management cluster test. I believe we need to rethink this design and decouple them.

test/framework/workload.go
```diff
@@ -54,15 +61,96 @@ func (w *WorkloadCluster) DeleteClusterWithKubectl() {
...
+	// Marshall w.ClusterConfig and write it to the repo path
+	w.buildClusterConfigFileInManagementClusterForGit()
+
+	if err := w.addConfigChangesInManagementClusterToGit(ctx); err != nil {
+		w.T.Errorf("Error pushing local changes to remote git repo: %v", err)
+	}
+}
+
+// UpgradeClusterWithAPIGitOps uses gitops to upgrade a cluster with the API.
+func (w *WorkloadCluster) UpgradeClusterWithAPIGitOps(clusterOpts ...ClusterE2ETestOpt) {
```
g-gaston, line 91: Why is this different than create? isn't it the same, just writing a file to disk, committing and pushing? The one who knows which operation is a create a which one is an upgrade is the test itself, since that's the entity who orchestrates the operations.

test/framework/workload_flux.go
```diff
@@ -0,0 +1,98 @@
+package framework
```
g-gaston, line 1: I don't think we need a new file for this, this is still part of the workload test code. If you see yourself making the original file too big, that's a good sign your entity has grown too big.

test/framework/workload_flux.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if err := w.pushConfigChangesInManagementCluster(ctx, "EKS-A E2E Flux test configuration removed"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (w *WorkloadCluster) pushConfigChangesInManagementCluster(ctx context.Context, commitMessage string) error {
+	g := w.ManagementCluster.GitClient
```
g-gaston, line 58: same comment about accessing the management cluster. all this code suffers from deep coupling and unclear responsibility boundary. Let's rethink this.

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,220 @@
...
+				api.WithStackedEtcdTopology(),
+			),
+			vsphere.WithUbuntu124(),
+		),
+	)
+
+	runSimpleMulticlusterFlow(test)
+}
+
+func TestVSphereMulticlusterWorkloadClusterAPIGitFlux(t *testing.T) {
```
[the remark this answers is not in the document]
cxbrowne1207: Okay, makes sense.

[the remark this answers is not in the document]
cxbrowne1207: Okay, makes sense.

pkg/filewriter/writer.go
```diff
@@ -38,6 +38,18 @@ func (w *writer) Write(fileName string, content []byte, opts ...FileOptionsFunc)
 	return filePath, nil
 }
 
+// Remove removes the file with the given path rooted at w's base direcftion.
+func (w *writer) Remove(fileName string, opts ...FileOptionsFunc) error {
```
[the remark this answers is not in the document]
g-gaston: why would you need that? The write method returns the path the file is written to. So you already have everything you need to delete it. One could argue we don't even need the delete method here.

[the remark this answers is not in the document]
cxbrowne1207: Good point. The original idea was that it was a convenience function, but the full path is also actually being build and used with `clusterConfigGitPathInManagementCluster`

test/framework/clustervalidator.go
```diff
@@ -44,7 +44,7 @@ func (c *ClusterValidator) WithWorkloadClusterValidations() {}
 func (c *ClusterValidator) WithExpectedObjectsExist() {
 	c.WithValidation(validateClusterReady, 5*time.Second, 60)
 	c.WithValidation(validateEKSAObjects, 5*time.Second, 60)
-	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 60)
+	c.WithValidation(validateControlPlaneNodes, 5*time.Second, 120)
```
mitalipaygude, line 47: Was there some timeout with the previous 60 sec?

test/framework/flux.go
```diff
@@ -247,6 +260,123 @@ func (e *ClusterE2ETest) initGit(ctx context.Context) {
...
+	if _, err := os.Stat(p); errors.Is(err, os.ErrNotExist) {
+		err := os.MkdirAll(p, os.ModePerm)
+		if err != nil {
+			w.T.Errorf("creating directory [%s]: %v", g.Dir(), err)
+		}
+	}
+
+	_, err := g.Write(e.workloadClusterConfGitPath(w), b, filewriter.PersistentFile)
+	if err != nil {
+		w.T.Errorf("Error writing cluster config file to local git folder: %v", err)
```
mitalipaygude, line 285: nit:
```suggestion
		w.T.Errorf("error writing cluster config file to local git folder: %v", err)
```

test/framework/clustervalidator.go
cxbrowne1207: Yeah, I experienced a timeout here when testing upgrading control planes nodes, so I increased this a bit.

This is a smell and I'll attempt to decouple the validations from the validator in another PR

test/e2e/vsphere_workload_api_flux_test.go
```diff
@@ -0,0 +1,220 @@
...
+	test.ManagementCluster.StopIfFailed()
+	test.DeleteManagementCluster()
+}
+
+func TestVSphereMulticlusterWorkloadClusterAPIGitHubFlux(t *testing.T) {
+	vsphere := framework.NewVSphere(t)
+	managementCluster := framework.NewClusterE2ETest(
+		t, vsphere,
+	)
+	managementCluster.WithClusterConfig(
```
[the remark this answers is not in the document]
cxbrowne1207: Had to decouple `WithFluxGithub` from the test to do this first.

test/framework/flux.go
cxbrowne1207: Since this is logging to the console, would it not make sense to have it capitalized like all the other logs?

test/framework/flux.go
```diff
@@ -247,6 +291,123 @@ func (e *ClusterE2ETest) initGit(ctx context.Context) {
 	e.GitClient = g.Client
 }
 
+func (e *ClusterE2ETest) workloadClusterConfGitPath(w *WorkloadCluster) string {
```
taneyland, line 294: nit: 
```suggestion
func (e *ClusterE2ETest) workloadClusterConfigGitPath(w *WorkloadCluster) string {
```

test/framework/flux.go
```diff
@@ -915,12 +1076,16 @@ func (e *ClusterE2ETest) gitBranch() string {
 	return e.ClusterConfig.FluxConfig.Spec.Branch
 }
 
-func (e *ClusterE2ETest) clusterConfGitPath() string {
+func (e *ClusterE2ETest) clusterConfGitPathFromName(clusterName string) string {
```
taneyland, line 1079: nit: I know we use `Conf` in that other method but can we spell it out in these new methods?
```suggestion
func (e *ClusterE2ETest) clusterConfigGitPathFromName(clusterName string) string {
```

test/framework/flux.go
```diff
@@ -247,6 +291,123 @@ func (e *ClusterE2ETest) initGit(ctx context.Context) {
...
+
+	if err := e.pushStagedChanges(ctx, "EKS-A E2E Flux test workload configuration deleted"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *ClusterE2ETest) pushStagedChanges(ctx context.Context, commitMessage string) error {
+	g := e.GitClient
+	if err := g.Commit(commitMessage); err != nil {
```
taneyland, line 348: Can we maybe add these commit & push operations to a separate helper method to use both here & in the existing `pushConfigChanges` method? Seems like it's doing the same thing at that point

cxbrowne1207: What do you mean?

The way I see it, `pushedStagedChanges` is the helper method since it includes the common commit and push flow in both. I could update `pushConfigChanges` to use it

taneyland: The current lines 348-359 in this file are basically the same as this: https://github.com/aws/eks-anywhere/blob/main/test/framework/flux.go#L861-L872

So I wanted your thoughts on maybe extracting those steps into their own little helper method to avoid duplicating code

taneyland (approved):

PR #4504, opened 2022-12-22
Add validation to only support unstacked etcd with static ip for snow

jiayiwang7:
*Issue #, if available:*

Currently CAPAS only supports creating an unstacked etcd machine with static IP.

*Description of changes:*

Add validation so that etcd machine using DHCP will fail the preflight check.

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/cluster/snow.go
+++ b/pkg/cluster/snow.go
@@ -77,6 +77,9 @@ func snowEntry() *ConfigManagerEntry {
 			func(c *Config) error {
 				return ValidateSnowMachineRefExists(c)
 			},
+			func(c *Config) error {
+				return validateSnowUnstackedEtcd(c)
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/cluster/snow.go
```diff
@@ -268,3 +271,21 @@ func ValidateSnowMachineRefExists(c *Config) error {
 	}
 	return nil
 }
+
+func validateSnowUnstackedEtcd(c *Config) error {
+	if c.Cluster.Spec.ExternalEtcdConfiguration == nil || c.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef == nil {
+		return nil
+	}
+
+	mc := c.SnowMachineConfig(c.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name)
```
vivek-koppuru, line 280: Why can this be nil?

pkg/cluster/snow.go
```diff
@@ -268,3 +271,21 @@ func ValidateSnowMachineRefExists(c *Config) error {
...
+	if mc == nil {
+		return nil
+	}
+
+	for _, dni := range mc.Spec.Network.DirectNetworkInterfaces {
+		if dni.DHCP || dni.IPPoolRef == nil {
```
vivek-koppuru, line 286: Should these be different checks? The second one seems like it might just be missing and not really configured to DHCP.

pkg/cluster/snow.go
jiayiwang7: on testing mainly. also bc mc is a pointer, it can be nil. i don't wanna this method to depend on side affect

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #4516, opened 2022-12-28
Add Prometheus e2e update workflow

ivyostosh:
Also refactor some helper functions to better follow DRY principle

*Issue #, if available:* https://github.com/aws/eks-anywhere-packages/issues/639

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -41,7 +41,7 @@ func TestCPackagesPrometheusVSphereKubernetes123BottleRocketSimpleFlow(t *testin
 	runCuratedPackagesPrometheusInstallSimpleFlow(test)
 }
 
-func TestCPackagesPrometheusVSphereKubernetes122UbuntuSimpleFlow(t *testing.T) {
+func TestCPackagesPrometheusVSphereKubernetes122UbuntuUpdateFlow(t *testing.T) {
 	framework.CheckCuratedPackagesCredentials(t)
 	test := framework.NewClusterE2ETest(t,
 		framework.NewVSphere(t, framework.WithUbuntu122()),
@@ -50,7 +50,7 @@ func TestCPackagesPrometheusVSphereKubernetes122UbuntuSimpleFlow(t *testing.T) {
 			EksaPackageControllerHelmChartName, EksaPackageControllerHelmURI,
 			EksaPackageControllerHelmVersion, EksaPackageControllerHelmValues),
 	)
-	runCuratedPackagesPrometheusInstallSimpleFlow(test)
+	runCuratedPackagesPrometheusUpdateFlow(test)
 }
 
 func TestCPackagesPrometheusCloudStackRedhatKubernetes121SimpleFlow(t *testing.T) {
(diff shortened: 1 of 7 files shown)


test/framework/cluster.go
```diff
@@ -1729,6 +1724,75 @@ func (e *ClusterE2ETest) ValidateClusterState() {
...
+	ctx := context.Background()
+	packageMetadatNamespace := fmt.Sprintf("%s-%s", "eksa-packages", e.ClusterName)
+
+	e.T.Log("Apply changes to package", packageName)
+	err := e.KubectlClient.ApplyKubeSpecFromBytesWithNamespace(ctx, e.cluster(), PackageFile, packageMetadatNamespace)
+	if err != nil {
+		e.T.Fatalf("Error upgrading package: %s", err)
+		return
+	}
+	time.Sleep(30 * time.Second) // Add sleep to allow package to change state
```
junshun, line 1738: Instead of adding sleep lets just check the package state.

test/framework/cluster.go
```diff
@@ -1729,6 +1724,75 @@ func (e *ClusterE2ETest) ValidateClusterState() {
...
+}
+
+// CurlEndpointByBusyBox creates a busybox pod with command to curl the target endpoint,
+// and returns the created busybox pod name.
+func (e *ClusterE2ETest) CurlEndpointByBusyBox(endpoint string, namespace string) string {
```
junshun, line 1743: With these functions refactored can we udpate the VerifyHelloPackageInstalled to also use these as well

test/e2e/prometheus_test.go
```diff
@@ -113,13 +113,38 @@ func runCuratedPackagesPrometheusInstall(test *framework.ClusterE2ETest) {
...
+	test.InstallCuratedPackage(packageName, packageFullName,
+		kubeconfig.FromClusterName(test.ClusterName), packageTargetNamespace,
+		"--set server.persistentVolume.storageClass=local-path")
+
+	test.ApplyPrometheusPackageServerStatefulSetFile(packageFullName, packageTargetNamespace)
+	test.VerifyPrometheusPackageInstalled(packageFullName, packageTargetNamespace)
+	test.VerifyPrometheusPrometheusServerStates(packageFullName, packageTargetNamespace, "statefulset")
+	test.VerifyPrometheusNodeExporterStates(packageFullName, packageTargetNamespace)
+
+	test.ApplyPrometheusPackageServerDeploymentFile(packageFullName, packageTargetNamespace)
```
junshun, line 134: Is there any conflicts with both of these installed at the same time? Would it be better to uninstall the other package before moving on to this test?

test/framework/testdata/prometheus_package_deployment.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: packages.eks.amazonaws.com/v1alpha1
+kind: Package
+metadata:
+  name: generated-prometheus
```
junshun, line 4: might be easier to debug failures in the future if this had a -deployment appended to name to differentiate it from the stateful set when doing k get packages

test/e2e/prometheus_test.go
ivyostosh: This test was designed to test the switch between statefulset and deployment for prometheus-server (along with other config changes). So, no uninstalling the package first will diminish the purpose of this test.

test/framework/cluster.go
ivyostosh: I mainly use this method for package update, which means the package was in installed state first, then changing to installing, then installed again. If I check the package state too fast, the package will show in installed state, although it’s not truly ready. I don’t like using sleep, but not sure what is a better approach here?

test/framework/cluster.go
ivyostosh: Updated

test/framework/testdata/prometheus_package_deployment.yaml
ivyostosh: I like the idea, but this is not easy to change without some thorough testing (as many tests referenced this name and they were written to be mode-agnostic).

I created https://github.com/aws/eks-anywhere-packages/issues/712 and https://github.com/aws/eks-anywhere-packages/issues/711 to track this comment. Is it ok I address them in a separate PR?

test/framework/cluster.go
junshun: Since we check immediately after a helper function that first checks that it goes into the installing state might work? So we can have a wait on installing state then it should go into installed state.

junshun: The only major concern I see here is that this function is meant to be reusable but in some cases like the harbor install 30 seconds might not be enough where as in the hello case its a bit long.

junshun (commented):
I am okay with this if maybe we just add an issue for a helper function to update packages and checking state, dealing with that in a seperate PR

ivyostosh:
> I am okay with this if maybe we just add an issue for a helper function to update packages and checking state, dealing with that in a seperate PR

@junshun Yes, please see issue at https://github.com/aws/eks-anywhere-packages/issues/713 and my response at https://github.com/aws/eks-anywhere/pull/4516/files#diff-4438221d080eeaeb7e58673f2b24e69bf482eb366de82660ae94342614f9ef7dR1718

ivyostosh (commented):
@junshun Hmm looks like this comment didn't get published properly. Just FYI.

ivyostosh:
This should also close https://github.com/aws/eks-anywhere-packages/issues/724

PR #4532, opened 2022-12-30
Update e2e instance count

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


mitalipaygude:
We are increasing the count by 50? is this due to the timeout for e2e tests? was it caused by lack of resources?

taneyland:
> We are increasing the count by 50? is this due to the timeout for e2e tests? was it caused by lack of resources?

They're testing that out now. I think the e2e test issues are likely a matter of cleaning up resources. 
However, our current instance count is 217, which is why I increased it to 220.

TerryHowe:
> > We are increasing the count by 50? is this due to the timeout for e2e tests? was it caused by lack of resources?
> 
> They're testing that out now. I think the e2e test issues are likely a matter of cleaning up resources. However, our current instance count is 217, which is why I increased it to 220.


If current instance count is 217 and this changes the value from 170 to 220, what does this change do?

taneyland:
I now remember that the count in our e2e channel is the count of actual tests that failed, not the instance count because many of those tests are actually running on the same instance.So I'm going to close this PR. As folks look into the tests more, we can increase this as needed

PR #4539, opened 2022-12-31
Add Docker API e2e test upgrading worker node for mgmt cluster

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
We currently only support upgrading worker node config fields for mgmt clusters when using the API, so here's a docker e2e test for that.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/multiple_worker_node_groups_test.go
+++ b/test/e2e/multiple_worker_node_groups_test.go
@@ -78,6 +78,31 @@ func TestVSphereKubernetes124UbuntuUpgradeAndRemoveWorkerNodeGroupsAPI(t *testin
 	)
 }
 
+func TestDockerKubernetes124UpgradeAndRemoveWorkerNodeGroupsAPI(t *testing.T) {
+	provider := framework.NewDocker(t)
+	test := framework.NewClusterE2ETest(
+		t, provider,
+	).WithClusterConfig(
+		api.ClusterToConfigFiller(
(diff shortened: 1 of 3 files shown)


test/e2e/multiple_worker_node_groups_test.go
```diff
@@ -78,6 +78,31 @@ func TestVSphereKubernetes124UbuntuUpgradeAndRemoveWorkerNodeGroupsAPI(t *testin
...
+		),
+		provider.WithWorkerNodeGroup(framework.WithWorkerNodeGroup("worker-1", api.WithCount(2))),
+		provider.WithWorkerNodeGroup(framework.WithWorkerNodeGroup("worker-2", api.WithCount(1))),
+	)
+
+	runUpgradeFlowWithAPI(
+		test,
+		api.ClusterToConfigFiller(
+			api.RemoveWorkerNodeGroup("worker-2"),
+			api.WithWorkerNodeGroup("worker-1", api.WithCount(1)),
```
g-gaston, line 100: It could be a good idea to add labels and taints to this test to extend the coverage without using more resources

g-gaston (approved):

test/e2e/multiple_worker_node_groups_test.go
cxbrowne1207: Good idea, I'll submit another PR updating this test

PR #4548, opened 2023-01-04
Remove method from provider reconciler interface

g-gaston:
*Description of changes:*
This method won't be needed and not used anywhere.

### DIFF
--- a/controllers/mocks/provider.go
+++ b/controllers/mocks/provider.go
@@ -9,7 +9,6 @@ import (
 	reflect "reflect"
 
 	v1alpha1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
-	cluster "github.com/aws/eks-anywhere/pkg/cluster"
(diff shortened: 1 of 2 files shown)


chrisdoherty4:
Aren't there a bunch of implementations around this? How is CNI reconciled now?

cxbrowne1207 (approved):

g-gaston:
> Aren't there a bunch of implementations around this? How is CNI reconciled now?

There aren't, there are no consumers for this method from the interface. It just happens that all the current implementations of this interface also implement the `ReconcileCNI` method, but that's not required.

CNI reconciliation is supposed to be part of the responsibilities of the `Reconcile` implementation of this interface. For reference, the actual reconciliation logic is in the `CNIReconciler` and the existing implementations of `ProviderClusterReconciler` get the `CNIReconciler` injected as a dependency.

mitalipaygude (approved):

PR #4550, opened 2023-01-04
update-bundle-golden-files

TerryHowe:
Need a bot for this


TerryHowe:
Some test making web requests looks like
```
--- FAIL: TestReconcilerReconcileWorkerNodesSuccess (0.16s)
    client.go:91: Try updating status but failed with a 404 error for [distro.eks.amazonaws.com/v1alpha1, Kind=Release name=test namespace=eksa-system] object, most probably because it doesn't have a defined status subresource
```

TerryHowe:
moving target

PR #4551, opened 2023-01-04
Remove TODO that is not going to happen

TerryHowe:
This TODO I think was done already.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -195,7 +195,6 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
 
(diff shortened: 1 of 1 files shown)


chrisdoherty4:
Should we be removing the `replace` directive and depending on a released thing if this is done?

TerryHowe:
I had to dig through the past

https://github.com/aws/eks-anywhere/blob/v0.5.0/go.mod#L39:40

TerryHowe:
Yeh, we aren't going to used the versioned module and if we have to, we can address that at that time, no need for a TODO. I generally don't like TODOs, they should be issues if they are issues.

PR #4560, opened 2023-01-05
Add docker e2e api tests workload cluster k8s upgrade, taints and labels

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR renames a file adds two docker e2e tests for the full lifecycle API, updates a vSphere API test with labels and taints to improve coverage without using more resources, and implements the `UpdateKubeConfig` to e2e Provider interface, which is required implementation for the docker provider. 

This is part 1 of a bigger PR. There will be another one adding validations for labels and taints to the controlplane and worker nodes validations.

`--- PASS: TestDockerUpgradeWorkloadClusterLabelsAndTaintsAPI (1086.90s)
`

`--- PASS: TestDockerUpgradeKubernetes123to124WorkloadClusterScaleWorkerNodesAPI (1418.77s)
`
*Testing (if applicable):*

### DIFF
--- a/test/e2e/multiple_worker_node_groups_test.go
+++ b/test/e2e/multiple_worker_node_groups_test.go
@@ -65,6 +65,7 @@ func TestVSphereKubernetes124UbuntuUpgradeAndRemoveWorkerNodeGroupsAPI(t *testin
 		),
 		provider.WithWorkerNodeGroup("worker-1", framework.WithWorkerNodeGroup("worker-1", api.WithCount(2))),
 		provider.WithWorkerNodeGroup("worker-2", framework.WithWorkerNodeGroup("worker-2", api.WithCount(1))),
+		provider.WithWorkerNodeGroup("worker-3", framework.WithWorkerNodeGroup("worker-3", api.WithCount(1), api.WithLabel("tier", "frontend"))),
 		provider.WithUbuntu124(),
 	)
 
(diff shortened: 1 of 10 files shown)


test/e2e/multiple_worker_node_groups_test.go
```diff
@@ -89,17 +89,26 @@ func TestDockerKubernetes124UpgradeAndRemoveWorkerNodeGroupsAPI(t *testing.T) {
 			api.WithControlPlaneCount(1),
 			api.RemoveAllWorkerNodeGroups(), // This gives us a blank slate
 		),
-		provider.WithWorkerNodeGroup(framework.WithWorkerNodeGroup("worker-1", api.WithCount(2))),
+		provider.WithWorkerNodeGroup(
+			framework.WithWorkerNodeGroup(
+				"worker-1",
+				api.WithCount(2),
+				api.WithLabel("tier", "frontend"),
```
vivek-koppuru, line 96: Why are we adding labels and taints to this test?

test/e2e/workload_api_test.go
```diff
@@ -0,0 +1,181 @@
...
+		framework.NewClusterE2ETest(
+			t, vsphere, framework.WithClusterName(test.NewWorkloadClusterName()),
+		).WithClusterConfig(
+			api.ClusterToConfigFiller(
+				api.WithManagementCluster(managementCluster.ClusterName),
+				api.WithControlPlaneCount(1),
+				api.WithWorkerNodeCount(1),
+				api.WithStackedEtcdTopology(),
+			),
+			vsphere.WithUbuntu121(),
```
vivek-koppuru, line 54: This also updates the kubernetes version in the cluster spec?

test/framework/docker.go
```diff
@@ -35,6 +41,18 @@ func (d *Docker) CleanupVMs(_ string) error {
 	return nil
 }
 
+// UpdateKubeConfig customizes generated kubeconfig for the provider. This is required for the docker provider.
```
vivek-koppuru, line 44: How does it customize the kubeconfig? Maybe explain that here?

test/e2e/multiple_worker_node_groups_test.go
cxbrowne1207: We thought that it might be a good idea to increase coverage without increasing resources for this particular case upgrading a self managed cluster with the api.

I think I need to decouple the upgrades though. Shouldn't be scaling the node group and adding a labels and taints to one worker.

test/e2e/workload_api_test.go
cxbrowne1207: Yes, it updates the kubernetes version on the cluster and all of the VSphereMachineConfigs

vivek-koppuru (approved):

PR #4563, opened 2023-01-06
Build Tinkerbell datacenter controller

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Build Tinkerbell datacenter controller
The controller does not call the Tinkerbell IP validations yet.

*Testing (if applicable):*

### DIFF
--- a/controllers/factory_test.go
+++ b/controllers/factory_test.go
@@ -52,6 +52,26 @@ func TestFactoryBuildAllDockerReconciler(t *testing.T) {
 	g.Expect(reconcilers.DockerDatacenterReconciler).NotTo(BeNil())
 }
 
+func TestFactoryBuildAllTinkerbellReconciler(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	logger := nullLog()
+	ctrl := gomock.NewController(t)
+	manager := mocks.NewMockManager(ctrl)
+	manager.EXPECT().GetClient().AnyTimes()
+	manager.EXPECT().GetScheme().AnyTimes()
+
+	f := controllers.NewFactory(logger, manager).
+		WithTinkerbellDatacenterReconciler()
+
+	// testing idempotence
+	f.WithTinkerbellDatacenterReconciler()
+
+	reconcilers, err := f.Build(ctx)
+	g.Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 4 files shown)


controllers/tinkerbell_datacenter_controller.go
```diff
@@ -0,0 +1,20 @@
+package controllers
+
+import (
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TinkerbellDatacenterReconciler reconciles a TinkerbellDatacenterConfig object.
+type TinkerbellDatacenterReconciler struct {
```
ddjjia, line 9: Does datacenter controller need a validator?

controllers/factory_test.go
```diff
@@ -52,6 +52,26 @@ func TestFactoryBuildAllDockerReconciler(t *testing.T) {
...
+
+	f := controllers.NewFactory(logger, manager).
+		WithTinkerbellDatacenterReconciler()
+
+	// testing idempotence
+	f.WithTinkerbellDatacenterReconciler()
+
+	reconcilers, err := f.Build(ctx)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(reconcilers.TinkerbellDatacenterReconciler).NotTo(BeNil())
```
cxbrowne1207, line 72: How is this testing idemptotence? Would the factory `Build` method return an err if there is more than one `WithTinkerbellDatacenterReconciler` registered?

controllers/tinkerbell_datacenter_controller.go
mitalipaygude: we will include the validations once we start the implementation, this is more of a skeleton to build the controller.

controllers/tinkerbell_datacenter_controller.go
```diff
@@ -0,0 +1,20 @@
...
+	log    logr.Logger
```
g-gaston, line 10: better to extract the logger from the context in the way the cluster controller does it
That ensures we inherit the values set controller-runtime

controllers/factory_test.go
mitalipaygude: yes it would.

g-gaston (approved):

PR #4567, opened 2023-01-06
E2E Tagging busybox image with tag for consistency

junshun:
E2E test failures for metric server and autoscaler caused by new busybox image not triggering the scaling targets for test. Tagging to an older known working version and to avoid using the latest image.


jonathanmeier5 (approved):
/lgtm
/approve

PR #4569, opened 2023-01-06
Move Tinkerbell datacenter config validations

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Move Tinkerbell datacenter config validations. This makes all datacenter config validations invokable from once place for webhooks.
*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/validate.go
+++ b/pkg/providers/tinkerbell/validate.go
@@ -3,17 +3,32 @@ package tinkerbell
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 
 	tinkv1alpha1 "github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/logger"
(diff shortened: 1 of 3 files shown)


pkg/providers/tinkerbell/validate.go
```diff
@@ -3,17 +3,32 @@ package tinkerbell
...
+		if _, err := url.ParseRequestURI(config.Spec.OSImageURL); err != nil {
+			return fmt.Errorf("parsing osImageOverride: %v", err)
+		}
+	}
+
+	if config.Spec.HookImagesURLPath != "" {
+		if _, err := url.ParseRequestURI(config.Spec.HookImagesURLPath); err != nil {
+			return fmt.Errorf("parsing hookOverride: %v", err)
+		}
+		logger.Info("hook path override set", "path", config.Spec.HookImagesURLPath)
```
pokearu, line 29: If we going to call this from a webhook, can we use logger? Or is it better to just return errors?

mitalipaygude: I see [this](https://github.com/aws/eks-anywhere/blob/74af3394729dab77473e9435f254a0f708572abe/pkg/api/v1alpha1/vspheremachineconfig.go#L123) mutation webhook using logger, so I can keep it for now, and then remove it (if reqd) during the webhook implementation by testing it.

vivek-koppuru (approved):

PR #4574, opened 2023-01-07
Update worker node validation for API e2e tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Previous worker node validation didn't work for BottleRocket. This PR updates it and also does some refactoring.

*Testing (if applicable):*
```
--- PASS: TestVSphereUpgradeLabelsTaintsBottleRocketAPI (1615.58s)
PASS
```

### DIFF
--- a/test/e2e/workload_api_test.go
+++ b/test/e2e/workload_api_test.go
@@ -100,7 +100,7 @@ func TestVSphereMulticlusterWorkloadClusterAPI(t *testing.T) {
 	test.DeleteManagementCluster()
 }
 
-func TestVSphereUpgradeLabelsTaintsUbuntuAPI(t *testing.T) {
+func TestVSphereUpgradeLabelsTaintsBottleRocketAPI(t *testing.T) {
 	vsphere := framework.NewVSphere(t)
 
 	managementCluster := framework.NewClusterE2ETest(
@@ -111,7 +111,7 @@ func TestVSphereUpgradeLabelsTaintsUbuntuAPI(t *testing.T) {
 			api.WithWorkerNodeCount(1),
 			api.WithStackedEtcdTopology(),
 		),
-		vsphere.WithUbuntu124(),
+		vsphere.WithBottleRocket124(),
 	)
 
 	test := framework.NewMulticlusterE2ETest(t, managementCluster)
@@ -128,7 +128,7 @@ func TestVSphereUpgradeLabelsTaintsUbuntuAPI(t *testing.T) {
 			vsphere.WithWorkerNodeGroup("worker-0", framework.WithWorkerNodeGroup("worker-0", api.WithCount(2), api.WithLabel("key1", "val2"), api.WithTaint(framework.NoScheduleTaint()))),
 			vsphere.WithWorkerNodeGroup("worker-1", framework.WithWorkerNodeGroup("worker-1", api.WithCount(1))),
 			vsphere.WithWorkerNodeGroup("worker-2", framework.WithWorkerNodeGroup("worker-2", api.WithCount(1), api.WithLabel("key2", "val2"), api.WithTaint(framework.PreferNoScheduleTaint()))),
-			vsphere.WithUbuntu124(),
+			vsphere.WithBottleRocket124(),
 		),
 	)
 
--- a/test/framework/clustervalidator.go
+++ b/test/framework/clustervalidator.go
@@ -173,7 +173,7 @@ func validateWorkerNodes(ctx context.Context, vc ClusterValidatorConfig) error {
 	// deduce the worker node group configuration to node mapping via the machine deployment and machine set
 	for _, w := range wn {
 		workerGroupCount := 0
-		ms, err := GetWorkerNodeMachineSets(ctx, vc, w)
+		ms, err := getWorkerNodeMachineSets(ctx, vc, w)
 		if err != nil {
 			return fmt.Errorf("failed to get machine sets when validating worker node: %v", err)
 		}
@@ -226,8 +226,8 @@ func validateClusterDoesNotExist(ctx context.Context, vc ClusterValidatorConfig)
 	return nil
 }
 
-// GetWorkerNodeMachineSets gets a list of MachineSets corresponding the provided WorkerNodeGroupConfiguration from the management cluster.
-func GetWorkerNodeMachineSets(ctx context.Context, vc ClusterValidatorConfig, w v1alpha1.WorkerNodeGroupConfiguration) ([]v1beta1.MachineSet, error) {
+// getWorkerNodeMachineSets gets a list of MachineSets corresponding the provided WorkerNodeGroupConfiguration from the management cluster.
+func getWorkerNodeMachineSets(ctx context.Context, vc ClusterValidatorConfig, w v1alpha1.WorkerNodeGroupConfiguration) ([]v1beta1.MachineSet, error) {
 	md := &v1beta1.MachineDeployment{}
 	mdName := fmt.Sprintf("%s-%s", vc.ClusterSpec.Cluster.Name, w.Name)
 	key := types.NamespacedName{Name: mdName, Namespace: constants.EksaSystemNamespace}
--- a/test/framework/clustervalidator.go
+++ b/test/framework/clustervalidator.go
@@ -14,26 +14,31 @@ import (
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/controller"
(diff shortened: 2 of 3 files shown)


test/framework/clustervalidator.go
```diff
@@ -109,79 +145,84 @@ func validateEKSAObjects(ctx context.Context, validateOpts clusterOpts) error {
...
 	if err != nil {
 		return fmt.Errorf("failed to retrieve cluster %s", err)
 	}
-
 	if capiCluster == nil {
-		return fmt.Errorf("cluster %s does not exist", validateOpts.ClusterSpec.Cluster.Name)
+		return fmt.Errorf("cluster %s does not exist", clus.Name)
 	}
-
-	if err := checkClusterReady(capiCluster); err != nil {
```
tatlat, line 135: Maybe it's because it's Monday, but I don't see the difference between this and what replaced it. Is there a reason for the replacement?

test/framework/clustervalidator.go
```diff
@@ -88,19 +93,50 @@ func NewClusterValidator(opts ...ClusterValidatorOpts) *ClusterValidator {
...
+// If the cluster is self-managed, it returns the clusters own client.
+func (c *ClusterValidatorConfig) GetManagementClusterClient() client.Client {
+	if c.ClusterSpec.Cluster.IsManaged() {
+		return c.ManagementClusterClient
 	}
+	return c.ClusterClient
+}
 
-	for _, obj := range validateOpts.ClusterSpec.ChildObjects() {
+// TODO: Move Validations to separate package
```
taneyland, line 134: Since we are already altering all of the validations in this PR, can we go ahead & move it to the other package in this PR? Or will that involve a significant amount of additional changes?

test/framework/clustervalidator.go
```diff
@@ -109,79 +145,84 @@ func validateEKSAObjects(ctx context.Context, validateOpts clusterOpts) error {
...
 		}
 	}
+	logger.V(4).Info("Control plane nodes validated", "cluster", clus.Name)
 	return nil
 }
 
-// TODO: Update this to also validate BottleRocket nodes
-func validateWorkerNodes(ctx context.Context, validateOpts clusterOpts) error {
-	var workerNodes []corev1.Node
+func validateWorkerNodes(ctx context.Context, vc ClusterValidatorConfig) error {
```
taneyland, line 196: Are we removing the worker node count validation?

test/framework/clustervalidator.go
cxbrowne1207: There is no difference really. I got rid of the function wrapping the logic entirely because it wasn't being used anywhere else.

taneyland:
Should we update one of the existing API tests or add one that uses Bottlerocket as the osFamily to ensure that this works for the Bottlerocket tests?

test/framework/clustervalidator.go
cxbrowne1207: My thoughts were to move the validations and the cluster validator out of the test folder as well. If we do that, then unit tests would be required for all of the validations, so I'm thinking it would be significant.

test/framework/clustervalidator.go
cxbrowne1207: Ah, thanks for catching that. Wasn't my intention, I'll add that back in.

test/framework/clustervalidator.go
taneyland: Ok interesting! So since you're thinking to move it outside of the test folder, are we thinking to use the cluster validator for more than just e2e tests?

cxbrowne1207: Yeah, that's what I'm thinking

taneyland: Cool, thanks!

cxbrowne1207:
> Should we update one of the existing API tests or add one that uses Bottlerocket as the osFamily to ensure that this works for the Bottlerocket tests?

Actually, I had tested this locally. But you're right, I should include it. I think I'll go ahead and add all the BR tests to this one.

test/e2e/workload_api_test.go
```diff
@@ -100,7 +100,7 @@ func TestVSphereMulticlusterWorkloadClusterAPI(t *testing.T) {
 	test.DeleteManagementCluster()
 }
 
-func TestVSphereUpgradeLabelsTaintsUbuntuAPI(t *testing.T) {
```
mitalipaygude, line 103: Are we not testing for Ubuntu at all then?

test/framework/clustervalidator.go
```diff
@@ -88,19 +93,50 @@ func NewClusterValidator(opts ...ClusterValidatorOpts) *ClusterValidator {
...
 
-func validateEKSAObjects(ctx context.Context, validateOpts clusterOpts) error {
-	clusterClient := validateOpts.ClusterClient
-	if validateOpts.ClusterSpec.Cluster.IsManaged() {
-		clusterClient = validateOpts.ManagementClusterClient
+// GetWorkerNodeMachineSets gets a list of MachineSets corresponding the provided WorkerNodeGroupConfiguration from the management cluster.
+func (c *ClusterValidatorConfig) GetWorkerNodeMachineSets(ctx context.Context, w v1alpha1.WorkerNodeGroupConfiguration) ([]v1beta1.MachineSet, error) {
+	md := &v1beta1.MachineDeployment{}
+	mgmtClusterClient := c.GetManagementClusterClient()
+	mdName := fmt.Sprintf("%v-%v", c.ClusterSpec.Cluster.Name, w.Name)
```
mitalipaygude, line 107: nit: we can use %s here as both will be strings.

test/framework/clustervalidator.go
```diff
@@ -88,19 +93,50 @@ func NewClusterValidator(opts ...ClusterValidatorOpts) *ClusterValidator {
...
+	}
+	ms := &v1beta1.MachineSetList{}
+	err := mgmtClusterClient.List(ctx, ms, client.InNamespace(constants.EksaSystemNamespace), client.MatchingLabels{
+		"cluster.x-k8s.io/deployment-name": md.Name,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get machine sets for deployment %s: %v", md.Name, err)
+	}
+	if len(ms.Items) == 0 {
+		return nil, fmt.Errorf("invalid number of machine sets associated with worker node configuration %v", w.Name)
```
mitalipaygude, line 120: nit: same as above for %s

test/framework/clustervalidator.go
```diff
@@ -109,79 +145,89 @@ func validateEKSAObjects(ctx context.Context, validateOpts clusterOpts) error {
...
+		for _, node := range nodes.Items {
+			ownerName, ok := node.Annotations["cluster.x-k8s.io/owner-name"]
+			if ok {
+				// there will be multiple machineSets present on a cluster following an upgrade.
+				// find the one that is associated with this worker node, and execute the validations.
+				for _, machineSet := range ms {
+					if ownerName == machineSet.Name {
+						workerGroupCount++
+						if err := validateNodeReady(node, vc.ClusterSpec.Cluster.Spec.KubernetesVersion); err != nil {
+							return fmt.Errorf("failed to validate worker node ready %s", err)
```
mitalipaygude, line 220:
```suggestion
							return fmt.Errorf("failed to validate worker node ready %v", err)
```

test/e2e/workload_api_test.go
cxbrowne1207: Yes. We had decided to update some API tests to use BR rather than create a new test for each.

g-gaston (commented):
The validator/validations is growing quite a lot and increasing in complexity. We might want to start writing some unit tests. If not, eventually it will become too difficult to change.

test/framework/clustervalidator.go
```diff
@@ -88,19 +93,50 @@ func NewClusterValidator(opts ...ClusterValidatorOpts) *ClusterValidator {
...
 	ManagementClusterClient client.Client // the client for the management cluster
 	ClusterSpec             *cluster.Spec // the cluster spec
 }
 
-func validateEKSAObjects(ctx context.Context, validateOpts clusterOpts) error {
-	clusterClient := validateOpts.ClusterClient
-	if validateOpts.ClusterSpec.Cluster.IsManaged() {
-		clusterClient = validateOpts.ManagementClusterClient
+// GetWorkerNodeMachineSets gets a list of MachineSets corresponding the provided WorkerNodeGroupConfiguration from the management cluster.
+func (c *ClusterValidatorConfig) GetWorkerNodeMachineSets(ctx context.Context, w v1alpha1.WorkerNodeGroupConfiguration) ([]v1beta1.MachineSet, error) {
```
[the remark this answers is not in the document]
cxbrowne1207: I see, I think I'll make it a standalone method for now. I also had planned to move the cluster validator and validations into separate packages outside of the tests as a part of another task soon. This one would have unit-tests associated with them.

taneyland (approved):

PR #4577, opened 2023-01-09
Add Docker API E2E tests for scaling nodes and worker node groups

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR adds an e2e test for the docker API  or scaling nodes and add/remove worker node group configurations.

*Testing (if applicable):*
`--- PASS: TestDockerUpgradeKubernetes123to124WorkloadClusterScaleupAPI (2017.01s)
`
`--- PASS: TestDockerUpgradeWorkloadClusterScaleAddRemoveWorkerNodeGroupsAPI (1167.16s)
`

### DIFF
--- a/test/e2e/workload_api_test.go
+++ b/test/e2e/workload_api_test.go
@@ -182,7 +182,7 @@ func TestVSphereUpgradeScaleWorkersUbuntuAPI(t *testing.T) {
 	)
 }
 
-func TestDockerUpgradeKubernetes123to124WorkloadClusterScaleWorkerNodesAPI(t *testing.T) {
+func TestDockerUpgradeKubernetes123to124WorkloadClusterScaleupAPI(t *testing.T) {
 	provider := framework.NewDocker(t)
 	managementCluster := framework.NewClusterE2ETest(
 		t, provider,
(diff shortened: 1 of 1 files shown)


test/e2e/workload_api_test.go
```diff
@@ -204,8 +204,7 @@ func TestDockerUpgradeKubernetes123to124WorkloadClusterScaleWorkerNodesAPI(t *te
 				api.WithManagementCluster(managementCluster.ClusterName),
 				api.WithControlPlaneCount(1),
 				api.RemoveAllWorkerNodeGroups(), // This gives us a blank slate
-				api.WithWorkerNodeGroup("worker-0", api.WithCount(2)),
-				api.WithWorkerNodeGroup("worker-1", api.WithCount(1)),
+				api.WithWorkerNodeGroup("worker-0", api.WithCount(1)),
```
taneyland, line 207: Why are we changing this to remove the scale down case?

cxbrowne1207: This one was changed to primarily test the k8s upgrade, so I figured scaling up is all that's needed. The new test added tests scaling up and down, as well as adding / remove worker node groups,

test/e2e/workload_api_test.go
```diff
@@ -262,3 +260,44 @@ func TestDockerUpgradeWorkloadClusterLabelsAndTaintsAPI(t *testing.T) {
...
+	test := framework.NewMulticlusterE2ETest(t, managementCluster)
+	test.WithWorkloadClusters(
+		framework.NewClusterE2ETest(
+			t, provider, framework.WithClusterName(test.NewWorkloadClusterName()),
+		).WithClusterConfig(
+			api.ClusterToConfigFiller(
+				api.WithKubernetesVersion(v1alpha1.Kube124),
+				api.WithManagementCluster(managementCluster.ClusterName),
+				api.WithControlPlaneCount(1),
+				api.RemoveAllWorkerNodeGroups(), // This gives us a blank slate
```
mitalipaygude, line 285: Out of curiosity, why do we need to remove all worker nodes?

cxbrowne1207: There is a worker node group defined by default in the base configurations. So removing all worker node groups gives a clean slate.

PR #4592, opened 2023-01-10
Fix snow flaky test

jiayiwang7:
*Issue #, if available:*

`TestUpgradeNeededMachineConfigChanged` fails half of the time because the method being tested loop a machineconfig map and order of the machine config is not fixed.

*Description of changes:*

Loose check on machine config of any name and validate.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/snow/snow_test.go
+++ b/pkg/providers/snow/snow_test.go
@@ -977,18 +977,7 @@ func TestUpgradeNeededMachineConfigChanged(t *testing.T) {
 	tt.kubeconfigClient.EXPECT().
 		Get(
 			tt.ctx,
-			"test-cp",
-			"test-namespace",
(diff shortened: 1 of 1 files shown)


pkg/providers/snow/snow_test.go
```diff
@@ -977,18 +977,7 @@ func TestUpgradeNeededMachineConfigChanged(t *testing.T) {
 	tt.kubeconfigClient.EXPECT().
 		Get(
 			tt.ctx,
```
TerryHowe, line 979: I think this will normally call kubectl the 3 times

jiayiwang7: only once. 3 times need `Times(3)`

TerryHowe (approved):
/lgtm
/approve

PR #4596, opened 2023-01-10
Add validation for management cluster referenced in workload cluster spec to be valid

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Added additional validations to ensure that we are able to retrieve the EKS-A cluster object from the management cluster and also confirm that the cluster we have referenced is indeed a management cluster by checking if the management cluster name for that cluster references itself.

*Testing (if applicable):*
unit tests and testing these scenarios functionally

### DIFF
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -1,6 +1,7 @@
 package validations
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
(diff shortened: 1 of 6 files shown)


pkg/validations/cluster.go
```diff
@@ -56,6 +58,18 @@ func ValidateAuthenticationForRegistryMirror(clusterSpec *cluster.Spec) error {
...
 }
 
+// ValidateManagementClusterName checks if the management cluster specified in the workload cluster spec is valid.
+func ValidateManagementClusterName(ctx context.Context, k KubectlClient, mgmtCluster *types.Cluster, mgmtClusterName string) error {
+	cluster, err := k.GetEksaCluster(ctx, mgmtCluster, mgmtClusterName)
+	if err != nil {
+		return err
+	}
+	if cluster.Name != cluster.Spec.ManagementCluster.Name {
+		return fmt.Errorf("%v is not a valid management cluster", mgmtClusterName)
```
mitalipaygude, line 68: nit: we could use %s for the string

cxbrowne1207 (approved):

mitalipaygude (approved):

PR #4597, opened 2023-01-10
Add bundle and charts to `copy packages`

TerryHowe:
I removed copy package bundles and helm charts from the previous PR to reduce the complexity of the last PR. The previous PR only copies the images. This PR adds the bundles and charts to the copy.
* Deprecates --include-packages and removes support from import images (won't work anyway)
* Deprecates --include-package and removes support from download images (won't work anyway)
* Meeting test coverage criteria was not easy; does not entirely make sense to me how that is calculated.


pkg/curatedpackages/reader.go
```diff
@@ -28,16 +30,25 @@ type ManifestReader interface {
...
 		ManifestReader: mr,
+		AllImages:      true,
 	}
 }
 
 func (r *PackageReader) ReadImagesFromBundles(ctx context.Context, b *releasev1.Bundles) ([]releasev1.Image, error) {
-	images, err := r.ManifestReader.ReadImagesFromBundles(ctx, b)
+	var err error
+	var images []releasev1.Image
+	if r.AllImages {
```
ewollesen, line 49: It's unclear what the difference is between ReadImagesFromBundles and fetchImagesFromBundles.

I'm guessing the former are for curated packages, while the latter are for EKS-A core images?

If so, perhaps some renaming could make it clearer what's going on:

```
r.AllImages => r.IncludePackagesImages
r.ManifestReader.ReadImagesFromBundles => ReadPackagesImagesFromBundles
```

pkg/curatedpackages/reader.go
```diff
@@ -53,29 +64,46 @@ func (r *PackageReader) ReadImagesFromBundles(ctx context.Context, b *releasev1.
...
+		images = append(images, eksaCharts...)
+	}
+
+	// Generate bundle repository
+	for _, chart := range eksaCharts {
+		if strings.Contains(chart.Repository(), "eks-anywhere-packages") {
+			var bundleRepository releasev1.Image
+			chart.DeepCopyInto(&bundleRepository)
+			bundleRepository.URI = strings.Replace(bundleRepository.URI, "eks-anywhere-packages", "eks-anywhere-packages-bundles", 1)
+			bundleRepository.URI = strings.Replace(bundleRepository.URI, bundleRepository.Version(), "v1-24-latest", 1)
```
ewollesen, line 83: Hard coding this value will likely come back to bite us.

pkg/curatedpackages/reader.go
```diff
@@ -53,29 +64,46 @@ func (r *PackageReader) ReadImagesFromBundles(ctx context.Context, b *releasev1.
...
+			bundleRepository.ImageDigest = ""
```
ewollesen, line 84: Minor: this feels like a nice function to extract.

pkg/curatedpackages/reader.go
```diff
@@ -122,13 +151,29 @@ func (r *PackageReader) fetchImagesFromBundle(ctx context.Context, versionsBundl
...
 				URI:         fmt.Sprintf("%s/%s@%s", getRegistry(ctrl.URI), version.Repository, version.Digest),
+				ImageDigest: version.Digest,
 			}
 			images = append(images, image)
 		}
 	}
 	return images, nil
 }
 
+func removeDuplicateImages(strSlice []releasev1.Image) []releasev1.Image {
```
ewollesen, line 162:
```suggestion
func removeDuplicateImages(images []releasev1.Image) []releasev1.Image {
```

pkg/curatedpackages/reader.go
```diff
@@ -122,13 +151,29 @@ func (r *PackageReader) fetchImagesFromBundle(ctx context.Context, versionsBundl
...
+	allKeys := make(map[string]releasev1.Image)
+	var list []releasev1.Image
```
ewollesen, line 164:
```suggestion
	var uniqImages []releasev1.Image
```

pkg/curatedpackages/reader.go
```diff
@@ -122,13 +151,29 @@ func (r *PackageReader) fetchImagesFromBundle(ctx context.Context, versionsBundl
...
+				ImageDigest: version.Digest,
 			}
 			images = append(images, image)
 		}
 	}
 	return images, nil
 }
 
+func removeDuplicateImages(strSlice []releasev1.Image) []releasev1.Image {
+	allKeys := make(map[string]releasev1.Image)
```
ewollesen, line 163: Nit: there's no need to copy the images into the map. You could just use:
```go
map[string]struct{}
```
Which wouldn't have to copy as many things around, saving both memory and time.
That would change line 167 to something more like: 
```go
allKeys[item.VersionedImage()] = struct{}{}
```

pkg/curatedpackages/reader.go
TerryHowe: I agree this code is odd, but I don't want to refactor it.

pkg/curatedpackages/reader.go
TerryHowe: Done.

pkg/curatedpackages/reader.go
TerryHowe: Done.

pkg/curatedpackages/reader.go
TerryHowe: Done.

pkg/curatedpackages/reader.go
TerryHowe: Another time perhaps.

pkg/curatedpackages/reader.go
TerryHowe: Yes.

TerryHowe: Fixed

cmd/eksctl-anywhere/cmd/downloadimages.go
```diff
@@ -43,7 +41,8 @@ func init() {
 		log.Fatalf("Cannot mark 'output' flag as required: %s", err)
 	}
 
-	downloadImagesCmd.Flags().BoolVar(&downloadImagesRunner.includePackages, "include-packages", false, "Flag to indicate inclusion of curated packages in downloaded images")
+	downloadImagesCmd.Flags().BoolVar(&downloadImagesRunner.includePackages, "include-packages", false, "Flag no long works, use copy packages")
```
ewollesen, line 44:
```suggestion
	downloadImagesCmd.Flags().BoolVar(&downloadImagesRunner.includePackages, "include-packages", false, "this flag no longer works, use copy packages instead")
```

cmd/eksctl-anywhere/cmd/import_images.go
```diff
@@ -48,6 +48,7 @@ func init() {
 		log.Fatalf("Cannot mark 'bundles' as required: %s", err)
 	}
 	importImagesCmd.Flags().BoolVar(&importImagesCommand.includePackages, "include-packages", false, "Flag to indicate inclusion of curated packages in imported images")
+	importImagesCmd.Flag("include-packages").Deprecated = "Use copy packages command"
```
ewollesen, line 51:
```suggestion
	importImagesCmd.Flag("include-packages").Deprecated = "use copy packages command"
```
Just for consistency with the other deprecation on line 45 of downloadimages.go

ewollesen (commented):
/lgtm

pkg/curatedpackages/reader.go
```diff
@@ -28,105 +30,119 @@ type ManifestReader interface {
 
 type PackageReader struct {
 	ManifestReader
+	AllImages bool
```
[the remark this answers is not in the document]
TerryHowe: Deleted.

pkg/curatedpackages/packagereader_test.go
```diff
@@ -2,183 +2,166 @@ package curatedpackages_test
...
+	bundles.Spec.VersionsBundles[0].KubeVersion = "1"
 
 	images, err := tt.command.ReadImagesFromBundles(tt.ctx, bundles)
 
 	tt.Expect(err).To(BeNil())
 	tt.Expect(images).To(BeEmpty())
 }
 
-func TestPackageReaderReadImagesFromBundlesFailWhenWrongBundle(t *testing.T) {
+func TestPackageReader_ReadImagesFromBundlesBadRegistry(t *testing.T) {
```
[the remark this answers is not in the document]
TerryHowe: Yeh, it just hits the continue on failures and logs a message. I do not want to change that behavior right now.

pkg/curatedpackages/reader.go
```diff
@@ -28,105 +30,117 @@ type ManifestReader interface {
...
-	ctrl := versionsBundle.PackageController.Controller
-	bundle := &packagesv1.PackageBundle{}
-	err = yaml.Unmarshal(data, bundle)
+	bundle := packagesv1.PackageBundle{}
+	err = yaml.Unmarshal(data, &bundle)
 	if err != nil {
-		return nil, err
+		return "", nil, err
 	}
+	return artifact.VersionedImage(), &bundle, nil
```
a-cool-train, line 97: What is the artifact.VersionedImage() for? Wouldn't returning just the bundle and the error suffice for this method?

TerryHowe: The bundle does not contain the bundles location.

PR #4603, opened 2023-01-11
Add docker API E2E test upgrade from latest minor release

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR adds a docker API E2E test for upgrading from the latest minor release test. The runs through the flow of upgrading a workload cluster to a new version of EKS-A through the API after first upgrading the management cluster.

*Testing (if applicable):*
```
--- PASS: TestDockerKubernetes123to124UpgradeFromLatestMinorReleaseAPI (1291.08s)
PASS
```

### DIFF
--- a/test/e2e/upgrade_from_latest_test.go
+++ b/test/e2e/upgrade_from_latest_test.go
@@ -24,6 +24,35 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
 	test.DeleteCluster()
 }
 
+func runMulticlusterUpgradeFromReleaseFlowAPI(test *framework.MulticlusterE2ETest, latestRelease *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, opts ...framework.ClusterE2ETestOpt) {
+	test.CreateManagementClusterForVersion(latestRelease.Version, framework.ExecuteWithEksaRelease(latestRelease))
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.ApplyClusterManifest()
+		wc.WaitForKubeconfig()
+		wc.ValidateClusterState()
+	})
+	// Adding this manual wait because old versions of the cli don't wait long enough
+	// after creation, which makes the upgrade preflight validations fail
+	test.ManagementCluster.WaitForControlPlaneReady()
+	test.ManagementCluster.UpgradeCluster(opts)
+	test.ManagementCluster.ValidateCluster(wantVersion)
+	test.ManagementCluster.StopIfFailed()
+	cluster := test.ManagementCluster.GetEKSACluster()
+	// Upgrade bundle workload clusters now because they still have the old versions of the bundle.
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.UpdateClusterConfig(
+			api.ClusterToConfigFiller(
+				api.WithBundlesRef(cluster.Spec.BundlesRef.Name, cluster.Spec.BundlesRef.Namespace, cluster.Spec.BundlesRef.APIVersion),
+			),
+		)
+		wc.ApplyClusterManifest()
+		wc.ValidateClusterState()
+		wc.DeleteClusterWithKubectl()
+		wc.ValidateClusterDelete()
+	})
+	test.DeleteManagementCluster()
+}
+
 func runUpgradeWithFluxFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, clusterOpts ...framework.ClusterE2ETestOpt) {
 	test.GenerateClusterConfigForVersion(latestRelease.Version, framework.ExecuteWithEksaRelease(latestRelease))
 	test.CreateCluster(framework.ExecuteWithEksaRelease(latestRelease))
--- a/test/e2e/upgrade_from_latest_test.go
+++ b/test/e2e/upgrade_from_latest_test.go
@@ -24,18 +24,15 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
 	test.DeleteCluster()
 }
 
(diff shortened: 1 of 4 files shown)


test/e2e/upgrade_from_latest_test.go
```diff
@@ -24,6 +24,35 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
...
+func runMulticlusterUpgradeFromReleaseFlowAPI(test *framework.MulticlusterE2ETest, latestRelease *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, opts ...framework.ClusterE2ETestOpt) {
+	test.CreateManagementClusterForVersion(latestRelease.Version, framework.ExecuteWithEksaRelease(latestRelease))
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.ApplyClusterManifest()
+		wc.WaitForKubeconfig()
+		wc.ValidateClusterState()
+	})
+	// Adding this manual wait because old versions of the cli don't wait long enough
+	// after creation, which makes the upgrade preflight validations fail
+	test.ManagementCluster.WaitForControlPlaneReady()
```
vivek-koppuru, line 36: Do we still need this?

test/e2e/upgrade_from_latest_test.go
```diff
@@ -24,6 +24,35 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
...
+	test.ManagementCluster.UpgradeCluster(opts)
+	test.ManagementCluster.ValidateCluster(wantVersion)
+	test.ManagementCluster.StopIfFailed()
+	cluster := test.ManagementCluster.GetEKSACluster()
+	// Upgrade bundle workload clusters now because they still have the old versions of the bundle.
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.UpdateClusterConfig(
+			api.ClusterToConfigFiller(
+				api.WithBundlesRef(cluster.Spec.BundlesRef.Name, cluster.Spec.BundlesRef.Namespace, cluster.Spec.BundlesRef.APIVersion),
```
vivek-koppuru, line 45: So we are reading bundle for the management cluster because the management cluster has the latest but the workload cluster doesn't? Where do we define the one for the workload cluster again?

test/e2e/upgrade_from_latest_test.go
g-gaston: I don't think so. And if we do, it should go before creating the workload clusters

test/e2e/workload_api_test.go
```diff
@@ -301,3 +301,35 @@ func TestDockerUpgradeWorkloadClusterScaleAddRemoveWorkerNodeGroupsAPI(t *testin
...
 	)
 }
+
+func TestDockerKubernetes123to124UpgradeFromLatestMinorReleaseAPI(t *testing.T) {
+	release := latestMinorRelease(t)
+	provider := framework.NewDocker(t)
+	managementCluster := framework.NewClusterE2ETest(
+		t,
+		provider,
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube123)),
```
g-gaston, line 311: Pls use `WithClusterConfig` for this

test/e2e/upgrade_from_latest_test.go
```diff
@@ -24,6 +24,35 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
...
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		wc.ApplyClusterManifest()
+		wc.WaitForKubeconfig()
+		wc.ValidateClusterState()
+	})
+	// Adding this manual wait because old versions of the cli don't wait long enough
+	// after creation, which makes the upgrade preflight validations fail
+	test.ManagementCluster.WaitForControlPlaneReady()
+	test.ManagementCluster.UpgradeCluster(opts)
+	test.ManagementCluster.ValidateCluster(wantVersion)
```
g-gaston, line 38: can't we extract `wantVersion` from the cluster config?

test/e2e/upgrade_from_latest_test.go
```diff
@@ -24,6 +24,35 @@ func runUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *re
 	test.DeleteCluster()
 }
 
+func runMulticlusterUpgradeFromReleaseFlowAPI(test *framework.MulticlusterE2ETest, latestRelease *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, opts ...framework.ClusterE2ETestOpt) {
```
g-gaston, line 27:
```suggestion
func runMulticlusterUpgradeFromReleaseFlowAPI(test *framework.MulticlusterE2ETest, release *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, opts ...framework.ClusterE2ETestOpt) {
```

test/e2e/upgrade_from_latest_test.go
cxbrowne1207: Ah, let me double check. At some point I was experiencing upgrade preflight validations failure, so I had decided to include this.

test/e2e/upgrade_from_latest_test.go
cxbrowne1207: Yes. It define the bundles for the workload cluster anywhere. After, the management and workload clusters are created, they have the same bundles. However, when the management cluster is updated, it then has the latest but the workload cluster still has the old bundles. The flow then continues to update the workload clusters with the bundles referenced by the management cluster.

test/e2e/workload_api_test.go
cxbrowne1207: I used this for this test because `CreateManagementClusterForVersion` calls `GenerateClusterConfigForVersion` which seems to have some version cleanup logic in that function. Does that matter?

If not, I can simply go with using `WithClusterConfig`  and then use `test.CreateManagementCluster(framework.ExecuteWithEksaRelease(release))`

test/e2e/upgrade_from_latest_test.go
cxbrowne1207: Not needed after all, will remove.

vivek-koppuru (approved):

g-gaston (approved):

PR #4604, opened 2023-01-11
Use concrete structs for govc devices.info

abhinavmpandey08:
*Description of changes:*
Use predefined structs instead of generic `map[string]interface{}` to unmarshal `govc devices.info` response.

### DIFF
--- a/pkg/executables/govc.go
+++ b/pkg/executables/govc.go
@@ -227,18 +227,34 @@ func (g *Govc) ResizeDisk(ctx context.Context, datacenter, template, diskName st
 	return nil
 }
 
-func (g *Govc) DevicesInfo(ctx context.Context, datacenter, template string) (interface{}, error) {
+type deviceInfoResponse struct {
+	Devices []VirtualDevice
+}
+
+// VirtualDevice describes a virtual device for a VM.
+type VirtualDevice struct {
+	Name         string
+	DeviceInfo   deviceInfo
+	CapacityInKB float64
+}
+
+type deviceInfo struct {
+	Label string
+}
+
+// DevicesInfo returns device info for the provided virtual machine.
+func (g *Govc) DevicesInfo(ctx context.Context, datacenter, template string) ([]VirtualDevice, error) {
 	response, err := g.exec(ctx, "device.info", "-dc", datacenter, "-vm", template, "-json")
 	if err != nil {
 		return nil, fmt.Errorf("getting template device information: %v", err)
 	}
 
-	var devicesInfo map[string]interface{}
+	var devicesInfo deviceInfoResponse
 	err = yaml.Unmarshal(response.Bytes(), &devicesInfo)
 	if err != nil {
 		return nil, fmt.Errorf("unmarshalling devices info: %v", err)
 	}
-	return devicesInfo["Devices"], nil
+	return devicesInfo.Devices, nil
 }
 
 func (g *Govc) TemplateHasSnapshot(ctx context.Context, template string) (bool, error) {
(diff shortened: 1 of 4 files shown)


pkg/executables/govc_test.go
```diff
@@ -107,20 +110,8 @@ func newDeployTemplateTest(t *testing.T) *deployTemplateTest {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "eksa-unit-test",
-		},
-		Spec: v1alpha1.VSphereMachineConfigSpec{
-			Template: "/SDDC-Datacenter/vm/Templates/ubuntu-v1.19.8-eks-d-1-19-4-eks-a-0.0.1.build.38-amd64",
-		},
+		diskName:                 "disk-31000-1",
+		diskSize:                 20,
 	}
 }
```
abhinavmpandey08, line 116: This was unused

pkg/executables/govc.go
```diff
@@ -227,18 +227,34 @@ func (g *Govc) ResizeDisk(ctx context.Context, datacenter, template, diskName st
...
+	Name         string
+	DeviceInfo   deviceInfo
+	CapacityInKB float64
+}
+
+type deviceInfo struct {
+	Label string
+}
+
+// DevicesInfo returns the device info for te provided virtual machine.
```
ptrivedi, line 245:
```suggestion
// DevicesInfo returns the device info for the provided virtual machine.
```

pkg/executables/govc.go
```diff
@@ -315,14 +331,13 @@ func (g *Govc) DeployTemplateFromLibrary(ctx context.Context, templateDir, templ
...
+		for _, device := range devicesInfo {
+			deviceLabel := device.DeviceInfo.Label
 			// Get the name of the hard disk and resize the disk to 20G
 			if strings.EqualFold(deviceLabel, "Hard disk 1") {
-				disk1 = deviceInfo.(map[string]interface{})["Name"].(string)
+				disk1 = device.Name
 			} else if strings.EqualFold(deviceLabel, "Hard disk 2") {
-				disk2 = deviceInfo.(map[string]interface{})["Name"].(string)
+				disk2 = device.Name
 				break
```
ptrivedi, line 341: why is the break within the 'if else'? If the idea is that after the disk#2 is found we want to break, then are we sure it  is returned in sequence?

ptrivedi: Also the log message below on line 345/360 for the error case during resizeDisk needs to be fixed, it currently says 20G.. But disk2 is being resized to 22G. Need to use diskSizeInGB var there.

abhinavmpandey08: This is the same logic we had before. I think the idea is that if Disk 2 is there, then we want to resize disk 2. So it doesn't matter if there are other disks as well.

abhinavmpandey08: > Also the log message below on line 345/360 for the error case during resizeDisk needs to be fixed, it currently says 20G.. But disk2 is being resized to 22G. Need to use diskSizeInGB var there.

fixed

pkg/executables/govc.go
```diff
@@ -227,18 +227,34 @@ func (g *Govc) ResizeDisk(ctx context.Context, datacenter, template, diskName st
 	return nil
 }
 
-func (g *Govc) DevicesInfo(ctx context.Context, datacenter, template string) (interface{}, error) {
+type deviceInfoResponse struct {
+	Devices []VirtualDevice
+}
+
+// VirtualDevice describes a virtual device for a VM.
```
chrisdoherty4, line 234: Is there some reference material in VMware's documentation we can link to?

PR #4605, opened 2023-01-11
Fix package controller client test

TerryHowe:
This test was failing if you had REGISTRY_USERNAME etc set.


TerryHowe:
merged in another

PR #4606, opened 2023-01-11
prefer os.ReadFile, as it's not deprecated

ewollesen:
My (possibly newer than the original author's) version of gofumpt complains about this.


TerryHowe (approved):
/lgtm
/approve

PR #4607, opened 2023-01-11
Fixes a platform-dependent test

ewollesen:
This fix isn't much better, as it matches against the text in a generic error, but it should work regardless of platform.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


TerryHowe (approved):
/lgtm

PR #4614, opened 2023-01-12
Update usage of control plane node role

taneyland:
*Issue #, if available:*

*Description of changes:*
For kubernetes 1.25+, `node-role.kubernetes.io/master`is deprecated, but it is still the only default taint on control plane nodes for 1.22, so we must add this toleration for both of those options.

*Testing (if applicable):*

### DIFF
--- a/pkg/awsiamauth/config/aws-iam-authenticator.yaml
+++ b/pkg/awsiamauth/config/aws-iam-authenticator.yaml
@@ -92,22 +92,16 @@ spec:
       # run on the host network (don't depend on CNI)
       hostNetwork: true
 
-      # run on each master node
+      # run on each control plane node
       nodeSelector:
-{{- if .kubeVersion124 }}
-        node-role.kubernetes.io/control-plane: ""
-{{- else }}
-        node-role.kubernetes.io/master: ""
-{{- end }}
+      {{ .cpNodeSelector | indent 2 }}
       tolerations:
       - effect: NoSchedule 
         key: node-role.kubernetes.io/master
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/control-plane
       - key: CriticalAddonsOnly
         operator: Exists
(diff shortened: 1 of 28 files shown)


pkg/awsiamauth/template.go
```diff
@@ -46,6 +46,7 @@ func (t *TemplateBuilder) GenerateManifest(clusterSpec *cluster.Spec, clusterID
 		"backendMode":        strings.Join(clusterSpec.AWSIamConfig.Spec.BackendMode, ","),
 		"partition":          clusterSpec.AWSIamConfig.Spec.Partition,
 		"kubeVersion124":     clusterSpec.Cluster.Spec.KubernetesVersion == v1alpha1.Kube124,
+		"kubeVersion125":     clusterSpec.Cluster.Spec.KubernetesVersion == v1alpha1.Kube125,
```
mitalipaygude, line 49: Maybe we can use a Kube version comparison to check if the version is greater than or equal to 1.24. It might not be the best idea to keep adding versions here for the check.

mitalipaygude:
Looks good!

pkg/awsiamauth/config/aws-iam-authenticator.yaml
```diff
@@ -104,7 +104,7 @@ spec:
         key: node-role.kubernetes.io/master
       - key: CriticalAddonsOnly
         operator: Exists
-{{- if .kubeVersion124 }}
+{{- if or .kubeVersion124 .kubeVersion125 }}
```
g-gaston, line 107: can't we just add both tolerations?

pkg/awsiamauth/template.go
g-gaston: Can we just pass the label as a value and run the logic here?

taneyland: Yes, incorporated both suggestions!

PR #4617, opened 2023-01-12
Fix credentials and bundle fetch

TerryHowe:
Sorry about the size of this PR, but code coverage.
* The credential store was not being passed around to everyone that was using it causing some failures
* The bundle fetch was failing to get the bundle blob


pkg/curatedpackages/reader.go
```diff
@@ -143,7 +145,12 @@ func removeDuplicateImages(images []releasev1.Image) []releasev1.Image {
 	return list
 }
 
-func getRegistry(uri string) string {
+func getChartRegistry(_ string) string {
+	// For now, we only support prod as a source
+	return publicProdECR
```
a-cool-train, line 150: This wouldn't work for development then since the images might not be available in prod. Any reason we can't use dev registry?

a-cool-train: We could also have this on a follow up PR if that will be easier.

TerryHowe: Yeh, this could be better, but not the highest priority atm.

TerryHowe: https://github.com/aws/eks-anywhere/issues/4620

pkg/curatedpackages/reader.go
```diff
@@ -47,7 +49,7 @@ func (r *PackageReader) ReadImagesFromBundles(ctx context.Context, b *releasev1.
 	for _, vb := range b.Spec.VersionsBundles {
 		bundleURI, bundle, err := r.getBundle(ctx, vb)
 		if err != nil {
-			logger.Info("Warning: Failed getting bundle reference", "error", err)
+			logger.Info("Warning: Failed getting image reference", "error", err)
```
a-cool-train, line 52: nit: This error stems from a lack of a bundle and not an image.

TerryHowe: Fixed.

pkg/curatedpackages/reader.go
TerryHowe: Fixed.
Closes: https://github.com/aws/eks-anywhere/issues/4620

PR #4621, opened 2023-01-12
Setup default ssh key based on cluster name and uuid for snow

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

This is to address the issue where a user creates a new cluster but doesn't have access to the previously generated key by default as it is stored in AWS. There is no way to retrieve the key to access the nodes in those situations. Instead, we are adding the cluster name to the default key so that there are more unique default keys. We are also appending a generated uuid in the cases that you have multiple clusters from different admin machines with the same name for additional uniqueness. 

Also needed to move around the code a little to resolve a cyclic complexity lint issue, which doesn't like too many if statements in one function. 

*Testing (if applicable):*
unit test and functional testing

### DIFF
--- a/pkg/providers/snow/defaults.go
+++ b/pkg/providers/snow/defaults.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/uuid"
+
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/aws"
 	"github.com/aws/eks-anywhere/pkg/cluster"
@@ -16,6 +18,7 @@ type Defaulters struct {
 	clientRegistry ClientRegistry
 	writer         filewriter.FileWriter
 	keyGenerator   SshKeyGenerator
+	uuid           uuid.UUID
 }
 
 type SshKeyGenerator interface {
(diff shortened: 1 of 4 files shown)


vivek-koppuru:
/release-0.14 

@d8660091 this is a bug found from our snow related testing

pkg/providers/snow/defaults.go
```diff
@@ -42,12 +46,26 @@ func WithKeyGenerator(generator SshKeyGenerator) DefaultersOpt {
...
+// GenerateDefaultSSHKeys generates ssh key if it doesn't exist already.
+func (d *Defaulters) GenerateDefaultSSHKeys(ctx context.Context, machineConfigs map[string]*v1alpha1.SnowMachineConfig, clusterName string) error {
 	md := NewMachineConfigDefaulters(d)
 
 	for _, m := range machineConfigs {
-		if err := md.SetupDefaultSshKey(ctx, m); err != nil {
-			return err
+		if md.keyGenerated {
+			// Generate a random string so that other clusters with same name and devices won't conflict in case user doesn't have access to previous key
+			m.Spec.SshKeyName = md.defaultSSHKeyName(clusterName)
```
jiayiwang7, line 63: if a machineconfig has a `sshKeyName`, this line overrides it with a default key which should not be expected behavior

vivek-koppuru: I needed to move due to a lint issue that was complaining about cyclic complexity (too many if conditions). Will fix the current condition

jiayiwang7 (approved):

PR #4630, opened 2023-01-13
Patch erroneous machine deployment state check on upgrade

chrisdoherty4:
Preflights check if machine deployments are in an appropriate state before an upgrade can proceed. The check eneds to retrieve machines associated with the target cluster only, not all machine deployments.


lewisdiamond:
Failed to cherry pick to 0.14
/cherry-pick release-0.14

Issue #3467, opened 2022-09-28
Allow users to choose between `LinkedClone` or `FullClone` mode for vSphere provider


abhinavmpandey08:
**What would you like to be added**:
Add option in EKS Anywhere on vSphere to allow users to specify whether to use LinkedClone or FullClone mode for OS images.

**Why is this needed**:
Currently, EKS Anywhere on vSphere is hardcoded to use `LinkedClone` mode. This makes the VM provisioning times faster but doesn't allow users to configure the disk size on VMs.

Because of the way LinkedClone works, if the template has snapshots, then the VM size isn't configurable.
So, if the users want to modify the disk size, they have to delete the template snapshots for the VMs to fall back to FullClone.

abhinavmpandey08:
more context on Full vs Linked clone mode: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/fae6ef88467e608665e2902e2bb0aaeb4cee67ed/apis/v1alpha4/types.go#L43-L52

echel0n:
I agree that this is a important feature to be added in as I find the default 25GB disk space is not enough for some clusters and keeps causing disk pressure

abhinavmpandey08:
This ticket should be resolved with this PR https://github.com/aws/eks-anywhere/pull/4634
These changes will be available in v0.15 release

PR #4634, opened 2023-01-14
Allow users to specify cloneMode for vSphere

abhinavmpandey08:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/3467

*Description of changes:*
Currently, EKS-A for vSphere defaults to LinkedClone for cloning the node VMs. This falls back to FullClone if some conditions are not met. 
With this change, users will be able to explicitly set the cloneMode to either `fullClone` or `linkedClone`. 

Alternatively, they can also skip the cloneMode and EKS-A will determine which cloneMode to use based on the following parameters:
 - if the `VSphereMachineConfig.Spec.Template` has a snapshot and its diskSize matches the `VSphereMachineConfig.Spec.DiskGiB`, `linkedClone` is used.
- else `fullClone` is used.

### DIFF
--- a/pkg/providers/vsphere/defaults.go
+++ b/pkg/providers/vsphere/defaults.go
@@ -34,9 +34,11 @@ func (d *Defaulter) setDefaultsForMachineConfig(ctx context.Context, spec *Spec)
 			return err
 		}
 
-		if err := d.setDiskDefaults(ctx, m); err != nil {
+		if err := d.setCloneModeDefaults(ctx, m, spec.VSphereDatacenter.Spec.Datacenter); err != nil {
 			return err
 		}
+
+		d.setDiskSizeDefaults(m)
 	}
 
 	return nil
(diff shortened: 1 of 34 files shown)


pkg/providers/vsphere/defaults.go
```diff
@@ -99,26 +101,69 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
+	case anywherev1.FullClone:
+		// do nothing
+
+	case anywherev1.LinkedClone:
+		if !templateHasSnapshot {
+			return fmt.Errorf(
+				"linkedClone can not be used for VSphereMachineConfig %s has the template %s has no snapshots. Ensure that the template has snapshots or change the cloneMode to fullClone",
+				machineConfig.Name,
+				machineConfig.Spec.Template,
+			)
```
chrisdoherty4, line 142:
```suggestion
			return fmt.Errorf(
				"cannot use 'linkedClone' for VSphereMachineConfig (%s) because its template (%s) has no snapshots; create snapshots or change the cloneMode to 'fullClone'",
				machineConfig.Name,
				machineConfig.Spec.Template,
			)
```

pkg/providers/vsphere/defaults.go
```diff
@@ -99,26 +101,69 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
 		}
-	} else if machineConfig.Spec.DiskGiB != 25 {
-		logger.Info("Warning: Your VM template includes snapshot(s). LinkedClone mode will be used. DiskGiB cannot be customizable as disks cannot be expanded when using LinkedClone mode. Using default of 25 for DiskGiBs.")
-		machineConfig.Spec.DiskGiB = 25
+		if machineConfig.Spec.DiskGiB != diskSize {
+			return fmt.Errorf(
+				"diskGiB can not be customized for VSphereMachineConfig %s when using linkedClone. Either change the cloneMode to fullClone or change the diskGiB to match %s's disk size of %d GiB",
```
chrisdoherty4, line 146:
```suggestion
				"diskGiB cannot be customized for VSphereMachineConfig '%s' when using 'linkedClone'; change the cloneMode to 'fullClone' or the diskGiB to match the template's (%s) disk size of %d GiB",
```

pkg/providers/vsphere/defaults.go
```diff
@@ -99,26 +101,69 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
+	default:
+		return fmt.Errorf("cloneMode %s is not supported for VSphereMachineConfig %s. Supported clone modes: [%s, %s]", machineConfig.Spec.CloneMode, machineConfig.Name, anywherev1.LinkedClone, anywherev1.FullClone)
 	}
 
 	return nil
 }
 
+func (d *Defaulter) setDiskSizeDefaults(machineConfig *anywherev1.VSphereMachineConfig) {
+	if machineConfig.Spec.DiskGiB < 20 {
+		logger.Info("Warning: VSphereMachineConfig DiskGiB cannot be less than 20. Defaulting to 20.")
```
chrisdoherty4, line 162: Is there not a warning method/function we can use? Prepending `Warning` typically leads to inconsistencies as people do `warn:` `WARN:` etc over time.

abhinavmpandey08: I don't think we have this yet. I just moved this call from here https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/defaults.go#L110-L113
I can create an issue to track this separately since it will require updating the other log messages not included as part of this PR

PR #4648, opened 2023-01-17
Close controller dependencies before manager exits

g-gaston:
*Description of changes:*

This guarantees that things like govc sessions are destroyed before the container is stopped, avoiding lingering resources.

### DIFF
--- a/manager/main.go
+++ b/manager/main.go
@@ -123,7 +123,13 @@ func main() {
 	// Setup the context that's going to be used in controllers and for the manager.
 	ctx := ctrl.SetupSignalHandler()
 
-	setupReconcilers(ctx, setupLog, mgr)
+	closer := setupReconcilers(ctx, setupLog, mgr)
+	defer func() {
+		setupLog.Info("Closing reconciler dependencies")
+		if err := closer.Close(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


manager/main.go
```diff
@@ -135,16 +141,27 @@ func main() {
...
-		setupFullLifecycleReconcilers(ctx, setupLog, mgr)
-	} else {
-		setupLog.Info("Setting up legacy cluster controller")
-		setupLegacyClusterReconciler(setupLog, mgr)
+		return setupFullLifecycleReconcilers(ctx, setupLog, mgr)
 	}
+
+	setupLog.Info("Setting up legacy cluster controller")
+	setupLegacyClusterReconciler(setupLog, mgr)
+	return noOpCloser{}
```
taneyland, line 161: just to clarify, we're returning the no-op here because the legacy controller already has logic to handle closing the dependencies?

g-gaston: Because the legacy controller doesn't have any dependency to close (afaik)

cxbrowne1207 (commented):
Looks good to me. There's a question that was asked that I'm curious about as well.

taneyland (approved):

Issue #4655, opened 2023-01-18
Modify EKS-A generate hardware to remove setting metadata.state


pokearu:
## Description
We want to remove setting `.spec.metadata.state` on the hardware CRD with the EKS-A generate hardware from CSV command.

This helps prevent users from accidentally setting a provisioned hardware object back to `provisioning` by regenerating them from an older CSV when adding more hardware to their management cluster.

PR #4656, opened 2023-01-18
Removing setting provisioning field on the hardware metadata

pokearu:
*Issue #4655*

*Description of changes:*
Removing metadata.state from the default generated hardware objects for tinkerbell.

*Testing:*
```
eksctl-anywhere generate hardware -z .ignore/node1214hardware.csv
```


pokearu:
We are now setting the value from CRD API spec.
https://github.com/tinkerbell/tink/pull/669

mitalipaygude (approved):

PR #4671, opened 2023-01-19
Set user-agent in all file readers

g-gaston:
## Description of changes
This should fix the "eks-a-unknown" user-agent in requests for the release and bundles manifests.

There is still cleanup due in the e2e framework where too many readers are instantiated. The cluster.Spec also does too many things, we should remove the read/load functionality it has and make it a pure data struct. However, that would blow up the scope of this change, so leaving that for other PRs.


TerryHowe (approved):
/lgtm

danbudris (approved):

PR #4672, opened 2023-01-19
Revamp (again) API defaults and validations

g-gaston:
In the past we have followed multiple approaches for validations and defaults but none of them seem to have succeeded. This proposed a holistic approach and a plan to make the different implementations consistent.

## TODOS
- [x] Decide if we will provide our own errors package.
- [x] Decide if the validation framework will be generic or tailored to `cluster.Spec`.
- [x] Decide if validations and defaults share the same underlying runner implementation or not.
- [x] Decide on the signature for default handlers
- [x] Decide if we need to abstract concurrency code into a `WorkerPool` construct
- [x] Write `doc.go` for all new packages including `pkg/cli`
- [x] `DeepCopy` `O` in validations and panic if it changes

### DIFF
--- a/designs/validations.md
+++ b/designs/validations.md
@@ -1,0 +1,361 @@
+# EKS-A API validations
+## Problem statement
+
+The current state of the CLI flow is undesirable. Validations and defaults are invoked from multiple places and at different stages with some validations contain hidden dependencies. This creates an inconsistent user experience (error logging, etc.) and, more importantly, a significant maintenance issue. Given a big part of EKS-A's added value is its validations, this topic is of tremendous importance for the project.
+
+There have been previous efforts on this area, however, they only focused on the **type** of validation and **where** to put the validation code, not about **when** or **how** to run them. Although this set the foundations for a better separation of validations between webhooks and controllers, it failed to solve the CLI inconsistency and coupling problems. Moreover, it didn't align well with the team's intent of keeping provider specific validations in their own package.
+
+This document attempts to solve this issue by proposing a wholistic approach for defaulting and validating both CLI and API operations.
+
+### Previous work
+
+https://github.com/aws/eks-anywhere/pull/1343
+
+> ## Validations
+> 
+> There are two types of validations:
+> - Context agnostic validations: there are no external dependencies, having the API object is enough to validate it: mutually exclusive fields, string format, empty fields…
+>    - These should live in the pkg/api/version package
+ >   - These should be the ones called in the validation webhook
+> - Context aware validations: these are dependent on other entities like clients or even other API objects
+>   - These could leave in different places, I’m not making a decision right now: providers, validations package… This is probably worth a separate discussion.
+>   - Try to keep these modular and reusable (eg. Don’t make them part of the providers struct’s api)
+>   - If they only need other API objects from the same package, they can live in the api package
+>   - Controller: these shouldn’t be run in a webhook and must be part of the main reconciliation loop
+
+### Current state
+Validations have grown organically over time. However, the natural evolution of requirements and usecases without taking the time to refactor previous implementations, has resulted in an incohesive array of frameworks, ideas and patterns. This doc doesn't intend to collect or represent them all in detail, but let this list be an example of the lack of cohesion and variety of usecases as well as a data to drive the solution:
+* `pkg/validations`
+	* It holds validation of many different domains (feature gates, certificates, docker daemon, local disk files, command line args, etc.), all implemented as standalone functions.
+	* It also offers some validation "utils/framework", like a validation runner, reporting, error types, etc.
+	* The validations are imported from validation compositions as well as directly from the commands in `cmd`.
+* `pkg/validations/createvalidations`
+	* This implements both *cluster create* validations (as standalone functions) as well as a "validation composition" that allows to build and run "preflight" validations for the *create* operation.
+	* This uses `validations.Validation` and `validations.ValidationResult` but not the `validations.Runner`.
+	* The compiled validations (`createvalidations.CreateValidations`) is used in the `cluster create` command, injected as a dependency to the corresponding workflow and called from the `SetAndValidateTask` task (more on this later).
+* `pkg/validations/upgradevalidations`
+	* This implements both *cluster upgrade* validations (as standalone functions) as well as a "validation composition" that allows to build and run "preflight" validations for the *upgrade* operation.
+	* Strangely, this uses the `validations.ValidationResult` but not the `validations.Runner` and works slightly different than  `CreateValidations`.
+	* The compiled validations (`upgradevalidations.UpgradeValidations`) is used in the `cluster upgrade` command, injected as a dependency to the corresponding workflow and called from the `setupAndValidateTasks` task (more on this later).
+* `pkg/validations/createcluster`
+	* This is a validations composition. It attempts to group all validations for `create cluster` (the ones in `pkg/createcluster`, provider validations, docker executable validations, kubeconfig path, `cluster.Config` validations and Flux validations).
+	* It uses `validations.Runner` to register, run the validations sequentially and report the validations result with `validations.ValidationResult`.
+	* It's a step in the good direction in terms of pattern and semantics vs other previous ideas but it might still be limited in scope.
+* `pkg/api/v1alpha1`
+	* Most eks-a API objects implement the `Validate` method, which is supposed to run only static data validations (or "context unaware"). Unfortunately most of these methods return at the first error as opposed to run them all and aggregate the errors.
+	* Most eks-a API objects implement the `SetDefaults` method.
+	* Most webhooks are also implemented in this package. They are composed of a combination of `object.Validate` calls and field immutability checks (for `ValidateUpgrade`). For defaults, they call `SetDefaults`.
+* `pkg/cluster`
+	* This package exports `cluster.Config` and allows to set defaults and validate it with `cluster.SetDefaults` and `cluster.Validate`. These internally call a `ConfigManager`, which allows to register logic dynamically to parse from yaml, set defaults and validate a `cluster.Config`.
+	* Validations in the package either call `.Validate` in the API objects or implement static validations but at the `Config` level (eg. validations that need multiple API objects).
+	* `ConfigManager`'s defaults also call the `SetDefaults` method in API objects.
+* `pkg/gitops/flux`
+	* Exports validations through `Validations` method, which returns a slice of `validations.Validation`, hence designed to be used in conjunction with `validations.Runner`, etc. These are run from the create cluster workflow.
+* `pkg/providers`
+	* All providers implement `SetupAndValidateCreateCluster` and `SetupAndValidateUpgradeCluster`. They are called from the *create* and *upgrade* workflows.
+	* These methods mix API defaults, validations and internal state changes (equivalent to an `Init` method).
+	* The internal implementation of these method variates from provider to provider. Some of them have a `Validator` and a `Defaulter` and Snow, for example, uses the `cluster.ConfigManager` to organize defaults and validations.
+* `pkg/workflows`
+	* Compose and run validations and defaults with `validations.Runner` from a workflow task.
+* `test/framework`
+	* `ClusterValidator` is yet another validator runner. This is one centered in validating the state of a cluster against a given cluster specification by using a k8s client to inspect the cluster API.
+* Special note is due for commands in `cmd`, since these are the entities with more knowledge and they are a good example of the fragmentation addressed in this doc. Currently, they handle different types of validations and at different stages:
+	* Independently run validations, like:
+		* `FileExists` for the cluster config yaml.
+		* Parsing and Cluster config API validation (`GetAndValidateClusterConfig`) 
+		* Tinkerbell CLI flags.
+		* Registry mirror auth.
+		* Validations to `cluster.Config` with `cluster.Validate`. Defaults are applied by just calling `cluster.NewSpecFromClusterConfig`.
+	* "Grouped" validations in `commonValidations`
+		* Docker daemon config
+		* File exists, parsing and Cluster config API validation (`GetAndValidateClusterConfig`). This is duplicated.
+		* Kubeconfig file and format validation.
+	* Delegated validations to the workflows through `interfaces.Validator`.
+
+## Goals
+* Define a standard for types of validation and defaults.
+* Define where validations and defaults should be defined within the code base.
+* Define a standard for when and where to run validations and defaults for both CLI commands and API requests.
+* Define what validations and defaults should and should not do (scope).
+* Propose a mechanism to run defaults from a single place.
+	* This needs to allow for defaults defined in different packages.
+* Propose a mechanism to run validations from a single place.
+	* This needs to allow for validations defined in different packages.
+	* This should allow to run all validations and collect all errors, as opposed to return after the first error.
+* Propose an execution plan to move all existing validations to such model.
+
+## Proposal
+
+### Kinds
+Validations and defaults can be classified in:
+* Static data: they only require pure API objects and have no other dependencies.
+* *Context aware*: they need to read information from the context (env vars, files, external APIs, etc.). Generally the depend on clients/adapters and a `context.Context`.
+
+Although this doc focuses on API validations, it's worth mentioning *CLI input* validation. This entails command arguments, flags and env vars. In general, these should be fast and have no dependencies. However, it's possible config files (provided through a path) might need to be parsed in order to run conditional validations (for example, depending on the provider specified in the cluster config, extra flags or env vars might be needed).
+
+### Implementation (*where*)
+Static data validations/defaults that only require one single API object **should** live in the `Validate` and `SetDefaults` api struct methods, thus in the `pkg/api/v1**` package. (*Note:* I'm more than open to relax this constraint if there is a good reason to, but given this is one of the only things that seem to resemble some kind consistency with regards to this whole topic, it seems worth to maintain it).
+
+Immutability validations have been treated differently (for some unknown reason) in the CLI than in webhooks. These checks **should** be implemented in the `api` package and **should not** be duplicated in two places. If the same checks are needed in both webhook and CLI, implement as standalone functions and reuse in both places. In the CLI, just read the object with a client and pass both old and new like the webhook does.
+
+The rest of validations and defaults (static data for a group of API structs and the context aware ones) **can** live anywhere, as long the dependency directions make sense, with a few constraints:
+* Context aware validation **cannot** live in `pkg/api/v1**`. This is to keep this package dependencies at a minimum.
+* **Avoid** packages like `validations`. Don't group validations for what they are, group them by what they validate (their context). For example, for docker validations, don't use `validations.ValidateDockerDesktopVersion`, think about `docker.ValidateDesktopVersion` in `pkg/docker`.
+* **Avoid** the `cmd` package. These tend to end up being untested and lack structure.
+
+Webhooks that only require static data validations/defaults **should** live in the `pkg/api/v1**` (where all present webhooks live today). If they have any other dependency the **cannot** live in the `api` package. (*Note*: we could possibly define a `webhooks` folder elsewhere if we think that could avoid problems in the future, leaving this up for discussion)
+
+### Scope (*what*)
+#### Defaults
+* Defaults should not **modify** data provided by the user, they can only **add** information. Any transformation needed should be moved to runtime operations and never get reflected back on the API objects.
+* They should have not secondary effects on anything else except the provided API object/s, including the internal state of the entity running the validation.
+* API objects should only be modified by the defaulting logic. This guarantees they are invariants during the CLI execution, allowing for workflow reentry/re-execution (think command fails and on the next execution we continue from where it stopped).
+
+#### Validations
+* They must not have secondary effects on such as modifying the object under validation.
+* API struct `Validate` methods should return a [`field.ErrorList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#ErrorList). This facilitates using it from webhooks and ensures the API user facing error is as complete as possible.
+
+### Calling (*when*)
+#### Defaults
+ * CLI
+	* At the beginning of the command flow, in most cases before any other business logic.
+	* Always before validations.
+	* Defaults will be grouped into a single usecase per CLI flow: one for create cluster, another one for upgrade, etc. The individual defaulting logic can be reused by importing it and composing it into multiple usecases. These will live in `pkg/cli`, and be named after its flow: `CreateClusterDefaulter`, `UpgradeClusterDefaulter`, etc. 
+	* This doc doesn't prescribe which entity should trigger the defaults (command in `cmd` package, workflow, etc.). But they do need to be run before the API objects get injected in any other entity as a dependency (think *service* objects). Given the approach the provider's refactor project has taken to inject the `cluster.Spec` and any other non dynamic data (this is, known at the beginning of the command execution) as state in the service objects, `cmd`  seems like the obvious place to do this since this is where service objects are constructed.
+* API
+	* Webhook: static data defaults (only depend on the given object) and sometimes, defaults that only require kubernetes API calls (to the same cluster where the object lives). Remember that only the given object should be altered from the webhook.
+	* Controllers: any other defaults requiring extra information, calls to external systems or simply *too slow* for a webhook.
+
+#### Validations
+*  CLI
+	* At the beginning of the command flow, in most cases before any other business logic.
+	* Always after defaults.
+	* Validations will be grouped into a single usecase per CLI flow: one for create cluster, another one for upgrade, etc. The individual validation logic can be reused by importing it and composing it into multiple usecases. These will live in `pkg/cli`, and be named after its flow: `CreateClusterValidator`, `UpgradeClusterValidator`, etc. 
+	* This doc does not prescribe the entity invoking validations (command in `cmd` package, workflow, etc.). The decision is left for the provider refactor team to make later.
+* API
+	* Webhook: static data validations (only depend on the given object) and sometimes, validations that only require kubernetes API calls (to the same cluster where the object lives).
+		* When using a client to read from the API server, be very careful. Webhooks are supposed to be very fast, remember they are synchronous.
+			* Only run *efficient* queries (avoid full scans) and always think about the number of results you expect to get (depending on the type, the difference can be of orders of magnitude).
+			* If your design performance is not very obvious, benchmark it.
+	* Controllers: any other validations requiring extra information, calls to external systems or simply *too slow* for a webhook.
+
+### Composing validations
+This section describes the validation framework and how we will be using it. Requirements:
+* Allow to compose and run validations implemented in different places.
+* Make implementing validations as easy as possible.
+* Design mainly for the CLI use-case. Optimize defaults for this scenario and possibly allow custom configuration for others.
+
+This approach is similar to `cluster.ConfigManager`, and some other approaches, in that it takes a two step process where validations are registered before they are run.
+
+```go
+// Runner allows to compose and run validations.
+type Runner struct
+```
+
+```go
+runner := validation.NewRunner()
+runner.Register(myValidation)
+runner.Register(otherValidations...)
+
+err := runner.RunAll(ctx, spec)
+```
+
+#### Validation signature
+Requirements:
+* They might need to contact external systems. In that case they should make use of a `context.Context`.
+* They might need access to the full specification of an EKS-A cluster. In that case, they should use `cluster.Spec`.
+* Validations should be able to, at least, return an error if the input is not valid.
+
+This boils down to take a context and a cluster spec as input and return some type of error as output. We could implement it with functions or an interface, but given that we want to make registering validations as easy as possible and keep the framework API surface as simple as possible, we opt out for a function:
+
+```go
+type Validation func(ctx context.Context, spec *cluster.Spec) error
+```
+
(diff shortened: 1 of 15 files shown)


chrisdoherty4:
> Write go.doc for all new packages including pkg/cli

Is `go.doc` a typo or a new thing?

g-gaston:
> > Write go.doc for all new packages including pkg/cli
> 
> Is `go.doc` a typo or a new thing?

fixed

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+* They should have not secondary effects on **anything**, including the internal state of the entity running the validation.
+* API struct `Validate` methods should return a [`field.ErrorList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#ErrorList). This facilitates using it from webhooks and ensures the API user facing error is as complete as possible.
+
+### Calling (*when*)
+#### Defaults
+ * CLI
+	* At the beginning of the command flow, in most cases before any other business logic.
+	* Always before validations.
+	* This doc doesn't prescribe which entity should trigger the defaults (command in `cmd` package, workflow, etc.). But they do need to be run before the API objects get injected in any other entity as a dependency (think *service* objects). Given the approach the provider's refactor project has taken to inject the `cluster.Spec` and any other non dynamic data (this is, known at the beginning of the command execution) as state in the service objects, `cmd`  seems like the obvious place to do this since this is where service objects are constructed.
+* API
```
vivek-koppuru, line 121: Is this also before respective validations in webhook and controllers or not the same as cli?

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+	* Webhook: static data defaults (only depend on the given object) and sometimes, defaults that only require kubernetes API calls (to the same cluster where the object lives). Remember that only the given object should be altered from the webhook.
```
vivek-koppuru, line 122: By default, the webhook doesn't have the client object injected to be able to call regular kubernetes API calls. Does this call for creating a custom webhook handler or do we have an example in our code that can in fact make kubernetes API calls?

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Validation error meta information
+The existing `validations.ValidationResult` allows handlers to return a `Remediation` message. Given we want to just use the `error` interface to make it easy to register validations and that `Remediation` is empty for most `ValidationResult`'s , we chose to make this optional. Validations wanting to provide a possible remediation for a failure, can return an `error` that also implements the `Fixable` interface.
```
vivek-koppuru, line 198:
```suggestion
The existing `validations.ValidationResult` allows handlers to return a `Remediation` message. Given we want to just use the `error` interface to make it easy to register validations and that `Remediation` is empty for most `ValidationResult`s , we chose to make this optional. Validations wanting to provide a possible remediation for a failure, can return an `error` that also implements the `Fixable` interface.
```

vivek-koppuru: Also I am confused by this, so where is `Fixable` returned for the validator runner to process remediation and print? Or is that the job of the framework callers?

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Framework implementation
+Given the framework logic is very much independent from the object being validated (`cluster.Spec`), we propose to use generics so the same framework can be used for different types. The complexity added is minimal and even in the case where the framework grows in a direction where generics don't make sense, moving back to a concrete type would be trivial. There are already places in the codebase where we envision make use of this:
+
+```go
+// Validation is the logic for a validation of a type O.
+type Validation[O any] func(ctx context.Context, obj O) error
+```
+
+In order to be able to prevent validations from modifying `obj` and detecting when if that happens, we add a constrain to `O` that allows us to `DeepCopy` the object so we can compare after running all validations:
```
vivek-koppuru, line 219: In what situation do we need to compare the objects after running all validations? It would only be during right, and as you mentioned above the validations should not be modifying the objects.

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+In addition, and following good practices for functions that are supposed to update a given object:
+* Don't rely on updating a given pointer value but take a `cluster.Spec` as input and return a `cluster.Spec` as output, with the necessary updates.
+
+```go
+type Default func(ctx context.Context, spec cluster.Spec) (cluster.Spec, error)
+```
+
+##### Problems with this approach
+
+Defaults should only be set in the API objects configurable by the user. For example, the `Bundles` (which compiles all eks-a dependencies) shouldn't be modified. However, all of these are part of the `cluster.Spec`, giving defaulting logic de ability to modify them. We could enforce this by separating them and providing a `cluster.Config` (all user inputable API objects) and a `Bundles` (or something else that includes the rest of the cluster specification).
```
vivek-koppuru, line 246:
```suggestion
Defaults should only be set in the API objects configurable by the user. For example, the `Bundles` (which compiles all eks-a dependencies) shouldn't be modified. However, all of these are part of the `cluster.Spec`, giving defaulting logic the ability to modify them. We could enforce this by separating them and providing a `cluster.Config` (all user inputable API objects) and a `Bundles` (or something else that includes the rest of the cluster specification).
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+// Default is the logic for a default for a type O. It should return a value of O
+// wether it updates it or not. When there is an error, return the zero value of O
+// and the error.
+type Defaulter[O any] func(ctx context.Context, obj O) (O, error)
+```
+
+#### Concurrency
+All defaults will be sequential given that they need to update the `cluster.Spec` and changes need to be propagated. However, defaults shouldn't rely on order. If such dependency exists between two or more defaults, then make them one single default.
+
+Default handlers can chose to implement concurrency but they are responsible for handling data races.
```
vivek-koppuru, line 261:
```suggestion
Default handlers can choose to implement concurrency but they are responsible for handling data race conditions.
```

designs/validations.md
vivek-koppuru: I think I see it now in the example below, but essentially we check for whether this is a Fixable interface or not with remediation?

vivek-koppuru (commented):
Wow this is a great direction I am excited for!

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+
+Immutability validations have been treated differently (for some unknown reason) in the CLI than in webhooks. These checks **should** be implemented in the `api` package and **should not** be duplicated in two places. If the same checks are needed in both webhook and CLI, implement as standalone functions and reuse in both places. In the CLI, just read the object with a client and pass both old and new like the webhook does.
+
+The rest of validations and defaults (static data for a group of API structs and the context aware ones) **can** live anywhere, as long the dependency directions make sense, with a few constrains:
+* Context aware validation **cannot** live in `pkg/api/v1**`. This is to keep this package dependencies at a minimum.
+* **Avoid** packages like `validations`. Don't group validations for what they are, group them by what they validate (their context). For example, for docker validations, don't use `validations.ValidateDockerDesktopVersion`, think about `docker.ValidateDesktopVersion` in `pkg/docker`.
```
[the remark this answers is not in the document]
chrisdoherty4: I think we expect the `executables` package to be broken up (we'll probably have a separate discussion on how to organize the broader code base but this proposal gives hints on the general direction). We may, eventually, see `type Client interface` and `type Executable struct` both in `/pkg/docker`.

The `executables` package is representative of a layered architecture. E.g. In MVC, models, views, and controllers each have their own respective packages. That's generally not the norm for organizing Go programs.

designs/validations.md
chrisdoherty4: > I think I see it now in the example below, but essentially we check for whether this is a Fixable interface or not with remediation?

That's the gist.

designs/validations.md
chrisdoherty4: > In what situation do we need to compare the objects after running all validations?

Every time the set of validations run. The goal is to make it very difficult if not impossible to abuse validations ensuring the have an appropriately narrow scope.

designs/validations.md
g-gaston: Sorry, I don't follow. The API bullet point only refers to "webhooks + controllers"

designs/validations.md
g-gaston: We don't have an example, we haven't needed this yet (or decided not to do it).
As of how to write this kind of webhook, we just need to implement the `webhook.CustomDefaulter` interface from controller-runtime and register it with `NewWebhookManagedBy(mgr)` against our controller manager.

designs/validations.md
g-gaston: The runner doesn't process errors or print them or do anything with them except flatten them if they are aggregated. It's up to the caller to decide what to do with those errors.

And exactly, we offer a util method that inspects errors to see if they implement `Fixable` or not, and if they do, return the remediation message. That is supposed to be used by the caller, in case it's interested in that kind of info.

pkg/validation/runner.go
```diff
@@ -0,0 +1,143 @@
...
+	if numWorkers > len(r.validations) {
+		numWorkers = len(r.validations)
+	}
+
+	wg.Add(numWorkers)
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			for validate := range validations {
+				if err := validate(ctx, obj); err != nil {
+					for _, err := range flatten(err) {
```
chrisdoherty4, line 97: I think we can leave flattening to `RunAll`.

pkg/validation/runner.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// RunAll runs all validations concurrently and waits until they all finish,
+// aggregating the errors if present. obj must not be modified. If it is, this
+// indicates a programming error and the method will panic.
+func (r *Runner[O]) RunAll(ctx context.Context, obj O) errors.Aggregate {
+	copyObj := obj.DeepCopy()
+	var allErr []error
+	result := r.run(ctx, obj)
+	for err := range result {
```
chrisdoherty4, line 72:
```suggestion
	for err := range r.run(ctx, obj) {
```

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/aws/eks-anywhere/internal/test"
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	eksaerrors "github.com/aws/eks-anywhere/pkg/errors"
+	"github.com/aws/eks-anywhere/pkg/validation"
+)
+
+func TestRunnerRuanAllSuccess(t *testing.T) {
```
chrisdoherty4, line 17:
```suggestion
func TestRunnerRunAllSuccess(t *testing.T) {
```

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+	g := NewWithT(t)
+	ctx := context.Background()
+	r := validation.NewRunner[*cluster.Spec]()
```
chrisdoherty4, line 20: Could we avoid `cluster.Spec` and use a test object instead to decouple them from `cluster.Spec` changes?

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+	)
+
+	spec := test.NewClusterSpec()
+
+	err := r.RunAll(ctx, spec)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Errors()).To(ConsistOf(e1, e2, e3))
+}
+
+func TestRunnerRuanAllPanicAfterModifyingObject(t *testing.T) {
```
chrisdoherty4, line 107:
```suggestion
func TestRunnerRunAllPanicAfterModifyingObject(t *testing.T) {
```

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	})
+	err := r.RunAll(ctx, spec)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err).To(MatchError(ContainSubstring("invalid 1")))
+	g.Expect(err).To(MatchError(ContainSubstring("invalid 2")))
+	g.Expect(callCounter).To(Equal(2))
+}
+
+func TestRunnerRuanAllAggregatedError(t *testing.T) {
```
chrisdoherty4, line 83:
```suggestion
func TestRunnerRunAllAggregatedError(t *testing.T) {
```

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+		s.Cluster.Spec.KubernetesVersion = anywherev1.Kube124
+		s.Cluster.Spec.BundlesRef = &anywherev1.BundlesRef{
+			Name: "bundles-1",
+		}
+	})
+
+	g.Expect(r.RunAll(ctx, spec)).To(Succeed())
+}
+
+func TestRunnerRuanAllSequentially(t *testing.T) {
```
chrisdoherty4, line 48:
```suggestion
func TestRunnerRunAllSequentially(t *testing.T) {
```

pkg/validation/runner.go
```diff
@@ -0,0 +1,143 @@
...
+	return func(c *RunnerConfig) {
+		c.maxJobs = m
+	}
+}
+
+// NewRunner constructs a new Runner.
+func NewRunner[O Validatable[O]](opts ...RunnerOpt) *Runner[O] {
+	r := &Runner[O]{
+		config: &RunnerConfig{
+			maxJobs: 8,
```
chrisdoherty4, line 49: Magic number.

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+		* Registry mirror auth.
+		* Validations to `cluster.Config` with `cluster.Validate`. Defaults are applied by just calling `cluster.NewSpecFromClusterConfig`.
+	* "Grouped" validations in `commonValidations`
+		* Docker daemon config
+		* File exists, parsing and Cluster config API validation (`GetAndValidateClusterConfig`). This is duplicated.
+		* Kubeconfig file and format validation.
+	* Delegated validations to the workflows through `interfaces.Validator`.
+
+## Goals
+* Define a standard for types of validations/defaults and where and **where not** to put them.
```
chrisdoherty4, line 76: We might want to replace the `/` in some of these goals to make clear we're addressing 2 independent concepts, not a pseudonym as `/` can represents.

E.g.

```
* Define a standard for types of validation and defaults.
* Define where validations and defaults should be defined within the code base.
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+Webhooks that only require static data validations/defaults **should** live in the `pkg/api/v1**` (where all present webhooks live today). If they have any other dependency the **cannot** live in the `api` package. (*Note*: we could possibly define a `webhooks` folder elsewhere if we think that could avoid problems in the future, leaving this up for discussion)
+
+### Scope (*what*)
+#### Defaults
+* Defaults should not **modify** data provided by the user, they can only **add** information. Any transformation needed should be moved to runtime operations and never get reflected back on the API objects.
+* They should have not secondary effects on anything else except the provided API object/s, including the internal state of the entity running the validation.
+* API objects should only be modified by the defaulting logic. This guarantees they are invariants during the CLI execution, allowing for workflow reentry/re-execution (think command fails and on the next execution we continue from where it stopped).
+
+#### Validations
+* They should have not secondary effects on **anything**, including the internal state of the entity running the validation.
```
chrisdoherty4, line 112:
```suggestion
* They must not have secondary effects on such as modifying the object under validation.
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+* API struct `Validate` methods should return a [`field.ErrorList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field#ErrorList). This facilitates using it from webhooks and ensures the API user facing error is as complete as possible.
+
+### Calling (*when*)
```
chrisdoherty4, line 115: Do we want to address the rare cases we have where some CLI flag parsing code wants needs to know what provider is in-use before deciding if a flag is required?

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+* API
+	* Webhook: static data defaults (only depend on the given object) and sometimes, defaults that only require kubernetes API calls (to the same cluster where the object lives). Remember that only the given object should be altered from the webhook.
+	* Controllers: any other defaults requiring extra information, calls to external systems or simply *too slow* for a webhook.
+
+#### Validations
+*  CLI
+	* At the beginning of the command flow, in most cases before any other business logic.
+	* Always after defaults.
+	* This doc is not perspective as of which entity should trigger the validations (command in `cmd` package, workflow, etc.). The decision is left for the provider refactor team to make later.
```
chrisdoherty4, line 129:
```suggestion
	* This doc does not prescribe the entity invoking validations (command in `cmd` package, workflow, etc.). The decision is left for the provider refactor team to make later.
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+			* Only run *efficient* queries (avoid full scans) and always think about the number of results you expect to get (depending on the type, the difference can be of orders of magnitude).
+			* If your design performance is not very obvious, benchmark it.
+	* Controllers: any other validations requiring extra information, calls to external systems or simply *too slow* for a webhook.
+
+### Composing validations
+This section lays down the requirements for a framework for validations. Some requisites:
+* Allow to compose and run validations implemented in different places.
+* Make implementing validations as easy as possible.
+* Design mainly for the CLI use-case. Optimize defaults for this scenario and possibly allow custom configuration for others.
+* We want to follow a similar idea to what `cluster.ConfigManager` and other attempts to validation framework propose: a two step process, where validations/defaults can be registered and then run:
```
chrisdoherty4, line 142:
```suggestion

This approach is similar to `cluster.ConfigManager`, and some other approaches, in that it takes a two step process where validations are registered before they are run.
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+#### Composability
+We want the framework to allow, not only to register validations implemented in different places, but also validations that are composed of multiple validations. This, although it introduces some complexity, it offers a lot of flexibility to registrars to aggregate validations in different ways and it facilitates reusing existing validations (that tend to be already aggregated in one big method).
+
+##### Aggregating errors
+For validations to be registered as one handler but still be considered as different in the runner result, they should return an `errors.Aggregate`. The framework will understand this and will flatten the result. This allows for multiple levels of nesting.
+
+##### Concurrency
+Most *preflight* validations we currently run in the CLI can be run concurrently as they are not depending on each other. Given most of the time is spent waiting for external calls (mostly HTTP), this could potentially speed up the process.
```
chrisdoherty4, line 171:
```suggestion
Most *preflight* validations we currently run in the CLI can be run concurrently as they are not dependent on each other. For validations that make external calls this can be slow, concurrent invocation of validations offers a convenient speed improvement.
```

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Validation error meta information
+The existing `validations.ValidationResult` allows handlers to return a `Remediation` message. Given we want to just use the `error` interface to make it easy to register validations and that `Remediation` is empty for most `ValidationResult`s , we chose to make this optional. Validations wanting to provide a possible remediation for a failure, can return an `error` that also implements the `Fixable` interface.
+
+```go
+type Fixable type {
```
chrisdoherty4, line 201: Would you be open to `Remediable`?

chrisdoherty4 (commented):
I didn't quite make it to the end of the proposal but here's a mixture of touch-ups and code suggestions.

pkg/validation/runner.go
```diff
@@ -0,0 +1,143 @@
...
+// Validatable is anything that can be validated.
+type Validatable[O any] interface {
+	Copiable[O]
+}
+
+// Copiable is anything that can be copied.
+type Copiable[O any] interface {
+	DeepCopy() O
+}
+
```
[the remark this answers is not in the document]
chrisdoherty4: Looks like the proposal details it this way also.

chrisdoherty4:
One thing I noticed is that its not necessarily abundantly clear how the 'use-cases' can leverage pieces being provided to them from core logic like the validation framework. It might be worth elaborating on the concrete use-case some.

pkg/validation/log.go
```diff
@@ -0,0 +1,20 @@
+package validation
```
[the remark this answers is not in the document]
g-gaston: Ummm fair, I'll move it 👍🏻

pkg/validation/runner.go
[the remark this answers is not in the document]
g-gaston: Yeah, for the proposal I just wrote the shorter version.

As of motivation, I liked the expressivity of saying that something is Validatable (you can validate it) as long as is Copiable (you can copy it).

If you feel strongly about it, I'll change it, no big deal.

pkg/validation/runner.go
```diff
@@ -0,0 +1,143 @@
...
+// Register adds validations to the Runner.
+func (r *Runner[O]) Register(validations ...Validation[O]) {
+	r.validations = append(r.validations, validations...)
+}
+
+// RunAll runs all validations concurrently and waits until they all finish,
+// aggregating the errors if present. obj must not be modified. If it is, this
+// indicates a programming error and the method will panic.
+func (r *Runner[O]) RunAll(ctx context.Context, obj O) errors.Aggregate {
```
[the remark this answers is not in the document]
g-gaston: Good question. I was thinking originally that we could possibly offer a `Run` that would just return errors through a channel, so the caller can handle them as they come as opposed to block and wait.

But tbh, since I'm not sure we'll ever do that, I'm fine changing it to `Run` if you prefer it.

pkg/validation/runner.go
g-gaston: Fair. I only did it this way because of the possibility of having a `Run` as mentioned in https://github.com/aws/eks-anywhere/pull/4672#discussion_r1110151211
I'll update this accordingly to the resolution of that comment

pkg/validation/runner_test.go
g-gaston: Yes sir

pkg/validation/runner_test.go
```diff
@@ -0,0 +1,126 @@
...
+	g := NewWithT(t)
+	ctx := context.Background()
+	r := validation.NewRunner[*cluster.Spec](validation.WithMaxJobs(10))
+
+	callCounter := 0
+
+	r.Register(
```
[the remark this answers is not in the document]
g-gaston: Added another validation to exemplify how to add a group of sequential validations in a concurrent runner. Although that's why we have an examples file and the examples already show that.

pkg/validation/runner.go
g-gaston: Defaulting to `runtime.GOMAXPROCS`

designs/validations.md
g-gaston: Yes, changing

designs/validations.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Composing validations
+This section lays down the requirements for a framework for validations. Some requisites:
+* Allow to compose and run validations implemented in different places.
+* Make implementing validations as easy as possible.
+* Design mainly for the CLI use-case. Optimize defaults for this scenario and possibly allow custom configuration for others.
+* We want to follow a similar idea to what `cluster.ConfigManager` and other attempts to validation framework propose: a two step process, where validations/defaults can be registered and then run:
+
+```go
+runner := validation.NewRunner()
```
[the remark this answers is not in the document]
g-gaston: Added something, let me know

designs/validations.md
g-gaston: Ok, if we can get an light consensus quickly on in, I'm down. here is my proposal (feel free to tweak wording):

> * In the special case where the code parsing/validating another type of CLI user input ( one that is not part of the EKS-A API, for example, command flags) requires information only available in the API (cluster config file):
>   1. Read the cluster config file
>   1. Parse the yaml content into a `cluster.Config`
>   1. Parse/validate the command flags
>   1. Construct a `cluster.Spec` with this `cluster.Config`
>   1. Continue with API defaults, validations and the rest of the command flow.
>   * Note how this requires complete decoupling between parsing the yaml content and the defaults and validations. Something that historically we haven't followed.

pkg/validation/errors.go
```diff
@@ -0,0 +1,36 @@
+package validation
```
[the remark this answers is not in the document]
chrisdoherty4: > That allows package consumers to implement that however they wont, as opposed to be restricted to use the helpers.

Unexporting the interface doesn't prohibit another package from satisfying it. The interface can be documented and still implemented by custom errors (this is perhaps a not so obvious benefit of duck typing).

```go
package validation

type remediable interface {
	Remediation() string
}

func IsRemediable(err error) bool {
	_, ok := err.(remediable)
	return ok
}
```

```go
package foo

// This works just fine with IsRemediable()
type myRemediableError struct {}
func (myRemediableError) Remediation() string { ... }
func (myRemediableError) Error() string { ... }
```

Regarding exposure of the interface, that gives an opportunity for consumers to return it which should be discouraged in favor of opaque errors (rule of thumb). That also mean any utility funcs like `NewRemediableErr` should return `error`.

The utility functions are just that. You can ignore them and define your own remediable error, or use them for convenience in error handling.

Dave Cheney does a great job explaining this kind of error handling: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully.

> It allows to check if the error is remediable and get the message in one call. As opposed to two, which implies checking for the interface implementation twice.

My motivation is readability at the consumer site and my feeling was the second form below is more readable than the first - I'm OK either way, though I'm not sure there's a problem with a double interface check.

```go
if rerr, ok := validation.ErrorRemediation(err); ok {
	handleRemediation(rerr)
}
```
vs
```go
if validation.IsRemediable(err) {
  handleRemediation(validation.Remediation(err))
}
```

pkg/validation/runner.go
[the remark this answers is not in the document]
chrisdoherty4: I would go with `Validatable` only because its simpler and nothing in the package consumes `Copiable`.

pkg/validation/runner.go
chrisdoherty4: I was thinking more about having a constant for it instead of defining in-place but on second look I don't think it's going to make a difference.

designs/validations.md
chrisdoherty4: It was just a thought I had. It can totally be addressed independently.

pkg/validation/errors.go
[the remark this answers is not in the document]
g-gaston: > Unexporting the interface doesn't prohibit another package from satisfying it. The interface can be documented and still implemented by custom errors (this is perhaps a not so obvious benefit of duck typing).

Sure you can implement the unexported interface, but you need to know about it. In that case, why wouldn't you use the code as documentation? Plus documenting it but not exposing in the package's API creates a hidden dependency. If we were to change our `Remediable` interface, the code implementing it could break silently.

> Regarding exposure of the interface, that gives an opportunity for consumers to return it which should be discouraged in favor of opaque errors (rule of thumb).

You mention trying to avoid other code from returning it in favor of opaque errors. Isn't that going a bit too far? Even if we didn't export it, other code can define its own copy of the interface and return it. I'm in favor of using the type system to ensure correctness when possible, but trying to enforce a particular error pattern seems too much.

> That also mean any utility funcs like NewRemediableErr should return error.

I don't think I agree here. I buy the argument for the code that actually generates the error since it helps decouple that package from the caller. But this is just a constructor for the error. Callers might actually want *this* error.

> I'm OK either way, though I'm not sure there's a problem with a double interface check.

Performance was my concern. But let's go for legibility and we can come back to it if it becomes a problem.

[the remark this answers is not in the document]
g-gaston: Ummm I just started playing around with the code.

Now I realize that in order to return remediable errors as an interface from the constructors, I need to embed `error` into `Remediable` or else the returned value doesn't directly implement `error`.

I don't like that, it pollutes the `Fixable` interface just to be able to use the returned value as an `error` elsewhere. I'm gonna change it to return `error` as you suggest, sorry for the back and forth.

g-gaston:
> One thing I noticed is that its not abundantly clear how the 'use-cases' can leverage pieces being provided to them from core logic like the validation framework. It might be worth elaborating on the concrete use-case some and how they extract information from data structures they receive.

Do you think adding an example in the proposal doc would help with this?

pkg/validation/errors.go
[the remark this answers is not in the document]
chrisdoherty4: > Sure you can implement the unexported interface, but you need to know about it. In that case, why wouldn't you use the code as documentation? Plus documenting it but not exposing in the package's API creates a hidden dependency. If we were to change our Remediable interface, the code implementing it could break silently.

FWIW, its a reasonably widely accepted pattern and the standard library also buys into it: [`errors.Is()`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/errors/wrap.go;l=53), [`errors.As()`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/errors/wrap.go;l=113) and [`errors.Unwrap()`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/errors/wrap.go;l=16) all document their expectations without any explicit interface.

> If we were to change our Remediable interface, the code implementing it could break silently.

Yes, absolutely. The interface is small and simple though, so the risk seems low.

chrisdoherty4:
> Do you think adding an example in the proposal doc would help with this?

That could help along with a brief explanation.

_Sorry for the slow response, somehow missed the notification.

chrisdoherty4 (approved):
I think this is looking really good.

chrisdoherty4 (approved):

PR #4677, opened 2023-01-19
Bump oras modules and remove containerd replace

g-gaston:
*Description of changes:*
Fixes CVE in github.com/containerd/containerd
We had an old replace directive, so I removed it. It wasn't necessary, but I also bumped `oras`, which are the modules that need containerd.


TerryHowe (approved):
/lgtm

PR #4684, opened 2023-01-19
Add passing the insecure flag to packages

TerryHowe:
The cluster spec has `insecureSkipVerify`, but we were not passing that to packages

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -168,7 +168,7 @@ func (pc *PackageControllerClient) CreateHelmOverrideValuesYaml() (string, []byt
 
 func (pc *PackageControllerClient) generateHelmOverrideValues() ([]byte, error) {
 	var err error
-	endpoint, username, password, caCertContent := "", "", "", ""
+	endpoint, username, password, caCertContent, insecureSkipVerify := "", "", "", "", "false"
 	if pc.registryMirror != nil {
 		endpoint = pc.registryMirror.BaseRegistry
 		username, password, err = config.ReadCredentials()
@@ -176,6 +176,9 @@ func (pc *PackageControllerClient) generateHelmOverrideValues() ([]byte, error) 
 			return []byte{}, err
 		}
 		caCertContent = pc.registryMirror.CACertContent
+		if pc.registryMirror.InsecureSkipVerify {
+			insecureSkipVerify = "true"
+		}
 	}
 	templateValues := map[string]interface{}{
 		"eksaAccessKeyId":     base64.StdEncoding.EncodeToString([]byte(pc.eksaAccessKeyID)),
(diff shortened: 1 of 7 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -158,14 +158,17 @@ func (pc *PackageControllerClient) CreateHelmOverrideValuesYaml() (string, []byt
...
+	endpoint, username, password, caCertContent, insecureSkipVerify := "", "", "", "", "false"
 	if pc.registryMirror != nil {
 		endpoint = pc.registryMirror.BaseRegistry
 		username, password, err = config.ReadCredentials()
 		if err != nil {
 			return []byte{}, err
 		}
 		caCertContent = pc.registryMirror.CACertContent
+		if pc.registryMirror.InsecureSkipVerify {
+			insecureSkipVerify = "true"
```
a-cool-train, line 170: You can just pass pc.registryMirror.InsecureSkipVerify down to the helm installation.

pkg/curatedpackages/testdata/values_empty.yaml
```diff
@@ -3,6 +3,7 @@ registryMirrorSecret:
   username: ""
   password: ""
   cacertcontent: ""
+  insecure: "ZmFsc2U="
```
a-cool-train, line 6: This should only true or false, right? Should we just check for if it is set to true and otherwise default to false.

pkg/curatedpackages/packagecontrollerclient.go
TerryHowe: InsecureSkipVerify is a bool, so at the very least I need to covert it to a string. There could be a more clever way to do that than this ham fisted way, but not sure it would be a huge improvement

pkg/curatedpackages/testdata/values_empty.yaml
TerryHowe: The handling on our side should only accept true, but I think it is a bit more clear to set it to false rather than not set it at all.

pkg/curatedpackages/packagecontrollerclient.go
TerryHowe: I looked at that, but this block checks for `if pc.registryMirror != nil {`, so I think this way is better.

pkg/curatedpackages/testdata/values_empty.yaml
a-cool-train: But currently it looks like as long as this field is set, we set it to true which doesn't seem like it would make sense. I think it is better to check for the true or t explicit fields or otherwise assume it is false so let it be false.

TerryHowe: I'm not following.
```
 % echo 'ZmFsc2U=' | base64 -d
false%                                            
```

a-cool-train: oh I see so it is base64 encoded.

PR #4690, opened 2023-01-19
adding external etcd unsupported for Tinkerbell check

raymond-zhang00:
*Issue #, if available:*
1033

*Description of changes:*
Tinkerbell external etcd is currently unsupported. Added a validation to prevent this configuration.

*Testing (if applicable):*
Added tests to cluster validations

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -187,7 +187,7 @@ func NewCluster(clusterName string) *Cluster {
 var clusterConfigValidations = []func(*Cluster) error{
 	validateClusterConfigName,
 	validateControlPlaneEndpoint,
-	validateUnsupportedFeatures,
+	validateExternalEtcdSupport,
 	validateMachineGroupRefs,
 	validateControlPlaneReplicas,
 	validateWorkerNodeGroups,
@@ -364,7 +364,7 @@ func validateClusterConfigName(clusterConfig *Cluster) error {
 	return nil
 }
 
-func validateUnsupportedFeatures(cluster *Cluster) error {
+func validateExternalEtcdSupport(cluster *Cluster) error {
 	if cluster.Spec.DatacenterRef.Kind == TinkerbellDatacenterKind {
 		if cluster.Spec.ExternalEtcdConfiguration != nil {
 			return errors.New("tinkerbell external etcd configuration is unsupported")
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -187,6 +187,7 @@ func NewCluster(clusterName string) *Cluster {
 var clusterConfigValidations = []func(*Cluster) error{
 	validateClusterConfigName,
 	validateControlPlaneEndpoint,
+	validateUnsupportedFeatures,
 	validateMachineGroupRefs,
 	validateControlPlaneReplicas,
 	validateWorkerNodeGroups,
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -363,6 +364,16 @@ func validateClusterConfigName(clusterConfig *Cluster) error {
 	return nil
 }
 
+func validateUnsupportedFeatures(cluster *Cluster) error {
```
cxbrowne1207, line 367: Why don't we name this something more specific? Like `validateExternalEtcdSupport`

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -93,6 +93,49 @@ func TestClusterNameLength(t *testing.T) {
...
 }
 
+func TestUnsupportedFeatures(t *testing.T) {
+	tests := []struct {
+		name        string
+		wantCluster *Cluster
+		wantErr     error
+	}{
+		{
+			name: "SuccessTinkerbellDatacenter",
```
cxbrowne1207, line 103:
```suggestion
			name: "tinkerbell config without external etcd",
```

This is how similar tests are described.

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -93,6 +93,49 @@ func TestClusterNameLength(t *testing.T) {
...
+					DatacenterRef: Ref{
+						Kind: TinkerbellDatacenterKind,
+						Name: "eksa-unit-test",
+					},
+				},
+			},
+			wantErr: nil,
+		},
+		{
+			name: "FailureTinkerbellExternalEtcd",
```
cxbrowne1207, line 115:
```suggestion
			name: "tinkerbell config with external etcd",
```

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -93,6 +93,49 @@ func TestClusterNameLength(t *testing.T) {
 	}
 }
 
+func TestUnsupportedFeatures(t *testing.T) {
```
cxbrowne1207, line 96:
```suggestion
func TestValidateExternalEtcdSupport(t *testing.T) {
```

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -93,6 +93,49 @@ func TestClusterNameLength(t *testing.T) {
...
+			wantErr: errors.New("tinkerbell external etcd configuration is unsupported"),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(tt *testing.T) {
+			got := validateUnsupportedFeatures(tc.wantCluster)
+			if !reflect.DeepEqual(tc.wantErr, got) {
+				t.Errorf("%v got = %v, want %v", tc.name, got, tc.wantErr)
+			}
```
cxbrowne1207, line 134: Can we use `gomega` here instead? Something like 

`g.Expect(got).To(MatchError(tt.wantErr))
`

pkg/api/v1alpha1/cluster.go
raymond-zhang00: I think that would be good - it will be more descriptive now, and can be modified later if we need to fit other things under it.

pkg/api/v1alpha1/cluster_test.go
```diff
@@ -93,6 +93,52 @@ func TestClusterNameLength(t *testing.T) {
...
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(tt *testing.T) {
+			got := validateExternalEtcdSupport(tc.wantCluster)
+			if tc.wantErr {
+				g.Expect(got).To(MatchError(ContainSubstring("unsupported")))
```
cxbrowne1207, line 134:
```suggestion
				g.Expect(got).To(MatchError(ContainSubstring("external etcd configuration is unsupported")))
```

What about using more of the string like this? I think this would be more robust since this test is looking for external etcd configuration support specifically

cxbrowne1207 (approved):

PR #4696, opened 2023-01-20
Add mention of the copy packages command

TerryHowe:
This could be a better, but I want to get something out there


ewollesen (commented):
/lgtm

PR #4701, opened 2023-01-20
Bump helm dependency in release module

g-gaston:
*Description of changes:*
Same CVE as the other day, but in the `release` module, due to `helm.sh/helm/v3`.
Since there are no new `v3.10`, I bumped to `v3.11.0`. Based on the changelog, there are no breaking changes.

Bumping helm created quite a mess with other k8s.io dependencies.

* `helm.sh/helm/v3@v3.11.0` only requires go 1.17, but it imports `k8s.io` modules at `v0.26.0`, which require go 1.19 (it uses atomic bool).
(like `k8s.io/component-base
* `sigs.k8s.io/controller-runtime` is not compatible with `k8s.io/client-go@v0.26.0` (due to a change in an interface's signature). We would need to bump controller-runtime to `v0.14.0` but that also requires go 1.19.

I bypassed this with replaces. Although not ideal, it gets the CVE fix merged asap and decouples this from a go version upgrade.

About go 1.19: I have been using it locally for quite some time and I haven't had any problems. Given it seems like most dependencies have already moved to it and some even require it, I will follow up with a go version update for the whole repo.


TerryHowe (approved):
/lgtm

Helm upgrades to make me nervous

PR #4716, opened 2023-01-20
Move cluster validator out of test framework

cxbrowne1207:
*Issue #, if available:*
[#1015
](https://github.com/orgs/aws/projects/19/views/10?pane=issue&itemId=17464335)

*Description of changes:*
This PR is the first part addressing the issue linked above. It moves the cluster validator out of the test framework in addition to one of the validations.

*Testing (if applicable):*


pkg/clustervalidator/clustervalidator.go
```diff
@@ -0,0 +1,84 @@
+package clustervalidator
```
chrisdoherty4, line 1: We should refrain from creating packages for types. Is there a reason this can't live in the  `Cluster` package? I would think things consuming the `cluster.Spec` would want to consume `cluster.Validator`.

pkg/clustervalidations/cluster.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/eks-anywhere/pkg/clustervalidator"
+	"github.com/aws/eks-anywhere/pkg/controller"
+)
+
+// ValidateClusterReady get the CAPICluster from the client then validates that it is in a ready state.
+func ValidateClusterReady(ctx context.Context, vc clustervalidator.Config) error {
```
chrisdoherty4, line 12: Is this going to be conditionally added or is it something we need to always check?

pkg/clustervalidator/clustervalidator.go
```diff
@@ -0,0 +1,84 @@
...
+			if err != nil {
+				errList = append(errList, err)
+			}
+		}()
+	}
+	wg.Wait()
+	return errors.NewAggregate(errList)
+}
+
+func (c *ClusterValidator) runValidation(ctx context.Context, v RetriableClusterValidation) error {
```
chrisdoherty4, line 53: This is a little peculiar. I would think the `ClusterValidator` has a set of `ClusterValidation` funcs and those funcs are responsible for retrying if they need to.

We could add a method to `RetriableClusterValidation` that matches the `ClusterValidation` signature and wraps up the retry algorithm. That makes the `RetriableClusterValidation` a helper of sorts that can still be registered with `ClusterValidator`. Then make `ClusterValidator` only concerned with `ClusterValidation`s and drop the `runValidation` func creating a much clearer separation of concerns.

E.g.
```go
// Add a method complimenting the ClusterValidation 
func (RetriableClusterValidation) Validate(ctx context.Context, vc Config) error { ... }

// Change to only care about ClusterValidation funcs.
type ClusterValidator struct {
	Config      Config
	validations []ClusterValidation
}

// Register the retriable validation Validate method with the validator instead.
func doValidation() {  
  retriable := RetriableValidation{
    Retrier: NewRetrier(),
    Validation: func(ctx context.Context, vc Config) error {
      // do validation
    }
  }

  validator := NewClusterValidator()
  validator.WithValidations(retriable.Validate)
  
}
```

pkg/clustervalidations/cluster.go
cxbrowne1207: That depends on how the cluster validator is used. You could opt to not register this validation in a specific use case.

pkg/clustervalidator/clustervalidator.go
```diff
@@ -0,0 +1,84 @@
...
+	var wg sync.WaitGroup
+	wg.Add(len(c.validations))
+	errList := make([]error, 0)
+	for _, validation := range c.validations {
+		v := validation
+		go func() {
+			defer wg.Done()
+			err := c.runValidation(ctx, v)
+			if err != nil {
+				errList = append(errList, err)
```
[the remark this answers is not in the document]
cxbrowne1207: Yeah, thanks for catching that. This was originally a synchronous operation, and ended up missing this.

pkg/clustervalidator/clustervalidator.go
cxbrowne1207: Okay, interesting. This would definitely be preferred as well for registering validations that possible don't require retries

cxbrowne1207:
/hold

I think it probably makes sense to move and write unit tests for the existing validations in this one PR

test/framework/clusterstatevalidations/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	yaml := vc.ClusterSpec.Cluster
+	cm := &corev1.ConfigMap{}
+	key := types.NamespacedName{Namespace: "kube-system", Name: "cilium-config"}
+	err := clusterClient.Get(ctx, key, cm)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve configmap %s", err)
+	}
+
+	clusterCilium := cm.Data["enable-policy"]
+	yamlCilium := string(yaml.Spec.ClusterNetwork.CNIConfig.Cilium.PolicyEnforcementMode)
```
g-gaston, line 143: is there any way we can leverage the code in the cilium package to do this?
I believe we have code to do all this
Plus we could benefit from checking the version as well, useful to for upgrades

test/framework/clusterstatevalidations/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	if !strings.Contains(kubeletVersion, string(kubeVersion)) {
+		return fmt.Errorf("validating node version: kubernetes version %s does not match expected version %s", kubeletVersion, kubeVersion)
+	}
+	return nil
+}
+
+// getWorkerNodeMachineSets gets a list of MachineSets corresponding the provided WorkerNodeGroupConfiguration from the management cluster.
+func getWorkerNodeMachineSets(ctx context.Context, vc clusterf.StateConfig, w v1alpha1.WorkerNodeGroupConfiguration) ([]v1beta1.MachineSet, error) {
+	md := &v1beta1.MachineDeployment{}
+	mdName := fmt.Sprintf("%s-%s", vc.ClusterSpec.Cluster.Name, w.Name)
```
g-gaston, line 170: do we have a method in the clusterapi package to generate this name?

test/framework/clusterstatevalidations/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	key := types.NamespacedName{Name: mdName, Namespace: constants.EksaSystemNamespace}
+	if err := vc.ManagementClusterClient.Get(ctx, key, md); err != nil {
```
g-gaston, line 172: why do we need to read the machine deployment if the only thing we need is its name? We are using the name to read it, so we are not gaining anything from that extra call

test/framework/clusterstatevalidations/main_test.go
```diff
@@ -0,0 +1,94 @@
...
+	objs = append(objs, tt.config.ClusterSpec.Cluster)
+	objs = append(objs, tt.otherObjs...)
+	for _, o := range childObjects {
+		objs = append(objs, o)
+	}
+	return objs
+}
+
+func TestMain(m *testing.M) {
+	os.Exit(envtest.RunWithEnvironment(m, envtest.WithAssignment(&env)))
```
g-gaston, line 93: do we need envtest for these tests? Can't we just use fakeclient?

test/framework/clusterstatevalidations/main_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+var env *envtest.Environment
+
+const (
+	clusterName      = "test-cluster"
+	clusterNamespace = "test-namespace"
+)
+
+type clusterValidationTest struct {
```
g-gaston, line 26: I would put this somewhere else and leave `main_test` only for the `TestMain` function

test/framework/clusterstatevalidations/cluster.go
```diff
@@ -0,0 +1,186 @@
+package clusterstatevalidations
```
g-gaston, line 1: what about putting this in `test/framework/cluster/validations`?

test/framework/clusterstatevalidations/vsphere.go
```diff
@@ -0,0 +1,46 @@
...
+
+	clusterf "github.com/aws/eks-anywhere/test/framework/cluster"
+)
+
+// ValidateCSI checks that the CSI is enabled/disabled based on the settings in the cluster.Spec by
+// by checking if CSI related objects components exist in the cluster or not.
+func ValidateCSI(ctx context.Context, vc clusterf.StateConfig) error {
+	clusterClient := vc.ClusterClient
+
+	yaml := vc.ClusterSpec.Config.VSphereDatacenter
```
g-gaston, line 18: why is this called yaml?

test/framework/clusterstatevalidations/vsphere.go
```diff
@@ -0,0 +1,46 @@
...
+	dsRes := handleCSIError(dsErr, yamlCSI)
+	if dsRes != nil {
+		return dsRes
+	}
+
+	return nil
+}
+
+func handleCSIError(err error, disabled bool) error {
+	if (disabled && err == nil) || (!disabled && err != nil) {
```
g-gaston, line 41: there could be an error for other reasons that not found
I feel like you could also handle the `NamespacedName` and client.Get here and simplify the code in the caller.

test/framework/clustervalidator.go
```diff
@@ -3,277 +3,104 @@ package framework
...
-func (c *ClusterValidator) Validate(ctx context.Context) error {
-	for _, v := range c.validations {
-		err := retrier.Retry(v.maxRetries, v.backoffPeriod, func() error {
-			return v.validation(ctx, c.Config)
-		})
-		if err != nil {
-			return fmt.Errorf("validation failed %v", err)
-		}
+// WithValidationsForExpectedObjects registers a set validations for the existence of various cluster objects.
+func WithValidationsForExpectedObjects(cv *clusterf.StateValidator) *clusterf.StateValidator {
```
g-gaston, line 21: nit: I would return a validation slice instead of doing the registering here
That way you have all the register calls in one place and avoid methods with side effects

test/framework/clustervalidator.go
```diff
@@ -3,277 +3,104 @@ package framework
...
+
+// WithValidationsForClusterDoesNotExist registers validations to check that a cluster does not exist or has been deleted.
+func WithValidationsForClusterDoesNotExist(cv *clusterf.StateValidator) *clusterf.StateValidator {
+	cv.WithValidations(
+		clusterf.RetriableStateValidation(retrier.NewWithMaxRetries(60, 5*time.Second), csvalidations.ValidateClusterDoesNotExist),
+	)
+	return cv
+}
+
+func (e *ClusterE2ETest) initClusterStateConfig(ctx context.Context) {
```
g-gaston, line 42: I would advise against this `init` method
I would just have a `clusterStateConfig(ctx)` that checks if cluster state config is memoized and build it if it's not

test/framework/flux.go
```diff
@@ -376,6 +376,9 @@ func (e *ClusterE2ETest) PushWorkloadClusterToGit(w *WorkloadCluster, opts ...ap
 
 	// Update the cluster config with the provided api.ClusterConfigFillers
 	w.UpdateClusterConfig(opts...)
+	if w.clusterStateConfig != nil {
+		w.updateClusterStateConfig()
```
g-gaston, line 380: why is this needed?

test/framework/workload.go
```diff
@@ -63,6 +67,42 @@ func (w *WorkloadCluster) WaitForKubeconfig() {
...
 		w.T.Fatalf("Failed waiting for cluster kubeconfig: %s", err)
 	}
+	w.initClusterStateConfig(ctx)
+}
+
+// ValidateClusterDelete verifies the cluster has been deleted.
+func (w *WorkloadCluster) ValidateClusterDelete() {
+	ctx := context.Background()
+	w.T.Logf("Validating cluster deletion %s", w.ClusterName)
+	if w.clusterStateConfig == nil {
```
g-gaston, line 77: What is check for?

test/framework/workload.go
```diff
@@ -63,6 +67,42 @@ func (w *WorkloadCluster) WaitForKubeconfig() {
...
+		return
+	}
+	clusterStateValidator := newClusterStateValidator(w.clusterStateConfig)
+	WithValidationsForClusterDoesNotExist(clusterStateValidator)
+	if err := clusterStateValidator.Validate(ctx); err != nil {
+		w.T.Fatalf("failed to validate cluster deletion %v", err)
+	}
+	w.clusterStateConfig = nil
```
g-gaston, line 85: why does this method have side effects on the test object?

test/framework/workload.go
```diff
@@ -63,6 +67,42 @@ func (w *WorkloadCluster) WaitForKubeconfig() {
...
+}
+
+func (w *WorkloadCluster) initClusterStateConfig(ctx context.Context) {
```
g-gaston, line 88: same here, I would avoid this and just have a method that returns the clusterStateConfig

pkg/clients/kubernetes/scheme.go
```diff
@@ -32,6 +33,7 @@ var schemeAdders = []schemeAdder{
 	vspherev1.AddToScheme,
 	etcdv1.AddToScheme,
 	addonsv1.AddToScheme,
+	appsv1.AddToScheme,
```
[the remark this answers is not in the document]
cxbrowne1207: Ah, I think this is something I forgot to remove. I had gotten an error I thought was related to a scheme needing to be registered while writing unit tests, but I was wrong.

test/framework/clusterstatevalidations/cluster.go
cxbrowne1207: i also meant it as a check to make sure that the corresponding machine deployment exists. do you think that's unnecessary for this validation?

test/framework/clusterstatevalidations/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	clus := vc.ClusterSpec.Cluster
+	mgmtClusterClient := vc.ManagementClusterClient
+	capiCluster, err := controller.GetCAPICluster(ctx, mgmtClusterClient, clus)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster %s", err)
+	}
+	if capiCluster == nil {
+		return fmt.Errorf("cluster %s does not exist", clus.Name)
+	}
+	for _, condition := range capiCluster.GetConditions() {
```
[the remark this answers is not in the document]
cxbrowne1207: Ah yes, I was unaware.

test/framework/cluster/state_validator.go
```diff
@@ -0,0 +1,81 @@
...
+	var mu sync.Mutex
+	wg.Add(len(c.validations))
+	errList := make([]error, 0)
+	for _, validation := range c.validations {
+		validate := validation
+		go func() {
+			defer wg.Done()
+			err := validate(ctx, c.Config)
+			if err != nil {
+				mu.Lock()
```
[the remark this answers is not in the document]
cxbrowne1207: I was not aware of channels in Go yet honestly. If they're preferred, I could use those instead

test/framework/clustervalidator.go
cxbrowne1207: Curious, what is your reasoning behind being strongly against this `init` method? I could simply build when `ValidateClusterState` is called, but I was going for keeping the spec in the `clusterValidatorConfig` up to date with the in the running cluster when the API is used.

test/framework/flux.go
cxbrowne1207: I was going for keeping the spec in the `clusterValidatorConfig` up to date with the in the running cluster when the API is used.

test/framework/workload.go
cxbrowne1207: When `clusterStateConfig` is required for this to work. Are you thinking it's better to leave out this check because it should error in that case?

test/framework/clusterstatevalidations/cluster.go
cxbrowne1207: Possible we can mark this down for improvement in another PR? This one is just moving the validations.

test/framework/clusterstatevalidations/vsphere.go
cxbrowne1207: I'll update this.

test/framework/clusterstatevalidations/vsphere.go
cxbrowne1207: Should we handle this as a fix in another PR as well? This PR is just moving the validations.

test/framework/clusterstatevalidations/main_test.go
cxbrowne1207: I'll see if we can do that instead!

test/framework/clusterstatevalidations/cluster.go
g-gaston: IMO that is an implementation detail. This validations are checking that the state of the cluster complies with the cluster specification. Machine deployments are not part of the cluster state, they are a mean to manage/control that state.

test/framework/cluster.go
```diff
@@ -1673,22 +1656,13 @@ func (e *ClusterE2ETest) CombinedAutoScalerMetricServerTest(autoscalerName strin
...
-	providerValidations := e.Provider.ClusterValidations()
-	e.clusterValidator.WithValidations(providerValidations...)
-
-	if err := e.clusterValidator.Validate(ctx); err != nil {
+	e.buildClusterStateValidationConfig(ctx)
+	clusterStateValidator := newClusterStateValidator(e.clusterStateValidationConfig)
+	validations := validationsForExpectedObjects()
+	providerValidations := e.Provider.ClusterStateValidations()
+	validations = append(validations, providerValidations...)
+	clusterStateValidator.WithValidations(validations...)
```
g-gaston, line 1664: nit, but I think it's a bit cleaner
```suggestion
	clusterStateValidator := newClusterStateValidator(e.clusterStateValidationConfig)
	clusterStateValidator.WithValidations(validationsForExpectedObjects()...)
	clusterStateValidator.WithValidations(e.Provider.ClusterStateValidations()...)
```

test/framework/cluster/state_validator.go
```diff
@@ -0,0 +1,71 @@
...
+// WithValidations registers multiple validations to the StateValidator that will be run when Validate is called.
+func (c *StateValidator) WithValidations(validations ...StateValidation) {
+	c.validations = append(c.validations, validations...)
+}
+
+// Validate runs through the set registered validations and returns an error if any of them fail after a number of retries.
+func (c *StateValidator) Validate(ctx context.Context) error {
+	errList := make([]error, 0)
+	for _, validation := range c.validations {
+		validate := validation
```
g-gaston, line 41:
```suggestion
	for _, validate := range c.validations {
```

g-gaston (approved):

test/framework/cluster/state_validator.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+)
+
+// StateValidation defines a validation that can be registered to the StateValidator.
+type StateValidation = func(ctx context.Context, vc StateValidationConfig) error
```
chrisdoherty4, line 14: Any reason for the alias as opposed to definition?

test/framework/cluster/state_validator.go
```diff
@@ -0,0 +1,70 @@
...
+		err := validate(ctx, c.Config)
+		if err != nil {
+			errList = append(errList, err)
+		}
+	}
+	return errors.NewAggregate(errList)
+}
+
+// Opt represents is a function that represents an option to configure a StateValidator.
+type Opt = func(cv *StateValidator)
```
chrisdoherty4, line 50: Any reason for the alias as opposed to definition? 

Could we also expand to `Option` leaving abbreviations primarily for variable names.

chrisdoherty4: Do we have `Option` implementations somewhere? If not we don't really need options. I also noted we accept the `StateValidationConfig` so users can just configure that as needed. Generally, options are only needed when configuring a set of defaults you want to expose configurability for.

test/framework/clustervalidator.go
```diff
@@ -3,277 +3,94 @@ package framework
...
-				}
-			}
-		}
-		if workerGroupCount != *w.Count {
-			return fmt.Errorf("worker node group %s count does not match expected: %d of %d", w.Name, workerGroupCount, *w.Count)
-		}
-	}
-	logger.V(4).Info("Worker nodes validated", "cluster", clusterName)
-	return nil
+func newClusterStateValidator(config *clusterf.StateValidationConfig) *clusterf.StateValidator {
```
chrisdoherty4, line 62: Are callers able to call `clusterf.NewStateValidator` directly?

test/framework/workload.go
```diff
@@ -65,6 +65,19 @@ func (w *WorkloadCluster) WaitForKubeconfig() {
...
+// ValidateClusterDelete verifies the cluster has been deleted.
+func (w *WorkloadCluster) ValidateClusterDelete() {
+	ctx := context.Background()
+	w.T.Logf("Validating cluster deletion %s", w.ClusterName)
+	clusterStateValidator := newClusterStateValidator(w.clusterStateValidationConfig)
+	clusterStateValidator.WithValidations(
+		validationsForClusterDoesNotExist()...,
+	)
+	if err := clusterStateValidator.Validate(ctx); err != nil {
+		w.T.Fatalf("failed to validate cluster deletion %v", err)
```
chrisdoherty4, line 77:
```suggestion
		w.T.Fatalf("failed to validate cluster deletion: %v", err)
```

test/framework/cluster/validations/cluster.go
```diff
@@ -0,0 +1,199 @@
...
+	clusterf "github.com/aws/eks-anywhere/test/framework/cluster"
+)
+
+// ValidateClusterReady gets the CAPICluster from the client then validates that it is in a ready state.
+func ValidateClusterReady(ctx context.Context, vc clusterf.StateValidationConfig) error {
+	clus := vc.ClusterSpec.Cluster
+	mgmtClusterClient := vc.ManagementClusterClient
+	capiCluster, err := controller.GetCAPICluster(ctx, mgmtClusterClient, clus)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster %s", err)
```
chrisdoherty4, line 30:
```suggestion
		return fmt.Errorf("failed to retrieve cluster: %s", err)
```

PR #4725, opened 2023-01-23
Support both dev and prod

vincentni:


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -145,7 +145,7 @@ func (pc *PackageControllerClient) GetCuratedPackagesRegistries() (sourceRegistr
 			defaultImageRegistry = gatedOCINamespace
 		}
 	} else {
-		defaultImageRegistry = strings.ReplaceAll(constants.DefaultCuratedPackagesRegistryRegex, "*", pc.eksaRegion)
+		defaultImageRegistry = strings.ReplaceAll(defaultImageRegistry, "us-west-2", pc.eksaRegion)
```
junshun, line 148: Is there region replication yet on the dev images?

TerryHowe: @jonahjon would know, but either way we should support it and have replication to at least one region.

junshun:
/lgtm
/approve

We can add support for dev registry mirroring later

Issue #4709, opened 2023-01-20
Machine gets deleted when exceeding default node start timeout in machine health check


jiayiwang7:
**What happened**:

Currently in EKS-A, we use a default `nodeStartupTimeout: 10m0s` in machine health check for all the CP and worker node machines.

When creating a big cluster, e.g. a Snow cluster with 10+ worker nodes, since worker machines get created in the beginning (in pending state), highly likely some of the worker nodes will not be ready within the default 10 mins, causing the machine health check fails. CAPI controller then deletes those worker machines and creates new ones. However, these deleted machines' corresponding nodes could not be removed from the cluster, and got stuck in `NotReady` state (upstream CAPI issue, https://github.com/aws/eks-anywhere/issues/4708). In this weird stage, cluster is running normally with the desired count of Ready nodes, but later the upgrade preflight validation will fail due to "node is not ready" error.

```sh
$ kubectl --kubeconfig=chy-cluster/chy-cluster-eks-a-cluster.kubeconfig get nodes -owide -A
NAME                                STATUS     ROLES                  AGE     VERSION                INTERNAL-IP     EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION    CONTAINER-RUNTIME
chy-cluster-control-plane-1-4kws4   Ready      control-plane,master   9m37s   v1.23.13-eks-6022eca   192.168.1.186   <none>        Ubuntu 20.04.5 LTS   5.15.0-1026-aws   containerd://1.5.9
chy-cluster-control-plane-1-ct26v   Ready      control-plane,master   6m44s   v1.23.13-eks-6022eca   192.168.1.111   <none>        Ubuntu 20.04.5 LTS   5.15.0-1026-aws   containerd://1.5.9
chy-cluster-control-plane-1-jrfdf   Ready      control-plane,master   12m     v1.23.13-eks-6022eca   192.168.1.112   <none>        Ubuntu 20.04.5 LTS   5.15.0-1026-aws   containerd://1.5.9
chy-cluster-control-plane-1-x5t7k   Ready      control-plane,master   31m     v1.23.13-eks-6022eca   192.168.1.108   <none>        Ubuntu 20.04.5 LTS   5.15.0-1026-aws   containerd://1.5.9
[...]

PR #4727, opened 2023-01-23
Add option to override default nodeStartupTimeout in machine health c…

jiayiwang7:
…heck in create and upgrade commands

*Issue #, if available:*

https://github.com/aws/eks-anywhere/issues/4709

*Description of changes:*

*Testing (if applicable):*


vivek-koppuru (approved):
I think in the future when we add another timeout, we need to definitely create a timeouts struct, or have it as an action item.

PR #4744, opened 2023-01-24
Fix various gci import problems

TerryHowe:
Various gci problems in the code

### DIFF
--- a/internal/aws-sdk-go-v2/internal/endpoints/v2/endpoints.go
+++ b/internal/aws-sdk-go-v2/internal/endpoints/v2/endpoints.go
@@ -5,9 +5,8 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/aws/smithy-go/logging"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/smithy-go/logging"
 )
 
 // DefaultKey is a compound map key of a variant and other values.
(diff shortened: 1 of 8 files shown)


internal/aws-sdk-go-v2/internal/endpoints/v2/endpoints.go
```diff
@@ -5,9 +5,8 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/aws/smithy-go/logging"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/smithy-go/logging"
```
g-gaston, line 9: this is supposed to be a copy of library, I think we should keep it unchanged and just ignore it in the linter config (I think it already is?)

TerryHowe: I just ran gci on everything and this came up. I'm not sure why I'd make an exception for historical reasons. What is the difference between smithy-go/logging copy and aws-sdk-go-v2? They are both libraries managed by aws. It is a much easier rule to follow.

g-gaston: I meant `internal/aws-sdk-go-v2` is a copy of a library. We copied it here because it's not available upstream yet (I think that was the reason).

PR #4764, opened 2023-01-26
Generate tinkerbell control plane CAPI objects

ddjjia:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/985

*Description of changes:*

*Testing (if applicable):*


pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// BaseControlPlane represents a CAPI Tinkerbell control plane.
+type BaseControlPlane = clusterapi.ControlPlane[*tinkerbellv1.TinkerbellCluster, *tinkerbellv1.TinkerbellMachineTemplate]
+
+// ControlPlane holds the Tinkerbell specific objects for a CAPI Tinkerbell control plane.
+type ControlPlane struct {
+	BaseControlPlane
+	Secrets []*corev1.Secret
```
taneyland, line 29: There is only one secret object related to Tinkerbell CP objects so can we just add the individual secret obj instead of defining this as a llst?

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,204 @@
...
+		err := diskExtractor.Register(controlPlaneMachineSpec.HardwareSelector)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return controlPlaneMachineSpec, nil
+}
+
+func getWorkerNodeGroupMachineSpec(clusterSpec *cluster.Spec, diskExtractor *hardware.DiskExtractor) (map[string]anywherev1.TinkerbellMachineConfigSpec, error) {
```
taneyland, line 60: Should we move this to a `workers.go` file since it should be separate from cp objects? Also double-check with @tatlat to ensure that this doesn't conflict with his workers task

go.mod
```diff
@@ -33,13 +33,14 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.12.0
 	github.com/stretchr/testify v1.8.1
-	github.com/tinkerbell/rufio v0.0.0-20220606134123-599b7401b5cc
-	github.com/tinkerbell/tink v0.7.1-0.20221004171112-6deeea887dac
+	github.com/tinkerbell/cluster-api-provider-tinkerbell v0.4.0
```
pokearu, line 36: I am not sure we want to update these tags. There is like a whole initiative to upgrade tinkerbell dependencies and the stack itself. There were some complications due to API changes.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,204 @@
...
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return etcdMachineSpec, nil
+}
+
+func generateTemplateBuilder(clusterSpec *cluster.Spec) (providers.TemplateBuilder, error) {
```
taneyland, line 92: What do you think about moving all of these methods needed to generate the template builder to `template.go`?

pkg/providers/tinkerbell/controlplane.go
tatlat: So I actually already have something like this for my task. I don't think templateBulider.GenerateCAPISpecControlPlane() uses WorkerNodeGroupMachineSpecs. I might be wrong but I think you could probably pass a nil value and it would still generate the control plane objects. Maybe it would be better to put all the methods for getting machine specs (cp, etcd, workers) and generating the template builder into one file that we could share.

go.mod
mitalipaygude: This change was probably added because of importing CAPT to AddToScheme for tinkerbell.

pkg/providers/tinkerbell/controlplane_test.go
```diff
@@ -0,0 +1,462 @@
...
+			expected: []kubernetes.Object{
+				capiCluster(),
+				tinkerbellCluster(),
+				kubeadmControlPlane(),
+				tinkerbellMachineTemplate("controlplane-machinetemplate"),
+				secret(),
+			},
+		},
+		{
+			name: "unstacked etcd",
```
mitalipaygude, line 51: We don't support unstacked etcd for Tinkerbell.

go.mod
ddjjia: The changes are introduced by importing CAPT, but I required rufio revision at the bottom

pokearu: But there is a breaking api change in rufio that will cause other issues. Can we just import an older version of CAPT instead? That will simplify things here.

ddjjia: That's why I require an older version of rufio at [bottom](https://github.com/aws/eks-anywhere/pull/4764/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R214). 
We didn't have CAPT imported, is there a specific version in your mind, or the latest that doesn't change rufio?

pkg/providers/tinkerbell/controlplane.go
ddjjia: sounds good. I'll move the template builder out to a common methods

go.mod
pokearu: If we want to have minimum impact we might want to use the same commit we build CAPT on.

https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/tinkerbell/cluster-api-provider-tinkerbell/GIT_TAG

mitalipaygude:
Can we increase the unit test coverage more?

ddjjia:
> Can we increase the unit test coverage more?

Added some tests on template builder. Please lmk if you have specific cases want to cover. @mitalipaygude

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,120 @@
...
+	templateBuilder, err := generateTemplateBuilder(clusterSpec)
+	if err != nil {
+		return nil, errors.Wrap(err, "generating tinkerbell template builder")
+	}
+
+	controlPlaneYaml, err := templateBuilder.GenerateCAPISpecControlPlane(
+		clusterSpec,
+		func(values map[string]interface{}) {
+			values["controlPlaneTemplateName"] = clusterapi.ControlPlaneMachineTemplateName(clusterSpec.Cluster)
+			values["etcdTemplateName"] = clusterapi.EtcdMachineTemplateName(clusterSpec.Cluster)
```
chrisdoherty4, line 47: We don't support external etcd so I suspect we can remove this.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,120 @@
...
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// BaseControlPlane represents a CAPI Tinkerbell control plane.
+type BaseControlPlane = clusterapi.ControlPlane[*tinkerbellv1.TinkerbellCluster, *tinkerbellv1.TinkerbellMachineTemplate]
+
+// ControlPlane holds the Tinkerbell specific objects for a CAPI Tinkerbell control plane.
+type ControlPlane struct {
```
chrisdoherty4, line 23: Do we anticipate needing to consume this in other packages?

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	yamlcapi "github.com/aws/eks-anywhere/pkg/clusterapi/yaml"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/yamlutil"
+)
+
+// BaseControlPlane represents a CAPI Tinkerbell control plane.
+type BaseControlPlane = clusterapi.ControlPlane[*tinkerbellv1.TinkerbellCluster, *tinkerbellv1.TinkerbellMachineTemplate]
```
chrisdoherty4, line 20: Do we anticipate needing to consume this in other packages? If not can we unexport.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,120 @@
...
+
+// ControlPlane holds the Tinkerbell specific objects for a CAPI Tinkerbell control plane.
+type ControlPlane struct {
+	BaseControlPlane
+	Secrets *corev1.Secret
```
chrisdoherty4, line 25: Could we use singular form given we aren't referencing multiple.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -0,0 +1,120 @@
...
+// Objects returns the control plane objects associated with the Tinkerbell cluster.
+func (p ControlPlane) Objects() []kubernetes.Object {
+	o := p.BaseControlPlane.Objects()
+	o = append(o, p.Secrets)
+
+	return o
+}
+
+// ControlPlaneSpec builds a Tinkerbell ControlPlane definition based on an eks-a cluster spec.
+func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes.Client, clusterSpec *cluster.Spec) (*ControlPlane, error) {
```
chrisdoherty4, line 37: Whats the rationale for deviating from `New*` funcs?

pkg/providers/tinkerbell/machinetemplate.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/pkg/errors"
+	tinkerbellv1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/api/v1beta1"
+	"k8s.io/apimachinery/pkg/api/equality"
+
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+)
+
+// GetMachineTemplate gets a TinkerbellMachineTemplate object using the provided client
+// If the object doesn't exist, it returns a NotFound error.
+func GetMachineTemplate(ctx context.Context, client kubernetes.Client, name, namespace string) (*tinkerbellv1.TinkerbellMachineTemplate, error) {
```
chrisdoherty4, line 15: Is this used externally to this package by anything?

pkg/providers/tinkerbell/controlplane.go
ddjjia: Shall we remove this logic for CLI as well? https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/tinkerbell.go#L130

pkg/providers/tinkerbell/controlplane.go
ddjjia: It will be used in reconciler

pkg/providers/tinkerbell/controlplane.go
ddjjia: I try to align with other provider naming convention.

pkg/providers/tinkerbell/machinetemplate.go
ddjjia: `WorkersSpec` would use it as well. https://github.com/aws/eks-anywhere/pull/4764#discussion_r1088339772

PR #4768, opened 2023-01-26
Add pointer to Nutanix getting started guide

chrisnegus:
*Description of changes:* I noticed an EKS Anywhere on Nutanix blog pointing to the Nutanix reference page instead of to the getting started guide. I added a pointer to the getting started guide in case someone lands on that reference page first.


abhinavmpandey08 (approved):
/lgtm
/approve

PR #4773, opened 2023-01-26
Add baremetal airgapped test

d8660091:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


test/framework/cluster.go
```diff
@@ -1970,3 +1963,15 @@ func (e *ClusterE2ETest) AssertAirgappedNetwork() {
 		e.T.Fatalf("Airgap user is not airgapped")
 	}
 }
+
+func dumpFile(description string, path string, t T) {
```
abhay-krishna, line 1967: Follow up with `defer file.Close()`.

abhay-krishna: Any reason we couldn't do
```golang
b, err := os.ReadFile(path)
```
?

cmd/integration_test/build/buildspecs/tinkerbell-test-eks-a-cli.yml
```diff
@@ -60,10 +61,14 @@ phases:
...
       - ${CODEBUILD_SRC_DIR}/cmd/integration_test/build/script/start_docker.sh
       - export CLUSTER_NAME_PREFIX="${BRANCH_NAME//./-}"
-      - >
-        ./bin/test e2e cleanup vsphere
-        -n ${CLUSTER_NAME_PREFIX}
-        -v 4
+      - make build-integration-test-binary
+      - make build-eks-a-for-e2e
+      - make e2e-tests-binary E2E_TAGS="e2e tinkerbell" E2E_OUTPUT_FILE=bin/tinkerbell/e2e.test
+      - curl -s https://dev-release-prod-pdx.s3.us-west-2.amazonaws.com/bundle-release.yaml?versionId=7ZBFyTiC5_3RRfPKGvzMaA6DlmW2yExs -o bin/local-bundle-release.yaml
```
abhay-krishna, line 67: Try without the version override? I think it's been fixed.

d8660091: I didn't try, but this versionId was newer than the one you used. It should be good

internal/test/e2e/run.go
```diff
@@ -411,6 +417,34 @@ func splitTinkerbellTests(awsSession *session.Session, testsList []string, conf
 	return runConfs, nil
 }
 
+func runAirgappedTinkerbellTests(awsSession *session.Session, testsList []string, conf ParallelRunConf, testRunnerConfig *TestInfraConfig, runConfs []instanceRunConf) ([]instanceRunConf, error) {
```
jiayiwang7, line 420: Can we rename the method? it sounds like it launches and runs the actual tests but in fact it only appends test runconfig to the runConf list

internal/test/e2e/run.go
```diff
@@ -411,6 +417,34 @@ func splitTinkerbellTests(awsSession *session.Session, testsList []string, conf
...
+	if len(tinkerbellTests) == 0 {
+		conf.Logger.V(1).Info("No tinkerbell airgapped test to run")
+		return runConfs, nil
+	}
+
+	conf.Logger.V(1).Info("INFO:", "tinkerbellAirGappedTests", len(tinkerbellTests))
+
+	err := s3.DownloadToDisk(awsSession, os.Getenv(tinkerbellAirgappedHardwareS3FileKeyEnvVar), conf.StorageBucket, e2eAirgappedHardwareCsvFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to download tinkerbell airgapped hardware csv: %v", err)
```
jiayiwang7, line 431: nit: i think we are avoiding "failed to xxx" for all error statement

test/e2e/airgap.go
```diff
@@ -4,11 +4,20 @@
...
+	"fmt"
+	"io"
+	"net/http"
 	"os"
+	"regexp"
 
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
+const bottlerocketOSFileName = "bottlerocket.img.gz"
```
jiayiwang7, line 17:
```suggestion
const (
    bottlerocketOSFileName = "bottlerocket.img.gz"
    ...
)
```

internal/test/e2e/run.go
```diff
@@ -321,16 +322,21 @@ func splitTests(testsList []string, conf ParallelRunConf) ([]instanceRunConf, er
 	}
 
 	if strings.EqualFold(conf.BranchName, conf.BaremetalBranchName) {
-		runConfs, err = splitTinkerbellTests(awsSession, testsList, conf, testRunnerConfig, runConfs)
+		runConfs, err = splitNonAirgapTinkerbellTests(awsSession, testsList, conf, testRunnerConfig, runConfs)
```
jiayiwang7, line 325: Confused by the name and method: we generate the non airgap test runConfs first, then append the airgap test runConfigs after?

This is kinda hard to review. Can we decouple the 2 and write methods that only deal w each type of testing? e.g.

```go
func airgapRunConfs(airgapTests []string) []runConf 

func nonAirgapRunConfs(nonAirgapTests []string) []runConf  
```

d8660091: I have renamed the two functions to make them easier to reason about. I think a more serious refactor would be nice, but it's pretty hard to include the refactoring in this PR.

internal/test/e2e/run.go
d8660091: I don't mind changing the error statement, but I'm not sure which is our preferred alternative. And this error message is not user facing, I think it should be fine.

jiayiwang7:
```suggestion
		return nil, fmt.Errorf("downloading tinkerbell airgapped hardware csv: %v", err)
```

jiayiwang7 (approved):

PR #4795, opened 2023-01-30
Flatten arch array to string

abhay-krishna:
Architecture is an array so we should flatten it to a string for use with `Sprintf`.


jonathanmeier5 (approved):
/lgtm
/approve

abhay-krishna:
Force merging to expedite release.

PR #4797, opened 2023-01-30
Update packagecontroller install command for registry mirror

junshun:
Uses registry mirror for package controller install if specified in cluster spec.


TerryHowe (approved):
/lgtm
/approve

PR #4798, opened 2023-01-31
Determine diskGiB default using template disk size for vSphere

abhinavmpandey08:
*Description of changes:*
vSphere fullClone currently fails if the user doesn't specify `diskGiB` for a machine and the template disk size is greater than 20 GiB. This is because `diskGiB` for a machine can't be less than the template's disk size but EKS-A defaults diskGiB to 20, if not specified.
```
Error getting disk config spec for primary disk: can't resize template disk down, initial capacity is larger: 23068672KiB > 20971520KiB
```

With this change, the new default for `diskGiB` will be the `max(20, templateDiskSize)`

### DIFF
--- a/pkg/providers/vsphere/defaults.go
+++ b/pkg/providers/vsphere/defaults.go
@@ -12,6 +12,8 @@ import (
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+const minDiskGib int = 20
+
 type Defaulter struct {
 	govc ProviderGovcClient
 }
@@ -34,11 +36,9 @@ func (d *Defaulter) setDefaultsForMachineConfig(ctx context.Context, spec *Spec)
 			return err
 		}
 
-		if err := d.setCloneModeDefaults(ctx, m, spec.VSphereDatacenter.Spec.Datacenter); err != nil {
+		if err := d.setCloneModeAndDiskSizeDefaults(ctx, m, spec.VSphereDatacenter.Spec.Datacenter); err != nil {
 			return err
 		}
-
-		d.setDiskSizeDefaults(m)
 	}
 
 	return nil
@@ -101,10 +101,24 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
 	return nil
 }
 
-func (d *Defaulter) setCloneModeDefaults(ctx context.Context, machineConfig *anywherev1.VSphereMachineConfig, datacenter string) error {
-	diskSize, err := d.govc.GetVMDiskSizeInGB(ctx, machineConfig.Spec.Template, datacenter)
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func (d *Defaulter) setCloneModeAndDiskSizeDefaults(ctx context.Context, machineConfig *anywherev1.VSphereMachineConfig, datacenter string) error {
+	templateDiskSize, err := d.govc.GetVMDiskSizeInGB(ctx, machineConfig.Spec.Template, datacenter)
 	if err != nil {
 		return fmt.Errorf("getting disk size for template %s: %v", machineConfig.Spec.Template, err)
+	}
+
+	minDiskSize := max(minDiskGib, templateDiskSize)
+
+	if machineConfig.Spec.DiskGiB < minDiskSize {
+		logger.Info("Warning: VSphereMachineConfig DiskGiB cannot be less than %s. Defaulting to %s.", minDiskSize, minDiskSize)
+		machineConfig.Spec.DiskGiB = minDiskSize
 	}
 
 	templateHasSnapshot, err := d.govc.TemplateHasSnapshot(ctx, machineConfig.Spec.Template)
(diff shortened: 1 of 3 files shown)


pkg/providers/vsphere/defaults.go
```diff
@@ -101,12 +101,27 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
 	return nil
 }
 
-func (d *Defaulter) setCloneModeDefaults(ctx context.Context, machineConfig *anywherev1.VSphereMachineConfig, datacenter string) error {
+func max(a, b int) int {
```
vivek-koppuru, line 104: Why not just use the built-in `Math.Max()` function?

pkg/providers/vsphere/defaults.go
```diff
@@ -101,12 +101,27 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
+	if a > b {
+		return a
+	}
+	return b
+}
+
+//nolint:gocyclo // skip cyclomatic complexity check for this function
```
vivek-koppuru, line 111: This needs to be skipped? What is the error that lint is complaining about and how many is it?

pkg/providers/vsphere/defaults.go
```diff
@@ -101,12 +101,27 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
+func (d *Defaulter) setCloneModeAndDiskSizeDefaults(ctx context.Context, machineConfig *anywherev1.VSphereMachineConfig, datacenter string) error {
 	diskSize, err := d.govc.GetVMDiskSizeInGB(ctx, machineConfig.Spec.Template, datacenter)
```
vivek-koppuru, line 113: Does this actually return GB or GiB? There is a difference in units if GB so we might need to do some conversion here to do a proper check.

pkg/providers/vsphere/defaults.go
abhinavmpandey08: [Math.Max ](https://pkg.go.dev/math#Max)operates on float64s only. The values being compared here are ints to it's would require a bunch of type conversions. This just makes things easier.

pkg/providers/vsphere/defaults.go
abhinavmpandey08: The [additional if condition](https://github.com/aws/eks-anywhere/pull/4798/files#diff-9d67b5f25923ee894de7e47cdf75a45f3224cde45ac7d518a97dfc8704bc8254R120-R123) bumps up the cyclomatic complexity to 11. I think 10 is the allowed value.

pkg/providers/vsphere/defaults.go
abhinavmpandey08: this part is a bit ambiguous. The govc command we use to retrieve this info is 
```bash
$ govc device.info -vm bottlerocket-kube-v1-24 -json | grep CapacityInKB
      "CapacityInKB": 23068672,
```
which returns `CapacityInKB`.
vSphere UI also shows the disk size in GB not GiB. It's just CAPV asks for disk size in GiB. But when I create a machine with `25GiB` using CAPV, the vSphere UI and that govc command shows the disk size as `25GB`

pkg/providers/vsphere/defaults.go
vivek-koppuru: Yea but you can pass ints too. It shouldn't change the result right?

pkg/providers/vsphere/defaults.go
vivek-koppuru: Oh wow we create a disk with gb instead of gib? Sounds like a bug then right? We can probably create an issue for that then

pkg/providers/vsphere/defaults.go
vivek-koppuru: Instead of bypassing the lint can we separate some part to a different function?

pkg/providers/vsphere/defaults.go
abhinavmpandey08: Passing ints doesn't work: https://go.dev/play/p/9xwiJu-dDGf
So I would have to convert the inputs to float64 and the output back to int, which doesn't look very clean. Here's what that would look like: https://go.dev/play/p/ylG5-_FhjFd

pkg/providers/vsphere/defaults.go
abhinavmpandey08: yeah I split some of the logic in separate functions so it's good now

pkg/providers/vsphere/defaults.go
abhinavmpandey08: Idk if it's a bug or just ambiguity in the terminology between CAPV and vSphere.
That will require a bit more research so I can create a separate issue to look into that.

pkg/providers/vsphere/defaults.go
vivek-koppuru: I was confused with github.com/pkg/math, this is fine then

vivek-koppuru (approved):

pkg/providers/vsphere/defaults.go
abhinavmpandey08: Create the ticket here: https://github.com/aws/eks-anywhere/issues/4839

pkg/providers/vsphere/defaults.go
```diff
@@ -101,59 +101,74 @@ func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machin
...
+func (d *Defaulter) setCloneModeAndDiskSizeDefaults(ctx context.Context, machineConfig *anywherev1.VSphereMachineConfig, datacenter string) error {
+	templateDiskSize, err := d.govc.GetVMDiskSizeInGB(ctx, machineConfig.Spec.Template, datacenter)
 	if err != nil {
 		return fmt.Errorf("getting disk size for template %s: %v", machineConfig.Spec.Template, err)
 	}
 
+	minDiskSize := max(minDiskGib, templateDiskSize)
+
+	if machineConfig.Spec.DiskGiB < minDiskSize {
+		logger.Info("Warning: VSphereMachineConfig DiskGiB cannot be less than %s. Defaulting to %s.", minDiskSize, minDiskSize)
```
vivek-koppuru, line 120: We need to change this to an fmt.Sprintf, I will follow up with a PR with this change.

vivek-koppuru: https://github.com/aws/eks-anywhere/pull/4883

PR #4808, opened 2023-01-31
validate worker tinkerbell IP matches management

raymond-zhang00:
*Issue #, if available:*
[1041](https://github.com/aws/eks-anywhere-internal/issues/1041)

*Description of changes:*
Validates that worker cluster Tinkerbell IP matches that of the management cluster for the cluster full lifecycle project. 

*Testing (if applicable):*
unit tests

### DIFF
--- a/pkg/providers/tinkerbell/reconciler/reconciler.go
+++ b/pkg/providers/tinkerbell/reconciler/reconciler.go
@@ -64,6 +64,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *an
 		r.ipValidator.ValidateControlPlaneIP,
 		r.ValidateClusterSpec,
 		r.ValidateHardware,
+		r.ValidateDatacenterConfig,
 		r.ReconcileControlPlane,
 		r.CheckControlPlaneReady,
 		r.ReconcileCNI,
@@ -114,6 +115,20 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger, 
 	return controller.Result{}, nil
 }
 
+// ValidateDatacenterConfig updates the cluster status if the TinkerbellDatacenter status indicates that the spec is invalid.
+func (r *Reconciler) ValidateDatacenterConfig(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "validateDatacenterConfig")
+
+	if err := r.validateTinkerbellIPMatch(ctx, clusterSpec); err != nil {
+		log.Error(err, "Invalid TinkerbellDatacenterConfig")
+		failureMessage := err.Error()
+		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
+		return controller.ResultWithReturn(), nil
+	}
+
+	return controller.Result{}, nil
+}
+
 // ReconcileCNI reconciles the CNI to the desired state.
 func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
 	log = log.WithValues("phase", "reconcileCNI")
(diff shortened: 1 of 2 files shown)


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -97,6 +99,44 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
...
+		err = r.client.Get(ctx, client.ObjectKey{
+			Namespace: clusterSpec.Cluster.Namespace,
+			Name:      managementClusterSpec.Spec.DatacenterRef.Name,
+		}, managementDatacenterConfig)
+		if err != nil {
+			return controller.Result{}, err
+		}
+
+		if dataCenterConfig.Spec.TinkerbellIP != managementDatacenterConfig.Spec.TinkerbellIP {
+			log.Error(err, "Invalid TinkerbellDatacenterConfig")
```
ddjjia, line 131: err should be nil here.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -97,6 +99,44 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
...
+			return controller.Result{}, fmt.Errorf("workload cluster Tinkerbell IP must match managment cluster Tinkerbell IP")
```
ddjjia, line 132:
```suggestion
			return controller.Result{}, errors.New("workload cluster Tinkerbell IP must match managment cluster Tinkerbell IP")
```

pkg/providers/tinkerbell/reconciler/reconciler.go
raymond-zhang00: You're right, updating this log statement

raymond-zhang00: I ended up splitting the validation logic into it's own function used this log statement to catch the error from that.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -97,6 +99,20 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
 	return controller.Result{}, nil
 }
 
+// ValidateDatacenterConfig updates the cluster status if the TinkerbellDatacenter status indicates that the spec is invalid.
+func (r *Reconciler) ValidateDatacenterConfig(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
```
vivek-koppuru, line 103: Do we need any other validations on the ip itself?

raymond-zhang00:
Rebased and updated the tests to reference the new tt.cleanup() function.

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: You mean this one? https://github.com/aws/eks-anywhere/blob/25e7072b79804fe1a057d268688a672a56bab815/pkg/providers/tinkerbell/reconciler/reconciler.go#L62

vivek-koppuru: This is tinkerbell ip right? Isn't it different?

mitalipaygude: it is different yes

mitalipaygude:
I am wondering if we should block reconciliation till the user gives the correct Tinkerbell IP in the cluster spec. Since we are fetching the management cluster's datacenter config, we have access to the Tink IP we compare. Instead of checking if the user's provided IP matches the mgmt cluster's IP or not, we could just set the value of Tink IP to mgmt cluster's tink IP the way we do it in the [CLI](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/create.go#L172) today.

However this will cause inconsistency in what the user has in their cluster spec vs what the tink IP should be. we could throw a log statement out there which could be a potential way of letting user know about this override.

cxbrowne1207:
> I am wondering if we should block reconciliation till the user gives the correct Tinkerbell IP in the cluster spec. Since we are fetching the management cluster's datacenter config, we have access to the Tink IP we compare. Instead of checking if the user's provided IP matches the mgmt cluster's IP or not, we could just set the value of Tink IP to mgmt cluster's tink IP the way we do it in the [CLI](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/create.go#L172) today.
> 
> However this will cause inconsistency in what the user has in their cluster spec vs what the tink IP should be. we could throw a log statement out there which could be a potential way of letting user know about this override.

I would vote for blocking reconciliation until the user gives the correct TinkerbellIP, for the reason that there would be an inconsistency in the users cluster config if we override. It seems more like an invalid value case imo and I also don't think we should be creating discrepancies and expect the user to fix their version. I'm thinking that it's possible for the user to miss the log an remain unaware of the change?

If the TinkerbellIP was optional and not specified, then I would vote for an override in that case.

mitalipaygude:
Let's create a follow up task to discuss this and prioritize it later in the implementation. We should talk about the discussions that went into defaulting the Tink IP in the CLI workflow, and make sure we are using the best practices for this situation.

vivek-koppuru (approved):

PR #4819, opened 2023-02-01
Update eksa objects in cluster for skipped upgrade

g-gaston:
*Description of changes:*
This makes sure that even when a rolling upgrade is not needed, the eks-a objects reflect the latest state driven by the CLI. This includes new Bundles.

Before this, if there was a new Bundles that updated dependencies like CAPI or eks-a controller but didn't require a rolling upgrade, we didn't create the new Bundles in the cluster or update the cluster object to reference it. It's quite an edge case since most new Bundles come with new eks-d releases, which do trigger a rolling upgrade, so this probably never happened in prod.

*Testing (if applicable):*
Verified this works with docker provider and a new Bundles that only updates capi and cert-manager.

### DIFF
--- a/pkg/workflows/upgrade.go
+++ b/pkg/workflows/upgrade.go
@@ -110,7 +110,10 @@ type deleteBootstrapClusterTask struct {
 
 type updateClusterAndGitResources struct{}
 
-type resumeEksaReconcile struct {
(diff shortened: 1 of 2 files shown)


jiayiwang7 (approved):
logic looks good

taneyland (approved):
Looks good to me! Will let @jiayiwang7 take a look too

g-gaston:
unhold

PR #4822, opened 2023-02-01
Implement Tinkerbell reconcile CP

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR implements the Tinkerbell Cluster Reconciler phase: `ReconcileControlPlane`.

Additionally, a `TinkerbellUseDiskExtractorDefaultDisk` flag has been added that forces the diskExtractor to use `/dev/sda` disk by default as discussed. This is temporary until the diskExtractor logic can be removed from the reconciler.

*Testing (if applicable):*
Unit tests

### DIFF
--- a/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
@@ -10,6 +10,7 @@ import (
 
 	eksav1alpha1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/features"
 )
 
 // DiskExtractor represents a hardware labels and map it with the appropriate disk given in the hardware csv file.
@@ -79,6 +80,12 @@ func (d *DiskExtractor) InsertProvisionedHardwareDisks(hardware *tinkv1alpha1.Ha
 // GetDisk returns the disk cached for selector. If selector has no disk cached ErrDiskNotFound
 // is returned.
 func (d *DiskExtractor) GetDisk(selector eksav1alpha1.HardwareSelector) (string, error) {
+	// TODO: To be removed with the diskExtractor logic. For now this is a temporary solution
+	// while having to use the diskExtractor in the controller.
+	if features.TinkerbellUseDiskExtractorDefaultDisk().IsActive() {
+		return "/dev/sda", nil
+	}
+
 	key, err := serializeHardwareSelector(selector)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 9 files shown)


pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -79,6 +80,13 @@ func (d *DiskExtractor) InsertProvisionedHardwareDisks(hardware *tinkv1alpha1.Ha
 // GetDisk returns the disk cached for selector. If selector has no disk cached ErrDiskNotFound
 // is returned.
 func (d *DiskExtractor) GetDisk(selector eksav1alpha1.HardwareSelector) (string, error) {
+	// TODO: To be removed with the diskExtractor logic. For now this is a temporary solution
+	// while having to use the diskExtractor in the controller.
+	if features.TinkerbellUseDiskExtractorDefaultDisk().IsActive() {
+		fmt.Printf("HEEHFDES")
```
taneyland, line 86:
```suggestion

```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -108,3 +114,19 @@ func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, clusterS
 
 	return r.cniReconciler.Reconcile(ctx, log, client, clusterSpec)
 }
+
+func toClientControlPlane(cp *tinkerbell.ControlPlane) *clusters.ControlPlane {
+	other := make([]client.Object, 0, 1)
+	if cp.Secrets != nil {
```
taneyland, line 120: Are there scenarios where we don't set this Secret, causing it to equal `nil`?

cxbrowne1207: When I run the ReconcileControlPlane test, the secret is nil without error, so I assumed it could be. Would you happen to know if this is field is required or not?

pkg/providers/tinkerbell/reconciler/reconciler_test.go
```diff
@@ -44,7 +53,7 @@ func TestReconcilerReconcileSuccess(t *testing.T) {
 	tt.ipValidator.EXPECT().ValidateControlPlaneIP(tt.ctx, logger, tt.buildSpec()).Return(controller.Result{}, nil)
 
 	tt.remoteClientRegistry.EXPECT().GetClient(
-		tt.ctx, client.ObjectKey{Name: "workload-cluster", Namespace: "eksa-system"},
+		tt.ctx, client.ObjectKey{Name: "workload-cluster", Namespace: constants.EksaSystemNamespace},
```
tatlat, line 56: Not super necessary but you could assign assign "workload-cluster" to a variable since you use it multiple times within this function.

pkg/providers/tinkerbell/reconciler/reconciler.go
cxbrowne1207: After some investigation, I can confirm it can be `nil`.registryAuth is not set for the template [here](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/config/template-cp.yaml#L409
)

taneyland (approved):

PR #4825, opened 2023-02-01
update registry auth docs

ahreehong:
*Description of changes:*
Update registry mirror docs 


TerryHowe (approved):
/lgtm
/hold

I would think we'd want to merge this when we create a new bundle with new BR, no?

PR #4836, opened 2023-02-02
Fixing panic errors for tinkerbell FLC

pokearu:
*Description of changes:*
Ran into `Index out of bound` and `invalid memory address or nil pointer dereference [recovered]` errors when testing the controller for tinkerbell. Added fixes.

### DIFF
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
@@ -108,16 +108,21 @@ func validateImmutableFieldsTinkerbellMachineConfig(new, old *TinkerbellMachineC
 		allErrs = append(allErrs, field.Forbidden(specPath.Child("OSFamily"), "field is immutable"))
 	}
 
+	if len(new.Spec.Users) != len(old.Spec.Users) {
+		allErrs = append(allErrs, field.Forbidden(specPath.Child("Users"), "field is immutable"))
+		return allErrs
+	}
+
+	if new.Spec.Users[0].Name != old.Spec.Users[0].Name {
+		allErrs = append(allErrs, field.Forbidden(specPath.Child("Users[0].Name"), "field is immutable"))
+	}
+
 	if len(new.Spec.Users[0].SshAuthorizedKeys) != len(old.Spec.Users[0].SshAuthorizedKeys) {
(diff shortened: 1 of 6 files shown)


pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
```diff
@@ -108,7 +108,11 @@ func validateImmutableFieldsTinkerbellMachineConfig(new, old *TinkerbellMachineC
 		allErrs = append(allErrs, field.Forbidden(specPath.Child("OSFamily"), "field is immutable"))
 	}
 
-	if new.Spec.Users[0].SshAuthorizedKeys[0] != old.Spec.Users[0].SshAuthorizedKeys[0] {
+	if len(new.Spec.Users[0].SshAuthorizedKeys) != len(old.Spec.Users[0].SshAuthorizedKeys) {
```
vivek-koppuru, line 111: Should we check for Users len too then?

pokearu: I was curious about this. 
What is the idea around this? Because we dont validate this for other providers, it can be said that if the user generates their new clusterconfig, they would have users set by default but they need to fill in SSH keys and they can remove it or set it as `[]` which causes this to panic.

vivek-koppuru: yea my concern is if it is set to `[]` it would panic. If we dont have this for other providers sounds like something we need to fix there as well

vivek-koppuru: But i don't remember if we check for this before to avoid this place to run into that issue

pokearu: You mean we dont check for Users[0] anywhere so there is a lesser chance of panic? 
I could just add a len check just to be safe 👍

chrisdoherty4: Other places _should_ be checking this. Its bad practice to trust user input.

vivek-koppuru (approved):

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #4838, opened 2023-02-02
Tinkerbell doesn't support external etcd

ddjjia:
*Issue #, if available:*
- Remove the logic that build tinkerbell template on etcd.
- Secret in control plane should be singular.

*Description of changes:*

*Testing (if applicable):*


tatlat:
This looks good to remove unless we still plan to block etcd instead of removing it.

PR #4855, opened 2023-02-06
Tinkerbell controller hardware validations 

pokearu:
*Issue: https://github.com/aws/eks-anywhere-internal/issues/1045*

*Description of changes:*
Adding hardware validations for tinkerbell controller.

*Testing:*
`make docker-build` and tested the controller to ensure validations work.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -129,6 +129,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 // +kubebuilder:rbac:groups=distro.eks.amazonaws.com,resources=releases,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awssnowclusters;awssnowmachinetemplates;awssnowippools;vsphereclusters;vspheremachinetemplates;dockerclusters;dockermachinetemplates;tinkerbellclusters;tinkerbellmachinetemplates,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",namespace=eksa-system,resources=secrets,verbs=delete;
+// +kubebuilder:rbac:groups=tinkerbell.org,resources=hardware;hardware/status,verbs=get;list;watch;update;patch
 func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	log := ctrl.LoggerFrom(ctx)
 	// Fetch the Cluster object
--- a/pkg/providers/tinkerbell/hardware/etcd.go
+++ b/pkg/providers/tinkerbell/hardware/etcd.go
@@ -10,6 +10,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// OwnerNameLabel is the label set by CAPT to mark a hardware as part of a cluster.
+const OwnerNameLabel string = "v1alpha1.tinkerbell.org/ownerName"
+
 // ETCDReader reads the tinkerbell hardware objects from the cluster.
 // It holds the objects in a catalogue.
 type ETCDReader struct {
@@ -62,7 +65,7 @@ func (er *ETCDReader) getUnprovisionedTinkerbellHardware(ctx context.Context) ([
 	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
 		MatchExpressions: []metav1.LabelSelectorRequirement{
 			{
-				Key:      "v1alpha1.tinkerbell.org/ownerName",
+				Key:      OwnerNameLabel,
 				Operator: metav1.LabelSelectorOpDoesNotExist,
 			},
 		},
--- a/pkg/providers/tinkerbell/hardware/etcd.go
+++ b/pkg/providers/tinkerbell/hardware/etcd.go
@@ -1,0 +1,79 @@
+package hardware
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	tinkv1alpha1 "github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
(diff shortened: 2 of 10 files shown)


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -130,3 +133,51 @@ func toClientControlPlane(cp *tinkerbell.ControlPlane) *clusters.ControlPlane {
...
+	if err := etcdReader.NewCatalogueFromETCD(ctx); err != nil {
+		log.Error(err, "Hardware validation failure")
+		failureMessage := err.Error()
+		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
+
+		return controller.ResultWithReturn(), nil
+	}
+
+	var v tinkerbell.ClusterSpecValidator
+	v.Register(
```
mitalipaygude, line 163: These are the checks for create workflow right? Upgrade validations are different?

pokearu: There is one more assertion. We havent decided how we are going to distinguish create and upgrade for now though.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -130,3 +133,51 @@ func toClientControlPlane(cp *tinkerbell.ControlPlane) *clusters.ControlPlane {
...
+	}
+	if capiCluster != nil {
+		// If CAPI cluster exists, the hardware has been validated
+		// and it's possibly already in use so no need to validate it again.
+		log.V(3).Info("CAPI cluster already exists, skipping hardware validations")
+		return controller.Result{}, nil
+	}
+
+	// We need a new reader each time so that the catalogue gets recreated.
+	etcdReader := hardware.NewETCDReader(r.client)
```
panktishah26, line 153: Not as a scope of this PR but do we have plans to change current flow of reading hardwares from ETCD reader instead of directly from Catalogue?

pkg/providers/tinkerbell/hardware/etcd.go
```diff
@@ -0,0 +1,79 @@
...
+	return er.catalogue
+}
+
+// getUnprovisionedTinkerbellHardware fetches the tinkerbell hardware objects on the cluster which do not have an ownerName label.
+func (er *ETCDReader) getUnprovisionedTinkerbellHardware(ctx context.Context) ([]tinkv1alpha1.Hardware, error) {
+	var selectedHardware tinkv1alpha1.HardwareList
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "v1alpha1.tinkerbell.org/ownerName",
```
panktishah26, line 65: Can we store this label to a constant variable and reference it here?

panktishah26 (commented):
Your changes looks perfect, Thanks!

pkg/providers/tinkerbell/reconciler/reconciler.go
pokearu: Well not right now. We eventually want to have this handle concurrent cluster creates as well, so we might not use the catalogue at that point.

controllers/cluster_controller.go
```diff
@@ -129,6 +129,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger)
 // +kubebuilder:rbac:groups=distro.eks.amazonaws.com,resources=releases,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awssnowclusters;awssnowmachinetemplates;awssnowippools;vsphereclusters;vspheremachinetemplates;dockerclusters;dockermachinetemplates;tinkerbellclusters;tinkerbellmachinetemplates,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",namespace=eksa-system,resources=secrets,verbs=delete;
+// +kubebuilder:rbac:groups=tinkerbell.org,resources=hardware;hardware/status,verbs=get;list;watch;update;patch
```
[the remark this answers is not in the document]
abhinavmpandey08: Also do we need to give it bmc permissions since the catalog reads that as well?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -130,3 +133,51 @@ func toClientControlPlane(cp *tinkerbell.ControlPlane) *clusters.ControlPlane {
...
+	if err := etcdReader.NewCatalogueFromETCD(ctx); err != nil {
```
abhinavmpandey08, line 154: This function is just populating the catalog right? It's not doing any validations itself?

pokearu: oh not this is just for populating. It basically reads the objects from etcd and adds it to the catalogue.

controllers/cluster_controller.go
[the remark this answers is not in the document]
pokearu: > Also do we need to give it bmc permissions since the catalog reads that as well?

I was planning on a separate PR for this. With the BMC checks included.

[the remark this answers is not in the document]
pokearu: > should we restrict this to hardware objects only in the `eksa-system` ns?

I think we could? We do always expect hardware to be in the eksa namesapce.

pkg/providers/tinkerbell/hardware/etcd.go
```diff
@@ -0,0 +1,79 @@
...
+				Operator: metav1.LabelSelectorOpDoesNotExist,
+			},
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("converting label selector: %w", err)
+	}
+
+	if err := er.client.List(ctx, &selectedHardware, &client.ListOptions{LabelSelector: selector}); err != nil {
```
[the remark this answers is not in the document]
pokearu: `List` API call I think be default looks at all namespaces, but we can pass in an option to look at a specific namespace. 
Are you suggesting to try and limit to eksa-system?

abhinavmpandey08 (approved):
/lgtm
/approve
/woof
/woof

PR #4868, opened 2023-02-06
Validate control plane ip not conflict with snow admin instance ip

jiayiwang7:
*Issue #, if available:*

Part 2 of #4844 

*Description of changes:*

This PR validates the control plane ip should not conflict with snow admin instance ip in CLI preflight check.
A `LocalIMDSClient` is added in the snow validator only for CLI factory. We do not validate this in eks-a controller as the IMDS only connects to the local metadata service, and the eks-a webhook controller cannot fetch the remote admin instance IMDS.

*Testing (if applicable):*

- Unit tests
- Manual e2e test on snow device

```
2023-02-07T14:37:25.306Z	V0	❌ Validation failed	{"validation": "snow Provider setup is valid", "error": "setting defaults and validate snow config: invalid cluster config: control plane host ip cannot be same as the admin instance ip", "remediation": ""}
```


pkg/providers/snow/validator.go
```diff
@@ -12,17 +12,42 @@ const (
...
+func WithIMDS(imds LocalIMDSClient) ValidatorOpt {
+	return func(c *Validator) {
+		c.imds = imds
+	}
 }
 
-func NewValidator(clientRegistry ClientRegistry) *AwsClientValidator {
-	return &AwsClientValidator{
+// NewValidator creates a snow validator.
+func NewValidator(clientRegistry ClientRegistry, opts ...ValidatorOpt) *Validator {
```
chrisdoherty4, line 37: Is the imds client actually optional? `ValidateControlPlaneIP` will panic if that's missing.

pkg/aws/imds.go
```diff
@@ -19,6 +19,18 @@ func NewIMDSClient(config aws.Config) *imds.Client {
 	return imds.NewFromConfig(config)
 }
 
+// BuildIMDS builds or overrides the imds client in the Client with default aws config.
+func (c *Client) BuildIMDS(ctx context.Context) error {
```
chrisdoherty4, line 23: Is this required to make the `*Client` valid?

jiayiwang7: yes, not sure if im following what you mean here

pkg/providers/snow/validator.go
jiayiwang7: added nil check + unit test

chrisdoherty4: I think we want to make it an explicit requirement by specifying it as a parameter. That will ensure consumers can't accidentally create invalid objects.

Generally, optional arguments should be for optional things only.

jiayiwang7: the `imds` is optional. The validator is being used in both CLI and controller. As I mentioned in the PR description, we do not need an `imds` client for controller validation -- that's why i made it optional.

chrisdoherty4: I see, sorry for the confusion.

chrisdoherty4 (approved):

PR #4873, opened 2023-02-07
Patch tls flag in Tinkerbell stack deployment

chrisdoherty4:
The TLS flag is required in v0.8.0 of the Tink Server.


TerryHowe:
No way to set that to true?

abhinavmpandey08 (approved):
/lgtm
/approve

chrisdoherty4:
> No way to set that to true?

Nope. This is intentional. Bare Metal's security posture isn't great. The use of TLS is really a security fallacy because there is no secure way to get a root of trust on to the machine currently.

PR #4879, opened 2023-02-07
Tinkerbell Reconcile Workers Full Lifecycle

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -106,12 +107,30 @@ func (r *Reconciler) CheckControlPlaneReady(ctx context.Context, log logr.Logger
...
-	// Implement reconcile worker nodes here
+	log = log.WithValues("provider", "tinkerbell", "reconcile type", "workers")
+	clusterSpec, err := c.BuildSpec(ctx, clientutil.NewKubeClient(r.client), cluster)
+	if err != nil {
+		return controller.Result{}, errors.Wrap(err, "building cluster Spec for worker node reconcile")
+	}
 
-	return controller.Result{}, nil
+	return controller.NewPhaseRunner().Register(
+		r.ValidateClusterSpec,
```
mitalipaygude, line 119: We should be validating hardware here too as this reconciles worker nodes for self managed clusters.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -106,12 +107,30 @@ func (r *Reconciler) CheckControlPlaneReady(ctx context.Context, log logr.Logger
...
+		r.ReconcileWorkers,
+	).Run(ctx, log, clusterSpec)
+}
+
+// ReconcileWorkers applies the worker CAPI objects to the cluster.
+func (r *Reconciler) ReconcileWorkers(ctx context.Context, log logr.Logger, spec *c.Spec) (controller.Result, error) {
```
chrisdoherty4, line 125: The name of this func has a remarkable similarity to `ReconcileWorkerNodes`. Is there another name we can use?

---

Does this need to be exported?

cxbrowne1207 (commented):
Looks good, aside from needing to add validating the hardware like @mitalipaygude  mentioned.

pkg/providers/tinkerbell/reconciler/reconciler.go
tatlat: Yes, we do need to validate hardware here. I just forgot to add it here.

mitalipaygude: Let's make a back up task to track this once we figure out how to detect upgrades and link it here.
LGTM otherwise.

mitalipaygude:
Can we squash the commits into 1 before merging?

PR #4882, opened 2023-02-08
Remove disk constraints from Tinkerbell provider

chrisdoherty4:
## Overview

The Tinkerbell provider enforces a constraint that ensures all disk types used for a particular group of nodes (such as all control plane nodes) are the same. This is rooted in the re-use of the CAPT `TinkerbellMachineTemplate` for every node of the same type. Due to a limitation in Tinkerbell, it was necessary to statically define the disk on the template objects requiring all machines of the same type to use the same disk. 

Tinkerbell has since been upgraded to inject disk information at template rendering time. This facilitates the transition to templated values instead of static values for disks thereby removing the need for disk constraints in a CAPT context and subsequently EKS-A.

## Change set

* Change all default template generation to use template values instead of static disk values.
* Remove the disk extractor that had the responsibility of extracting the disks from all hardware used for a particular node grouping.
* Remove validation to enforce a single disk among node groupings.
* On Ubuntu OSs we now use reboot only instead of deciding between reboot or kexec. The original decision logic was an optimization where reboot was only used for NVMe drives due to bug. All other OSs uses a full reboot. Furthermore, we've had recommendations to avoid kexec as it can leave devices in an unknown state forcing the newly loaded kernel to recover.

## Manual Testing

* Created a management cluster and manually verified the workflows are rendered correctly.
* Created a workload cluster and manually verified the workflows are rendered correctly.
* Upgraded a management cluster.


pkg/api/v1alpha1/tinkerbelltemplateconfig_defaults.go
```diff
@@ -21,76 +21,80 @@ warnings:
...
-func GetDefaultActionsFromBundle(b v1alpha1.VersionsBundle, disk, osImageOverride, tinkerbellLocalIp, tinkerbellLBIp string, osFamily OSFamily) []ActionOpt {
-	var diskPart string
+	// During workflow reconciliation when the Tinkerbell template is rendered, the Workflow
+	// Controller injects a subset of data from the Hardware resource. This lets us use Go template
+	// language to render the disks enabling mix'n'match disk types for templates that represent
+	// the same kind of machine such as control plane nodes.
+	//
+	// The devicePath disk index and the storagePartitionPath disk index should match.
+	const devicePath = "{{ index .Hardware.Disks 0 }}"
+	const paritionPathFmt = "{{ formatPartition ( index .Hardware.Disks 0 ) %s }}"
```
vivek-koppuru, line 43: So these are template strings now instead of the actual values so yaml can replace them? Also any reason why they are consts defined within a function vs just a variable, you want to avoid mutability of it? If so, why not define them above where it can be used by the unit tests as well?

chrisdoherty4: We can expand the variable names further, although I'm not sure I see the value in the addition given the scope of the variable.

I'm not quite following what the 'Condition' part is referencing, could you elaborate?

They're denoted as constants so they're allocated once. I don't think they need to be made part of the package API though, particularly given this is the API package and already bloated.

chrisdoherty4: For clarity. 'devicePath' is being used because it's used in place of various Tinkerbell template device path parameters. The fact it's a templated value felt of less importance.

jacobweinstock (approved):

PR #4891, opened 2023-02-08
Validate snow instance type has at least 2 default vCPU

jiayiwang7:
*Issue #, if available:*

Add vCPU check for snow instance type as kubelet requires at least 2 vCPU.

*Description of changes:*

*Testing (if applicable):*

Unit tests + manual e2e

### DIFF
--- a/pkg/providers/snow/validator.go
+++ b/pkg/providers/snow/validator.go
@@ -6,7 +6,6 @@ import (
 	"reflect"
 
 	"github.com/pkg/errors"
-	"k8s.io/utils/strings/slices"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 )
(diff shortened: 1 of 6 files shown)


pkg/providers/snow/validator.go
```diff
@@ -134,6 +133,27 @@ func (v *Validator) ValidateDeviceIsUnlocked(ctx context.Context, m *v1alpha1.Sn
...
+	if err != nil {
+		return fmt.Errorf("fetching supported instance types for device [%s]: %v", deviceIP, err)
+	}
+
+	for _, it := range instanceTypes {
+		if instanceType != it.Name {
+			continue
+		}
+
+		if it.DefaultVCPU != nil && *it.DefaultVCPU < minimumVCPU {
```
vivek-koppuru, line 147: Is it not possible to do `it.DefaultVCPU < minimumVCPU ` instead?

vivek-koppuru: Nvm I see why, the int is a pointer.

vivek-koppuru (approved):

PR #4899, opened 2023-02-09
Add defaults for Tinkerbell Machine Config

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Add defaults for Tinkerbell Machine Config
*Testing (if applicable):*
- Unit tests

### DIFF
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig.go
@@ -129,3 +129,9 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
 
 	return nil
 }
+
+func setTinkerbellMachineConfigDefaults(machineConfig *TinkerbellMachineConfig) {
+	if machineConfig.Spec.OSFamily == "" {
+		machineConfig.Spec.OSFamily = Bottlerocket
+	}
+}
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig_types_test.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig_types_test.go
@@ -8,7 +8,7 @@ import (
 )
 
 func TestTinkerbellMachineConfigValidateSucceed(t *testing.T) {
-	machineConfig := createTinkerbellMachineConfig()
+	machineConfig := CreateTinkerbellMachineConfig()
 
 	g := NewWithT(t)
 	g.Expect(machineConfig.Validate()).To(Succeed())
@@ -22,42 +22,42 @@ func TestTinkerbellMachineConfigValidateFail(t *testing.T) {
 	}{
 		{
 			name: "Invalid object meta",
-			machineConfig: createTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
+			machineConfig: CreateTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
 				mc.ObjectMeta.Name = ""
 			}),
 			expectedErr: "TinkerbellMachineConfig: missing name",
 		},
 		{
 			name: "Empty hardware selector",
-			machineConfig: createTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
+			machineConfig: CreateTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
 				mc.Spec.HardwareSelector = nil
 			}),
 			expectedErr: "TinkerbellMachineConfig: missing spec.hardwareSelector",
 		},
 		{
 			name: "Multiple hardware selectors",
-			machineConfig: createTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
+			machineConfig: CreateTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
 				mc.Spec.HardwareSelector["type2"] = "cp2"
 			}),
 			expectedErr: "TinkerbellMachineConfig: spec.hardwareSelector must contain only 1 key-value pair",
 		},
 		{
 			name: "Empty OS family",
-			machineConfig: createTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
+			machineConfig: CreateTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
 				mc.Spec.OSFamily = ""
 			}),
 			expectedErr: "TinkerbellMachineConfig: missing spec.osFamily",
 		},
 		{
 			name: "Invalid OS family",
-			machineConfig: createTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
+			machineConfig: CreateTinkerbellMachineConfig(func(mc *TinkerbellMachineConfig) {
 				mc.Spec.OSFamily = "invalid OS"
 			}),
 			expectedErr: "unsupported spec.osFamily (invalid OS); Please use one of the following",
 		},
 		{
 			name: "Invalid hostOSConfiguration",
-			machineConfig: createTinkerbellMachineConfig(
+			machineConfig: CreateTinkerbellMachineConfig(
 				withHostOSConfiguration(
 					&HostOSConfiguration{
 						NTPConfiguration: &NTPConfiguration{},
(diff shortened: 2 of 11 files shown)


pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -125,3 +127,27 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
...
+	if len(machineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
+		machineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
+	}
+
+	if machineConfig.Spec.OSFamily == "" {
+		logger.Info("Warning: OS family not specified in machine config specification. Defaulting to Bottlerocket.")
+		machineConfig.Spec.OSFamily = Bottlerocket
+	}
+
+	if len(machineConfig.Spec.Users) == 0 || machineConfig.Spec.Users[0].Name == "" {
```
pokearu, line 145: We don’t need to check len of users here again right? 
Since if it was 0 we set a default above.

mitalipaygude: We do because if the users array is an empty array, the 2nd condition machineConfig.Spec.Users[0].Name == "" could throw a panic error as it does not have a `Name` field in the empty array.

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -125,3 +127,27 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
 
 	return nil
 }
+
+func setTinkerbellMachineConfigDefaults(machineConfig *TinkerbellMachineConfig) {
```
vivek-koppuru, line 131: Are we doing these defaults just to be in sync with the cli or we determined that these are the defaults we want to move forward with?

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -125,3 +127,27 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
...
+	if len(machineConfig.Spec.Users) <= 0 {
+		machineConfig.Spec.Users = []UserConfiguration{{}}
+	}
+
+	if len(machineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
+		machineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
```
vivek-koppuru, line 137: So we are setting the sshauthorized keys for the first element separately from the name value? Shouldn't we just set it at the same time? Also if this is default to empty, do we generate an ssh key?

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -125,3 +127,27 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
...
+		machineConfig.Spec.OSFamily = Bottlerocket
+	}
+
+	if len(machineConfig.Spec.Users) == 0 || machineConfig.Spec.Users[0].Name == "" {
+		if machineConfig.Spec.OSFamily == Bottlerocket {
+			machineConfig.Spec.Users[0].Name = BottlerocketDefaultUser
+		} else {
+			machineConfig.Spec.Users[0].Name = UbuntuDefaultUser
+		}
+		logger.V(1).Info("SSHUsername is not set or is empty for TiinkerbellMachineConfig, using default", "machineConfig", machineConfig.Name, "user", machineConfig.Spec.Users[0].Name)
```
vivek-koppuru, line 151:
```suggestion
		logger.V(1).Info("SSHUsername is not set or is empty for TinkerbellMachineConfig, using default", "machineConfig", machineConfig.Name, "user", machineConfig.Spec.Users[0].Name)
```

pkg/api/v1alpha1/tinkerbellmachineconfig.go
mitalipaygude: We're doing this to get in sync with the cli for the most part. We mainly error out on webhooks if we don't create a users field. We also have these today for other providers.

pkg/api/v1alpha1/tinkerbellmachineconfig.go
mitalipaygude: Currently if it's empty we don't generate keys for any providers via the controller. We are prioritizing that effort soon though to generate the keys if the user doesn't specify the key.

pkg/api/v1alpha1/vspheremachineconfig.go
```diff
@@ -17,8 +17,8 @@ const (
 	DefaultVSphereNumCPUs    = 2
 	DefaultVSphereMemoryMiB  = 8192
 	DefaultVSphereOSFamily   = Bottlerocket
-	bottlerocketDefaultUser  = "ec2-user"
-	ubuntuDefaultUser        = "capv"
+	BottlerocketDefaultUser  = "ec2-user"
+	UbuntuDefaultUser        = "capv"
```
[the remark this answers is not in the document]
mitalipaygude: We are using it for both though, there is a reference in the same file below.l

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -125,3 +127,27 @@ func validateTinkerbellMachineConfig(config *TinkerbellMachineConfig) error {
...
+	if len(machineConfig.Spec.Users) <= 0 {
+		machineConfig.Spec.Users = []UserConfiguration{{}}
+	}
+
+	if len(machineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
+		machineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
+	}
+
+	if machineConfig.Spec.OSFamily == "" {
+		logger.Info("Warning: OS family not specified in machine config specification. Defaulting to Bottlerocket.")
```
g-gaston, line 141: I don't think we need this here
If we do, can we inject a logger?

pkg/api/v1alpha1/tinkerbellmachineconfig_types_test.go
```diff
@@ -67,7 +67,7 @@ func TestTinkerbellMachineConfigValidateFail(t *testing.T) {
 
 type tinkerbellMachineConfigOpt func(mc *TinkerbellMachineConfig)
 
-func createTinkerbellMachineConfig(options ...tinkerbellMachineConfigOpt) *TinkerbellMachineConfig {
+func CreateTinkerbellMachineConfig(options ...tinkerbellMachineConfigOpt) *TinkerbellMachineConfig {
```
g-gaston, line 70: curious why make this exported?

pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
```diff
@@ -36,7 +36,15 @@ func (r *TinkerbellMachineConfig) SetupWebhookWithManager(mgr ctrl.Manager) erro
...
 }
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+//+kubebuilder:webhook:path=/mutate-anywhere-eks-amazonaws-com-v1alpha1-tinkerbellmachineconfig,mutating=true,failurePolicy=fail,sideEffects=None,groups=anywhere.eks.amazonaws.com,resources=tinkerbellmachineconfigs,verbs=create;update,versions=v1alpha1,name=mutation.tinkerbellmachineconfig.anywhere.amazonaws.com,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Defaulter = &TinkerbellMachineConfig{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type.
+func (r *TinkerbellMachineConfig) Default() {
+	tinkerbellmachineconfiglog.Info("Setting up Tinkerbell Machine Config defaults", "name", r.Name, "namespace", r.Namespace)
```
g-gaston, line 45: prefer `klog.KObj` :)

pkg/api/v1alpha1/vspheremachineconfig.go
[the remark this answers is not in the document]
g-gaston: nit: if there are not exclusive for vsphere I would move them to a separate file

pkg/api/v1alpha1/tinkerbellmachineconfig_types_test.go
mitalipaygude: we use the same method in `tinkerbellmachineconfig_types_test` and `tinkerbellmachineconfig_webhook_test`

vivek-koppuru (approved):

PR #4901, opened 2023-02-10
Add management cluster name validation to controller

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
This is the same validation from https://github.com/aws/eks-anywhere/pull/4596 except from the controller side. Made some minor refactoring as well as fix a mockgen issue due to https://github.com/golang/mock/issues/494, from which I followed https://github.com/golang/mock#reflect-vendoring-error

Chose to do validation in controller instead of webhook as we have access to client there that will allow us to retrieve the older object. 

*Testing (if applicable):*
functional testing and unit testing


controllers/cluster_controller_test.go
```diff
@@ -67,7 +67,7 @@ func newVsphereClusterReconcilerTest(t *testing.T, objs ...runtime.Object) *vsph
 		Add(anywherev1.VSphereDatacenterKind, reconciler).
 		Build()
 
-	r := controllers.NewClusterReconciler(cl, &registry, iam)
+	r := controllers.NewClusterReconciler(cl, &registry, iam, clusters.NewClusterValidator(cl))
```
taneyland, line 70: should we use the mock cluster validator for the tests in this file?

taneyland (approved):

cxbrowne1207 (approved):
/lgtm

controllers/cluster_controller.go
```diff
@@ -54,12 +55,18 @@ type AWSIamConfigReconciler interface {
 	ReconcileDelete(ctx context.Context, logger logr.Logger, cluster *anywherev1.Cluster) error
 }
 
+// ClusterValidator runs cluster level preflight validations before it goes to provider reconciler.
+type ClusterValidator interface {
+	ValidateManagementClusterName(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error
+}
+
```
cxbrowne1207, line 62: This may clash with the ClusterValidator located in the e2e tests. It may be a good idea to merge them at some point

cxbrowne1207 (commented):
Left a comment FYI

controllers/cluster_controller.go
vivek-koppuru: I think it should be fine for now because they are independent enough, don't expect the interface to clash

PR #4906, opened 2023-02-10
Upload Hardware YAML to GitOps 

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR uploads the a generated hardware YAML to git during the CLI flow and enables management of hardware via GitOps. It adds hardware.yaml under resources of the `kuztomization.yaml ` file when using GitOPs to allow the Flux to recognize and apply changes when the user writes to hardware.yaml file in the repo. Without it flux does nothing when the hardware file is written.

*Testing (if applicable):*
Unit testing


pkg/providers/tinkerbell/hardware/csv.go
```diff
@@ -87,3 +88,23 @@ func ensureRequiredColumnsInCSV(unmatched []string) error {
...
+
+func BuildHardwareManifestFromCSV(path string) ([]byte, error) {
+	reader, err := NewNormalizedCSVReaderFromFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("csv: %v", err)
+	}
+
+	var b bytes.Buffer
+	bufferedWriter := io.Writer(&b)
+	writer := NewTinkerbellManifestYAML(bufferedWriter)
```
chrisdoherty4, line 100:
```suggestion
	writer := NewTinkerbellManifestYAML(&bufferedWriter)
```

pkg/gitops/flux/files.go
```diff
@@ -85,7 +88,13 @@ func (g *FileGenerator) WriteEksaFiles(clusterSpec *cluster.Spec, datacenterConf
 		return err
 	}
 
-	if err := g.WriteEksaKustomization(); err != nil {
+	if hardwareCSVPath != "" {
```
chrisdoherty4, line 91: If the caller knows whether or not a hardware CSV is available, would it make sense to just have that decide to invoke `WriteHardware` instead of pushing it through to here?

cxbrowne1207: I think that maintaining the write in this method is ultimately better, since it's responsible for writing all the files under the `eksa-system` folder.

PR #4907, opened 2023-02-11
Add Docker airgapped test

abhay-krishna:
Adding airgapped test for Docker. This also adds support for configuring registry mirror on Docker.


pkg/providers/docker/config/template-cp.yaml
```diff
@@ -138,6 +138,31 @@ spec:
 {{ .auditPolicy | indent 8 }}
       owner: root:root
       path: /etc/kubernetes/audit-policy.yaml
+{{- if .registryCACert }}
```
d8660091, line 141: Interesting that Docker didn't support registry mirror

test/e2e/airgap.go
```diff
@@ -22,3 +22,15 @@ func runVSphereAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs strin
...
 	test.DeleteCluster()
 }
+
+func runDockerAirgapConfigFlow(test *framework.ClusterE2ETest) {
+	test.GenerateClusterConfig()
+	test.DownloadArtifacts()
+	test.ExtractDownloadedArtifacts()
+	test.DownloadImages()
+	test.ImportImages()
+	test.ChangeInstanceSecurityGroup("sg-0f3b5c1368db406ee")
```
d8660091, line 32: Yet another environment variable. But I am not against it, I think it's good to have multiple ways to airgap the network

test/e2e/airgap.go
```diff
@@ -22,3 +22,15 @@ func runVSphereAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs strin
 	)
 	test.DeleteCluster()
 }
+
+func runDockerAirgapConfigFlow(test *framework.ClusterE2ETest) {
+	test.GenerateClusterConfig()
+	test.DownloadArtifacts()
+	test.ExtractDownloadedArtifacts()
+	test.DownloadImages()
```
d8660091, line 30: We need to add this DownloadImages to runVSphereAirgapConfigFlow

test/e2e/airgap.go
abhay-krishna: Ahh this is just for E2E tests, this simulates the situation where the admin machine might be initially open to the internet for importing artifacts but airgapped after that.

test/e2e/airgap.go
```diff
@@ -22,3 +23,15 @@ func runVSphereAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs strin
...
+	test.ImportImages()
```
d8660091, line 32: Since we now separate the downloading and Importing images, I think we should run ImportImages() after airgapping

test/framework/cluster.go
```diff
@@ -584,29 +587,49 @@ func (e *ClusterE2ETest) WithClusterConfig(fillers ...api.ClusterConfigFiller) *
...
+	} else {
+		e.T.Logf("Downloaded images archive saved at %s", defaultDownloadImagesOutputLocation)
 	}
+}
+
+func (e *ClusterE2ETest) ImportImages(opts ...CommandOpt) {
+	clusterConfig := e.ClusterConfig.Cluster
+	registyMirrorEndpoint, registryMirrorPort := clusterConfig.Spec.RegistryMirrorConfiguration.Endpoint, clusterConfig.Spec.RegistryMirrorConfiguration.Port
+	registryMirrorHost := net.JoinHostPort(registyMirrorEndpoint, registryMirrorPort)
+	importImagesArgs := []string{"import images", "--input", defaultDownloadImagesOutputLocation, "--bundles", defaultBundleReleaseManifestFile, "--registry", registryMirrorHost, "--insecure"}
```
d8660091, line 626: import images might fail in airgapped environment if defaultBundleReleaseManifestFile is used, since it may contain url to remote yaml files

abhay-krishna: yeah I changed this logic to use the one from extracted artifacts download.

test/framework/cluster.go
```diff
@@ -583,29 +586,59 @@ func (e *ClusterE2ETest) WithClusterConfig(fillers ...api.ClusterConfigFiller) *
...
-	if _, err := os.Stat("eks-anywhere-downloads.tar.gz"); err != nil {
+	e.T.Log("Extracting downloaded artifacts")
+	e.Run("tar", "-xf", defaultDownloadArtifactsOutputLocation)
+}
+
+// DownloadImages runs the EKS-A `download images` command with appropriate args.
+func (e *ClusterE2ETest) DownloadImages(opts ...CommandOpt) {
+	downloadImagesArgs := []string{"download", "images", "-o", defaultDownloadImagesOutputLocation}
+	if getBundlesOverride() == "true" {
+		downloadImagesArgs = append(downloadImagesArgs, "--bundles-override", defaultBundleReleaseManifestFile)
```
d8660091, line 613: download images should take into account of the bundle released downloaded from "download artifact" command as well. There is opportunity to reuse the code.

abhay-krishna: This is currently reusing the bundles override code that the create/upgrade/delete workflows are using for tests. It's going to be the same bundle in either case

d8660091: It's not "necessary" to change if either case the same, but, they are not the same (either due to some bug or by design). Manifesturl in artifact bundlerelease.yaml points to a local file, while the bundle-release downloaded through curl points to a remote URL.

abhay-krishna: Oh I just meant, as far as images goes it should ideally preserve the same URIs right? only the S3 URI will be replaced by local filepaths

abhay-krishna: I am not opposed to changing if it makes it more proper, just an observation

abhay-krishna: Done!

test/e2e/airgap.go
```diff
@@ -4,15 +4,18 @@
...
 )
 
 // runVSphereAirgapConfigFlow run airgap deployment but allow bootstrap cluster to access local peers.
 func runVSphereAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs string) {
 	test.GenerateClusterConfig()
 	test.DownloadArtifacts()
-	test.ImportImages()
 	test.ExtractDownloadedArtifacts()
+	test.DownloadImages()
+	test.ImportImages()
```
d8660091, line 18: nit: move ImportImages() before CreateCluster

test/framework/cluster.go
```diff
@@ -583,29 +586,65 @@ func (e *ClusterE2ETest) WithClusterConfig(fillers ...api.ClusterConfigFiller) *
...
+	e.Run("tar", "-xf", defaultDownloadArtifactsOutputLocation)
+}
+
+// DownloadImages runs the EKS-A `download images` command with appropriate args.
+func (e *ClusterE2ETest) DownloadImages(opts ...CommandOpt) {
+	downloadImagesArgs := []string{"download", "images", "-o", defaultDownloadImagesOutputLocation}
+	if getBundlesOverride() == "true" {
+		var bundleManifestLocation string
+		if _, err := os.Stat(defaultDownloadArtifactsOutputLocation); err == nil {
+			bundleManifestLocation = "eks-anywhere-downloads/bundle-release.yaml"
```
d8660091, line 615: the 2 lines should execute if getBundlesOverride() == "false"

abhay-krishna: `getBundlesOverride()` method just reads the T_BUNDLES_OVERRIDE env var, but we are providing a choice of which local bundle manifest to use as override. For air-gapped, we will use the one from download-artifacts, but otherwise the one from `bin/local-bundle-release.yaml`

d8660091: fair enough, I'm not too much concernd about this, let's see how it goes

PR #4908, opened 2023-02-13
Rufio machines validations FLC

pokearu:
*Issue https://github.com/aws/eks-anywhere-internal/issues/1054:*

*Description of changes:*
Adding validations to check baseboard management for hardware are contactable in the controller reconcile flow.

*Testing:*
`make docker-build` and tested the controller.


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -256,3 +258,44 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, clus
...
+	etcdReader := hardware.NewETCDReader(r.client)
+	if err := etcdReader.RufioMachinesFromEtcd(ctx); err != nil {
+		log.Error(err, "rufio machine check failure")
+		failureMessage := err.Error()
+		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
+
+		return controller.ResultWithReturn(), nil
+	}
+
+	for _, rm := range etcdReader.GetCatalogue().AllBMCs() {
```
pokearu, line 275: Currently this checks all the BM objects on the cluster each time since thats how the main CLI flow does it.
I am open to refactor this to only check the current clusters BM.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.16.14
 	github.com/aws/aws-sdk-go-v2/config v1.15.3
 	github.com/aws/aws-sdk-go-v2/credentials v1.11.2
+	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.3
```
chrisdoherty4, line 11: Was this accidental?

pkg/providers/tinkerbell/hardware/etcd.go
```diff
@@ -82,3 +78,19 @@ func (er *ETCDReader) getUnprovisionedTinkerbellHardware(ctx context.Context) ([
 
 	return selectedHardware.Items, nil
 }
+
+// RufioMachinesFromEtcd fetches rufio machine objects from the cluster and inserts into catalogue.
+func (er *ETCDReader) RufioMachinesFromEtcd(ctx context.Context) error {
```
chrisdoherty4, line 83: Do we need to append `FromEtcd` to the methods? 

Side note: this reads from Kube API Server as opposed to etcd. We should consider renaming to something a little more appropriate.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -256,3 +259,44 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, clus
 
 	return controller.Result{}, nil
 }
+
+// CheckRufioMachinesContactable checks to ensure all the Rufio machines condition contactable is True.
+func (r *Reconciler) CheckRufioMachinesContactable(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
```
chrisdoherty4, line 264: What's the rationale for deviating from `Validate`?

pkg/providers/tinkerbell/hardware/etcd.go
```diff
@@ -36,17 +36,13 @@ func NewETCDReader(client client.Client) *ETCDReader {
...
+// HardwareFromETCD fetches the unprovisioned tinkerbell hardware objects and inserts in to ETCDReader catalogue.
+func (er *ETCDReader) HardwareFromETCD(ctx context.Context) error {
 	hwList, err := er.getUnprovisionedTinkerbellHardware(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to build catalogue: %v", err)
 	}
 
-	if len(hwList) == 0 {
-		return errors.New("no available hardware")
-	}
```
abhinavmpandey08, line 48: Why is this not needed?

pkg/providers/tinkerbell/hardware/etcd.go
```diff
@@ -82,3 +78,19 @@ func (er *ETCDReader) getUnprovisionedTinkerbellHardware(ctx context.Context) ([
...
+func (er *ETCDReader) RufioMachinesFromEtcd(ctx context.Context) error {
+	var rufioMachines rufiov1alpha1.MachineList
+	if err := er.client.List(ctx, &rufioMachines, &client.ListOptions{Namespace: constants.EksaSystemNamespace}); err != nil {
+		return fmt.Errorf("listing rufio machines: %v", err)
+	}
+
+	for i := range rufioMachines.Items {
+		if err := er.catalogue.InsertBMC(&rufioMachines.Items[i]); err != nil {
+			return err
+		}
```
abhinavmpandey08, line 92: Does this work?
```suggestion
	for _,  rufioMachine := range rufioMachines.Items {
		if err := er.catalogue.InsertBMC(&rufioMachine); err != nil {
			return err
		}
```

abhinavmpandey08 (commented):
Do we have already have validations to check if the hardware's BMC ref exists? And BMC's Secret Ref?

go.mod
pokearu: It came up from a go mod tidy run. But looks like it got merged in now.

pkg/providers/tinkerbell/hardware/etcd.go
pokearu: I tested without it, so we run assertion `MinimumHardwareAvailableAssertionForCreate`. This also covers the case where there are no hardware available, so the explicit check became redundant.

pokearu:
> Do we have already have validations to check if the hardware's BMC ref exists? And BMC's Secret Ref?

So BMCs are still optional. But we have static validations for those refs being correctly mapped when we run hardware generate.

pkg/providers/tinkerbell/reconciler/reconciler.go
pokearu: Um I dont have a strong preference here. My rational was we have `CheckControlPlaneReady` which checks on a conditions and hence since we do the same here I named it `Check`.

I can make it validate 👍

pkg/providers/tinkerbell/hardware/etcd.go
pokearu: As discussed I have renamed the struct and the methods 👍

pkg/providers/tinkerbell/hardware/etcd.go
pokearu: Actually this caused an issue. Since we are using pointers InsertBMC inserts the same rufioMachine pointer each time. Thereby each object inserted was always the last one in the list.

abhinavmpandey08: oh wait that's how it works? So `InsertBMC()` takes a pointer to the BMC and not the BMC object itself?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -256,3 +259,44 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, clus
...
+	if err := kubeReader.LoadRufioMachines(ctx); err != nil {
+		log.Error(err, "rufio machine check failure")
+		failureMessage := err.Error()
+		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
+
+		return controller.ResultWithReturn(), nil
+	}
+
+	for _, rm := range kubeReader.GetCatalogue().AllBMCs() {
+		if err := r.checkContactable(rm); err != nil {
```
abhinavmpandey08, line 277: So is this validating that all the rufio machines are contactable before reconciling the cluster? Or only the new rufio machines that are added?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -256,3 +259,44 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, clus
...
 	return controller.Result{}, nil
 }
+
+// ValidateRufioMachines checks to ensure all the Rufio machines condition contactable is True.
+func (r *Reconciler) ValidateRufioMachines(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "validateRufioMachines")
+
+	kubeReader := hardware.NewKubeReader(r.client)
+	if err := kubeReader.LoadRufioMachines(ctx); err != nil {
+		log.Error(err, "rufio machine check failure")
```
panktishah26, line 269: I believe `LoadRufioMachines` loads the existing rufio machines to the kubeReader object. Can we change the error log here, something like?
```suggestion
		log.Error(err, "loading existing rufio machines from the cluster")
```

pkg/providers/tinkerbell/hardware/etcd.go
pokearu: Yah it takes a pointer. It was hard to notice but when I tested it caused issues and I needed to change it back.

pkg/providers/tinkerbell/reconciler/reconciler.go
pokearu: Actually it checks all. Thats what i wanted to discuss here.

https://github.com/aws/eks-anywhere/pull/4908#discussion_r1104126052

abhinavmpandey08 (approved):
/lgtm
/woof

PR #4930, opened 2023-02-13
Attempt at fix for dev helm chart in dev bundles

jonahjon:

*Description of changes:*

This will do 3 things which should help with package controller image tag/sha mismatch from the bundle and the helm chart that is causing installation/e2e issues.

1. Will override the Image SHA in the values.yaml for the packages helm chart on dev bundles. This is the function `OverWriteChartValuesImageSha` on line 447 of `release/pkg/helm/helm.go`

2. Will lookup and use the correct latest version/sha of the Token Refresher in the Dev Bundle when running a dev Release. This is on lines 58-61, and 84-89 on  `release/pkg/bundles/package-controller.go`

3. Will override the Image Version and SHA in the values.yaml for the packages helm chart on dev bundles, and push the helm chart to Public ECR that will be used in the dev bundle. We are re-using logic we use to fix the Helm Chart in production by overwriting tags here. This can be found on lines 106-136 on `release/pkg/bundles/package-controller.go`

What this results in is the Helm Chart being referenced in the Dev Bundle getting it's `values.yaml` image tag/sha referenced being forced to the exact version referenced in the Dev Bundle which **Should** :crossed_fingers:  help fix the issues we've been seeing where installations fail due to image not being found.

### DIFF
--- a/release/pkg/bundles/package-controller.go
+++ b/release/pkg/bundles/package-controller.go
@@ -23,6 +23,7 @@ import (
 	anywherev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 	"github.com/aws/eks-anywhere/release/pkg/aws/ecr"
 	"github.com/aws/eks-anywhere/release/pkg/constants"
+	"github.com/aws/eks-anywhere/release/pkg/helm"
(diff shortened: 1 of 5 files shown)


abhay-krishna (approved):
/approve

jonahjon:
I hope this works

/woof

PR #4931, opened 2023-02-13
Strip nulls from hardware yaml when generated from CSV

cxbrowne1207:
*Issue #, if available:*
#1036

*Description of changes:*
This PR strips all null fields from the hardware.yaml produced by the `eksctl anywhere generate -z hardware.csv` command.

Currently, sing the *eksctl anywhere generate hardware -z hardware.csv* command, generates a yaml that has null for fields `.metadata.creationTimestamp` and `.spec.bmfRef`. To enable management of hardware through GitOps, these fields must be removed from the generated hardware.yaml because flux throws an `invalid value error.: body must be of type string` in this case.

*Testing (if applicable):*
Unit testing and running the `eksctl anywhere generate` command 


pkg/providers/tinkerbell/hardware/csv.go
```diff
@@ -87,3 +90,24 @@ func ensureRequiredColumnsInCSV(unmatched []string) error {
 
 	return nil
 }
+
+// BuildHardwareYaml builds a hardware yaml from the csv at the provided path.
+func BuildHardwareYaml(path string) ([]byte, error) {
```
chrisdoherty4, line 95:
```suggestion
// BuildHardwareYAML builds a hardware yaml from the csv at the provided path.
func BuildHardwareYAML(path string) ([]byte, error) {
```

pkg/providers/tinkerbell/hardware/csv.go
```diff
@@ -87,3 +90,24 @@ func ensureRequiredColumnsInCSV(unmatched []string) error {
...
+	reader, err := NewNormalizedCSVReaderFromFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading csv: %v", err)
+	}
+
+	var b bytes.Buffer
+	bufferedWriter := io.Writer(&b)
+	writer := NewTinkerbellManifestYAML(bufferedWriter)
```
chrisdoherty4, line 103:
```suggestion
	var b bytes.Buffer
	writer := NewTinkerbellManifestYAML(&b)
```

pkg/utils/yaml/yaml.go
```diff
@@ -25,3 +28,31 @@ func Serialize[T any](objs ...T) ([][]byte, error) {
...
+		v := val.MapIndex(key)
+		if v.IsNil() {
+			delete(m, key.String())
+			continue
+		}
+		switch t := v.Interface().(type) {
+		// If key is a JSON object (Go Map), use recursion to go deeper
+		case map[string]interface{}:
+			stripNullFromMap(t)
+		}
```
chrisdoherty4, line 56:
```suggestion
		// If key is a JSON object (Go Map), use recursion to go deeper
		if t, ok := v.Interface().(map[string]interface{}); ok {
			stripNullFromMap(t)
		}
```

pkg/utils/yaml/yaml.go
```diff
@@ -25,3 +28,31 @@ func Serialize[T any](objs ...T) ([][]byte, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("converting yaml to unstructured: %v", err)
+	}
+	for _, u := range uList {
+		stripNullFromMap(u.Object)
+	}
+	return unstructuredutil.UnstructuredToYaml(uList)
+}
+
+func stripNullFromMap(m map[string]interface{}) {
```
chrisdoherty4, line 44:
```suggestion
func stripNull(m map[string]interface{}) {
```

pkg/utils/yaml/yaml.go
```diff
@@ -25,3 +28,31 @@ func Serialize[T any](objs ...T) ([][]byte, error) {
 	}
 	return r, nil
 }
+
+// StripNull removes all null fields from the provided yaml.
+func StripNull(resources []byte) ([]byte, error) {
+	uList, err := unstructuredutil.YamlToUnstructured(resources)
```
chrisdoherty4, line 34: The underlying implementations of `Unstructured` expect to be working with Kubernetes objects. This package should contain YAML generic functionality only so I don't think its appropriate for here.

We can either remove the dependency on unstructured; I don't think that would be too difficult. Alternatively, move to a more appropriate location such as the `unstructured` util package.

chrisdoherty4 (changes requested):
The changes are OK but it does indicate a broader problem upstream that we should focus on fixing as opposed to only patching in our code base. 

I suspect the upstream `Hardware` YAML has some Go tags that cause it to render the `createTimestamp` etc fields. Would you be able to investigate that?

pkg/providers/tinkerbell/hardware/csv.go
cxbrowne1207: From what I understand from investigating, the `CreationTimestamp` in the `metav1.ObjectMeta` is optional and it's tagged with `omitempty`, but because it isn't a pointer it will never satisfy the empty check. There's an issue open but it seems to be difficult to do:

https://github.com/kubernetes/kubernetes/issues/67610

As for apiGroup, there is no `omitempty` tag on it [here](https://pkg.go.dev/k8s.io/api/core/v1#TypedLocalObjectReference), and I don't think we use it

vivek-koppuru: Yea that's why we have a separate ObjectMeta object here: https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/object_meta.go#L8

Are we saying we don't want to use this instead to simply get rid of the field as this is a patch approach? Since we use this for others it feels like we are doing a lot of extra work for hardware, but then again it could be the right way. Either way, we probably want to create an issue to be in sync with the others.

vivek-koppuru (approved):

pkg/providers/tinkerbell/hardware/csv.go
chrisdoherty4: I think this is fine for now but marshalling into JSON shouldn't result in nul fields output and I think that's probably something that needs addressing upstream.

chrisdoherty4 (approved):

PR #4935, opened 2023-02-13
Wait for eksa-packages-[clusterName] namespace

vincentni:
wait for a bundle to be active as well as eksa-packages-[clusterName] namespace to be created


lewisdiamond:
There are issues with this implementation. Using `readyCnt` has two potential issue, if the ActiveBundle isn't "" but the namespace isn't created, the count will be incremented again after and would return without the namespace existing. If the 2nd run both the namespace and the ActiveBundle is valid, count becomes 3 and the function will continue waiting and incrementing until it timesout.

PR #4943, opened 2023-02-14
Cert manager e2e

a-cool-train:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


test/e2e/vsphere_test.go
```diff
@@ -373,6 +373,46 @@ func TestVSphereKubernetes121BottleRocketWorkloadClusterCuratedPackagesSimpleFlo
 	runCuratedPackageRemoteClusterInstallSimpleFlow(test)
 }
 
+func TestVSphereKubernetes121BottleRocketWorkloadClusterCuratedPackagesCertManagerSimpleFlow(t *testing.T) {
```
ivyostosh, line 376: Are you going to add Ubuntu tests in a separate PR? If not, try to do a mix of BottleRocket and Ubuntu tests here for better coverage.

test/framework/cluster.go
```diff
@@ -1446,6 +1446,151 @@ func (e *ClusterE2ETest) VerifyPrometheusPackageInstalled(packageName string, ta
...
+	if err != nil {
+		return fmt.Errorf("error creating cert manager let's encrypt issuer: %v", err)
+	}
+
+	err = e.KubectlClient.ApplyKubeSpecFromBytes(ctx, e.Cluster(), certManagerLetsEncryptCert)
+	if err != nil {
+		return fmt.Errorf("error creating cert manager let's encrypt issuer: %v", err)
+	}
+
+	err = e.KubectlClient.WaitJSONPathLoop(ctx, e.Cluster().KubeconfigFile, "5m", "status.conditions[0].status", "True",
```
lewisdiamond, line 1585: This will succeed even if the certificate is in issuing state.

test/e2e/vsphere_test.go
a-cool-train: I think a single OS and a single provider should suffice Cert manager since cert manager doesn't have any dependency on OS or provider.

test/framework/cluster.go
a-cool-train: That has not been my experience. I have generally seen that this status is False when the certificate is not ready. Do you know a better way of checking than checking the status conditions.

lewisdiamond: https://github.com/cert-manager/cert-manager/blob/b5d1b1e27467e7fd9be4124f0041ec58524105c4/pkg/apis/certmanager/v1/types_certificate.go#L457
You must check that the type of the condition is `"Ready"` as well as status being `True`. It can also be in `"Issuing"` and status `True`, for example: https://github.com/cert-manager/cert-manager/blob/master/pkg/controller/certificates/trigger/trigger_controller.go#L204

test/framework/cluster.go
```diff
@@ -1446,6 +1446,149 @@ func (e *ClusterE2ETest) VerifyPrometheusPackageInstalled(packageName string, ta
...
+		mgmtCluster, prefix+"-"+packageName, "5m", ns)
+	if err != nil {
+		e.T.Fatalf("waiting for cert-manager package timed out: %s", err)
+	}
+
+	e.T.Log("Waiting for Package", packageName, "Deployment to be healthy")
+
+	for _, name := range deployments {
+		go func(name string) {
+			wg.Add(1)
```
lewisdiamond, line 1473: `wg.Add(1)` must be before the goroutine.

lewisdiamond, line 1473:
```suggestion
	for _, name := range deployments {
		wg.Add(1)
		go func(name string) {
```

test/framework/cluster.go
```diff
@@ -1446,6 +1446,151 @@ func (e *ClusterE2ETest) VerifyPrometheusPackageInstalled(packageName string, ta
...
+
+	e.T.Log("Waiting for Let's Encrypt certificate to be issued")
+	err = e.verifyLetsEncryptCert(mgmtCluster)
+	if err != nil {
+		errCh <- err
+	}
+
+	go func() {
+		wg.Wait()
+		okCh <- "completed"
```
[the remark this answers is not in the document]
a-cool-train: I think 2 channels makes it easily readable.

PR #4951, opened 2023-02-14
Adding function for doubletagged images

jonahjon:

*Description of changes:*

Sometimes when the image details gets checked it will multiple tags.

`["0.0.0.abcde", "latest"]`

This would technically pass the filter requirement where we are looking for the presence of the of the prefix `0.0.0` however it might return back the tag of latest for the actual bundle.

This function will check the slice once a match has been made, and remove out an items from the slice which also don't match the prefix requirement.

var newSlice []string
newSlice = removeStringSlice(["0.0.0.abcde", "latest"] "latest")

`["0.0.0.abcde", "latest"]`  becomes.... `["0.0.0.abcde"]`

### DIFF
--- a/release/pkg/aws/ecr/ecr.go
+++ b/release/pkg/aws/ecr/ecr.go
@@ -119,6 +119,18 @@ func FilterECRRepoByTagPrefix(ecrClient *ecr.ECR, repoName, prefix string, hasTa
 	} else {
 		filteredImageDetails = imageTagFilterWithout(imageDetails, prefix)
 	}
+
+	// Filter out any tags that don't match our prefix for doubletagged scenarios
+	for _, detail := range filteredImageDetails {
+		if len(detail.ImageTags) > 0 {
+			for j, tag := range detail.ImageTags {
+				if !strings.HasPrefix(*tag, prefix) {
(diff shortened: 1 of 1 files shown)


release/pkg/aws/ecr/ecr.go
```diff
@@ -119,6 +119,18 @@ func FilterECRRepoByTagPrefix(ecrClient *ecr.ECR, repoName, prefix string, hasTa
 	} else {
 		filteredImageDetails = imageTagFilterWithout(imageDetails, prefix)
 	}
+
+	// Filter out any tags that don't match our prefix for doubletagged scenarios
+	for _, detail := range filteredImageDetails {
+		if len(detail.ImageTags) > 0 {
```
TerryHowe, line 125: Doesn't range work on an empty slice?

TerryHowe (approved):
/lgtm

release/pkg/aws/ecr/ecr.go
jonahjon: Are you asking if this would fail on an empty slice?

jonahjon: O I see what your'e saying it would just do 0 iterations and go to the next step.

PR #4970, opened 2023-02-14
curated packages full cluster lifecycle adaptations

ewollesen:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


ewollesen:
Related to https://github.com/aws/eks-anywhere-packages/issues/807

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -72,7 +72,7 @@ func installPackageController(ctx context.Context) error {
 	}
 
 	curatedpackages.PrintLicense()
-	err = ctrlClient.EnableCuratedPackages(ctx)
+	err = ctrlClient.Enable(ctx)
```
a-cool-train, line 75: Any reason for the name change?

controllers/cluster_controller.go
```diff
@@ -61,12 +68,13 @@ type ClusterValidator interface {
 }
 
 // NewClusterReconciler constructs a new ClusterReconciler.
-func NewClusterReconciler(client client.Client, registry ProviderClusterReconcilerRegistry, awsIamAuth AWSIamConfigReconciler, clusterValidator ClusterValidator) *ClusterReconciler {
+func NewClusterReconciler(client client.Client, registry ProviderClusterReconcilerRegistry, awsIamAuth AWSIamConfigReconciler, clusterValidator ClusterValidator, pkgs PackagesClient) *ClusterReconciler {
```
a-cool-train, line 71: Is the packagesclient an optional item or will this be required for the lifecycle management?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
ewollesen: The old name was super long, and it didn't feel useful to have packagesControllerClient.EnableCuratedPackages, as per

> For instance, the buffered reader type in the bufio package is called Reader, not BufReader, because users see it as bufio.Reader, which is a clear, concise name. 

https://go.dev/doc/effective_go#names

controllers/cluster_controller.go
ewollesen: It's required.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+			return controller.Result{}, err
+		}
+	}
+
 	return controller.Result{}, nil
 }
 
+// PackagesClient handles curated packages operations from within the cluster
+// controller.
+type PackagesClient interface {
```
a-cool-train, line 284: Any reason this is defined here? It might make more sense to define it somewhere up there or where it is getting used.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	EnableFullLifecycle(context.Context, logr.Logger, string, string, *v1alpha1.Image, *registrymirror.RegistryMirror, ...curatedpackages.PackageControllerClientOpt) error
```
a-cool-train, line 285: Should we opt for defining the variables that will be consumed by this function? That will make the readibility easier.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+}
+
+// getBundleFromCluster based on the cluster's k8s version.
+func (r *ClusterReconciler) getBundleFromCluster(ctx context.Context, cluster *anywherev1.Cluster) (*v1alpha1.Image, error) {
+	bundles := &v1alpha1.Bundles{}
+	nn := types.NamespacedName{
+		Name:      cluster.Spec.BundlesRef.Name,
+		Namespace: cluster.Spec.BundlesRef.Namespace,
+	}
+	if err := r.client.Get(ctx, nn, bundles); err != nil {
```
a-cool-train, line 323: Doesn't the Kube client support jsonpath? That way you wouldn't need to have two separate GET, and FIND but instead have a single GET with jsonpath.

a-cool-train, line 323: Also should we create a separate file for all kubernetes go client execution?

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	if verBundle == nil {
+		return nil, fmt.Errorf("no bundle for kube version %q", kubeVersion)
+	}
+
+	return verBundle, nil
+}
+
+// writeKubeConfigSecret so that the helm executable can pick it up to install
+// curated packages helm charts.
+func (r *ClusterReconciler) writeKubeConfigSecret(ctx context.Context, cluster *anywherev1.Cluster) (kubeconfigFile, error) {
```
a-cool-train, line 352: Why do we need the kubeconfig secret for the workload cluster since all the package lifecycle goes through the management cluster? Should we potentially create the kubeconfig secret for management cluster?

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	kubeConfigSecret := &corev1.Secret{}
+	kubeConfigNN := types.NamespacedName{
+		Namespace: constants.EksaSystemNamespace,
+		Name:      cluster.Name + "-kubeconfig",
+	}
+	if err := r.client.Get(ctx, kubeConfigNN, kubeConfigSecret); err != nil {
```
a-cool-train, line 358: I would recommend moving kube client stuff to a different file for ease of read and structure.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+		return nil, fmt.Errorf("getting kubeconfig secret: %w", err)
+	}
+	secretBytes, err := yaml.Marshal(kubeConfigSecret)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling secret %w", err)
+	}
+	f, err := os.CreateTemp("", "kubeconfig-*.yaml")
```
a-cool-train, line 365: Why is this necessary?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +57,79 @@ type PackageControllerClient struct {
...
+	Delete(ctx context.Context, kubeconfigFilePath, installName, namespace string) error
+}
+
+// NewPackageControllerClientFullLifecycle creates a PackageControllerClient
+// for the Full Cluster Lifecycle controller.
+//
+// It differs because the CLI use case has far more information available at
+// instantiation, while the FCL use case has less information at
+// instantiation, and the rest when cluster creation is triggered.
+func NewPackageControllerClientFullLifecycle(logger logr.Logger, chartInstaller ChartInstaller, deleter ChartInstallationDeleter, kubectl KubectlRunner) *PackageControllerClient {
```
a-cool-train, line 83: This feels like it warrants its own file.

a-cool-train:
I would recommend having someone from the CLI also reviewing this PR.

controllers/cluster_controller.go
ewollesen: Just let me know where you'd like it to be, and I'll move it there. It's used in several places, and "up there" is rather vague, but if you can give me a line number in this file, I'll try to get it as close as I can.

controllers/cluster_controller.go
ewollesen: > Should we opt for defining the variables that will be consumed by this function? That will make the readibility easier.

Please see line 99 of this file for that information. I didn't want to have duplicates because then if I change the name in one place, I'll undoubtedly forget to change the name in the other, then I have an even bigger readability problem on my hands.

I generally only have names on the implementations, not the interfaces.

controllers/cluster_controller.go
ewollesen: That's news to me. Do you have a link to any docs for me? I've not run across that before.

ewollesen: > Also should we create a separate file for all kubernetes go client execution?

I don't see how there's any problem with what we have now...?

controllers/cluster_controller.go
ewollesen: > Why do we need the kubeconfig secret for the workload cluster since all the package lifecycle goes through the management cluster? Should we potentially create the kubeconfig secret for management cluster?

Fair question. Let me do a little digging. I think we might not need any kubeconfig written at all, which would be even better.

controllers/cluster_controller.go
ewollesen: > I would recommend moving kube client stuff to a different file for ease of read and structure.

I don't really know what you mean. How is it easier to read if it lives in another file? What exactly would you move there? The client was already in this controller, it's not like I added it, I'm just using the existing resources.

controllers/cluster_controller.go
ewollesen: > Why is this necessary?

Not sure what you're asking, but I'll explain what's going on...

This is pretty standard code for marshaling (or serializing if you prefer) a Go struct to YAML text, and then writing it to a file on disk. The Go struct in question is kubeconfig data that I need to write to the disk so that the helm executable can read it.

I don't know of a better/simpler way to get that data to the helm executable.

However, as per [one of your earlier comments](https://github.com/aws/eks-anywhere/pull/4970#discussion_r1123495122), this code might be removed soon. I'll be testing that possibility ASAP.

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Why?

I like having it close to the NewPackageControllerClient, as it makes it easier to compare the two functions and see how they're similar or different. Plus they're both creating the same object.

ewollesen:
> I would recommend having someone from the CLI also reviewing this PR.

No recommendations?

I've posted the PR for everyone, no one from the CLI side has stepped up (yet).

controllers/cluster_controller.go
ewollesen: Monday morning followup. Wow. SUPER-MEGA-ULTRA-Brain fart on my part. I see what you're asking now. 

I must not have had my head on right when I wrote this.

I'm guessing you were confused, because I took what was already just YAML text, not a Go struct, and double serialized it then wrote it to disk, and how could that possibly work?

On the bright side, this means I can almost 99% certain just get rid of this.

However, the helm executable interface still needs a file to pass on the command line. So I'll re-work the ChartInstaller implementation to skip passing the kubeconfig flag if the filename is the empty string.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
 }
 
+// PackagesClient handles curated packages operations from within the cluster
+// controller.
+type PackagesClient interface {
+	EnableFullLifecycle(context.Context, logr.Logger, string, string, *v1alpha1.Image, *registrymirror.RegistryMirror, ...curatedpackages.PackageControllerClientOpt) error
+	ReconcileDelete(context.Context, logr.Logger, client.Client, *anywherev1.Cluster) error
+}
+
+func (r *ClusterReconciler) postReconcilePackagesForWorkloadCluster(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (err error) {
```
g-gaston, line 289: could we move this to a new entity in the curatedpackages package? Something like `curatedpackages.Reconciler`.
With probably two methods, `Reconcile` and `ReconcileDelete`?

That would fit better the idea of having this controller as an orchestrator that calls other services for the specific reconcile logic at specific points in time.

g-gaston, line 289: Could we have a guard at the beggining of this method that returns early if "full lifecycle" is already enabled?

That allows to avoid writing the kubeconfig to disk and all that jazz if we already "enabled" curated packages for that cluster.

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
 		}
 	}
 
+	if cluster.IsPackagesEnabled() {
+		if err := r.postReconcilePackagesForWorkloadCluster(ctx, log, cluster); err != nil {
```
g-gaston, line 274: does this need to happen for both management and workload cluster?
if it's only for workload then I would do
```suggestion
	if cluster.IsPackagesEnabled() && cluster.IsManaged() {
		if err := r.postReconcilePackagesForWorkloadCluster(ctx, log, cluster); err != nil {
```

controllers/cluster_controller.go
```diff
@@ -262,9 +270,135 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+
+	verBundle, err := r.findMatchingBundle(bundles, string(cluster.Spec.KubernetesVersion))
+	if err != nil {
+		return nil, err
+	}
+
+	return &verBundle.PackageController.HelmChart, nil
+}
+
+func (r *ClusterReconciler) findMatchingBundle(bundles *v1alpha1.Bundles, kubeVersion string) (*v1alpha1.VersionsBundle, error) {
```
g-gaston, line 335: nit: I think there is helper method in the `cluster` package for this
I think `GetVersionsBundle`

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -70,6 +70,12 @@ func (c *Cluster) HasAWSIamConfig() bool {
 	return false
 }
 
+// IsPackagesEnabled checks if the configuration supports curated packages
+// installation.
+func (c *Cluster) IsPackagesEnabled() bool {
+	return !c.IsSelfManaged() && (c.Spec.Packages == nil || !c.Spec.Packages.Disable)
```
g-gaston, line 76: Does this mean that packages can't be enabled in management clusters?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +57,79 @@ type PackageControllerClient struct {
...
+// information is known at PackageControllerClient initialization, and the
+// Full Cluster Lifecycle use case, where there's limited information at
+// initialization.
+func (pc *PackageControllerClient) EnableFullLifecycle(ctx context.Context, log logr.Logger, clusterName, kubeConfig string, chart *releasev1.Image, registryMirror *registrymirror.RegistryMirror, options ...PackageControllerClientOpt) (err error) {
+	log.V(6).Info("enabling curated package full lifecycle", "clusterName", clusterName)
+	defer func(err *error) {
+		if err != nil && *err != nil {
+			log.V(6).Error(*err, "enabling curated packages full lifecycle", "clusterName", clusterName)
+		} else {
+			log.V(6).Info("success enabling curated packages full lifecycle")
```
g-gaston, line 105:
```suggestion
			log.V(6).Info("Success enabling curated packages full lifecycle")
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +57,79 @@ type PackageControllerClient struct {
...
+//
+// This method fills in the gaps between the original CLI use case, where all
+// information is known at PackageControllerClient initialization, and the
+// Full Cluster Lifecycle use case, where there's limited information at
+// initialization.
+func (pc *PackageControllerClient) EnableFullLifecycle(ctx context.Context, log logr.Logger, clusterName, kubeConfig string, chart *releasev1.Image, registryMirror *registrymirror.RegistryMirror, options ...PackageControllerClientOpt) (err error) {
+	log.V(6).Info("enabling curated package full lifecycle", "clusterName", clusterName)
+	defer func(err *error) {
+		if err != nil && *err != nil {
+			log.V(6).Error(*err, "enabling curated packages full lifecycle", "clusterName", clusterName)
```
g-gaston, line 103:
```suggestion
			log.V(6).Error(*err, "Enabling curated packages full lifecycle", "clusterName", clusterName)
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +57,79 @@ type PackageControllerClient struct {
...
+}
+
+// EnableFullLifecycle wraps Enable to handle run-time arguments.
+//
+// This method fills in the gaps between the original CLI use case, where all
+// information is known at PackageControllerClient initialization, and the
+// Full Cluster Lifecycle use case, where there's limited information at
+// initialization.
+func (pc *PackageControllerClient) EnableFullLifecycle(ctx context.Context, log logr.Logger, clusterName, kubeConfig string, chart *releasev1.Image, registryMirror *registrymirror.RegistryMirror, options ...PackageControllerClientOpt) (err error) {
+	log.V(6).Info("enabling curated package full lifecycle", "clusterName", clusterName)
```
g-gaston, line 100: You don't need the `clusterName` in the logs
`controller-runtime` injects that (and the kind and namespace) and log is passed down from the controller.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +57,79 @@ type PackageControllerClient struct {
...
+			log.V(6).Error(*err, "enabling curated packages full lifecycle", "clusterName", clusterName)
+		} else {
+			log.V(6).Info("success enabling curated packages full lifecycle")
+		}
+	}(&err)
+	pc.skipWaitForPackageBundle = true
+	pc.clusterName = clusterName
+	pc.kubeConfig = kubeConfig
+	pc.chart = chart
+	pc.registryMirror = registryMirror
```
g-gaston, line 112: This is not thread safe. If we decide to run our controller with more than one routine (I think right now it defaults to 1), this can become problematic. That's likely future, I wouldn't ask otherwise.

Is there any way we can avoid it?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -352,6 +425,24 @@ func (pc *PackageControllerClient) GetPackageControllerConfiguration() (result s
 	return result, err
 }
 
+// ReconcileDelete removes resources after a full cluster lifecycle cluster is deleted.
+func (pc *PackageControllerClient) ReconcileDelete(ctx context.Context, logger logr.Logger, client client.Client, cluster *anywherev1.Cluster) error {
+	namespace := "eksa-packages-" + cluster.Name
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	if err := client.Delete(ctx, ns); err != nil {
```
g-gaston, line 432: I don't think delete is idempotent so this will fail if called twice (it can happen for a reentry reconciliation for a deleted cluster, like if a transient error happens).

I think you can make it idempotent but just ignoring with `apierrors.IsNotFound`

pkg/executables/helm.go
```diff
@@ -21,6 +21,8 @@ type Helm struct {
 	registryMirror *registrymirror.RegistryMirror
 	env            map[string]string
 	insecure       bool
+	// skipCRDs passes the --skip-crds flag to the helm executable.
+	skipCRDs bool
```
g-gaston, line 25: what does this do?
By building the helm client with this, all consumers will use this flag (there is only one instance of the helm client per factory)

pkg/executables/helm.go
```diff
@@ -126,7 +138,10 @@ func (h *Helm) InstallChartFromName(ctx context.Context, ociURI, kubeConfig, nam
 // InstallChart installs a helm chart to the target cluster.
 func (h *Helm) InstallChart(ctx context.Context, chart, ociURI, version, kubeconfigFilePath, namespace, valueFilePath string, values []string) error {
 	valueArgs := GetHelmValueArgs(values)
-	params := []string{"install", chart, ociURI, "--version", version}
+	params := []string{"upgrade", "--install", chart, ociURI, "--version", version}
```
g-gaston, line 141: for my own understanding, what's the difference between install and upgrade?
This is going to be a change for all consumers of this method.

pkg/api/v1alpha1/cluster_types.go
ewollesen: No, they can be and are, it's just that it's handled by the CLI for management clusters.

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Yeah, I have a follow-up already for this, just haven't merged it yet. :)

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Hmm. I'll take a look. If it's as easy as wrapping things in a mutex, then it shouldn't be too bad to do.

controllers/cluster_controller.go
ewollesen: Tests indicate that indeed, this isn't necessary at all. Will remove.

controllers/cluster_controller.go
ewollesen: Hmm, the packagescontrollerclient already has ReconcileDelete, so I suppose moving Reconcile over there has a nice symmetry. Does that work for you?

ewollesen: So what I mean to say by the above, is that I think your concern about writing kubeconfig when it isn't necessary is moot now, since that logic is removed. If you have another issue, or I've misunderstood, please let me know.

controllers/cluster_controller.go
ewollesen: It needs only run for workload clusters. Packages installation via FCL isn't supported for management clusters yet, as FCL isn't supported for management clusters yet. :)

IsPackagesEnabled is simply:

```return !c.IsSelfManaged() && (c.Spec.Packages == nil || !c.Spec.Packages.Disable)```

I felt that extracting the proposition to its own method read better.

File: controllers/cluster_controller.go (hunk 12 above)
Comment (ewollesen) [reply controllers/cluster_controller.go:335]: Thanks.

File: pkg/curatedpackages/packagecontrollerclient.go (hunk 16 above)
Comment (ewollesen) [reply pkg/curatedpackages/packagecontrollerclient.go:100]: Noted.Fixed

File: pkg/curatedpackages/packagecontrollerclient.go (hunk 17 above)
Comment (ewollesen) [reply pkg/curatedpackages/packagecontrollerclient.go:112]: I have a solution in place. Doesn't seem too bad. Please let me know if you have ideas.

File: pkg/executables/helm.go (hunk 20 above)
Comment (ewollesen) [reply pkg/executables/helm.go:141]: `upgrade --install` is an "upsert" of sorts. If the release doesn't exist, it's created. If the release exists, it's upgraded. That is to say, it attempts to be idempotent.

`install` will fail if the release already exists, which doesn't work well with a reconcile situation.

File: pkg/executables/helm.go (hunk 19 above)
Comment (ewollesen) [reply pkg/executables/helm.go:25]: It does what it says on the tin... If the helm chart contains CRDs, they'll be skipped (i.e. not installed on the cluster).

Let me see what I can manage. It'd probably mean moving it as a flag to `Install`.

Comment (ewollesen) [reply pkg/executables/helm.go:25]: Yeah, moving it to `InstallChart` makes the most sense to me. That's where we set a few other run-time flags, like `kubeconfig`.

File: config/manifest/eksa-components.yaml [hunk 21]
```diff
@@ -5983,6 +6002,30 @@ rules:
...
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
```
danbudris, line 6020: curious about what's driving the need for such broad permissions against the `role` API

danbudris: is there any way to get around adding these permissions to the controller role? It might be convenient to have them here, but this is extremely broad, given it allows the controller to update ANY role (including this role), effectively allowing it to have unbounded permissions

ewollesen: @g-gaston asked me to document the reasons for a bunch of these changes, so I'll do that with some code comments, and ping both of you when it's up.

danbudris: ok great thanks Eric!

ewollesen: Here's a "more to the point" link: https://github.com/aws/eks-anywhere/blob/99d1c71b48867c6c531cf2085fc47d6921af857f/controllers/cluster_controller_legacy.go#L64

ewollesen: I was able to get rid of the update, but I'll have to do a round of testing to be safe.

g-gaston: Thanks for this!
Few questions
* We need to create a Role for a particular namespace. Does that mean the namespace name is dynamic?
* The role is for the package controller manager. Does that mean there is one copy of the packages controller (one deployment) running in the management cluster per workload cluster? Or does the new role add permissions to the existing packages controller?
* It seems like it only need to delete rolebindings. If that’s the case, how are the new created roles assigned to anything if it’s not creating new rolebindings?
* For the CRD create operation, couldn’t we create all CRDs when we install packages from the CLI? I feel like at that point we should already know all the CRDs we need.
* What are the secrets that it needs to delete? If it’s not creating any (it doesn’t have permissions to do it), where do those secrets come from?
* For namespaces, the doc mentions it creates a namespace in the workload cluster. Does it also create a namespace in the management cluster? If so, is a dynamically named namespace?

ewollesen: > We need to create a Role for a particular namespace. Does that mean the namespace name is dynamic?

Yes, the namespace is of the form "eksa-packages-<workload cluster name>".

> The role is for the package controller manager. Does that mean there is one copy of the packages controller (one deployment) running in the management cluster per workload cluster? Or does the new role add permissions to the existing packages controller?

There is one deployment, regardless of how many clusters. But there is a package bundle controller resource per cluster. So in a situation where you have 1 management and 2 workload clusters, there will be 1 deployment, 1 controller running, but 3 packagebundle controller resources.

> It seems like it only need to delete rolebindings. If that’s the case, how are the new created roles assigned to anything if it’s not creating new rolebindings?

You know, I assumed it already had create permissions from somewhere else? My empirical testing showed that in order to prevent weird errors in the logs I had to be able to remove rolebindings, so I added the permission. It didn't occur to me to track down where the creation permission came from.

> For the CRD create operation, couldn’t we create all CRDs when we install packages from the CLI? I feel like at that point we should already know all the CRDs we need.

Good question. I'll run that by @lewisdiamond as he's got the most knowledge in that area.

> What are the secrets that it needs to delete? If it’s not creating any (it doesn’t have permissions to do it), where do those secrets come from?

There's a k8s cron job installed with the helm chart that keeps registry secrets up to date. They need to be deleted when the cluster is deleted.

> For namespaces, the doc mentions it creates a namespace in the workload cluster. Does it also create a namespace in the management cluster? If so, is a dynamically named namespace?

Yes and yes.

In the management cluster, it creates eksa-packages-<cluster name>. In the workload cluster, it creates "eksa-packages".

ewollesen: I found a regression that was causing the CRDs to be installed, even though they already were. Fixed that, and now we don't need CRDs permissions at all. Yay.

I also looked more closely at the Roles being created. The key role being created is the `&lt;workload cluster name&gt;-eks-anywhere-packages-manager-role`, which is created within the workload-cluster-specific namespace `eksa-packages-&lt;workload cluster name&gt;` So that's a dynamically named role in a dynamically create namespace, hence the create and matching delete permissions, and since we don't know the namespace name ahead of time, we can't restrict them to particular namespaces.

The code's comment is updated to reflect these changes.

pkg/executables/helm.go
```diff
@@ -116,19 +118,26 @@ func (h *Helm) SaveChart(ctx context.Context, ociURI, version, folder string) er
 }
 
 func (h *Helm) InstallChartFromName(ctx context.Context, ociURI, kubeConfig, name, version string) error {
-	params := []string{"install", name, ociURI, "--version", version, "--kubeconfig", kubeConfig}
+	params := []string{"upgade", "--install", name, ociURI, "--version", version, "--kubeconfig", kubeConfig}
```
a-cool-train, line 121: Maybe a typo? upgrade instead?

ewollesen: Indeed. Thank you for the keen eyes! Fixed.

config/manifest/eksa-components.yaml
ewollesen: After resetting the permissions model in my head, plus a lot of chats with @g-gaston and @lewisdiamond, I stripped things down to brass tacks, and was able to remove nearly all the permissions. The only one that remains is full control over package bundle controller resources, which signal the packages controller that it needs to install things or remove them or whatever. So that seems totally reasonable to me. No new roles. :)

The namespace of the package bundle controller resources could be further restricted with a namespace (eksa-pacakges), but with the way kubebuilder works, that means we'd be required to define a matching role named "eksa-packages", and I don't think that buys us anything, since no one but us should be modifying package bundle controller resources anyway, as we're the ones that created the CRD in the first place. :)

controllers/cluster_controller_legacy.go
```diff
@@ -60,6 +60,13 @@ func NewClusterReconcilerLegacy(client client.Client, log logr.Logger, scheme *r
 // +kubebuilder:rbac:groups="",namespace=eksa-system,resources=secrets,verbs=delete;
 // +kubebuilder:rbac:groups=tinkerbell.org,resources=hardware;hardware/status,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=bmc.tinkerbell.org,resources=machines;machines/status,verbs=get;list;watch
+//
+// For the full cluster lifecycle to support Curated Packages, the controller
+// must be able to create, delete, update, and patch package bundle controller
+// resources, which will trigger the curated packages controller to do the
+// rest.
+//
+// +kubebuilder:rbac:groups=packages.eks.amazonaws.com,resources=packagebundlecontrollers,verbs=create;delete;get;list;patch;update;watch;
```
vivek-koppuru, line 69: What is the purpose of adding this comment then? Generally we don't need to backport to legacy controller anyways.

danbudris (approved):
looks good from the perspective of the RBAC configuration scope!

pkg/api/v1alpha1/cluster_types.go
vivek-koppuru: This is confusing to me then, why is there a method for `IsPackagesEnabled` then? That makes it sound like it should apply for any cluster that has packages enabled or not, regardless of if it's handled by the CLI.

Is `IsPackagesSupported` a better name based on the comment? I think understanding the use case better will help explain it.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -32,9 +41,11 @@ const (
...
 
 type PackageControllerClient struct {
-	kubeConfig            string
-	chart                 *releasev1.Image
-	chartInstaller        ChartInstaller
+	kubeConfig     string
+	chart          *releasev1.Image
+	chartInstaller ChartInstaller
+	// deleter handles helm chart install deletion.
+	deleter               ChartInstallationDeleter
```
vivek-koppuru, line 48: Can we just call this `ChartUninstaller`?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +60,102 @@ type PackageControllerClient struct {
...
+		eksaRegion:               eksaDefaultRegion,
+		clientBuilder:            clientBuilder,
+	}
+}
+
+// EnableFullLifecycle wraps Enable to handle run-time arguments.
+//
+// This method fills in the gaps between the original CLI use case, where all
+// information is known at PackageControllerClient initialization, and the
+// Full Cluster Lifecycle use case, where there's limited information at
```
vivek-koppuru, line 112: What are some examples of this limited information?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +60,102 @@ type PackageControllerClient struct {
...
+// initialization.
+func (pc *PackageControllerClient) EnableFullLifecycle(ctx context.Context, log logr.Logger, clusterName, kubeConfig string, chart *releasev1.Image, registryMirror *registrymirror.RegistryMirror, options ...PackageControllerClientOpt) (err error) {
+	log.V(6).Info("enabling curated packages full lifecycle")
+	defer func(err *error) {
+		if err != nil && *err != nil {
+			log.Error(*err, "Enabling curated packages full lifecycle", "clusterName", clusterName)
+		} else {
+			log.Info("Success enabling curated packages full lifecycle")
```
vivek-koppuru, line 120:
```suggestion
			log.Info("Successfully enabled curated packages full lifecycle")
```

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -49,19 +60,102 @@ type PackageControllerClient struct {
...
+		}
+	}(&err)
+	pc.mu.Lock()
+	// This anonymous function ensures that the pc.mu is unlocked before
+	// Enable is called, preventing deadlocks in the event that Enable tries
```
vivek-koppuru, line 125: Not sure if I quite understand why we need to use a lock here. Just curious on the use case.

pkg/dependencies/factory.go
```diff
@@ -1049,6 +1049,7 @@ func (f *Factory) WithPackageControllerClient(spec *cluster.Spec, kubeConfig str
 			return err
 		}
 		f.dependencies.PackageControllerClient = curatedpackages.NewPackageControllerClient(
+			f.dependencies.Helm,
```
vivek-koppuru, line 1052: Why do we have two helm dependencies being passed in here?

pkg/executables/helm.go
```diff
@@ -116,19 +118,26 @@ func (h *Helm) SaveChart(ctx context.Context, ociURI, version, folder string) er
 }
 
 func (h *Helm) InstallChartFromName(ctx context.Context, ociURI, kubeConfig, name, version string) error {
-	params := []string{"install", name, ociURI, "--version", version, "--kubeconfig", kubeConfig}
+	params := []string{"upgrade", "--install", name, ociURI, "--version", version, "--kubeconfig", kubeConfig}
```
vivek-koppuru, line 121: Can we have a comment explaining that this will install if it doesn't exist and upgrade if it does, and any side effects of that?

controllers/cluster_controller_legacy.go
ewollesen: This was the only place I found to add these comments that was actually reflected in the final output. Can you suggest a better place to put them?

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: I guess? Helm has weird wording for a chart installation or whatever. I guess I don't have a strong feeling on it, it's just all confusing.

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Basically, if you look at `NewPackageControllerClient`'s parameters, and subtract these parameters out, you're left with the set of parameters that full cluster lifecycle doesn't know at instantiation. Things like the cluster name. Some of this is because of different timings, and some is because when used by full cluster lifecycle, this package controller client might be reused to build multiple clusters, whereas in a CLI usage, it builds a single cluster, then exits.

There's probably a lot of refactoring that could be done here, but this PR was big enough as is.

pkg/dependencies/factory.go
ewollesen: The first is for installing helm charts, the second is for removing helm charts. They can require different parameters when they're instantiated, so you can't assume that you can use the same object to do both. Kinda annoying, but grants flexibility.

pkg/executables/helm.go
ewollesen: Of course.

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Guillermo pointed out that this isn't thread safe. While we don't have any concurrent code at the moment, he asked that I ensure thread-safety if possible, as there are designs towards concurrency for the future. Since it was relatively simple to add, I did so. This is far from the only thing in our codebase that isn't thread-safe though. ;)

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: I'll add something to the comment.

pkg/dependencies/factory.go
ewollesen: Adding comments

controllers/cluster_controller_legacy.go
vivek-koppuru: This file is for the old controller, so I would suggest to try in cluster_controller.go. Or if it's not applicable then omit.

pkg/dependencies/factory.go
vivek-koppuru: What's a case where different need to be specified? I am wondering if we should only stick to one right now if we don't have that use case for two different right now. Whoever needs two separate can then decide to separate this.

vivek-koppuru (approved):

vivek-koppuru (approved):

controllers/cluster_controller_legacy.go
ewollesen: I'll try again, but the last time I made modifications there, they plain just didn't show up... Which seemed weird, but maybe I was using the wrong makefile target... or something.

pkg/dependencies/factory.go
ewollesen: Removing via followup. There was a case during development, but then that case was eliminated, so they can be collapsed into one now.

PR #4971, opened 2023-02-15
Check Tinkerbell upgrade/create

ddjjia:
*Issue #, if available:*
[Detect upgrade changes and call upgrade specific validations#1044](https://github.com/aws/eks-anywhere-internal/issues/1044)

[Design Doc](https://quip-amazon.com/tPdvA2zEwdd7/Tinkerbell-Cluster-Upgrade) 

*Description of changes:*
- if k8s version changes, rolling upgrade, assert extra hardware available, omit machine template.
- if node number changes, scale upgrade, assert hardware for scale.

*Testing (if applicable):*


pkg/api/v1alpha1/cluster.go
```diff
@@ -833,3 +834,11 @@ func validateMDUpgradeRolloutStrategy(w *WorkerNodeGroupConfiguration) error {
 
 	return nil
 }
+
+// IsRollingNotScale compares current CAPI objects and desired CAPI objects
+// returns true on rolling upgrade, false on scale upgrade.
+// should error if rolling and scale at the same time
+// todo: IsRollingNotScale
+func IsRollingNotScale(ctx context.Context, client client.Client, cluster *Cluster) (bool, error) {
```
vivek-koppuru, line 842: Why are we importing the client object here? Also, how do we check for the previous cluster fields? This is the api package so it's more for operations that don't require communicating with the Kubernetes cluster.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -34,7 +38,7 @@ func (p ControlPlane) Objects() []kubernetes.Object {
 }
 
 // ControlPlaneSpec builds a Tinkerbell ControlPlane definition based on an eks-a cluster spec.
-func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes.Client, clusterSpec *cluster.Spec) (*ControlPlane, error) {
+func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client client.Client, clusterSpec *cluster.Spec) (*ControlPlane, error) {
```
vivek-koppuru, line 41: How does this affect cli flow? How can you make it reusable, unless we need to introduce a different method.

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -51,6 +55,23 @@ func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes
 		return nil, errors.Wrap(err, "generating tinkerbell control plane yaml spec")
 	}
 
+	capiCluster, err := controller.GetCAPICluster(ctx, client, clusterSpec.Cluster)
+	if err != nil {
+		return nil, err
+	}
+	if capiCluster != nil {
+		rollingOverScale, err := anywherev1.IsRollingNotScale(ctx, client, clusterSpec.Cluster)
```
vivek-koppuru, line 63: Where is the comparison happening?

pkg/api/v1alpha1/cluster.go
ddjjia: Raymond moved this method under tinkerbell/upgrade, which should be a better place, since it's very provider specific.

pkg/providers/tinkerbell/controlplane.go
ddjjia: This method is only used in controller, not cli

pkg/providers/tinkerbell/controlplane.go
ddjjia: please find detailed comparator here. https://github.com/aws/eks-anywhere/pull/4994

pkg/providers/tinkerbell/workers.go
```diff
@@ -43,6 +46,23 @@ func WorkersSpec(ctx context.Context, logger logr.Logger, client kubernetes.Clie
...
+		if err != nil {
+			return nil, errors.Wrap(err, "checking upgrade behavior")
+		}
+		if rollingOverScale {
+			workersYaml, err = omitTinkerbellMachineTemplate(workersYaml)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
```
g-gaston, line 64: this logic is repeated int the CP
maybe that's a clue we should rethink how these 3 pieces interact

pkg/api/v1alpha1/cluster.go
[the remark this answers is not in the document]
ddjjia: please find detailed comparator here. https://github.com/aws/eks-anywhere/pull/4994

pkg/providers/tinkerbell/controlplane.go
```diff
@@ -51,6 +55,23 @@ func ControlPlaneSpec(ctx context.Context, logger logr.Logger, client kubernetes
...
+		if err != nil {
+			return nil, errors.Wrap(err, "checking upgrade behavior")
+		}
+		if rollingOverScale {
+			controlPlaneYaml, err = omitTinkerbellMachineTemplate(controlPlaneYaml)
```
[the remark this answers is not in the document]
ddjjia: This maps to [current CLI logic ](https://github.com/aws/eks-anywhere/blob/75eb25f76ffc39344302098cdaddce463efcfea7/pkg/providers/tinkerbell/template.go#L292
)

pkg/providers/tinkerbell/workers.go
ddjjia: will extract the logic to reconciler

pkg/providers/tinkerbell/controlplane.go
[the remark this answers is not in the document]
g-gaston: I see. However I don't think that's a good reason to do it this way. The CLI logic operates over yaml because it was never able to operate over api structs. This code that you are adding here is basically adding that capability.

pkg/api/v1alpha1/cluster.go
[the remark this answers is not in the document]
g-gaston: It seems like that PR is taking a different approach? The signature of the method is radically different. Maybe this one need to wait for that PR? Or viceversa?

chrisdoherty4, line 842: I see some commentary on this moving around. Whereever it lands, could we name it `IsRollingUpgrade`, there's really no need to describe both the affirmative and negative in a name and it generally makes it less usable elsewhere.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -235,8 +224,33 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, clus
...
+			v.Register(
+				tinkerbell.AssertionsForScaleUpDown(etcdReader.GetCatalogue(), clusterSpec, false),
+			)
+		}
+	} else {
+		// CAPI cluster doesn't exists
+		v.Register(
+			tinkerbell.MinimumHardwareAvailableAssertionForCreate(etcdReader.GetCatalogue()),
+		)
+	}
```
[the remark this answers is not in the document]
ddjjia: CLI separates assertion registration for create and upgrade, do you want to just extract the logic for upgrade?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -109,17 +119,54 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	tinkerbellScope.controlPlane = cp
+
+	w, err := tinkerbell.WorkersSpec(ctx, log, clientutil.NewKubeClient(r.client), spec)
+	if err != nil {
+		return controller.Result{}, errors.Wrap(err, "generating workers spec")
+	}
+	tinkerbellScope.workers = w
+
+	err = detectOperation(ctx, log, tinkerbellScope)
+	if err != nil && apierrors.IsNotFound(err) {
```
g-gaston, line 141: why only for `IsNotFound`?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -109,17 +119,54 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+		return controller.Result{}, errors.Wrap(err, "Detecting operation type")
+	}
+
+	// omit machine template on scale update
+	if tinkerbellScope.operationType == scaleUpdate {
+		tinkerbellScope.controlPlane.ControlPlaneMachineTemplate = nil
+		wg := tinkerbellScope.workers.Groups
```
g-gaston, line 148: imo this should be part of the tinkerbell package, in the spec generation code

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -109,17 +119,54 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+		return controller.Result{}, errors.Wrap(err, "generating workers spec")
+	}
+	tinkerbellScope.workers = w
+
+	err = detectOperation(ctx, log, tinkerbellScope)
+	if err != nil && apierrors.IsNotFound(err) {
+		return controller.Result{}, errors.Wrap(err, "Detecting operation type")
+	}
+
+	// omit machine template on scale update
```
g-gaston, line 145: why do they need to be nullified? can't then just have the same value as before?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -33,15 +40,17 @@ type IPValidator interface {
 
 // Scope object for Tinkerbell reconciler.
 type Scope struct {
-	ClusterSpec      *c.Spec
-	isRollingUpgrade bool
+	ClusterSpec   *c.Spec
+	operationType string
```
g-gaston, line 44: this ties again to the conversation we had about "operations/events" vs difference between desired and current state

I think these are the wrong semantics. By introducing the concept of "operation" we give the impression this reconcile loop reacts to independent indivisible events, which I don't think it's true.

I would prefer if we use terms like "cluster needs more nodes", "cluster has extra nodes", "cluster need new nodes", etc.

Note: my comment is not necessarily about how to reflect this in code, but about the semantics the code proposes.

g-gaston, line 44: if you want to use this as something like an enum, I would suggest a new type like `type clusterOperation string`

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -186,7 +229,8 @@ func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, tinkerbe
 	return r.cniReconciler.Reconcile(ctx, log, client, clusterSpec)
 }
 
-func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, clusterSpec *c.Spec) error {
+func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, tinkerbellScope *Scope) error {
```
g-gaston, line 232: does this actually need the scope?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -150,20 +197,16 @@ func (r *Reconciler) ReconcileWorkers(ctx context.Context, log logr.Logger, tink
...
+	return clusters.ReconcileWorkersForEKSA(ctx, log, r.client, spec.Cluster, clusters.ToWorkers(tinkerbellScope.workers))
 }
 
 // ValidateDatacenterConfig updates the cluster status if the TinkerbellDatacenter status indicates that the spec is invalid.
 func (r *Reconciler) ValidateDatacenterConfig(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
 	clusterSpec := tinkerbellScope.ClusterSpec
 	log = log.WithValues("phase", "validateDatacenterConfig")
 
-	if err := r.validateTinkerbellIPMatch(ctx, clusterSpec); err != nil {
+	if err := r.validateTinkerbellIPMatch(ctx, NewScope(clusterSpec)); err != nil {
```
g-gaston, line 209: why create a new scope if you already have one?

pkg/providers/tinkerbell/reconciler/reconciler.go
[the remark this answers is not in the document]
g-gaston: well, it doesn't need to
As long as the validations are the same, it just needs a shared format for the input

pkg/providers/tinkerbell/reconciler/reconciler.go
raymond-zhang00: I think we want to actually ignore NotFound errors, because they could indicate a creating a new cluster or worker node instead of upgrades.

I can put the apierrors.IsNotFound(err) catch in my detectOperation implementation or closer to the client.Get() themselves.

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: actually no, but all phases take the same input

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: sure. I moved it to tinkerbell package but don't let provider consume it yet. I think this could be done in a separate pr, maybe when we clean up all providers and need to have all providers using api cp and workers.

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: My understanding is we don't want to create new machine template on scale. @pokearu may have more context here.

pkg/providers/tinkerbell/reconciler/reconciler.go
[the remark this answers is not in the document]
ddjjia: I notice AssertionsForScaleUpDown also needs some changes for controller to use. I'll address it along with common validation register in another issue. [Make AssertionsForScaleUpDown capable for controller#1060](https://github.com/aws/eks-anywhere-internal/issues/1060)

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: [Clean up] All providers use CAPI api#1061](https://github.com/aws/eks-anywhere-internal/issues/1061)

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -31,17 +39,20 @@ type IPValidator interface {
 	ValidateControlPlaneIP(ctx context.Context, log logr.Logger, spec *c.Spec) (controller.Result, error)
 }
 
+type clusterChange string
```
chrisdoherty4, line 42: What's the motivation to unexport this?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -109,17 +121,49 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
 	return controller.Result{}, nil
 }
 
-// ReconcileControlPlane applies the control plane CAPI objects to the cluster.
-func (r *Reconciler) ReconcileControlPlane(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
+// DetectOperationAndGenerateSpec generates control plane and workers spec, then detect create or upgrade type.
+func (r *Reconciler) DetectOperationAndGenerateSpec(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
```
chrisdoherty4, line 125: Could this be modeled as atomic phases: `DetectOperation` and `GenerateSpec`?

pkg/providers/tinkerbell/reconciler/reconciler.go
chrisdoherty4: If I recall correctly its simply because the existing template is sufficient to perform the scale up. 

The generation logic needs rethinking (separate to these changes) because generate and omit under specific circumstances feels convoluted and error prone.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -239,32 +280,25 @@ func (r *Reconciler) ValidateHardware(ctx context.Context, log logr.Logger, tink
...
 
 	var v tinkerbell.ClusterSpecValidator
-	v.Register(
-		tinkerbell.MinimumHardwareAvailableAssertionForCreate(etcdReader.GetCatalogue()),
-		tinkerbell.HardwareSatisfiesOnlyOneSelectorAssertion(etcdReader.GetCatalogue()),
-	)
+	v.Register(tinkerbell.HardwareSatisfiesOnlyOneSelectorAssertion(etcdReader.GetCatalogue()))
+
+	// todo: AssertionsForScaleUpDown
+	switch *tinkerbellScope.ClusterChange {
```
chrisdoherty4, line 296: Having this decision logic here makes the decision to add the validations execution context dependent (controller vs CLI).

I would think this logic would reside in some shared function, and possibly some other execution context agnostic validations like the `HardwareSatisfiesOnlyOneSelectorAssertion`.

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: to use it as enum https://github.com/aws/eks-anywhere/pull/4971#discussion_r1113710109

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
+package tinkerbell
```
mitalipaygude, line 1: Are we merging the CAPI comparator changes in this PR?

ddjjia: yes

pkg/providers/tinkerbell/comparator_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	g.Expect(KCPComparator.Replicas()).To(Equal(1))
+}
+
+func TestKubeadmControlPlaneVersion(t *testing.T) {
+	g := NewWithT(t)
+
+	KCPComparator := &tinkerbell.KubeadmControlPlane{comparatorKubeadmControlPlane()}
+
+	g.Expect(KCPComparator.Version()).To(Equal("1.22"))
```
mitalipaygude, line 27: nit: we can use kube versions from [here](https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L651-L661) in the places we are hardcoding the version in this test file.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	return workerWantScaleChange
+}
+
+// OmitMachineTemplate omits control plane and worker machine template on scaling update.
+func (r *Reconciler) OmitMachineTemplate(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
+	log = log.WithValues("phase", "OmitMachineTemplate")
+	log.Info("Omit machine template on scaling")
+	if tinkerbellScope.ClusterChange == NeedMoreOrLessNode {
+		tinkerbell.OmitTinkerbellCPMachineTemplate(tinkerbellScope.ControlPlane)
+		tinkerbell.OmitTinkerbellWorkersMachineTemplate(tinkerbellScope.Workers)
```
g-gaston, line 213: can't we just have this as part of the capi spec generation methods?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	if currentKCP == nil {
+		tinkerbellScope.ClusterChange = ReconcileNewCluster
+		return controller.Result{}, nil
+	}
+
+	tinkCurrentKCP := tinkerbell.KubeadmControlPlane{KubeadmControlPlane: currentKCP}
+	tinkDesiredKCP := tinkerbell.KubeadmControlPlane{KubeadmControlPlane: tinkerbellScope.ControlPlane.KubeadmControlPlane}
+	wantVersionChange := tinkerbell.HasKubernetesVersionChange(&tinkCurrentKCP, &tinkDesiredKCP)
+	cpWantScaleChange := tinkerbell.ReplicasDiff(&tinkCurrentKCP, &tinkDesiredKCP)
+	workerWantScaleChange := r.workerReplicasDiff(ctx, tinkerbellScope)
```
g-gaston, line 172: Looking at this, IMO the `ClusterChange` adds unnecessary complexity.

The only thing that is caching is the result of a couple if else's that compare strings and int's that have been already fetched.
Unless we can prove that's having a performance impact, I would vote to remove it and just make the calculation twice. Specially now that is encapsulated in the tinkerbell package.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	if wantVersionChange {
+		if cpWantScaleChange != 0 || workerWantScaleChange {
+			return controller.Result{}, errors.Errorf("cannot perform scale up or down during k8s version change")
+		}
+		tinkerbellScope.ClusterChange = NeedNewNode
+	} else if cpWantScaleChange != 0 || workerWantScaleChange {
+		tinkerbellScope.ClusterChange = NeedMoreOrLessNode
+	} else {
+		return controller.Result{}, errors.Errorf("cannot detect operation type")
+	}
```
chrisdoherty4, line 183: This is a little confusing because both functions mention `diff` with 1 returning `int` (the diff) and the other returning `bool`. Given the boolean logic, it makes me think we aren't really looking for a replica diff either. 

Here's a suggestion that might also be a little clearer combined with some other suggestions, and that embraces the [return early philosophy](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_return_early_rather_than_nesting_deeply):

```suggestion
	// Ensure we're performing (1) at least 1 operation and (2) that only 1 kind of
	// operation being requested.
	//
	// TODO: Add detail on _why_.
	
	upgrade := tinkerbell.HasKubernetesVersionChange(&tinkCurrentKCP, &tinkDesiredKCP)
	scaleCp := tinkerbell.ReplicasDiff(&tinkCurrentKCP, &tinkDesiredKCP) != 0
	scaleWkrs := r.workersHaveReplicaDiff(ctx, tinkerbellScope)
	scale := scaleCp || scaleWkrs
	
	if !upgrade && !scale {
		return controller.Result{}, errors.Errorf("cannot detect operation type")
	}
	
	if upgrade && scale {
		return controller.Result{}, errors.Errorf("cannot upgrade kubernetes version and scale the cluster concurrently")
	}
	
	switch {
	case upgrade:
		scope.Operation = K8sVersionUpgradeOperation
	case scale:
		scope.Operation = ScaleOperation
	}
```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	tinkerbellScope.Workers = w
 
-	return clusters.ReconcileControlPlane(ctx, r.client, toClientControlPlane(cp))
+	return controller.Result{}, nil
+}
+
+// DetectOperation detects change type.
+func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
+	log = log.WithValues("phase", "DetectOperation")
+	log.Info("Detecting operation type")
```
chrisdoherty4, line 157: Small suggestion that minimizes regurgitation in the log event. This can be applied to all the phase begining logs we have.

```suggestion
	log.Info("Begin phase")
```

pkg/controller/clusters/controlplane.go
```diff
@@ -44,7 +45,10 @@ type ControlPlane struct {
 // AllObjects returns all the control plane objects.
 func (cp *ControlPlane) AllObjects() []client.Object {
 	objs := make([]client.Object, 0, 6+len(cp.Other))
-	objs = append(objs, cp.Cluster, cp.ProviderCluster, cp.KubeadmControlPlane, cp.ControlPlaneMachineTemplate)
+	objs = append(objs, cp.Cluster, cp.ProviderCluster, cp.KubeadmControlPlane)
+	if !reflect.ValueOf(cp.ControlPlaneMachineTemplate).IsNil() {
```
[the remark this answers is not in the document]
ddjjia: when omit machine template

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: machine template would be needed in hardware validation. @chrisdoherty4  @raymond-zhang00

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: In order to get the if else result, we also get the current capi cp and md from client. So basically trade off between get those capi objects multiple times or cache the result

chrisdoherty4: We're caching either way. If we need to cache the current state for later then we could have a `scope.GetOperation()` and remove this phase.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -2,21 +2,34 @@ package reconciler
...
 
+const (
+	// ReconcileNewCluster means need to create nodes for a new cluster.
+	ReconcileNewCluster clusterChange = "create-new-cluster"
+	// NeedNewNode means node has k8s version change.
+	NeedNewNode clusterChange = "k8s-version-change"
+	// NeedMoreOrLessNode means node group scales up or scales down.
+	NeedMoreOrLessNode clusterChange = "scale-update"
+)
+
```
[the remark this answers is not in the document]
ddjjia: We're trying to avoid "operations/events" and address difference between desired and current state.
https://github.com/aws/eks-anywhere/pull/4971#discussion_r1113709519

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
-	return clusters.ReconcileControlPlane(ctx, r.client, toClientControlPlane(cp))
+	return controller.Result{}, nil
+}
+
+// DetectOperation detects change type.
+func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinkerbellScope *Scope) (controller.Result, error) {
+	log = log.WithValues("phase", "DetectOperation")
+	log.Info("Detecting operation type")
+
+	currentKCP, err := controller.GetKubeadmControlPlane(ctx, r.client, tinkerbellScope.ClusterSpec.Cluster)
```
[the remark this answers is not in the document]
ddjjia: sure. I can wrap additional context here but `controller.GetKubeadmControlPlane` returns <nil, nil> if can't find the resource

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -111,17 +130,97 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, t
...
+	tinkDesiredKCP := tinkerbell.KubeadmControlPlane{KubeadmControlPlane: tinkerbellScope.ControlPlane.KubeadmControlPlane}
+	wantVersionChange := tinkerbell.HasKubernetesVersionChange(&tinkCurrentKCP, &tinkDesiredKCP)
+	cpWantScaleChange := tinkerbell.ReplicasDiff(&tinkCurrentKCP, &tinkDesiredKCP)
+	workerWantScaleChange := r.workerReplicasDiff(ctx, tinkerbellScope)
+
+	if wantVersionChange {
+		if cpWantScaleChange != 0 || workerWantScaleChange {
+			return controller.Result{}, errors.Errorf("cannot perform scale up or down during k8s version change")
+		}
+		tinkerbellScope.ClusterChange = NeedNewNode
```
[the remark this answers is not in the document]
ddjjia: ditto https://github.com/aws/eks-anywhere/pull/4971#discussion_r1127373658

pkg/providers/tinkerbell/reconciler/reconciler.go
ddjjia: can't really isolating the data retrieval on worker node diff, since we iterate desired worker node group, see if it exists and any replica changes. then break whenever on a change

pkg/providers/tinkerbell/reconciler/reconciler.go
raymond-zhang00: I think I was mistaken, the omits only affect the desired CAPI objects rather than the current CAPI objects. The `AssertionsForScaleUpDown` just need the current machine templates in order to retrieve the hardware selector. It gets desired state hardware selector via the Tinkerbell ClusterSpec. 

We should actually be okay to remove this before the assertion.

pkg/providers/tinkerbell/reconciler/reconciler.go
raymond-zhang00: I think running a something like GetOperation() works, but we would still need to retrieve the current state CAPI objects to cache them for comparison. Does it make sense to keep the phase around to retrieve the current state CAPI? We still need the current state CAPI for the `AssertionsForScaleUpDown` validation.

g-gaston: In case it helps with the conversation, the client we use is a cached client. So, without going too much into how its cache invalidation works (which I'm not an expert in either), it's safe to assume the same object won't be read twice in the same reconciliation loop the majority of the time.

That said, if you still want to "cache" something in the scope to avoid going through the client's cache (which might or might not be more inefficient), I think caching the objects themselves (current CAPI state) is a better idea than caching the "change" type. A couple reasons:
* It's simpler.
* It's more reusable. The current cluster change only allows for one change at the same time. By introducing some assumptions in the abstraction (like you can't have scale up/down at the same time as a kubernetes upgrade), it becomes a bit inflexible.
* It doesn't create a new abstraction.

I think this might be one of those times where we tried to optimized too early. If the only gain between the set of solutions we are considering is performance, my vote is to implement whatever is simpler and then measure performance once we have a working controller.

pkg/providers/tinkerbell/reconciler/reconciler.go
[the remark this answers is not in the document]
chrisdoherty4: We're determining what the user is trying to do so we can enforce appropriate constraints (namely preventing a K8s version upgrade concurrently with a scale). All the while we impose the constraint the concept of an `Operation` will be necessary. Naming it something else, which is arguably more ambiguous, doesn't change that fact.

If we want to model current and desired state only it will be necessary to remove the concurrent operation constraint and I'm not sure we want to do that.

pkg/providers/tinkerbell/reconciler/reconciler.go
pokearu: IMO I think we dont need this log here? We dont usually add any begin and end logs in other phases.

pkg/providers/tinkerbell/reconciler/reconciler.go
pokearu: I am kinda confused here. Technically we dont need to omit right? Isn't there a logic in the template generation to just fetch reuse the same machine template on the cluster already?

I believe it was here https://github.com/aws/eks-anywhere/blob/main/pkg/clusterapi/controlplane.go#L51

And as of the comparator we just wanted to add a condition based on scale up/rolling? 
https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/machinetemplate.go#L25

ddjjia: What about still omitting it in a separate phase, but before assertion?

ddjjia:
Clean up commits to another PR https://github.com/aws/eks-anywhere/pull/5179

PR #4975, opened 2023-02-15
Add vSphere airgapped testing

d8660091:
*Description of changes:* The current vSphere RgistryMirror test doesn't airgap admin machine. To test the scenario where admin machine is airgapped, the new added test airgap the bootstrap cluster and the process through iptables.

*Testing (if applicable):* Manually invoked the codebuild test, running against the downstream branch. The manually triggered test succeeded.


test/e2e/airgap.go
```diff
@@ -0,0 +1,20 @@
+package e2e
+
+import (
+	"github.com/aws/eks-anywhere/test/framework"
+)
+
+// Run airgap deployment but allow bootstrap cluster to access local peers.
```
jonathanmeier5, line 7: small nit: could we follow the convention we use elsewhere and start with the function name? `runVSphereAirgapConfigFlow`?

test/e2e/airgap.go
```diff
@@ -0,0 +1,20 @@
...
+func runVSphereAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs string) {
+	test.GenerateClusterConfig()
+	test.DownloadArtifacts()
+	test.ImportImages()
```
jonathanmeier5, line 11: Based on offline conversation I think we eventually want this changed to use `import images` instead of `import-images`, right? Happy to merge as is, just flagging.

jonathanmeier5 (approved):
Just a few small nits/notes.

Discussed with @d8660091 adding a step to verify that the environment is indeed airgapped, for instance by making a http call against something on the public internet with a short timeout. He's going to do it in a fast follow PR.

/lgtm

test/e2e/airgap.go
d8660091: > Just a few small nits/notes.
> 
> Discussed with @d8660091 adding a step to verify that the environment is indeed airgapped, for instance by making a http call against something on the public internet with a short timeout. He's going to do it in a fast follow PR.
> 
> /lgtm

The task is created: https://github.com/aws/eks-anywhere/issues/4980

d8660091:
> Just a few small nits/notes.
> 
> Discussed with @d8660091 adding a step to verify that the environment is indeed airgapped, for instance by making a http call against something on the public internet with a short timeout. He's going to do it in a fast follow PR.
> 
> /lgtm

The task is created: https://github.com/aws/eks-anywhere/issues/4980

test/e2e/airgap.go
d8660091: Yes, we will deprecate import-images perhaps all at once, which is out of the scope of this PR

test/e2e/airgap.go
```diff
@@ -0,0 +1,20 @@
+package e2e
```
jonathanmeier5, line 1: I think we need the e2e build tags here and the linter should stop complaining:

```suggestion
//go:build e2e

package e2e
```

Issue #1108, opened 2022-01-31
Snow E2E tests


jiayiwang7:
`*` are the new features introduced after Snow beta and should be prioritized. `TestSnow*` are the existing E2E tests that are written but haven't been ran with the latest code.

###  * BottleRocket
- [x] * with stacked Etcd and DHCP
- [ ] * with stacked Etcd and static IPs
- [ ] * with unstacked Etcd and static IPs
- [ ] * with proxy
- [ ] * with registry mirror
- [x] * with node labels and taints

### Ubuntu
- [x] * with stacked Etcd and DHCP
- [ ] * with stacked Etcd and static IPs
- [ ] * with unstacked Etcd and static IPs
- [ ] with proxy
    - `TestSnowKubernetes121UbuntuProxyConfig`
- [ ] with registry mirror
- [x] with node labels and taints
    - `TestSnowKubernetes123LabelsUbuntu`
    - `TestSnowKubernetes123TaintsUbuntu`

(unstacked Etcd with DHCP is not supported)

### Cluster Full Life Cycle (EKS-A Controller)
- [ ] * use GitOps creating/upgrading a workload cluster
- [x] * use kubectl creating/upgrading a workload cluster 

### Other Features that are OS generic
- [ ] * air-gapped env (import images, create cluster with registry mirror)
- [ ] with AWS IAM
    - `TestSnowKubernetes121UbuntuAWSIamAuth`
    - `TestSnowKubernetes122To123AWSIamAuthUpgrade`
- [ ] with OIDC
    - `TestSnowKubernetes121OIDC`
- [ ] with multiple worker groups
    - `TestSnowKubernetes123UbuntuRemoveWorkerNodeGroups`

### K8s Versions
- [ ] 1.21
    - `TestSnowKubernetes121SimpleFlow`
- [ ] 1.22
    - `TestSnowKubernetes122SimpleFlow`
- [ ] 1.23
    - `TestSnowKubernetes123SimpleFlow`
- [ ] 1.24

### Cluster Management
- [ ] scale up/down nodes in cluster
- [ ] upgrade K8s version
- [ ] upgrade cluster with features (proxy, AWS IAM, GitOps, etc.)
- [ ] create a workload cluster through CLI
- [ ] delete a cluster

PR #4979, opened 2023-02-15
Add snow static ip e2e tests

jiayiwang7:
*Issue #, if available:*

*Description of changes:*

#1108 

*Testing (if applicable):*

### DIFF
--- a/internal/pkg/api/snow.go
+++ b/internal/pkg/api/snow.go
@@ -13,6 +13,7 @@ import (
 type SnowConfig struct {
 	datacenterConfig *anywherev1.SnowDatacenterConfig
 	machineConfigs   map[string]*anywherev1.SnowMachineConfig
+	ipPools          map[string]*anywherev1.SnowIPPool
 }
 
 type SnowFiller func(config SnowConfig)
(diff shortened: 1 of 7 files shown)


test/framework/snow.go
```diff
@@ -22,6 +22,10 @@ const (
 	snowPodCidr          = "T_SNOW_POD_CIDR"
 	snowCredentialsFile  = "EKSA_AWS_CREDENTIALS_FILE"
 	snowCertificatesFile = "EKSA_AWS_CA_BUNDLES_FILE"
+	snowIPPoolIPStart    = "T_SNOW_IPPOOL_IPSTART"
```
jonathanmeier5, line 25: Do you think it's worth documenting what these values should be when configuring your snow e2e setup? I realize we don't actually have a lab to run against right now.

internal/pkg/api/snow.go
```diff
@@ -117,3 +123,28 @@ func WithChangeForAllSnowMachines(change SnowMachineConfigFiller) SnowFiller {
 		}
 	}
 }
+
+// WithSnowIPPool sets a SnowIPPool.
+func WithSnowIPPool(name, ipStart, ipEnd, gateway, subnet string) SnowFiller {
```
jonathanmeier5, line 128: Do we already have unit test coverage here? If not, do you think it's worth adding?

internal/pkg/api/snowmachines.go
```diff
@@ -66,6 +66,21 @@ func WithDHCP() SnowMachineConfigFiller {
 	}
 }
 
+// WithStaticIP configures one single primary DNI using static ip for IP allocation.
+func WithStaticIP(poolName string) SnowMachineConfigFiller {
```
jonathanmeier5, line 70: Same here. Do we already have unit test coverage, or are we ok with just e2e?

jonathanmeier5 (commented):
A few small questions.

Adding lgtm in case we are comfortable with current test coverage. Happy to give another lgtm if we decide to add more unit tests.

/lgtm

test/framework/snow.go
jiayiwang7: updated readme: https://github.com/aws/eks-anywhere/pull/4979/files#diff-7d9a55ec9712c1ccc931efe126dbea316dab465e6ab265abd089736be5511f47R158

PR #4986, opened 2023-02-16
Add go.work File

jonathanmeier5:
*Description of changes:*

The eks-anywhere repo includes two separate go modules located at ./ and ./release.
By default, vscode will download and resolve dependencies for the module at ./, but throws errors unable to resolve dependencies for packages in the ./release module not present in ./.

We can tell vscode that our workspace actually includes two modules by adding a go.work file pointing at our two modules.

With this file present, vscode knows to look for a go.mod in each module directory and download the dependencies.

Read more about go.work [here](https://go.dev/blog/get-familiar-with-workspaces) and how it works with gopls and vscode [here](https://github.com/golang/tools/blob/268cb0b82f51be725f8ec432c155a92075cd54c6/gopls/doc/workspace.md?plain=1#L25)


TerryHowe:
I wonder if that works for Goland

PR #4994, opened 2023-02-16
Tinkerbell Upgrade CAPI Comparator

raymond-zhang00:
*Issue #, if available:*
[Design Doc](https://quip-amazon.com/tPdvA2zEwdd7/Tinkerbell-Cluster-Upgrade) 
[Custom comparator for CAPI spec for Upgrade#1042](https://github.com/aws/eks-anywhere-internal/issues/1042)

*Description of changes:*
This implements comparator functions for CAPI objects to help detect Tinkerbell rolling/scaling upgrades.

*Testing (if applicable):*
added unit tests

### DIFF
--- a/pkg/clusterapi/apibuilder.go
+++ b/pkg/clusterapi/apibuilder.go
@@ -133,7 +133,7 @@ func KubeadmControlPlane(clusterSpec *cluster.Spec, infrastructureObject APIObje
 			Kind:       kubeadmControlPlaneKind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      KubeadmControlPlaneName(clusterSpec),
+			Name:      KubeadmControlPlaneName(clusterSpec.Cluster),
 			Namespace: constants.EksaSystemNamespace,
 		},
 		Spec: controlplanev1.KubeadmControlPlaneSpec{
--- a/pkg/controller/clusterapi.go
+++ b/pkg/controller/clusterapi.go
@@ -6,6 +6,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
@@ -40,4 +41,23 @@ func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
 		Name:      clusterapi.ClusterName(cluster),
 		Namespace: constants.EksaSystemNamespace,
 	}
+}
+
+// GetKubeadmControlPlane reads a cluster-api KubeadmControlPlane for an eks-a cluster using a kube client
+// If the KubeadmControlPlane is not found, the method returns (nil, nil).
+func GetKubeadmControlPlane(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) (*controlplanev1.KubeadmControlPlane, error) {
+	kubeadmControlPlaneName := clusterapi.KubeadmControlPlaneName(cluster)
+
+	kubeadmControlPlane := &controlplanev1.KubeadmControlPlane{}
+	key := types.NamespacedName{Namespace: constants.EksaSystemNamespace, Name: kubeadmControlPlaneName}
+
+	err := client.Get(ctx, key, kubeadmControlPlane)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return kubeadmControlPlane, nil
 }
--- a/pkg/providers/tinkerbell/comparator.go
+++ b/pkg/providers/tinkerbell/comparator.go
@@ -1,0 +1,51 @@
+package tinkerbell
+
+import (
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+)
+
+// Versionable enables kubernetes version retrieval.
+type Versionable interface {
+	Version() string
+}
+
+// Replicable enables replica number retrieval.
+type Replicable interface {
+	Replicas() int
+}
+
+// KubeadmControlPlane wraps the KubeadmControlPlane to enable Comparator functions.
+type KubeadmControlPlane struct {
+	*controlplanev1.KubeadmControlPlane
+}
+
+// Version retrieves KubeadmControlPlane replicas.
+func (k *KubeadmControlPlane) Version() string {
+	return k.Spec.Version
+}
+
+// Replicas retrieves KubeadmControlPlane replicas.
+func (k *KubeadmControlPlane) Replicas() int {
+	return int(*k.Spec.Replicas)
+}
+
+// MachineDeployment implements Replicable by decorating a MachineDeployment.
+type MachineDeployment struct {
+	*clusterv1.MachineDeployment
+}
+
+// Replicas retrieves MachineDeployment replicas.
+func (m *MachineDeployment) Replicas() int {
+	return int(*m.Spec.Replicas)
+}
+
+// HasKubernetesVersionChange detects kubernetes version has changed between two Comparator objects.
+func HasKubernetesVersionChange(current Versionable, desired Versionable) bool {
+	return current.Version() != desired.Version()
+}
+
+// ReplicasDiff returns the difference in replica count between new and old Comparator objects.
+func ReplicasDiff(current Replicable, desired Replicable) int {
+	return desired.Replicas() - current.Replicas()
+}
--- a/pkg/providers/tinkerbell/comparator_test.go
+++ b/pkg/providers/tinkerbell/comparator_test.go
@@ -1,0 +1,106 @@
+package tinkerbell_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/providers/tinkerbell"
+	"github.com/aws/eks-anywhere/pkg/utils/ptr"
+)
+
+func TestKubeadmControlPlaneReplicas(t *testing.T) {
+	g := NewWithT(t)
+
+	KCPComparator := &tinkerbell.KubeadmControlPlane{comparatorKubeadmControlPlane()}
+
+	g.Expect(KCPComparator.Replicas()).To(Equal(1))
+}
+
+func TestKubeadmControlPlaneVersion(t *testing.T) {
+	g := NewWithT(t)
+
+	KCPComparator := &tinkerbell.KubeadmControlPlane{comparatorKubeadmControlPlane()}
+
+	g.Expect(KCPComparator.Version()).To(Equal("1.22"))
+}
+
+func TestMachineDeploymentReplicas(t *testing.T) {
+	g := NewWithT(t)
+
+	MDComparator := &tinkerbell.MachineDeployment{comparatorMachineDeployment()}
+
+	g.Expect(MDComparator.Replicas()).To(Equal(1))
+}
+
+func TestKubernetesVersionChangeFalse(t *testing.T) {
+	g := NewWithT(t)
+
+	oldComparator := &testComparable{version: "1.22"}
+	newComparator := &testComparable{version: "1.22"}
+
+	g.Expect(tinkerbell.HasKubernetesVersionChange(oldComparator, newComparator)).To(Equal(false))
+}
+
+func TestKubernetesVersionChangeTrue(t *testing.T) {
+	g := NewWithT(t)
+
+	oldComparator := &testComparable{version: "1.22"}
+	newComparator := &testComparable{version: "1.23"}
+
+	g.Expect(tinkerbell.HasKubernetesVersionChange(oldComparator, newComparator)).To(Equal(true))
+}
+
+func TestReplicasChangeUnchanged(t *testing.T) {
+	g := NewWithT(t)
+
+	oldComparator := &testComparable{replicas: 1}
+	newComparator := &testComparable{replicas: 1}
+
+	g.Expect(tinkerbell.ReplicasDiff(oldComparator, newComparator)).To(Equal(0))
+}
+
+func TestReplicasChangeChanged(t *testing.T) {
+	g := NewWithT(t)
+
+	oldComparator := &testComparable{replicas: 1}
+	newComparator := &testComparable{replicas: 3}
+
+	g.Expect(tinkerbell.ReplicasDiff(oldComparator, newComparator)).To(Equal(2))
+}
+
+func comparatorKubeadmControlPlane() *controlplanev1.KubeadmControlPlane {
+	k := &controlplanev1.KubeadmControlPlane{
+		Spec: controlplanev1.KubeadmControlPlaneSpec{
+			Replicas: ptr.Int32(1),
+			Version:  "1.22",
+		},
+	}
+
+	return k
+}
+
+func comparatorMachineDeployment() *clusterv1.MachineDeployment {
+	m := &clusterv1.MachineDeployment{
+		Spec: clusterv1.MachineDeploymentSpec{
+			Replicas: ptr.Int32(1),
+		},
+	}
+
+	return m
+}
+
+type testComparable struct {
+	version  string
+	replicas int
+}
+
+func (t *testComparable) Version() string {
+	return t.version
+}
+
+func (t *testComparable) Replicas() int {
+	return t.replicas
+}
--- a/pkg/clusterapi/name.go
+++ b/pkg/clusterapi/name.go
@@ -72,8 +72,9 @@ func DefaultObjectName(baseName string) string {
 	return ObjectName(baseName, 1)
 }
 
-func KubeadmControlPlaneName(clusterSpec *cluster.Spec) string {
-	return clusterSpec.Cluster.GetName()
+// KubeadmControlPlaneName generates the kubeadmControlPlane name for an EKSA Cluster.
+func KubeadmControlPlaneName(cluster *v1alpha1.Cluster) string {
+	return cluster.GetName()
 }
 
 // EtcdClusterName sets the default EtcdCluster object name.
@@ -110,7 +111,7 @@ func WorkerMachineTemplateName(clusterSpec *cluster.Spec, workerNodeGroupConfig 
 }
 
 func ControlPlaneMachineHealthCheckName(clusterSpec *cluster.Spec) string {
-	return fmt.Sprintf("%s-kcp-unhealthy", KubeadmControlPlaneName(clusterSpec))
+	return fmt.Sprintf("%s-kcp-unhealthy", KubeadmControlPlaneName(clusterSpec.Cluster))
 }
 
 func WorkerMachineHealthCheckName(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1.WorkerNodeGroupConfiguration) string {
--- a/pkg/clusterapi/name_test.go
+++ b/pkg/clusterapi/name_test.go
@@ -133,7 +133,7 @@ func TestKubeadmControlPlaneName(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			g := newApiBuilerTest(t)
-			g.Expect(clusterapi.KubeadmControlPlaneName(g.clusterSpec)).To(Equal(tt.want))
+			g.Expect(clusterapi.KubeadmControlPlaneName(g.clusterSpec.Cluster)).To(Equal(tt.want))
 		})
 	}
 }
--- a/pkg/controller/clusterapi_test.go
+++ b/pkg/controller/clusterapi_test.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	_ "github.com/aws/eks-anywhere/internal/test/envtest"
@@ -46,6 +47,36 @@ func TestGetCAPIClusterError(t *testing.T) {
 	g.Expect(err).To(MatchError(ContainSubstring("no kind is registered for the type")))
 }
 
+func TestGetKubeadmControlPlaneSuccess(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	eksaCluster := eksaCluster()
+	kubeadmControlPlane := kubeadmControlPlane()
+	client := fake.NewClientBuilder().WithObjects(eksaCluster, kubeadmControlPlane).Build()
+
+	g.Expect(controller.GetKubeadmControlPlane(ctx, client, eksaCluster)).To(Equal(kubeadmControlPlane))
+}
+
+func TestGetKubeadmControlPlaneMissingKubeadmControlPlane(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	eksaCluster := eksaCluster()
+	client := fake.NewClientBuilder().WithObjects(eksaCluster).Build()
+
+	g.Expect(controller.GetKubeadmControlPlane(ctx, client, eksaCluster)).To(BeNil())
+}
+
+func TestGetKubeadmControlPlaneError(t *testing.T) {
+	g := NewWithT(t)
+	ctx := context.Background()
+	eksaCluster := eksaCluster()
+	// This should make the client fail because CRDs are not registered
+	client := fake.NewClientBuilder().WithScheme(runtime.NewScheme()).Build()
+
+	_, err := controller.GetKubeadmControlPlane(ctx, client, eksaCluster)
+	g.Expect(err).To(MatchError(ContainSubstring("no kind is registered for the type")))
+}
+
 func TestGetCapiClusterObjectKey(t *testing.T) {
 	g := NewWithT(t)
 	eksaCluster := eksaCluster()
@@ -76,6 +107,19 @@ func capiCluster() *clusterv1.Cluster {
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Cluster",
 			APIVersion: clusterv1.GroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-cluster",
+			Namespace: "eksa-system",
+		},
+	}
+}
+
+func kubeadmControlPlane() *controlplanev1.KubeadmControlPlane {
+	return &controlplanev1.KubeadmControlPlane{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "KubeadmControlPlane",
+			APIVersion: controlplanev1.GroupVersion.String(),
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "my-cluster",


pkg/clusterapi/comparator.go
```diff
@@ -0,0 +1,27 @@
...
+	Workers[M]
+}
+
+// KubernetesVersionModified checks if kubernetes version differs from current state KubeadmControlPlane.
+func KubernetesVersionModified(oldKubeadm controlplanev1.KubeadmControlPlane, newKubeadm controlplanev1.KubeadmControlPlane) bool {
+	return oldKubeadm.Spec.Version != newKubeadm.Spec.Version
+}
+
+// ControlPlaneScaleUpDown returns the change in number of replicas of kubeadm.
+func ControlPlaneScaleUpDown(oldKubeadm controlplanev1.KubeadmControlPlane, newKubeadm controlplanev1.KubeadmControlPlane) int32 {
```
ddjjia, line 20: any reason not return bool?

raymond-zhang00: I think the number of replicas to scale up/down will be needed for later Tinkerbell scaling upgrade hardware validations and I wanted to make this reusable for that.

pkg/providers/tinkerbell/apibuilder.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return kcp, nil
+}
+
+func MachineDeployment(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1.WorkerNodeGroupConfiguration, kubeadmConfigTemplate *bootstrapv1.KubeadmConfigTemplate, tinkerbellMachineTemplate *tinkerbellv1.TinkerbellMachineTemplate) *clusterv1.MachineDeployment {
+	return clusterapi.MachineDeployment(clusterSpec, workerNodeGroupConfig, kubeadmConfigTemplate, tinkerbellMachineTemplate)
+}
+
+// MachineTemplate builds a snowMachineTemplate based on an eks-a snowMachineConfig and a capasIPPool.
+func MachineTemplate(name string, machineConfig *v1alpha1.TinkerbellMachineConfig) *tinkerbellv1.TinkerbellMachineTemplate {
```
ddjjia, line 36: What's the usage of this one?

pkg/providers/tinkerbell/apibuilder.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, fmt.Errorf("generating KubeadmControlPlane: %v", err)
+	}
+	return kcp, nil
+}
+
+func MachineDeployment(clusterSpec *cluster.Spec, workerNodeGroupConfig v1alpha1.WorkerNodeGroupConfiguration, kubeadmConfigTemplate *bootstrapv1.KubeadmConfigTemplate, tinkerbellMachineTemplate *tinkerbellv1.TinkerbellMachineTemplate) *clusterv1.MachineDeployment {
+	return clusterapi.MachineDeployment(clusterSpec, workerNodeGroupConfig, kubeadmConfigTemplate, tinkerbellMachineTemplate)
+}
+
+// MachineTemplate builds a snowMachineTemplate based on an eks-a snowMachineConfig and a capasIPPool.
```
ddjjia, line 35:
```suggestion
// MachineTemplate builds a tinkerbellMachineTemplate based on an eks-a tinkerbellMachineConfig.
```

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -506,3 +512,52 @@ func toRufioMachines(items []rufiounreleased.BaseboardManagement) []rufiov1.Mach
...
+	if err != nil {
+		return false, err
+	}
+
+	currentKCP := &controlplanev1.KubeadmControlPlane{}
+	err = client.Get(ctx, runtimeclient.ObjectKey{
+		Name:      cp.KubeadmControlPlane.Name,
+		Namespace: cp.KubeadmControlPlane.Namespace,
+	}, currentKCP)
+	if apierrors.IsNotFound(err) {
```
ddjjia, line 536: Can we return false and err here? This could be a better place to place cluster existence check, to remove duplicate logic in reconciler/cp/worker. I'll update my pr and check error for `IsNotFound`.

pkg/providers/tinkerbell/apibuilder.go
raymond-zhang00: It's to generate a TinkerbellMachineTemplate, which is used by capi.KubeadmControlPlane() and capi.MachineDeployment(). I think another function to generate KubeadmConfigTemplate will also be necessary for MachineDeployment

pkg/clusterapi/comparator.go
raymond-zhang00: I think this should return 0 if there is no scale up/down, and then a positive or negative integer depending on whether we are scaling up or down - I was hoping this could be used for the assertion later to calculate the number of hardware we need. 

Should I make another function/replace this one with one that returns bool for use with IsRollingNotScale()?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -89,6 +91,19 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, c
...
 
+// ValidateDatacenterConfig updates the cluster status if the TinkerbellDatacenter status indicates that the spec is invalid.
+func (r *Reconciler) RetrieveCAPIState(ctx context.Context, log logr.Logger, clusterSpec *c.Spec) (controller.Result, error) {
+	log = log.WithValues("phase", "retrieve CAPI state")
+
+	// Retrieve CAPI KubeadmControlPlane for control plane comparisons
+	// Retrieve CAPI MachineDeployments for worker node comparisons
+
+	// Store into ClusterApiState
+	r.clusterState = &tinkerbell.ClusterApiState{}
```
ddjjia, line 102: what is the cluster api state here? I thought we're going to store operation in reconciler? like rolling/scaling/others

pkg/controller/clusterapi.go
```diff
@@ -41,3 +42,22 @@ func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
 		Namespace: constants.EksaSystemNamespace,
 	}
 }
+
+// GetKubeadmControlPlane reads a cluster-api KubeadmControlPlane for an eks-a cluster using a kube client
+// If the KubeadmControlPlane is not found, the method returns (nil, nil).
+func GetKubeadmControlPlane(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) (*controlplanev1.KubeadmControlPlane, error) {
```
ddjjia, line 48: This is not needed right?

raymond-zhang00: I was going to use this in r.DetectUpgrade we use to pull back the kubeadmControlPlane. Is there a better place I should put this logic?

ddjjia: I c, this one tries to get the existing CAPI.

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -506,3 +506,20 @@ func toRufioMachines(items []rufiounreleased.BaseboardManagement) []rufiov1.Mach
 	}
 	return machines
 }
+
+// IsRollingNotScale compares current CAPI objects and desired CAPI objects
+// returns true on rolling upgrade, false on scale upgrade or create
+// should error if rolling and scale at the same time
+// todo: IsRollingNotScale
+func IsRollingNotScale(currentState *ClusterApiState, clusterSpec *cluster.Spec) (bool, error) {
```
g-gaston, line 514: I would suggest using an interface and comparing two of the same interface
That allows you to change the types you use to represent current and desired state. You are already using method to access the data, so that actually comes for free.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -89,6 +91,29 @@ func (r *Reconciler) ValidateClusterSpec(ctx context.Context, log logr.Logger, c
...
+
+	rollingUpgrade, err := tinkerbell.IsRollingNotScale(clusterState, clusterSpec)
+	if err != nil {
+		log.Error(err, "Invalid Upgrade")
+		failureMessage := err.Error()
+		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
+		return controller.ResultWithReturn(), nil
+	}
+
+	r.isRollingUpgrade = rollingUpgrade
```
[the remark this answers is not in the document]
ddjjia: That's true. Reconciler should be stateless. We need the operation detection at 3 places (hardware validation, control plane generation and worker generation). So we were thinking just run the comparator once and save it.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,22 @@
...
+
+func getClusterSpecCAPIKubeVersion(clusterSpec *cluster.Spec) string {
+	return clusterSpec.VersionsBundle.KubeDistro.Kubernetes.Tag
+}
+
+func (c *ClusterApiState) GetControlPlaneReplicas() int {
+	return int(*c.KubeadmControlPlane.Spec.Replicas)
+}
+
+func (c *ClusterApiState) GetControlPlaneVersion() string {
```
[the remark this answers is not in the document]
raymond-zhang00: Removed comparator.go since we can compare CAPI to CAPI

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,22 @@
+package tinkerbell
+
+import (
+	"github.com/aws/eks-anywhere/pkg/cluster"
+)
+
+type ClusterApiState struct {
```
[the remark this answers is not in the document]
raymond-zhang00: Using Scope to track Tinkerbell Cluster objects now

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -888,6 +888,10 @@ func (c *Cluster) EtcdAnnotation() string {
 	return etcdAnnotation
 }
 
+func (c *Cluster) ControlPlaneReplicas() int {
```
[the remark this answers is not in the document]
raymond-zhang00: Removed this from API.

pkg/controller/clusterapi.go
```diff
@@ -41,3 +42,22 @@ func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
...
+	kubeadmControlPlaneName := clusterapi.ClusterName(cluster)
```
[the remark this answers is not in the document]
raymond-zhang00: Added KubeadmControlPlane name.

pkg/providers/tinkerbell/reconciler/reconciler.go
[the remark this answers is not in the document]
raymond-zhang00: We updated the phase runner to accept generic objects and implemeted a Scope object in the tinkerbell reconciler to hold our state information.

pkg/providers/tinkerbell/reconciler/reconciler.go
raymond-zhang00: State info is stored in Scope now

pkg/clusterapi/comparator.go
raymond-zhang00: No longer using bool to track operation.

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -33,16 +41,32 @@ type IPValidator interface {
...
+	workers       *tinkerbell.Workers
 }
 
 // NewScope creates a new Tinkerbell Reconciler Scope.
 func NewScope(clusterSpec *c.Spec) *Scope {
 	return &Scope{
-		ClusterSpec:      clusterSpec,
-		isRollingUpgrade: false,
+		ClusterSpec:   clusterSpec,
+		operationType: createOperation,
```
ddjjia, line 54: why default to create?

raymond-zhang00: You're right, I added this to help troubleshoot some tests but my setOperationType() function was defaulting to Create if it does not detect scaling or rolling. But in reality it should stay as "" to prevent create hardware validations from being run.

pkg/clusterapi/apibuilder.go
```diff
@@ -133,7 +133,7 @@ func KubeadmControlPlane(clusterSpec *cluster.Spec, infrastructureObject APIObje
 			Kind:       kubeadmControlPlaneKind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      KubeadmControlPlaneName(clusterSpec),
+			Name:      KubeadmControlPlaneName(clusterSpec.Cluster),
```
ddjjia, line 136: Why do we need changes under clusterapi?

raymond-zhang00: This was so I can use the `KubeadmControlPlaneName()` function in `GetKubeadmControlPlane` with the cluster instead of requiring a clusterSpec: https://github.com/aws/eks-anywhere/pull/4994/files#diff-11f996f58e1db6221bc9e3ea4a62440e4071227bf110ecf71efb26024c34218cR49

Just like `GetCAPICluster()`

pkg/controller/clusterapi.go
```diff
@@ -41,3 +42,39 @@ func CapiClusterObjectKey(cluster *anywherev1.Cluster) client.ObjectKey {
...
+
+	if err != nil {
+		return nil, err
+	}
+	return kubeadmControlPlane, nil
+}
+
+// GetMachineDeployment reads a cluster-api MachineDeployment of a given name using a kube client
+// If the MachineDeployment is not found, the method returns (nil, nil).
+func GetMachineDeployment(ctx context.Context, client client.Client, name string) (*clusterv1.MachineDeployment, error) {
```
g-gaston, line 67: do we need a method for this? It's not encapsulating a particular logic to derive one object from another, you are just passing the name.
It's basically a wrapper around the client that just checks for not found error.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,30 @@
+package tinkerbell
+
+import (
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+)
+
+// ControlPlaneVersionChange detects kubernetes version has changed between new and old KubeadminControlPlane objects.
+func ControlPlaneVersionChange(oldKCP *controlplanev1.KubeadmControlPlane, newKCP *controlplanev1.KubeadmControlPlane) bool {
+	if oldKCP != nil && newKCP != nil {
```
g-gaston, line 10: why do we need to check for nil?

pkg/providers/tinkerbell/comparator_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestControlPlaneChanged(t *testing.T) {
+	g := NewWithT(t)
+
+	oldKubeadmControlPlane := comparatorKubeadmControlPlane()
+	newKubeadmControlPlane := comparatorKubeadmControlPlane()
+	newKubeadmControlPlane.Spec.Version = "1.23"
+	newKubeadmControlPlane.Spec.Replicas = ptr.Int32(2)
+
+	g.Expect(tinkerbell.ControlPlaneVersionChange(oldKubeadmControlPlane, newKubeadmControlPlane), true)
+	g.Expect(tinkerbell.ControlPlaneReplicasChange(oldKubeadmControlPlane, newKubeadmControlPlane), true)
```
g-gaston, line 33: could we have independent test per tested method?

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,30 @@
...
+		return oldKCP.Spec.Version != newKCP.Spec.Version
+	}
+	return false
+}
+
+// ControlPlaneReplicasChange detects whether replica count has changed between new and old KubeadminControlPlane objects.
+func ControlPlaneReplicasChange(oldKCP *controlplanev1.KubeadmControlPlane, newKCP *controlplanev1.KubeadmControlPlane) bool {
```
g-gaston, line 17: I thought you were gonna make these comparisons not dependent on the types, like implementing an interface?

pkg/providers/tinkerbell/comparator_test.go
raymond-zhang00: Yeah I'll split those to be per method.

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: I'll take the nil check out here. I had the nil checks to allow me to pass values from funcs like `GetKubeadmControlPlane()` directly to the comparison but that is mixing logic.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,30 @@
...
+	if oldKCP != nil && newKCP != nil {
+		return oldKCP.Spec.Replicas != newKCP.Spec.Replicas
+	}
+	return false
+}
+
+// WorkerNodeReplicasChange detects whether replica count has changed between new and old MachineDeployment objects.
+func WorkerNodeReplicasChange(oldMD *clusterv1.MachineDeployment, newMD *clusterv1.MachineDeployment) bool {
```
cxbrowne1207, line 25: I remember we had a previous discussion about returning an `int` instead of a `bool` to detect change. Did we decide on using a bool instead?

raymond-zhang00: Thanks for catching that. I think `int` would be more reusable, updating to return that.

pkg/controller/clusterapi.go
raymond-zhang00: I'll take this out. Calling it in `detectOperation()` will still require us to do a nil check, same as apierror.IsNotFound, so it won't save much trouble.

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Implemented comparator interfaces for kubernetes version and replica count retrieval

cxbrowne1207 (commented):
Left a comment

pkg/providers/tinkerbell/comparator_test.go
```diff
@@ -0,0 +1,98 @@
...
+	g.Expect(tinkerbell.KubernetesVersionChange(oldKCP, newKCP), true)
+}
+
+func TestKubeadmControlPlaneReplicasUnchanged(t *testing.T) {
+	g := NewWithT(t)
+
+	oldKCPComparator := &tinkerbell.KubeadmControlPlaneComparator{comparatorKubeadmControlPlane()}
+	newKCPComparator := &tinkerbell.KubeadmControlPlaneComparator{comparatorKubeadmControlPlane()}
+
+	g.Expect(tinkerbell.ReplicasChange(oldKCPComparator, newKCPComparator), false)
```
cxbrowne1207, line 41: Should a `boolean` be expected here?

raymond-zhang00: Ah, that should definitely be a int. Thank you.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
+package tinkerbell
+
+import (
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+)
+
+// VersionComparator enables kubernetes version retrieval.
+type VersionComparator interface {
```
g-gaston, line 9: nit: this is not a comparator
Is more a "comparable"?

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	GetVersion() string
+}
+
+// ReplicaComparator enables replica number retrieval.
+type ReplicaComparator interface {
```
g-gaston, line 14: what kind of replicas? or are you using this for different comparisons?

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	GetReplicas() int
+}
+
+// KubeadmControlPlaneComparator wraps the KubeadmControlPlane to enable Comparator functions.
+type KubeadmControlPlaneComparator struct {
+	*controlplanev1.KubeadmControlPlane
+}
+
+// GetVersion retrieves KubeadmControlPlane replicas.
+func (k *KubeadmControlPlaneComparator) GetVersion() string {
```
g-gaston, line 24: prefer accessors without the `Get`
```suggestion
func (k *KubeadmControlPlaneComparator) Version() string {
```

https://go.dev/doc/effective_go#Getters

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	return k.Spec.Version
+}
+
+// GetReplicas retrieves KubeadmControlPlane replicas.
+func (k *KubeadmControlPlaneComparator) GetReplicas() int {
+	return int(*k.Spec.Replicas)
+}
+
+// MachineDeploymentComparator wraps the MachineDeployment to enable Comparator functions.
```
g-gaston, line 33: it only enable one
maybe something like
```suggestion
// MachineDeploymentComparator implements ReplicaComparator by decorating a MachineDeployment.
```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -34,15 +34,13 @@ type IPValidator interface {
 
 // Scope object for Tinkerbell reconciler.
 type Scope struct {
-	ClusterSpec      *c.Spec
-	isRollingUpgrade bool
+	ClusterSpec *c.Spec
```
g-gaston, line 37: nit: we probably don't need these changes in this PR?

g-gaston (commented):
niceeee!

pkg/providers/tinkerbell/comparator.go
chrisdoherty4: What's the rationale for an int? What makes an int re-usable? Where does this need to be re-used?

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	return int(*m.Spec.Replicas)
+}
+
+// KubernetesVersionChange detects kubernetes version has changed between two Comparator objects.
+func KubernetesVersionChange(oldObj VersionComparator, newObj VersionComparator) bool {
+	return oldObj.GetVersion() != newObj.GetVersion()
+}
+
+// ReplicasChange returns the difference in replica count between new and old Comparator objects.
+func ReplicasChange(oldObj ReplicaComparator, newObj ReplicaComparator) int {
```
chrisdoherty4, line 49: Would you be open to `ReplicasDiff`? 'Diff' seems more representative vs 'Change' and is more common place.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	*clusterv1.MachineDeployment
+}
+
+// GetReplicas retrieves MachineDeployment replicas.
+func (m *MachineDeploymentComparator) GetReplicas() int {
+	return int(*m.Spec.Replicas)
+}
+
+// KubernetesVersionChange detects kubernetes version has changed between two Comparator objects.
+func KubernetesVersionChange(oldObj VersionComparator, newObj VersionComparator) bool {
```
chrisdoherty4, line 44: Would you be open to `HasKubernetesVersionChange`? This makes the readability at the site of consumption a little easier.

```
if HasKubernetesVersionChange() { ... }

// vs

if KubernetesVersionChange() { ... }

---

```suggestion
func HasKubernetesVersionChange(oldObj, newObj VersionComparator) bool {
```

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Updated naming for this to "Versionable" to match "Replicatable"

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Updating naming to "Replicatable". I was using this for getting the machine replicas, whether control plane or worker node. I'll update to `MachineReplicaComparator`

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Yes, that name is a lot more descriptive.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	*clusterv1.MachineDeployment
+}
+
+// Replicas retrieves MachineDeployment replicas.
+func (m *MachineDeploymentComparator) Replicas() int {
+	return int(*m.Spec.Replicas)
+}
+
+// HasKubernetesVersionChange detects kubernetes version has changed between two Comparator objects.
+func HasKubernetesVersionChange(oldObj VersionComparable, newObj VersionComparable) bool {
```
g-gaston, line 44:
```suggestion
func HasKubernetesVersionChange(current VersionComparable, desired VersionComparable) bool {
```

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	return oldObj.Version() != newObj.Version()
+}
+
+// ReplicasDiff returns the difference in replica count between new and old Comparator objects.
+func ReplicasDiff(oldObj MachineReplicaComparable, newObj MachineReplicaComparable) int {
```
g-gaston, line 49:
```suggestion
func ReplicasDiff(current MachineReplicaComparable, desired MachineReplicaComparable) int {
```

pkg/providers/tinkerbell/comparator.go
g-gaston, line 44: where are these two methods going to be used?

raymond-zhang00: This would be to detect the Tinkerbell Upgrade Operation, whether it's rolling or scaling. So these will be to compare the CAPI current vs desired state via KubeadmControlPlane for control plane and MachineDeployment for worker nodes.

pkg/providers/tinkerbell/comparator.go
chrisdoherty4: I think we should double down on this terminology in our code!

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
+package tinkerbell
+
+import (
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+)
+
+// VersionComparable enables kubernetes version retrieval.
+type VersionComparable interface {
```
chrisdoherty4, line 9: I think we should [stick with Go idioms](https://go.dev/doc/effective_go#interface-names) in this instance. I'm not sold on the use of `Comparable` because it really doesn't add anything to the type vs `Versionable`.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	Version() string
+}
+
+// MachineReplicaComparable enables replica number retrieval.
+type MachineReplicaComparable interface {
```
chrisdoherty4, line 14: Ditto on naming: `Replicable`

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	Replicas() int
+}
+
+// KubeadmControlPlaneComparator wraps the KubeadmControlPlane to enable Comparator functions.
+type KubeadmControlPlaneComparator struct {
```
chrisdoherty4, line 19: Similar to other comments, I don't think `Comparator` is adding any value. Its also a misrepresentation because this object doesn't provide any comparison behavior.

pkg/providers/tinkerbell/comparator.go
```diff
@@ -0,0 +1,51 @@
...
+	return k.Spec.Version
+}
+
+// Replicas retrieves KubeadmControlPlane replicas.
+func (k *KubeadmControlPlaneComparator) Replicas() int {
+	return int(*k.Spec.Replicas)
+}
+
+// MachineDeploymentComparator implements MachineReplicaComparable by decorating a MachineDeployment.
+type MachineDeploymentComparator struct {
```
chrisdoherty4, line 34: Ditto to the comment on use of `Comparator`.

chrisdoherty4 (changes requested):

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Updated naming to `Versionable`

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Updated naming to `Replicable`

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Was having a difficult time coming up with an appropriate name for this. Updated this to `KubeadmControlPlaneClusterInfoRetriever` does that work?

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: Updated to `MachineDeploymentClusterInfoRetreiver`, but definitely open to new names

raymond-zhang00: Renaming to `MachineDeployment`

pkg/providers/tinkerbell/comparator.go
raymond-zhang00: renaming to `KubeadmControlPlane`

chrisdoherty4 (commented):
Not sure what the final decision was around using this but all my comments we're resolved. Good job!

raymond-zhang00:
> Not sure what the final decision was around using this but all my comments we're resolved. Good job!

Thanks for the help Chris! I think the plan is Jia will use the comparator in [her PR](https://github.com/aws/eks-anywhere/pull/4971) and we can see how it looks in practice and we can decide to use or move away from it there.

PR #4995, opened 2023-02-16
Generate hardware yaml spec for writing during CLI workflows

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR generates a hardware yaml spec in preparation for writing, which is part of an effort to upload it to git for managing hardware via GitOps.

*Testing (if applicable):*

- Unit Testing
- Creating a management cluster

### DIFF
--- a/pkg/providers/nutanix/provider.go
+++ b/pkg/providers/nutanix/provider.go
@@ -496,6 +496,10 @@ func (p *Provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
 	return nil
 }
 
+func (p *Provider) HardwareSpec() []byte {
+	return nil
+}
+
 func (p *Provider) Version(clusterSpec *cluster.Spec) string {
 	return clusterSpec.VersionsBundle.Nutanix.Version
 }
--- a/pkg/providers/provider.go
+++ b/pkg/providers/provider.go
@@ -26,6 +26,7 @@ type Provider interface {
 	PostWorkloadInit(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 	BootstrapClusterOpts(clusterSpec *cluster.Spec) ([]bootstrapper.BootstrapClusterOption, error)
 	UpdateKubeConfig(content *[]byte, clusterName string) error
+	HardwareSpec() []byte
 	Version(clusterSpec *cluster.Spec) string
 	EnvMap(clusterSpec *cluster.Spec) (map[string]string, error)
 	GetDeployments() map[string][]string
--- a/pkg/providers/tinkerbell/create.go
+++ b/pkg/providers/tinkerbell/create.go
@@ -64,6 +64,9 @@ func (p *Provider) applyHardware(ctx context.Context, cluster *types.Cluster) er
 			return fmt.Errorf("waiting for baseboard management to be contactable: %v", err)
 		}
 	}
+	if err := p.generateHardwareSpec(ctx, cluster); err != nil {
+		return err
+	}
 	return nil
 }
 
--- a/pkg/providers/tinkerbell/hardware/catalogue_bmc.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_bmc.go
@@ -1,7 +1,7 @@
 package hardware
 
 import (
-	"github.com/tinkerbell/rufio/api/v1alpha1"
+	rufiov1alpha1 "github.com/tinkerbell/rufio/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -14,7 +14,7 @@ func (c *Catalogue) IndexBMCs(index string, fn KeyExtractorFunc) {
 }
 
 // InsertBMC inserts BMCs into the catalogue. If any indexes exist, the BMC is indexed.
-func (c *Catalogue) InsertBMC(bmc *v1alpha1.Machine) error {
+func (c *Catalogue) InsertBMC(bmc *rufiov1alpha1.Machine) error {
 	if err := c.bmcIndex.Insert(bmc); err != nil {
 		return err
 	}
@@ -23,23 +23,23 @@ func (c *Catalogue) InsertBMC(bmc *v1alpha1.Machine) error {
 }
 
 // AllBMCs retrieves a copy of the catalogued BMC instances.
-func (c *Catalogue) AllBMCs() []*v1alpha1.Machine {
-	bmcs := make([]*v1alpha1.Machine, len(c.bmcs))
+func (c *Catalogue) AllBMCs() []*rufiov1alpha1.Machine {
+	bmcs := make([]*rufiov1alpha1.Machine, len(c.bmcs))
 	copy(bmcs, c.bmcs)
 	return bmcs
 }
 
 // LookupBMC retrieves BMC instances on index with a key of key. Multiple BMCs _may_
 // have the same key hence it can return multiple BMCs.
-func (c *Catalogue) LookupBMC(index, key string) ([]*v1alpha1.Machine, error) {
+func (c *Catalogue) LookupBMC(index, key string) ([]*rufiov1alpha1.Machine, error) {
 	untyped, err := c.bmcIndex.Lookup(index, key)
 	if err != nil {
 		return nil, err
 	}
 
-	bmcs := make([]*v1alpha1.Machine, len(untyped))
+	bmcs := make([]*rufiov1alpha1.Machine, len(untyped))
 	for i, v := range untyped {
-		bmcs[i] = v.(*v1alpha1.Machine)
+		bmcs[i] = v.(*rufiov1alpha1.Machine)
 	}
 
 	return bmcs, nil
@@ -56,7 +56,7 @@ const BMCNameIndex = ".ObjectMeta.Name"
 func WithBMCNameIndex() CatalogueOption {
 	return func(c *Catalogue) {
 		c.IndexBMCs(BMCNameIndex, func(o interface{}) string {
-			bmc := o.(*v1alpha1.Machine)
+			bmc := o.(*rufiov1alpha1.Machine)
 			return bmc.ObjectMeta.Name
 		})
 	}
@@ -83,18 +83,18 @@ func (w *BMCCatalogueWriter) Write(m Machine) error {
 	return nil
 }
 
-func toRufioMachine(m Machine) *v1alpha1.Machine {
+func toRufioMachine(m Machine) *rufiov1alpha1.Machine {
 	// TODO(chrisdoherty4)
 	// 	- Set the namespace to the CAPT namespace.
 	// 	- Patch through insecure TLS.
-	return &v1alpha1.Machine{
+	return &rufiov1alpha1.Machine{
 		TypeMeta: newMachineTypeMeta(),
 		ObjectMeta: v1.ObjectMeta{
 			Name:      formatBMCRef(m),
 			Namespace: constants.EksaSystemNamespace,
 		},
-		Spec: v1alpha1.MachineSpec{
-			Connection: v1alpha1.Connection{
+		Spec: rufiov1alpha1.MachineSpec{
+			Connection: rufiov1alpha1.Connection{
 				Host: m.BMCIPAddress,
 				AuthSecretRef: corev1.SecretReference{
 					Name:      formatBMCSecretRef(m),
--- a/pkg/providers/tinkerbell/hardware/machine.go
+++ b/pkg/providers/tinkerbell/hardware/machine.go
@@ -4,6 +4,11 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/collection"
+	rufiov1alpha1 "github.com/tinkerbell/rufio/api/v1alpha1"
+	tinkv1alpha1 "github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // Machine is a machine configuration with optional BMC interface configuration.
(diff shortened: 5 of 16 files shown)


pkg/providers/provider.go
```diff
@@ -26,6 +26,7 @@ type Provider interface {
 	PostWorkloadInit(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 	BootstrapClusterOpts(clusterSpec *cluster.Spec) ([]bootstrapper.BootstrapClusterOption, error)
 	UpdateKubeConfig(content *[]byte, clusterName string) error
+	HardwareSpec() []byte
```
ddjjia, line 29: How's `HardwareSpec()` get called? `GetRufioMachine` does't seem use it

cxbrowne1207: So, I've split the task into two PRs and this is the first, so it's not used here. The CLI workflows has access to an implementer of the `Provider` interface, and the `HardwareSpec` method will be called to retrieve the hardware spec to upload the file to git during the appropriate step.

vivek-koppuru: If this returns a `[]byte`, do we need to reevaluate how we use a method like `MachineConfigs` and generalize all to a `GetProviderSpec` for marshaling purposes and return `[]byte`? This way we can avoid introducing a new function every time we are dealing with a new type. But if it is too much work, I think the refactoring project @chrisdoherty4 is leading might take care of it.

pkg/providers/tinkerbell/hardware/catalogue_bmc.go
```diff
@@ -1,7 +1,7 @@
 package hardware
 
 import (
-	"github.com/tinkerbell/rufio/api/v1alpha1"
+	rufiov1alpha1 "github.com/tinkerbell/rufio/api/v1alpha1"
```
vivek-koppuru, line 4: Is this needed if we don't reference our own spec in the file?

pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -29,6 +35,39 @@ type Machine struct {
...
+	labels := make(map[string]string, len(hw.Labels)-len(excludedLabels))
+	for key, value := range hw.Labels {
+		if !excludedLabels.Contains(key) {
+			labels[key] = value
+		}
+	}
+
+	return Machine{
+		Hostname:     hw.Name,
+		IPAddress:    hw.Spec.Interfaces[0].DHCP.IP.Address,
```
vivek-koppuru, line 57: So it is safe to only assume the first index? This length is validated beforehand?

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -286,3 +291,66 @@ func (p *Provider) ChangeDiff(currentSpec, newSpec *cluster.Spec) *types.Compone
...
 }
+
+// HardwareSpec returns the hardware yaml containing configurations for the hardware in the cluster.
+func (p *Provider) HardwareSpec() []byte {
+	return p.hardwareSpec
+}
+
+// generateHardwareSpec reads the hardware information from the available cluster, generates a yaml that can be submitted to a kubernetes cluster
+// and caches it on the Provider.
+func (p *Provider) generateHardwareSpec(ctx context.Context, cluster *types.Cluster) error {
```
vivek-koppuru, line 302: I think we should return the spec and then cache it in the caller of this function.

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -286,3 +291,66 @@ func (p *Provider) ChangeDiff(currentSpec, newSpec *cluster.Spec) *types.Compone
...
+	if err != nil {
+		return err
+	}
+	p.hardwareSpec = hardwareSpec
+
+	return nil
+}
+
+// readMachineCatalogueFromCluster fetches all the hardware, bmc and related secret objects and inserts in to ETCDReader catalogue.
+func (p *Provider) readMachineCatalogueFromCluster(ctx context.Context, cluster *types.Cluster, catalogue *hardware.Catalogue) error {
```
vivek-koppuru, line 323: Same here, not possible to return value and set it in the caller?

pkg/providers/tinkerbell/hardware/catalogue_bmc.go
cxbrowne1207: It's not needed, but I think this is what we should do since we have multiple `v1alpha1` references in our codebase. It makes it easier to tell which one is being used at a glance

pkg/providers/tinkerbell/hardware/machine.go
cxbrowne1207: Yes. When hardware is created, these values are usually loaded from the CSV file, and the `Interfaces` slice on the `Hardware` is always loaded with one entry.

pkg/providers/provider.go
cxbrowne1207: Yeah, I think that'd be too much effort for now. I did notice the "new workflows" that use hooks, which seems to be a good solution for this but it's currently feature gated.

pkg/providers/nutanix/provider.go
```diff
@@ -496,6 +496,11 @@ func (p *Provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
 	return nil
 }
 
+// HardwareSpec returns the hardware yaml containing configurations for the hardware in the cluster.
+func (p *Provider) HardwareSpec() []byte {
```
abhinavmpandey08, line 500: `HardwareSpec()` doesn't make much sense for any providers besides tinkerbell. Is there a different way we can get the hardware spec without adding this func to the provider interface?

pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -29,6 +35,39 @@ type Machine struct {
 	VLANID       string `csv:"vlan_id, omitempty"`
 }
 
+// NewMachineFromHardware creates a new Machine from the given hardware and it's dependencies, if any.
+func NewMachineFromHardware(hw tinkv1alpha1.Hardware, rm *rufiov1alpha1.Machine, s *corev1.Secret) Machine {
+	var bmcIPAddress, bmcUsername, bmcPassword string
+	if rm != nil && s != nil {
+		bmcIPAddress = rm.Spec.Connection.Host
+		bmcUsername = string(s.Data["username"])
+		bmcPassword = string(s.Data["password"])
```
abhinavmpandey08, line 44: I think these are base64 encoded. Does the Machine struct below expect these to be in base64 format?

pkg/providers/tinkerbell/create.go
```diff
@@ -45,7 +45,15 @@ func (p *Provider) PreCAPIInstallOnBootstrap(ctx context.Context, cluster *types
 }
 
 func (p *Provider) PostBootstrapSetup(ctx context.Context, clusterConfig *v1alpha1.Cluster, cluster *types.Cluster) error {
-	return p.applyHardware(ctx, cluster)
+	if err := p.applyHardware(ctx, cluster); err != nil {
+		return err
+	}
+	hardwareSpec, err := p.generateHardwareSpec(ctx, cluster)
```
abhinavmpandey08, line 51: I think `p.applyHardware()` also generates the hardware spec since it has to apply it. Can we consolidate these two calls somehow so we are not generating the hardware spec twice?

pkg/providers/tinkerbell/create.go
```diff
@@ -45,7 +45,15 @@ func (p *Provider) PreCAPIInstallOnBootstrap(ctx context.Context, cluster *types
...
+	if err != nil {
+		return fmt.Errorf("generating hardware spec: %v", err)
+	}
+	p.hardwareSpec = hardwareSpec
```
abhinavmpandey08, line 55: What's the motivation behind storing the hardwareSpec? Will this be used in the controller to get available hardware or something for workload clusters?

cxbrowne1207: We want to enable hardware management via GitOps and consequently upload the hardware YAML to the git repo. This PR is the first part of that initiative, and a subsequent PR will use this to upload the file to GitOps. Check the draft [here](https://github.com/aws/eks-anywhere/pull/4906) for details.

I've updated the description to include the motive.

pkg/providers/tinkerbell/create.go
cxbrowne1207: The spec that is generated for `p.applyHardware()` is partial and deals with the hardware fed from CSV file (or in general added hardware in the case of upgrade I think). The hardware yaml we want to generate must include all of the hardware in the management cluster  and the related objects. We probably don't want `p.applyHardware` to try to apply existing hardware, hence this approach. So I'm not sure if we can merge them. and

pkg/providers/nutanix/provider.go
cxbrowne1207: I figured it would be an okay approach, since we've done something similar with `UpdateKubeConfig`.

This is how I planned on using it. in the CLI flow:
```
func (s *InstallGitOpsManagerTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
...
	err := commandContext.GitOpsManager.InstallGitOps(..., ..., ..., ..., ..., commandContext.Provider.HardwareSpec())
...
}
```
Another approach is to have it only on the Tinkerbell Provider type and then use a type assertion to access it from the CLI flow when needed if that's preferred?

pkg/providers/tinkerbell/hardware/machine.go
cxbrowne1207: The machine struct expects the unencoded data, and `Data` is a map[string][]byte so I think `s.Data["username"]` returns the expected output here. Through manual testing, I see that the output YAML is generated as expected

pkg/providers/provider.go
cxbrowne1207: Discussed with @abhinavmpandey08. Will change `HardwareSpec()` -> `AdditionalFiles()`, so it makes more sense to be on the `Provider` type for now

pkg/providers/tinkerbell/create.go
panktishah26: As a `PostBootstrapSetup` process, we apply hardwares by marshaling the catalogue object. Could you elaborate a little about what is the main purpose of generateHardwareSpec and buildHardwareMachineFromCluster functions?

abhinavmpandey08 (commented):
looks good!

abhinavmpandey08 (approved):

abhinavmpandey08 (approved):

pkg/providers/provider.go
```diff
@@ -26,6 +26,7 @@ type Provider interface {
 	PostWorkloadInit(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 	BootstrapClusterOpts(clusterSpec *cluster.Spec) ([]bootstrapper.BootstrapClusterOption, error)
 	UpdateKubeConfig(content *[]byte, clusterName string) error
+	AdditionalFiles() map[string][]byte
```
g-gaston, line 29: What is this being used for? Sorry I can't find it
I'm just concerned about adding even more stuff to this interface, specially when is only used by one provider

pkg/providers/tinkerbell/hardware/catalogue_bmc.go
g-gaston: pls use `rufiov1` :)

g-gaston:
Could you elaborate a bit on exactly when this file is generated and where is being written?
For the readers without much context like me :)

pkg/providers/provider.go
cxbrowne1207: Sorry, the usage isn't in this PR. I split it off to keep the PR focused on generating the hardware yaml. 

For context, this was originally`HardwareSpec()`, but we decided to make it AdditionalFiles instead so it would make more sense being on the provider interface. It is going to be used to retrieve the generated hardwareSpec in the CLI workflow during to upload it to GitOps like so:

```
func (s *InstallGitOpsManagerTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
...
        var hardwareSpec []byte
        if commandContext.Provider.AdditionalFiles != nil {
            hardwareSpec = commandContext.Provider.AdditionalFiles["hardware"]
        }
        
	err := commandContext.GitOpsManager.InstallGitOps(..., ..., ..., ..., ..., hardwareSpec)
...
}
```

cxbrowne1207:
/test all

> Could you elaborate a bit on exactly when this file is generated and where is being written? For the readers without much context like me :)

I've updated the PR description, let me know if any more explanation is needed

pkg/providers/tinkerbell/hardware/machine.go
cxbrowne1207: When the hardware is generated from csv initially, it is validated to have the required columns. Then the data is marshaled into the following struct [here](https://pkg.go.dev/github.com/aws/eks-anywhere/pkg/providers/tinkerbell/hardware#Machine)

Then `hardware.Machine` is used to build `tinkv1alpha1.Hardware` that can be applied to the cluster [here](https://github.com/aws/eks-anywhere/blob/v0.14.2/pkg/providers/tinkerbell/hardware/catalogue_hardware.go#LL250C6-L250C25) in a way where we can safely assume the first index.

Does that answer your question?

vivek-koppuru: Yea just want to make sure that there isn't a situation now or in the future where we have this index a non-existent entry, so was wondering if we need to add a simple check for that.

cxbrowne1207: Updated with some checks. Also renamed this method to `MachineFromHardware`

cxbrowne1207:
\hold

pausing review for further discussion

cxbrowne1207:
/hold

Need to have some further discussion on this.

chrisdoherty4:
Is this PR still active? Seems like it can be closed based on discussions to discontinue hardware management efforts - for now at least.

PR #4997, opened 2023-02-16
Add airgap assertion to vsphere e2e test

d8660091:
*Issue #, if available:*

*Description of changes:* Iptables works like magic, we need a check to make sure that the admin machine's network is really airgapped.

*Testing (if applicable):* Manually invoked the e2e test against just the newly added check. And the new added check was working as expected


jonathanmeier5 (approved):
nice 👌
/lgtm
/approve

PR #5007, opened 2023-02-17
Add region support to packages reader

TerryHowe:


vincentni (approved):
/lgtm
/approve

ewollesen (commented):
I wonder if it'd be much work to read the AWS_REGION env var for this information. It seems like that's more what the customer would expect, at least if the CLI flag isn't specified.

TerryHowe:
> I wonder if it'd be much work to read the AWS_REGION env var for this information? It seems like that's more what the customer would expect, at least if the CLI flag isn't specified.

If the parameter is not set, it defaults to the value of `EKSA_AWS_REGION` which is what the CLI generally uses. I thought about support AWS_REGION, but that would be inconsistent with the CLI otherwise I think.

PR #5013, opened 2023-02-18
updating phase runner to accept generic objects

raymond-zhang00:
*Issue #, if available:*

*Description of changes:*
Tinkerbell upgrade validations require comparisons between the desired Cluster Spec state and the existing current state pulled from the CAPI objects at multiple phases of the reconciler. Caching the result of our comparisons and/or the desired CAPI state earlier on will allow us to reference these objects in later phases without repeatedly retrieving the CAPI state and running our comparisons.

The phase runner currently runs off the Cluster Spec object. We propose allowing the object type to be specified as needed to allow custom fields and values outside of Cluster Spec to be stored, referenced and modified in different phases - while being constrained to the scope of the `reconciler.Reconcile()` function for thread safety.

This update enables phase runner to accept a specifiable type instead of the previous `*cluster.Spec` type as well as introduces the `Scope` struct for Tinkerbell to store our additional information

*Testing (if applicable):*
Existing automated unit tests

### DIFF
--- a/pkg/controller/runner.go
+++ b/pkg/controller/runner.go
@@ -4,34 +4,32 @@ import (
 	"context"
 
 	"github.com/go-logr/logr"
-
-	"github.com/aws/eks-anywhere/pkg/cluster"
 )
 
 // Phase represents a generic reconciliation phase for a cluster spec.
-type Phase func(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec) (Result, error)
+type Phase[O any] func(ctx context.Context, log logr.Logger, obj O) (Result, error)
 
 // PhaseRunner allows to execute Phases in order.
-type PhaseRunner struct {
-	phases []Phase
+type PhaseRunner[O any] struct {
+	phases []Phase[O]
 }
 
 // NewPhaseRunner creates a new PhaseRunner without any Phases.
-func NewPhaseRunner() PhaseRunner {
-	return PhaseRunner{}
+func NewPhaseRunner[O any]() PhaseRunner[O] {
+	return PhaseRunner[O]{}
 }
 
 // Register adds a phase to the runnner.
-func (r PhaseRunner) Register(phases ...Phase) PhaseRunner {
+func (r PhaseRunner[O]) Register(phases ...Phase[O]) PhaseRunner[O] {
 	r.phases = append(r.phases, phases...)
 	return r
 }
 
 // Run will execute phases in the order they were registered until a phase
 // returns an error or a Result that requests to an interruption.
-func (r PhaseRunner) Run(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec) (Result, error) {
+func (r PhaseRunner[O]) Run(ctx context.Context, log logr.Logger, obj O) (Result, error) {
 	for _, p := range r.phases {
-		if r, err := p(ctx, log, clusterSpec); r.Return() {
+		if r, err := p(ctx, log, obj); r.Return() {
 			return r, nil
 		} else if err != nil {
 			return Result{}, err
--- a/pkg/providers/tinkerbell/reconciler/reconciler.go
+++ b/pkg/providers/tinkerbell/reconciler/reconciler.go
@@ -163,7 +163,7 @@ func (r *Reconciler) ValidateDatacenterConfig(ctx context.Context, log logr.Logg
 	clusterSpec := tinkerbellScope.ClusterSpec
 	log = log.WithValues("phase", "validateDatacenterConfig")
 
-	if err := r.validateTinkerbellIPMatch(ctx, NewScope(clusterSpec)); err != nil {
+	if err := r.validateTinkerbellIPMatch(ctx, clusterSpec); err != nil {
 		log.Error(err, "Invalid TinkerbellDatacenterConfig")
 		failureMessage := err.Error()
 		clusterSpec.Cluster.Status.FailureMessage = &failureMessage
@@ -186,8 +186,7 @@ func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, tinkerbe
 	return r.cniReconciler.Reconcile(ctx, log, client, clusterSpec)
 }
 
-func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, tinkerbellScope *Scope) error {
-	clusterSpec := tinkerbellScope.ClusterSpec
+func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, clusterSpec *c.Spec) error {
 	if clusterSpec.Cluster.IsManaged() {
 
 		// for workload cluster tinkerbell IP must match management cluster tinkerbell IP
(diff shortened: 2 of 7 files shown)


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -31,6 +31,18 @@ type IPValidator interface {
 	ValidateControlPlaneIP(ctx context.Context, log logr.Logger, spec *c.Spec) (controller.Result, error)
 }
 
+// Scope object for Tinkerbell reconciler.
+type Scope struct {
```
ddjjia, line 35: tinkerbell scope need a operation bool at least

ddjjia:
This is part of Tinkerbell Upgrade process as a generic phase runner allows to pass object between phases. That way CAPI and operation detected at early phase could be used at later reconcile phase.

pkg/controller/runner.go
```diff
@@ -4,34 +4,32 @@ import (
...
 
 // Phase represents a generic reconciliation phase for a cluster spec.
-type Phase func(ctx context.Context, log logr.Logger, clusterSpec *cluster.Spec) (Result, error)
+type Phase[O any] func(ctx context.Context, log logr.Logger, obj O) (Result, error)
 
 // PhaseRunner allows to execute Phases in order.
-type PhaseRunner struct {
-	phases []Phase
+type PhaseRunner[O any] struct {
+	phases []Phase[O]
```
cxbrowne1207, line 14: I'm not opposed. Have we considered another approach where we simply use add state to the PhaseRunner struct?

E.g.

```
type PhaseRunner struct {
  phases []Phase
  isRollingUpdate: bool
}
```

ddjjia: This would add isRollingUpdate to phase runner in all providers, but it's just tinkerbell specific

raymond-zhang00: This may be good to keep as an alternative, I think Chris saw some code smells with how I'm using the generic phase runner. I ended up still wrapping the existing IPValidatorPhase which takes in clusterSpec as an argument with a Tinkerbell phase just so I could run it, which might be moving backwards. 

Maybe we an additional field called additionalDependencies or something could do what you're suggesting.

```
type PhaseRunner struct {
  phases []Phase
 additionalDependencies  any
}

type TinkerbellAdditionalDeps struct {
  isRollingUpdate  bool
  controlPlane   CP
  workerNodes  W
}
```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -31,6 +31,20 @@ type IPValidator interface {
...
+// NewScope creates a new Tinkerbell Reconciler Scope.
+func NewScope(clusterSpec *c.Spec) *Scope {
+	return &Scope{
+		ClusterSpec:      clusterSpec,
+		isRollingUpgrade: false,
+	}
+}
+
 // Reconciler for Tinkerbell.
 type Reconciler struct {
```
taneyland, line 49: I could be missing something, but why can't we just add the `isRollingUpgrade` bool field to this Tinkerbell Reconciler struct? 

If we are computing isRollout and the desired CAPI as one of the early phases (which seems to be the plan), then that value should persist across phases since we are using the same Reconciler `r` for every phase

raymond-zhang00: Guillermo was able to catch that as an issue in a previous PR I was working on. There were issues with thread safety and introducing state to the controller
https://github.com/aws/eks-anywhere/pull/4994/files#r1110388725

pkg/controller/runner.go
cxbrowne1207: Yeah, I realize it's not exactly provider specific, but the generic does seem like an overkill solution.

@raymond-zhang00 Yeah. I would even make that a `map[string]interface{}`, if we decide to go with this as an alternative.

pkg/providers/tinkerbell/reconciler/reconciler.go
taneyland: Ok good to know! That makes sense. Thanks!

pkg/controller/runner.go
raymond-zhang00: Ended up having a discussing the code smell with Chris and utilizing a generic phase runner should be a solid approach so I don't think we'll need to go down this path.

cxbrowne1207: Okay, gotcha!

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -60,20 +74,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *an
...
+		r.ValidateControlPlaneIP,
 		r.ValidateClusterSpec,
 		r.ValidateHardware,
 		r.ValidateDatacenterConfig,
 		r.ReconcileControlPlane,
 		r.CheckControlPlaneReady,
 		r.ReconcileCNI,
 		r.ReconcileWorkers,
-	).Run(ctx, log, clusterSpec)
+	).Run(ctx, log, NewScope(clusterSpec))
```
ddjjia, line 86: We should init a new scope before phase register so the values in the scope can be updated thru phases. This way `isRollingUpgrade` will be reset to false

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -116,15 +138,16 @@ func (r *Reconciler) ReconcileWorkerNodes(ctx context.Context, log logr.Logger,
 		return controller.Result{}, errors.Wrap(err, "building cluster Spec for worker node reconcile")
 	}
 
-	return controller.NewPhaseRunner().Register(
+	return controller.NewPhaseRunner[*Scope]().Register(
 		r.ValidateClusterSpec,
 		r.ValidateHardware,
 		r.ReconcileWorkers,
-	).Run(ctx, log, clusterSpec)
+	).Run(ctx, log, NewScope(clusterSpec))
```
ddjjia, line 145: same here. need to init scope before phase register

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -161,7 +186,8 @@ func (r *Reconciler) ReconcileCNI(ctx context.Context, log logr.Logger, clusterS
 	return r.cniReconciler.Reconcile(ctx, log, client, clusterSpec)
 }
 
-func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, clusterSpec *c.Spec) error {
+func (r *Reconciler) validateTinkerbellIPMatch(ctx context.Context, tinkerbellScope *Scope) error {
```
ddjjia, line 189: why is this change needed?

raymond-zhang00: You're right, this doesn't need to be changed since it's not a phase. Changing it back

pkg/controller/runner.go
vivek-koppuru: Was this based on a design that @g-gaston shared recently?

pkg/providers/tinkerbell/reconciler/reconciler.go
vivek-koppuru: Why is it called scope?

pkg/controller/runner.go
raymond-zhang00: Yes, we discussed that we wanted an object to hold data from phase to phase that's thread safe but didn't want to add on to ClusterSpec.

PR #5031, opened 2023-02-21
Implement packages cluster spec

TerryHowe:


cmd/eksctl-anywhere/cmd/installpackagecontroller.go
```diff
@@ -72,7 +72,7 @@ func installPackageController(ctx context.Context) error {
 	}
 
 	curatedpackages.PrintLicense()
-	err = ctrlClient.EnableCuratedPackages(ctx)
+	err = ctrlClient.EnableCuratedPackages(ctx, &clusterSpec.Cluster.Spec)
```
a-cool-train, line 75: I think injecting this into ctrlClient will make this so much easier.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -51,6 +52,54 @@ func GetVersionBundle(reader Reader, eksaVersion string, spec *v1alpha1.Cluster)
...
+
+// GetPackageControllerConfiguration returns the default kubernetes version for a Cluster.
+func GetPackageControllerConfiguration(clusterSpec *v1alpha1.ClusterSpec) (result string, err error) {
+	if clusterSpec == nil {
+		return "", nil
+	}
+
+	if clusterSpec.Packages.Controller != nil {
+		result += "controller:\n"
+		result += formatYamlLine("  ", "digest", clusterSpec.Packages.Controller.Digest)
```
a-cool-train, line 70: Is the goal to return a yaml formatted controller config? Maybe yaml marshaller will come in handy here?

cmd/eksctl-anywhere/cmd/installpackagecontroller.go
TerryHowe: Done.

pkg/curatedpackages/curatedpackages.go
TerryHowe: I tried taht, but this approach was much easier to understand and test.

pkg/curatedpackages/curatedpackages.go
```diff
@@ -51,6 +52,74 @@ func GetVersionBundle(reader Reader, eksaVersion string, spec *v1alpha1.Cluster)
...
+		resources.Requests.CPU != "" || resources.Requests.Memory != "" {
+		result += "  resources:\n"
+		result += formatImageResource(&resources.Limits, "limits")
+		result += formatImageResource(&resources.Requests, "requests")
+	}
+	return result
+}
+
+// GetPackageControllerConfiguration returns the default kubernetes version for a Cluster.
+func GetPackageControllerConfiguration(clusterSpec *v1alpha1.ClusterSpec) (result string, err error) {
```
a-cool-train, line 93: I would recommend moving this method to pkgControllerClient since functionally this method is dealing with a packagecontroller.

pkg/curatedpackages/curatedpackages.go
a-cool-train: I have used the yaml marshall's functionality and it was easy to understand and also test. We can sync up offline.

pkg/curatedpackages/curatedpackages.go
TerryHowe: Done.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -335,3 +348,71 @@ func WithValuesFileWriter(writer filewriter.FileWriter) func(client *PackageCont
...
+		resources.Requests.CPU != "" || resources.Requests.Memory != "" {
+		result += "  resources:\n"
+		result += formatImageResource(&resources.Limits, "limits")
+		result += formatImageResource(&resources.Requests, "requests")
+	}
+	return result
+}
+
+// GetPackageControllerConfiguration returns the default kubernetes version for a Cluster.
+func GetPackageControllerConfiguration(clusterSpec *v1alpha1.ClusterSpec) (result string, err error) {
```
a-cool-train, line 390: Make this a method and then u won't need the clusterSpec parameter.

TerryHowe: I think in the long road, PCC is not the place for this. Refactoring a solution between PackageInstaller and PCC is out of scope for this effort and I'd sooner make as few changes here to minimize the chances of breaking something right now.

a-cool-train: I don't understand why a refactor is needed to make this change?

TerryHowe: This could be added to the PCC easy enough, but I don't think that is the right place for it. There should be a new class created that combines the some of the logic of PCC and PackageInstaller. I don't want to start ripping all this apart right now.

TerryHowe: nvm, I moved it over there

pkg/curatedpackages/packagecontrollerclient_test.go
```diff
@@ -149,7 +161,7 @@ func newPackageControllerTests(t *testing.T) []*packageControllerTest {
 			kubectl:        k,
 			chartInstaller: ci,
 			command: curatedpackages.NewPackageControllerClient(
-				ci, k, clusterName, kubeConfig, chartDev,
+				ci, k, clusterName, nil, kubeConfig, chartDev,
```
a-cool-train, line 164: From usage, looks like clusterSpec is not used by all. Should we make it an option in PCC?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -274,6 +287,75 @@ func (pc *PackageControllerClient) IsInstalled(ctx context.Context) bool {
...
+
+// GetPackageControllerConfiguration returns the default kubernetes version for a Cluster.
+func (pc *PackageControllerClient) GetPackageControllerConfiguration() (result string, err error) {
+	clusterSpec := pc.clusterSpec
+	if clusterSpec == nil || clusterSpec.Packages == nil {
+		return "", nil
+	}
+
+	if clusterSpec.Packages.Controller != nil {
+		result += "controller:\n"
```
a-cool-train, line 335: I am still not comfortable with custom parsing instead of using the classic json marshaller.

pkg/curatedpackages/packagecontrollerclient_test.go
TerryHowe: it is only used in the generate yaml method. I don't want to create one for tests that don't use it.

a-cool-train: What I am referring to is in the NewPackageControllerClient should we make the clusterSpec a PackageControllerClientOpt so that you don't need to pass the nil everywhere.

TerryHowe: oooh

PR #5035, opened 2023-02-22
Add watches for CloudStack resources to controller

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/config/manifest/eksa-components.yaml
+++ b/config/manifest/eksa-components.yaml
@@ -3806,6 +3806,75 @@ spec:
                   name:
                     type: string
                 type: object
+              packages:
+                description: PackageConfiguration for installing EKS Anywhere curated
+                  packages.
+                properties:
+                  controller:
+                    description: Controller package controller configuration
+                    properties:
+                      digest:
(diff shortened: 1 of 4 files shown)


cxbrowne1207 (approved):

config/manifest/eksa-components.yaml
```diff
@@ -3806,6 +3806,75 @@ spec:
                   name:
                     type: string
                 type: object
+              packages:
```
cxbrowne1207, line 3809: Looks good. qq: Why do we need this?

cxbrowne1207, line 3809: qq: Why do we need this?

cxbrowne1207 (commented):
Looks good. Left a quick question

config/manifest/eksa-components.yaml
taneyland: This was added when I ran `make generate-manifests`, so it must've been added recently and we just never generated the new manifest then

PR #5036, opened 2023-02-22
Decouple a unit test from its environment

ewollesen:
Having this variable set would cause test failures, as the "noflags" variant of the test would assume that the value was unset.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/eksa_installer_test.go
+++ b/pkg/clustermanager/eksa_installer_test.go
@@ -2,6 +2,7 @@ package clustermanager_test
 
 import (
 	"context"
+	"os"
 	"testing"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


pkg/clustermanager/eksa_installer_test.go
```diff
@@ -252,6 +253,7 @@ func TestSetManagerFlags(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			os.Unsetenv(features.FullLifecycleAPIEnvVar)
```
g-gaston, line 256: I'm assuming this is to prevent the current env from where we run the test to affect the results of the test.

If that's the case, (nit) can we move this line to the beginning of the test function? That removes ambiguity about this line trying to cleanup between sub-tests vs setting up the environment appropriately before the whole test.

g-gaston (approved):

pkg/clustermanager/eksa_installer_test.go
ewollesen: I put it here because it should be checked and fixed on a per-test basis. Each test should ensure the environment is pristine, regardless of what other tests do. That property is lost if its moved to the beginning of the test function.

PR #5037, opened 2023-02-22
Cleanup cluster.Spec

g-gaston:
## Description of changes

This addresses one of the standing problems with `cluster.Spec`'s scope: it acts both as a data object and a "manifest" reader (understand manifest as in a yaml document read from a local file or HTTP). Being the goal to make the Spec a pure data object, we are removing these capabilities while maintaining the same overall program functionality:

* The `Spec`'s ability to read manifests. This is delegated to a file reader that any caller in need of manifests will need to accept as a dependency.
* The `Spec`'s ability to "build itself". Although the construction was invoked through a constructor, the `Spec`'s ability to read manifests was use to build itself. This made the problem even worse, since the CLI configuration to build the Spec (Bundles URL, etc.) bled in to the `Spec`'s internals. Now this is all done through a `FileSpecBuilder`. For now this lives in the `cluster` package. Once the `cli` package is available, that would probably be a way better place for it since this builder logic is exclusive to the CLI flows.

## Testing

Ran a full flow e2e test with docker.


cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -110,31 +116,28 @@ func readAndValidateClusterSpec(clusterConfigPath string, cliVersion version.Inf
...
 		return nil, fmt.Errorf("unable to get cluster config from file: %v", err)
 	}
 
 	if clusterSpec.Cluster.IsManaged() && options.managementKubeconfig == "" {
 		options.managementKubeconfig = kubeconfig.FromEnvironment()
-		clusterSpec.ManagementCluster = &types.Cluster{
-			Name:               clusterSpec.Cluster.ManagedBy(),
-			KubeconfigFile:     options.managementKubeconfig,
-			ExistingManagement: true,
+	}
```
jiayiwang7, line 131: what if it's a workload cluster spec and management kubeconfig is empty? shouldn't we throw an error?

pkg/cluster/builder.go
```diff
@@ -0,0 +1,119 @@
...
+// manifest is skipped.
+func WithOverrideBundlesManifest(url string) FileSpecBuilderOpt {
+	return func(b *FileSpecBuilder) {
+		b.bundlesManifestURL = url
+	}
+}
+
+// NewFileSpecBuilder builds a new [FileSpecBuilder].
+// cliVersion is used to chose the right Bundles from the the Release manifest.
+func NewFileSpecBuilder(reader manifests.FileReader, cliVersion version.Info, opts ...FileSpecBuilderOpt) FileSpecBuilder {
```
jiayiwang7, line 43: whats the reason to return `FileSpecBuilder` instead of `*FileSpecBuilder` here?

jiayiwang7 (approved):

pkg/cluster/builder.go
g-gaston: It doesn't need to be a pointer because it doesn't need to modify itself after creation. Also, the amount of data it stores doesn't require passing it by reference, the copy shouldn't be a performance problem.

I usually go back to this when in doubt of what to use: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

cmd/eksctl-anywhere/cmd/options.go
g-gaston: I didn't change the logic here (or I didn't intend to)
That might be a validation we do somewhere else or one we don't do at all?

jiayiwang7: i don't think we do it at all

PR #5064, opened 2023-02-24
Move cloudstackmachineconfig validations to API

taneyland:
*Issue #, if available:*

*Description of changes:*
Moving machineconfig validations to api to use in both CLI workflow and when managing clusters using cluster full lifecycle api

*Testing (if applicable):*


pkg/providers/cloudstack/validator.go
```diff
@@ -176,35 +144,15 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, cloudStac
...
 
 	for _, machineConfig := range cloudStackClusterSpec.machineConfigsLookup {
-		if machineConfig.Namespace != cloudStackClusterSpec.Cluster.Namespace {
-			return fmt.Errorf(
-				"CloudStackMachineConfig %s and Cluster objects must have the same namespace: CloudStackMachineConfig namespace=%s; Cluster namespace=%s",
-				machineConfig.Name,
-				machineConfig.Namespace,
-				cloudStackClusterSpec.Cluster.Namespace,
-			)
-		}
```
taneyland, line 186: Removing ip, namespace & machineGroupRef checks because that is handled in the config manager validations for the cluster and cloudstackmachineconfig entries: https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/default_manager.go#L17

cxbrowne1207 (approved):
/lgtm

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -203,7 +203,7 @@ func (c *CloudStackMachineConfig) GetName() string {
 }
 
 func (c *CloudStackMachineConfig) Validate() error {
```
ddjjia, line 205: Tinkkerbell provider needs to leverage this so the logic won't be missed in provider.

PR #5067, opened 2023-02-24
Add validation for HookImagesURLPath for proxy configuration

panktishah26:
*Description of changes:*
Add validation for HookImagesURLPath if proxy configuration is specified in cluster config file. Hook image need to be hosted locally and should be accessible without using proxy if proxy is specified. 

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/assert.go
+++ b/pkg/providers/tinkerbell/assert.go
@@ -3,6 +3,7 @@ package tinkerbell
 import (
 	"errors"
 	"fmt"
+	"net/http"
 
 	tinkerbellv1 "github.com/tinkerbell/cluster-api-provider-tinkerbell/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
@@ -119,6 +120,36 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
 	if tinkerbellIP == controlPlaneIP {
 		return fmt.Errorf("controlPlaneConfiguration.endpoint.host and tinkerbellIP are the same (%s), please provide two unique IPs", tinkerbellIP)
 	}
+	return nil
+}
+
+// AssertHookRetrievableWithoutProxy ensures the executing machine can retrieve Hook
+// from the host URL without a proxy configured. It does not guarantee the target node
+// will be able to download Hook.
+func AssertHookRetrievableWithoutProxy(spec *ClusterSpec) error {
+	if spec.Cluster.Spec.ProxyConfiguration == nil {
+		return nil
+	}
+
+	// return an error if hookImagesURLPath field is not specified for during Proxy configuration.
+	if spec.DatacenterConfig.Spec.HookImagesURLPath == "" {
+		return fmt.Errorf("locally hosted hookImagesURLPath is required to support ProxyConfiguration")
+	}
+
+	// verify hookImagesURLPath is accessible locally too
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.Proxy = nil
+	client := &http.Client{
+		Transport: transport,
+	}
+
+	resp, err := client.Get(spec.DatacenterConfig.Spec.HookImagesURLPath)
+	if err != nil {
+		return fmt.Errorf("HookImagesURLPath: %s needs to be hosted locally while specifiying Proxy configuration: %v", spec.DatacenterConfig.Spec.HookImagesURLPath, err)
+	}
+
+	defer resp.Body.Close()
+
 	return nil
 }
 
--- a/pkg/providers/tinkerbell/assert_test.go
+++ b/pkg/providers/tinkerbell/assert_test.go
@@ -286,6 +286,44 @@ func TestAssertPortsNotInUse_Fails(t *testing.T) {
 	g.Expect(assertion(clusterSpec)).ToNot(gomega.Succeed())
 }
 
+func TestAssertAssertHookImageURLProxyNonAirgappedURLSuccess(t *testing.T) {
+	g := gomega.NewWithT(t)
+
+	clusterSpec := NewDefaultValidClusterSpecBuilder().Build()
+	clusterSpec.Cluster.Spec.ProxyConfiguration = &eksav1alpha1.ProxyConfiguration{
+		HttpProxy:  "2.3.4.5",
+		HttpsProxy: "2.3.4.5",
+	}
+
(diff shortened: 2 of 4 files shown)


pkg/providers/tinkerbell/assert.go
```diff
@@ -119,6 +120,31 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
 	return nil
 }
 
+// AssertHookImageURLForProxyConfiguration ensures locally hosted HookImagesURLPath is present for Proxy configuration.
+func AssertHookImageURLForProxyConfiguration(spec *ClusterSpec) error {
```
chrisdoherty4, line 124: I think we talked about this but could you clarify how this verifies the image is hosted locally? The implementation currently verifies if the URL is reachable without any proxy configuration.

pkg/providers/tinkerbell/assert.go
```diff
@@ -119,6 +120,31 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
...
+	if spec.Cluster.Spec.ProxyConfiguration != nil {
```
chrisdoherty4, line 125: Generally we want to be returning early as opposed to nesting deeply. In this instance, it means preferring the following.

```go
if spec.Cluster.Spec.ProxyConfiguration == nil {
  return nil
}

// The rest of the algorithm
```

pkg/providers/tinkerbell/assert.go
```diff
@@ -119,6 +120,31 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
...
+		}
+
+		// verify hookImagesURLPath is accessible locally too
+		transport := http.DefaultTransport
+		backupProxy := transport.(*http.Transport).Proxy
+		transport.(*http.Transport).Proxy = nil
+		client := &http.Client{
+			Transport: transport,
+		}
+		_, err := client.Get(spec.DatacenterConfig.Spec.HookImagesURLPath)
```
chrisdoherty4, line 138: We need the first return value so we can close the `Body` field which should be one after the error is handled.

pkg/providers/tinkerbell/assert.go
```diff
@@ -119,6 +120,31 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
...
+// AssertHookImageURLForProxyConfiguration ensures locally hosted HookImagesURLPath is present for Proxy configuration.
+func AssertHookImageURLForProxyConfiguration(spec *ClusterSpec) error {
+	if spec.Cluster.Spec.ProxyConfiguration != nil {
+		// return an error if hookImagesURLPath field is not specified for during Proxy configuration.
+		if spec.DatacenterConfig.Spec.HookImagesURLPath == "" {
+			return fmt.Errorf("locally hosted hookImagesURLPath is required to support ProxyConfiguration")
+		}
+
+		// verify hookImagesURLPath is accessible locally too
+		transport := http.DefaultTransport
```
chrisdoherty4, line 132: I think we should create a client from scratch instead of modifying the `http` packages state given it isn't thread protected.

chrisdoherty4 (changes requested):

pkg/providers/tinkerbell/assert.go
panktishah26: We want to verify if the hookImageURLPath specified by a user is hosted locally so to check that I am verifying that the URL is accessible even without the proxy environment variables.

pkg/providers/tinkerbell/assert.go
panktishah26: In case if `spec.Cluster.Spec.ProxyConfiguration` is nil, wouldn't it be returned immediately since it won't go into the if statement?

pkg/providers/tinkerbell/assert.go
chrisdoherty4: I'm not making the connection on how that verifies its hosted locally. It seems the algorithm checks "Can be accessed without proxy" as opposed to "Is hosted locally" and I'm not sure how useful that is.

pkg/providers/tinkerbell/assert.go
chrisdoherty4: This is in reference to [code structure](https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply). 

Functionally, both work.

pkg/providers/tinkerbell/assert.go
```diff
@@ -122,6 +123,35 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
 	return nil
 }
 
+// AssertHookImageURLForProxyConfiguration ensures locally hosted HookImagesURLPath is present for Proxy configuration.
+func AssertHookImageURLForProxyConfiguration(spec *ClusterSpec) error {
```
chrisdoherty4, line 127:
```suggestion
// AssertHookRetrievableWithoutProxy ensures the executing machine can retrieve Hook  
// from the host URL without a proxy configured. It does not guarantee the target node
// will be able to download Hook.
func AssertHookRetrievableWithoutProxy(spec *ClusterSpec) error {
```

pkg/providers/tinkerbell/assert.go
```diff
@@ -122,6 +123,35 @@ func AssertTinkerbellIPAndControlPlaneIPNotSame(spec *ClusterSpec) error {
...
+		return fmt.Errorf("locally hosted hookImagesURLPath is required to support ProxyConfiguration")
+	}
+
+	// verify hookImagesURLPath is accessible locally too
+	newHTTPClient := &http.Client{Transport: http.DefaultTransport.(*http.Transport).Clone()}
+	transport := newHTTPClient.Transport
+	transport.(*http.Transport).Proxy = nil
+	client := &http.Client{
+		Transport: transport,
+	}
```
chrisdoherty4, line 143:
```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.Proxy = nil
	client := &http.Client{Transport: transport}
```

pkg/providers/tinkerbell/assert.go
panktishah26: For future reference, with this PR we are checking if the given hook image URL is accessible from the admin machine without use of proxy configuration. This doesn't guarantee that nodes we are trying to provision will have an access to given hook URLs. We will add this in the doc for the reference.

PR #5069, opened 2023-02-24
Update best practice to simplify and reference established guidelines.

chrisdoherty4:
Simplify the best practice and lean into existing resources in an effort to stay up-to-date and reduce regurgitation.

### DIFF
--- a/docs/developer/best-practice.md
+++ b/docs/developer/best-practice.md
@@ -2,6 +2,25 @@
 
 Software development is like nature: it's in a constant state of flux and continuously changing. As software engineers, it's our responsibility to embrace the flux and write clean code that is _optimized for change_. In doing so, we can rapidly innovate and satisfy customer needs.
 
+## Contents
+
+- [Development Best Practice](#development-best-practice)
+  - [Contents](#contents)
+  - [Guiding principles](#guiding-principles)
+  - [Defining _"clean code that is optimized for change"_](#defining-_clean-code-that-is-optimized-for-change_)
+  - [Additions](#additions)
+    - [Global state](#global-state)
+    - ["Accept interfaces, return structs"](#accept-interfaces-return-structs)
+    - [Unit tests](#unit-tests)
+    - [Avoid appending error messages with the word 'error'](#avoid-appending-error-messages-with-the-word-error)
+    - [Avoid boolean parameters](#avoid-boolean-parameters)
+    - [Comment package APIs](#comment-package-apis)
+    - [Don't use naked returns](#dont-use-naked-returns)
+    - [Panicking](#panicking)
+    - [`replace` directive](#replace-directive)
+    - [Line length](#line-length)
+    - [File naming](#file-naming)
+
 ## Guiding principles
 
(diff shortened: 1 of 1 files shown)


docs/developer/best-practice.md
```diff
@@ -22,258 +41,86 @@ Finally, we have delivery commitments. Larger more intentional refactors may be
...
-### Testing
-
-Software development is costly and hazardous. Tests provide example usage, guard rails, and assurances. Without tests, you cannot have high quality code. For this reason, all code we spend time designing and writing is worth testing. Similarly, if it can be unit tested, it should be unit tested. Unit tests are the first line of defense and can be complimented with additional integration or end-to-end tests.
+### Unit tests
 
 Unit tests should:
 
 - focus on the public API of a package.
-- use the `_test` idiom for testing files.
+- prefer the `_test` when testing the public API of the package.
```
g-gaston, line 80: curious why lowering it from "use" to "prefer"

docs/developer/best-practice.md
```diff
@@ -22,258 +41,86 @@ Finally, we have delivery commitments. Larger more intentional refactors may be
...
 - treat the package as a black box.
-- cover at least 80% of the public API.
-- named after the entity they're testing; for example, `TestFooBarDependencyErrors` tests that an error occurs when `Foo` is executed and its `Bar` dependency errors.
```
g-gaston, line 121: I would prefer to keep this
I don't really care what the common format is, but it does help to have one. And we have it (theoretically)

g-gaston (commented):
Rest looks good

docs/developer/best-practice.md
chrisdoherty4: Its to remove contradictory statements. We say 'use' but then we also say 'if somethings complex and unexported feel free to not use and do another thing that could include putting the tests in the package.

This was most if not all of my motivation.

docs/developer/best-practice.md
chrisdoherty4: Understood. I'm OK keeping it but perhaps tweaking the wording? Naming after the entity under test isn't applicable for functions, for example.

I agree though, clarity and consistency is useful. 

Do you have any ideas on tweaks to make it more general without being so focused on 'entities'?

docs/developer/best-practice.md
chrisdoherty4: Another though, specifically calling out the public API test may mean 'use' is appropriate.

docs/developer/best-practice.md
g-gaston: Maybe?

> Test names:
> * If you are testing a function, the test name should start with the name of that function followed by the scenario you are testing; for example, `TestFooInvalidInputError` tests that an error occurs when `Foo` is executed with invalid input.
> * If you are testing a method, the test name should start with the name of the receiver type followed by the method name; for example, `TestFooBarDependencyErrors` tests that an error occurs when `Foo` is executed and its `Bar` errors because of a dependency.

g-gaston (approved):

PR #5092, opened 2023-02-28
Add a design proposal on adding CredentialRef to NutanixDatacenter

thunderboltsid:
The CredentialRef property will allow to transparently link credentials for a Prism Central with its corresponding EKS-A abstraction: NutanixDatacenter.

### DIFF
--- a/designs/nutanix-datacenterconfig-credentialref.md
+++ b/designs/nutanix-datacenterconfig-credentialref.md
@@ -7,56 +7,45 @@ using same credentials to authenticate to both instances is not possible. To rem
 with corresponding Nutanix Datacenter Configs.
 
 ## Overview of Solution
-With this feature, a user can specify credentials reference (i.e. Name and Namespace of the secret containing the credentials)
+With this feature, a user can specify credentials reference (i.e. Name of the secret in eksa-system namespace containing the Prism credentials)
 in the Nutanix Datacenter Config. This info will be fetched and used to authenticate to the Nutanix Prism Central instance
 when creating the workload cluster.
 
 ### Solution Details
-Extend the `NutanixDatacenterConfig` CRD to include a new field `credentialRef` storing name, namespace, and kind (similar to how
-`credentialsRef` is defined in the `NutanixCluster` CRD).
+Extend the `NutanixDatacenterConfig` CRD to include a new field `credentialRef` storing name (name of kubernetes secret object) and
+kind ("Secret"). This is similar to how `credentialsRef` is defined in the `NutanixCluster` CRD minus the namespace. The namespace
+of the secret will always be `eksa-system`. Keeping the namespace constant will allow us to avoid granting EKS-A controller access to
+secrets in other namespaces. The `Ref` type used in Snow provider is reused here for this purpose.
 
 ```go
 // NutanixDatacenterConfigSpec defines the desired state of NutanixDatacenterConfig.
 type NutanixDatacenterConfigSpec struct {
     ...
-    // CredentialRef is the reference to the secret that contains the credentials
-    // for the Nutanix Prism Central
+    // CredentialRef is the reference to the secret name that contains the credentials
+    // for the Nutanix Prism Central. The namespace for the secret is assumed to be a constant i.e. eksa-system.
     // +kubebuilder:validation:Optional
     // +optional
-    CredentialRef *CredentialReference `json:"credentialRef,omitempty"`
+    CredentialRef *Ref `json:"credentialRef,omitempty"`
 }
 
-// CredentialKind is the kind of the credential
-type CredentialKind string
-
 const (
-    // CredentialKindSecret is the kind of the credential that corresponds to a Kubernetes Secret
-    CredentialKindSecret = CredentialKind("Secret")
+    // RefKindSecret is the kind of the credential that corresponds to a Kubernetes Secret
+    RefKindSecret = "Secret"
 )
 
-// CredentialReference is the reference to the secret that contains the credentials
-// +kubebuilder:object:generate=true
-type CredentialReference struct {
-    // Kind of the Nutanix credential 
-    // +kubebuilder:validation:Required
-    // +kubebuilder:validation:Enum=Secret
-    // +kubebuilder:default:=Secret
-    Kind CredentialKind `json:"kind"`
-    // Name of the credential.
-    // +kubebuilder:validation:Required
-    // +kubebuilder:validation:MinLength=1
-    Name string `json:"name"`
-    // Namespace of the credential.
-    // +kubebuilder:validation:Optional
-    // +kubebuilder:default:=eksa-system
-    // +optional
-    Namespace string `json:"namespace, omitemp"`
+func (in *NutanixDatacenterConfig) Validate() error {
+    ...
+    if in.Spec.CredentialRef != nil && in.Spec.CredentialRef.Kind != RefKindSecret {
+        return errors.New("only Secret is supported as a credential kind")
+    }
+    ...
 }
 ```
 
 When creating a workload cluster, the `NutanixDatacenterConfig` referenced in the `Cluster` CRD will be fetched and
 the `credentialRef` field will be used to fetch the secret to get the credentials needed to authenticate with Prism Central
-to create resources needed for the cluster. If the `credentialRef` field is not present when creating a workload cluster, we can
-fail the cluster creation request. During management cluster creation, we can fetch the secrets from the
-environment, create the secret manifests, and add the `credentialRef` to the corresponding `NutanixDatacenterConfig` if not present.
-This will allow us to assume that `credentialRef` will always be set in the `NutanixDatacenterConfig` in all cases.
+to create resources needed for the cluster. If the `credentialRef` field is not present when creating a workload cluster,
+validation webhook will return a validation failure. During management cluster creation, we can fetch the secrets from the
+environment, create the secret manifests in `eksa-system` as we do now, and add the `credentialRef` to the corresponding
+`NutanixDatacenterConfig` if not present. This will allow us to assume that `credentialRef` will always be set in the
+`NutanixDatacenterConfig` in all cases.
--- a/designs/nutanix-datacenterconfig-credentialref.md
+++ b/designs/nutanix-datacenterconfig-credentialref.md
@@ -1,0 +1,51 @@
+# Nutanix Datacenter Config Credential Reference
+
+## Introduction
+
+**Problem:** When a management cluster and a workload cluster are present on two different Nutanix Prism Central instances,
+using same credentials to authenticate to both instances is not possible. To remedy this, we need to associate credentials
+with corresponding Nutanix Datacenter Configs.
+
+## Overview of Solution
+With this feature, a user can specify credentials reference (i.e. Name and Namespace of the secret containing the credentials)
+in the Nutanix Datacenter Config. This info will be fetched and used to authenticate to the Nutanix Prism Central instance
+when creating the workload cluster.
+
+### Solution Details
+Extend the `NutanixDatacenterConfig` CRD to include a new field `credentialRef` storing name, namespace, and kind (similar to how
+`credentialsRef` is defined in the `NutanixCluster` CRD).
+
+```go
+// NutanixDatacenterConfigSpec defines the desired state of NutanixDatacenterConfig.
+type NutanixDatacenterConfigSpec struct {
+    ...
+    // CredentialRef is the reference to the secret that contains the credentials
+    // for the Nutanix Prism Central
+    // +optional
+    CredentialRef *NutanixCredentialReference `json:"credentialRef,omitempty"`
+}
+
+// NutanixCredentialKind is the kind of the Nutanix credential
+type NutanixCredentialKind string
+
+// NutanixCredentialReference is the reference to the secret that contains the credentials
+// +kubebuilder:object:generate=true
+type NutanixCredentialReference struct {
+    // Kind of the Nutanix credential
+    // +kubebuilder:validation:Enum=Secret
+    Kind NutanixCredentialKind `json:"kind"`
+    // Name of the credential.
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    Name string `json:"name"`
+    // namespace of the credential.
+    // +optional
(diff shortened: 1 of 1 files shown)


designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+`credentialsRef` is defined in the `NutanixCluster` CRD).
+
+```go
+// NutanixDatacenterConfigSpec defines the desired state of NutanixDatacenterConfig.
+type NutanixDatacenterConfigSpec struct {
+    ...
+    // CredentialRef is the reference to the secret that contains the credentials
+    // for the Nutanix Prism Central
+    // +optional
+    CredentialRef *NutanixCredentialReference `json:"credentialRef,omitempty"`
```
g-gaston, line 25: Since this is optional, what should be the behavior when not present? I'm assuming defaulting to a secret with a fixed name in a fixed namespace?

g-gaston, line 25: Could we use a non specific nutanix type for this?
I believe we have already a ref (or more than one) types, I would love to reuse it.
If for whatever reason they don't serve you purpose, maybe we can go with `type CredentialsRef struct`?

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// NutanixCredentialKind is the kind of the Nutanix credential
+type NutanixCredentialKind string
```
g-gaston, line 29: Can we define a `Secret` kind const?

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+    // Kind of the Nutanix credential
+    // +kubebuilder:validation:Enum=Secret
+    Kind NutanixCredentialKind `json:"kind"`
+    // Name of the credential.
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    Name string `json:"name"`
+    // namespace of the credential.
+    // +optional
+    Namespace string `json:"namespace"`
```
g-gaston, line 43: If this is optional, what's the default behavior? `default`, `eksa-system`?
Also, could you add the `omitempty` tag?

designs/nutanix-datacenterconfig-credentialref.md
g-gaston: I see that this is explained bellow. Could we add that piece of info here as well?

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+}
+```
+
+When creating a workload cluster, the `NutanixDatacenterConfig` referenced in the `Cluster` CRD will be fetched and
+the `credentialRef` field will be used to fetch the secret to get the credentials needed to authenticate with Prism Central
+to create resources needed for the cluster. If the `credentialRef` field is not present, the credentials from the
+cluster (i.e. eksa-system/cluster-name) will be used as default to preserve existing behaviour.
```
g-gaston, line 50: I like this default behavior. However, what do you think about filling this field also for management clusters? The CLI can do that. And for existing workload clusters, we can add a guard in the controller that fills this with the default secret name if it's not present.

That simplifies the rest of the logic since we don't need to handle two cases everywhere, we just make sure the data exists at the beginning of the reconciliation.

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+// +kubebuilder:object:generate=true
+type NutanixCredentialReference struct {
+    // Kind of the Nutanix credential
+    // +kubebuilder:validation:Enum=Secret
+    Kind NutanixCredentialKind `json:"kind"`
+    // Name of the credential.
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    Name string `json:"name"`
+    // namespace of the credential.
```
g-gaston, line 41:
```suggestion
    // Namespace of the credential.
```

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+    // +optional
+    Namespace string `json:"namespace"`
+}
+```
+
+When creating a workload cluster, the `NutanixDatacenterConfig` referenced in the `Cluster` CRD will be fetched and
+the `credentialRef` field will be used to fetch the secret to get the credentials needed to authenticate with Prism Central
+to create resources needed for the cluster. If the `credentialRef` field is not present, the credentials from the
```
chrisdoherty4, line 49: Defaulting to credentials for a different cluster could be undesirable and unintended by the end user.

What do we think about _requiring_ `credentialRef` when creating workload clusters?

designs/nutanix-datacenterconfig-credentialref.md
g-gaston, line 25: I think we should document that the controller won't react to changes in the secret itself, and that a new secret needs to be created and the reference updated for the controller to reconcile those changes.

In fact, we should probably recommend that the secrets are created [immutable](https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable).

g-gaston (commented):
Love this!

designs/nutanix-datacenterconfig-credentialref.md
chrisdoherty4: I don't think secrets need to be immutable. You could re-use the same secret reference for secret rotation, for example. The constraint feels a little unnecessary.

designs/nutanix-datacenterconfig-credentialref.md
jiayiwang7: +1 to @chrisdoherty4 and that is how we do it for Snow provider: the `identityRef` is a required field when creating workload clusters and when you create one through cluster full lifecycle, you need to specify it with a secret object.

thunderboltsid: I'm not opposed to the idea. If we were to mandate the credentialRef, it simplifies the handling for workload clusters.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: @jiayiwang7 suggests we use `Ref`. I'm game for common data structures, however, `Ref` seems to lack `Namespace` https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_types.go#L741

thunderboltsid: Sure, we can add it here. @jiayiwang7 looking at Ref, lack of Namespace is a potential concern.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: For sure.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: Yeah, good point on omitempty. Will add those tags. The `default` should be eksa-system.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: That's a good point. I'm game for also filling it in for management clusters. That simplifies the logic elsewhere.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: Marking secrets as immutable seems a tad bit unnecessary.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: @g-gaston @chrisdoherty4  How do we feel about the following? 
```
When creating a workload cluster, the `NutanixDatacenterConfig` referenced in the `Cluster` CRD will be fetched and
the `credentialRef` field will be used to fetch the secret to get the credentials needed to authenticate with Prism Central
to create resources needed for the cluster. If the `credentialRef` field is not present when creating a workload cluster, we can
fail the cluster creation request. During management cluster creation, we can fetch the secrets from the
environment, create the secret manifests, and add the `credentialRef` to the corresponding `NutanixDatacenterConfig` if not present.
This will allow us to assume that `credentialRef` will always be set in the `NutanixDatacenterConfig` in all cases.
```

designs/nutanix-datacenterconfig-credentialref.md
abhinavmpandey08: do we need the namespace? We mainly operate on objects/CRDs that are in the `eksa-system` namespace. This also prevents eks-a controller from needing access to secrets in all namespaces and limits the access it needs. 
From EKS-A side, we can ask users to create the secret in the `eksa-system` ns.

designs/nutanix-datacenterconfig-credentialref.md
g-gaston: I like it, one minor clarification:
`we can fail the cluster creation request` -> ca we do this in the validation webhook?

@jiayiwang7 this is exactly what we do for snow today, right? Require it for workloads and fill it in from the CLI for management.

chrisdoherty4: @thunderboltsid That seems OK to me.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: If we are fine with the secrets always being created in `eksa-system` namespace, we can definitely use the `Ref` type.

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: Updated to explicitly spell out validation will be done using validation webhook.

thunderboltsid:
@chrisdoherty4 @g-gaston @jiayiwang7 @abhinavmpandey08 can you guys take one more look and see if this works?

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+    // for the Nutanix Prism Central. The namespace for the secret is assumed to be a constant i.e. eksa-system.
+    // +kubebuilder:validation:Optional
+    // +optional
+    CredentialRef *Ref `json:"credentialRef,omitempty"`
+}
+
+const (
+    // RefKindSecret is the kind of the credential that corresponds to a Kubernetes Secret
+    RefKindSecret = "Secret"
+)
```
thunderboltsid, line 34: actually can also just use `constants.SecretKind` from `pkg/constants`.

designs/nutanix-datacenterconfig-credentialref.md
g-gaston: @thunderboltsid how does this sound?

* We allow secrets to be mutable (and immutable). It’s a user’s choice.
* We watch these referenced secrets, using `OwnerReferences` to link back from secrets to `Cluster`s.
* We won’t reuse the same instance of the secret the user provides for the nutanix provider. EKS-A will have its own secret (user provided) and our controller will create a separate secret for the nutanix provider, making any format changes if necessary.
* The same “eks-a secret” can be referenced in multiple eks-a `Cluster`s.

About the watch part, we can maybe implement it in two phases if timelines are a concern. We can chat offline since that's out of scope for this design.

abhinavmpandey08 (approved):

designs/nutanix-datacenterconfig-credentialref.md
g-gaston: @thunderboltsid could you add those points (reworded however you see fit) to the design?
The rest looks good to me

thunderboltsid: @g-gaston can you please take a look again?

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,51 @@
...
+        return errors.New("only Secret is supported as a credential kind")
+    }
+    ...
+}
+```
+
+When creating a workload cluster, the `NutanixDatacenterConfig` referenced in the `Cluster` CRD will be fetched and
+the `credentialRef` field will be used to fetch the secret to get the credentials needed to authenticate with Prism Central
+to create resources needed for the cluster. If the `credentialRef` field is not present when creating a nutanix datacenter config, a
+mutation webhook will add the `credentialRef` field to the `NutanixDatacenterConfig` with the `nutanix-credentials` as the
```
g-gaston, line 43: I thought we said there wasn't going to be any defaulting? The user has to set credentialsRef, and if they don't, the webhook will error out.

For management clusters, since they are created through the CLI, the CLI will take care of filling the credentialsRef and create the secret, pulling the info from the current env vars.

g-gaston (commented):
rest looks good

designs/nutanix-datacenterconfig-credentialref.md
thunderboltsid: Got it; removed the usage of defaulting webhook.

g-gaston (approved):

designs/nutanix-datacenterconfig-credentialref.md
```diff
@@ -0,0 +1,58 @@
...
+secrets in other namespaces. The `Ref` type used in Snow provider is reused here for this purpose.
+
+```go
+// NutanixDatacenterConfigSpec defines the desired state of NutanixDatacenterConfig.
+type NutanixDatacenterConfigSpec struct {
+    ...
+    // CredentialRef is the reference to the secret name that contains the credentials
+    // for the Nutanix Prism Central. The namespace for the secret is assumed to be a constant i.e. eksa-system.
+    // +kubebuilder:validation:Optional
+    // +optional
```
chrisdoherty4, line 27: We don't need both of these and `+optional` is typically used.

```suggestion
    // +optional
```

chrisdoherty4 (commented):
1 small comment. Everything else lgtm.

thunderboltsid:
> 1 small comment. Everything else lgtm.

fixed

abhinavmpandey08 (approved):
/lgtm
/approve

PR #5109, opened 2023-03-01
set a default region for PackageControllerClient

ewollesen:
As the region setting is optional, it can't be assumed to be set, as was happening in places.

Since it seems safe to have a default value set, and there was a default value already defined, I've gone ahead and set it.

Now things can assume the value of the eksaRegion field is a non-zero value.

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -64,6 +64,7 @@ func NewPackageControllerClient(chartInstaller ChartInstaller, kubectl KubectlRu
 		chartInstaller: chartInstaller,
 		kubectl:        kubectl,
 		registryMirror: registryMirror,
+		eksaRegion:     eksaDefaultRegion,
 	}
 
 	for _, o := range options {
@@ -148,7 +149,9 @@ func (pc *PackageControllerClient) GetCuratedPackagesRegistries() (sourceRegistr
 			defaultImageRegistry = gatedOCINamespace
 		}
 	} else {
-		defaultImageRegistry = strings.ReplaceAll(defaultImageRegistry, eksaDefaultRegion, pc.eksaRegion)
+		if pc.eksaRegion != eksaDefaultRegion {
+			defaultImageRegistry = strings.ReplaceAll(defaultImageRegistry, eksaDefaultRegion, pc.eksaRegion)
+		}
 	}
 	return sourceRegistry, defaultRegistry, defaultImageRegistry
 }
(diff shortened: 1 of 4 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -371,8 +374,6 @@ func WithEksaRegion(eksaRegion string) func(client *PackageControllerClient) {
 	return func(config *PackageControllerClient) {
 		if eksaRegion != "" {
 			config.eksaRegion = eksaRegion
-		} else {
-			config.eksaRegion = eksaDefaultRegion
```
a-cool-train, line 375: Isn't this already setting a default value for this field?

ewollesen: It doesn't, no.

If a user calls `NewPackageControllerClient` without any options, the `eksaRegion` is just `""`. 

In order to get the benefit of this code fragment, one has to call `NewPackageControllerClient(..., curatedpackages.WithEksaRegion(""))`

Why would someone ever call `curatedpackages.WithEksaRegion` with an empty value? That's very unintuitive to me.

a-cool-train: The intention of the WithEksaRegion is to allow the reading of configs from environment variables: [factory.go](https://github.com/aws/eks-anywhere/blob/d4e1f9cd05761dc16dfd25709609623bb511a3ef/pkg/dependencies/factory.go#L1000) then passing this forward to the WithEksaRegion to determine whether to use the default region or not. So the issue here is not calling with an empty value but rather determining from the environment variable whether or not to set to default. The only use cases I am currently aware of this functionality is through the factory.go by the creation of packagecontrollerclient. Is there another use case I am not aware of here?

ewollesen: Yes. Full cluster lifecycle doesn't have the ability to use environment variables. 

All the same, I don't see any reason this isn't an improvement? Does it break the desired functionality?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -148,7 +149,9 @@ func (pc *PackageControllerClient) GetCuratedPackagesRegistries() (sourceRegistr
 			defaultImageRegistry = gatedOCINamespace
 		}
 	} else {
-		defaultImageRegistry = strings.ReplaceAll(defaultImageRegistry, eksaDefaultRegion, pc.eksaRegion)
+		if pc.eksaRegion != eksaDefaultRegion {
+			defaultImageRegistry = strings.ReplaceAll(defaultImageRegistry, eksaDefaultRegion, pc.eksaRegion)
```
a-cool-train, line 153: nit: This is really not necessary. Even if the check wasn't done, this will still work.

vincentni: the same

vincentni (approved):

pkg/curatedpackages/packagecontrollerclient.go
ewollesen: Agreed, but I thought it better captured the intention.

PR #5113, opened 2023-03-02
Backup CAPI objects of management cluster

panktishah26:
*Description of changes:*
Backup CAPI objects before moving from management cluster to bootstrap cluster during upgrade

*Testing (if applicable):*
I have tested this feature by create/upgrade/delete of management and workload cluster.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/root.go
+++ b/cmd/eksctl-anywhere/cmd/root.go
@@ -24,7 +24,6 @@ var rootCmd = &cobra.Command{
 		if outputFilePath == "" {
 			return
 		}
-
 		if err := os.Remove(outputFilePath); err != nil {
 			fmt.Printf("Failed to cleanup log file %s: %s", outputFilePath, err)
 		}
--- a/pkg/executables/clusterctl.go
+++ b/pkg/executables/clusterctl.go
@@ -176,6 +176,26 @@ func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait ti
 	return false, 0
 }
 
+// BackupManagement save CAPI resources of a workload cluster before moving it to the bootstrap cluster during upgrade.
+func (c *Clusterctl) BackupManagement(ctx context.Context, cluster *types.Cluster, managementStatePath string) error {
+	filePath := filepath.Join(".", cluster.Name, managementStatePath)
+	err := os.MkdirAll(filePath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("could not create backup file for CAPI objects: %v", err)
+	}
+
+	_, err = c.Execute(
+		ctx, "move",
+		"--to-directory", filePath,
+		"--kubeconfig", cluster.KubeconfigFile,
+		"--namespace", constants.EksaSystemNamespace,
+	)
+	if err != nil {
+		return fmt.Errorf("failed taking backup of CAPI objects: %v", err)
+	}
+	return nil
+}
+
 func (c *Clusterctl) MoveManagement(ctx context.Context, from, to *types.Cluster) error {
 	params := []string{"move", "--to-kubeconfig", to.KubeconfigFile, "--namespace", constants.EksaSystemNamespace}
 	if from.KubeconfigFile != "" {
--- a/pkg/task/task.go
+++ b/pkg/task/task.go
@@ -27,24 +27,25 @@ type Task interface {
 
 // Command context maintains the mutable and shared entities.
 type CommandContext struct {
-	Bootstrapper       interfaces.Bootstrapper
-	Provider           providers.Provider
-	ClusterManager     interfaces.ClusterManager
-	GitOpsManager      interfaces.GitOpsManager
-	Validations        interfaces.Validator
-	Writer             filewriter.FileWriter
-	EksdInstaller      interfaces.EksdInstaller
-	PackageInstaller   interfaces.PackageInstaller
-	EksdUpgrader       interfaces.EksdUpgrader
-	CAPIManager        interfaces.CAPIManager
-	ClusterSpec        *cluster.Spec
-	CurrentClusterSpec *cluster.Spec
-	UpgradeChangeDiff  *types.ChangeDiff
-	BootstrapCluster   *types.Cluster
-	ManagementCluster  *types.Cluster
-	WorkloadCluster    *types.Cluster
-	Profiler           *Profiler
-	OriginalError      error
+	Bootstrapper              interfaces.Bootstrapper
+	Provider                  providers.Provider
+	ClusterManager            interfaces.ClusterManager
+	GitOpsManager             interfaces.GitOpsManager
+	Validations               interfaces.Validator
+	Writer                    filewriter.FileWriter
+	EksdInstaller             interfaces.EksdInstaller
+	PackageInstaller          interfaces.PackageInstaller
+	EksdUpgrader              interfaces.EksdUpgrader
+	CAPIManager               interfaces.CAPIManager
+	ClusterSpec               *cluster.Spec
+	CurrentClusterSpec        *cluster.Spec
+	UpgradeChangeDiff         *types.ChangeDiff
+	BootstrapCluster          *types.Cluster
+	ManagementCluster         *types.Cluster
+	WorkloadCluster           *types.Cluster
+	Profiler                  *Profiler
+	OriginalError             error
+	ManagementClusterStateDir string
 }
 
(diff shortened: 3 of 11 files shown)


cmd/eksctl-anywhere/cmd/root.go
```diff
@@ -51,10 +50,14 @@ func initLogger() error {
 		return fmt.Errorf("failed to create logs folder: %v", err)
 	}
 
-	outputFilePath := filepath.Join(".", "eksa-cli-logs", fmt.Sprintf("%s.log", time.Now().Format("2006-01-02T15_04_05")))
+	timeNow := time.Now().Format("2006-01-02T15_04_05")
+	outputFilePath := filepath.Join(".", "eksa-cli-logs", fmt.Sprintf("%s.log", timeNow))
+	managementClusterState := fmt.Sprintf("management-cluster-state-save-%s", timeNow)
```
vivek-koppuru, line 55: Can we store this in the mgmt cluster directory? maybe under `mgmt-cluster/cluster-state-backup/` or something?

pkg/executables/clusterctl.go
```diff
@@ -176,6 +177,26 @@ func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait ti
 	return false, 0
 }
 
+// BackupManagement save CAPI resources of a workload cluster before moving it to the bootstrap cluster during upgrade.
+func (c *Clusterctl) BackupManagement(ctx context.Context, cluster *types.Cluster) error {
+	filePath := filepath.Join(".", cluster.Name, logger.GetManagementClusterStatePath())
+	err := os.MkdirAll(filePath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("could not create backup file for CAPI objects: %v", err)
```
vivek-koppuru, line 185:
```suggestion
		return fmt.Errorf("creating backup directory for management objects: %v", err)
```

pkg/executables/clusterctl.go
```diff
@@ -176,6 +177,26 @@ func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait ti
...
+	}
+
+	_, err = c.Execute(
+		ctx, "move",
+		"--to-directory", filePath,
+		"--kubeconfig", cluster.KubeconfigFile,
+		"--namespace", constants.EksaSystemNamespace,
+	)
+	if err != nil {
+		return fmt.Errorf("failed taking backup of CAPI objects: %v", err)
```
vivek-koppuru, line 195:
```suggestion
		return fmt.Errorf("saving backup for management objects: %v", err)
```

pkg/workflows/upgrade.go
```diff
@@ -449,8 +451,15 @@ func (s *installCAPITask) Restore(ctx context.Context, commandContext *task.Comm
 }
 
 func (s *moveManagementToBootstrapTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Backing up workload cluster's management resources before moving to bootstrap cluster")
```
vivek-koppuru, line 454: Let's make this a 1 verbosity log

pkg/workflows/upgrade.go
```diff
@@ -652,6 +662,11 @@ func (s *deleteBootstrapClusterTask) Run(ctx context.Context, commandContext *ta
 			// We don't necessarily need to return with an error here and abort
 			logger.Info(fmt.Sprintf("%v", err))
 		}
+
+		capiObjectFile := filepath.Join(commandContext.BootstrapCluster.Name, logger.GetManagementClusterStatePath())
+		if err := os.RemoveAll(capiObjectFile); err != nil {
+			logger.Info(fmt.Sprintf("management cluster CAPI backup file not found: %v", err))
```
vivek-koppuru, line 668:
```suggestion
			logger.Info(fmt.Sprintf("deleting management cluster CAPI backup: %v", err))
```

cmd/eksctl-anywhere/cmd/root.go
panktishah26: Yeah, I do store it in the management cluster directory. I prepend management cluster directory from clusterctl.go file. https://github.com/aws/eks-anywhere/pull/5113/files#diff-682bd4902d8a65030e531aeae6a0c303cb4c74ce5b469387b51ef56fb604d7aaR182
I will change the folder name to `cluster-state-backup`.

pkg/workflows/upgrade.go
panktishah26: Since this is not throwing any error, does this look okay?
`logger.Info(fmt.Sprintf("management objects backup folder not found: %v", err))`

vivek-koppuru: How about `Warning: Unable to find management objects backup folder due to: %v`?

vivek-koppuru (approved):

pkg/logger/output_file.go
```diff
@@ -28,3 +30,17 @@ func GetOutputFilePath() string {
...
 }
+
+func setManagementClusterStatePath(path string) {
+	managementClusterStateMtx.Lock()
+	defer managementClusterStateMtx.Unlock()
+	managementClusterState = path
+}
+
+// GetManagementClusterStatePath returns the path to the fiolder where management cluster's CAPI resources backup is.
+// Deprecated: The function will be removed to avoid using package state.
```
vivek-koppuru, line 41: Is this really deprecated?

chrisdoherty4 (changes requested):
Could we shift to injecting paths into the data structures that need them as opposed to using the `logger` packages state?

pkg/logger/init.go
```diff
@@ -69,6 +69,7 @@ type Options struct {
 	// OutputFilePath is an absolute file path. The file will be created if it doesn't exist.
 	// All logs available at level 9 will be written to the file.
 	OutputFilePath string
+	ManagementClusterState string
```
chrisdoherty4, line 72: It doesn't seem like this is used by anything in the logger package - what's its purpose?

panktishah26: The purpose of using the logger package is the flow is kind of similar to eksa-cli.log file we generate and delete after cluster creation succeeds and keep it if cluster creation fails. Also, we want this folder from many other packages like workflows, clustermanager, executables, etc so logger package seemed suitable for that.

chrisdoherty4: I appreciate the motivation, however logger output file path configuration and management cluster state snapshots are 2 very different concepts with the latter not really having anything to do with logging.

We want to inject the path into the data structures that require it as opposed to have them rely on global state.

panktishah26: As discussed, moved ManagementClusterState to task package.

pkg/task/task.go
```diff
@@ -27,24 +27,25 @@ type Task interface {
...
+	CAPIManager            interfaces.CAPIManager
+	ClusterSpec            *cluster.Spec
+	CurrentClusterSpec     *cluster.Spec
+	UpgradeChangeDiff      *types.ChangeDiff
+	BootstrapCluster       *types.Cluster
+	ManagementCluster      *types.Cluster
+	WorkloadCluster        *types.Cluster
+	Profiler               *Profiler
+	OriginalError          error
+	ManagementClusterState string
```
vivek-koppuru, line 48: I think now that it is part of the command context, we can be more clear with what the string represents.
Either `ManagementClusterStateDir` or `ManagementClusterStateFile`

chrisdoherty4 (commented):
I'm still not thrilled about embedding knowledge of how a backup is done across the code base.

Can we raise a ticket to address that problem and then be explicit as Vivek has suggested about paths and we should be OK?

pkg/executables/clusterctl.go
```diff
@@ -176,6 +176,26 @@ func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait ti
 	return false, 0
 }
 
+// BackupManagement save CAPI resources of a workload cluster before moving it to the bootstrap cluster during upgrade.
+func (c *Clusterctl) BackupManagement(ctx context.Context, cluster *types.Cluster, managementStatePath string) error {
+	filePath := filepath.Join(".", cluster.Name, managementStatePath)
+	err := os.MkdirAll(filePath, os.ModePerm)
```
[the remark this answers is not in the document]
panktishah26: As discussed, filewriter package creates a dir in generated folder so we would not be able to use that.

vivek-koppuru (approved):

PR #5141, opened 2023-03-03
Align root README Kubernetes images

chrisdoherty4:
Align Kubernetes support images on the root README to avoid creating lots of whitespace.


abhinavmpandey08 (approved):
/approve

PR #5151, opened 2023-03-03
Introduce --no-timeouts flag in create and upgrade CLI

jiayiwang7:
*Issue #, if available:*

#5104 

*Description of changes:*

Introduce `--no-timeouts` flag to disable timeout for all the cluster operations.

This PR
- overrides the default cluster_manager retriers with a new retrier with no timeout, when the flag is presented.
- overrides all kubectl wait timeout with max time

*Testing (if applicable):*

Unit tests

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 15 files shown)


pkg/clustermanager/eksa_installer.go
```diff
@@ -38,7 +38,7 @@ func NewEKSAInstaller(client KubernetesClient, reader manifests.FileReader) *EKS
 }
 
 // Install configures and applies eks-a components in a cluster accordingly to a spec.
-func (i *EKSAInstaller) Install(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error {
+func (i *EKSAInstaller) Install(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec, deploymentTimeout string) error {
```
g-gaston, line 41: should we just make this configurable through the constructor instead?

If you prefer to pass it to the method, can we document what it does in the doc comment. It's not super obvious without reading the code.

pkg/clustermanager/eksa_installer.go
```diff
@@ -38,7 +38,7 @@ func NewEKSAInstaller(client KubernetesClient, reader manifests.FileReader) *EKS
 }
```
g-gaston, line 38: Do we need to update the KubernetesClient as well? I believe that's a retrier client as well.

jiayiwang7: oh yea good catch

pkg/dependencies/factory.go
```diff
@@ -846,7 +883,12 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster, opts ...cl
 			},
 			clustermanager.DefaultRetrier(),
 		)
-		installer := clustermanager.NewEKSAInstaller(client, f.dependencies.FileReader)
+
+		if timeoutOpts != nil && timeoutOpts.NoTimeout {
+			client.Retrier = retrier.NewWithNoTimeout()
```
g-gaston, line 888: nit: it would be cleaner if just pass the right retrier to the constructor
And ideally make `client.Retrier` -> `client.retrier`

pkg/dependencies/factory_test.go
```diff
@@ -214,7 +214,7 @@ func TestFactoryBuildWithMultipleDependencies(t *testing.T) {
 		WithLocalExecutables().
 		WithBootstrapper().
 		WithCliConfig(&tt.cliConfig).
-		WithClusterManager(tt.clusterSpec.Cluster).
+		WithClusterManager(tt.clusterSpec.Cluster, nil).
```
g-gaston, line 217: maybe add a test with options?

pkg/clustermanager/cluster_manager.go
```diff
@@ -228,6 +241,25 @@ func WithRetrier(retrier *retrier.Retrier) ClusterManagerOpt {
 	}
 }
 
+// WithNoTimeout disables the timeout for all the waits and retries in cluster manager.
+func WithNoTimeout() ClusterManagerOpt {
+	return func(c *ClusterManager) {
+		noTimeoutRetrier := retrier.NewWithNoTimeout()
+		maxTime := time.Duration(math.MaxInt64)
+
+		c.clusterClient.Retrier = noTimeoutRetrier
```
g-gaston, line 250: I think we can omit this now since the factory takes care of this

pkg/dependencies/factory.go
jiayiwang7: i will do the client.Retrier -> retrier in separate PR

g-gaston (approved):

PR #5161, opened 2023-03-06
Add mutation webhook for CloudStack machine configs

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/config/manifest/eksa-components.yaml
+++ b/config/manifest/eksa-components.yaml
@@ -6292,6 +6292,48 @@ webhooks:
     service:
       name: eksa-webhook-service
       namespace: eksa-system
+      path: /mutate-anywhere-eks-amazonaws-com-v1alpha1-cloudstackdatacenterconfig
+  failurePolicy: Fail
+  name: mutation.cloudstackdatacenterconfig.anywhere.amazonaws.com
+  rules:
+  - apiGroups:
+    - anywhere.eks.amazonaws.com
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - cloudstackdatacenterconfigs
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: eksa-webhook-service
+      namespace: eksa-system
+      path: /mutate-anywhere-eks-amazonaws-com-v1alpha1-cloudstackmachineconfig
+  failurePolicy: Fail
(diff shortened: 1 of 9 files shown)


pkg/api/v1alpha1/cloudstackmachineconfig.go
```diff
@@ -85,6 +85,15 @@ func GetCloudStackMachineConfigs(fileName string) (map[string]*CloudStackMachine
 	return configs, nil
 }
 
+func setCloudStackMachineConfigDefaults(machineConfig *CloudStackMachineConfig) {
+	if len(machineConfig.Spec.Users) <= 0 {
```
mitalipaygude, line 89: Setting the ssh key to empty string means the user can never access nodes without the ssh keys right? Instead of these defaults, I vote we add validation for the controller where the user should provide a ssh key, atleast till we start generating ssh keys for the controller.

cxbrowne1207: Do we want to do this for all the other providers as well? If so, maybe it would be good idea to make those changes in this PR or in another PR all at once.

taneyland: I believe we set the set defaults in the CLI to avoid a nil pointer when generating the keys.

But yeah for the controller, I agree that it makes more sense to validate rather than defaulting to empty. Will add that validation & I'll check the other providers to do the same if it doesn't already

config/manifest/eksa-components.yaml
```diff
@@ -6292,9 +6292,9 @@ webhooks:
     service:
       name: eksa-webhook-service
       namespace: eksa-system
-      path: /mutate-anywhere-eks-amazonaws-com-v1alpha1-cluster
+      path: /mutate-anywhere-eks-amazonaws-com-v1alpha1-cloudstackdatacenterconfig
   failurePolicy: Fail
-  name: mutation.cluster.anywhere.amazonaws.com
```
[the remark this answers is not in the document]
taneyland: It doesnt remove them. It just moves everything else down since `cloudstack` comes first alphabetically

pkg/api/v1alpha1/cloudstackmachineconfig.go
mitalipaygude: I do have the code changes to fix it for other providers. I picked up that action item when I noticed this default issue for baremetal.

taneyland: Ok awesome. I can close this PR out then & let you do it for CloudStack too in that PR

PR #5163, opened 2023-03-06
Patch containerd version to 1.6.19

chrisdoherty4:


TerryHowe (approved):
/lgtm

We'll see how it goes with https://github.com/aws/eks-anywhere/blob/main/release/go.mod#L40

PR #5183, opened 2023-03-07
Tinkerbell: patch missing mount paths for BottleRocket upgrade

chrisdoherty4:
When using BottleRocket we are required to alter kubeconfig paths that are mounted into the scheduler and controller-manager containers launched by the kubelet. This is because BottleRocket uses different paths.

Other providers patch using the YAML template ensuring new nodes created during an upgrade have the correct host mount paths.

### DIFF
--- a/pkg/providers/tinkerbell/config/template-cp.yaml
+++ b/pkg/providers/tinkerbell/config/template-cp.yaml
@@ -98,6 +98,29 @@ spec:
             name: awsiamcert
             readOnly: false
 {{- end}}
+{{- /*
+  BottleRocket uses different host paths for kubeconfigs requiring host mount path overwrites for
+  the scheduler and controller-manager static pods.
+*/}}
+{{- if ( eq .format "bottlerocket" ) }}
+      controllerManager:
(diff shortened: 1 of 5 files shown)


jacobweinstock (approved):

HiteshRepo:
Hi @chrisdoherty4 

While scaling up a node group (post successful cluster provisioning), I am facing an issue in the new VMs.
The VMs are moving to `Provisioned` phase but not to `Running` phase while hard-drive boot after restart.
The reason is `kubelet failed to start`.

Unable to ssh into the VM to debug as ssh is not responding.

I tried injecting the user info via workflow action as mentioned here (https://github.com/aws/eks-anywhere/blob/main/examples/tinkerbell/adding-a-user-to-tinkerbelltemplateconfiguration.md) but the workflow fails.
I believe the above user info injection is specific to Ubuntu, I am using bottlerocket OS.

Any idea how to get through this issue?

eksctl version: latest
anywhere plugin release version: 0.13.0
anywhere plugin release number: 25

PR #5185, opened 2023-03-07
Adding packages recover function

jonahjon:
*Issue #, if available:*

*Description of changes:*

Now that the packages e2e relys on the dev versions when creating a dev bundle we've seen a few failures in the dev-release job due to packages images not existing properly. This happens when there is a failure in the aws-packages-dev pipeline (due to new development/changes), and the images not getting copied over to ecr public. 

This failure should not block the aws-dev-release job, so this is adding a dive and recover functionality when one of the dev artifacts failed to make it to ecr public. Tested locally be deleting artifacts out of ECR public, and appears to work as intended.

```
Modifying helm chart for eks-anywhere-packages
Starting to modifying helm chart public.ecr.aws/f5b7k4z5/eks-anywhere-packages
Pulling helm chart public.ecr.aws/f5b7k4z5/eks-anywhere-packages:v0.0.0-501fae1fe0f60aea370b3fd576f2a7c58d870980
Error finding remote Helm chart, fallback to use default version: pulling the helm chart: running the Helm LocateChart command

.....

    packageController:
      helmChart:
        description: Helm chart for eks-anywhere-packages
        imageDigest: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
        name: eks-anywhere-packages
        uri: public.ecr.aws/f5b7k4z5/eks-anywhere-packages:0.3.0-eks-a-v0.0.0-dev-build.1
      packageController:
        arch:
        - amd64
        - arm64
        description: Container image for eks-anywhere-packages image
        imageDigest: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
        name: eks-anywhere-packages
        os: linux
        uri: public.ecr.aws/f5b7k4z5/eks-anywhere-packages:v0.3.0-eks-a-v0.0.0-dev-build.1
      tokenRefresher:
        arch:
        - amd64
        - arm64
        description: Container image for ecr-token-refresher image
        imageDigest: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
        name: ecr-token-refresher
        os: linux
        uri: public.ecr.aws/f5b7k4z5/ecr-token-refresher:v0.3.0-eks-a-v0.0.0-dev-build.1

### CODE REVIEW

File: release/pkg/bundles/package-controller.go [hunk 1]
```diff
@@ -191,3 +204,60 @@ func replaceTag(uri, tag string) string {
...
+func getTag(uri string) string {
+	NewURIList := strings.Split(uri, ":")
+	if len(NewURIList) < 2 {
+		return uri
+	}
+	return NewURIList[len(NewURIList)-1]
+}
+
+// recoverBundle is a function used if we have a helm error to revert the bundle back to it's original state using Git Tag version.
+func recoverBundle(r *releasetypes.ReleaseConfig, imageDigests map[string]string, helmtag, imagetag, tokentag string) (map[string]anywherev1alpha1.Image, []string) {
```
jonathanmeier5, line 218: I'm probably misunderstanding the problem as a whole, but is there a reason we don't just always point the release branch at the prod packages bundle? As a prod build, I feel like it might make sense to explicitly point it at the prod packages bundle?

If we are committed to this recovery logic, could we get a unit test on it?

release/pkg/bundles/package-controller.go
```diff
@@ -191,3 +204,60 @@ func replaceTag(uri, tag string) string {
...
 	return uri
 }
+
+// getTag is used to replace get the tag of an ImageURI
+func getTag(uri string) string {
+	NewURIList := strings.Split(uri, ":")
+	if len(NewURIList) < 2 {
+		return uri
+	}
+	return NewURIList[len(NewURIList)-1]
```
chrisdoherty4, line 214:
```suggestion
	newURIList := strings.Split(uri, ":")
	if len(newURIList) < 2 {
		return uri
	}
	return newURIList[len(newURIList)-1]
```

release/pkg/bundles/package-controller.go
```diff
@@ -191,3 +204,60 @@ func replaceTag(uri, tag string) string {
...
+	bundleImageArtifacts := map[string]anywherev1alpha1.Image{}
+	artifactHashes := []string{}
+	for _, componentName := range sortedComponentNames {
+		for _, artifact := range artifacts[componentName] {
+			if artifact.Image != nil {
+				imageArtifact := artifact.Image
+				bundleImageArtifact := anywherev1alpha1.Image{}
+				if strings.HasSuffix(imageArtifact.AssetName, "helm") {
+					imageArtifact.ReleaseImageURI = replaceTag(imageArtifact.ReleaseImageURI, helmtag)
+					Digest := imageDigests[imageArtifact.ReleaseImageURI]
```
chrisdoherty4, line 233:
```suggestion
					digest := imageDigests[imageArtifact.ReleaseImageURI]
```

release/pkg/bundles/package-controller.go
chrisdoherty4: I'm a little unclear on why this is needed too. If packages is meant to be an independently maintained and releasable artifact, would we not want to depend on a stable release in EKS-A?

jonahjon: So this is a scenario where the dev-packages pipeline runs into issues and is failing, we don't want to block the EKS-A team or e2e tests since the package controller install is part of the cluster creation process.

release/pkg/bundles/package-controller.go
```diff
@@ -191,3 +204,60 @@ func replaceTag(uri, tag string) string {
...
+					assetName := strings.TrimSuffix(imageArtifact.AssetName, "-helm")
+					bundleImageArtifact = anywherev1alpha1.Image{
+						Name:        assetName,
+						Description: fmt.Sprintf("Helm chart for %s", assetName),
+						URI:         imageArtifact.ReleaseImageURI,
+						ImageDigest: Digest,
+					}
+				} else {
+					if strings.HasSuffix(imageArtifact.AssetName, "eks-anywhere-packages") {
```
abhay-krishna, line 242: Do we expect it to be at the end of the asset name? Should we do `Contains` or exact match instead?

jonahjon: This was so it wouldn't get matched with `eks-anywhere-packages-helm`, but I guess there is a check up above the the string `helm` so this wouldn't catch anyway?

PR #5206, opened 2023-03-09
Simplify logger package

chrisdoherty4:
This change tidies the logger package in preparation for adding the ability to pause logging to the console so we can ask the user a question. The pause will facilitate asking the user if they are certain they wish to exit the program during upgrades.

The change includes moving away from a `sync.Once`, which made little sense in the context of unexported functions, and introduces mutexes to protect package state.

The use of package state to communicate output file paths is undesirable, but I've maintained the existing package API as there are consumer I don't want to refactor.


jacobweinstock (approved):
Looking good, just has some build job failures.

PR #5207, opened 2023-03-09
Confirm cancelling when user cancels (Ctrl+C) CLI execution

chrisdoherty4:
_Depends on: #5206_

This PR changes the warning currently output when a user `Ctrl`+`C`s into an interactive Q&A asking the user to confirm they really want to cancel. The warning still exists in its original form but is followed by a question. If the user responds with anything other than a capital `Y` the program continues. This ensures the user has a combination keystroke (`Shift`+`Y`) mitigating the risk of entering the wrong thing.

While the question is posed, console logging is redirected to a buffer. If the user chooses to continue logging collected while paused is dumped to the console before resuming normal logging.

```
Warning: Terminating this operation may leave the cluster in an irrecoverable state.
Are you sure you want to exit? [Y/n]: Y
```

The logic is executed on every command as it is instrumented at the top level root cmd at a similar point to the original warning.

It maintains `SIGTERM` handling which models `SIGKILL` behavior. We should improve on this with a separate PR and attempt to perform a graceful shutdown making `SIGKILL` the only way to forcefully terminate a process (this aligns with unix expectations).

### DIFF
--- a/cmd/eksctl-anywhere/cmd/root.go
+++ b/cmd/eksctl-anywhere/cmd/root.go
@@ -51,7 +51,7 @@ func rootPersistentPreRun(cmd *cobra.Command, args []string) {
 	signals.On(func() {
 		resume := stdout.Pause()
 
-		fmt.Println("Warning: Terminating this operation may leave the cluster in an irrecoverable state.")
+		logger.Info("Warning: Terminating this operation may leave the cluster in an irrecoverable state.")
 		if console.Confirm("Are you sure you want to exit?", os.Stdout, os.Stdin) {
 			os.Exit(-1)
 		}
@@ -66,6 +66,13 @@ func rootPersistentPreRun(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(os.Stderr, "Failed to resume stdout logging: %s", err)
 		}
 	}, syscall.SIGINT)
+
+	// We've historically performed a forceful process termination on SIGTERM. We should be able to
+	// improve this and trigger a graceful shutdown instead.
+	signals.On(func() {
+		logger.Info("Warning: Terminating may have left your cluster in an irrecoverable state.")
+		os.Exit(-1)
+	}, syscall.SIGTERM)
 }
 
 func initLogger(consoleWriter io.Writer) error {
--- a/cmd/eksctl-anywhere/cmd/root.go
+++ b/cmd/eksctl-anywhere/cmd/root.go
@@ -3,15 +3,19 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


chrisdoherty4:
I don't recall if Cobra does anything with CTRL+C. We'll need to validate the behavior given many things direct their context from it and if that gets cancelled it could kill the application anyway.

chrisdoherty4:
I had a thought about `SIGTERM`. We probably want to maintain the behavior of a hard exit on `SIGTERM` simply because its how the system currently behaves. We can improve that situation and gracefully exit in the future.

cmd/eksctl-anywhere/cmd/root.go
```diff
@@ -38,24 +42,55 @@ func init() {
...
 
 func rootPersistentPreRun(cmd *cobra.Command, args []string) {
-	if err := initLogger(); err != nil {
+	stdout := logger.NewPausableWriter(os.Stdout)
+
+	if err := initLogger(stdout); err != nil {
 		log.Fatal(err)
 	}
+
+	signals.On(func() {
```
chrisdoherty4, line 51: This isn't really relevant for commands beyond create delete and upgrade so we can probably move to those locations.

pkg/signals/signals.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"os"
+	"os/signal"
+	"sync"
+)
+
+// On registers a function to be called when one of sigs is received. If no sigs are provided On
+// is a noop. It returns a cancellation func that can be used to unregister the signals. The
+// cancellation func is idempotent.
+func On(fn func(), sigs ...os.Signal) func() {
```
g-gaston, line 12: I like this method. The idea of abstracting the code that gets executed on signals from how that registering is done (in Go) is beautiful.

But, when multiple functions are registered for the same signal, how do we guarantee that they all have time to run if they call `os.Exit` (something quite common for certain signals)? Usually you try to have only one shutdown handler and maybe "wait" an arbitrary amount of time before exiting for other subroutine shutdown.

Maybe we don't need to solve this problem just yet since we only have one. But I wonder if we can do something smart to orchestrate them since we are "making" them all go through this method.

g-gaston (commented):
This is great, very clean

pkg/signals/signals.go
chrisdoherty4: Thanks for the feedback.

> how do we guarantee that they all have time to run if they call os.Exit

We could probably have a lengthy discussion on this. For now, I've purposely made this the callers responsibility. The rationale is rooted in the fact all use-cases at present are quite simplistic and so the responsibility falls on the consumer to ensure no 1 handler calls `os.Exit()` in a way that upsets other handlers.

> maybe "wait" an arbitrary amount of time before exiting for other subroutine shutdown

That sounds like a graceful shutdown. Again, I think that's on the consumer to orchestrate, not this registration function. I made a comment to the effect of "we can do better" in `root.go` FWIW.

Issue #5147, opened 2023-03-03
Flake Test: `TestReconcilerValidateHardwareNoHardware`


chrisdoherty4:
```
--- FAIL: TestReconcilerValidateHardwareNoHardware (0.05s)
    reconciler_test.go:500: 
        Expected
            <string>: minimum hardware count not met for selector '{"type":"worker"}': have 0, require 1
        to contain substring
            <string>: minimum hardware count not met for selector '{"type":"cp"}': have 0, require 1
FAIL
coverage: 88.7% of statements
Stopping the test environment
FAIL	github.com/aws/eks-anywhere/pkg/providers/tinkerbell/reconciler	18.111s
```

chrisdoherty4:
https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-anywhere/5163/eks-anywhere-presubmit/1632803990557167616

chrisdoherty4:
https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-anywhere/5183/eks-anywhere-presubmit/1633506131353538560

chrisdoherty4:
https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-anywhere/5197/eks-anywhere-presubmit/1633522327696183296

chrisdoherty4:
https://github.com/aws/eks-anywhere/actions/runs/4366883145/jobs/7637472822

PR #5211, opened 2023-03-09
Modified error msg check for unit test

pokearu:
*Issue #5147*

*Description of changes:*
The unit test `TestReconcilerValidateHardwareNoHardware` is flaky because it checks for an error string and the error string depends on the order of a Map.

https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/validate.go#L136

The requirements object is a map and it swaps its order of objects, so sometimes the worker node condition gets checked first and we are looking for the control plane condition in the unit test error message.

### DIFF
--- a/pkg/providers/tinkerbell/reconciler/reconciler.go
+++ b/pkg/providers/tinkerbell/reconciler/reconciler.go
@@ -2,6 +2,7 @@ package reconciler
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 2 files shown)


abhinavmpandey08 (approved):
/lgtm
/woof

TerryHowe (approved):
/lgtm

Issue #1647, opened 2022-03-28
Introduce InsecureSkipVerify option for RegistryMirrorConfiguration for all providers


jiayiwang7:
**What would you like to be added**:

Follow up on #1637 where we introduced `InsecureSkipVerify` only for snow provider. We shall support this option for all providers.

**Why is this needed**:

It's a fairly common use case to skips the registry certificate verification for isolated testing or in a tightly controlled, air-gapped environment.

jonathanmeier5:
Easy docs for setting up registry with self-signed certs: https://docs.docker.com/registry/insecure/

PR #5235, opened 2023-03-13
Enable InsecureSkipVerify option for vSphere and Docker

cxbrowne1207:
*Issue #, if available:*
#1647 following up on the introduction of `InsecureSkipVerify` flag, and introducing it to the other providers.

*Description of changes:*
This PR adds support for vSphere, which is only being added for Ubuntu, because BottleRocket does not support `insecure_skip_verify` just yet (see the issue [here](https://github.com/bottlerocket-os/bottlerocket/issues/2859)) and Docker. 

Work on the other providers will continue in other PRs

*Testing (if applicable):*

- Unit tests
- Local e2e tests


pkg/providers/docker/config/template-md.yaml
```diff
@@ -42,10 +42,15 @@ spec:
             [plugins."io.containerd.grpc.v1.cri".registry.mirrors."{{ $orig }}"]
               endpoint = ["https://{{ $mirror }}"]
             {{- end }}
-            {{- if .registryCACert }}
+            {{- if or .registryCACert .insecureSkip }}
```
d8660091, line 45: is this if redundant?

cxbrowne1207: I don't think so. The line directly below should be included if registryCACert or insecureSkip is provided.

d8660091: Right, the line below should be included if registryCACert or insecureSkip is provided. But the line may still be included if neither registryCACert nor insecureSkip is provided

jiayiwang7: i dont think we need the line below if neither is provided

jiayiwang7 (approved):

pkg/api/v1alpha1/cluster.go
```diff
@@ -724,9 +723,12 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 		return fmt.Errorf("registry mirror port %s is invalid, please provide a valid port", clusterConfig.Spec.RegistryMirrorConfiguration.Port)
 	}
 
-	if !features.RegistryMirrorInsecureSkipVerifySupport().IsActive() {
-		if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify && clusterConfig.Spec.DatacenterRef.Kind != SnowDatacenterKind {
-			return errors.New("insecureSkipVerify is only supported for snow provider")
+	if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify {
```
jiayiwang7, line 726: can we add the OS validation?

pkg/api/v1alpha1/cluster.go
```diff
@@ -724,9 +723,12 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
...
+		switch clusterConfig.Spec.DatacenterRef.Kind {
```
jiayiwang7, line 727: instead of switch statement, can we maintain a supported insecureSkipVerify provider list and if not within the list, return error

jiayiwang7 (approved):

PR #5242, opened 2023-03-13
Increase package e2e timeout

jonahjon:
*Issue #, if available:*

*Description of changes:*

- Increase package timeout to 10m on slower cluster runs
- Add defer func to output deployment/package outputs on test ends.


test/framework/cluster.go
```diff
@@ -1295,35 +1301,59 @@ func (e *ClusterE2ETest) VerifyHarborPackageInstalled(prefix string, namespace s
 	}
 }
 
+func (e *ClusterE2ETest) printPackageSpec(ctx context.Context, params []string) {
+	bytes, _ := e.KubectlClient.Execute(ctx, params...)
+	response := &packagesv1.Package{}
+	_ = json.Unmarshal(bytes.Bytes(), response)
```
a-cool-train, line 1307: This could potentially fail if the previous command Execute doesn't work.

PR #5243, opened 2023-03-13
Remove hard coded Tinkerbell provider version

chrisdoherty4:
The hard coded version populates the version key for CAPT in the bundle. It is used by EKS-A to determine if there's a change in the provider and if a new version needs rolling out. Because we always use the same version number, we never roll out the updated CAPT and consequently CAPT fails to launch correctly. This occurs during an upgrade from a v0.14 cluster to main only as that's the first time we've updated CAPT.


abhinavmpandey08 (approved):
/lgtm
/approve

PR #5246, opened 2023-03-14
Fixing tink reconcile worker template omit and logs

pokearu:
*Description of changes:*
The tinkerbell reconcile phase omitting worker node machine templates was setting `nil` to the loop iterator and not the scope object.
Added `nil` for the right objects and also added Nil checks when applying worker node spec.

*Testing (if applicable):*
Built the controller and tested changes.

### DIFF
--- a/pkg/controller/clusterapi.go
+++ b/pkg/controller/clusterapi.go
@@ -61,3 +61,20 @@ func GetKubeadmControlPlane(ctx context.Context, client client.Client, cluster *
 	}
 	return kubeadmControlPlane, nil
 }
+
+// GetMachineDeployment reads a cluster-api MachineDeployment for an eks-a cluster using a kube client.
+// If the MachineDeployment is not found, the method returns (nil, nil).
+func GetMachineDeployment(ctx context.Context, client client.Client, machineDeploymentName string) (*clusterv1.MachineDeployment, error) {
+	machineDeployment := &clusterv1.MachineDeployment{}
+	key := types.NamespacedName{Namespace: constants.EksaSystemNamespace, Name: machineDeploymentName}
+
+	err := client.Get(ctx, key, machineDeployment)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return machineDeployment, nil
+}
--- a/pkg/controller/clusters/workers.go
+++ b/pkg/controller/clusters/workers.go
@@ -2,6 +2,7 @@ package clusters
 
 import (
 	"context"
+	"reflect"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -46,11 +47,13 @@ type WorkerGroup struct {
 }
 
 func (g *WorkerGroup) objects() []client.Object {
-	return []client.Object{
-		g.KubeadmConfigTemplate,
-		g.MachineDeployment,
-		g.ProviderMachineTemplate,
+	objs := []client.Object{g.KubeadmConfigTemplate, g.MachineDeployment}
+
+	if !reflect.ValueOf(g.ProviderMachineTemplate).IsNil() {
+		objs = append(objs, g.ProviderMachineTemplate)
 	}
+
+	return objs
 }
 
 // ToWorkers converts the generic clusterapi Workers definition to the concrete one defined
--- a/pkg/providers/tinkerbell/reconciler/reconciler.go
+++ b/pkg/providers/tinkerbell/reconciler/reconciler.go
@@ -3,6 +3,7 @@ package reconciler
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -29,8 +30,6 @@ const (
 	NewClusterOperation Operation = "NewCluster"
 	// K8sVersionUpgradeOperation indicates to upgrade all nodes to a new Kubernetes version.
 	K8sVersionUpgradeOperation Operation = "K8sVersionUpgrade"
-	// ScaleOperation indicates to change the node size of the cluster.
-	ScaleOperation Operation = "Scale"
 	// NoChange indicates no change made to cluster during periodical sync.
 	NoChange Operation = "NoChange"
 )
@@ -101,7 +100,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *an
 		r.ValidateClusterSpec,
 		r.GenerateSpec,
 		r.ValidateHardware,
-		r.OmitMachineTemplate,
 		r.ValidateDatacenterConfig,
 		r.ValidateRufioMachines,
 		r.CleanupStatusAfterValidate,
@@ -157,6 +155,11 @@ func (r *Reconciler) GenerateSpec(ctx context.Context, log logr.Logger, tinkerbe
 	}
 	tinkerbellScope.Workers = w
 
+	err = r.omitTinkerbellMachineTemplates(ctx, tinkerbellScope)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
 	return controller.Result{}, nil
 }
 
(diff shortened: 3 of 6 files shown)


pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
-	if err != nil {
-		return "", err
-	}
-	wantScaleChange := workerWantScaleChange || cpWantScaleChange
 
 	// The restriction that not allowing scaling and rolling is covered in webhook.
 	op := NoChange
-	switch {
-	case wantScaleChange:
-		op = ScaleOperation
```
ddjjia, line 187: Why scale operation not detected in DetectOperation? We'll never hit scale assertion in hardware validation

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
-func (r *Reconciler) WorkerReplicasDiff(ctx context.Context, tinkerbellScope *Scope) (bool, error) {
-	workerWantScaleChange := false
-	for _, wnc := range tinkerbellScope.ClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
-		md := &clusterv1.MachineDeployment{}
-		mdName := clusterapi.MachineDeploymentName(tinkerbellScope.ClusterSpec.Cluster, wnc)
-		key := types.NamespacedName{Namespace: constants.EksaSystemNamespace, Name: mdName}
-		err := r.client.Get(ctx, key, md)
+func (r *Reconciler) omitTinkerbellMachineTemplates(ctx context.Context, tinkerbellScope *Scope) error { //nolint:gocyclo
+	currentKCP, err := controller.GetKubeadmControlPlane(ctx, r.client, tinkerbellScope.ClusterSpec.Cluster)
+	if err != nil {
```
ddjjia, line 193: Do we need to consider not found err for cluster creation?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
+	}
+
+	if cpMachineTemplate != nil {
+		tinkerbellScope.ControlPlane.ControlPlaneMachineTemplate = nil
+		tinkerbellScope.ControlPlane.KubeadmControlPlane.Spec.MachineTemplate.InfrastructureRef.Name = cpMachineTemplate.GetName()
+	}
+
+	for i, wg := range tinkerbellScope.Workers.Groups {
+		machineDeployment, err := controller.GetMachineDeployment(ctx, r.client, tinkerbellScope.ClusterSpec.Cluster, wg.MachineDeployment.GetName())
 		if err != nil {
```
ddjjia, line 213: Would this also be a scale update?

pkg/providers/tinkerbell/reconciler/reconciler.go
mitalipaygude: yeah we are bundling up no change and scale up together, we changed how we check the assertions below.

pkg/providers/tinkerbell/reconciler/reconciler.go
mitalipaygude: That is already getting handled in the `GetKubeadmControlPlane` defined inside clusterapi.

pkg/providers/tinkerbell/reconciler/reconciler.go
mitalipaygude: It can be a scale up which is why we only check for nil/rolling case below on line 216. we omit the templates otherwise for scale up.

pkg/controller/clusterapi.go
```diff
@@ -61,3 +61,20 @@ func GetKubeadmControlPlane(ctx context.Context, client client.Client, cluster *
 	}
 	return kubeadmControlPlane, nil
 }
+
+// GetMachineDeployment reads a cluster-api MachineDeployment for an eks-a cluster using a kube client.
+// If the MachineDeployment is not found, the method returns (nil, nil).
+func GetMachineDeployment(ctx context.Context, client client.Client, cluster *anywherev1.Cluster, machineDeploymentName string) (*clusterv1.MachineDeployment, error) {
```
abhinavmpandey08, line 67: nit: remove unused param
```suggestion
func GetMachineDeployment(ctx context.Context, client client.Client, machineDeploymentName string) (*clusterv1.MachineDeployment, error) {
```

pkg/controller/clusters/workers.go
```diff
@@ -46,11 +47,14 @@ type WorkerGroup struct {
 }
 
 func (g *WorkerGroup) objects() []client.Object {
-	return []client.Object{
-		g.KubeadmConfigTemplate,
-		g.MachineDeployment,
-		g.ProviderMachineTemplate,
+	objs := make([]client.Object, 0, 3)
+	objs = append(objs, g.KubeadmConfigTemplate, g.MachineDeployment)
```
abhinavmpandey08, line 51: nit:
```suggestion
	objs := []client.Object{g.KubeadmConfigTemplate, g.MachineDeployment}
```

pkg/controller/clusters/workers.go
```diff
@@ -46,11 +47,14 @@ type WorkerGroup struct {
...
+
+	if !reflect.ValueOf(g.ProviderMachineTemplate).IsNil() {
```
abhinavmpandey08, line 53: is this basically checking `g.ProviderMachineTemplate != nil`?

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
-	}
-	wantScaleChange := workerWantScaleChange || cpWantScaleChange
 
 	// The restriction that not allowing scaling and rolling is covered in webhook.
 	op := NoChange
-	switch {
-	case wantScaleChange:
-		op = ScaleOperation
-	case wantVersionChange:
+	if wantVersionChange {
```
abhinavmpandey08, line 184: can you check version change directly in the if condition?
```suggestion
	if currentKCP.Spec.Version != tinkerbellScope.ControlPlane.KubeadmControlPlane.Spec.Version {
```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
 		op = K8sVersionUpgradeOperation
 	}
 	log.Info("Operation detected", "operation", op)
```
abhinavmpandey08, line 187: This log message won't be printed for create operation cuz that's returning instead of just modifying the op https://github.com/aws/eks-anywhere/pull/5246/files#diff-912930497d9a5035c7a0b2e57b0740c683c897b56baf1628991e8e43146c5f68R176-R177

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -173,62 +178,58 @@ func (r *Reconciler) DetectOperation(ctx context.Context, log logr.Logger, tinke
...
 
-// WorkerReplicasDiff indicates if there's difference between current and desired worker node groups.
-func (r *Reconciler) WorkerReplicasDiff(ctx context.Context, tinkerbellScope *Scope) (bool, error) {
-	workerWantScaleChange := false
-	for _, wnc := range tinkerbellScope.ClusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
-		md := &clusterv1.MachineDeployment{}
-		mdName := clusterapi.MachineDeploymentName(tinkerbellScope.ClusterSpec.Cluster, wnc)
-		key := types.NamespacedName{Namespace: constants.EksaSystemNamespace, Name: mdName}
-		err := r.client.Get(ctx, key, md)
+func (r *Reconciler) omitTinkerbellMachineTemplates(ctx context.Context, tinkerbellScope *Scope) error { //nolint:gocyclo
```
abhinavmpandey08, line 191: nit
```suggestion
//nolint:gocyclo
func (r *Reconciler) omitTinkerbellMachineTemplates(ctx context.Context, tinkerbellScope *Scope) error { 
```

abhinavmpandey08 (commented):
looks good! Just some minor comments

pkg/controller/clusters/workers.go
mitalipaygude: yes but ProviderMachineTemplate is not a pointer and the `g.ProviderMachineTemplate != nil` check fails causing nil panics when we tried it.

mitalipaygude (approved):

PR #5262, opened 2023-03-15
Feature: Add unmanaged CNI for FLC

chrisdoherty4:
This PR adds the business logic for Full Lifecycle Controller handling of the Cilium `skipUpgrade` flag in the EKS-A API.

The change relies on an annotation populated whenever we observe EKS-A Embedded Cilium is installed in the cluster. Using the annotation combined with the skip flag, we can determine if we need to install Cilium (for successful cluster creation) or treat the CNI as unmanaged and do nothing.

An edge case exists where a user may have uninstalled EKS-A Embedded Cilium and toggled skip on. In that case, we cannot identify if Cilium was ever installed and consequently will try to install it. Users can work around the problem by manually populating the annotation.

**Manual Tested Scenarios**

* Create a workload cluster and observe EKS-A Cilium installed and annotation present.
* Create a workload cluster, then remove the annotation and observe its put back.
* Create a workload cluster, uninstall Cilium and remove annotation and observe both are installed/put back.
* Create a workload cluster with skip enabled and observe EKS-A Cilium installed but then skipped in future reconciliation.
* Create a workload cluster, enable skip, remove the annotation and observe the annotation get reapplied.
* Create a workload cluster, enable skip, uninstall Cilium and install OSS Cilium and observe upgrades skipped.
* Create a workload cluster, enable skip, remove the annotation, remove Cilium and observe Cilium get reinstalled and the annotation reapplied (this is a bork case that requires a workaround if the user doesn't want Cilium to be reinstalled).

### DIFF
--- a/pkg/networking/cilium/installation.go
+++ b/pkg/networking/cilium/installation.go
@@ -1,6 +1,8 @@
 package cilium
 
 import (
+	"strings"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 )
@@ -12,9 +14,15 @@ type Installation struct {
 	ConfigMap *corev1.ConfigMap
 }
 
-// Installed determines if all Cilium components are present.
-// If the ConfigMap doesn't exist we still considered Cilium is installed.
-// The installation might not be complete but it can be functional.
+// Installed determines if all EKS-A Embedded Cilium components are present. It identifies
+// EKS-A Embedded Cilium by the image name. If the ConfigMap doesn't exist we still considered
+// Cilium is installed. The installation might not be complete but it can be functional.
 func (i Installation) Installed() bool {
-	return i.DaemonSet != nil && i.Operator != nil
+	var isEKSACilium bool
+	if i.DaemonSet != nil {
+		for _, c := range i.DaemonSet.Spec.Template.Spec.Containers {
+			isEKSACilium = isEKSACilium || strings.Contains(c.Image, "eksa")
+		}
+	}
+	return i.DaemonSet != nil && i.Operator != nil && isEKSACilium
 }
--- a/pkg/networking/cilium/reconciler/marker.go
+++ b/pkg/networking/cilium/reconciler/marker.go
@@ -1,0 +1,27 @@
+package reconciler
+
+import (
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
+)
+
+// EKSACiliumInstalledAnnotation indicates a cluster has previously been observed to have
+// EKS-A Cilium installed irrespective of whether its still installed.
+const EKSACiliumInstalledAnnotation = "anywhere.eks.amazonaws.com/eksa-cilium"
+
+// ciliumWasInstalled checks cluster for the EKSACiliumInstalledAnnotation.
+func ciliumWasInstalled(ctx context.Context, cluster *v1alpha1.Cluster) bool {
+	if cluster.Annotations == nil {
+		return false
+	}
+	_, ok := cluster.Annotations[EKSACiliumInstalledAnnotation]
+	return ok
+}
+
+// markCiliumInstalled populates the EKSACiliumInstalledAnnotation on cluster. It may trigger
+// anothe reconciliation event.
+func markCiliumInstalled(ctx context.Context, cluster *v1alpha1.Cluster) {
+	clientutil.AddAnnotation(cluster, EKSACiliumInstalledAnnotation, "")
+}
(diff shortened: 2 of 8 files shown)


pkg/networking/cilium/installation.go
```diff
@@ -12,9 +14,15 @@ type Installation struct {
...
-// The installation might not be complete but it can be functional.
+// Installed determines if all EKS-A Embedded Cilium components are present. It identifies
+// EKS-A Embedded Cilium by the image name. If the ConfigMap doesn't exist we still considered
+// Cilium is installed. The installation might not be complete but it can be functional.
 func (i Installation) Installed() bool {
-	return i.DaemonSet != nil && i.Operator != nil
+	var isEKSACilium bool
+	if i.DaemonSet != nil {
+		for _, c := range i.DaemonSet.Spec.Template.Spec.Containers {
+			isEKSACilium = isEKSACilium || strings.Contains(c.Image, "eksa")
```
g-gaston, line 24: This is ok for now due to time constrains, but what do you think about replacing this by an annotation in the DS? It seems more robust that relying on a substring comparison.

I don't see why the image name wouldn't include `eksa` but we don't have control over that and it's subject to change and if it does, we would need to remember this is here.

pkg/networking/cilium/reconciler/marker.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+// EmbeddedCiliumInstalledAnnotation indicates a cluster has previously been observed to have
+// EKS-A Embedded Cilium installed irrespective of whether its still installed.
+const EmbeddedCiliumInstalledAnnotation = "anywhere.eks.amazonaws.com/embedded-cilium"
```
g-gaston, line 11: Curious why adding the word Embedded as opposed to just "EKS-A Cilium"?
I suspect the moment we introduce this term, we are all gonna start using it.

pkg/networking/cilium/reconciler/marker.go
```diff
@@ -0,0 +1,29 @@
...
+	return ok
+}
+
+// markCiliumInstalled populates the embeddedCiliumInstalledAnnotation on cluster. It may trigger
+// anothe reconciliation event.
+func markCiliumInstalled(ctx context.Context, cluster *v1alpha1.Cluster) {
+	if cluster.Annotations == nil {
+		cluster.Annotations = make(map[string]string)
+	}
+	cluster.Annotations[EmbeddedCiliumInstalledAnnotation] = "true"
```
g-gaston, line 28: there is a `clientutil.AddAnnotation` if you want it

pkg/networking/cilium/reconciler/reconciler_test.go
```diff
@@ -283,13 +299,19 @@ func TestReconcilerReconcileUpgradeInvalidCiliumInstalledVersion(t *testing.T) {
...
+	manifest := buildManifest(tt.WithT, ds, operator)
+	tt.templater.EXPECT().GenerateManifest(tt.ctx, tt.spec).Return(manifest, nil)
+
 	tt.makeCiliumDaemonSetReady()
 	tt.makePreflightDaemonSetReady()
 	tt.makePreflightDeploymentReady()
 
 	result, err := tt.reconciler.Reconcile(tt.ctx, test.NewNullLogger(), tt.client, tt.spec)
 	tt.Expect(result).To(Equal(controller.Result{}))
-	tt.Expect(err).To(MatchError(ContainSubstring("installed cilium DS has an invalid version tag")))
```
g-gaston, line 292: Ummm I think this is still a valid test case, right?

Instead of changing the expectations, I would change the test data in line 287
`ds.Spec.Template.Spec.Containers[0].Image = "cilium:eksa-invalid-version"`

g-gaston (commented):
Looks good, minor questions

pkg/networking/cilium/installation.go
chrisdoherty4: That sounds sensible. It shifts the concern toward deployment which seems saner than depending on an image build process to use a specific name.

I'll raise a ticket.

pkg/networking/cilium/reconciler/marker.go
chrisdoherty4: The 'embedded' term has been floated in a few meetings. I can strip it out entirely.

pkg/networking/cilium/reconciler/reconciler_test.go
```diff
@@ -290,6 +306,9 @@ func TestReconcilerReconcileUpgradeInvalidCiliumInstalledVersion(t *testing.T) {
 	result, err := tt.reconciler.Reconcile(tt.ctx, test.NewNullLogger(), tt.client, tt.spec)
 	tt.Expect(result).To(Equal(controller.Result{}))
 	tt.Expect(err).To(MatchError(ContainSubstring("installed cilium DS has an invalid version tag")))
+	// tt.expectDaemonSetSemanticallyEqual(ds)
+	// tt.expectOperatorSemanticallyEqual(operator)
+	// tt.expectCiliumInstalledAnnotation()
```
g-gaston, line 311: delete?

g-gaston (approved):

pkg/networking/cilium/reconciler/reconciler_test.go
chrisdoherty4: Whoops. Will follow up.

Issue #5073, opened 2023-02-26
EKS Anywhere cluster upgrade failed


hainabc:
**What happened**: 

I tried to upgrade our vSphere Ubuntu cluster using eksctl anywhere upgrade command

EKS-D change diff 	{"oldVersion ": "kubernetes-1-22-eks-11", "newVersion ": "kubernetes-1-22-eks-19"}

Upgrade plan output shows

NAME                 CURRENT VERSION   NEXT VERSION
EKS-A                v0.12.0+1665cb9   v0.14.2+0f5d583
cert-manager         v1.9.1+4f37615    v1.9.1+ee92264
cluster-api          v1.2.0+b1e1d8e    v1.2.0+ca45580
kubeadm              v1.2.0+ea0245c    v1.2.0+2edcdda
vsphere              v1.3.1+0d6a713    v1.3.1+243da93
kubeadm              v1.2.0+649554d    v1.2.0+361b9ed
etcdadm-bootstrap    v1.0.5+4f4a848    v1.0.5-rc4+a1b2d1e
etcdadm-controller   v1.0.4+a69a9f8    v1.0.4-rc4+6ad0a3f
cilium               v1.11.8-eksa.1    v1.11.10-eksa.2 

The upgrade failed with this error:

docker	{"stderr": "Warning: resource kubeadmcontrolplanes/k8s-lower-dal is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.\nError from server (Invalid): error when creating \"STDIN\": ClusterResourceSet.addons.cluster.x-k8s.io \"k8s-lower-dal-csi\" is invalid: spec.strategy: Unsupported value: \"Reconcile\": supported values: \"ApplyOnce\"\nError from server (Invalid): error when creating \"STDIN\": ClusterResourceSet.addons.cluster.x-k8s.io \"k8s-lower-dal-cpi\" is invalid: spec.strategy: Unsupported value: \"Reconcile\": supported values: \"ApplyOnce\"\n"}

[...]

PR #5263, opened 2023-03-15
Validate Mgmt Cluster Bundles Version on Upgrade

jonathanmeier5:
Validates that the management cluster's bundle is the same version or newer than the bundle version used to upgrade a workload cluster.

*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/5105

*Description of changes:*
Validates that a management cluster's bundles rev >= the bundle used by CLI to upgrade a workload cluster.

*Testing (if applicable):*
Unit test and two manual tests:
- [x] upgrading workload cluster with a bundle manifest built in at build time that should fail validation
- [x] upgrading workload cluster with a bundles override that should fail the validation
- [x] upgrading workload cluster with a bundles override that should pass the validation

Example error:
```
2023-03-15T17:13:16.583-0400    V0      ❌ Validation failed    {"validation": "validate management cluster bundle version compatibility", "error": "cannot upgrade workload cluster with bundle 29 while management cluster mgmt-v-jwmeier is on older bundle 28", "remediation": "upgrade management cluster mgmt-v-jwmeier before upgrading workload cluster wk-v-jwmeier"}
```

### DIFF
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -79,3 +79,27 @@ func ValidateK8s126Support(clusterSpec *cluster.Spec) error {
 	}
 	return nil
 }
+
+// ValidateManagementClusterBundlesVersion checks if management cluster's bundle version
+// is greater than or equal to the bundle version used to upgrade a workload cluster.
+func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClient, mgmtCluster *types.Cluster, workload *cluster.Spec) error {
+	cluster, err := k.GetEksaCluster(ctx, mgmtCluster, mgmtCluster.Name)
+	if err != nil {
+		return err
+	}
+
+	if cluster.Spec.BundlesRef == nil {
+		return fmt.Errorf("management cluster bundlesRef cannot be nil")
+	}
+
+	mgmtBundles, err := k.GetBundles(ctx, mgmtCluster.KubeconfigFile, cluster.Spec.BundlesRef.Name, cluster.Spec.BundlesRef.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if mgmtBundles.Spec.Number < workload.Bundles.Spec.Number {
+		return fmt.Errorf("cannot upgrade workload cluster with bundle spec.number %d while management cluster %s is on older bundle spec.number %d", workload.Bundles.Spec.Number, mgmtCluster.Name, mgmtBundles.Spec.Number)
+	}
+
+	return nil
+}
--- a/pkg/validations/cluster_test.go
+++ b/pkg/validations/cluster_test.go
@@ -13,10 +13,12 @@ import (
 	"github.com/aws/eks-anywhere/internal/test"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
 	"github.com/aws/eks-anywhere/pkg/validations/mocks"
+	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
 type clusterTest struct {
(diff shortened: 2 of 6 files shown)


pkg/validations/cluster_test.go
```diff
@@ -232,3 +234,106 @@ func TestValidateK8s126SupportActive(t *testing.T) {
...
+			wkBundlesName:     "bundles-27",
+			wkBundlesNumber:   27,
+			wantErr:           "",
+		},
+		{
+			mgmtBundlesName:   "bundles-28",
+			mgmtBundlesNumber: 28,
+			wkBundlesName:     "bundles-29",
+			wkBundlesNumber:   29,
+			wantErr:           "cannot upgrade workload cluster with bundle 29 while management cluster management-cluster is on older bundle 28",
```
chrisdoherty4, line 262: I'm not sure this error reads well with the double `management cluster management-cluster`.

pkg/validations/cluster.go
```diff
@@ -79,3 +79,26 @@ func ValidateK8s126Support(clusterSpec *cluster.Spec) error {
 	}
 	return nil
 }
+
+// ValidateManagementClusterBundlesVersion checks if management cluster's bundle version is greater than or equal to the bundle version used to upgrade a workload cluster.
+func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClient, mgmtCluster *types.Cluster, clusterSpec *cluster.Spec) error {
```
chrisdoherty4, line 84: Would you mind clarifying the cluster spec is for the workload cluster -  perhaps with a variable change (e.g. `workload`)?

chrisdoherty4:
Why does the workload cluster bundle need to be >= to the management cluster? How does this play with `--bundles-override`?

pkg/validations/upgradevalidations/preflightvalidations.go
```diff
@@ -94,5 +94,17 @@ func (u *UpgradeValidations) PreflightValidations(ctx context.Context) []validat
 			}
 		},
 	}
+
+	if u.Opts.Spec.Cluster.IsManaged() {
```
jiayiwang7, line 98: what do u think abt moving the if condition in the `ValidateManagementClusterBundlesVersion` ?

pkg/validations/upgradevalidations/preflightvalidations_test.go
```diff
@@ -1135,10 +1138,15 @@ func TestPreflightValidationsVsphere(t *testing.T) {
...
 		}
+		s.Cluster.Spec.BundlesRef = &anywherev1.BundlesRef{
+			Name:      "bundles-28",
+			Namespace: constants.EksaSystemNamespace,
+		}
 
 		s.GitOpsConfig = defaultGitOps
 		s.OIDCConfig = defaultOIDC
 		s.AWSIamConfig = defaultAWSIAM
+		// s.Bundles = s.Cluster.Spec.BundlesRef
```
jiayiwang7, line 1149: nit: remove?

pkg/validations/cluster.go
```diff
@@ -79,3 +79,26 @@ func ValidateK8s126Support(clusterSpec *cluster.Spec) error {
...
+		return fmt.Errorf("management cluster bundlesref cannot be nil")
+	}
+
+	mgmtBundles, err := k.GetBundles(ctx, mgmtCluster.KubeconfigFile, cluster.Spec.BundlesRef.Name, cluster.Spec.BundlesRef.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if mgmtBundles.Spec.Number < clusterSpec.Bundles.Spec.Number {
+		return fmt.Errorf("cannot upgrade workload cluster with bundle %d while management cluster %s is on older bundle %d", clusterSpec.Bundles.Spec.Number, mgmtCluster.Name, mgmtBundles.Spec.Number)
```
jiayiwang7, line 100:
```suggestion
		return fmt.Errorf("cannot upgrade workload cluster with bundle spec.number %d while management cluster %s is on older bundle %d", clusterSpec.Bundles.Spec.Number, mgmtCluster.Name, mgmtBundles.Spec.Number)
```

pkg/validations/cluster.go
```diff
@@ -79,3 +79,26 @@ func ValidateK8s126Support(clusterSpec *cluster.Spec) error {
...
+	if cluster.Spec.BundlesRef == nil {
+		return fmt.Errorf("management cluster bundlesref cannot be nil")
+	}
+
+	mgmtBundles, err := k.GetBundles(ctx, mgmtCluster.KubeconfigFile, cluster.Spec.BundlesRef.Name, cluster.Spec.BundlesRef.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if mgmtBundles.Spec.Number < clusterSpec.Bundles.Spec.Number {
```
jiayiwang7, line 99: if not specifying bundleOverride flag, will the workload `clusterSpec` here has a valid bundle number? like does the clusterSpec being properly built with bundle workload cluster bundle fetched from existing mgmt cluster?

pkg/validations/cluster_test.go
jonathanmeier5: `management-cluster` is the name of the management cluster.

jonathanmeier5:
> Why does the workload cluster bundle need to be >= to the management cluster?
> 
> How does this play with `--bundles-override` where a user is expecting us to use what they provide? I think I'd expect this validation to be irrelevant in that case.

Great question. I should've linked to this in the PR originally.

This ticket came out of this customer issue: https://github.com/aws/eks-anywhere/issues/5073

The management cluster's CRDs hadn't been upgraded yet because it was running off an older bundle.

When the CLI went to make an upgrade using a newer bundle, it brought the cluster into a bad state because a CRD was missing a field that was introduced in the next bundle.

For the above reasons, I actually don't think we want to ignore this validation even on bundles override. I'm open to changing this, but I think we always want the management cluster to be on the same or higher bundle version, even when doing a bundle override.

Thoughts?

pkg/validations/upgradevalidations/preflightvalidations.go
jonathanmeier5: I don't have a strong opinion either way. I was following the pattern we established over in the create preflight validations flow: https://github.com/aws/eks-anywhere/blob/368c9b899494f99d17be80adce6599c172b598eb/pkg/validations/createvalidations/preflightvalidations.go#L48-L48

If left up to me I'd probably leave as-is, but again not a strong opinion.

chrisdoherty4:
We had a chat offline and concluded users are expected to upgrade management clusters prior to upgrading workload clusters so there shouldn't be an issue with enforcing this constraint.

Thanks @jiayiwang7 @jonathanmeier5 .

chrisdoherty4 (approved):

g-gaston:
> Looks good, before merging: why do we only apply this for upgrades? The user can run into the same problem by trying to create a workload cluster with a newer CLI, right? Or does `create cluster` use the management's cluster bundle?

Chatted offline. We want to do this for create as well. @jonathanmeier5 is gonna explain this in this PR's description and will follow up with another PR for create.
/lgtm

jonathanmeier5:
Linking this for posterity. This is the problem that caused us to create this feature: https://github.com/aws/eks-anywhere/issues/5073

Essentially someone tried to run an upgrade on a workload cluster before upgrading the mgmt cluster.

When they did this, the workload cluster upgrade failed because the CLI was using a newer bundle that expected certain CRDs to have new fields that the mgmt cluster's older bundle did not have. Because the older mgmt cluster's CRDs did not have these fields present, the upgrade failed.

This validation was added to ensure this does not happen.

PR #5275, opened 2023-03-16
More 1.26 e2e tests

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


test/e2e/docker_test.go
```diff
@@ -17,14 +17,14 @@ import (
 )
 
 // Labels
-func TestDockerKubernetes125Labels(t *testing.T) {
+func TestDockerKubernetesLabels(t *testing.T) {
```
mitalipaygude, line 20: Are we not renaming this to 126Labels because there is the only test for Docker labels?

taneyland: Yeah exactly & we always just update it to the latest version

mitalipaygude:
Shouldnt we be removing all the feature flags right before the release? Is this going out in v0.15.0 release?

taneyland:
> Shouldn't we be removing all the feature flags right before the release? Is this going out in v0.15.0 release?

Yes, I added a line to apply it right before we run the create/upgrade commands so we no longer have to specify it in every single new test

PR #5296, opened 2023-03-17
Changing logic for removing hardware from catalogue

pokearu:
*Description of changes:*
`Removehardwares` was throwing an index out of bound panic because we were reducing the length of the `catalogue.Hardware` in-place. 
Fixed by storing a diff list and replacing.

### DIFF
--- a/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
@@ -51,24 +51,28 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
 func (c *Catalogue) RemoveHardwares(hardware []tinkv1alpha1.Hardware) error {
 	m := make(map[string]bool, len(hardware))
 	for _, hw := range hardware {
-		m[hw.Name+":"+hw.Namespace] = true
+		m[getRemoveKey(hw)] = true
 	}
 
 	diff := []*tinkv1alpha1.Hardware{}
 	for i, hw := range c.hardware {
-		key := hw.Name + ":" + hw.Namespace
+		key := getRemoveKey(*hw)
 		if _, ok := m[key]; !ok {
 			diff = append(diff, c.hardware[i])
 		} else {
 			if err := c.hardwareIndex.Remove(c.hardware[i]); err != nil {
 				return err
 			}
-			delete(m, key)
 		}
 	}
 
 	c.hardware = diff
 	return nil
+}
+
+// getRemoveKey returns key used to search and remove hardware.
+func getRemoveKey(hardware tinkv1alpha1.Hardware) string {
+	return hardware.Name + ":" + hardware.Namespace
 }
 
 // AllHardware retrieves a copy of the catalogued Hardware instances.
--- a/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
+++ b/pkg/providers/tinkerbell/hardware/catalogue_hardware.go
@@ -47,38 +47,27 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -47,38 +47,27 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
...
-	for _, hw := range hardware {
+	diff := []*tinkv1alpha1.Hardware{}
+	for i, hw := range c.hardware {
 		key := hw.Name + ":" + hw.Namespace
-		if _, ok := m[key]; ok {
-			if err := c.RemoveHardware(c.hardware[m[key]], m[key]); err != nil {
+		if _, ok := m[key]; !ok {
+			diff = append(diff, c.hardware[i])
+		} else {
+			if err := c.hardwareIndex.Remove(c.hardware[i]); err != nil {
```
jiayiwang7, line 63:
```suggestion
			if err := c.hardwareIndex.Remove(hw); err != nil {
```

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -47,38 +47,27 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
...
-		m[hw.Name+":"+hw.Namespace] = i
+	m := make(map[string]bool, len(hardware))
+	for _, hw := range hardware {
+		m[hw.Name+":"+hw.Namespace] = true
 	}
 
-	for _, hw := range hardware {
+	diff := []*tinkv1alpha1.Hardware{}
+	for i, hw := range c.hardware {
 		key := hw.Name + ":" + hw.Namespace
```
jiayiwang7, line 59: can we define the key in a method and use it in both set and get?

```
func hardwareKey(hw) string {
  return hw.Name + ":" + hw.Namespace
}
```

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -47,38 +47,27 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
...
-		if _, ok := m[key]; ok {
-			if err := c.RemoveHardware(c.hardware[m[key]], m[key]); err != nil {
+		if _, ok := m[key]; !ok {
+			diff = append(diff, c.hardware[i])
+		} else {
+			if err := c.hardwareIndex.Remove(c.hardware[i]); err != nil {
 				return err
 			}
 			delete(m, key)
```
jiayiwang7, line 66: why do we need to delete the key from m?

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
```diff
@@ -47,38 +47,27 @@ func (c *Catalogue) InsertHardware(hardware *tinkv1alpha1.Hardware) error {
...
-}
-
 // RemoveHardwares removes a slice of hardwares from the catalogue.
 func (c *Catalogue) RemoveHardwares(hardware []tinkv1alpha1.Hardware) error {
-	m := make(map[string]int, len(c.hardware))
-	for i, hw := range c.hardware {
-		m[hw.Name+":"+hw.Namespace] = i
+	m := make(map[string]bool, len(hardware))
+	for _, hw := range hardware {
+		m[hw.Name+":"+hw.Namespace] = true
```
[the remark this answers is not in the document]
pokearu: Ideally yes. We want to look at only unique hw. We do not want to validate any duplicate hw.

pkg/providers/tinkerbell/hardware/catalogue_hardware.go
pokearu: Actually we should not be. In case there are duplicates in c.Hardware, we want to remove all the entries 👍

jiayiwang7 (approved):

PR #5298, opened 2023-03-17
feature: add skip upgrade cilium CNI flag to API

chrisdoherty4:
This adds a flag - `skipUpgrade` - to the Cilium configuration in our v1alpha1 API that will enable self managed CNIs.


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -678,6 +678,12 @@ type CNIConfig struct {
 type CiliumConfig struct {
 	// PolicyEnforcementMode determines communication allowed between pods. Accepted values are default, always, never.
 	PolicyEnforcementMode CiliumPolicyEnforcementMode `json:"policyEnforcementMode,omitempty"`
+
+	// SkipUpgrade indicicates that Cilium maintenance should be skipped during upgrades. This can
+	// be used when operators wish to self manage the Cilium installation.
+	// +kubebuilder:default=false
+	// +optional
+	SkipUpgrade *bool `json:"skipUpgrade,omitempty"`
```
jiayiwang7, line 686: why do we need a pointer here?

jiayiwang7 (commented):
can we run `make release-manifests` to auto update the crd and eksa-components.yaml?

pkg/api/v1alpha1/cluster_types.go
chrisdoherty4: To align with Kubernetes API conventions.

https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#optional-vs-required

PR #5300, opened 2023-03-17
update cloudstack immutable fields validation

tatlat:
*Issue #, if available:*

*Description of changes:*
Updated webhook validation so that all AvailabilityZones are immutable instead of a minimum of one

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
+++ b/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
@@ -136,9 +136,11 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 					allErrs,
 					field.Forbidden(specPath.Child("availabilityZone", oldAz.Name), "availabilityZone is immutable"),
 				)
+				break
 			}
 		} else {
 			allAzOverlap = false
+			break
 		}
 	}
 	if !allAzOverlap {
--- a/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
+++ b/pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
@@ -128,22 +128,23 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 	for _, az := range new.Spec.AvailabilityZones {
 		newAzMap[az.Name] = az
 	}
-	atLeastOneAzOverlap := false
+	allAzOverlap := true
 	for _, oldAz := range old.Spec.AvailabilityZones {
 		if newAz, ok := newAzMap[oldAz.Name]; ok {
-			atLeastOneAzOverlap = true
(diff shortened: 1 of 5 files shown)


pkg/api/v1alpha1/cloudstackdatacenterconfig_webhook.go
```diff
@@ -128,22 +128,23 @@ func validateImmutableFieldsCloudStackCluster(new, old *CloudStackDatacenterConf
 	for _, az := range new.Spec.AvailabilityZones {
 		newAzMap[az.Name] = az
 	}
-	atLeastOneAzOverlap := false
+	allAzOverlap := true
 	for _, oldAz := range old.Spec.AvailabilityZones {
 		if newAz, ok := newAzMap[oldAz.Name]; ok {
-			atLeastOneAzOverlap = true
 			if !newAz.Equal(&oldAz) {
```
ddjjia, line 134: We can exit early on any new/unmatch az.

cxbrowne1207 (commented):
I may have missed this discussion somewhere, but, why the change from at least one AZ must overlap to all must be the same? I'm assuming the CLI has been using the former logic.

tatlat:
> I may have missed this discussion somewhere, but, why the change from at least one AZ must overlap to all must be the same? I'm assuming the CLI has been using the former logic.

We want alignment on how immutable fields are validated between the webhook and provider. In the provider, it checked to see all az are the same.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -551,10 +551,23 @@ func (p *cloudstackProvider) needsNewKubeadmConfigTemplate(workerNodeGroupConfig
...
+// NeedNewMachineTemplate Used by EKS-A controller and CLI upgrade workflow to compare generated CSDC/CSMC's from
 // CAPC resources in fetcher.go with those already on the cluster when deciding whether or not to generate and apply
 // new CloudStackMachineTemplates.
-func AnyImmutableFieldChanged(generatedCsdc, actualCsdc *v1alpha1.CloudStackDatacenterConfig, generatedCsmc, actualCsmc *v1alpha1.CloudStackMachineConfig, log logr.Logger) bool {
+func NeedNewMachineTemplate(generatedCsdc, actualCsdc *v1alpha1.CloudStackDatacenterConfig, generatedCsmc, actualCsmc *v1alpha1.CloudStackMachineConfig, log logr.Logger) bool {
+	if notMatch := compareDatacenterConfig(generatedCsdc, actualCsdc, log); notMatch {
+		return true
+	}
+
+	if err := v1alpha1.ValidateImmutableFieldsCloudStackMachineConfig(generatedCsmc, actualCsmc); len(err) > 0 {
```
chrisdoherty4, line 562: Based on the name alone, I would expect this to never return an error because none of the fields it checks should have changed and that should probably be validated prior to executing this logic. 

What do you think?

PR #5305, opened 2023-03-20
Feature: Implement unmanaged CNI for CLI based clusters

chrisdoherty4:
Implement the behavior for the `SkipUpgrade` Cilium flag on CLI based clusters (i.e. not FLC).

When creating a cluster without specifying `SkipUpgrade` the cluster should continue updating the embedded Cilium. When `SkipUpgrade` is true on create the system should install embedded Cilium but not upgrade it. If a user toggles on `SkipUpgrade` after creating the cluster we should no longer attempt to upgrade it.

Toggling off `SkipUpgrade` is disabled until we've explored the implications.

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -658,12 +658,24 @@ func validateCNIConfig(cniConfig *CNIConfig) error {
 }
 
 func validateCiliumConfig(cilium *CiliumConfig) error {
+	if cilium == nil {
+		return nil
+	}
+
+	if !cilium.IsManaged() {
+		if cilium.PolicyEnforcementMode != "" {
+			return errors.New("when using skipUpgrades for cilium all other fields must be empty")
+		}
+	}
+
 	if cilium.PolicyEnforcementMode == "" {
 		return nil
 	}
+
 	if !validCiliumPolicyEnforcementModes[cilium.PolicyEnforcementMode] {
 		return fmt.Errorf("cilium policyEnforcementMode \"%s\" not supported", cilium.PolicyEnforcementMode)
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -658,12 +658,24 @@ func validateCNIConfig(cniConfig *CNIConfig) error {
 }
 
 func validateCiliumConfig(cilium *CiliumConfig) error {
+	if cilium == nil {
+		return nil
+	}
+
+	if !cilium.IsManaged() {
+		if cilium.PolicyEnforcementMode != "" {
+			return fmt.Errorf("when using skipUpgrades for cilium all other fields must be empty")
```
jiayiwang7, line 667:
```suggestion
			return errors.New("when using skipUpgrades for cilium all other fields must be empty")
```

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -489,7 +489,22 @@ func (n *CiliumConfig) Equal(o *CiliumConfig) bool {
...
 		return false
 	}
-	return n.PolicyEnforcementMode == o.PolicyEnforcementMode
+
+	if n.PolicyEnforcementMode != o.PolicyEnforcementMode {
+		return false
+	}
+
+	// If 1 is nil, both should be nil.
+	if (n.SkipUpgrade == nil || o.SkipUpgrade == nil) && n.SkipUpgrade != o.SkipUpgrade {
```
jiayiwang7, line 498: what abt one is nil, the other is *false? this should return true right?

abhay-krishna:
/override eks-anywhere-cli-attribution-presubmit eks-anywhere-e2e-presubmit eks-anywhere-mocks-presubmits

Quelling the retesting churn

PR #647, opened 2021-11-16
New controller poc

g-gaston:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5314, opened 2023-03-20
Add validation for OS when using registry mirror InsecureSkipVerify

cxbrowne1207:
*Issue #, if available:*
#647 following up on the introduction of InsecureSkipVerify flag, and introducing it to the other providers. However, bottlerocket is not supported

*Description of changes:*
Adds a preflight validation that checks if the OS is valid for the provided registry mirror configuration with insecure skip verify enabled.

*Testing (if applicable):*
Unit tests
Manual testing

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -723,15 +723,6 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 		return fmt.Errorf("registry mirror port %s is invalid, please provide a valid port", clusterConfig.Spec.RegistryMirrorConfiguration.Port)
 	}
 
-	if clusterConfig.Spec.RegistryMirrorConfiguration.InsecureSkipVerify {
-		switch clusterConfig.Spec.DatacenterRef.Kind {
-		case DockerDatacenterKind, NutanixDatacenterKind, VSphereDatacenterKind, TinkerbellDatacenterKind, CloudStackDatacenterKind, SnowDatacenterKind:
-			break
-		default:
-			return fmt.Errorf("insecureSkipVerify is only supported for docker, nutanix, snow, tinkerbell, cloudstack and vsphere providers")
-		}
-	}
-
 	mirrorCount := 0
 	ociNamespaces := clusterConfig.Spec.RegistryMirrorConfiguration.OCINamespaces
 	for _, ociNamespace := range ociNamespaces {
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -2,6 +2,7 @@ package validations
 
 import (
 	"context"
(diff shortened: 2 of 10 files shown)


pkg/validations/cluster.go
```diff
@@ -12,6 +12,42 @@ import (
...
+	case v1alpha1.CloudStackDatacenterKind:
+		osFamily = clusterSpec.CloudStackMachineConfigs[cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name].OSFamily()
+	case v1alpha1.SnowDatacenterKind:
+		osFamily = clusterSpec.SnowMachineConfigs[cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name].OSFamily()
+	default:
+		return v1alpha1.InsecureSkipVerifyProviderSupportError()
+	}
+
+	if osFamily == v1alpha1.Bottlerocket {
+		return fmt.Errorf("InsecureSkipVerify is not supported for bottlerocket")
```
jiayiwang7, line 45:
```suggestion
		return errors.New("InsecureSkipVerify is not supported for bottlerocket")
```

pkg/validations/cluster.go
```diff
@@ -12,6 +12,42 @@ import (
...
+	if !cluster.Spec.RegistryMirrorConfiguration.InsecureSkipVerify ||
+		cluster.Spec.DatacenterRef.Kind == v1alpha1.DockerDatacenterKind {
+		return nil
+	}
+
+	var osFamily v1alpha1.OSFamily
+
+	switch cluster.Spec.DatacenterRef.Kind {
+	case v1alpha1.NutanixDatacenterKind:
+		osFamily = clusterSpec.NutanixMachineConfigs[cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name].OSFamily()
```
jiayiwang7, line 31: do we support each machine config to have different OS family?

pkg/api/v1alpha1/cluster.go
```diff
@@ -728,7 +733,7 @@ func validateMirrorConfig(clusterConfig *Cluster) error {
 		case DockerDatacenterKind, NutanixDatacenterKind, VSphereDatacenterKind, TinkerbellDatacenterKind, CloudStackDatacenterKind, SnowDatacenterKind:
```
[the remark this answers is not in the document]
cxbrowne1207: That makes sense. We can just remove it now

pkg/validations/cluster.go
cxbrowne1207: yeah, went through some testing to verify and it seems as though we do. I've updated the code to take that into account.

It's more clean now :)

jiayiwang7 (approved):

PR #5315, opened 2023-03-20
change signaling method for FCL curated packages support

ewollesen:
Previously, the creation a of package bundle controller was executed via a helm chart release. Now that helm chart will create a package instead. The end result being the same, that the packages controller will notice the newly created resource and do what it needs to support curated packages on the new workload cluster.

One additional value is passed to the helm invocation to help the packages controller take the appropriate actions.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/config/manifest/eksa-components.yaml
+++ b/config/manifest/eksa-components.yaml
@@ -6030,7 +6030,7 @@ rules:
 - apiGroups:
   - packages.eks.amazonaws.com
   resources:
-  - packagebundlecontrollers
+  - packages
   verbs:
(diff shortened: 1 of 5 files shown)


ewollesen:
/hold

Still need validation testing, but getting this out earlier for others to test with.

ewollesen:
/unhold

Validation of CLI and FCL indicates that the flag is being passed as expected for workload clusters.

g-gaston (approved):

PR #5324, opened 2023-03-21
Use default credential names if credential names are not set

thunderboltsid:
Use `nutanix-credentials` as NutanixDatacenterConfig credentialRef name if it is not present. This ensures we don't get a nil-pointer dereference during upgrade and upgrade can happen smoothly.

**How has this been tested?**
* Created a cluster using a EKS-A CLI build before FCL changes were merged (NutanixDatacenterConfig spec had no credentialRef)
* Upgraded the cluster successfully using an EKS-A CLI build with these changes included and using the same cluster spec as above (no credentialRef set on NutanixDatacenterConfig). 

Full upgrade logs: https://app.warp.dev/block/w7d8Ic3TrSign6eYTeRv55

### DIFF
--- a/pkg/providers/nutanix/template.go
+++ b/pkg/providers/nutanix/template.go
@@ -116,7 +116,10 @@ func (ntb *TemplateBuilder) GenerateEKSASpecSecret(clusterSpec *cluster.Spec, bu
 // EKSASecretName returns the name of the secret containing the credentials for the nutanix prism central and is used by the
 // EKS-Anywhere controller.
 func EKSASecretName(spec *cluster.Spec) string {
-	return spec.NutanixDatacenter.Spec.CredentialRef.Name
+	if spec.NutanixDatacenter.Spec.CredentialRef != nil {
+		return spec.NutanixDatacenter.Spec.CredentialRef.Name
+	}
+	return constants.NutanixCredentialsName
 }
 
 func (ntb *TemplateBuilder) generateSpecSecret(secretName string, creds credentials.BasicAuthCredential, buildOptions ...providers.BuildMapOption) ([]byte, error) {
--- a/pkg/providers/nutanix/template_test.go
+++ b/pkg/providers/nutanix/template_test.go
@@ -98,6 +98,25 @@ func TestNewNutanixTemplateBuilderGenerateSpecSecretFailure(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestNewNutanixTemplateBuilderGenerateSpecSecretDefaultCreds(t *testing.T) {
+	dcConf, machineConf, workerConfs := minimalNutanixConfigSpec(t)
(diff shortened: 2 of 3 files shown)


thunderboltsid:
> Thanks Sid! You PR looks good. Can you pl update our doc to include this field under [NutanixDatacenter spec?](https://anywhere.eks.amazonaws.com/docs/reference/clusterspec/nutanix/#nutanixdatacenterconfig-fields)

Sure, should we do that in a separate PR?

pkg/providers/nutanix/template.go
```diff
@@ -116,7 +116,10 @@ func (ntb *TemplateBuilder) GenerateEKSASpecSecret(clusterSpec *cluster.Spec, bu
 // EKSASecretName returns the name of the secret containing the credentials for the nutanix prism central and is used by the
 // EKS-Anywhere controller.
 func EKSASecretName(spec *cluster.Spec) string {
-	return spec.NutanixDatacenter.Spec.CredentialRef.Name
+	if spec.NutanixDatacenter.Spec.CredentialRef != nil {
```
chrisdoherty4, line 119: Does this adhere to the behavior described in https://github.com/aws/eks-anywhere/pull/5092#discussion_r1126935945?

thunderboltsid: @chrisdoherty4 My understanding of that comment was restricted to not having a defaulting webhook for new cluster NutanixDatacenterConfig specs. We do have a reconcile for older NutanixDatacenterConfig objects that will add those fields to default values to ensure that the field is set everywhere.

chrisdoherty4:
Could you clarify the situation this arises? 

> Use nutanix-credentials as NutanixDatacenterConfig credentialRef name if it is not present.

thunderboltsid:
> Could you clarify the situation this arises?
> 
> > Use nutanix-credentials as NutanixDatacenterConfig credentialRef name if it is not present.

An EKS-A cluster created before the 

> Could you clarify the situation this arises?
> 
> > Use nutanix-credentials as NutanixDatacenterConfig credentialRef name if it is not present.

This is in a brownfield upgrade path. So consider a case where a EKS-A cluster spec was generated and a cluster created with a EKS-A CLI built prior to our credentialRef+ FCL changes. Now upgrade the cluster with the latest EKS-A CLI changes.

pkg/providers/nutanix/template.go
thunderboltsid: Just to clarify, we are not creating defaults for new clusters. If new cluster specs don't have a credentialRef they will still fail early with validation failures.
```
❌ Validation failed	{"validation": "nutanix Provider setup is valid", "error": "failed to validate cluster spec: credentialRef must be provided", "remediation": ""}
```

chrisdoherty4: Gotcha. Thanks for clarifying.

PR #5326, opened 2023-03-21
Fix kubectl get call to point to full API name

vignesh-goutham:


pkg/executables/kubectl.go
```diff
@@ -1893,7 +1893,7 @@ func (k *Kubectl) CheckProviderExists(ctx context.Context, kubeconfigFile, name,
 		return false, nil
 	}
 
-	params = []string{"get", "provider", "--namespace", namespace, fmt.Sprintf("--field-selector=metadata.name=%s", name), "--kubeconfig", kubeconfigFile}
+	params = []string{"get", "providers.clusterctl.cluster.x-k8s.io", "--namespace", namespace, fmt.Sprintf("--field-selector=metadata.name=%s", name), "--kubeconfig", kubeconfigFile}
```
jiayiwang7, line 1896: can we declare a var on top instead?

```
capiProvidersResourceType             = fmt.Sprintf("providers.%s", clusterv1.GroupVersion.Group)
```

PR #5331, opened 2023-03-21
set management cluster name helm value

ewollesen:
Passes the management cluster's name when installing curated packages on a workload cluster. Removes the workloadOnly flag, as it's no longer needed in this case, as the FCL/CLI curated packages installation for workload clusters is now triggered from the workloadPackageOnly flag.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


ewollesen:
/hold
Holding for validation testing

PR #5345, opened 2023-03-22
Allow smooth upgrades to new kube-proxy with nft 

g-gaston:
### Symptoms
When upgrading a Cloudstack cluster (either in k8s 1.22 or 1.23) from eks-a `v0.14.x` to `v0.15.0` (code in `main`), the api-server looses connectivity to the webhooks. This makes impossible to create/update CRDs with validation/mutation webhooks. In particular, the eks-a process fails when moving back the management resources to the management cluster with `clusterctl`, since this recreates the capi objects.

### Cause

Preface: upgrading from v0.14.3 to v0.15.0 involves upgrading from eks-d `v1.23.16-eks-1-23-16` to `v1.23.16-eks-1-23-17`. The new eks-d version includes the new `kube-proxy` with [support for iptables nft](https://github.com/aws/eks-distro-build-tooling/pull/885). The old `kube-proxy` always uses iptables legacy.

During an upgrade, when the new machine for the new CP node is started, if the machine has iptables nft as the default, the kubelet will use it. Then, before capi updates the `kube-proxy` image version in the DS (this doesn't happen until the CP upgrade is finished), the old `kube-proxy` is scheduled in the node. This old `kube-proxy` doesn't support nft and always uses iptables legacy. When it starts, it adds legacy iptables rules. However, at this point the kubelet has already added iptables-nft rules.

After the CP has been updated, capi updates the `kube-proxy` DS to the new version. This new version has the new wrapper, which detects the rules introduced by the kubelet, so it starts using nft.

The hypothesis is that these leftover legacy rules break the k8s service IP "redirection". It's also possible that Cilium is getting messed up when it starts since it also depends on the rules introduced by `kube-proxy`: https://github.com/cilium/cilium/pull/20123

So it turns out this problem is not exclusive to Cloudstack, it will happen with any OS that defaults to nft. It just happens that the only provider where we have that kind of image right now is Cloudstack.

## Solution

**TLDR**: before the upgrade, schedule a DS with the old kube proxy only in the old nodes and  schedule the existing DS with the new kube-proxy version only in the new nodes.

Upgrade flow changes:
1. Run everything we do today until before the actual capi upgrade/CP upgrade (the capi objects upgrade not the capi components upgrade).
2. Add annotation to KCP to stop `kube-proxy` reconciliations. If not, the kcp controller will undo our next changes.
3. Add special label to all nodes `anywhere.eks.amazonaws.com/iptableslegacy=true`. This allow us to identify the old nodes later.
4. Add node (anti) affinity to main `kube-proxy` DS using `requiredDuringSchedulingIgnoredDuringExecution` and `DoesNotExist`. This ensures the original `kube-proxy` is not scheduled in the old nodes after we update the image version.
4. Delete `kube-proxy` pods in old nodes. Node affinity should remove them, but this removes race conditions.
5. Create new DS `kube-proxy-iptables-legacy` with old `kube-proxy` version and using node selector with  `anywhere.eks.amazonaws.com/iptableslegacy=true`. This DS is only scheduled in the current (pre-upgrade) nodes.
6. Update main `kube-proxy` DS with new eks-d version `v1.23.16-eks-1-23-18`.
7. Continue with the upgrade flow.
8. Delete `kube-proxy-iptables-legacy`.
9. Remove `nodeAffinity` from `kube-proxy`  DS.
10. Remove annotation from KCP to resume `kube-proxy` reconciliations.

### Implementation notes
* I have used a go client for the k8s api calls. I know this is a semi new thing, specially in the CLI and it requires a group conversation and agreement before it gets introduced in the codebase. However, I went with it because I knew the implementation was gonna be significantly faster and this issue is a release blocker. If there are concerns with using this client, I would ask the reviewers to write them down but let the PR continue its course. And if required, I'll come back and refactor this to use `kubectl` calls.
* I haven't put a lot of thought into naming and the placement of the code, both physically (package) and logically (from where do we call the new code). If you have better ideas please voice them out. I'm willing to move this around as necessary if the timing allows it. If it doesn't, I'll create a ticket to come back and addressed the agreed alternative.

## Testing
I have tested manually the present code a few times and it works just fine.

## TODOs
- [x] Cleanup code
- [x] Add cleanup phase after upgrade implementing 8, 9 and 10.
- [x] Run e2e test locally
- [x] Add k8s -> eks-d mapping for all versions
- [x] Unit tests
- [x] (Maybe) add retrying to the client calls or retry the whole "prepare" call and make the code reentrant. This will depend on timing. So far I haven't found issues during testing.


controllers/resource/reconciler_wb_test.go
g-gaston, line 1: Apologies for doing this kind of unexported method testing
But this reconciler and the legacy controller are extremely difficult to test without gigantic setups. Since there were no tests for this, there was no setup I could reuse.

This controller should go away on the next iteration, so I opted to test the private method directly.

pkg/clustermanager/retrier_client.go
```diff
@@ -12,15 +12,15 @@ import (
 
 // RetrierClient wraps around a ClusterClient, offering retry functionality for some operations.
 type RetrierClient struct {
-	*client
+	*clusterManageClient
```
g-gaston, line 15: This is just renaming to avoid collision with the client package

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+		return errors.Wrap(err, "reading the kubeadm control plane for an upgrade")
+	}
+
+	_, newVersion := oci.Split(spec.VersionsBundle.KubeDistro.KubeProxy.URI)
+
+	// If the new spec doesn't include the new kube-proxy or if the current cluster already has it, skip this
+	if needsPrepare, err := needsKubeProxyPreUpgrade(spec, kcp); err != nil {
+		return err
+	} else if !needsPrepare {
+		log.V(4).Info("Kube-proxy upgrade doesn't need special handling", "currentVersion", kcp.Spec.Version, "newVersion", newVersion)
```
vivek-koppuru, line 190: Isn't this just the following? Or did you want to highlight the special handling between old to new differently?
```suggestion
		log.V(4).Info("skipping kube-proxy upgrade setup as it contains same version", "currentVersion", kcp.Spec.Version, "newVersion", newVersion)
```

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+	time.Sleep(u.updateKubeProxyBackoff)
+	return retrier.Retry(u.updateKubeProxyRetries, u.updateKubeProxyBackoff, func() error {
+		kubeProxy, err = getKubeProxy(ctx, client)
+		if err != nil {
+			return err
+		}
+
+		currentImage := kubeProxy.Spec.Template.Spec.Containers[0].Image
+		if currentImage == newKubeProxyImage {
+			log.V(4).Info("Kube-proxy image update seems stable", "wantImage", newKubeProxyImage, "currentImage", currentImage)
```
vivek-koppuru, line 493: How do you ensure stability here? So with these retries is this enough to ensure that the new image has not been reverted? This won't run into race conditions right or can it?

pkg/clustermanager/retrier_client.go
vivek-koppuru:
```suggestion
	*clusterManagerClient
```

pkg/clustermanager/kube_proxy.go
g-gaston: It doesn't necessarily contain the same version
It can contain a different version but none of them including the new kube-proxy
Or it can also contain two different versions but both including the new kube-proxy

pkg/clustermanager/kube_proxy.go
g-gaston: It waits for two seconds after every update and then checks if the image is the expected one
This tries to mitigate a race condition where the KCP controller hasn't yet read the skip annotation so it undoes our update to the kube-proxy image. So we repeat this update in a loop and verify that the image doesn't get changed back.

If you are asking if this guarantees correctness: as almost any wait based mechanism, it doesn't. It's still possible, although unlikely, that the KCP controller won't see the skip annotation in 2 seconds and might try to undo the change. The longer the wait, the more confidence, but it will never be 100%. I couldn't figure out any algorithm that guaranteed the KCP controller had read the skip annotation before we update the DS.

pkg/clustermanager/kube_proxy.go
g-gaston: To clarify: this prepare process is only necessary when there is a transition between a kube-proxy that always uses legacy (the "old" one) and a kube-proxy that uses the new iptables wrapper (the "new" one), which adds support for nft.

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+		u.updateKubeProxyRetries = retries
+		u.updateKubeProxyBackoff = backoff
+	}
+}
+
+// PrepareForUpgrade gets the workload cluster ready for a smooth transition between the
+// old kube-proxy that always uses iptables legacy and the new one that detects the host preference
+// and is able to work with nft as well. This is idempotent, so it can be called in a loop if transient
+// errors are a risk.
+func (u KubeProxyUpgrader) PrepareForUpgrade(ctx context.Context, log logr.Logger, managementClusterClient, workloadClusterClient client.Client, spec *cluster.Spec) error {
```
jiayiwang7, line 178: does this handle case like upgrading from one eks-a version to another one with node count scaling up?

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+	if err := client.Update(ctx, kubeProxy); err != nil {
+		return errors.Wrap(err, "updating main kube-proxy version before upgrade")
+	}
+
+	return nil
+}
+
+func (u KubeProxyUpgrader) ensureUpdateKubeProxyVersion(ctx context.Context, log logr.Logger, client client.Client, spec *cluster.Spec) error {
+	newKubeProxyImage := spec.VersionsBundle.KubeDistro.KubeProxy.URI
+	kubeProxy, err := getKubeProxy(ctx, client)
```
jiayiwang7, line 474: why needing the first getKubeProxy, updateKubeProxyVersion separately from the retrier?

pkg/clustermanager/kube_proxy.go
g-gaston: I haven't tested that specific scenario but yes. It doesn't matter if the number of new nodes is the same as the old nodes or not: all old nodes will use the "legacy" kube-proxy and all new nodes will use the "new" kube-proxy version from the original DS.

pkg/clustermanager/kube_proxy.go
g-gaston: We don't need it. It avoids having the retrier always fail on the first iteration.
That's not a big deal so I changed the logging a bit to make it more explicit and simplified the code. Let me know.

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+	workloadClusterClient, err = u.clientFactory.BuildClientFromKubeconfig(workloadClusterKubeconfigPath)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return managementClusterClient, workloadClusterClient, nil
+}
+
+// NewKubeProxyUpgrader builds a new KubeProxyUpgrader.
+func NewKubeProxyUpgrader(opts ...KubeProxyUpgraderOpt) KubeProxyUpgrader {
```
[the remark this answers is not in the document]
g-gaston: I don't think it will need to be used in the controller. I think we can remove this on the next release.

If we decide not to, the logic to implement this in the controller will be similar, but not the same. For example, we won't use a retrier in the "ensureUpdate" method.

The reason it's two objects is it because it helps separate concerns (business logic from retrying logic). Originally the non CLI one was just a standalone function, but I made it a method to be able to introduce the "timing" options, only because of testing.

pkg/clustermanager/kube_proxy.go
```diff
@@ -0,0 +1,505 @@
...
+	log           logr.Logger
+	retrier       retrier.Retrier
+}
+
+// KubeProxyCLIUpgraderOpt allows to customize a KubeProxyCLIUpgraderOpt
+// on construction.
+type KubeProxyCLIUpgraderOpt func(*KubeProxyCLIUpgrader)
+
+// KubeProxyCLIUpgraderRetrier allows to use a custom retrier.
+func KubeProxyCLIUpgraderRetrier(retrier retrier.Retrier) KubeProxyCLIUpgraderOpt {
```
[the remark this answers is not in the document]
jiayiwang7: IMO we should. Cases like sudden lost network connection to workload cluster as you are fetching and updating workload cluster objects from admin machine can exhaust retry of this upgrade. This is a valid scenario some of our customers specifically want us to handle without failure/timeout.

pkg/clustermanager/kube_proxy.go
vivek-koppuru: Is 2 seconds enough to have this check then? Just wondering if we need longer or if KCP modifies another property when it sees the skip annotation. If not, yea then this seems fine.

pkg/clustermanager/kube_proxy.go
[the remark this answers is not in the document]
g-gaston: Done

jiayiwang7 (approved):

g-gaston:
`eks-anywhere-e2e-presubmit` seems like a flake when pushing the image
/retest

PR #5350, opened 2023-03-22
Add 1.26 Bottlerocket e2e tests

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


test/e2e/tinkerbell_test.go
```diff
@@ -54,12 +54,12 @@ func TestTinkerbellKubernetes123BottleRocketAWSIamAuth(t *testing.T) {
 	runTinkerbellAWSIamAuthFlow(test)
 }
 
-func TestTinkerbellKubernetes125BottleRocketAWSIamAuth(t *testing.T) {
+func TestTinkerbellKubernetes126BottleRocketAWSIamAuth(t *testing.T) {
```
cxbrowne1207, line 57: Do we want none of tests for 125? I noticed that there is currently still one for 123

taneyland: There isn't really a set protocol. We update most of the tests to use the latest version, but we also leave a few tests with older versions just to have insight and coverage of the behavior with other versions, just to ensure that we aren't introducing breaking changes for older versions.

PR #5409, opened 2023-03-27
Fix panic on Tinkerbell custom resource conversion

chrisdoherty4:
When Hardware are submitted to the cluster without baseboard management data they won't have a baseboard management controller reference. The upgrade procedure from pre v0.15 clusters to v0.15+ will panic if the reference is missing on Hardware objects. This change ensures a reference exists before trying to update it.


jiayiwang7:
@chrisdoherty4 pls follow up with unit tests, thanks!

PR #5424, opened 2023-03-28
Fix check for conditional CLI tarball upload path

abhay-krishna:
After the provider-specific e2e test split, it is the build stage that uploads artifacts and not the test stage, to avoid duplicate artifacts upload. So this check needs to test for the build stage.


cxbrowne1207 (commented):
/lgtm
/approve

PR #5436, opened 2023-03-29
Update actions/labeler Githb Action version

chrisdoherty4:
Update the `actions/labeler` action to migrate away from the deprecated Node JS 12 Github Action base.


TerryHowe (approved):
/lgtm

PR #5448, opened 2023-03-29
Adding a case for no substring match

pokearu:
*Description of changes:*
`eksa-test-tool` panics when fetching tinkerbell artifacts for use cases where `/Test` is not found in the object key. This happens when tests have a SSM time out.

### DIFF
--- a/test/e2e/tools/eks-anywhere-test-tool/pkg/filewriter/writer.go
+++ b/test/e2e/tools/eks-anywhere-test-tool/pkg/filewriter/writer.go
@@ -7,6 +7,8 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+
+	"github.com/aws/eks-anywhere/pkg/logger"
 )
 
 type writer struct {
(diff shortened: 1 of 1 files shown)


test/e2e/tools/eks-anywhere-test-tool/pkg/filewriter/writer.go
```diff
@@ -60,6 +62,11 @@ func (t *writer) Dir() string {
 // This method writes the e2e test artifacts from S3 to files in a directory named after the e2e test name.
 func (t *writer) WriteTestArtifactsS3ToFile(key string, data []byte) error {
 	i := strings.LastIndex(key, "/Test")
+	if i == -1 {
+		logger.Info("Failed writing object to file", "key", key)
+		return nil
```
abhinavmpandey08, line 67: why log and return nil instead of returning err?

pokearu: Because we want to fetch the other ones after?

PR #5452, opened 2023-03-30
Add cloudstack validator to controller

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Building validator in controller to use cmk to run validations against Cloudstack. 

Currently in progress, better approach is to build client each time we call a validate function similar to how we do for Snow, but that requires a bigger refactor (https://github.com/aws/eks-anywhere/pull/2859, https://github.com/aws/eks-anywhere/pull/2887). I reused the validator factory that is defined in the provider package. I also removed the error check when building the cmk executable because we are already checking for the validity of execConfig (at least from what I understand) whenever we get it, so it's better to not have errors thrown in the constructor. But we need to make sure that there is enough validation for whether the profile exists or not in the cmk methods. The validator may also need to be built in the ClusterProviderReconciler as there are certain validations that are not isolated to just datacenter config, so that would be the best place to contain all the objects associated with the cluster.

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/cmk.go
+++ b/pkg/executables/cmk.go
@@ -338,10 +338,7 @@ func (c *Cmk) ValidateAccountPresent(ctx context.Context, profile string, accoun
 }
 
 // NewCmk initializes CloudMonkey executable to query CloudStack via CLI.
-func NewCmk(executable Executable, writer filewriter.FileWriter, config *decoder.CloudStackExecConfig) (*Cmk, error) {
-	if config == nil {
-		return nil, fmt.Errorf("nil exec config for CloudMonkey, unable to proceed")
-	}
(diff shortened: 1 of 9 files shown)


pkg/executables/cmk.go
```diff
@@ -338,10 +338,7 @@ func (c *Cmk) ValidateAccountPresent(ctx context.Context, profile string, accoun
 }
 
 // NewCmk initializes CloudMonkey executable to query CloudStack via CLI.
-func NewCmk(executable Executable, writer filewriter.FileWriter, config *decoder.CloudStackExecConfig) (*Cmk, error) {
-	if config == nil {
```
vivek-koppuru, line 342: This error check might still be needed, need to see if this is the right way to implement.

ddjjia:
the changes are picked up at another [pr](https://github.com/aws/eks-anywhere/pull/5722), so close this one.

PR #5454, opened 2023-03-30
Run Helm push command using CLI instead of library

abhay-krishna:
Running Helm push command using CLI instead of library to make it more robust when working credentials from the env. This will unblock the bundle release.

### DIFF
--- a/release/pkg/helm/helm.go
+++ b/release/pkg/helm/helm.go
@@ -17,6 +17,7 @@ package helm
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


TerryHowe:
I assume something broke because of a helm upgrade, but you are just pushing the problem to whenever someone updates what is the build tooling (or if they are running locally to whatever version they happen to have installed). I'm not against this change, but not thrilled either.

abhay-krishna:
> I assume something broke because of a helm upgrade, but you are just pushing the problem to whenever someone updates what is the build tooling (or if they are running locally to whatever version they happen to have installed). I'm not against this change, but not thrilled either.

Surprisingly the helm CLI worked even at the latest version while the latest go mod broke things.

Issue #5313, opened 2023-03-20
Upgrade CloudStack controlPlaneConfigurations.machineGroupRef.name field failed


jiayiwang7:
For Cloudstack, upgrade of `controlPlaneConfigurations.machineGroupRef.name` field fails with below error

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x31e954c]

goroutine 1 [running]:
github.com/aws/eks-anywhere/pkg/providers/cloudstack.(*cloudstackProvider).validateMachineConfigNameUniqueness(0xc000597e50?, {0x46639d0?, 0xc000132008?}, {0xc000544a10, 0xe}, 0xa?)
        /Users/pjshah/workspace/eks-anywhere/pkg/providers/cloudstack/cloudstack.go:1359 +0x4c
github.com/aws/eks-anywhere/pkg/providers/cloudstack.(*cloudstackProvider).validateMachineConfigsNameUniqueness(0xc00081c200, {0x46639d0, 0xc000132008}, 0xc0007a4a00?, 0xc0007a4a00)
```

Upon debugging, it looks it fails at https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/cloudstack.go#L1325 line. `clusterSpec.Cluster.GetNamespace()` throws an error and thus cluster upgrade fails with panic.

PR #5478, opened 2023-03-31
Fix worker node machine group ref name update

taneyland:
*Issue #, if available:*

#5313 

*Description of changes:*
When updating the worker machine group ref names, we would run into an error:
`Error: failed to upgrade cluster: generating capi spec: generating cluster api spec contents: getting eksa vsphere cluster Error from server (NotFound): vspheremachineconfigs.anywhere.eks.amazonaws.com "taylor1" not found`

This happened because we were referring to the new machine config as the old one and vice versa. So when we tried to `kubectl get` the old machine config to compare it to the new one, we were actually attempting to get the new one which doesn't exist yet, causing the error above.

This PR assigns the appropriate configs and the upgrade is now successful.

*Testing (if applicable):*
Upgraded all machine group ref names on a vsphere cluster :white_check_mark:

### DIFF
--- a/internal/test/timer.go
+++ b/internal/test/timer.go
@@ -3,7 +3,14 @@ package test
 import "time"
 
 var defaultNow = time.Unix(0, 1234567890000*int64(time.Millisecond))
+var newNow = time.Unix(0, 987654321000*int64(time.Millisecond))
 
 func FakeNow() time.Time {
 	return defaultNow
+}
+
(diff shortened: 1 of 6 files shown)


cxbrowne1207 (commented):
Can we add a unit test to cover this case?

jiayiwang7 (approved):

jiayiwang7:
lets also backport it to 0.15

PR #5488, opened 2023-03-31
Clone default transport to honor http proxy config

g-gaston:
*Description of changes:*
We need to preserve some of the defaults in the default transport. In particular Proxy, which is set to http.ProxyFromEnvironment. This will make the client honor the HTTP_PROXY, HTTPS_PROXY and NO_PROXY env variables.

### DIFF
--- a/pkg/files/reader.go
+++ b/pkg/files/reader.go
@@ -1,6 +1,8 @@
 package files
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"embed"
 	"fmt"
 	"io"
@@ -9,6 +11,8 @@ import (
 	"os"
 	"strings"
 	"time"
+
+	"golang.org/x/net/http/httpproxy"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/files/reader.go
```diff
@@ -43,16 +47,58 @@ func WithEKSAUserAgent(eksAComponent, version string) ReaderOpt {
 	return WithUserAgent(eksaUserAgent(eksAComponent, version))
 }
 
+// WithRootCACerts configures the HTTP client's trusted CAs. Note that this will ovewrite
```
jiayiwang7, line 50:
```suggestion
// WithRootCACerts configures the HTTP client's trusted CAs. Note that this will overwrite
```

abhinavmpandey08 (approved):
/lgtm

jiayiwang7 (approved):

jiayiwang7 (approved):

PR #5495, opened 2023-04-01
Kubernetes version skew check for cluster webhook

mitalipaygude:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/1252

*Description of changes:*
Kubernetes version skew check for cluster webhook. This validation is a part of our cli validations, including it for the controller as well.

*Testing (if applicable):*
- unit tests
- Controller validations for docker
```
mpaygude@bcd0746652e3 eks-anywhere-1 % k apply -f docker-w1.yaml
dockerdatacenterconfig.anywhere.eks.amazonaws.com/docker124-w1 unchanged
The Cluster "docker124-w1" is invalid: spec: Invalid value: "1.26": only 1 minor version skew is supported
mpaygude@bcd0746652e3 eks-anywhere-1 % k apply -f docker-w1.yaml
dockerdatacenterconfig.anywhere.eks.amazonaws.com/docker124-w1 unchanged
The Cluster "docker124-w1" is invalid: spec: Invalid value: "2.26": major version upgrades are not supported
mpaygude@bcd0746652e3 eks-anywhere-1 % k apply -f docker-w1.yaml
dockerdatacenterconfig.anywhere.eks.amazonaws.com/docker124-w1 unchanged
The Cluster "docker124-w1" is invalid: spec: Invalid value: "1.23": minor version downgrade is not supported. Upgrades should meet the supported version increment of +1
```

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -21,6 +21,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apimachinery/pkg/util/version"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -94,6 +95,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 	allErrs = append(allErrs, validateImmutableFieldsCluster(r, oldCluster)...)
 
 	allErrs = append(allErrs, validateBundlesRefCluster(r, oldCluster)...)
+
+	allErrs = append(allErrs, ValidateKubernetesVersionSkew(r, oldCluster)...)
 
 	if len(allErrs) != 0 {
 		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
@@ -340,4 +343,42 @@ func (r *Cluster) ValidateDelete() error {
 	clusterlog.Info("validate delete", "name", r.Name)
 
 	return nil
+}
+
+// ValidateKubernetesVersionSkew validates Kubernetes version skew between upgrades.
+func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+
+	path := field.NewPath("spec")
+
+	oldVersion := old.Spec.KubernetesVersion
+	newVersion := new.Spec.KubernetesVersion
+
+	parsedOldVersion, err := version.ParseGeneric(string(oldVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing cluster version: %v", err.Error())))
+		return allErrs
+	}
+
+	parsedNewVersion, err := version.ParseGeneric(string(newVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing comparison version: %v", err.Error())))
+		return allErrs
+	}
+
+	if parsedNewVersion.Minor() == parsedOldVersion.Minor() && parsedNewVersion.Major() == parsedOldVersion.Major() {
+		return allErrs
+	}
+
+	if err := ValidateVersionSkew(parsedOldVersion, parsedNewVersion); err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, err.Error()))
+	}
+
+	return allErrs
 }
--- a/pkg/api/v1alpha1/cluster_webhook_test.go
+++ b/pkg/api/v1alpha1/cluster_webhook_test.go
@@ -29,7 +29,7 @@ func TestClusterDefault(t *testing.T) {
 func TestClusterValidateUpdateManagementValueMutableExperimental(t *testing.T) {
 	features.ClearCache()
 	t.Setenv(features.ExperimentalSelfManagedClusterUpgradeEnvVar, "true")
-	cOld := createCluster()
+	cOld := baseCluster()
 	cOld.Spec.ControlPlaneConfiguration.Labels = map[string]string{"Key1": "Val1"}
 	cOld.Spec.ControlPlaneConfiguration.Taints = []v1.Taint{
 		{
@@ -56,7 +56,7 @@ func TestClusterValidateUpdateManagementValueMutableExperimental(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementValueImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{}
+	cOld := baseCluster()
 	cOld.SetSelfManaged()
 
 	c := cOld.DeepCopy()
@@ -67,7 +67,7 @@ func TestClusterValidateUpdateManagementValueImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementOldNilNewTrueSuccess(t *testing.T) {
-	cOld := createCluster()
+	cOld := baseCluster()
 	c := cOld.DeepCopy()
 	c.SetSelfManaged()
 
@@ -76,7 +76,8 @@ func TestClusterValidateUpdateManagementOldNilNewTrueSuccess(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementOldNilNewFalseImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{}
+	cOld := baseCluster()
+	cOld.SetManagedBy("")
 	c := cOld.DeepCopy()
 	c.SetManagedBy("management-cluster")
 
@@ -85,7 +86,7 @@ func TestClusterValidateUpdateManagementOldNilNewFalseImmutable(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementBothNilImmutable(t *testing.T) {
-	cOld := createCluster()
+	cOld := baseCluster()
 	c := cOld.DeepCopy()
 
 	g := NewWithT(t)
@@ -129,18 +130,28 @@ func TestManagementNilClusterValidateUpdateKubernetesVersionImmutable(t *testing
 }
 
 func TestWorkloadClusterValidateUpdateKubernetesVersionSuccess(t *testing.T) {
-	cOld := createCluster()
+	cOld := baseCluster()
 	cOld.SetManagedBy("management-cluster")
 
 	c := cOld.DeepCopy()
-	c.Spec.KubernetesVersion = v1alpha1.Kube120
+	c.Spec.KubernetesVersion = v1alpha1.Kube122
+
+	g := NewWithT(t)
+	g.Expect(c.ValidateUpdate(cOld)).To(Succeed())
+}
+
+func TestWorkloadClusterValidateUpdateNoUpdateSuccess(t *testing.T) {
+	cOld := baseCluster()
+	cOld.SetManagedBy("management-cluster")
+
+	c := cOld.DeepCopy()
 
 	g := NewWithT(t)
 	g.Expect(c.ValidateUpdate(cOld)).To(Succeed())
 }
 
 func TestManagementClusterValidateUpdateControlPlaneConfigurationEqual(t *testing.T) {
-	cOld := createCluster()
+	cOld := baseCluster()
 	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{
 		Count:           3,
 		Endpoint:        &v1alpha1.Endpoint{Host: "1.1.1.1/1"},
@@ -160,7 +171,7 @@ func TestManagementClusterValidateUpdateControlPlaneConfigurationEqual(t *testin
 }
 
 func TestWorkloadClusterValidateUpdateControlPlaneConfigurationEqual(t *testing.T) {
-	cOld := createCluster()
+	cOld := baseCluster()
 	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{
 		Count:           3,
 		Endpoint:        &v1alpha1.Endpoint{Host: "1.1.1.1/1"},
@@ -180,14 +191,11 @@ func TestWorkloadClusterValidateUpdateControlPlaneConfigurationEqual(t *testing.
 }
 
 func TestClusterValidateUpdateControlPlaneConfigurationImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{
-		Spec: v1alpha1.ClusterSpec{
-			ControlPlaneConfiguration: v1alpha1.ControlPlaneConfiguration{
-				Count:           3,
-				Endpoint:        &v1alpha1.Endpoint{Host: "1.1.1.1/1"},
-				MachineGroupRef: &v1alpha1.Ref{Name: "test", Kind: "MachineConfig"},
-			},
-		},
+	cOld := baseCluster()
+	cOld.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{
+		Count:           3,
+		Endpoint:        &v1alpha1.Endpoint{Host: "1.1.1.1/1"},
+		MachineGroupRef: &v1alpha1.Ref{Name: "test", Kind: "MachineConfig"},
 	}
 	c := cOld.DeepCopy()
 	c.Spec.ControlPlaneConfiguration = v1alpha1.ControlPlaneConfiguration{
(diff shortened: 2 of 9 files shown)


pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -27,7 +27,7 @@ func TestClusterDefault(t *testing.T) {
 }
 
 func TestClusterValidateUpdateManagementValueImmutable(t *testing.T) {
-	cOld := &v1alpha1.Cluster{}
+	cOld := createCluster()
```
g-gaston, line 30: I know you didn't write this, but could we rename this to just `cluster()` or `baseCluster()`. I find that name misleading since it's not creating anything :)

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -1960,6 +1802,57 @@ func TestClusterValidateUpdateSkipUpgradeImmutability(t *testing.T) {
...
+	cOld.Spec.ManagementCluster.Name = "mgmt2"
+	cOld.Spec.KubernetesVersion = "test"
+
+	cNew := cOld.DeepCopy()
+	g := NewWithT(t)
+	err := cNew.ValidateUpdate(cOld)
+	g.Expect(err).To(MatchError(ContainSubstring("parsing comparison version: could not parse \"test\" as version")))
+}
+
```
g-gaston, line 1855: Could we add a couple tests for the positive cases? no upgrade and one minor version bump
I think we are also missing a test for major version upgrade

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +352,47 @@ func (r *Cluster) ValidateDelete() error {
 
 	return nil
 }
+
+func validateServerVersionSkew(new, old *Cluster, log logr.Logger) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if old.IsSelfManaged() {
```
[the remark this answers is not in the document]
mitalipaygude: This field is immutable for management clusters, so set it for managed clusters only.

pkg/api/v1alpha1/cluster_webhook_test.go
mitalipaygude: Yes I had checked that the positive case already exists as a test case here
https://github.com/aws/eks-anywhere/blob/main/pkg/api/v1alpha1/cluster_webhook_test.go#L101

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +349,45 @@ func (r *Cluster) ValidateDelete() error {
...
+	newVersion := new.Spec.KubernetesVersion
+
+	parsedOldVersion, err := version.ParseGeneric(string(oldVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing comparison version: %v", err.Error())))
+		return allErrs
+	}
+
```
cxbrowne1207, line 372: can we do the minor version difference check right after parsing it to keep related logic nearby?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +349,45 @@ func (r *Cluster) ValidateDelete() error {
...
+	parsedNewVersion, err := version.ParseGeneric(string(newVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing comparison version: %v", err.Error())))
+		return allErrs
+	}
+
+	majorVersionDifference := math.Abs(float64(parsedNewVersion.Major()) - float64(parsedOldVersion.Major()))
+	minorVersionDifference := float64(parsedNewVersion.Minor()) - float64(parsedOldVersion.Minor())
```
cxbrowne1207, line 382: can we do the minor version difference check right after parsing it to keep related logic nearby?

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
g-gaston: But why would this validation have to know about that? This validation should be independent from the other validations if possible.

If we went and make that field mutable, how would we know that we need to update this piece too?
IMO this coupling is not necessary, but I might be missing something. What's the benefit you see on having this guard?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +346,40 @@ func (r *Cluster) ValidateDelete() error {
...
 }
+
+func validateServerVersionSkew(new, old *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if old.IsSelfManaged() {
+		return allErrs
+	}
+
+	path := field.NewPath("spec")
```
g-gaston, line 357: shouldn't this be `field.NewPath("spec", "kubernetesVersion")`?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +346,40 @@ func (r *Cluster) ValidateDelete() error {
...
+
+	oldVersion := old.Spec.KubernetesVersion
+	newVersion := new.Spec.KubernetesVersion
+
+	parsedOldVersion, err := version.ParseGeneric(string(oldVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing cluster version: %v", err.Error())))
```
g-gaston, line 366: shouldn't this be
```suggestion
			field.Invalid(path, old.Spec.KubernetesVersion, fmt.Sprintf("parsing cluster version: %v", err.Error())))
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -343,3 +346,40 @@ func (r *Cluster) ValidateDelete() error {
 
 	return nil
 }
+
+func validateServerVersionSkew(new, old *Cluster) field.ErrorList {
```
g-gaston, line 350: Can't this be reused in the CLI as well? it only takes two clusters, that's independent from the webhook

g-gaston: It seems that we still have the parsing logic duplicated

pkg/api/v1alpha1/cluster_webhook_test.go
g-gaston: That's testing minor version bump, what about no version bump?
And we seem to still miss the major version upgrade?

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,37 @@
...
+const supportedMinorVersionIncrement = 1
+
+// ValidateVersionSkew validates Kubernetes version skew between valid non-nil versions.
+func ValidateVersionSkew(oldVersion, newVersion *version.Version) error {
+	newVersionMajor := newVersion.Major()
+	newVersionMinor := newVersion.Minor()
+	oldVersionMajor := oldVersion.Major()
+	oldVersionMinor := oldVersion.Minor()
+
+	majorVersionDifference := math.Abs(float64(newVersionMajor) - float64(oldVersionMajor))
```
g-gaston, line 19: there is probably something I'm missing here, but why do we need to do a float subtraction to then abs it? major versions are already ints

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,37 @@
...
+
+	if majorVersionDifference > 0 {
+		return fmt.Errorf("MAJOR version upgrades are not supported. Major version difference between upgrade version (%d.%d) and server version (%d.%d) is %f", newVersionMajor, newVersionMinor, oldVersionMajor, oldVersionMinor, majorVersionDifference)
+	}
+
+	minorVersionDifference := float64(newVersionMinor) - float64(oldVersionMinor)
```
g-gaston, line 25: same here about floats

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,37 @@
...
+func ValidateVersionSkew(oldVersion, newVersion *version.Version) error {
+	newVersionMajor := newVersion.Major()
+	newVersionMinor := newVersion.Minor()
+	oldVersionMajor := oldVersion.Major()
+	oldVersionMinor := oldVersion.Minor()
+
+	majorVersionDifference := math.Abs(float64(newVersionMajor) - float64(oldVersionMajor))
+
+	if majorVersionDifference > 0 {
+		return fmt.Errorf("MAJOR version upgrades are not supported. Major version difference between upgrade version (%d.%d) and server version (%d.%d) is %f", newVersionMajor, newVersionMinor, oldVersionMajor, oldVersionMinor, majorVersionDifference)
```
g-gaston, line 22: curious why use capital letters for MAJOR?
Same for MINOR

g-gaston, line 22: The fact that we need to do `(%d.%d)` makes me thing that this method should probably take `anywherev1.KubernetesVersion` instead of `*version.Version`

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,37 @@
...
+// ValidateVersionSkew validates Kubernetes version skew between valid non-nil versions.
+func ValidateVersionSkew(oldVersion, newVersion *version.Version) error {
+	newVersionMajor := newVersion.Major()
+	newVersionMinor := newVersion.Minor()
+	oldVersionMajor := oldVersion.Major()
+	oldVersionMinor := oldVersion.Minor()
+
+	majorVersionDifference := math.Abs(float64(newVersionMajor) - float64(oldVersionMajor))
+
+	if majorVersionDifference > 0 {
```
g-gaston, line 21: what about `< 0`

g-gaston: Wouldn't this code be simplify if you just did `newVersion.LessThan(oldVersion)`?

pkg/validations/upgradevalidations/preflightvalidations_test.go
```diff
@@ -27,7 +26,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/validations"
 	"github.com/aws/eks-anywhere/pkg/validations/mocks"
 	"github.com/aws/eks-anywhere/pkg/validations/upgradevalidations"
-	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+	releaseanywherev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
```
g-gaston, line 29: I believe we usually call this `releasev1`

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,37 @@
...
+		return fmt.Errorf("MAJOR version upgrades are not supported. Major version difference between upgrade version (%d.%d) and server version (%d.%d) is %f", newVersionMajor, newVersionMinor, oldVersionMajor, oldVersionMinor, majorVersionDifference)
+	}
+
+	minorVersionDifference := float64(newVersionMinor) - float64(oldVersionMinor)
+
+	if minorVersionDifference < 0 {
+		return fmt.Errorf("MINOR version downgrade is not supported. Difference between upgrade version (%d.%d) and server version (%d.%d) should meet the supported version increment of +%d", newVersionMajor, newVersionMinor, oldVersionMajor, oldVersionMinor, supportedMinorVersionIncrement)
```
g-gaston, line 28: nit: this error message sounds very programmer oriented
Can it be simplified to something like "blablabla only %d minor version skew is supported blabla"

pkg/validations/upgradevalidations/versions_test.go
```diff
@@ -51,7 +63,7 @@ func TestValidateVersionSkew(t *testing.T) {
 			fileContent := test.ReadFile(t, tc.serverVersionResponse)
 			e.EXPECT().Execute(ctx, []string{"version", "-o", "json", "--kubeconfig", cluster.KubeconfigFile}).Return(*bytes.NewBufferString(fileContent), nil)
 			err := upgradevalidations.ValidateServerVersionSkew(ctx, tc.upgradeVersion, cluster, k)
-			if !reflect.DeepEqual(err, tc.wantErr) {
+			if err != nil && !reflect.DeepEqual(err.Error(), tc.wantErr.Error()) {
```
g-gaston, line 66: why use deep equal for strings?

pkg/validations/upgradevalidations/versions_test.go
```diff
@@ -43,6 +43,18 @@ func TestValidateVersionSkew(t *testing.T) {
...
+			name:                  "FailureParsingVersion",
+			wantErr:               errors.New("parsing comparison version: could not parse \"test\" as version"),
+			upgradeVersion:        "test",
+			serverVersionResponse: "testdata/kubectl_version_server_119.json",
+		},
+		{
+			name:                  "FailureParsingServerVersion",
+			wantErr:               errors.New("parsing cluster version: could not parse \"test\" as version"),
+			upgradeVersion:        v1alpha1.Kube119,
+			serverVersionResponse: "testdata/kubectl_invalid_server.json",
```
g-gaston, line 56: I'm very confused by this
Why are we passing jsons to the test and why are we adding new test cases for invalid json?

pkg/api/v1alpha1/versionskew.go
mitalipaygude: To make this method common, I am sending in the `*version.Version`. For the CLI we get the server version from the cluster which is usually of format similar to `v1.18.16-eks-1-18-4`. With the webhook however both formats are of `1.18`. We use `ParseGeneric` for webhook vs `ParseSemantic` and `ParseGeneric` for the CLI.

pkg/validations/upgradevalidations/versions_test.go
mitalipaygude: The invalid test case was added to cover the parsing cluster version error for CLI specifically. This is not entirely necessary because we usually assume valid kube version coming in from the cluster.

As for why we pass jsons is I think to mock the response received from the server. This use case is covered in the unit test where the main logic resides so this extra one can be removed.

pkg/validations/upgradevalidations/preflightvalidations_test.go
mitalipaygude: yeah i think this was an involuntary. in main this is actually named as `releasev1alpha1`

pkg/api/v1alpha1/versionskew.go
mitalipaygude: If we were to use `newVersion.LessThan(oldVersion)` we would not be able to separate out errors for major version decrement vs minor version decrement.

pkg/api/v1alpha1/versionskew.go
mitalipaygude: basically to emphasize which version is causing the issue. we were initially saying `WARNING`, which I figured instead made sense to display the version type.

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: The Kubernetes version is added for those individual fields when returning error at the end of the method.

pkg/api/v1alpha1/cluster_webhook.go
mitalipaygude: the parsing logic for getting the old server version in the CLI is using a kubectl call to get older version. with the webhooks however we can avoid this call because we have the cluster objects available.

pkg/api/v1alpha1/versionskew.go
mitalipaygude: sure, I can make it simpler.

pkg/api/v1alpha1/cluster_webhook_test.go
mitalipaygude: These 2 are covered in the `versionskew_test.go` where the main logic lies. Do we want to also test it out again for the cli and controller separately? this will kinda of have duplicate test cases.

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
mitalipaygude: Yeah the only reason was to avoid running through this at all for the self managed clusters. Instead of this I can add a check to run through the logic only if the kube versions differ from each other.

pkg/api/v1alpha1/versionskew.go
mitalipaygude: versions are actually uint. units type can not have a value of less than 0. Since we want to compare all the versions and check for conditions of negative and positive difference, we convert them to float64.

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
g-gaston: Great idea!

abhinavmpandey08 (approved):
looks good!

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,28 @@
+package v1alpha1
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/version"
+)
+
+// SupportedMinorVersionIncrement represents the minor version skew for kubernets version upgrades.
```
abhinavmpandey08, line 9: nit
```suggestion
// SupportedMinorVersionIncrement represents the minor version skew for kubernetes version upgrades.
```

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,28 @@
...
+	if newVersion.LessThan(oldVersion) {
+		return fmt.Errorf("kubernetes version downgrade is not supported (%s) -> (%s)", oldVersion, newVersion)
+	}
+
+	newVersionMinor := newVersion.Minor()
+	oldVersionMinor := oldVersion.Minor()
+
+	minorVersionDifference := int(newVersionMinor) - int(oldVersionMinor)
+
+	if minorVersionDifference != SupportedMinorVersionIncrement && minorVersionDifference != 0 {
```
abhinavmpandey08, line 23: shouldn't the `minorVersionDifference` <= `SupportedMinorVersionIncrement`
The check works in this case because `SupportedMinorVersionIncrement` is 1 but if this was changed, it would be nice to have that handled as well. Something like this:
```suggestion
	if minorVersionDifference < 0 || minorVersionDifference > SupportedMinorVersionIncrement {
```

pkg/api/v1alpha1/versionskew.go
```diff
@@ -0,0 +1,28 @@
...
+		return fmt.Errorf("only +%d minor version skew is supported", SupportedMinorVersionIncrement)
```
abhinavmpandey08, line 24: Can we also report the `minorVersionDifference` in the error?

abhinavmpandey08 (commented):
looks good! Just a couple small comments

abhinavmpandey08 (approved):
/lgtm
/approve
/woof

PR #5509, opened 2023-04-03
Remove 1.26 feature flag and set it as default version

taneyland:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


mitalipaygude:
/lgtm

Maybe we should think how we can kind of automate this feature flag setting for future versions so that we don't keep repeatedly adding and removing this similar amount of code every time.

abhinavmpandey08 (commented):
Can you also the feature flag from here? https://github.com/aws/eks-anywhere/blob/main/test/e2e/conformance_test.go#L318

abhinavmpandey08:
holding for the fix. Feel free to unhold

/hold

taneyland:
> /lgtm
> 
> Maybe we should think how we can kind of automate this feature flag setting for future versions so that we don't keep repeatedly adding and removing this similar amount of code every time.

agreed! i'll take a look into how we can do that

/unhold

abhinavmpandey08 (approved):

cxbrowne1207:
Looking good! I'll let the reviewers handle the final sign off

PR #5434, opened 2023-03-29
Generate cloudstack CP CAPI objects

tatlat:
*Issue #, if available:*
[Generate cp capi objects#971](https://github.com/aws/eks-anywhere-internal/issues/971)

*Description of changes:*

*Testing (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5544, opened 2023-04-05
Generate CloudStack Workers CAPI objects

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR generates implements `WorkerSpec` function that generates a CloudStack specific CAPI spec for an EKS-A cluster worker nodes. It talks to the cluster with a client to detect changes in immutable objects and generates new names for them.

It also refactors the `generateTemplateBuilder` function to not return an error, as these checks are already being made during cluster config validations. Additional CloudStackMachineConfig validations will also be made as a phase in the reconcile before reconciling the workers. Additionally, we've also decided to refactor the `CloudStackTemplateBuilder` in the near future. Its methods will rely only on the `cluster.Spec`, therefore, soon `generateTemplateBuilder` will no longer be necessary and will be removed.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/cloudstack/cloudstack.go
+++ b/pkg/providers/cloudstack/cloudstack.go
@@ -238,9 +238,9 @@ type ProviderKubectlClient interface {
 
 // NewProvider initializes the CloudStack provider object.
 func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, validator ProviderValidator, writer filewriter.FileWriter, now types.NowFunc, log logr.Logger) *cloudstackProvider { //nolint:revive
-	controlPlaneMachineSpec, _ := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
-	etcdMachineSpec, _ := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
-	workerNodeGroupMachineSpecs, _ := getWorkerNodeGroupMachineSpec(clusterConfig.Spec, machineConfigs)
+	controlPlaneMachineSpec := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
+	etcdMachineSpec := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
+	workerNodeGroupMachineSpecs := make(map[string]v1alpha1.CloudStackMachineConfigSpec, len(machineConfigs))
 	return &cloudstackProvider{
 		datacenterConfig:      datacenterConfig,
 		machineConfigs:        machineConfigs,
--- a/pkg/providers/cloudstack/template.go
+++ b/pkg/providers/cloudstack/template.go
@@ -8,7 +8,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/providers"
 )
 
-func generateTemplateBuilder(clusterSpec *cluster.Spec) (providers.TemplateBuilder, error) {
+func generateTemplateBuilder(clusterSpec *cluster.Spec) providers.TemplateBuilder {
 	spec := v1alpha1.ClusterSpec{
 		ControlPlaneConfiguration:     clusterSpec.Cluster.Spec.ControlPlaneConfiguration,
 		WorkerNodeGroupConfigurations: clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations,
@@ -26,7 +26,7 @@ func generateTemplateBuilder(clusterSpec *cluster.Spec) (providers.TemplateBuild
 		workerNodeGroupMachineSpecs,
 		time.Now,
 	)
-	return templateBuilder, nil
+	return templateBuilder
 }
 
 func getEtcdMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) *v1alpha1.CloudStackMachineConfigSpec {
--- a/pkg/providers/cloudstack/template.go
+++ b/pkg/providers/cloudstack/template.go
@@ -3,8 +3,6 @@ package cloudstack
 import (
 	"time"
 
-	"github.com/pkg/errors"
-
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/providers"
@@ -16,20 +14,10 @@ func generateTemplateBuilder(clusterSpec *cluster.Spec) (providers.TemplateBuild
 		WorkerNodeGroupConfigurations: clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations,
 		ExternalEtcdConfiguration:     clusterSpec.Cluster.Spec.ExternalEtcdConfiguration,
 	}
-	controlPlaneMachineSpec, err := getControlPlaneMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
-	if err != nil {
-		return nil, errors.Wrap(err, "generating control plane machine spec")
-	}
 
-	workerNodeGroupMachineSpecs, err := getWorkerNodeGroupMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
-	if err != nil {
-		return nil, errors.Wrap(err, "generating worker node group machine specs")
-	}
-
-	etcdMachineSpec, err := getEtcdMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
-	if err != nil {
-		return nil, errors.Wrap(err, "generating etcd machine spec")
-	}
+	controlPlaneMachineSpec := getControlPlaneMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
+	workerNodeGroupMachineSpecs := getWorkerNodeGroupMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
+	etcdMachineSpec := getEtcdMachineSpec(spec, clusterSpec.CloudStackMachineConfigs)
 
 	templateBuilder := NewCloudStackTemplateBuilder(
 		&clusterSpec.CloudStackDatacenter.Spec,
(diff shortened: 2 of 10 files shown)


pkg/providers/cloudstack/cloudstack.go
```diff
@@ -240,7 +240,7 @@ type ProviderKubectlClient interface {
 func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, validator ProviderValidator, writer filewriter.FileWriter, now types.NowFunc, log logr.Logger) *cloudstackProvider { //nolint:revive
 	controlPlaneMachineSpec, _ := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
 	etcdMachineSpec, _ := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
-	workerNodeGroupMachineSpecs, _ := getWorkerNodeGroupMachineSpec(clusterConfig.Spec, machineConfigs)
+	workerNodeGroupMachineSpecs := make(map[string]v1alpha1.CloudStackMachineConfigSpec, len(machineConfigs))
```
cxbrowne1207, line 243: Reverting this because it was changed in [this PR](https://github.com/aws/eks-anywhere/pull/5434). We don't want any side effects happening from this change in the CLI flow.

pkg/providers/cloudstack/template.go
```diff
@@ -59,15 +60,19 @@ func getEtcdMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[str
...
+		} else {
+			allErrs = append(allErrs, errors.Errorf("machine group ref is not provided"))
 		}
 	}
 
-	if len(workerNodeGroupMachineSpecs) == 0 {
-		return workerNodeGroupMachineSpecs, errors.Errorf("getting worker node group configs")
+	if len(allErrs) > 0 {
+		aggregate := utilerrors.NewAggregate(allErrs)
+		return workerNodeGroupMachineSpecs, errors.Errorf("getting worker node group configs: %v", aggregate)
```
cxbrowne1207, line 75: Updating to adequately capture and error if any MachineGroupRef were missing. This would pass the case where there are multiple `WorkerNodeMachineGroupConfigurations` and any of them has a complete `MachineGroupRef` defined.

pkg/providers/tinkerbell/workers.go
```diff
@@ -24,20 +24,12 @@ type (
...
-		workerTemplateNames[wc.Name] = clusterapi.WorkerMachineTemplateName(spec, wc)
-		kubeadmTemplateNames[wc.Name] = clusterapi.DefaultKubeadmConfigTemplateName(spec, wc)
-	}
-
 	templateBuilder, err := generateTemplateBuilder(spec)
 	if err != nil {
 		return nil, errors.Wrap(err, "generating tinkerbell template builder")
 	}
 
+	workerTemplateNames, kubeadmTemplateNames := clusterapi.InitialTemplateNamesForWorkers(spec)
```
cxbrowne1207, line 32: The same logic was duplicated across all our applicable providers thus far for the controller

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -238,9 +238,9 @@ type ProviderKubectlClient interface {
 
 // NewProvider initializes the CloudStack provider object.
 func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, validator ProviderValidator, writer filewriter.FileWriter, now types.NowFunc, log logr.Logger) *cloudstackProvider { //nolint:revive
-	controlPlaneMachineSpec, _ := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
-	etcdMachineSpec, _ := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
-	workerNodeGroupMachineSpecs, _ := getWorkerNodeGroupMachineSpec(clusterConfig.Spec, machineConfigs)
+	controlPlaneMachineSpec := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
+	etcdMachineSpec := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
+	workerNodeGroupMachineSpecs := make(map[string]v1alpha1.CloudStackMachineConfigSpec, len(machineConfigs))
```
cxbrowne1207, line 243: This was changed in https://github.com/aws/eks-anywhere/pull/5434, but we don't want any side effects happening from this change in the CLI flow, so reverting.

pkg/providers/cloudstack/template.go
```diff
@@ -66,20 +50,14 @@ func getWorkerNodeGroupMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConf
...
+	return workerNodeGroupMachineSpecs
 }
 
-func getControlPlaneMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) (*v1alpha1.CloudStackMachineConfigSpec, error) {
+func getControlPlaneMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) *v1alpha1.CloudStackMachineConfigSpec {
 	var controlPlaneMachineSpec *v1alpha1.CloudStackMachineConfigSpec
 	if clusterSpec.ControlPlaneConfiguration.MachineGroupRef != nil && machineConfigs[clusterSpec.ControlPlaneConfiguration.MachineGroupRef.Name] != nil {
 		controlPlaneMachineSpec = &machineConfigs[clusterSpec.ControlPlaneConfiguration.MachineGroupRef.Name].Spec
-	} else {
-		return controlPlaneMachineSpec, errors.Errorf("getting MachineGroupRef")
```
cxbrowne1207, line 81: The cluster config is already being validated [here](https://github.com/aws/eks-anywhere/blob/c98621273ade48012679a66f9bcc62e72f40a49d/pkg/api/v1alpha1/cluster.go#L384), so there is no need for this

pkg/providers/cloudstack/template.go
```diff
@@ -66,20 +50,14 @@ func getWorkerNodeGroupMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConf
 		}
 	}
 
-	if len(workerNodeGroupMachineSpecs) == 0 {
-		return workerNodeGroupMachineSpecs, errors.Errorf("getting worker node group configs")
-	}
```
cxbrowne1207, line 71: The cluster config is already being validated [here](https://github.com/aws/eks-anywhere/blob/c98621273ade48012679a66f9bcc62e72f40a49d/pkg/api/v1alpha1/cluster.go#L384), so there is no need for this

pkg/providers/cloudstack/template.go
```diff
@@ -41,22 +29,18 @@ func generateTemplateBuilder(clusterSpec *cluster.Spec) (providers.TemplateBuild
...
+func getEtcdMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) *v1alpha1.CloudStackMachineConfigSpec {
 	var etcdMachineSpec *v1alpha1.CloudStackMachineConfigSpec
 	if clusterSpec.ExternalEtcdConfiguration != nil {
 		if clusterSpec.ExternalEtcdConfiguration.MachineGroupRef != nil && machineConfigs[clusterSpec.ExternalEtcdConfiguration.MachineGroupRef.Name] != nil {
 			etcdMachineSpec = &machineConfigs[clusterSpec.ExternalEtcdConfiguration.MachineGroupRef.Name].Spec
-		} else {
-			return etcdMachineSpec, errors.Errorf("getting MachineGroupRef")
 		}
-	} else {
-		return etcdMachineSpec, errors.Errorf("getting external etcd config")
```
cxbrowne1207, line 53: The cluster config is already being validated [here](https://github.com/aws/eks-anywhere/blob/c98621273ade48012679a66f9bcc62e72f40a49d/pkg/api/v1alpha1/cluster.go#L384), so there is no need for this

pkg/providers/cloudstack/template.go
taneyland: This validation only checks that it isn't nil, so we could still see an error here if they don't specify a machineGroupRef name. But to resolve this, we just have to be sure to add this to the CloudStack reconciler when we add the additional validations: https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/validator.go#L113 , we do the same thing for vsphere

pkg/providers/cloudstack/template.go
```diff
@@ -66,20 +50,14 @@ func getWorkerNodeGroupMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConf
...
-	}
-
-	return workerNodeGroupMachineSpecs, nil
+	return workerNodeGroupMachineSpecs
 }
 
-func getControlPlaneMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) (*v1alpha1.CloudStackMachineConfigSpec, error) {
+func getControlPlaneMachineSpec(clusterSpec v1alpha1.ClusterSpec, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) *v1alpha1.CloudStackMachineConfigSpec {
 	var controlPlaneMachineSpec *v1alpha1.CloudStackMachineConfigSpec
 	if clusterSpec.ControlPlaneConfiguration.MachineGroupRef != nil && machineConfigs[clusterSpec.ControlPlaneConfiguration.MachineGroupRef.Name] != nil {
```
taneyland, line 58: Should we remove this first if case then since it'll never be nil at this point?

pkg/providers/cloudstack/template.go
cxbrowne1207: Yes, that is the plan.

Additionally, I'm mainly removing this here because then they would writing tests for these cases. This would be wasted effort because we're planning to simplify the `templateBuilder`, such that these functions would no longer be necessary just like with vSphere

pkg/providers/cloudstack/template.go
cxbrowne1207: I don't think this is necessary for now.

To clarify, the main reason I refactored this is to avoid the extra work of writing a test to cover these cases that would eventually be removed soon once the `CloudStackTemplateBuilder` has been simplified.

taneyland: ok sounds good!

taneyland (approved):

PR #5569, opened 2023-04-10
Fix regauth FLC

pokearu:
*Description of changes:*
When customers set up authenticated registry mirrors, the eks-a controller does not have access to the required env variables. Hence adding logic to read from registry-auth secret and setting Env on the controller pod.

*Testing:*
Built the eks-a controller and pushed to private registry. Tested changes on vsphere.


abhinavmpandey08 (approved):

mitalipaygude:
We might want to add a few unit tests for newly added code, but /lgtm otherwise!

PR #5574, opened 2023-04-10
Add notimeouts option for cilium

jiayiwang7:
*Issue #, if available:*

Part of #5565

*Description of changes:*

Respect `--no-timeouts` in cilium installer.

*Testing (if applicable):*

Unit tests


pkg/dependencies/factory.go
```diff
@@ -144,6 +145,10 @@ type executablesConfig struct {
 	mountDirs          []string
 }
 
+type timeoutConfig struct {
```
g-gaston, line 148: since there is only one timeout setting, maybe just call it `config`?
Up to you, no strong opinion

PR #5614, opened 2023-04-12
Refactor e2e for vSphere to use table driven tests:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This seeks to optimize for readability and less duplication of code. This doesn't give us an increase in the reliability of these tests. Also, this change seeks to teach about table driven testing.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/vsphere_test.go
+++ b/test/e2e/vsphere_test.go
@@ -17,181 +17,63 @@ import (
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
-// Autoimport
-func TestVSphereKubernetes122BottlerocketAutoimport(t *testing.T) {
-	provider := framework.NewVSphere(t,
-		framework.WithVSphereFillers(
-			api.WithTemplateForAllMachines(""),
-			api.WithOsFamilyForAllMachines(v1alpha1.Bottlerocket),
-		),
-	)
-	test := framework.NewClusterE2ETest(
-		t,
-		provider,
-		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube122)),
-	)
-	runAutoImportFlow(test, provider)
+// Auto Import tests
+func TestBottlerocketAutoImport(t *testing.T) {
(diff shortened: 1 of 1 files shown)


jaxesn (commented):
I like this train of thought! Id like to see how this could look across providers... in my simplistic view, ideally we have could have 1 set of tests that we want to run across every provider (providers can and will have their own specific ones too) to ensure that when we add new critical use cases, or even new providers, they are running across all without requiring engineers to remember to add everywhere.

jacobweinstock:
Looks like i have not taken into account how we allocate clusters for testing. Looks like we have tests as single functions so that we can parse them out for properly assigning clusters to tests in a serial way.

https://github.com/aws/eks-anywhere/blob/main/internal/test/e2e/list.go#L14

pokearu:
> Looks like i have not taken into account how we allocate clusters for testing. Looks like we have tests as single functions so that we can parse them out for properly assigning clusters to tests in a serial way. It also allows us to specify tests to skip at runtime. This allows for non code changes to skip tests.
> 
> https://github.com/aws/eks-anywhere/blob/main/internal/test/e2e/list.go#L14

That’s right. I don’t think we can leverage table tests, that would force all the tests to run in a single EC2. We had the tests as individual ones as they can run in parallel in their own environments.

jaxesn:
Let's think of other options though what we have now where we have to rely on keeping things in sync across versions and providers is less than ideal. Just to throw out something wild, what about just auto generating the test files from a known template so at least we could keep them in sync?

jacobweinstock:
I like the wild idea, @jaxesn ! We should discuss more. I found that running or skipping table driven tests (which in pure Go are just sub-tests) is no different than what we currently do. See this link for sub-test running: https://go.dev/blog/subtests
I believe the only thing left to solve, that would allow for table driven tests to be used, is that the way we determine all tests is by using `./bin/e2e.test -test.list` and this doesn't print out sub-tests. I'll keep investigating to see if i can find a solution.

PR #5620, opened 2023-04-13
Fix notimeouts for cilium

jiayiwang7:
*Issue #, if available:*

Part of #5565 

*Description of changes:*

When adding the networking dependency in factory, the `networkingBuilder` was called out of the `f.buildSteps` causing networking builder to be executed before `WithNoTimeouts` are applied. Fix the issue by moving `networkingBuilder` to the `buildSteps`, so that the build happens in the end when calling `f.Build()`.

*Testing (if applicable):*

Unit tests


jonathanmeier5 (approved):
/lgtm

jonathanmeier5 (approved):
/lgtm

g-gaston (approved):

PR #5622, opened 2023-04-13
Add design for workload cluster upgrade isolation

g-gaston:

### DIFF
--- a/designs/isolate-cluster-upgrades.md
+++ b/designs/isolate-cluster-upgrades.md
@@ -1,0 +1,276 @@
+# Isolate workload clusters from management cluster upgrades
+
+## Problem statement
+
+When a management cluster is upgraded, it's management components are also upgraded. This includes the EKS-A controller manager. Given this controller contains the logic to "convert" the EKS-A to the cluster-api API, any changes to this logic between one EKS-A version to the next one can trigger unexpected (for the user) cascading changes in the workload clusters. In some scenarios, those changes can cause machine rollouts.
+
+### The EKS-A cluster controller
+Reducing to the simples expression, the EKS-A controller follows this logic:
+1. Read the cluster config (EKS-A `Cluster` object plus all linked objects).
+1. Using the cluster config as input, generate the `cluster-api` cluster spec (the combination of all CAPI custom resources that define a cluster).
+1. Upsert these CAPI objects (create them if they don't exist, update otherwise).
+
+```mermaid
+graph TB
+    subgraph config [Cluster config]
+        direction TB
+        cluster{Cluster} -.-> dc[Datacenter]
+        dc -.-> creds[Credentials Secret]
+        cluster -.-> mc1[Machine Config 1]
+        cluster -.-> mc2[Machine Config 2]
+    end
+
+    config -- input --> generation[CAPI generation logic]
+
+    subgraph capi ["cluster-api spec"]
+        direction TB
+        capicluster[Cluster] -.-> kcp[KCP]
+        kcp -.-> cpmc[Machine Config]
+        md1[Machine Deployment 1] -.-> mdc1[Machine Config]
+        md1[Machine Deployment 2] -.-> mdc2[Machine Config]
+    end
+
+    generation  -- output --> capi
+
+    capi  -- input --> client[Client]
+    
+    client  -- upsert --> apiserver[Kube api-server]
+```
+For more detailed information, consult the [controller design doc](https://github.com/aws/eks-anywhere/blob/main/designs/full-cluster-lifecycle-api.md).
+
+### When is the issue more likely to happen?
+
+Any new feature we add that involves API changes is usually opt-in, or at least not enabled by default in existing clusters. This makes easier to write the generation logic in a way that produces the same output for a particular input, since we can count on the input to change before having to "configure" the new feature on the CAPI side.
+
+However, it's with defaults where this issue becomes more relevant. Not necessarily with defaults on the EKS-A API but with defaults we set on the CAPI side that are not a direct consequence of a particular EKS-A input. We could call this the base CAPI spec. These are the values that are set regardless of the EKS-A API input. Although these don't change very often, when they do, it becomes difficult to make those changes opt-in.
+### Has this happened before?
+There have been a few times where a change in the CAPI "generation logic" caused unwanted rollouts in workload clusters.
+
+For example, we made a change to the logic that generates the [KCP for Cloudstack](https://github.com/aws/eks-anywhere/pull/3566) in order to set correctly the hostname. Although a valid and useful addition, this change affects the bootstrap data which requires a machine rollout in order to reconcile the change.
+
+Another example could be [this](https://github.com/aws/eks-anywhere/pull/3402) (although it was fixed pre-release so the bug was never shipped). This time, we wanted to change the content of one of the files we copy to the machines. An API version changed in between Kubernetes versions and we just updated to newest one. Again, since this file is part of the bootstrap data, new machines needed to be rolled out to reconcile the changes.
+
+### Will this happen again?
+Although we should strive to make all new CAPI config "opt-in", so it only affects new cluster or cluster where the user specifically request such changes, it's possible that we might find situations where this is not possible. As mentioned before, the "base cluster-api spec" is where this is most likely to happen.
+
+Is for those edge cases where there is no other way but to change the cluster-api output that this design is trying to find a solution for. 
+
+## Tenets
+1. **Simple**. Prioritize (in order): simple for the user, simple to maintain, simple to implement.
+
+## Goals
+1. As a cluster operator, I want to upgrade my management clusters without new machines being rolled out in the workload clusters.
+2. As a cluster operator, I want EKS-A to fully take care of the upgrade process, I don't want to have to manually update the workload clusters before I upgrade the management cluster.
+
+## Scope
+### In
+* Design a system that prevents, working transparently to the user, new machines from being rolled out in workload clusters when the management cluster is upgraded.
+### Out
+* Decouple management components update from cluster components upgrade in management clusters.
+
+## Solution
+Use the Kubernetes concepts of `generation` and `observedGeneration` to determine when an cluster configuration has changed and only then, reconcile the state of the world to the desired state. This prevents workload clusters from being affected by a management cluster upgrade: if a workload cluster config has already been reconciled and its `generation` matches its `observedGeneration`, no changes will be applied to the CAPI definition of the cluster unless the EKS-A cluster config changes again, regardless if the CAPI cluster definition generated for the same EKS-A cluster config has changed between controller versions.
+
+### `generation` and `observedGeneration`
+* [`metadata.generation`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata): a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.
+	* [For CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource): The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
+* [`status.observedGeneration`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties): is the `generation` most recently observed by the component responsible for acting upon changes to the desired state of the resource. In other words, the value of the `observedGeneration` field indicates the last time the controller updated the status subresource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.
+
+### Workflow: how to use `observedGeneration` to only reconcile on changes
+This diagram represents the logic to follow in the EKS-A `Cluster` controller:
+
+```mermaid
+stateDiagram-v2
+    [*] --> Queue: Cluster reconcile request
+    
+    deleted: Cluster deleted?
+    delete: Reconcile delete
+    reconcile: Reconcile cluster
+    observedGeneration: generation == observedGeneration
+    updateGeneration: observedGeneration = generation
+    done: Reconciliation done?
+
+    Queue --> deleted
+    state deleted <<choice>> 
+        deleted --> observedGeneration: no
+        deleted --> delete: yes
+    state observedGeneration <<choice>>
+        observedGeneration --> reconcile: no
+        observedGeneration --> [*]: yes
+    delete --> done
+    reconcile --> done
+    state done <<choice>>
+        done --> updateGeneration: yes
+        done --> Queue: no
+    updateGeneration  --> [*]
+```
+
+We only perform the actual reconciliation logic (this is trying to change the state of the world) when `observedGeneration` is different than `generation`. This will only happen after the resource has changed.
+
+It's important to mention that we only update the value of `observedGeneration` when the reconciliation process is "done". Done here means that the state of the world is equal to the desired state: the cluster has been completely reconciled. This is to take into account the case where a full reconciliation takes multiple loops (for example, when the cluster is first created). Note that "reconcile not done" also includes errors.
+
+This might be slightly different of how other resources use the `observedGeneration` field, where the field is always set on every reconciliation loop. For multi state resources (like our `Cluster`) this is usually accompanied of a "state" indicator (note that `phases` is now discouraged in favor of `conditions`). However, we don't have that in our `Cluster`'s status. Although tempting to try to add this to the status, we encourage to tackle this problem separately and revisit the use of `observedGeneration` if necessary. Refer to the [alternatives](#alternative-solutions) section for other possible solutions.
+
+### Cluster vs cluster config: how to calculate an aggregated generation
+An astute reader might have realized that until now we have worked with a simplified model: one where cluster reconciliation only needs to occur when the resource been reconciled (the Cluster in this case) changes. Although this is true for most controllers in Kubernetes, our Cluster controller doesn't behave this way. Due to the particularities of our API and how we define a single entity (the cluster config) with multiple CRDs (all linked through the `Cluster` object), the Cluster controller reacts to changes not only in the `Cluster` but also in all its linked objects (we'll call them child objects for the rest of the document).
+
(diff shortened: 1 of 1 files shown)


designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+	// childrenObservedGeneration represents the sum of all cluster's linked
+	// objects's .metadata.generation the last time the cluster was successfully reconciled.
+	ChildrenObservedGeneration int64 `json:"childrenObservedGeneration,omitempty"`
+}
+```
+
+### Version skew
+Although it's possible today to continue upgrading a management cluster without updating the workload cluster's `Bundles` (the versions of all components in a EKS-A cluster), we lack a maximum skew version between management and workload clusters.
+
+Since this proposal makes the isolation between management and workload clusters even bigger, we propose to add to our documentation a maximum skew version support of one minor version. This means the management cluster can only be one EKS-A minor version after the workload clusters.
```
jiayiwang7, line 122: why does the proposal requires "the management cluster can only be one EKS-A minor version after the workload clusters."?

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Risks and doors
+Although it involves API changes, and as any other API change it should be approached with caution, we consider the proposed solution a two way door:
+
+* It's an internal mechanism not exposed to users that we are free to change as we please.
+* Even if this document is lengthy, the solution should be quite simple to implement, both in length and scope (overall codebase footprint). This should minimize the effort required to change it for a different solution.
+
+We can document the new API fields as internal and subject to change their meaning in future releases.
+
+#### Drift detection
```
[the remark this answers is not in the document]
g-gaston: Great question
The reasoning is we don't support manually manipulating CAPI objects. So loosing the ability to detect those is not necessarily a feature/capability that users can expect.
On the contrary, we do support upgrading management clusters and user should expect to be able to do that without needing to roll new machines for workload clusters.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+### `generation` and `observedGeneration`
+* [`metadata.generation`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata): a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.
+	* [For CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource): The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
+* [`status.observedGeneration`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties): is the `generation` most recently observed by the component responsible for acting upon changes to the desired state of the resource. In other words, the value of the `generation` field the last time the controller updated the status subresource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.
```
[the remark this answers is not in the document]
g-gaston:
```suggestion
* [`status.observedGeneration`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties): is the `generation` most recently observed by the component responsible for acting upon changes to the desired state of the resource. In other words, the value of the `observedGeneration` field indicates the last time the controller updated the status subresource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.
```

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
+# Isolate workload clusters from management cluster upgrades
+
+## Introduction
+When a management cluster is upgraded, it's management components are also upgraded. This includes the EKS-A controller manager. Given this controller contains the logic to "convert" the EKS-A to the cluster-api API, any changes to this logic between one EKS-A version to the next one can trigger a cascading effect in the workload clusters after upgrading the management cluster. In some scenarios, this can cause unexpected machine rollouts.
```
g-gaston, line 4: This needs better explanation and/or wording

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+It's important to mention that we only update the value of `observedGeneration` when the reconciliation process is "done". Done here means that the state of the world is equal to the desired state: the cluster has been completely reconciled. This is to take into account the case where a full reconciliation takes multiple loops (for example, when the cluster is first created). Note that "reconcile not done" also includes errors.
+
+This might be slightly different of how other resources use the `observedGeneration` field, where the field is always set on every reconciliation loop. For multi state resources (like our `Cluster`) this is usually accompanied of a "state" indicator (note that `phases` is now discouraged in favor of `conditions`). However, we don't have that in our `Cluster`'s status. Although tempting to try to add this to the status, we encourage to tackle this problem separately and revisit the use of `observedGeneration` if necessary. Refer to the [alternatives](#alternative-solutions) section for other possible solutions.
```
vivek-koppuru, line 60: I missed where you talk about using status in alternatives, unless that wasn't what you meant. Is it possible to talk about what is the additional effort required to use `observedGeneration` along with `ClusterStatus` because it seems like that might be the approach we want to go towards anyways.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Cluster vs cluster config: how to calculate an aggregated generation
+An astute reader might have realized that until now we have worked with a simplified model: where cluster reconciliation only needs to occur when the resource been reconciled `Cluster` object changes. Although this is true for most controllers in Kubernetes, our Cluster controller doesn't behave this way. Due to the particularities of our API and how we define a single entity (the cluster config) with multiple CRDs (all linked through the `Cluster` object), the Cluster controller reacts to changes not only in the `Cluster` but also in all its linked objects (we'll call them child objects for the rest of the document).
```
vivek-koppuru, line 63: I am confused by "when the resource been reconciled `Cluster` object changes". Are you saying it only needs to occur when `Cluster` object changes or when it has already been reconciled and then it changes?

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+Now, the question becomes: how do we reflect child objects changes in the `observedGeneration`? Since all our CRDs, including the child objects, have a `generation` field, we can simply add all of them into one aggregated cluster config `observed`. The are two different ways for a referenced object to change:
+* If a child object reference is changed to a different object, `Cluster` object's `generation` will change, increasing the aggregated cluster config's generation.
+* If the reference is kept pointing to the same object, but the child object is changed, this will increase its `generation`, which also increases the aggregated cluster config's generation.
+$$aggregatedGeneration = \sum_{n=1}^{clusterConfigObjs} generation_n$$
```
vivek-koppuru, line 68: Are you saying this a new field or using the same `observedGeneration` field in cluster config?

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+```go
+type ClusterStatus struct {
+	...
+	// observedGeneration represents the .metadata.generation the last time the
+	// cluster was successfully reconciled.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// childrenObservedGeneration represents the sum of all cluster's linked
+	// objects's .metadata.generation the last time the cluster was successfully reconciled.
+	ChildrenObservedGeneration int64 `json:"childrenObservedGeneration,omitempty"`
```
vivek-koppuru, line 115: Is this field what tracks aggregate? What about always incrementing `generation` whenever a child object changes, would that remove the need to track this here? Not sure if I missed this above.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+#### Pros
+* Allows drift detection.
+* No unexpected rollouts.
+* Potentially infinite extensibility to support a bigger version skew.
+#### Cons
+* Increases codebase maintenance work. When support for a certain version is removed, that version's "layer" logic should be incorporated into the base generation logic so the complexity doesn't grow forever. This can also be prone to regressions.
+* It becomes increasingly difficult to navigate the logic that goes from EKS-A input to CAPI output. The code is the single source of truth for this process and it becomes vital when trying to understand the system's behavior (when debugging an issue or simply trying to add a new feature). This approach makes building this mental model more difficult.
+* It requires more testing, since each version (layer) requires its own set of tests.
+
+It's important to note that this approach only works when generating CAPI objects using go structs. So it becomes almost a pre-requisite to move away from the yaml + go templates combo that EKS-A currently uses.
```
vivek-koppuru, line 214: Yea we need to move towards this anyways

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+	// cluster was successfully reconciled.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// childrenObservedGeneration represents the sum of all cluster's linked
+	// objects's .metadata.generation the last time the cluster was successfully reconciled.
+	ChildrenObservedGeneration int64 `json:"childrenObservedGeneration,omitempty"`
+}
+```
+
+### Version skew
```
vivek-koppuru, line 119: Is there a possibility of management cluster being one version older than the workload clusters? I assume that might have to do with if there are api changes for other components that would make it incompatible?

I am asking due to what we introduced in https://github.com/aws/eks-anywhere/pull/5263 where we set the expectation that management clusters need to be upgraded before workload clusters, but that expectation may or may not affect this design. Any thoughts around how this changes if we were to open up the restriction in the future where we maybe support an older version of a management cluster with a newer version of a workload cluster?

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+	return spec
+}
+```
+
+#### Pros
+* Allows drift detection.
+* No unexpected rollouts.
+* Potentially infinite extensibility to support a bigger version skew.
+#### Cons
```
jiayiwang7, line 209: Overall I agree that we should go with the proposed solution, since it requires least effort and solve the upgrade problem. Although from pure user/product perspective, this layered solution is more flexible regarding the version it can support, plus that it solves the upgrade problem + does not introduce a problem (drift detection).

I think at least the team can discuss about which solution makes more sense: approach w heavy dev cost and maintenance or simpler solution which also introduce a problem that user can be affected.

designs/isolate-cluster-upgrades.md
g-gaston: ah this is bad writing
I should say:
> An astute reader might have realized that until now we have worked with a simplified model: one where cluster reconciliation only needs to occur when the resource been reconciled (the `Cluster` in this case) changes.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
+# Isolate workload clusters from management cluster upgrades
+
+## Introduction
```
d8660091, line 3: Would machine rollout being triggered when you upgrade a cluster-api controller? If so, adding "generation" would not prevent this kind of machine rollout from happening

designs/isolate-cluster-upgrades.md
g-gaston: At this point this is talking about setting `observedGeneration` with this calculation.

designs/isolate-cluster-upgrades.md
g-gaston: The problem is `generation` is managed by kubernetes, we don't control its value.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+### Out
+* Decouple management components update from cluster components upgrade in management clusters.
+
+## Solution
+Use the Kubernetes concepts of `generation` and `observedGeneration` to determine when an cluster configuration has changed and only then, reconcile the state of the world to the desired state. This prevents workload clusters from being affected by a management cluster upgrade: if a workload cluster config has already been reconciled and its `generation` matches its `observedGeneration`, no changes will be applied to the CAPI definition of the cluster unless the EKS-A cluster config changes again, regardless if the CAPI cluster definition generated for the same EKS-A cluster config has changed between controller versions.
+
+### `generation` and `observedGeneration`
+* [`metadata.generation`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata): a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.
```
d8660091, line 23: Curious, can a user modify this "generation" or "observedGeneration" field explicitly?

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+        observedGeneration --> [*]: yes
+    delete --> done
+    reconcile --> done
+    state done <<choice>>
+        done --> updateGeneration: yes
+        done --> Queue: no
+    updateGeneration  --> [*]
+```
+
+We only perform the actual reconciliation logic (this is trying to change the state of the world) when `observedGeneration` is different than `generation`. This will only happen after the resource has changed.
```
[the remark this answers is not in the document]
g-gaston: We don't have that field yet

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+### `generation` and `observedGeneration`
+* [`metadata.generation`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata): a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.
+	* [For CRDs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource): The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
```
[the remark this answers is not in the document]
g-gaston: v1 to v2 requires creating a new object
So the new object will have a generation of 1 and once reconciled, an observedGeneration of 1

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+* If a child object reference is changed to a different object, `Cluster` object's `generation` will change, increasing the aggregated cluster config's generation.
+* If the reference is kept pointing to the same object, but the child object is changed, this will increase its `generation`, which also increases the aggregated cluster config's generation.
+$$aggregatedGeneration = \sum_{n=1}^{clusterConfigObjs} generation_n$$
+
+
+#### Hidden `generation` changes
+Although it covers most of the scenarios, this solution still leaves in edge case unsolved. While `generation` can only be increased for a certain object, by changing a child object reference to a different object we can make the aggregated generation go down, or even worse, stay the same. The latter is bad since we will identify that reconcile request as a no-op, hence no reacting to the changes. This can happen if a referenced gets changed to child object with a $generation = prevObjGeneration -1$. That decreased will be compensated by the increase in the `Cluster`'s generation.
```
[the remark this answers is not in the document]
g-gaston: right, but every object has a different monotonic sequence
So if we change the link to a different object, we start aggregating a different sequence

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+### Risks and doors
+Although it involves API changes, and as any other API change it should be approached with caution, we consider the proposed solution a two way door:
+
+* It's an internal mechanism not exposed to users that we are free to change as we please.
+* Even if this document is lengthy, the solution should be quite simple to implement, both in length and scope (overall codebase footprint). This should minimize the effort required to change it for a different solution.
+
+We can document the new API fields as internal and subject to change their meaning in future releases.
+
+#### Drift detection
+The biggest disadvantage of this solution is losing the ability to detect drift: if a user, or any other entity with enough permissions, changes the CAPI objects directly (hence affecting the real state of the cluster) we won't be able to react to this until any of the EKS-A objects composing the cluster config are updated. Although the controller shouldn't have any problem reconciling this new state of the cluster once the EKS-A cluster config is updated, this behavior is not ideal since it in a way breaks the expected Kubernetes control logic.
```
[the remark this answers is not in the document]
g-gaston: First point: agreed, but this is already present today and I believe it's something it should be tackled independently from this doc. I think the proposed design is compatible with more configuration in thee API.


Second point: agreed, we should clearly document it. But we don't support updating capi objects, user shouldn't care about the behavior if they do that, because they shouldn't do it in the first place. And if they do, there are no guarantees.

designs/isolate-cluster-upgrades.md
g-gaston: Agreed, we need alignment here.

designs/isolate-cluster-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+## Tenets
+1. **Simple**. Prioritize (in order): simple for the user, simple to maintain, simple to implement.
+
+## Goals
+1. As a cluster operator, I want to upgrade my management clusters without new machines being rolled out in the workload clusters.
+2. As a cluster operator, I want EKS-A to fully take care of the upgrade process, I don't want to have to manually update the workload clusters before I upgrade the management cluster.
+
+## Scope
+### In
+* Design a system that prevents, working transparently to the user, new machines from being rolled out in workload clusters when the management cluster is upgraded.
```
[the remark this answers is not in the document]
g-gaston: Perhaps, but it's also less clear IMO. Because it's not when the cluster has been upgraded, it's **just after it has been upgraded**. Not string opinion here, open to better wording.

designs/isolate-cluster-upgrades.md
g-gaston: Not sure about generation, since that's managed by kubernetes. But it's possible.

They can definitely update `observedGeneration` and the controller will revert it.

designs/isolate-cluster-upgrades.md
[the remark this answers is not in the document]
jiayiwang7: But then how would the generation go down?

jiayiwang7 (approved):

PR #5642, opened 2023-04-14
Add timeout to the e2e command

abhinavmpandey08:
*Description of changes:*
Add timeout to the e2e command. This makes sure that the e2e test command itself times out instead of the SSM command timing out. With this change, we will be able to get the results of the tests that finished even if the instance timed out.

### DIFF
--- a/internal/pkg/ssm/command.go
+++ b/internal/pkg/ssm/command.go
@@ -2,6 +2,8 @@ package ssm
 
 import (
 	"fmt"
+	"math"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -17,9 +19,10 @@ const ssmLogGroup = "/eks-anywhere/test/e2e"
 
 var initE2EDirCommand = "mkdir -p /home/e2e/bin && cd /home/e2e"
 
-func WaitForSSMReady(session *session.Session, instanceId string) error {
+// WaitForSSMReady waits for the SSM command to be ready.
+func WaitForSSMReady(session *session.Session, instanceID string, timeout time.Duration) error {
 	err := retrier.Retry(10, 20*time.Second, func() error {
-		return Run(session, logr.Discard(), instanceId, "ls")
+		return Run(session, logr.Discard(), instanceID, "ls", timeout)
 	})
 	if err != nil {
 		return fmt.Errorf("waiting for ssm to be ready: %v", err)
@@ -53,8 +56,9 @@ var nonFinalStatuses = map[string]struct{}{
 	ssm.CommandInvocationStatusInProgress: {}, ssm.CommandInvocationStatusDelayed: {}, ssm.CommandInvocationStatusPending: {},
 }
 
-func Run(session *session.Session, logger logr.Logger, instanceId, command string, opts ...CommandOpt) error {
-	o, err := RunCommand(session, logger, instanceId, command, opts...)
+// Run runs the command using SSM on the instance corresponding to the instanceID.
+func Run(session *session.Session, logger logr.Logger, instanceID, command string, timeout time.Duration, opts ...CommandOpt) error {
+	o, err := RunCommand(session, logger, instanceID, command, timeout, opts...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


internal/pkg/ssm/command.go
```diff
@@ -94,7 +94,7 @@ func RunCommand(session *session.Session, logger logr.Logger, instanceId, comman
 
 	logger.Info("Waiting for ssm command to finish")
 	var commandOut *ssm.GetCommandInvocationOutput
-	r := retrier.New(300*time.Minute, retrier.WithMaxRetries(2160, 60*time.Second))
+	r := retrier.New(310*time.Minute, retrier.WithMaxRetries(2160, 60*time.Second))
```
abhinavmpandey08, line 97: Making SSM timeout slightly longer than e2e timeout so e2e timeout gets triggered before SSM timeout

internal/test/e2e/run.go
```diff
@@ -205,7 +205,7 @@ func (e *E2ESession) runTests(regex string) (testCommandResult *testCommandResul
 	command := "GOVERSION=go1.16.6 gotestsum --junitfile=junit-testing.xml --raw-command --format=standard-verbose --hide-summary=all --ignore-non-json-output-lines -- test2json -t -p e2e ./bin/e2e.test -test.v"
 
 	if regex != "" {
-		command = fmt.Sprintf("%s -test.run \"%s\"", command, regex)
+		command = fmt.Sprintf("%s -test.run \"%s\" -test.timeout 300m", command, regex)
```
abhinavmpandey08, line 208: This sets the e2e timeout to 300 minutes for each e2e test instance

internal/pkg/ssm/command.go
g-gaston: Can we make this configurable?
That way we can pass down the timeouts from the a central place.
If not, these two numbers need to be in sync in the two files, it's a hidden dep. Someone might go and change one of them not knowing about this dependency.

g-gaston, line 97: I think we also need to update the `SendCommandInput`
I think we want `go test timeout` < `SSM command timeout` < `timeout in retrier that pulls the SSM command results`.
That guarantees the SSM command always gets the test results and that we always get the results of the SSM command.

g-gaston (commented):
Great idea!

internal/test/e2e/run.go
cxbrowne1207: Does this cover the duration of the individual tests or all the tests being run? If it's the timeout for all the tests, should we make it 6 hours? 5.5 is pretty close to the normal time it's been taking to run a few of the provider test durations.

internal/pkg/ssm/command.go
abhinavmpandey08: done

abhinavmpandey08: Is this what you meant? https://github.com/aws/eks-anywhere/pull/5642/files#diff-ec745d2a43e1da6e449129ed0106f64d40ac71e38690f7970f0d3d440ed44889R125-R130

internal/test/e2e/run.go
```diff
@@ -217,6 +225,7 @@ func (e *E2ESession) runTests(regex string) (testCommandResult *testCommandResul
 		e.logger.V(4),
 		e.instanceId,
 		command,
+		defaultSSMTimeout,
```
g-gaston, line 228: maybe we can do 
```suggestion
		e2eTimeout + e2eSSMTimeoutPadding,
```

and having that as a const (whichever value you decide)
```go
e2eSSMTimeoutPadding = 10 * Minute
```

internal/pkg/ssm/command.go
```diff
@@ -94,7 +98,7 @@ func RunCommand(session *session.Session, logger logr.Logger, instanceId, comman
 
 	logger.Info("Waiting for ssm command to finish")
 	var commandOut *ssm.GetCommandInvocationOutput
-	r := retrier.New(300*time.Minute, retrier.WithMaxRetries(2160, 60*time.Second))
+	r := retrier.New(timeout, retrier.WithMaxRetries(2160, 60*time.Second))
```
g-gaston, line 101: I would make this one slightly bigger
```suggestion
    // Making the retrier wait for longer than the provided SSM timeout to make sure
    // we always get the output results.
	r := retrier.New(timeout + 5 * time.Minute, retrier.WithMaxRetries(2160, 60*time.Second))
```

Now that I think about it, maybe we need to change the option as well

```suggestion
	// Making the retrier wait for longer than the provided SSM timeout to make sure
    // we always get the output results.
	r := retrier.New(timeout + 5 * time.Minute, retrier.WithMaxRetries(math.MaxInt, 60*time.Second))
```
That should guarantee "infinite" (not really) retries, so we only error out if max timeout is reached.

g-gaston (approved):

PR #5674, opened 2023-04-19
Simplify CloudStack template builder 

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
The [CloudStackTemplateBuilder](https://github.com/aws/eks-anywhere/blob/2ca366517b41fede3256ec0e47e5de37b9f59a2a/pkg/providers/cloudstack/cloudstack.go#L633) has dependencies on `controlPlaneMachineSpec`, `WorkerNodeGroupMachineSpecs`, and `etcdMachineSpecs`. According to [here](https://github.com/aws/eks-anywhere/blob/50ee3e1e0705b769039fdff04d53e471e198dd50/controllers/resource/template.go#L181), it seems as though it was made prohibiting control plane and ETCD updates, and these provided arguments help to override what's in the provided spec. Given we are working on the new controller, we don't need this limitation anymore, it also makes the code made the code more complex. 

This PR does the following:
1. Removes dependencies from the CloudStackTemplateBuilder, simplifying it to make use the API structs in the provided cluster.Spec instead of maintaining it's own versions of the data.
2. It moves SSH key normalization to the template builder level as opposed to during defaulting. This way user comments in the SSH keys will be kept in the CloudStackMachineConfigs as entered by the user but removed from machine templates. Also, now it's only done in one place.
3. Adds logic to assign a default username for the CloudStackMachineConfig user. If there were no users specified, the logic here would default the SSH key for the default user but not the username and there would be an error late during the creation process. This adds defaulting logic for the username as well to avoid that.
4. Remove CloudStack provider dependency on API structs. Given the cluster spec is passed to most methods as an arguments, this is easier to track as the single source of truth. This avoids bugs where updates are only performed in one place but not the other.

*Testing (if applicable):*
- Unit tests
- Run create/upgrade e2e test manually

### DIFF
--- a/controllers/resource/reconciler.go
+++ b/controllers/resource/reconciler.go
@@ -117,6 +117,8 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 		}
 		resources = append(resources, r...)
 	case anywherev1.CloudStackDatacenterKind:
+		// max len = len(workers) + CP + etcd
+		spec.CloudStackMachineConfigs = make(map[string]*anywherev1.CloudStackMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations)+2)
 		csdc := &anywherev1.CloudStackDatacenterConfig{}
 		cpCsmc := &anywherev1.CloudStackMachineConfig{}
 		etcdCsmc := &anywherev1.CloudStackMachineConfig{}
@@ -126,23 +128,31 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 		if err != nil {
 			return err
 		}
+		spec.CloudStackDatacenter = csdc
+
 		err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.ControlPlaneConfiguration.MachineGroupRef.Name}, cpCsmc)
 		if err != nil {
 			return err
 		}
+		spec.CloudStackMachineConfigs[cs.Spec.ControlPlaneConfiguration.MachineGroupRef.Name] = cpCsmc
+
 		for _, workerNodeGroupConfiguration := range cs.Spec.WorkerNodeGroupConfigurations {
 			err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: workerNodeGroupConfiguration.MachineGroupRef.Name}, workerCsmc)
 			if err != nil {
 				return err
 			}
 			workerCsmcs[workerNodeGroupConfiguration.MachineGroupRef.Name] = *workerCsmc
+			spec.CloudStackMachineConfigs[workerNodeGroupConfiguration.MachineGroupRef.Name] = workerCsmc
 		}
+
 		if cs.Spec.ExternalEtcdConfiguration != nil {
 			err = cor.FetchObject(ctx, types.NamespacedName{Namespace: objectKey.Namespace, Name: cs.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name}, etcdCsmc)
 			if err != nil {
 				return err
 			}
+			spec.CloudStackMachineConfigs[cs.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name] = etcdCsmc
 		}
+
 		r, err := cor.cloudStackTemplate.TemplateResources(ctx, cs, spec, *csdc, *cpCsmc, *etcdCsmc, workerCsmcs)
 		if err != nil {
 			return err
--- a/controllers/resource/template.go
+++ b/controllers/resource/template.go
@@ -179,7 +179,7 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster 
 	}
 	clusterSpec.CloudStackDatacenter = &csdc
 	// control plane and etcd updates are prohibited in controller so those specs should not change
-	templateBuilder := cloudstack.NewTemplateBuilder(&csdc.Spec, &cpCsmc.Spec, &etcdCsmc.Spec, workerNodeGroupMachineSpecs, r.now)
+	templateBuilder := cloudstack.NewTemplateBuilder(r.now)
 	clusterName := clusterSpec.Cluster.Name
 
 	oldCsdc, err := r.ExistingCloudStackDatacenterConfig(ctx, eksaCluster, clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0])
@@ -211,8 +211,6 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster 
 
 	cpOpt := func(values map[string]interface{}) {
 		values["controlPlaneTemplateName"] = controlPlaneTemplateName
-		values["cloudstackControlPlaneSshAuthorizedKey"] = sshAuthorizedKey(cpCsmc.Spec.Users)
-		values["cloudstackEtcdSshAuthorizedKey"] = sshAuthorizedKey(etcdCsmc.Spec.Users)
 		values["etcdTemplateName"] = etcdTemplateName
 	}
 
--- a/controllers/resource/testdata/cloudstackEtcdadmcluster.yaml
+++ b/controllers/resource/testdata/cloudstackEtcdadmcluster.yaml
@@ -28,7 +28,7 @@ preEtcdadmCommands:
 users:
   - name: capc
     sshAuthorizedKeys:
-      - 'ssh-rsa ssh_key_value'
+      - 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC8ZEibIrz1AUBKDvmDiWLs9f5DnOerC4qPITiDtSOuPAsxgZbRMavBfVTxodMdAkYRYlXxK6PqNo0ve0qcOV2yvpxH1OogasMMetck6BlM/dIoo3vEY4ZoG9DuVRIf9Iry5gJKbpMDYWpx1IGZrDMOFcIM20ii2qLQQk5hfq9OqdqhToEJFixdgJt/y/zt6Koy3kix+XsnrVdAHgWAq4CZuwt1G6JUAqrpob3H8vPmL7aS+35ktf0pHBm6nYoxRhslnWMUb/7vpzWiq+fUBIm2LYqvrnm7t3fRqFx7p2sZqAm2jDNivyYXwRXkoQPR96zvGeMtuQ5BVGPpsDfVudSW21+pEXHI0GINtTbua7Ogz7wtpVywSvHraRgdFOeY9mkXPzvm2IhoqNrteck2GErwqSqb19mPz6LnHueK0u7i6WuQWJn0CUoCtyMGIrowXSviK8qgHXKrmfTWATmCkbtosnLskNdYuOw8bKxq5S4WgdQVhPps2TiMSZndjX5NTr8='
     sudo: ALL=(ALL) NOPASSWD:ALL
 infrastructureTemplate:
   apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
--- a/pkg/api/v1alpha1/cloudstackmachineconfig.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig.go
@@ -9,6 +9,10 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// DefaultCloudStackUser is the default CloudStackMachingConfig username.
+const DefaultCloudStackUser = "capc"
+
+// CloudStackMachineConfigKind is the kind value for a CloudStackMachineConfig.
 const CloudStackMachineConfigKind = "CloudStackMachineConfig"
 
 // Taken from https://github.com/shapeblue/cloudstack/blob/08bb4ad9fea7e422c3d3ac6d52f4670b1e89eed7/api/src/main/java/com/cloud/vm/VmDetailConstants.java
--- a/pkg/providers/cloudstack/cloudstack.go
+++ b/pkg/providers/cloudstack/cloudstack.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 	"github.com/aws/eks-anywhere/pkg/providers/common"
@@ -56,7 +57,6 @@ var (
 
 type cloudstackProvider struct {
 	datacenterConfig      *v1alpha1.CloudStackDatacenterConfig
-	machineConfigs        map[string]*v1alpha1.CloudStackMachineConfig
 	clusterConfig         *v1alpha1.Cluster
 	providerKubectlClient ProviderKubectlClient
 	writer                filewriter.FileWriter
@@ -179,7 +179,7 @@ func (p *cloudstackProvider) ValidateNewSpec(ctx context.Context, cluster *types
 	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
 
 	for _, machineConfigRef := range clusterSpec.Cluster.MachineConfigRefs() {
-		machineConfig, ok := p.machineConfigs[machineConfigRef.Name]
+		machineConfig, ok := clusterSpec.CloudStackMachineConfigs[machineConfigRef.Name]
 		if !ok {
 			return fmt.Errorf("cannot find machine config %s in cloudstack provider machine configs", machineConfigRef.Name)
 		}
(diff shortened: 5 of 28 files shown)


controllers/resource/reconciler.go
```diff
@@ -117,6 +117,8 @@ func (cor *clusterReconciler) Reconcile(ctx context.Context, objectKey types.Nam
 		}
 		resources = append(resources, r...)
 	case anywherev1.CloudStackDatacenterKind:
+		// max len = len(workers) + CP + etcd
+		spec.CloudStackMachineConfigs = make(map[string]*anywherev1.CloudStackMachineConfig, len(cs.Spec.WorkerNodeGroupConfigurations)+2)
```
cxbrowne1207, line 121: The template builder makes use of the `CloudStackMachineConfigs`, so making sure to add them to the template builder here.

controllers/resource/testdata/cloudstackEtcdadmcluster.yaml
```diff
@@ -28,7 +28,7 @@ preEtcdadmCommands:
 users:
   - name: capc
     sshAuthorizedKeys:
-      - 'ssh-rsa ssh_key_value'
+      - 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC8ZEibIrz1AUBKDvmDiWLs9f5DnOerC4qPITiDtSOuPAsxgZbRMavBfVTxodMdAkYRYlXxK6PqNo0ve0qcOV2yvpxH1OogasMMetck6BlM/dIoo3vEY4ZoG9DuVRIf9Iry5gJKbpMDYWpx1IGZrDMOFcIM20ii2qLQQk5hfq9OqdqhToEJFixdgJt/y/zt6Koy3kix+XsnrVdAHgWAq4CZuwt1G6JUAqrpob3H8vPmL7aS+35ktf0pHBm6nYoxRhslnWMUb/7vpzWiq+fUBIm2LYqvrnm7t3fRqFx7p2sZqAm2jDNivyYXwRXkoQPR96zvGeMtuQ5BVGPpsDfVudSW21+pEXHI0GINtTbua7Ogz7wtpVywSvHraRgdFOeY9mkXPzvm2IhoqNrteck2GErwqSqb19mPz6LnHueK0u7i6WuQWJn0CUoCtyMGIrowXSviK8qgHXKrmfTWATmCkbtosnLskNdYuOw8bKxq5S4WgdQVhPps2TiMSZndjX5NTr8='
```
cxbrowne1207, line 31: This is required now because of the new SSH key logic in the template builder

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -1004,58 +1004,19 @@ func TestPreCAPIInstallOnBootstrap(t *testing.T) {
...
-	provider := givenProvider(t)
-	etcdMachineConfigName := clusterSpec.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name
-	tempKey := "ssh-rsa AAAA    B3NzaC1yc2EAAAADAQABAAACAQC1BK73XhIzjX+meUr7pIYh6RHbvI3tmHeQIXY5lv7aztN1UoX+bhPo3dwo2sfSQn5kuxgQdnxIZ/CTzy0p0GkEYVv3gwspCeurjmu0XmrdmaSGcGxCEWT/65NtvYrQtUE5ELxJ+N/aeZNlK2B7IWANnw/82913asXH4VksV1NYNduP0o1/G4XcwLLSyVFB078q/oEnmvdNIoS61j4/o36HVtENJgYr0idcBvwJdvcGxGnPaqOhx477t+kfJAa5n5dSA5wilIaoXH5i1Tf/HsTCM52L+iNCARvQzJYZhzbWI1MDQwzILtIBEQCJsl2XSqIupleY8CxqQ6jCXt2mhae+wPc3YmbO5rFvr2/EvC57kh3yDs1Nsuj8KOvD78KeeujbR8n8pScm3WDp62HFQ8lEKNdeRNj6kB8WnuaJvPnyZfvzOhwG65/9w13IBl7B1sWxbFnq2rMpm5uHVK7mAmjL0Tt8zoDhcE1YJEnp9xte3/pvmKPkST5Q/9ZtR9P5sI+02jY0fvPkPyC03j2gsPixG7rpOCwpOdbny4dcj0TDeeXJX8er+oVfJuLYz0pNWJcT2raDdFfcqvYA0B0IyNYlj5nWX4RuEcyT3qocLReWPnZojetvAG/H8XwOh7fEVGqHAKOVSnPXCSQJPl6s0H12jPJBDJMTydtYPEszl4/CeQ== testemail@test.com"
-	provider.machineConfigs[etcdMachineConfigName].Spec.Users[0].SshAuthorizedKeys[0] = tempKey
-	setupContext(t)
-
-	err := provider.SetupAndValidateCreateCluster(ctx, clusterSpec)
-	thenErrorExpected(t, "setting up SSH keys: ssh: no key found", err)
-}
-
```
cxbrowne1207, line 1045: Same here

pkg/providers/cloudstack/cloudstack_test.go
```diff
@@ -1748,60 +1709,6 @@ func TestSetupAndValidateUpgradeClusterEtcdSshNotExists(t *testing.T) {
...
-	setupContext(t)
-
-	mockCtrl := gomock.NewController(t)
-	kubectl := mocks.NewMockProviderKubectlClient(mockCtrl)
-	provider.providerKubectlClient = kubectl
-
-	cluster := &types.Cluster{}
-	err := provider.SetupAndValidateUpgradeCluster(ctx, cluster, clusterSpec, clusterSpec)
-	thenErrorExpected(t, "setting up SSH keys: ssh: no key found", err)
-}
```
cxbrowne1207, line 1767: The logic for stripping comments from SSH keys has been moved to the template builder so, this will never happen here. This is being scenario is now being tested in template.go file now

controllers/resource/template.go
```diff
@@ -211,8 +211,6 @@ func (r *CloudStackTemplate) TemplateResources(ctx context.Context, eksaCluster
 
 	cpOpt := func(values map[string]interface{}) {
 		values["controlPlaneTemplateName"] = controlPlaneTemplateName
-		values["cloudstackControlPlaneSshAuthorizedKey"] = sshAuthorizedKey(cpCsmc.Spec.Users)
-		values["cloudstackEtcdSshAuthorizedKey"] = sshAuthorizedKey(etcdCsmc.Spec.Users)
```
cxbrowne1207, line 215: Moving these to the template builder where there is now one logical way assigning the keys to the MachineTemplate

pkg/providers/cloudstack/validator.go
```diff
@@ -141,14 +142,17 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, cloudStac
...
 		if len(machineConfig.Spec.Users) <= 0 {
 			machineConfig.Spec.Users = []anywherev1.UserConfiguration{{}}
 		}
 		if len(machineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
 			machineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
 		}
-		if err = v.validateMachineConfig(ctx, cloudStackClusterSpec.datacenterConfig, machineConfig); err != nil {
+		if machineConfig.Spec.Users[0].Name == "" {
+			machineConfig.Spec.Users[0].Name = v1alpha1.DefaultCloudStackUser
+		}
```
cxbrowne1207, line 154: If there were no users specified, the logic here would default the SSH key for the default user but not the username and there would be an error late during the creation process. This adds defaulting logic for the username as well to avoid that.

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -232,25 +232,16 @@ type ProviderKubectlClient interface {
...
 
 // NewProvider initializes the CloudStack provider object.
-func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, machineConfigs map[string]*v1alpha1.CloudStackMachineConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, validator ProviderValidator, writer filewriter.FileWriter, now types.NowFunc, log logr.Logger) *cloudstackProvider { //nolint:revive
-	controlPlaneMachineSpec := getControlPlaneMachineSpec(clusterConfig.Spec, machineConfigs)
-	etcdMachineSpec := getEtcdMachineSpec(clusterConfig.Spec, machineConfigs)
-	workerNodeGroupMachineSpecs := make(map[string]v1alpha1.CloudStackMachineConfigSpec, len(machineConfigs))
+func NewProvider(datacenterConfig *v1alpha1.CloudStackDatacenterConfig, clusterConfig *v1alpha1.Cluster, providerKubectlClient ProviderKubectlClient, validator ProviderValidator, writer filewriter.FileWriter, now types.NowFunc, log logr.Logger) *cloudstackProvider { //nolint:revive
 	return &cloudstackProvider{
 		datacenterConfig:      datacenterConfig,
-		machineConfigs:        machineConfigs,
```
ddjjia, line 241: why taking off machine configs here?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -280,65 +271,21 @@ func (p *cloudstackProvider) MachineResourceType() string {
...
-			}
-		} else {
-			if useKeyGeneratedForControlplane { // use the same key as for controlplane
-				etcdUser.SshAuthorizedKeys[0] = controlPlaneUser.SshAuthorizedKeys[0]
-			} else if useKeyGeneratedForWorker {
-				etcdUser.SshAuthorizedKeys[0] = workerUser.SshAuthorizedKeys[0] // if cp key was provided by user, check if worker key was generated by cli and use that
-			} else {
-				p.log.Info("Provided etcd sshAuthorizedKey is not set or is empty, auto-generating new key pair...")
-				generatedKey, err := common.GenerateSSHAuthKey(p.writer)
+func (p *cloudstackProvider) generateSSHKeysIfNotSet(machineConfigs map[string]*v1alpha1.CloudStackMachineConfig) error {
```
ddjjia, line 274: Do we still need this if users are required to provide ssh keys?

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -633,7 +568,6 @@ func (p *cloudstackProvider) generateCAPISpecForCreate(ctx context.Context, clus
 	for _, workerNodeGroupConfiguration := range clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations {
 		workloadTemplateNames[workerNodeGroupConfiguration.Name] = common.WorkerMachineTemplateName(clusterSpec.Cluster.Name, workerNodeGroupConfiguration.Name, p.templateBuilder.now)
 		kubeadmconfigTemplateNames[workerNodeGroupConfiguration.Name] = common.KubeadmConfigTemplateName(clusterSpec.Cluster.Name, workerNodeGroupConfiguration.Name, p.templateBuilder.now)
-		p.templateBuilder.WorkerNodeGroupMachineSpecs[workerNodeGroupConfiguration.MachineGroupRef.Name] = p.machineConfigs[workerNodeGroupConfiguration.MachineGroupRef.Name].Spec
```
ddjjia, line 636: why removing this?

pkg/providers/cloudstack/validator.go
ddjjia: Could you check with @mitalipaygude on how webhook would validate the ssh key, only the ssh key itself or include user?

pkg/providers/cloudstack/cloudstack.go
cxbrowne1207: The same thing was done for the vSphere implementation a while back, of which cloudstack was essentially similar.

It's already in the` cluster.Spec` which is already passed and used in most methods, so it isn't needed. I had run into issues where updates were here but not in the cluster.Spec. Removing the `cloudstackProvider` dependency on API machineconfig structs avoids those bugs where updates are only performed in one place. Given the cluster spec is passed to most methods as an arguments, this is easier to track as the single source of truth.

pkg/providers/cloudstack/cloudstack.go
cxbrowne1207: To clarify, the ssh keys are required for the template builder. There is no validation blocking users from not entering SSH keys. My goal here was just to remove the stripping of comments from the ssh keys, as that is new done in the template builder and preserving the rest of current CLI behavior.

pkg/providers/cloudstack/cloudstack.go
cxbrowne1207: `WorkerNodeGroupMachineSpecs` is not a field on the template builder anymore. This information is now being pull directly from the cluster.Spec in the template builder

pkg/providers/cloudstack/validator.go
cxbrowne1207: Ok I'll follow up.  For reference, this is the error that occurs if the username is not specified, despite the CLI auto generating:

`Error: applying capi spec: executing apply: error: error validating "STDIN": error validating data: ValidationError(KubeadmControlPlane.spec.kubeadmConfigSpec.users[0]): missing required field "name" in io.x-k8s.cluster.controlplane.v1beta1.KubeadmControlPlane.spec.kubeadmConfigSpec.users; if you choose to ignore these errors, turn validation off with --validate=false`

pkg/providers/cloudstack/cloudstack.go
ddjjia: sorry let me rephrase my question: why taking of machine config but leave datacenter? Could you check if datacenter can be removed as well? not necessarily in this pr if you think that involves too much work, but can create an issue to track

pkg/providers/cloudstack/cloudstack.go
ddjjia: So are we going to remove `generateSSHKeysIfNotSet` once we add validation blocking users from not entering SSH keys? If so, could you sync with mitali and also leave a comment there?

cxbrowne1207: I don't think we're removing the function of generating ssh keys in the CLI anytime soon. We've only decided to block it in the webhook if an SSH key isn't specified. The team hasn't made a decision of how we'll achieve parity between the API and CLI as far as I know. I will double check with @mitalipaygude though

pkg/providers/cloudstack/template.go
```diff
@@ -236,6 +233,39 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec, controlPlaneMachineSpec, etcd
...
 }
 
+func buildControlPlaneTemplate(controlPlaneMachineSpec *v1alpha1.CloudStackMachineConfigSpec, values map[string]interface{}) (content []byte, err error) {
+	cpTemplateName, ok := values[cpTemplateNameKey]
+	if !ok {
+		return nil, fmt.Errorf("unable to determine control plane template name")
+	}
+	cpMachineTemplate := MachineTemplate(fmt.Sprintf("%s", cpTemplateName), controlPlaneMachineSpec)
+	cpMachineTemplateBytes, err := templater.ObjectsToYaml(cpMachineTemplate)
+	if err != nil {
```
chrisdoherty4, line 243: The variables here are all prefixed with `cp`, but in this context there's only 1 template being rendered, the control plane, so we can simplify everything to suit the context of the function.

pkg/api/v1alpha1/cloudstackmachineconfig.go
```diff
@@ -9,7 +9,10 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-const CloudStackMachineConfigKind = "CloudStackMachineConfig"
+const (
+	DefaultCloudStackUser       = "capc"                    // DefaultCloudStackUser is the default CloudStackMachingConfig username
```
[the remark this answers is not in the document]
cxbrowne1207: I moved it here because the previous package at `internal/package/api` seems to be primarily geared towards constructs and methods used in the testing modules. While here contains related code for the `CloudStackMachineConfig` in the main modules.

pkg/providers/cloudstack/template.go
```diff
@@ -19,35 +18,29 @@ import (
...
 }
 
-// TemplateBuilder is responsible for building the CAPI yaml templates.
-type TemplateBuilder struct {
-	controlPlaneMachineSpec     *v1alpha1.CloudStackMachineConfigSpec
-	WorkerNodeGroupMachineSpecs map[string]v1alpha1.CloudStackMachineConfigSpec
-	etcdMachineSpec             *v1alpha1.CloudStackMachineConfigSpec
-	now                         types.NowFunc
+// NewTemplateBuilder creates a new TemplateBuilder.
+func NewTemplateBuilder(now types.NowFunc) *TemplateBuilder {
```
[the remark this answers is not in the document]
cxbrowne1207: This can eventually be removed with the legacy controller, so I don't think it's worth refactoring but I will add a comment.

pkg/providers/cloudstack/spec.go
```diff
@@ -7,34 +7,34 @@ import (
...
+
+func etcdMachineConfig(s *cluster.Spec) *anywherev1.CloudStackMachineConfig {
 	if s.Cluster.Spec.ExternalEtcdConfiguration == nil || s.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef == nil {
 		return nil
 	}
-	return s.machineConfigsLookup[s.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name]
+	return s.CloudStackMachineConfigs[s.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name]
+}
+
+func controlPlaneMachineConfig(s *cluster.Spec) *anywherev1.CloudStackMachineConfig {
```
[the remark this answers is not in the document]
cxbrowne1207: Alright. I’m thinking, we may just want to get rid of the `Spec` construct all together. It previously contained more than just the `clusterSpec`, which is why it was useful, but now it’s just a wrapper really

pkg/providers/cloudstack/validator.go
cxbrowne1207: Confirmed with @mitalipaygude. The web hook would validate the ssh and default the user. The same is done in all the other providers.

pkg/providers/cloudstack/cloudstack.go
cxbrowne1207: It think we should do this in another PR, as a fast follow. It won't be too much work, but testing may take up some more time. Also, I do not want to extend the scope of this PR further.

PR #5677, opened 2023-04-20
use --password-stdin to prevent process table leaking

ewollesen:
Modifies the `helm registry login` command to use the `--password-stdin` flag which prevents leaking the password on the process table.

This might also be the cause of a bug: https://github.com/aws/eks-anywhere-packages/issues/923

Though as far as I can tell, the helm registry login command still accepts the `--password flag`, though it prints a warning.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


TerryHowe (approved):
/lgtm

Issue #5524, opened 2023-04-04
Update formula to calculate free datastore disk space in case of upgrade


panktishah26:
**What happened**:
For Vsphere provider, if we want to upgrade a cluster by scaling up worker node size for ex, by 2 count and 75 diskGib from 10 worker nodes and 100 diskGib , it requires free space in workload datastore by ((12 * 75) for the upgrade to succeed. Same is the case for CP and ETCD.

`2023-04-01T11:01:58.569Z        V0      ❌ Validation failed     {"validation": "vsphere provider validation", "error": "not enough space in datastore <datastore> for given diskGiB and count for respective machine groups", "remediation": ""}`

**What you expected to happen**:
We should calculate free available space in the respective datastore [here](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/validator.go#L263-L315)  by ((10 * 100)-(12 * 75)) which would be 100 in this case, rather than 900. 

**How to reproduce it (as minimally and precisely as possible)**: 

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

PR #5683, opened 2023-04-21
Update datastore usage validation to account for space that will free up during upgrade

taneyland:
*Issue #, if available:*
#5524 

*Description of changes:*
Update datastore usage validation to account for space that will free up during upgrade. 

When we validate datastore usage during upgrades, we will now check if the machine config already exists on the cluster. If it exists, we will get the space that it is already using and add that to the availableSpace to prevent inaccurate datastore usage errors when modifying the disk size/replica count in machine configs.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -303,6 +303,9 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
 	if err := p.validator.ValidateClusterMachineConfigs(ctx, vSphereClusterSpec); err != nil {
 		return err
 	}
+	if err := p.validateDatastoreUsageForCreate(ctx, vSphereClusterSpec); err != nil {
+		return fmt.Errorf("validating vsphere machine configs datastore usage: %v", err)
+	}
 
 	if err := p.generateSSHKeysIfNotSet(clusterSpec.VSphereMachineConfigs); err != nil {
 		return fmt.Errorf("failed setup and validations: %v", err)
@@ -403,6 +406,10 @@ func (p *vsphereProvider) SetupAndValidateUpgradeCluster(ctx context.Context, cl
 		return err
 	}
 
+	if err := p.validateDatastoreUsageForUpgrade(ctx, vSphereClusterSpec, cluster); err != nil {
+		return fmt.Errorf("validating vsphere machine configs datastore usage: %v", err)
+	}
+
 	err := p.validateMachineConfigsNameUniqueness(ctx, cluster, clusterSpec)
 	if err != nil {
 		return fmt.Errorf("failed validate machineconfig uniqueness: %v", err)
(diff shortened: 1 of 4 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -443,6 +450,80 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
 	return nil
 }
 
+func (p *vsphereProvider) getCurrentMachineConfigDatastoreUsage(ctx context.Context, machineConfig *v1alpha1.VSphereMachineConfig, vsphereClusterSpec *Spec, cluster *types.Cluster) (float64, error) {
```
chrisdoherty4, line 453: Its helpful for maintainers if we indicate the unit being returned via the signature. Some options: create a return parameter `unitGiB` or document it in a method comment.

pkg/providers/vsphere/vsphere.go
```diff
@@ -443,6 +450,80 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
...
+func (p *vsphereProvider) getMachineConfigDatastoreRequirements(ctx context.Context, machineConfig *v1alpha1.VSphereMachineConfig, count int) (available float64, need int, err error) {
+	availableSpace, err := p.providerGovcClient.GetWorkloadAvailableSpace(ctx, machineConfig.Spec.Datastore) // TODO: remove dependency on machineConfig
+	if err != nil {
+		return 0, 0, fmt.Errorf("getting datastore details: %v", err)
+	}
+	needGiB := machineConfig.Spec.DiskGiB * count
+	return availableSpace, needGiB, nil
+}
+
+func (p *vsphereProvider) validateDatastoreUsage(ctx context.Context, vsphereClusterSpec *Spec, cluster *types.Cluster) error {
```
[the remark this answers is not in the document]
taneyland: We need the `kubectl` client to search for the current machine config objects on the cluster during upgrades. In the vsphere validator, we only make external calls using GoVC. Any validations that require the kubectl client are defined at the provider level & called from there.

pkg/providers/vsphere/vsphere.go
```diff
@@ -443,6 +450,80 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
...
+func (p *vsphereProvider) calculateDatastoreUsage(ctx context.Context, machineConfig *v1alpha1.VSphereMachineConfig, vsphereClusterSpec *Spec, cluster *types.Cluster, usage map[string]*datastoreUsage) error {
+	availableSpace, needGiB, err := p.getMachineConfigDatastoreRequirements(ctx, machineConfig, vsphereClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Count)
+	if err != nil {
+		return err
+	}
+	currentUsage, err := p.getCurrentMachineConfigDatastoreUsage(ctx, machineConfig, vsphereClusterSpec, cluster)
+	if err != nil {
+		return err
+	}
+	availableSpace += currentUsage
```
jiayiwang7, line 476: why `availableSpace += currentUsage`?

pkg/providers/vsphere/vsphere.go
```diff
@@ -443,6 +450,80 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
...
+	if _, ok := usage[machineConfig.Spec.Datastore]; ok {
+		usage[machineConfig.Spec.Datastore].needGiBSpace += needGiB
+		usage[machineConfig.Spec.Datastore].availableSpace += availableSpace
```
jiayiwang7, line 479: what if multiple machine configs are sharing the same datastore? isn't this adding the same available space multiple times?

pkg/providers/vsphere/vsphere.go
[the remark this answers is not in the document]
jiayiwang7: then should we add the kubectl client to the validator?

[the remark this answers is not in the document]
taneyland: I figured we were explicitly not defining it there since we separate validating machine config name uniqueness due to the dependency of kubectl client. https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/vsphere.go#L413

If there are no foreseen issues with adding the kubectl client here, then i can make those changes

pkg/providers/vsphere/vsphere.go
taneyland: oh yes, good catch. This should be `usage[machineConfig.Spec.Datastore].availableSpace += currentUsage`. This will help to account for the current machine configs that will be replaced, freeing up space for any updates/new machine configs

pkg/providers/vsphere/vsphere.go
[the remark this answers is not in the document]
jiayiwang7: ah you are right, i think the reason not injecting kubectl to validator is bc controller code uses the validation methods and can't use the same `kubectl` client without refactoring

pkg/providers/vsphere/vsphere.go
```diff
@@ -443,6 +450,80 @@ func (p *vsphereProvider) validateMachineConfigsNameUniqueness(ctx context.Conte
...
+		return 0, err
+	}
+	if len(em) > 0 {
+		return float64(em[0].Spec.DiskGiB), nil
+	}
+	return 0, nil
+}
+
+func (p *vsphereProvider) calculateDatastoreUsage(ctx context.Context, machineConfig *v1alpha1.VSphereMachineConfig, vsphereClusterSpec *Spec, cluster *types.Cluster, usage map[string]*datastoreUsage) error {
+	availableSpace, needGiB, err := p.getMachineConfigDatastoreRequirements(ctx, machineConfig, vsphereClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Count)
```
abhinavmpandey08, line 468: This function is being called for cp, workers and etcd separately right? We shouldn't hardcode the control plane node count (`vsphereClusterSpec.Cluster.Spec.ControlPlaneConfiguration.Count`) here

pkg/providers/vsphere/vsphere.go
taneyland: This helps to account for space that's already used by current machines. We calculate the NeedGiB based on the machine `diskGiB * count`, so if the machine config already exists (thus already taking up datastore space), we are double dipping during the calculation without this line.

Issue #5310, opened 2023-03-20
Document user cert renewal process


jiayiwang7:

vignesh-goutham:
Related issue for Etcd - https://github.com/aws/eks-anywhere/issues/5400

We have to do for both kubeadm and etcdadm

PR #5687, opened 2023-04-24
Add auto-certs-renewal design doc

d8660091:
*Issue #, if available:* https://github.com/aws/eks-anywhere/issues/5310

*Description of changes:* Export quip doc to design doc

*Testing (if applicable):*


designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Context
+
+As a user who deployed K8s cluster with EKS-A, I want an option to automatically renew the certificates in etcd and control plane nodes before expiry in 1 year.
+
+### External ETCD node certificates
+
+* etcd node’s certificates are stored in /etc/etcd/pki (/var/lib/etcd/pki in bottlerocket)
+* each certificate comes in pair: key and crt
+* all etcd nodes contain the same key pairs no matter if it’s the first node (not in term of contents)
```
jiayiwang7, line 11:
```suggestion
* all etcd nodes contain the same key pair names no matter if it’s the first node (not in term of contents)
```
?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,121 @@
...
+
+However, not all certificates in cp node would be renewed even both etcd and cp nodes have changed, since apiserver-etcd-client key pair is created by etcdadm controller (refer the sequence diagram) and it always stays the same. We need to fix this bug by changing [etcdadm controller’s code](https://github.com/aws/etcdadm-controller/blob/main/controllers/certs.go#L81), it should update apiserver-etcd-client secret constantly in a short interval, such as every 24 hours. If we choose just update the secret during etcd nodes rolling out, then we have to invent mechanism to make sure the etcd nodes rolling out happens before cp nodes rolling out when it comes to renew certificates.
+
+### Cluster Yaml Spec Example
+
+```
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+spec:
+  certificatesConfiguration:
```
jiayiwang7, line 82: what other certConfig will there be?

d8660091: Such as specifying ETCD/CP CA certs like Chris mentioned, or something like customizing cert's expiring time, algorithm, subject etc. But most likely, we would not need those fields in the near future, so just "thinking big" and make it extensible here.

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,121 @@
...
+    1. apply machine annotation `machine.cluster.x-k8s.io/certificates-expiry` during etcd node creation.
+    2. check if roll out is needed. Refer https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/internal/controllers/controller.go#L380.
+    3. if roll out is needed for any etcd node, upgrade all etcd nodes.
+
+However, not all certificates in cp node would be renewed even both etcd and cp nodes have changed, since apiserver-etcd-client key pair is created by etcdadm controller (refer the sequence diagram) and it always stays the same. We need to fix this bug by changing [etcdadm controller’s code](https://github.com/aws/etcdadm-controller/blob/main/controllers/certs.go#L81), it should update apiserver-etcd-client secret constantly in a short interval, such as every 24 hours. If we choose just update the secret during etcd nodes rolling out, then we have to invent mechanism to make sure the etcd nodes rolling out happens before cp nodes rolling out when it comes to renew certificates.
```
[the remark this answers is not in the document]
d8660091: This is still necessary. It is a bug existing today: a user would have apiserer-etcd-client expired even if he did the rollout for etcd and cp everyday.

The proposed solution by rollout etcd only would only make sure that cp doesn't have an extra rollout

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,121 @@
...
+      // whether to rollout etcd and cp nodes before certs expiring
+      enabled: true/false (false by default, auto certs renew require opt-in) 
+      // allow adding finer control in the future
+```
+
+### Tech Details
+
+1. `certificatesConfiguration.rolloutBeforeExpiry.enabled` should be true when user run “eksctl anywhere generate clusterconfig”
+2. When `rolloutBeforeExpiry` is enabled, the cp/etcd nodes are rollout when certs are expiring in **7** days
+3. etcd certs renewal has a side-effect of renewing cp certs (etcd-rollout → etcd api change → cp rollout → cp certs renewal). We should only renew etcd certs to avoid unnecessary cp rollout.
```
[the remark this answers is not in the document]
d8660091: Added

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+
+The above certificate files are described in the official doc: https://kubernetes.io/docs/setup/best-practices/certificates/#all-certificates
+
+### Worker node certificates
+
+1. just ca.crt in /etc/kubernetes/pki, which should be generated after `kubeadm join`
+
+## Goal
+
+1. Auto renew all the above mentioned certificates except etcd CA and kube CA, which last for 10 years.
```
chrisdoherty4, line 45: Do we have an open issue for handling this also? Or, perhaps there's an easy win to address this at the same time?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+    1. apply machine annotation `machine.cluster.x-k8s.io/certificates-expiry` during etcd node creation.
+    2. check if roll out is needed. Refer https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/internal/controllers/controller.go#L380.
+    3. if roll out is needed for any etcd node, upgrade all etcd nodes.
+
+However, not all certificates in cp node would be renewed even both etcd and cp nodes have changed, since apiserver-etcd-client key pair is created by etcdadm controller (refer the sequence diagram) and it always stays the same. We need to fix this bug by changing [etcdadm controller’s code](https://github.com/aws/etcdadm-controller/blob/main/controllers/certs.go#L81), it should update apiserver-etcd-client secret constantly in a short interval, such as every 24 hours. If we choose just update the secret during etcd nodes rolling out, then we have to invent mechanism to make sure the etcd nodes rolling out happens before cp nodes rolling out when it comes to renew certificates.
```
chrisdoherty4, line 74: > However, not all certificates in cp node would be renewed even both etcd and cp nodes have changed, since apiserver-etcd-client key pair is created by etcdadm controller (refer the sequence diagram) and it always stays the same. 

Could you clarify this. Are we making a statement about the existing behavior or something that's the result of implementing part of a solution?

In conjunction with the clarification, I think we need to clearly state the problem before mentioning any solutions. Does the final sentence elaborate further on the problem for example? Perhaps combine that with the problem description from the first sentence.

chrisdoherty4, line 74: > We need to fix this bug by changing [etcdadm controller’s code](https://github.com/aws/etcdadm-controller/blob/main/controllers/certs.go#L81), it should update apiserver-etcd-client secret constantly in a short interval, such as every 24 hours

What is the expiry duration on the client certificate currently?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Cluster Yaml Spec Example
+
+```
+apiVersion: anywhere.eks.amazonaws.com/v1alpha1
+kind: Cluster
+spec:
+  certificatesConfiguration:
```
chrisdoherty4, line 82: Would we be open to just `certs` or `certificates`? This object is all about configuration and I don't think stating that explicitly in the name adds much.

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+    rolloutBeforeExpiry: 
+      // whether to rollout etcd and cp nodes before certs expiring
+      enabled: true/false (false by default, auto certs renew require opt-in) 
+      // allow adding finer control in the future
```
chrisdoherty4, line 86: What's the rationale for `rolloutBeforeExpiry` being a map vs a bool?

For example
```yaml
spec:
  certificates:
    rolloutBeforeExpiry: true
```

Also curious if, instead of a bool, we should use a time with [`time.Parse()`](https://pkg.go.dev/time#ParseDuration) semantics.

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+### Tech Details
+
+1. `certificatesConfiguration.rolloutBeforeExpiry.enabled` should be true when user run “eksctl anywhere generate clusterconfig”
```
chrisdoherty4, line 91: If the default is to enable, do we want the inverse? I.e. `disabled`?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+2. When `rolloutBeforeExpiry` is enabled, the cp/etcd nodes are rollout when certs are expiring in **7** days
```
chrisdoherty4, line 92:
```suggestion
2. When `rolloutBeforeExpiry` is enabled the nodes are rolled 7 days before expiry.
```

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,122 @@
...
+3. etcd certs renewal has a side-effect of renewing cp certs (etcd-rollout → etcd api change → cp rollout → cp certs renewal). We should only renew etcd certs to avoid unnecessary cp rollout.
+4. ETCD cluster rollout should undergo the same process as upgrading cluster, making sure that etcd cluster is accessible from cp, aka, no downtime for the k8s cluster.
+
+## Alternative Solutions
+
+### Cronjob
+
+The script run by the job would need to be tailored/tested for each provider/OS combinations. It’s error prone and hard to develop/maintain. Manually renew certs in bottlerocket is very tricky. Instead of running the binary locally, you have to pull the etcdadm/kubeadm image locally, mount folders, configuring container runtime, calling bottlerocket api etc to make it work. Besides, when there is bug or dependencies change, you would perhap need to test the fix against multiple environments.
```
chrisdoherty4, line 100: These points are really interesting, thanks for noting them. To be able to think critically about the alternatives I think we need a little elaboration on the 'why' and a little more explanation on the what. Where does this CronJob run? Why is it error prone? Why is renewing certs in BottleRocket tricky?

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
g-gaston: OK, help me walk through this pls, I might be missing something.
The etcd rollout to renew certs will always happen before the CP rollout. In fact, it's etcd the one that will cause the CO to rollout.

Then, why can't we just renew the apiserver-etcd-client secret just before the etcd rollout? That way it will be fresh when the CP nodes get rolled out.

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
g-gaston: > ETCD cluster rollout should undergo the same process as upgrading cluster

Could we go over that process in the doc?

It is the same process than upgrades, but the same exact solution won't work. Upgrades are initiated by an external entity (the user). This decision to renew certs comes from withing the controllers (eg. there are no external API changes).

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
d8660091: If we just renew apiserver-etcd-client cert before etcd rollout, the apiserver-etcd-client cert of control plane will be renewed. However, the bug still exists: as a user, when I rollout cp nodes, I want all certs in the cp nodes get renewed.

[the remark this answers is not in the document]
g-gaston: why is it a bug? There is no contract saying that the client cert always has to be fresh.

> However, the bug still exists: as a user, when I rollout cp nodes, I want all certs in the cp nodes get renewed.

Where is this requirement coming from?

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
d8660091: Right, it's not exactly the same process. The intended message to convey is in fact:
```
ETCD cluster rollout should be triggered similarly as it's triggered by upgrading etcd cluster
```

designs/auto-certs-renewal.md
d8660091: No such issue was created as far as I know.

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
d8660091: I would say it's an implicit requirement based on common sense.

If a user upgrades his cluster everyday (not enabling auto certs renew), and he still get this cert expired in 1 year, he would have a great chance calling this behaviour "bug".

designs/auto-certs-renewal.md
d8660091: 1 year

designs/auto-certs-renewal.md
d8660091: I think we should bias to use the same vocabulary as kubernetes' official documentation where certificates is prefered  :
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/

designs/auto-certs-renewal.md
d8660091: It's a decision that we made with PM, the rationale is making it easier for the user to opt-in and flexible if we get feedbacks from users that they want to specify when to renew the certs

designs/auto-certs-renewal.md
d8660091: The default is still "false". We only make it true when the user generates clusterconfig.

designs/auto-certs-renewal.md
d8660091: added

designs/auto-certs-renewal.md
chrisdoherty4: Are we able to handle it at the same time. I know we want to address the first problem, but both problems seem similar.

designs/auto-certs-renewal.md
chrisdoherty4: `certificates` is fine. I'm just thinking we should drop 'configuration'.

designs/auto-certs-renewal.md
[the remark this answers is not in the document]
chrisdoherty4: I think we need to talk through the problem and solution some more. Are we able to schedule a call?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,126 @@
+# Auto Certificates Renewal
+
+## Context
+
+As a user who deployed K8s cluster with EKS-A, I want an option to automatically renew the certificates in etcd and control plane nodes before expiry in 1 year.
```
jacobweinstock, line 5: Hey @d8660091, the problem this is seeking to solve is a bit unclear to me. I'd like to see these questions answered.

- What problems are you solving?
- How do they impact users?

designs/auto-certs-renewal.md
jacobweinstock: is 10 years configurable? what happens if there is a need to roll the certs before then? what if they dont want 10 years.

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Solution
+
+Etcd can be deployed in [two topologies.](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/ha-topology/) Certs renew mechanism is slightly different in those two scenarios.
+
+### Stacked etcd Nodes
+
+In Stacked etcd nodes topology, etcd is running inside the cp node. 
+
+We can renew all certificates by just redeploying cp nodes
```
jacobweinstock, line 61: is this what we're telling customers to do? I'd like to see more details on the customer experience.

designs/auto-certs-renewal.md
chrisdoherty4: How does this impact existing customers - particularly those that may not realize they need to enable this?

designs/auto-certs-renewal.md
chrisdoherty4: Thoughts @d8660091?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,126 @@
...
+
+1. Expose [rolloutBefore.certificatesExpiryDays](https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp.html) field from KubeadmControlPlane in EKSA cluster spec. Add pass the field to KubeadmControlPlane
+
+### External etcd Nodes
+
+In external etcd nodes topology, etcd is running in a separate machine besides cp machine. In addition to roll out cp nodes, we also need to roll out etcd nodes, with the following changes on top of above:
+
+1. Add rolloutBefore.certificatesExpiryDays field to [EtcdadmCluster](https://github.com/aws/etcdadm-controller/blob/main/api/v1beta1/etcdadmcluster_types.go#L38), exposing and passing it from EKSA. The business logic of this field and reconciler should be:
+    1. apply machine annotation `machine.cluster.x-k8s.io/certificates-expiry` during etcd node creation.
```
chrisdoherty4, line 70: What happens if I remove the annotation?

designs/auto-certs-renewal.md
```diff
@@ -0,0 +1,132 @@
...
+
+Etcd can be deployed in [two topologies.](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/ha-topology/) Certs renew mechanism is slightly different in those two scenarios.
+
+### Stacked etcd Nodes
+
+In Stacked etcd nodes topology, etcd is running inside the cp node. 
+
+We can renew all certificates by just redeploying cp nodes
+
+1. Expose [rolloutBefore.certificatesExpiryDays](https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp.html) field from KubeadmControlPlane in EKSA cluster spec. Add pass the field to KubeadmControlPlane
```
g-gaston, line 70: nit: we are not exposing `certificatesExpiryDays` directly in out spec. So maybe "control rolloutBefore.certificatesExpiryDays from EKS-A, setting a fixed value (7 days) when auto-renew certificates is enabled". Or something like that.

g-gaston (commented):
lgtm

PR #5689, opened 2023-04-24
Add Timeout to Delete Full Lifecycle API E2E

tatlat:
*Issue #, if available:*

*Description of changes:*
Add a 20 minute timeout to calls that delete clusters with kubectl for full lifecycle api e2e tests

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -348,8 +348,12 @@ func (k *Kubectl) ApplyKubeSpecFromBytesForce(ctx context.Context, cluster *type
 }
 
 // DeleteManifest uses client-side logic to delete objects defined in a yaml manifest.
-func (k *Kubectl) DeleteManifest(ctx context.Context, kubeconfigPath, manifestPath string, timeout string) error {
-	if _, err := k.Execute(ctx, "delete", "-f", manifestPath, "--kubeconfig", kubeconfigPath, "--timeout", timeout); err != nil {
+func (k *Kubectl) DeleteManifest(ctx context.Context, kubeconfigPath, manifestPath string, opts ...KubectlOpt) error {
+	params := []string{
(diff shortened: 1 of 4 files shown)


pkg/executables/kubectl.go
```diff
@@ -349,7 +349,7 @@ func (k *Kubectl) ApplyKubeSpecFromBytesForce(ctx context.Context, cluster *type
 
 // DeleteManifest uses client-side logic to delete objects defined in a yaml manifest.
 func (k *Kubectl) DeleteManifest(ctx context.Context, kubeconfigPath, manifestPath string) error {
-	if _, err := k.Execute(ctx, "delete", "-f", manifestPath, "--kubeconfig", kubeconfigPath); err != nil {
+	if _, err := k.Execute(ctx, "delete", "-f", manifestPath, "--kubeconfig", kubeconfigPath, "--timeout", "20m"); err != nil {
```
jonathanmeier5, line 352: Instead of hard-coding this, could we pass in opts like for the [wait command](https://github.com/aws/eks-anywhere/blob/5fb16d2fd0bc960e7c4609c5b44cf0991194e863/pkg/executables/kubectl.go#L482-L482)?

I know we're only using the function in the test framework, but I'd still rather avoid hardcoding this timeout in our application code.

Then in [DeleteClusterWithKubectl](https://github.com/aws/eks-anywhere/blob/5fb16d2fd0bc960e7c4609c5b44cf0991194e863/test/framework/workload.go#L52-L52) we can set a constant like `kubectlDeleteTimeout = 20m` and pass it into our call to `DeleteManifest` something like:
```
w.KubectlClient.DeleteManifest(ctx, w.ManagementClusterKubeconfigFile(), w.ClusterConfigLocation, "--timeout", kubectlDeleteTimeout)
```

PR #5719, opened 2023-04-26
Update Nutanix E2E to determine UUID from name

abhinavmpandey08:
*Description of changes:*
With this change, we won't have to go update the UUIDs manually in the CI every time we build a new image or change cluster or subnet.

*Testing (if applicable):*
Ran E2E locally and validated the results


internal/pkg/nutanix/prismclient.go
```diff
@@ -0,0 +1,112 @@
...
+
+// PrismClient is a interface that provides useful functions for performing Prism operations.
+type PrismClient interface {
+	GetImageUUIDFromName(ctx context.Context, imageName string) (*string, error)
+	GetClusterUUIDFromName(ctx context.Context, clusterName string) (*string, error)
+	GetSubnetUUIDFromName(ctx context.Context, subnetName string) (*string, error)
+}
+
+type client struct {
+	v3.Client
```
pokearu, line 23: Do we want struct `client` to have an `is a` relationship with `v3.Client`?  This exposes all the behavior of the `v3.Client` from struct client. 

I think this would work for our use case
```code
type client struct {
	v3 v3.Client
}
```

abhinavmpandey08: The client struct isn't exposed outside so it shouldn't matter. It just implements the `PrismClient` interface
So outside this package, you can only do 
```go
// initialize the client
prismClient := nutanix.NewPrismClient(endpoint, port, true)

// this is valid since you are calling the interface functions
prismClient.GetImageUUIDFromName()
prismClient.GetClusterUUIDFromName()
prismClient.GetSubnetUUIDFromName()

// this is not valid since the PrismClient interface doesn't have this function
// even though v3.Client has the ListSubnet() function
prismClient.V3.ListSubnet()
```

thunderboltsid (approved):
lgtm

internal/pkg/nutanix/prismclient.go
pokearu: Discussed offline. We can go with this approach.

PR #5736, opened 2023-04-28
Set default and validate control plane host port when generating CloudStack CAPI objects

cxbrowne1207:
*Issue #, if available:*
[Set Default And Validate Control Plane Host Port when generate CAPI object#1280](https://github.com/aws/eks-anywhere-internal/issues/1280)

*Description of changes:*
When creating a CloudStack cluster, the endpoint provided by the user should have a port specified. This is already done for the CLI, but not for clusters created through the API yet. However, it modifies the user input in the cluster.Spec to achieve this: This PR:
- Validates and sets a default control plane host port when generating the CloudStack CAPI objects. This will allow clusters created through the API to utilize a default port for the control plane endpoint if not specified, bringing it up to parity with the CLI. 
- Removes the side effect of editing the user input in the `cluster.Spec`. The cluster.Spec represents data input by the user to create the cluster. It follow good practice to not permanently modify user input, and to instead transform it into other  non user input constructs.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/cloudstack/template_test.go
+++ b/pkg/providers/cloudstack/template_test.go
@@ -8,6 +8,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/aws/eks-anywhere/internal/test"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/clusterapi"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack"
 )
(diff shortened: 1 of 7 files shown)


pkg/providers/cloudstack/template_test.go
```diff
@@ -124,3 +139,14 @@ func TestTemplateBuilderGenerateCAPISpecWorkersInvalidSSHKey(t *testing.T) {
...
+func TestTemplateBuilderGenerateCAPISpecWorkersInvalidEndpoint(t *testing.T) {
+	g := NewWithT(t)
+	templateBuilder := cloudstack.NewTemplateBuilder(time.Now)
+	clusterSpec := test.NewFullClusterSpec(t, path.Join(testDataDir, testClusterConfigMainFilename))
+	clusterSpec.Cluster.Spec.ProxyConfiguration = &v1alpha1.ProxyConfiguration{}
+	clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host = "1.1.1.1::"
+	machineTemplateNames, kubeadmConfigTemplateNames := clusterapi.InitialTemplateNamesForWorkers(clusterSpec)
+	_, err := templateBuilder.GenerateCAPISpecWorkers(clusterSpec, machineTemplateNames, kubeadmConfigTemplateNames)
+	g.Expect(err).To(MatchError(ContainSubstring("building template map for MD host 1.1.1.1:: is invalid: address 1.1.1.1::: too many colons in address")))
+}
```
ddjjia, line 152: Can we have two more success test cases to check the default port is not provided but specified in in the template or capi object

cxbrowne1207: Those test cases are currently being covered by the success tests in `cloudstack_test.go`. Do you think it's still worth it to add coverage for them here?

Issue #5055, opened 2023-02-23
Curated packages import does not support airgapped import


raykrueger:
**What would you like to be added**:
As a cluster admin I want to be able to pull curated packages from the internet, switch to my private network, and push to a private registry.

**Why is this needed**:
The import images process allows for a pull from the internet, disconnect, connect to airgapped environment and push to private registry process. The curated packages import does this in one-shot pull/push process. Which means we cannot support an airgapped install.

TerryHowe:
I was hoping to upstream that to oras, but if that fails, we could use the code in my POC link in that issue.

jiayiwang7:
related issue: https://github.com/aws/eks-anywhere/issues/5396

TerryHowe:
The code in the draft PRs I mentioned are what we should use if we are going to implement this. It will take a while to get this through the oras community and I'm kind of busy now to push that.

TerryHowe:
> The code in the draft PRs I mentioned are what we should use if we are going to implement this. It will take a while to get this through the oras community and I'm kind of busy now to push that.

What I mean by that is if we want a `download packages` right now, we should just take the code from the draft PR and use it. If we can wait, it should get pushed through eventually.

PR #5745, opened 2023-04-30
Add download packages command

TerryHowe:
This implementation of `download pacakges` uses a new go scripting interface to `oras`. With this interface, you can use ORAS in Go just like you would in shell.


TerryHowe:
Partial https://github.com/aws/eks-anywhere/issues/5055

cmd/eksctl-anywhere/cmd/downloadpackages.go
```diff
@@ -0,0 +1,122 @@
...
+type DownloadPackagesCommand struct {
+	destination   string
+	bundleFile    string
+	awsRegion     string
+	registryCache *registry.Cache
+}
+
+func runDownloadPackages(_ *cobra.Command, args []string) error {
+	ctx := context.Background()
+	downloadPackagesCommand.destination = args[0]
```
g-gaston, line 57: I would lean towards having the destination as flag instead of a positional arg
I think that fits the syntax of other existing commands better

cmd/eksctl-anywhere/cmd/downloadpackages.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+
+	c.registryCache = registry.NewCache()
+	bundleReader := curatedpackages.NewPackageReader(c.registryCache, credentialStore, c.awsRegion)
+
+	imageList := bundleReader.ReadChartsFromBundles(ctx, eksaBundle)
+
+	log.Printf("Downloading curated packages helm charts from public ECR to %s", c.destination)
```
g-gaston, line 87: could we use the shared logger here?
you can get it from the factory using `WithLogger`

cmd/eksctl-anywhere/cmd/downloadpackages.go
```diff
@@ -0,0 +1,122 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func downloadSingleImage(source, destination string) error {
+	args := []string{"copy", "--to-oci-layout", source, destination}
+	cmd := root.New()
```
g-gaston, line 115: would it possible to use `oras.Copy` directly instead?
Using the library helps detecting breaking api changes. Plus it makes it easier to deal with the extra behavior that a cli code has, like logs.

cmd/eksctl-anywhere/cmd/downloadpackages.go
```diff
@@ -0,0 +1,122 @@
...
+		destination = path.Join(destination, project, image.Repository)
+		err := downloadSingleImage(source, destination)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func downloadSingleImage(source, destination string) error {
```
g-gaston, line 113: if we could make this reusable by moving it somewhere in pkg that would be great

cmd/eksctl-anywhere/cmd/downloadpackages.go
TerryHowe: Well, it isn't that easy. You have to set up the structures it uses. That is what I did for https://github.com/aws/eks-anywhere/pull/4420 and that was 25 files modified on an XXL PR and this is 6 lines to do the same.

chrisdoherty4:
@TerryHowe Do you intend on fixing this up?

TerryHowe:
I think this could be tested with just functional e2e tests because there isn't much that can be tested here.

abhay-krishna:
/close
Closing this because this code doesn't support our migration to regional curated packages and will need significant changes to get there. If needed, it can be re-opened at a later time.

PR #5760, opened 2023-05-02
Add bundle manifest URL to EKS-A version command

abhay-krishna:
The EKS Anywhere deliverables include not just the CLI but also a bundle of artifacts used to bring up a working EKS-A cluster. There have been many cases where EKS-A customers want to know what bundle will be used by default in the cluster creation. In our docs, we have some steps to get the bundle manifest associated with an EKS-A release but this requires the customer to download additional tools like `yq`. Most customers find this only when running cluster operations but only when verbose logging is enabled, which is not ideal. Instead of all this, we can present them this info as an output in the `eksctl anywhere version` command, with an optional output flag to print in json. This will also send the message that the bundle is as much a part of EKS-A as the CLI. This is analogous to the output of Flux version which also includes the controller versions along with CLI version.
```
$ flux version
flux: v0.41.1
helm-controller: v0.12.0
kustomize-controller: v0.15.2
notification-controller: v0.17.0
source-controller: v0.16.0
```

#### Dev
```bash
$ eksctl anywhere version                                                             
Version: v0.0.0-dev
Bundle manifest URL: https://dev-release-assets.eks-anywhere.model-rocket.aws.dev/bundle-release.yaml
```
#### Prod
```bash
$ eksctl anywhere version                                                             
Version: v0.18.3
Bundle manifest URL: https://anywhere-assets.eks.amazonaws.com/releases/bundles/54/manifest.yaml
```
#### Dev/JSON
```bash
$ eksctl anywhere version -o json
{"version":"v0.0.0-dev","bundleManifestURL":"https://dev-release-assets.eks-anywhere.model-rocket.aws.dev/bundle-release.yaml"}
```
#### Dev/YAML
```bash
$ eksctl anywhere version -o yaml
bundleManifestURL: https://dev-release-assets.eks-anywhere.model-rocket.aws.dev/bundle-release.yaml
version: v0.0.0-dev
```

### DIFF
--- a/cmd/eksctl-anywhere/cmd/version.go
+++ b/cmd/eksctl-anywhere/cmd/version.go
@@ -1,27 +1,65 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
+	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
 
 	"github.com/aws/eks-anywhere/pkg/version"
 )
+
+type versionOptions struct {
+	output string
+}
+
+var vo = &versionOptions{}
 
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Get the eksctl anywhere version",
 	Long:  "This command prints the version of eksctl anywhere",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return printVersion()
+		return vo.printVersion()
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
+	versionCmd.Flags().StringVarP(&vo.output, "output", "o", "", "specifies the output format (valid option: json)")
 }
 
-func printVersion() error {
-	fmt.Println(version.Get().GitVersion)
+func (vo *versionOptions) printVersion() error {
+	versionInfo, bundlesErr := version.GetFullVersionInfo()
+	switch vo.output {
+	case "":
+		fmt.Printf("Version: %s\n", versionInfo.GitVersion)
+		if bundlesErr != nil {
+			return fmt.Errorf("error getting bundle manifest URL for version %s: %v", versionInfo, bundlesErr)
+		}
+		fmt.Printf("Bundle Manifest URL: %s\n", versionInfo.BundleManifestURL)
+	case "json":
+		versionInfoJSON, unmarshalErr := json.Marshal(versionInfo)
+		if unmarshalErr != nil {
+			return fmt.Errorf("error marshaling version info to JSON: %v", unmarshalErr)
+		}
+		fmt.Printf("%s\n", versionInfoJSON)
+		if bundlesErr != nil {
+			return fmt.Errorf("error getting bundle manifest URL for version %s: %v", versionInfo, bundlesErr)
+		}
+	case "yaml":
+		versionInfoYAML, unmarshalErr := yaml.Marshal(versionInfo)
+		if unmarshalErr != nil {
+			return fmt.Errorf("error marshaling version info to YAML: %v", unmarshalErr)
+		}
+		fmt.Printf("%s\n", versionInfoYAML)
+		if bundlesErr != nil {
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -1,13 +1,51 @@
...
+	}, nil
+}
+
+// GetFullVersionInfoJSON returns the complete version information for the
+// EKS Anywhere, including Git version and bundle manifest URL
+// associated with this release, formatted as a JSON string.
+func GetFullVersionInfoJSON() string {
+	versionInfo, err := GetFullVersionInfo()
+	if err != nil {
+		return ""
```
jacobweinstock, line 45: if `GetFullVersionInfo` returns an error it seems like it will be impossible to troubleshoot, no?

abhay-krishna: I can make the output similar to `GetFullVersionInfoJSON`'s error case, where it gets the EKS-A Git version atleast and then errors out.

cmd/eksctl-anywhere/cmd/version.go
```diff
@@ -8,20 +8,44 @@ import (
...
+		versionInfo, err := version.GetFullVersionInfo()
+		fmt.Printf("Version: %s\n", versionInfo.GitVersion)
+		if err != nil {
+			return fmt.Errorf("error getting bundle manifest URL for version %s: %v", versionInfo, err)
+		}
+		fmt.Printf("Bundle manifest URL: %s\n", versionInfo.BundleManifestURL)
+	case "json":
+		versionInfoJSON, err := version.GetFullVersionInfoJSON()
+		if err != nil {
+			return fmt.Errorf("error getting full version info in JSON: %v, err")
```
d8660091, line 43:
```suggestion
			return fmt.Errorf("error getting full version info in JSON: %v", er")
```

cmd/eksctl-anywhere/cmd/version.go
```diff
@@ -8,20 +8,44 @@ import (
...
-	fmt.Println(version.Get().GitVersion)
+func (vo *versionOptions) printVersion() error {
+	switch vo.output {
+	case "":
+		versionInfo, err := version.GetFullVersionInfo()
+		fmt.Printf("Version: %s\n", versionInfo.GitVersion)
+		if err != nil {
+			return fmt.Errorf("error getting bundle manifest URL for version %s: %v", versionInfo, err)
+		}
+		fmt.Printf("Bundle manifest URL: %s\n", versionInfo.BundleManifestURL)
```
jacobweinstock, line 39: the flux example has all components printed in lowercase and without any spaces. This is advantageous when parsing with command line tools. Might be something to consider here too. 

```bash
version: v0.0.0-dev
bundle-url: https://dev-release-assets.eks-anywhere.model-rocket.aws.dev/bundle-release.yaml
```

pkg/manifests/releases/read.go
```diff
@@ -41,6 +41,22 @@ func ReadReleasesFromURL(reader Reader, url string) (*releasev1.Release, error)
 	return release, nil
 }
 
+// GetBundleManifestURLForVersion fetches the bundle manifest URL pertaining to this
+// release version of EKS Anywhere.
+func GetBundleManifestURLForVersion(reader Reader, version string) (string, error) {
```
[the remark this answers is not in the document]
abhay-krishna: Done.

pkg/version/version.go
abhay-krishna: Fixed this by surfacing all errors. In case of an error, the Git Version will be printed out, and will be followed by the error log.

cmd/eksctl-anywhere/cmd/version.go
abhay-krishna: I made the version output look similar to `kubectl version` whose default format includes capitalization and spaces. I have included an `output` flag which allows printing the version in JSON or YAML, and the JSON tags used for the Info struct is easily parseable (`version`, `bundleManifestURL`, etc)

jaxesn (approved):

PR #5774, opened 2023-05-03
Encode git private key using base64

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Tinkerbell end to end tests are failing with an error for git private keys which are in invalid format. After looking into the ssm commands we run for e2e, what stood out was when the generated private key was getting written to a file on the instance, the format was getting messed up.

In this PR, we are encoding the generated private key, and then writing back the decoded value of that private key using base64. This will ensure that the format stays intact.

### DIFF
--- a/internal/test/e2e/fluxGit.go
+++ b/internal/test/e2e/fluxGit.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #5777, opened 2023-05-03
Isolate cluster upgrades for management and workload clusters

abhay-krishna:
Implementation of proposed solution in #5622.

/hold

### DIFF
--- a/config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
+++ b/config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
@@ -469,6 +469,15 @@ spec:
           status:
             description: ClusterStatus defines the observed state of Cluster.
             properties:
+              childrenReconciledGeneration:
+                description: 'ChildrenReconciledGeneration represents the sum of the
+                  .metadata.generation for all the linked objects for the cluster,
+                  observed the last time the cluster was successfully reconciled.
+                  NOTE: This field was added for internal use and we do not provide
+                  guarantees to its behavior if changed externally. Its meaning and
+                  implementation are subject to change in the future.'
+                format: int64
+                type: integer
               conditions:
                 items:
                   description: Condition defines an observation of a Cluster API resource
@@ -541,6 +550,15 @@ spec:
                 description: Descriptive message about a fatal problem while reconciling
                   a cluster
                 type: string
+              reconciledGeneration:
+                description: 'ReconciledGeneration represents the .metadata.generation
+                  the last time the cluster was successfully reconciled. It is the
+                  latest generation observed by the controller. NOTE: This field was
+                  added for internal use and we do not provide guarantees to its behavior
+                  if changed externally. Its meaning and implementation are subject
+                  to change in the future.'
+                format: int64
+                type: integer
             type: object
         type: object
     served: true
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -59,7 +59,7 @@ type ClusterReconciler struct {
 // PackagesClient handles curated packages operations from within the cluster
 // controller.
 type PackagesClient interface {
-	EnableFullLifecycle(ctx context.Context, log logr.Logger, clusterName string, kubeConfig string, chart *v1alpha1.Image, registry *registrymirror.RegistryMirror, options ...curatedpackages.PackageControllerClientOpt) error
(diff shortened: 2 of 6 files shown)


controllers/cluster_controller_test.go
```diff
@@ -115,6 +118,40 @@ func TestClusterReconcilerReconcileSelfManagedCluster(t *testing.T) {
 	g.Expect(result).To(Equal(ctrl.Result{}))
 }
 
+func TestClusterReconcilerSelfManagedClusterWithMatchingGenerations(t *testing.T) {
```
g-gaston, line 121: could we add child objs to this test? That makes it more realistic. IRL we will never reconcile a cluster without linked objects

Also, I would add tests for the different no matching generations:
- same generation, diff aggregated generation
- diff aggregation, same aggregated generation
- diff aggregation, diff aggregated generation

this can easily be a table test

controllers/cluster_controller.go
```diff
@@ -384,21 +416,26 @@ func (r *ClusterReconciler) reconcileDelete(ctx context.Context, log logr.Logger
...
-		return err
+		return nil, err
 	}
 
 	childObjs := config.ChildObjects()
+
+	return childObjs, nil
+}
+
+func (r *ClusterReconciler) ensureClusterOwnerReferences(ctx context.Context, clus *anywherev1.Cluster, childObjs []kubernetes.Object) error {
```
g-gaston, line 435: I would take the clusterConfig as input, which contains both the cluster and the child objs

controllers/cluster_controller.go
```diff
@@ -384,21 +416,26 @@ func (r *ClusterReconciler) reconcileDelete(ctx context.Context, log logr.Logger
 	return ctrl.Result{}, nil
 }
 
-func (r *ClusterReconciler) ensureClusterOwnerReferences(ctx context.Context, clus *anywherev1.Cluster) error {
+func (r *ClusterReconciler) getChildObjectsForCluster(ctx context.Context, clus *anywherev1.Cluster) ([]kubernetes.Object, error) {
```
g-gaston, line 419: this method is a bit terse
I would lean towards returning the cluster.Config like `buildClusterConfig`
And then let the callers use whatever they need from it

controllers/cluster_controller.go
```diff
@@ -236,14 +237,33 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+	// for all the child objects of this cluster.
+	for _, obj := range childObjs {
+		aggregatedGeneration += obj.GetGeneration()
+	}
+
+	// If there is no difference between the aggregated generation and childrenReconciledGeneration,
+	// and there is no difference in the reconciled generation and .metadata.generation of the cluster,
+	// then return without any further processing.
+	if aggregatedGeneration == cluster.Status.ChildrenReconciledGeneration && cluster.Status.ReconciledGeneration == cluster.Generation {
+		return ctrl.Result{}, nil
```
g-gaston, line 260: maybe worth adding a log here
```suggestion
        log.Info("Generation and child generation match reconciled generation, skipping reconciliation.")
		return ctrl.Result{}, nil
```

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
```diff
@@ -541,6 +547,12 @@ spec:
                 description: Descriptive message about a fatal problem while reconciling
                   a cluster
                 type: string
+              reconciledGeneration:
```
g-gaston, line 550: what do you think about adding here in the description that these are for internal use and there is no grantee around them. same in both fields

controllers/cluster_controller_test.go
abhay-krishna: Will update!

controllers/cluster_controller.go
abhay-krishna: Done.

controllers/cluster_controller.go
abhay-krishna: Done!

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
abhay-krishna: Good callout. will add a note

g-gaston:
Updated the tests myself
/unhold

taneyland (approved):

PR #5779, opened 2023-05-04
Add Proxy support for Nutanix

thunderboltsid:
Propagate Proxy config for Nutanix templates.

**How has this been tested?**
- Created a proxy with tinyproxy on a ubuntu server VM
- Configure `proxyConfiguration` on cluster spec to point to the proxy and create a cluster successfully ([logs](https://app.warp.dev/block/vBn52cJUC5EGcLJuEUZLSK))
- Correlate whether the traffic from cluster is being routed through  the proxy

### DIFF
--- a/pkg/providers/nutanix/template.go
+++ b/pkg/providers/nutanix/template.go
@@ -241,24 +241,9 @@ func buildTemplateMapCP(
 
 	if clusterSpec.Cluster.Spec.ProxyConfiguration != nil {
 		values["proxyConfig"] = true
-		capacity := len(clusterSpec.Cluster.Spec.ClusterNetwork.Pods.CidrBlocks) +
-			len(clusterSpec.Cluster.Spec.ClusterNetwork.Services.CidrBlocks) +
-			len(clusterSpec.Cluster.Spec.ProxyConfiguration.NoProxy) + 4
-		noProxyList := make([]string, 0, capacity)
-		noProxyList = append(noProxyList, clusterSpec.Cluster.Spec.ClusterNetwork.Pods.CidrBlocks...)
-		noProxyList = append(noProxyList, clusterSpec.Cluster.Spec.ClusterNetwork.Services.CidrBlocks...)
-		noProxyList = append(noProxyList, clusterSpec.Cluster.Spec.ProxyConfiguration.NoProxy...)
-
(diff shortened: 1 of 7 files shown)


pkg/providers/nutanix/testdata/eksa-cluster-proxy.yaml
```diff
@@ -0,0 +1,78 @@
...
+    cni: "cilium"
+    pods:
+      cidrBlocks:
+        - 192.168.0.0/16
+    services:
+      cidrBlocks:
+        - 10.96.0.0/12
+  proxyConfiguration:
+    httpProxy: proxy.nutanix.com:8888
+    httpsProxy: proxy.nutanix.com:8888
```
abhinavmpandey08, line 41: mind adding a noProxy here as well for validation?
```suggestion
    httpsProxy: proxy.nutanix.com:8888
    noProxy:
      - noProxy1
      - noProxy2
      - noProxy3
```

pkg/providers/nutanix/template.go
```diff
@@ -239,6 +239,28 @@ func buildTemplateMapCP(
...
+		// Add no-proxy defaults
+		noProxyList = append(noProxyList, clusterapi.NoProxyDefaults()...)
+		noProxyList = append(noProxyList,
+			datacenterSpec.Endpoint,
+			clusterSpec.Cluster.Spec.ControlPlaneConfiguration.Endpoint.Host,
+		)
+
+		values["httpProxy"] = clusterSpec.Cluster.Spec.ProxyConfiguration.HttpProxy
+		values["httpsProxy"] = clusterSpec.Cluster.Spec.ProxyConfiguration.HttpsProxy
+		values["noProxy"] = noProxyList
```
panktishah26, line 261: For control plane and worker node, since we use the same noProxy list, can we get it from a common place by creating a function like [this](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/template.go#L641-L659)?

panktishah26 (commented):
Thanks for this PR. Your changes looks good. I have added a small comment.

pkg/providers/nutanix/template.go
thunderboltsid: Sure; fixed! PTAL!

PR #5784, opened 2023-05-04
Add Users and SSH key validation in cloudstackmachineconfig webhook

cxbrowne1207:
*Issue #, if available:*
Part of [Provider validations for ssh key #1217]( https://github.com/aws/eks-anywhere-internal/issues/1217)

*Description of changes:*
This PR adds user and ssh key validation to the `cloudstackmachineconfig` webhook. This doesn't affect the current behavior for the CLI, because it defaults users and ssh keys if not provided.

*Testing (if applicable):*
Unit tests
- Create and upgrade a mgmt cluster:
- Create workload cluster with API.
    - No users, user with empty name,  user with empty SSHAuthorizedKeys throws webhook error
    - User specified with name and valid sshAuthorizationKey succeeds
- Update workload cluster with API.
    - No users, user with empty name,  user with empty SSHAuthorizedKeys throws webhook error
    - User specified with name and valid sshAuthorizationKey succeeds
    
### DIFF
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
@@ -217,7 +217,7 @@ func (c *CloudStackMachineConfig) Validate() error {
 }
 
 // SetUserDefaults initializes Spec.Users for the CloudStackMachineConfig with default values.
-// This only runs in the CLI, as we support don't support user defaults through the webhook.
+// This only runs in the CLI, as we don't support user defaults through the webhook.
 func (c *CloudStackMachineConfig) SetUserDefaults() {
 	c.Spec.Users = defaultMachineConfigUsers(DefaultCloudStackUser, c.Spec.Users)
 }
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
@@ -217,9 +217,9 @@ func (c *CloudStackMachineConfig) Validate() error {
 }
 
 // SetUserDefaults initializes Spec.Users for the CloudStackMachineConfig with default values.
-// This only runs in the CLI, as we support do support user defaults through the webhook.
+// This only runs in the CLI, as we support don't support user defaults through the webhook.
 func (c *CloudStackMachineConfig) SetUserDefaults() {
-	c.Spec.Users = defaultMachineConfigUsers(c.Spec.Users)
+	c.Spec.Users = defaultMachineConfigUsers(DefaultCloudStackUser, c.Spec.Users)
 }
 
 // CloudStackMachineConfigStatus defines the observed state of CloudStackMachineConfig.
--- a/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
+++ b/pkg/api/v1alpha1/cloudstackmachineconfig_types.go
@@ -147,6 +147,16 @@ func (r SymlinkMaps) Validate() (err error, field string, value string) {
 	return nil, "", ""
 }
 
+// ValidateUsers verifies a CloudStackMachineConfig object must have a users with ssh authorized keys.
+// This validation only runs in CloudStackMachineConfig validation webhook, as we support
+// auto-generate and import ssh key when creating a cluster via CLI.
+func (c *CloudStackMachineConfig) ValidateUsers() error {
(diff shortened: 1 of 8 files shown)


pkg/providers/cloudstack/validator.go
```diff
@@ -139,15 +138,6 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, clusterSp
...
 	for _, machineConfig := range clusterSpec.CloudStackMachineConfigs {
-		if len(machineConfig.Spec.Users) <= 0 {
-			machineConfig.Spec.Users = []anywherev1.UserConfiguration{{}}
-		}
-		if len(machineConfig.Spec.Users[0].SshAuthorizedKeys) <= 0 {
-			machineConfig.Spec.Users[0].SshAuthorizedKeys = []string{""}
-		}
-		if machineConfig.Spec.Users[0].Name == "" {
-			machineConfig.Spec.Users[0].Name = v1alpha1.DefaultCloudStackUser
-		}
```
cxbrowne1207, line 150: These aren't validations, so moving this logic to the SetupAndValidate provider methods

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -147,6 +148,22 @@ func (r SymlinkMaps) Validate() (err error, field string, value string) {
 	return nil, "", ""
 }
 
+// ValidateUsers verifies a CloudStackMachineConfig object must have a users with ssh authorized keys.
+// This validation only runs in CloudStackMachineConfig validation webhook, as we support
+// auto-generate and import ssh key when creating a cluster via CLI.
+func (c *CloudStackMachineConfig) ValidateUsers() error {
```
ddjjia, line 154: why not have a default webhook?

ddjjia: Do we need both `setCloudStackMachineConfigUserDefaults` and `ValidateUsers` to have the same logic?

cxbrowne1207: We won't support users for the web-hook because only the user name can be assigned by default and they are required to specify the `name` field anyway. If we can't default the users in full, I think it makes sense to just validate it.

It also decreases confusion on what to

cxbrowne1207: `setCloudStackMachineConfigUserDefaults` and `ValidateUsers` don't necessarily have the same logic. One mutates the `machineconfig`, while one just validates

cxbrowne1207: There might be a misunderstand. We're not using the validation webhook to perform defaulting.

The `users` are only being defaulted for the CLI because we auto generate the ssh keys. For the webhook, we opted to remove defaulting users entirely for just validations because we don't currently support auto generating ssh keys.

cxbrowne1207: Resolved in direct messages

pkg/api/v1alpha1/cloudstackmachineconfig_types.go
```diff
@@ -206,6 +216,12 @@ func (c *CloudStackMachineConfig) Validate() error {
 	return validateCloudStackMachineConfig(c)
 }
 
+// SetUserDefaults initializes Spec.Users for the CloudStackMachineConfig with default values.
+// This only runs in the CLI, as we support do support user defaults through the webhook.
```
vivek-koppuru, line 220:
```suggestion
// This only runs in the CLI, as we support user defaults through the webhook.
```

pkg/api/v1alpha1/machine_types.go
```diff
@@ -13,3 +15,30 @@ type UserConfiguration struct {
...
+
+func defaultMachineConfigUsers(users []UserConfiguration) []UserConfiguration {
+	if len(users) <= 0 {
+		users = []UserConfiguration{{}}
+	}
+	if len(users[0].SshAuthorizedKeys) <= 0 {
+		users[0].SshAuthorizedKeys = []string{""}
+	}
+	if users[0].Name == "" {
+		users[0].Name = DefaultCloudStackUser
```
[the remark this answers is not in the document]
cxbrowne1207: Ah, you're right. This was a mistake. Here's I'm trying to dry out this logic because it will be similar in other providers. I'll make the necessary adjustment

PR #5792, opened 2023-05-05
make cloudstack port defaulting in CAPI backward compatible

ddjjia:
*Issue #, if available:*
port is not backward compatible with [this change](https://github.com/aws/eks-anywhere/commit/4777d85689fe58b4effc2b60100431f8b754e87f)

*Description of changes:*
[one pager doc](https://quip-amazon.com/5LjTApbrKp8i/Control-Plane-Endpoint-Port)

CLI used to default cloudstack endpoint port in cluster object. Now the port only added when generating CAPI and cluster object remains the same with customer input. https://github.com/aws/eks-anywhere/commit/4777d85689fe58b4effc2b60100431f8b754e87f
This pr makes above change backward compatible when customers upgrade their cluster via CLI.
| Old Endpoint  | New Endpoint | Expected Result | 
| ------------- | ------------- | ------------- |
| no port | no port | depending on host |
| no port | custom port  | fail |
| no port | default port | depending on host |
| default port | no port or default port | depending on host |
| default port| custom port  | fail |
| custom port | no port | fail | 
| custom port | default port | fail |
| custom port | same custom port | depending on host |
| custom port | different custom port | fail |

*Testing (if applicable):*
unit test and manual testing on mgmt via CLI:
- created mgmt cluster without port provided
- observed the default port is added to cluster object
- updated cli to include this [change](https://github.com/aws/eks-anywhere/commit/4777d85689fe58b4effc2b60100431f8b754e87f)
- upgraded the cluster and had immutable error
- updated cli to this pr version
- upgraded the cluster successfully and no port in cluster object


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -87,6 +93,18 @@ func (n *Cluster) Equal(o *Cluster) bool {
 	if n.Spec.KubernetesVersion != o.Spec.KubernetesVersion {
 		return false
 	}
+
+	switch n.Spec.DatacenterRef.Kind {
```
pokearu, line 97: Can we move this to its own method and add it as part of the return??

```
return n.Count == o.Count && n.MachineGroupRef.Equal(o.MachineGroupRef) &&
		TaintsSliceEqual(n.Taints, o.Taints) && MapEqual(n.Labels, o.Labels) && CPEndpointEqual()
```

ddjjia: Do you mean add to `func (n *ControlPlaneConfiguration) Equal(o *ControlPlaneConfiguration) bool {`, and pass datacenterkind there?

pokearu (approved):

PR #5798, opened 2023-05-05
Add validation for tinkerbell ip for workload cluster to match management cluster

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
We are disallowing the ability to specify a different workload cluster tinkerbellip from the management cluster as we only look at the management cluster one as the source of truth. It is misleading when we override the value specified by the user in the CLI, so this will throw an error if the wrong ip is specified. 

This behavior matches what we do in the controller now: https://github.com/aws/eks-anywhere/pull/4808

*Testing (if applicable):*
unit tests

### DIFF
--- a/pkg/providers/tinkerbell/create.go
+++ b/pkg/providers/tinkerbell/create.go
@@ -152,8 +152,10 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		if err != nil {
 			return fmt.Errorf("getting TinkerbellIP of management cluster: %s", err)
 		}
-
-		p.datacenterConfig.Spec.TinkerbellIP = managementDatacenterConfig.Spec.TinkerbellIP
+		// Checking for empty first as that returns a different error in the datacenter config validate method below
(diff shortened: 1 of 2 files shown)


abhinavmpandey08 (approved):

pkg/providers/tinkerbell/create.go
```diff
@@ -152,8 +152,10 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 		if err != nil {
 			return fmt.Errorf("getting TinkerbellIP of management cluster: %s", err)
 		}
-
-		p.datacenterConfig.Spec.TinkerbellIP = managementDatacenterConfig.Spec.TinkerbellIP
+		// Checking for empty first as that returns a different error in the datacenter config validate method below
+		if p.datacenterConfig.Spec.TinkerbellIP != "" && p.datacenterConfig.Spec.TinkerbellIP != managementDatacenterConfig.Spec.TinkerbellIP {
```
chrisdoherty4, line 156: Do we have defaulting somewhere the populates the value if its empty?

chrisdoherty4 (approved):

pkg/providers/tinkerbell/create.go
vivek-koppuru: Not that I am aware of no, but we were doing this defaulting here below so I would save that until we have other defaulting.

PR #5809, opened 2023-05-09
Vsphere e2e tests

rahulbabu95:
*Issue #, if available:*
The change adds Adot packages test for Vsphere provider. While the update flow was covered for Adot packages in Vsphere, simple flow for Adot packages were missing in our E2E tets.
*Description of changes:*

*Testing (if applicable):*


rahulbabu95:
We have these tests running for other providers and saw we were missing it for Vsphere. We can definitely discuss if we want to have these or not but that would be a different discussion, since we have more tests than these lot running related to Adot across providers.?

PR #5813, opened 2023-05-09
Reduce controller's log verbosity

g-gaston:
## Description of changes

Defaulting to v=4 was printing a lot of client-go and controller-runtime "debug" logs that added a lot of noise but no value.

Now it defaults to 0 and for tilt we bump it to 2, which adds a little more information from controller-runtime, specially during startup, but still avoids unnecessary noise.

### DIFF
--- a/pkg/controller/clusters/ipvalidator.go
+++ b/pkg/controller/clusters/ipvalidator.go
@@ -41,7 +41,7 @@ func (i *IPValidator) ValidateControlPlaneIP(ctx context.Context, log logr.Logge
 	if capiCluster != nil {
 		// If CAPI cluster exists, the control plane IP has already been validated,
 		// and it's possibly already in use so no need to validate it again
-		log.V(3).Info("CAPI cluster already exists, skipping control plane IP validation")
+		log.Info("CAPI cluster already exists, skipping control plane IP validation")
 		return controller.Result{}, nil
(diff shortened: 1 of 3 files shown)


pkg/controller/clusters/ipvalidator.go
```diff
@@ -41,7 +41,7 @@ func (i *IPValidator) ValidateControlPlaneIP(ctx context.Context, log logr.Logge
 	if capiCluster != nil {
 		// If CAPI cluster exists, the control plane IP has already been validated,
 		// and it's possibly already in use so no need to validate it again
-		log.V(3).Info("CAPI cluster already exists, skipping control plane IP validation")
```
ddjjia, line 44: change the log level here as well?
https://github.com/aws/eks-anywhere/blob/c4a3d14e7ec7aa58d91e5ad02a550ce269109118/pkg/providers/vsphere/reconciler/reconciler.go#L221

g-gaston: Sorry, what is the question? :)
I'm changing the log level of this line so the log actually shows, since this was level 3 and now we default to 0.

ddjjia: I mean we should change the vsphere storage class log level to 0 as well

Issue #5757, opened 2023-05-01
Add VMware permissions validation in the upgrade workflow


taneyland:
Users want a way to validate VMWare permissions before attempting operations. We already added this for the create workflow in EKS-A v0.11.2 (PR- https://github.com/aws/eks-anywhere/pull/2907). We should add these same validations for upgrades too.

csplinter:
Is this just a miss from #2907 ?

taneyland:
> Is this just a miss from #2907 ?

Talked with @jonathanmeier5 offline, and yes I believe you are correct. We already added this in 0.11.2 for the create workflow. So I believe we will just have to add the same validations for upgrades.

PR #5838, opened 2023-05-10
Validate VSphere user privileges during upgrade

taneyland:
*Issue #, if available:*
#5757 

*Description of changes:*
Adding the method to validate Vsphere privs to the upgrade flow.
Also moving the method to validator.go

*Testing (if applicable):*


jonathanmeier5 (commented):
Overall looks good. Could we at least get the missing error cases test coverage covered in vsphere/vsphere.go and see where codecov looks from there?

PR #5841, opened 2023-05-10
Make registry mirror e2es use bundles override

ahreehong:
*Issue #, if available:*

*Description of changes:*
Make registry mirror e2es use bundles override to avoid failing with mismatched tags during create vs delete

*Testing (if applicable):*

### DIFF
--- a/test/e2e/registrymirror.go
+++ b/test/e2e/registrymirror.go
@@ -13,8 +13,8 @@ func runRegistryMirrorConfigFlow(test *framework.ClusterE2ETest) {
 	test.ExtractDownloadedArtifacts()
 	test.DownloadImages()
 	test.ImportImages()
-	test.CreateCluster()
-	test.DeleteCluster()
+	test.CreateCluster(framework.WithBundlesOverride(bundleReleasePathFromArtifacts))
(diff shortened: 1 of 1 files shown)


test/e2e/registrymirror.go
```diff
@@ -13,8 +13,8 @@ func runRegistryMirrorConfigFlow(test *framework.ClusterE2ETest) {
 	test.ExtractDownloadedArtifacts()
 	test.DownloadImages()
 	test.ImportImages()
-	test.CreateCluster()
-	test.DeleteCluster()
+	test.CreateCluster(framework.WithBundlesOverride("./eks-anywhere-downloads/bundle-release.yaml"))
```
panktishah26, line 16: Do we download bundle release file for each test so `./eks-anywhere-downloads/bundle-release.yaml` this path will always have the bundle release file?

ahreehong: yeah the bundle release file gets saved in that path when ExtractDownloadArtifacts() runs

abhinavmpandey08, line 16: nit: can you use this var instead https://github.com/aws/eks-anywhere/blob/main/test/e2e/airgap.go#L20

abhinavmpandey08 (approved):

abhinavmpandey08 (approved):

PR #5855, opened 2023-05-11
Add cloudstack workload cluster API e2e tests 

cxbrowne1207:
*Issue #, if available:*
[#1237 (internal)](https://github.com/aws/eks-anywhere-internal/issues/1237)

*Description of changes:*

This PR add an e2e test that that tests the following features for the CloudStack API. Here we create a management cluster and workload cluster(s) and test it's features using sub tests. The benefits of this approach is the test utilizes a single management cluster rather than spinning a new one up every time to test a workload cluster feature.
- create and delete workload clusters for all supported k8 versions
- add and remove labels and taints
- scale up cp and replace existing worker node groups
- scale down cp and scaling worker node groups

See the testing doc for the full list of scenarios we are going to eventually cover [here](https://quip-amazon.com/99bUA1xK7bt0/CloudStack-Full-Lifecycle-API-Testing
*Testing (if applicable):*
- Ran the e2e tests locally

```
--- PASS: TestCloudStackMulticlusterWorkloadClusterAPI (1846.19s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterAPI/add_and_remove_labels_and_taints (5.26s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterAPI/scale_up/down_cp_and_worker_node_group (415.61s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterAPI/replace_existing_worker_node_groups (115.70s)
PASS

```

```
--- PASS: TestCloudStackMulticlusterWorkloadClusterGitHubFluxAPI (2112.32s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterGitHubFluxAPI/add_and_remove_labels_and_taints (6.31s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterGitHubFluxAPI/scale_up_and_down_cp_and_worker_node_group (530.08s)
    --- PASS: TestCloudStackMulticlusterWorkloadClusterGitHubFluxAPI/replace_existing_worker_node_groups (140.74s)
PASS

```

### DIFF
--- a/test/e2e/cloudstack_test.go
+++ b/test/e2e/cloudstack_test.go
@@ -1083,9 +1083,7 @@ func TestCloudStackMulticlusterWorkloadClusterAPI(t *testing.T) {
 
 		for _, tt := range tests {
 			t.Run(tt.name, func(t *testing.T) {
-				wc.UpdateClusterConfig(tt.upgradeFiller)
-				wc.ApplyClusterManifest()
-				wc.ValidateClusterState()
+				runCloudStackAPIUpgradeTest(t, wc, tt)
 			})
 		}
 
@@ -1144,8 +1142,7 @@ func TestCloudStackMulticlusterWorkloadClusterGitHubFluxAPI(t *testing.T) {
 
 		for _, tt := range tests {
 			t.Run(tt.name, func(t *testing.T) {
-				test.PushWorkloadClusterToGit(wc, tt.upgradeFiller)
-				wc.ValidateClusterState()
+				runCloudStackAPIUpgradeTestWithFlux(t, test, wc, tt)
 			})
 		}
 
--- a/test/e2e/cloudstack_test.go
+++ b/test/e2e/cloudstack_test.go
@@ -1028,10 +1028,6 @@ func TestCloudStackKubernetes123UbuntuAirgappedRegistryMirror(t *testing.T) {
 	runAirgapConfigFlow(test, "10.0.0.1/8")
 }
 
-func runCloudStackMulticlusterWorkloadClusterAPITest() {
-
-}
-
 // Workload API
 
 func TestCloudStackMulticlusterWorkloadClusterAPI(t *testing.T) {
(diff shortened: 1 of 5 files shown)


test/framework/clustervalidator.go
```diff
@@ -81,14 +81,18 @@ func buildClusterClient(kubeconfigFileName string) (client.Client, error) {
...
 
 func buildClusterSpec(ctx context.Context, client client.Client, config *cluster.Config) (*cluster.Spec, error) {
-	clus := &v1alpha1.Cluster{}
-	key := machinerytypes.NamespacedName{Namespace: config.Cluster.Namespace, Name: config.Cluster.Name}
-	if err := client.Get(ctx, key, clus); err != nil {
-		return nil, fmt.Errorf("failed to get cluster to build spec: %s", err)
+	clusterConfig := config.DeepCopy()
+	// The cluster config built by the test does not have certain defaults like the bundle reference,
+	// so fetch that information from the cluster if missing. This is needed inorder to build the cluster spec.
+	if clusterConfig.Cluster.Spec.BundlesRef == nil {
```
ddjjia, line 87: How do we validate "default" bundle then, like the case that only update mgmt bundle, how do we make sure wl cluster bundle doesn't get effected?

cxbrowne1207: So, in that test we just grab the EKSA Cluster object from the cluster before upgrading the management cluster and read the bundleRef from there. After updating the management cluster, we grab it again for the updated bundle.

The CLI has some logic that "defaults" the bundle based on different version details and then applies it to the cluster. It's kind of complex to replicate so this code here is to sort bridge that gap in the test suite (I think the controller does something similar as well)

test/e2e/cloudstack_test.go
```diff
@@ -1039,3 +1040,116 @@ func TestCloudStackKubernetes123UbuntuAirgappedRegistryMirror(t *testing.T) {
...
+	test.RunConcurrentlyInWorkloadClusters(func(wc *framework.WorkloadCluster) {
+		test.PushWorkloadClusterToGit(wc)
+		wc.WaitForKubeconfig()
+		wc.ValidateClusterState()
+		tests := cloudStackAPIUpgradeTests(cloudstack)
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				runCloudStackAPIUpgradeTestWithFlux(t, test, wc, tt)
+			})
```
chrisdoherty4, line 1146: This logic won't create individual tests for those listed in `cloudStackAPIUpgradeTests()`, is that intentional?

chrisdoherty4 (approved):

PR #5890, opened 2023-05-19
Add cloudstack management cluster and upgrade latest minor release API e2e tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


test/framework/cloudstack.go
```diff
@@ -302,3 +302,14 @@ func (c *CloudStack) WithRedhat123() api.ClusterConfigFiller {
 		),
 	)
 }
+
```
ddjjia, line 305: why is this needed?

test/e2e/cloudstack_test.go
```diff
@@ -1801,6 +1822,45 @@ func TestCloudStackMulticlusterWorkloadClusterAPI(t *testing.T) {
...
+		),
+		cloudstack.WithRedhat123(),
+	)
+	test.WithWorkloadClusters(wc)
+
+	runMulticlusterUpgradeFromReleaseFlowAPI(
+		test,
+		release,
+		api.JoinClusterConfigFillers(),
+		api.JoinClusterConfigFillers(),
```
ddjjia, line 1860: dup?

test/framework/cloudstack.go
cxbrowne1207: Ah, are you referring to the comment? My bad, i forgot to remove it

test/framework/cluster/validations/cluster.go
```diff
@@ -60,6 +60,35 @@ func ValidateEKSAObjects(ctx context.Context, vc clusterf.StateValidationConfig)
 	return nil
 }
 
+// ValidateBundleKubernetesTag is a loose check for checking  that the bundle is in use by the cluster.
```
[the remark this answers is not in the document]
cxbrowne1207: for example, it would be good to check if the correct version of the CNI is installed and other key components.

I'm not really convinced the right solution for this yet is to validate the state of the cluster to the bundle because it is someone complex to do so. Therefore, the task was made to also explore exactly what needs to be done here as well.

PR #5908, opened 2023-05-23
Skip tests with known failures and limit the concurrent test runs for Cloudstack provider

rahulbabu95:
*Issue #, if available:*
Airgapped profile doesn't exist for CS provider and so the airgapped validations are failing. Upgrade from latest minor version tests for 1.24 will fail until a patch version of 16.x is released. Hence skipping these tests until then. Limit concurrent runs to 35 to check if CS env is able to handle the test runs consistently with stability.

*Description of changes:*

*Testing (if applicable):*


pokearu (approved):
/approve
/lgtm

🤞

PR #5910, opened 2023-05-23
Patch incorrect CloudStack upgrade comparison

chrisdoherty4:
# Summary

A subset of fields for the EKSA CloudStack API do not trigger creation of new CAPI templates when they are removed. The issue stems from the DeepDerivative function used to compare old with new to identify a diff. It does not compare fields when their type is defaulted meaning removal of the data all together isn't considered a change.

This change ensures we identify a diff when removing the fields that would have previously gone unnoticed.

__Example__

```go
old := CloudStackMachineConfig{
  ... 
  UserCustomDetails: map[string]string{"foo": "bar"}
}

new := CloudStackMachineConfig{
  ...
  UserCustomDetails: nil
}

// DeepDerivative(new, old) wouldn't identify a diff because UserCustomDetails is set
// to its default value.
```

Also see https://github.com/aws/eks-anywhere-internal/issues/1327.


chrisdoherty4:
/hold

When relying on default values in the CAPI/Infra providers set when the manifest is applied to the cluster, the DeepDerivative check in the inverse will see it as a change and roll new nodes.

PR #5914, opened 2023-05-24
Only move management cluster objects during management cluster upgrade

vignesh-goutham:
*Description of changes:*
When a management cluster is getting upgraded, in the current workflow, all the CAPI clusters, their respective dependents including machines, secrets, etc get moved from the workload management cluster to bootstrap cluster. After the workload cluster gets upgraded, the management components are moved back from bootstrap cluster back to the workload cluster. This process is error prone and an issue could result in all CAPI objects being lost. Another issue with this is checking and expecting the status of all clusters to be ready during management cluster upgrade which is not always possible, especially in a more established environment.

This change only checks for management cluster's status and only moves the management cluster objects back and forth to the bootstrap cluster. All other clusters are paused during this process.

/hold till build tooling is updated with change for `clusterctl move --filter-cluster` flag

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -99,7 +99,7 @@ type ClusterManager struct {
 type ClusterClient interface {
 	KubernetesClient
 	BackupManagement(ctx context.Context, cluster *types.Cluster, managementStatePath string) error
-	MoveManagement(ctx context.Context, org, target *types.Cluster) error
+	MoveManagement(ctx context.Context, from, target *types.Cluster, clusterName string) error
 	WaitForClusterReady(ctx context.Context, cluster *types.Cluster, timeout string, clusterName string) error
 	WaitForControlPlaneAvailable(ctx context.Context, cluster *types.Cluster, timeout string, newClusterName string) error
 	WaitForControlPlaneReady(ctx context.Context, cluster *types.Cluster, timeout string, newClusterName string) error
@@ -122,6 +122,8 @@ type ClusterClient interface {
 	SaveLog(ctx context.Context, cluster *types.Cluster, deployment *types.Deployment, fileName string, writer filewriter.FileWriter) error
 	GetMachines(ctx context.Context, cluster *types.Cluster, clusterName string) ([]types.Machine, error)
 	GetClusters(ctx context.Context, cluster *types.Cluster) ([]types.CAPICluster, error)
+	PauseCAPICluster(ctx context.Context, cluster, kubeconfig string) error
+	ResumeCAPICluster(ctx context.Context, cluster, kubeconfig string) error
 	GetEksaCluster(ctx context.Context, cluster *types.Cluster, clusterName string) (*v1alpha1.Cluster, error)
 	GetEksaVSphereDatacenterConfig(ctx context.Context, VSphereDatacenterName string, kubeconfigFile string, namespace string) (*v1alpha1.VSphereDatacenterConfig, error)
 	UpdateEnvironmentVariablesInNamespace(ctx context.Context, resourceType, resourceName string, envMap map[string]string, cluster *types.Cluster, namespace string) error
@@ -313,8 +315,8 @@ func (c *ClusterManager) MoveCAPI(ctx context.Context, from, to *types.Cluster, 
 		return err
 	}
 
-	logger.V(3).Info("Waiting for all clusters to be ready before move")
-	if err := c.waitForAllClustersReady(ctx, from, c.clusterWaitTimeout.String()); err != nil {
+	logger.V(3).Info("Waiting for management cluster to be ready before move")
+	if err := c.clusterClient.WaitForClusterReady(ctx, from, c.clusterWaitTimeout.String(), clusterName); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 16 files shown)


pkg/executables/clusterctl.go
```diff
@@ -172,12 +172,16 @@ func (c *Clusterctl) BackupManagement(ctx context.Context, cluster *types.Cluste
...
 }
 
-func (c *Clusterctl) MoveManagement(ctx context.Context, from, to *types.Cluster) error {
+func (c *Clusterctl) MoveManagement(ctx context.Context, from, to *types.Cluster, clusterName string) error {
 	params := []string{"move", "--to-kubeconfig", to.KubeconfigFile, "--namespace", constants.EksaSystemNamespace}
 	if from.KubeconfigFile != "" {
 		params = append(params, "--kubeconfig", from.KubeconfigFile)
 	}
 
+	if clusterName != "" {
```
g-gaston, line 183: do we have a usecase to move all clusters?
I would lean to make this argument mandatory. That would avoid errors where we end up passing an empty cluster name my accident.

And if we ever need to move all clusters, we just create a `MoveAllClusters` method. Which is more explicit.

pkg/executables/kubectl.go
```diff
@@ -2008,6 +2008,34 @@ func (k *Kubectl) ApplyTolerationsFromTaints(ctx context.Context, oldTaints []co
 	return nil
 }
 
+// PauseCAPICluster adds a `spec.Paused: true` to the CAPI cluster resource. This will cause all
+// downstream CAPI + provider controllers to skip reconciling on the paused cluster's objects
+func (k *Kubectl) PauseCAPICluster(ctx context.Context, cluster, kubeconfig string) error {
+	patch := fmt.Sprintf("{\"spec\":{\"paused\":%t}}", true)
```
g-gaston, line 2014: I thought the canonical way to pause a cluster was through the pause annotation?
Which also simplifies this since we won't need the merge path.

pkg/clustermanager/cluster_manager.go
```diff
@@ -122,6 +122,8 @@ type ClusterClient interface {
 	SaveLog(ctx context.Context, cluster *types.Cluster, deployment *types.Deployment, fileName string, writer filewriter.FileWriter) error
 	GetMachines(ctx context.Context, cluster *types.Cluster, clusterName string) ([]types.Machine, error)
 	GetClusters(ctx context.Context, cluster *types.Cluster) ([]types.CAPICluster, error)
+	PauseCAPICluster(ctx context.Context, cluster, kubeconfig string) error
+	ResumeCAPICluster(ctx context.Context, cluster, kubeconfig string) error
```
g-gaston, line 126: I think we need to implement these two through the retrier client, so they get retried in case of transient errors.

pkg/workflows/upgrade.go
```diff
@@ -568,6 +575,14 @@ func (s *moveManagementToWorkloadTask) Run(ctx context.Context, commandContext *
 		return &CollectDiagnosticsTask{}
 	}
 	commandContext.ManagementCluster = commandContext.WorkloadCluster
+
+	logger.V(3).Info("Resuming all workload clusters after moving management cluster resources from bootstrap to workload clusters")
```
g-gaston, line 579:
```suggestion
	logger.V(3).Info("Resuming all workload clusters after moving management cluster resources from bootstrap to management cluster")
```

pkg/executables/clusterctl.go
vignesh-goutham: Sure I like that

pkg/executables/kubectl.go
vignesh-goutham: I think the pause annotation is recommended way to pause individual objects. Most if not all controllers call the helper function over [here at CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/83b8694556bc7986ee8573e93e65538957154b59/util/annotations/helpers.go#L29) in their reconcile loops to validate cluster paused. This looks for `spec.Paused` on their owner cluster or pause annotation on that specific object. If we add this spec.Pause we can get away without having to annotate all the dependents.

pkg/executables/kubectl.go
```diff
@@ -2008,6 +2008,34 @@ func (k *Kubectl) ApplyTolerationsFromTaints(ctx context.Context, oldTaints []co
...
+// ResumeCAPICluster removes the `spec.Paused` on the CAPI cluster resource. This will cause all
+// downstream CAPI + provider controllers to resume reconciling on the paused cluster's objects
+// `spec.Paused` is set to `null` to drop the field instead of setting it to `false`.
+func (k *Kubectl) ResumeCAPICluster(ctx context.Context, cluster, kubeconfig string) error {
+	patch := "{\"spec\":{\"paused\":null}}"
+	return k.MergePatchResource(ctx, capiClustersResourceType, cluster, patch, kubeconfig, constants.EksaSystemNamespace)
+}
+
+// MergePatchResource patches named resource using merge patch.
+func (k *Kubectl) MergePatchResource(ctx context.Context, resource, name, patch, kubeconfig, namespace string) error {
```
g-gaston, line 2027: i would lean towards making this unexported

g-gaston (approved):

Issue #5517, opened 2023-04-03
Remove CSI and storage class management from EKS-A for vSphere provider


jiayiwang7:
Currently we have a specific storage class installed here https://github.com/aws/eks-anywhere/blob/main/pkg/providers/vsphere/config/defaultStorageClass.yaml which is not flexible, as well as unnecessary management of a driver that is not required for running a cluster. So we want to remove it from the EKS Anywhere create/upgrade flow so users can manage CSI on their own if they require it, with an implementation for VSphere listed [here](https://docs.vmware.com/en/VMware-vSphere-Container-Storage-Plug-in/2.0/vmware-vsphere-csp-getting-started/GUID-C44D8071-85E7-4933-83EA-6797518C1837.html).

drewvanstone:
We are going to no longer install the CSI driver by default and will issue a warning that it will be removed in an upcoming release. This issue is being moved to the 0.17 release

vivek-koppuru:
With #5918, we will give a warning to all our users when `disableCSI` is set to `false` or not specified to tell them to switch to managing their CSI driver and StorageClass on their own if they require it.

PR #5918, opened 2023-05-25
Add warning to deprecate disableCSI through CLI

vivek-koppuru:
*Issue #, if available:*
#5517 

*Description of changes:*
Adding a warning that this field and functionality will not be supporting starting from the next release. Making this the first warning that appears when running the CLI for create and upgrade if `disableCSI` is set to `false`, which means that the user expects to have the vsphere csi driver and storage class installed on the cluster. We have documentation today on what it entails to disable it and we will add more information to the docs about this deprecation: https://anywhere.eks.amazonaws.com/docs/reference/clusterspec/vsphere/#disablecsi-optional

Output:
```
⚠️  Warning: Installing CSI through EKS Anywhere is deprecated. Refer to the official documentation for more details on the disableCSI field in VSphereDatacenterConfig
✅ Connected to server
```

I looked into how to add a warning to the webhook when creating workload clusters through kubectl, and it doesn't seem to be straightforward. Openapi supports adding the [deprecated](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#fixed-fields-10) field, but I can't find how to enable that through Kubebuilder except for `kubebuilder:deprecatedversion`, which [deprecates the whole api](https://github.com/kubernetes-sigs/kubebuilder/issues/2116#issuecomment-1020111737). I looked into seeing if we can throw the warning through the Validating webhook to have the warning displayed, but it looks like it was just merged recently [here](https://github.com/kubernetes-sigs/controller-runtime/pull/2014). Our best option here instead of trying to find a different way to do it is to just throw the warning through the CLI and omit the warning for the controller, as this message will be displayed first to users through the CLI anyways when they create/upgrade their management clusters. 

*Testing (if applicable):*
unit testing and functional testing

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -274,7 +274,17 @@ func (p *vsphereProvider) PostMoveManagementToBootstrap(_ context.Context, _ *ty
 	return nil
 }
 
+// TODO: Remove this field for v0.17.x, adding a warning as of v0.16.0
+func warnIfCSIEnabled(disableCSI bool) {
+	if !disableCSI {
+		// Need to add spacing to make the log message look neat
+		logger.MarkWarning("  Warning: Installing CSI through EKS Anywhere is deprecated. Refer to the official documentation for more details on " +
+			"the disableCSI field in VSphereDatacenterConfig")
+	}
+}
+
 func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpec *cluster.Spec) error {
+	warnIfCSIEnabled(clusterSpec.VSphereDatacenter.Spec.DisableCSI)
 	if err := p.validator.validateUpgradeRolloutStrategy(clusterSpec); err != nil {
 		return fmt.Errorf("failed setup and validations: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -274,7 +274,17 @@ func (p *vsphereProvider) PostMoveManagementToBootstrap(_ context.Context, _ *ty
 	return nil
 }
 
+// TODO: Remove this field for v0.17.x, adding a warning as of v0.16.0
+func warnIfCSIEnabled(disableCSI bool) {
+	if !disableCSI {
+		// Need to add spacing to make the log message look neat
+		logger.MarkWarning("  Warning: Installing CSI through EKS Anywhere is deprecated. Refer to the official documentation for more details on " +
```
chrisnegus, line 281: I think that wording is fine. Saying that something "is deprecated" means that it will be going away in the future, but has not gone away yet.

chrisnegus (commented):
I think the wording is fine. Did you mean to put a link to the docs in the message as well. A simple search for CSI came up with the right pages I think, so I think it's okay as it is.

jiayiwang7 (approved):

vivek-koppuru:
> I think the wording is fine. Did you mean to put a link to the docs in the message as well. A simple search for CSI came up with the right pages I think, so I think it's okay as it is.

I plan on updating the documentation but I didn't want to link it because the log message would get more verbose than it already is.

chrisnegus:
> > I think the wording is fine. Did you mean to put a link to the docs in the message as well. A simple search for CSI came up with the right pages I think, so I think it's okay as it is.
> 
> I plan on updating the documentation but I didn't want to link it because the log message would get more verbose than it already is.

That makes sense. I think it's fine.

pkg/providers/vsphere/vsphere.go
chrisdoherty4: @vivek-koppuru Do we have a ticket users can track? Its useful to include that in warnings. No worries if not.

chrisdoherty4 (approved):

pkg/providers/vsphere/vsphere.go
vivek-koppuru: We have this issue https://github.com/aws/eks-anywhere/issues/5517 but if we feel that is not the right one, we can create a different one. I am still leaning towards putting it in our documentation and changelog and not here because I don't want to make the message bigger, unless we think that is not a problem.

PR #5923, opened 2023-05-25
Improve CloudStack E2E template selection logic

panktishah26:
*Description of changes:*
For CloudStack e2e tests we get the template names from the secret value. When we change eks-d version, template names get out of sync which makes e2e test failure until we change the template name in the secret.

To avoid this, this PR includes a logic to select the right template from the test runner itself and by not depending on the secret value. It looks for template name in below order,

1. Look for explicit configuration through an env var: "T_CLOUDSTACK_TEMPLATE_{osFamily}_{eks-d version}" eg. T_CLOUDSTACK_TEMPLATE_REDHAT_KUBERNETES_1_23_EKS_22. This should be used for explicit configuration, mostly in local development for overrides.

2. If not present, look for a template if the default templates: "{eks-d version}-{osFamily}" eg. kubernetes-1-23-eks-22-redhat. This is what should be used most of the time in CI, the explicit configuration is not present but the right template has already been imported to cloudstack.

3. If the template doesn't exist, default to the value of the default template env vars: eg. "T_CLOUDSTACK_TEMPLATE_REDHAT_1_23". This is a catch all condition. Mostly for edge cases where the bundle has been updated with a new eks-d version, but the the new template hasn't been imported yet. It also preserves backwards compatibility.

### DIFF
--- a/pkg/executables/cmk.go
+++ b/pkg/executables/cmk.go
@@ -36,6 +36,10 @@ type Cmk struct {
 	configMap  map[string]decoder.CloudStackProfileConfig
 }
 
+type listTemplatesResponse struct {
+	CmkTemplates []cmkTemplate `json:"template"`
+}
+
 func (c *Cmk) Close(ctx context.Context) error {
 	return nil
 }
@@ -64,9 +68,7 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
 		return fmt.Errorf("template %s not found", template)
 	}
 
-	response := struct {
-		CmkTemplates []cmkTemplate `json:"template"`
-	}{}
+	response := listTemplatesResponse{}
 	if err = json.Unmarshal(result.Bytes(), &response); err != nil {
 		return fmt.Errorf("parsing response into json: %v", err)
 	}
@@ -77,6 +79,37 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
 		return fmt.Errorf("template %s not found", template)
 	}
 	return nil
+}
+
+// SearchTemplate looks for a template by name or by id and returns template name if found.
+func (c *Cmk) SearchTemplate(ctx context.Context, profile string, template v1alpha1.CloudStackResourceIdentifier) (string, error) {
+	command := newCmkCommand("list templates")
+	applyCmkArgs(&command, appendArgs("templatefilter=all"), appendArgs("listall=true"))
+	if len(template.Id) > 0 {
+		applyCmkArgs(&command, withCloudStackId(template.Id))
+	} else {
+		applyCmkArgs(&command, withCloudStackName(template.Name))
+	}
+
+	result, err := c.exec(ctx, profile, command...)
+	if err != nil {
+		return "", fmt.Errorf("getting templates info - %s: %v", result.String(), err)
+	}
+	if result.Len() == 0 {
+		return "", nil
+	}
+
+	response := listTemplatesResponse{}
+	if err = json.Unmarshal(result.Bytes(), &response); err != nil {
+		return "", fmt.Errorf("parsing response into json: %v", err)
+	}
+	templates := response.CmkTemplates
+	if len(templates) > 1 {
+		return "", fmt.Errorf("duplicate templates %s found", template)
+	} else if len(templates) == 0 {
+		return "", nil
+	}
+	return templates[0].Name, nil
 }
 
 func (c *Cmk) ValidateServiceOfferingPresent(ctx context.Context, profile string, zoneId string, serviceOffering v1alpha1.CloudStackResourceIdentifier) error {
--- a/test/framework/cloudstack.go
+++ b/test/framework/cloudstack.go
@@ -2,14 +2,20 @@ package framework
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/aws/eks-anywhere/internal/pkg/api"
 	"github.com/aws/eks-anywhere/internal/test/cleanup"
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/executables"
 	"github.com/aws/eks-anywhere/pkg/retrier"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 	clusterf "github.com/aws/eks-anywhere/test/framework/cluster"
 	"github.com/aws/eks-anywhere/test/framework/cluster/validations"
 )
@@ -71,12 +77,15 @@ var requiredCloudStackEnvVars = []string{
 }
 
 type CloudStack struct {
-	t              *testing.T
-	fillers        []api.CloudStackFiller
-	clusterFillers []api.ClusterFiller
-	cidr           string
-	podCidr        string
-	serviceCidr    string
+	t                 *testing.T
+	fillers           []api.CloudStackFiller
+	clusterFillers    []api.ClusterFiller
+	cidr              string
+	podCidr           string
+	serviceCidr       string
+	cmkClient         *executables.Cmk
+	devRelease        *releasev1.EksARelease
+	templatesRegistry *templateRegistry
 }
 
 type CloudStackOpt func(*CloudStack)
@@ -122,8 +131,10 @@ func CloudStackCredentialsAz1() string {
 
 func NewCloudStack(t *testing.T, opts ...CloudStackOpt) *CloudStack {
 	checkRequiredEnvVars(t, requiredCloudStackEnvVars)
+	cmk := buildCmk(t)
 	c := &CloudStack{
-		t: t,
+		t:         t,
+		cmkClient: cmk,
 		fillers: []api.CloudStackFiller{
 			api.RemoveCloudStackAzs(),
 			api.WithCloudStackAzFromEnvVars(cloudstackAccountVar, cloudstackDomainVar, cloudstackZoneVar, cloudstackCredentialsVar, cloudstackNetworkVar,
@@ -136,11 +147,10 @@ func NewCloudStack(t *testing.T, opts ...CloudStackOpt) *CloudStack {
 	c.cidr = os.Getenv(cloudStackCidrVar)
 	c.podCidr = os.Getenv(podCidrVar)
 	c.serviceCidr = os.Getenv(serviceCidrVar)
-
+	c.templatesRegistry = &templateRegistry{cache: map[string]string{}, generator: c}
 	for _, opt := range opts {
 		opt(c)
 	}
-
 	return c
 }
 
@@ -156,7 +166,7 @@ func WithCloudStackWorkerNodeGroup(name string, workerNodeGroup *WorkerNodeGroup
 func WithCloudStackRedhat123() CloudStackOpt {
 	return func(c *CloudStack) {
 		c.fillers = append(c.fillers,
-			api.WithCloudStackStringFromEnvVar(cloudstackTemplateRedhat123Var, api.WithCloudStackTemplateForAllMachines),
+			api.WithCloudStackTemplateForAllMachines(c.templateForDevRelease(anywherev1.RedHat, anywherev1.Kube123)),
 		)
 	}
 }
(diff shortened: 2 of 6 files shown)


pkg/executables/cmk.go
```diff
@@ -79,6 +79,39 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
...
+
+	result, err := c.exec(ctx, profile, command...)
+	if err != nil {
+		return "", fmt.Errorf("getting templates info - %s: %v", result.String(), err)
+	}
+	if result.Len() == 0 {
+		return "", fmt.Errorf("template %s not found", template)
+	}
+
+	response := struct {
```
g-gaston, line 100: i would vote to define this as package unexported struct

pkg/executables/cmk.go
```diff
@@ -79,6 +79,39 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
 	return nil
 }
 
+// SearchTemplatePresent - looks for template name or id and returns template name if available.
```
g-gaston, line 82:
```suggestion
// SearchTemplatePresent looks for a template by name or by id and returns template name if found.
```

pkg/executables/cmk.go
```diff
@@ -79,6 +79,39 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
...
+func (c *Cmk) SearchTemplatePresent(ctx context.Context, profile string, template v1alpha1.CloudStackResourceIdentifier) (string, error) {
```
g-gaston, line 83:
```suggestion
func (c *Cmk) SearchTemplate(ctx context.Context, profile string, template v1alpha1.CloudStackResourceIdentifier) (string, error) {
```

pkg/executables/cmk.go
```diff
@@ -79,6 +79,39 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
...
+	} else {
+		applyCmkArgs(&command, withCloudStackName(template.Name))
+	}
+
+	result, err := c.exec(ctx, profile, command...)
+	if err != nil {
+		return "", fmt.Errorf("getting templates info - %s: %v", result.String(), err)
+	}
+	if result.Len() == 0 {
+		return "", fmt.Errorf("template %s not found", template)
```
g-gaston, line 97: should this just return `"", nil`?
It's a good way of allowing the caller to distinguish between a not found result and an actual error.

pkg/executables/cmk.go
```diff
@@ -79,6 +79,39 @@ func (c *Cmk) ValidateTemplatePresent(ctx context.Context, profile string, domai
...
+		CmkTemplates []cmkTemplate `json:"template"`
+	}{}
+	if err = json.Unmarshal(result.Bytes(), &response); err != nil {
+		return "", fmt.Errorf("parsing response into json: %v", err)
+	}
+	templates := response.CmkTemplates
+	if len(templates) > 1 {
+		return "", fmt.Errorf("duplicate templates %s found", template)
+	} else if len(templates) == 0 {
+		return "", fmt.Errorf("template %s not found", template)
```
g-gaston, line 110: same here about not found

test/framework/cloudstack.go
```diff
@@ -77,6 +83,9 @@ type CloudStack struct {
 	cidr           string
 	podCidr        string
 	serviceCidr    string
+	CmkClient      *executables.Cmk
```
g-gaston, line 86: why is this exported? could we keep it in?

test/framework/cloudstack.go
```diff
@@ -339,3 +351,101 @@ func (c *CloudStack) WithRedhatVersion(version anywherev1.KubernetesVersion) api
...
+// This should be used for explicit configuration, mostly in local development for overrides.
+//
+// 2. If not present, look for a template if the default templates: "{eks-d version}-{osFamily}" eg. kubernetes-1-23-eks-22-redhat
+// This is what should be used most of the time in CI, the explicit configuration is not present but the right template has already been
+// imported to cloudstack.
+//
+// 3. If the template doesn't exist, default to the value of the default template env vars: eg. "T_CLOUDSTACK_TEMPLATE_REDHAT_1_23".
+// This is a catch all condition. Mostly for edge cases where the bundle has been updated with a new eks-d version, but the
+// the new template hasn't been imported yet. It also preserves backwards compatibility.
+func (c *CloudStack) templateForRelease(osFamily anywherev1.OSFamily, release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion) string {
```
g-gaston, line 398: this method seems like it's a 1 to 1 with the vsphere logic
What do you think about trying to move it to a shared piece of logic?

test/framework/executables.go
```diff
@@ -60,3 +61,26 @@ func buildHelm(t T) *executables.Helm {
 func buildSSH(t T) *executables.SSH {
 	return executables.NewLocalExecutablesBuilder().BuildSSHExecutable()
 }
+
+func buildCmk(t T) *executables.Cmk {
+	ctx := context.Background()
+	tmpWriter, err := filewriter.NewWriter("unique-ip")
```
g-gaston, line 67:
```suggestion
	tmpWriter, err := filewriter.NewWriter("cmk")
```

pkg/executables/cmk.go
```diff
@@ -36,6 +36,10 @@ type Cmk struct {
 	configMap  map[string]decoder.CloudStackProfileConfig
 }
 
+type response struct {
```
g-gaston, line 39:
```suggestion
type listTemplatesResponse struct {
```

test/framework/template.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"os"
+	"testing"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
+)
+
+// ProviderTemplateNameGenerator is an interface for getting template name for each provider.
+type ProviderTemplateNameGenerator interface {
```
g-gaston, line 13: why does this need to be exported? same question for all the methods
I think this is all internal to the e2e framework, it doesn't need to be exposed

test/framework/template.go
```diff
@@ -0,0 +1,89 @@
...
+	EnvVarForTemplate(osFamilyStr, eksDName string) string
+	DefaultNameForTemplate(osFamilyStr, eksDName string) string
+	SearchTemplate(ctx context.Context, template string) (string, error)
+	DefaultEnvVarForTemplate(osFamilyStr string, kubeVersion anywherev1.KubernetesVersion) string
+}
+
+// TemplateCache is a map of template name.
+type TemplateCache struct {
```
g-gaston, line 21: I think this can be unexported as well
also, what do you think about

```suggestion
type templateRegistry struct {
```

This does more things that just caching values, it's also able to retrieve them
The cache is just an internal implementation detail of the registry

test/framework/template.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// 2. If not present, look for a template if the default templates: "{eks-d version}-{osFamily}"
+// eg. kubernetes-1-23-eks-22-redhat (CloudStack),  /SDDC-Datacenter/vm/Templates/kubernetes-1-23-eks-22-redhat (vSphere)
+// This is what should be used most of the time in CI, the explicit configuration is not present but the right template has already been
+// imported to cloudstack.
+//
+// 3. If the template doesn't exist, default to the value of the default template env vars: eg. "T_CLOUDSTACK_TEMPLATE_REDHAT_1_23".
+// This is a catch all condition. Mostly for edge cases where the bundle has been updated with a new eks-d version, but the
+// the new template hasn't been imported yet. It also preserves backwards compatibility.
+func (tc *TemplateCache) TemplateForRelease(t *testing.T, generator ProviderTemplateNameGenerator, osFamily anywherev1.OSFamily, release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion) string {
```
g-gaston, line 40: I think we can have `generator` as a field in the `TemplateCache` to avoid having to pass it every time

test/framework/vsphere.go
```diff
@@ -713,34 +645,54 @@ func WithBottlerocketFromRelease(release *releasev1.EksARelease, kubeVersion any
 
 func (v *VSphere) WithBottleRocketForRelease(release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion) api.ClusterConfigFiller {
 	return api.VSphereToConfigFiller(
-		api.WithTemplateForAllMachines(v.templateForRelease(anywherev1.Bottlerocket, release, kubeVersion)),
+		api.WithTemplateForAllMachines(v.templateForDevRelease(anywherev1.Bottlerocket, kubeVersion)),
```
g-gaston, line 648: why are we changing this?

test/framework/vsphere.go
```diff
@@ -713,34 +645,54 @@ func WithBottlerocketFromRelease(release *releasev1.EksARelease, kubeVersion any
...
 	)
 }
 
 func optionToSetTemplateForRelease(osFamily anywherev1.OSFamily, release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion) VSphereOpt {
 	return func(v *VSphere) {
 		v.fillers = append(v.fillers,
-			api.WithTemplateForAllMachines(v.templateForRelease(osFamily, release, kubeVersion)),
+			api.WithTemplateForAllMachines(v.templateForDevRelease(osFamily, kubeVersion)),
```
g-gaston, line 655: same here, we changing this? this is supposed to take any release as an argument and now it's forcing to only use the dev release

test/framework/vsphere.go
```diff
@@ -713,34 +645,54 @@ func WithBottlerocketFromRelease(release *releasev1.EksARelease, kubeVersion any
...
 }
 
-func defaultNameForVSphereTemplate(templatesFolder, osFamily, eksDName string) string {
-	return filepath.Join(templatesFolder, fmt.Sprintf("%s-%s", strings.ToLower(eksDName), strings.ToLower(osFamily)))
+// DefaultNameForTemplate looks for a template with the name path: "{folder}/{eks-d version}-{osFamily}"
+// eg: /SDDC-Datacenter/vm/Templates/kubernetes-1-23-eks-22-redhat.
+func (v *VSphere) DefaultNameForTemplate(osFamily, eksDName string) string {
+	folder := v.testsConfig.TemplatesFolder
+	if folder == "" {
+		v.t.Log("vSphere templates folder is not configured, can't continue template search.")
```
g-gaston, line 671:
```suggestion
		v.t.Log("vSphere templates folder is not configured.")
```
IMO this method shouldn't assume where is being called from

test/framework/vsphere.go
panktishah26: My bad! I thought both the release would be same only. Changed it back.

test/framework/cloudstack.go
```diff
@@ -122,25 +131,27 @@ func CloudStackCredentialsAz1() string {
...
+		t:         t,
+		cmkClient: cmk,
 		fillers: []api.CloudStackFiller{
 			api.RemoveCloudStackAzs(),
 			api.WithCloudStackAzFromEnvVars(cloudstackAccountVar, cloudstackDomainVar, cloudstackZoneVar, cloudstackCredentialsVar, cloudstackNetworkVar,
 				cloudstackManagementServerVar, api.WithCloudStackAz),
 			api.WithCloudStackStringFromEnvVar(cloudstackSshAuthorizedKeyVar, api.WithCloudStackSSHAuthorizedKey),
 			api.WithCloudStackStringFromEnvVar(cloudstackComputeOfferingLargeVar, api.WithCloudStackComputeOfferingForAllMachines),
 		},
+		templatesRegistry: &templateRegistry{cache: map[string]string{}},
```
g-gaston, line 145: super nit: can we move this down so the initialization of the templatesRegistry happens all at once?

test/framework/vsphere.go
```diff
@@ -120,11 +120,11 @@ func NewVSphere(t *testing.T, opts ...VSphereOpt) *VSphere {
 			api.WithTLSInsecure(config.TLSInsecure),
 			api.WithTLSThumbprint(config.TLSThumbprint),
 		},
-		templatesCache: make(map[string]string),
+		templatesRegistry: &templateRegistry{cache: map[string]string{}},
```
g-gaston, line 123: same here about init

g-gaston (approved):

PR #5992, opened 2023-06-06
Add setting webhook data via env and cli flags:

jacobweinstock:
This enables the use of the Rufio webhook provider. The webhook provider needs a secret and a URL defined. All BMC machines will use this one secret.

As part of enabling the webhook secret to be captured by the CLI as an environment variable, all CLI flags are now accessible as env vars. The env var for each flag is the flag name but in all uppercase and the dash "-" symbol converted to an underscore "_".

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/hardware/translate.go
+++ b/pkg/providers/tinkerbell/hardware/translate.go
@@ -23,9 +23,9 @@ type MachineValidator interface {
 
 // TranslateAll reads entries 1 at a time from reader and writes them to writer. When reader returns io.EOF,
 // TranslateAll returns nil. Failure to return io.EOF from reader will result in an infinite loop.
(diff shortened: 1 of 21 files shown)


pkg/providers/tinkerbell/hardware/translate.go
```diff
@@ -48,6 +48,9 @@ func Translate(reader MachineReader, writer MachineWriter, validator MachineVali
 		return fmt.Errorf("read: invalid hardware: %v", err)
 	}
 
+	for _, modify := range modifiers {
+		machine = modify(machine)
+	}
```
chrisdoherty4, line 53: What do you think about decorating the `MachineWriter` instead of passing modifiers?

jacobweinstock:
Superseded by https://github.com/aws/eks-anywhere/pull/6741

PR #5997, opened 2023-06-07
Add design for kindless upgrades

g-gaston:

### DIFF
--- a/designs/kindless-upgrades.md
+++ b/designs/kindless-upgrades.md
@@ -1,0 +1,150 @@
+# Kindless upgrades
+
+## Introduction
+
+EKS-A uses a `kind` cluster to perform self-managed cluster (aka management cluster) upgrades. This has turned to be a potential source of issues when the upgrade process doesn't go smoothly, making the system fragile and difficult to recover from errors. Our goal is to remove it.
+
+Recent manual testing has proved that is indeed possible to upgrade a self-managed cluster from within itself by leveraging the EKS-A new full lifecycle controller. The rest of this docs outlines the necessary work to integrate this feature with the CLI and proposes initiatives and improvements to mitigate risks as well as increase the confidence in this feature before releasing it officially.
+
+
+>This is not a formal technical design doc, just an outline of necessary work. Each of these work items might or might not require a design doc of their own.
+
+### History
+
+When the `upgrade` operation was designed, it was decided that upgrading a cluster from within itself was too risky. So to mitigate this risk, a bootstrap cluster is created, using kind, and the management resources are moved to it. Then the `upgrade` operation is orchestrated from within the bootstrap cluster. Once it completes successfully, resources are moved back to the original cluster and the bootstrap cluster is deleted.
+
+The CAPI community advised against upgrading a self-managed cluster from within due to the risk of leaving the cluster in an unrecoverable state. The theory claims that if something goes very wrong with the upgrade and it leaves either the api server or etcd in a non working state, the CAPI controllers would be incapable of healing the cluster due to their dependency on the API server (and as an extension, etcd). This would leave the operator with very little room to maneuver since they would have lost access to the data stored in the CAPI resources as well as the ability to interact with its API. Given EKS-A's dependency on CAPI, the same concern applies to all EKS-A clusters.
+
+The community's concerns seemed solid and resonated with the team's understanding of the system. Given the complexity described this feature entails, the limited time to implement this operation before GA and the limited experience of the team with CAPI, the team lacked the resources to try to confidently implement self-managed upgrades by themselves. Hence, it was decided to follow the community's advice.
+
+## CLI integration
+
+This section describes the work required to implement this new architecture in the CLI `upgrade` command.
+
+```mermaid
+sequenceDiagram
+    autonumber
+    User->>+CLI: eksctl anywhere upgrade cluster
+    CLI->>+CLI: Validate
+    box Management CLuster
+    participant apiserver as kube-api server
+    participant controller as EKS-A cluster controller
+    participant infra as Infrastructure
+    end
+    CLI->>+apiserver: Apply cluster spec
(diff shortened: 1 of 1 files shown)


designs/kindless-upgrades.md
```diff
@@ -0,0 +1,150 @@
...
+
+    1. Update EKS-A resources with the user provided user spec using a Kubernetes client.
+    2. Watch/pull the cluster status, following the upgrade process. Beware of race conditions/caches and make sure we leverage `generation`/`observedGeneration` appropriately.
+    3. Blocking until the cluster is `Ready` or timeout is reached.
+
+* `reconcileClusterDefinitions` needs to be split and some parts reordered.
+    * Creating the new `Bundles` and EKS-D manifest in the cluster (in case of EKS-A version upgrade) needs to be moved up in the workflow, before core components upgrade.
+    * Updating the EKS-A resources is not needed, since the upgrade task already does it.
+    * All the GitOps operations should remain in this task.
+* Consider keeping a task to perform a backup of CAPI resources before applying any changes to the EKS-A cluster spec.
```
vivek-koppuru, line 86: Also during the upgrade flow in between the different phases?

designs/kindless-upgrades.md
```diff
@@ -0,0 +1,150 @@
...
+
+Most of the time when an upgrade fails, it shows up in the first etcd node not coming up. By making the etcd changes a separate step in the upgrade process, we reduce the amount of changes to rollback in case of an issue. If the first etcd node doesn't come up, the etcd cluster spec can be manually changed to its previous version and the new machine deleted, leaving the cluster in the same state as before.
+
+This will require changing the way the etcdadm cluster, core CAPI cluster and KCP controllers interact, or at least make the current behavior optional. Right now, the system is designed to apply all the CP changes at once, so it orchestrates itself making that assumption. If this we don't change this, it will cause a double CP upgrade, one after etcd finishes the upgrade (when it updated the endpoints secret) and a second one after updating the KCP.
```
vivek-koppuru, line 122: I am a little confused how this is avoided in the new way, unless you are saying it doesn't until we actually change the reconciliation logic.

designs/kindless-upgrades.md
```diff
@@ -0,0 +1,150 @@
...
+
+This implementation needs to extended to look at the KCP instead of the CAPI cluster, making sure `generation` and `observedGeneration` are leveraged to avoid trusting outdated statuses.
+
+This could make the reconciliation process take longer for control planes with a lot of nodes, since the controller would wait until all new nodes have joined the cluster and all the old ones have been removed. In order to speed things up, we might not need to wait until the whole CP is upgraded before moving on with the next phases. It might be enough with just waiting for the first control plane node to be upgraded (so first new machine created and first old machine deleted).
+
+This gives us good confidence the CP upgrade is going well. That said, since the next phase involves upgrading the CNI at that can affect networking, we recommend starting with extra caution and just wait for the whole CP to be upgraded (all new machines created and ready, all old machines deleted).
+
+That said, the "first node wait" strategy can still be a great alternative for workload clusters, given their upgrade process is less risky.
+
+#### Leaving `kube-vip` CP node for last
```
vivek-koppuru, line 138: Is kube-vip the only concern or is the scheduling of management components on the newer nodes also something to think about? I think the main issue is api server connectivity even in the management components so maybe kube-vip is the only thing.

designs/kindless-upgrades.md
g-gaston: That's a good idea, however, that would be a new feature, since the CLI doesn't do that today. That said, it's not trivial to implement, since the phases are controlled by the controller. The CLI would somehow need to hook into those.

My 2c here would be if we want to go after this, discuss the feature separately (even in parallel) from the effort to remove the kind cluster.

designs/kindless-upgrades.md
g-gaston: Ah good point, I might need to rephrase this, jackson had a similar comment. Maybe you can help me find where the whole is.

Right now, we don't need to solve this problem, it's already solved. The 3 controllers coordinate themselves using a combination of annotations and objects to avoid double CP rollouts. But, the current orchestration only works if changes to etcdamdcluster and KCP are applies at the same time.

So, if we decide to execute this idea I propose here (spliting etcd from KCP in the eks-a controller), we will lose this "double rollout" protection. That's why I mention "It will require changing the way..." that this orchestration works.

designs/kindless-upgrades.md
g-gaston: Ah interesting, allow me to dig in here. What management components are you referring to? Like the EKS-A and CAPI controllers?

Assuming it's EKS-A and CAPI controllers, I didn't consider that a concern, but I might be missing some scenarios. I was thinking that if the logic to "continue" with the reconciliation is only present in the controllers, the reconciliation won't progress unless they get successfully scheduled in a new node.

Or were you saying that we might want to leave the node running the controllers to be deleted the last? That might be a good idea but it can become tricky, since they can probably be running in different nodes. I'm not sure what the logic should be here but something to consider. Thoughts?

designs/kindless-upgrades.md
vivek-koppuru: Oh I see okay, the section makes it seem like we are mitigating risk, but for this specific one we are adding a new concern, so maybe calling it out as a side effect or another action item might help? Not sure if you want to add another section below for specific points that need to be figured out still.

designs/kindless-upgrades.md
g-gaston: Ah got you. Yeah I'm not concerned about transient errors due to webhooks restarting, the control loop should requeue those requests and retry until the webhooks are available.

designs/kindless-upgrades.md
g-gaston: I reworded this a bit, let me know if it helps.

vivek-koppuru (approved):

PR #6013, opened 2023-06-09
Add Cluster status conditions prototype implementation

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Currently, prototyping. So there are no unit test or e2e tests here at the moment.

*Testing (if applicable):*
Created a management and workload cluster. Monitored the statuses as they transition.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -30,8 +30,6 @@ import (
 	"github.com/aws/eks-anywhere/pkg/controller/clusters"
 	"github.com/aws/eks-anywhere/pkg/controller/handlers"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
-	"github.com/aws/eks-anywhere/pkg/networking/cilium"
-	ciliumreconciler "github.com/aws/eks-anywhere/pkg/networking/cilium/reconciler"
 	"github.com/aws/eks-anywhere/pkg/registrymirror"
 	"github.com/aws/eks-anywhere/pkg/utils/ptr"
 	"github.com/aws/eks-anywhere/release/api/v1alpha1"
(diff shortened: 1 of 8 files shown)


controllers/cluster_controller.go
```diff
@@ -21,28 +22,31 @@ import (
...
+	"github.com/aws/eks-anywhere/pkg/networking/cilium"
+	ciliumreconciler "github.com/aws/eks-anywhere/pkg/networking/cilium/reconciler"
 	"github.com/aws/eks-anywhere/pkg/registrymirror"
 	"github.com/aws/eks-anywhere/pkg/utils/ptr"
 	"github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
 const (
-	defaultRequeueTime = time.Minute
+	defaultRequeueTime = 10 * time.Second
```
cxbrowne1207, line 41: We'll need a shorter re-queue time, as one minute is too long to wait to get an update on the Cluster's condition.

controllers/cluster_controller.go
```diff
@@ -211,8 +219,43 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+
+		// Use the condition result if the last reconcile does not specify requeue or an interruption
+		if result.IsZero() && reterr == nil {
+			result = conditionResult
+		}
+
+		// Always update the readyCondition by summarizing the state of other conditions.
+		conditions.SetSummary(cluster,
+			conditions.WithConditions(
+				clusterv1.ControlPlaneReadyCondition,
```
mitalipaygude, line 235: is there a reason to look for capi ready here instead of eksa ready condition for eksa cluster?

pkg/controller/clusterapi.go
```diff
@@ -91,3 +92,76 @@ func GetMachineDeployment(ctx context.Context, client client.Client, machineDepl
...
 	return machineDeployment, nil
 }
+
+// GetMachines reads a list cluster-api Machines for an eks-a cluster using a kube client.
+func GetMachines(ctx context.Context, c client.Client, cluster *anywherev1.Cluster) ([]clusterv1.Machine, error) {
+	// "get", capiMachinesType, "-o", "json", "--kubeconfig", cluster.KubeconfigFile,
+	// "--selector=cluster.x-k8s.io/cluster-name=" + clusterName,
+	// "--namespace", constants.EksaSystemNamespace,
+	machines := &clusterv1.MachineList{}
+	err := c.List(ctx, machines, client.MatchingLabels{clusterv1.ClusterNameLabel: cluster.Name}, client.InNamespace(constants.EksaSystemNamespace))
```
mitalipaygude, line 102: Can we use the unauth client to list these machines?

Ref: https://github.com/aws/eks-anywhere/blob/main/pkg/clients/kubernetes/unauth.go#L66

controllers/cluster_controller.go
cxbrowne1207: it looks for the a condition with condition.Type == "ControlPlaneReady". The package does not matter here since the strings are the same. We are going to use our own types for conditions  instead for though

cxbrowne1207:
This was only a prototype, so closing

Issue #6016, opened 2023-06-12
exporting Snow node AMI to S3 fails when following the build guide


jespereneberg:
**What happened**:
[When following the docs to build a Snow node image](https://anywhere.eks.amazonaws.com/docs/osmgmt/artifacts/#build-snow-node-images) (step 5-7) you get the error when you want to export your newly built AMI to a raw image to S3:
`An error occurred (NotExportable) when calling the ExportImage operation: The image ID provided (ami-xzy123) is a marketplace AMI and is not exportable.`
This export needs to work so  you can later upload your image to your snow device.

**What you expected to happen**:
My AMI be exported as a RAW image to s3.

**How to reproduce it (as minimally and precisely as possible)**:
Follow step 1-7 in the following guide: https://anywhere.eks.amazonaws.com/docs/osmgmt/artifacts/#build-snow-node-images

**Anything else we need to know?**:
Solved by using the following Ubuntu AWS account ID: `099720109477`[1] instead of `679593333241` (AWS Marketplace Account ID) in `ami.json`. This make sure you don't use a marketplace AMI which is not allowed the be exported[2].

[1] https://ubuntu.com/server/docs/cloud-images/amazon-ec2
[2] https://docs.aws.amazon.com/vm-import/latest/userguide/vmexport_image.html#limits-image-export

**Environment**:
- EKS Anywhere Release: -
- EKS Distro Release: -

PR #6017, opened 2023-06-12
Update Snow AMI filter doc

jespereneberg:
Issue #, if available:
#6016

Description of changes:
Updated the owner `ami_filter_owners` value to `099720109477` which is the Ubuntu AWS account ID publishing the Ubuntu AMIs outside of Marketplace.

This solves the issue of building AMIs and exporting them to S3 as RAW images.


TerryHowe (commented):
/ok-to-test

Someone needs to explain why the accounts are different.

TerryHowe (approved):
/lgtm

PR #6005, opened 2023-06-08
Cluster Status design proposal doc

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Documentation for proposing a`Cluster` status related to performing management operations using the API.

*Testing (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #5628, opened 2023-04-13
Cluster Status


drewvanstone:
The Full Cluster Lifecycle API allows customers to manage (creating, updating and deleting) EKS-Anywhere clusters with a native Kubernetes experience we want users to have. However, it is unclear how to assess the state of cluster managed in this way, unlike the CLI where the state of any operation is made known through the run CLI run logs. 

- [x] Implement Conditions
  - [x] ControlPlaneReady
  - [x] DefaultCNIConfigured
  - [x] WorkersReady
  - [x] Ready
- [x] Add FailureReason field
<s>- [ ] Add DefaultCNI field</s>

PR #6020, opened 2023-06-12
Implement conditions in Cluster status

cxbrowne1207:
*Issue #, if available:*
#5628 

*Description of changes:*
This PR  implement the Cluster status as per the proposal [here](https://github.com/aws/eks-anywhere/pull/6005).

Changes to the cluster.status
- Added `Conditions`: `ControlPlaneInitialized`, `DefaultCNIConfigured`, `ControlPlaneReady`, `WorkersReady`, `Ready`
- Added `ObservedGeneration` field

*Testing (if applicable):*

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -385,10 +385,10 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
 
 func (r *ClusterReconciler) updateStatus(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
 	// An EKS-A Cluster managed by the CLI is deleted after the cluster finalizer is removed, leaving the CAPI cluster behind. In this case, we do not want to update
-	// the status. If we do, patching will through a 404 NotFound error when trying to update the status of the already deleted
+	// the status. If we do, patching will throw a 404 NotFound error when trying to update the status of the already deleted
 	// Cluster object.
 	if !cluster.DeletionTimestamp.IsZero() && metav1.HasAnnotation(cluster.ObjectMeta, anywherev1.ManagedByCLIAnnotation) {
-		log.Info("Clusters is managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
+		log.Info("Cluster managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
 		return nil
 	}
 
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -12,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -35,7 +37,7 @@ import (
 )
 
 const (
-	defaultRequeueTime = time.Minute
+	defaultRequeueTime = 10 * time.Second
 	// ClusterFinalizerName is the finalizer added to clusters to handle deletion.
 	ClusterFinalizerName = "clusters.anywhere.eks.amazonaws.com/finalizer"
 )
@@ -178,6 +180,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 }
 
 // Reconcile reconciles a cluster object.
+// nolint:gocyclo //TODO: Reduce high cycomatic complexity.
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=clusters;snowmachineconfigs;snowippools;vspheredatacenterconfigs;vspheremachineconfigs;dockerdatacenterconfigs;tinkerbellmachineconfigs;tinkerbelldatacenterconfigs;cloudstackdatacenterconfigs;cloudstackmachineconfigs;nutanixdatacenterconfigs;nutanixmachineconfigs;bundles;awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=oidcconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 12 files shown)


pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,187 @@
...
+	if conditions.IsTrue(cluster, anywherev1.ControlPlaneInitializedCondition) {
+		return nil
+	}
+
+	if kcp == nil {
+		conditions.Set(cluster, waitingForCPInitializedCondition())
+		return nil
+	}
+
+	con, err := checkKubeadmControlPlaneError(cluster, anywherev1.ControlPlaneInitializedCondition, kcp, controlplanev1.AvailableCondition)
```
vivek-koppuru, line 41: What is `con` here? I am assuming condition, but we should call it `kcpCondition` or something.

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,187 @@
...
+		conditions.Set(cluster, waitingForCPInitializedCondition())
+		return nil
+	}
+
+	conditions.MarkTrue(cluster, anywherev1.ControlPlaneInitializedCondition)
+	return nil
+}
+
+func waitingForCPInitializedCondition() *anywherev1.Condition {
+	return conditions.FalseCondition(anywherev1.ControlPlaneInitializedCondition, anywherev1.WaitingForControlPlaneInitializedReason, clusterv1.ConditionSeverityInfo, "The first control plane instance is not available yet")
```
vivek-koppuru, line 62: Can we have the string as a variable above or in the other file with the condition_types? Not sure if that is the right place, but it feels like it can be overlooked here? But don't feel super strongly on that preference.

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return err
+	}
+
+	if con != nil {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, con.Reason, con.Severity, con.Message)
+		return nil
+	}
+
+	// The control plane should be marked ready when the total number of nodes specified in the spec is
```
vivek-koppuru, line 95: Do we have to do this math ourselves? Does cluster-api not keep track of this already?

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,187 @@
...
+// and number of ready vs expected replicas are accounted for.
+func UpdateWorkersReadyCondition(cluster *anywherev1.Cluster, machineDeployments []clusterv1.MachineDeployment) error {
+	expected := 0
+	for _, wng := range cluster.Spec.WorkerNodeGroupConfigurations {
+		expected += *wng.Count
+	}
+
+	// The workers condition should be marked ready when the total number of nodes specified in the spec is
+	// equal to the ready amount in the cluster and there are no lingering old nodes. We need to check this in
+	// two parts, since in the case of a rolling upgrade, there can be ready machines in the cluster with the old spec.
```
vivek-koppuru, line 131: How do we check old spec? Based on observed generation field or something else?

vivek-koppuru: Or is `md.Status.UpdatedReplicas` enough for that?

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,187 @@
...
+	ciliumCfg := cluster.Spec.ClusterNetwork.CNIConfig.Cilium
+
+	// Though it may be installed initially to successfully create the cluster,
+	// if the CNI is configured to skip upgrades, we want to mark the condition as "False"
+	// as maintenance is not being performed by EKS-A.
+
+	// Otherwise, we can assume the CNI has been
+	// configured at this point after confirming the control plane is ready
+	if !ciliumCfg.IsManaged() {
+		conditions.MarkFalse(cluster, anywherev1.DefaultCNIConfiguredCondition, anywherev1.SkipUpgradesForDefaultCNIConfiguredReason, clusterv1.ConditionSeverityWarning, "Maintenance will not be performed during upgrades")
```
vivek-koppuru, line 183: What does maintenance not being performed mean? Can we just say something like "Skipping upgrade as CNI is not managed by EKS-A"

controllers/cluster_controller.go
```diff
@@ -357,6 +385,102 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+		return nil
+	}
+
+	log.Info("Updating cluster status")
+	defer func() {
+		// Always update the readyCondition by summarizing the state of other conditions.
+		conditions.SetSummary(cluster,
+			conditions.WithConditions(
+				anywherev1.ControlPlaneReadyCondition,
+				anywherev1.WorkersReadyConditon,
```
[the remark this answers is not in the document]
cxbrowne1207: No, we want the Ready condition to be a summary of only `ControlPlaneReady` and `WorkersReady` for now. `DefaultCNIConfigured` can be false if configured to skip upgrades, so that should not be an ultimate determinator of cluster readiness. We could add `ControlPlaneInitialized`, though it would be redundant because `ControlPlaneReady` already covers for it.

controllers/cluster_controller.go
```diff
@@ -357,6 +385,102 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	}
+
+	if err = clusters.UpdateWorkersReadyCondition(cluster, machineDeployments); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updateCNIStatus(log logr.Logger, cluster *anywherev1.Cluster) {
```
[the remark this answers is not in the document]
cxbrowne1207: It's also updating the status if it's a self managed cluster. So I think 'update' would be more appropriate?

pkg/controller/clusters/status.go
cxbrowne1207: Then `md.Status.UpdatedReplicas` helps us determine how many nodes currently have the old spec. It's the total number of non-temrinated machines that have the desired template spec [according to CAPI](https://pkg.go.dev/sigs.k8s.io/cluster-api/api/v1beta1#MachineDeploymentStatus).

But checking the observed generation helps us make sure those values on the status is reflecting the status of the current desired spec first.

controllers/cluster_controller.go
```diff
@@ -357,6 +385,102 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
 	return controller.Result{}, nil
 }
 
+func (r *ClusterReconciler) updateStatus(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
+	// An EKS-A Cluster managed by the CLI is deleted after the cluster finalizer is removed, leaving the CAPI cluster behind. In this case, we do not want to update
+	// the status. If we do, patching will through a 404 NotFound error when trying to update the status of the already deleted
+	// Cluster object.
+	if !cluster.DeletionTimestamp.IsZero() && metav1.HasAnnotation(cluster.ObjectMeta, anywherev1.ManagedByCLIAnnotation) {
+		log.Info("Clusters is managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
```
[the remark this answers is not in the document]
cxbrowne1207: I don't think this is an edge case where it isn't properly cleaned up. The reconciler currently removes the finalizer from the EKS-A Cluster and skips deleting the CAPI object if it has the annotation indicating that it is managed by the CLI. From what I understand, the CLI will handle cleaning it up.

If the EKS-A Cluster object is already deleted, then we can't update the status.

pkg/controller/clusters/status.go
cxbrowne1207: Chris was very against naming the condition ManagedCNI, so I was trying ro keep away from the "managed" verbiage. But yes, we can probably use something more specific

controllers/cluster_controller.go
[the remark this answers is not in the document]
vivek-koppuru: Could be redundant but maybe it's better to just include anyways if this is a summary?

cxbrowne1207:
This was split into multiple separate PRs and merged in. Closing.

PR #6022, opened 2023-06-14
Add design doc for Out-of-band solution:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This doc details new Rufio functionality for out of band communication.

*Testing (if applicable):*


designs/bare-metal-out-of-band-webhook.md
```diff
@@ -0,0 +1,549 @@
...
+* `webhook_provider_secrets`
+* `webhook_consumer_url`
+* `webhook_consumer_tls_cert`
+
+Currently BMC credentials are stored in a csv file and are per machine. The paradigm for the webhook provider is a bit different. There will not be a user/password combo per machine, instead all machines will use the secret(s) provided in `webhook_provider_secrets`.
+
+The webhook provider will require at a minimum 2 things from the customer.
+
+1. One or more HMAC secrets for use in signing the payload (HMAC).
+    1. These secrets must be provided as a comma separated environment variable. `webhook_provider_secrets="secretOne,secretTwo,secretThree,etc"`
```
chrisdoherty4, line 335: Do we expect the secrets to be base64 encoded so the user can use whatever characters they choose?

designs/bare-metal-out-of-band-webhook.md
```diff
@@ -0,0 +1,549 @@
...
+    2. A minimum complexity and length for each HMAC secret will be required and validated.
+2. A URL for where we will post webhook notifications.
+    1. This URL must be provided as an environment variable. `webhook_consumer_url=https://mywebhookconsumer/webhooks"`
+    2. We will do a preflight validation check for connectivity to this URL.
```
chrisdoherty4, line 339: How do we envisage this working? Would we need to require webhook providers to implement a readiness type endpoint possibly configured via a `webhook_consumer_readiness_url` environment variable.

designs/bare-metal-out-of-band-webhook.md
```diff
@@ -0,0 +1,549 @@
...
+          description: Successful operation
+components:
+  schemas:
+    BMC:
+      type: object
+      properties:
+        host:
+          type: string
+          example: "192.168.2.3"
+          description: The IP or hostname of a BMC
```
chrisdoherty4, line 189: Are we confident customers are content documenting address information of their BMCs like this? Perhaps opaque identifiers could be used instead (which makes them webhook provider implementation dependent so they could still use hostnames etc) with the webhook providers dealing with conversion how they see fit.

jacobweinstock: yeah, the data type is a string and we wont do any validation around it so a customer could definitely use an opaque identifier instead of an IP. I will make a note of this so that we get docs around this.

designs/bare-metal-out-of-band-webhook.md
jacobweinstock: Great point here, thanks. In order to satisfy the bmclib interfaces we will implement the `Open` method. This is where i envision we make a call to the `webhook_consumer_url` to see if it's reachable. I understand this is a runtime call mid BMC interaction and maybe in an ideal world this `webhook_consumer_url` would be checked as early as possible but this is not the model that bmclib has. Im open to discuss alternatives/options/etc.

designs/bare-metal-out-of-band-webhook.md
jacobweinstock: Good question. We will be hex encode to strings. I dont think there is any restrictions in terms of characters. Let me investigate some though.

jacobweinstock:
```Go
import "crypto/hmac"

s := "secret"
hmac.New(sha256.New, []byte(s))
```

chrisdoherty4: Sorry, I should've been clearer. I mean specifically for the environment variable where their secret may include a comma. Requiring a base64 encoding means stringing together the secrets is safe in that case.

chrisdoherty4 (approved):

designs/bare-metal-out-of-band-webhook.md
jacobweinstock: ah i see. thats a good idea. let me think on the affects.

Issue #5847, opened 2023-05-11
Better representing CLI version and dependencies in EKS-A cluster


jiayiwang7:
The relationship between an EKS-A release and its corresponding bundle is not clear. Furthermore, it is difficult to tell what version of EKS-A was used to create/upgrade a cluster. For these reasons, the bundlesRef field of a cluster should be deprecated in favor of an eksaVersion field. The eksaVersion field should map to an EKSARelease resource on the cluster, which can contain a reference of the release's corresponding bundle.

tatlat:
Implemented in https://github.com/aws/eks-anywhere/pull/6262 and https://github.com/aws/eks-anywhere/pull/6268

PR #6036, opened 2023-06-16
Create Instances of EKSARelease

tatlat:
*Issue #, if available:*

#5847 

*Description of changes:*
Build and create instances of EKSARelease on Cluster

*Testing (if applicable):*
Manual testing

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/controller/handlers"
 	"github.com/aws/eks-anywhere/pkg/curatedpackages"
 	"github.com/aws/eks-anywhere/pkg/registrymirror"
+	"github.com/aws/eks-anywhere/pkg/utils/ptr"
 	"github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
@@ -200,7 +201,10 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awssnowclusters;awssnowmachinetemplates;awssnowippools;vsphereclusters;vspheremachinetemplates;dockerclusters;dockermachinetemplates;tinkerbellclusters;tinkerbellmachinetemplates;cloudstackclusters;cloudstackmachinetemplates;nutanixclusters;nutanixmachinetemplates,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=packages.eks.amazonaws.com,resources=packages,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=packages.eks.amazonaws.com,namespace=eksa-system,resources=packagebundlecontrollers,verbs=delete
-// +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,namespace=eksa-system,resources=eksareleases,verbs=get;list;watch
+// +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=eksareleases,verbs=get;list;watch
+// The eksareleases permissions are being moved to the ClusterRole due to client trying to list this resource from cache.
+// When trying to list resources not already in cache, it starts an informer for that type using the scope of the cache.
+// So if the manager is cluster-scoped, the new informers created by the cache will be cluster-scoped
 
 // Reconcile reconciles a cluster object.
 // nolint:gocyclo
@@ -266,9 +270,9 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	// AddFinalizer	is idempotent
 	controllerutil.AddFinalizer(cluster, ClusterFinalizerName)
 
-	if cluster.Spec.BundlesRef == nil {
-		if err = r.setBundlesRef(ctx, cluster); err != nil {
-			return ctrl.Result{}, err
+	if !cluster.IsSelfManaged() && cluster.Spec.BundlesRef == nil && cluster.Spec.EksaVersion == nil {
+		if err = r.setDefaultBundlesRefOrEksaVersion(ctx, cluster); err != nil {
+			return ctrl.Result{}, nil
 		}
 	}
 
@@ -543,15 +547,35 @@ func aggregatedGeneration(config *cluster.Config) int64 {
 	return aggregatedGeneration
 }
 
-func (r *ClusterReconciler) setBundlesRef(ctx context.Context, clus *anywherev1.Cluster) error {
+func getManagementCluster(ctx context.Context, clus *anywherev1.Cluster, client client.Client) (*anywherev1.Cluster, error) {
 	mgmtCluster := &anywherev1.Cluster{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
+	if err := client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
 		if apierrors.IsNotFound(err) {
 			failureMessage := fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name)
 			clus.SetFailure(anywherev1.ManagementClusterRefInvalidReason, failureMessage)
 		}
+		return nil, err
+	}
+
+	return mgmtCluster, nil
+}
+
+func (r *ClusterReconciler) setDefaultBundlesRefOrEksaVersion(ctx context.Context, clus *anywherev1.Cluster) error {
+	mgmtCluster, err := getManagementCluster(ctx, clus, r.client)
+	if err != nil {
 		return err
 	}
-	clus.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
-	return nil
+
+	if mgmtCluster.Spec.EksaVersion != nil {
+		clus.Spec.EksaVersion = mgmtCluster.Spec.EksaVersion
+		return nil
+	}
+
+	if mgmtCluster.Spec.BundlesRef != nil {
+		clus.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
+		return nil
+	}
+
+	clus.Status.FailureMessage = ptr.String("Management cluster must have either EksaVersion or BundlesRef")
+	return fmt.Errorf("could not set default values")
 }
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -140,6 +140,7 @@ func newVsphereClusterReconcilerTest(t *testing.T, objs ...runtime.Object) *vsph
 func TestClusterReconcilerReconcileSelfManagedCluster(t *testing.T) {
 	g := NewWithT(t)
 	ctx := context.Background()
+	version := test.DevEksaVersion()
 
 	selfManagedCluster := &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
@@ -149,6 +150,7 @@ func TestClusterReconcilerReconcileSelfManagedCluster(t *testing.T) {
 			BundlesRef: &anywherev1.BundlesRef{
 				Name: "my-bundles-ref",
 			},
+			EksaVersion: &version,
 			ClusterNetwork: anywherev1.ClusterNetwork{
 				CNIConfig: &anywherev1.CNIConfig{
 					Cilium: &anywherev1.CiliumConfig{},
@@ -693,7 +695,8 @@ func TestClusterReconcilerReconcileGenerations(t *testing.T) {
 	for _, tt := range testCases {
 		t.Run(tt.testName, func(t *testing.T) {
 			config, bundles := baseTestVsphereCluster()
-
+			version := test.DevEksaVersion()
+			config.Cluster.Spec.EksaVersion = &version
 			config.Cluster.Generation = tt.clusterGeneration
 			config.Cluster.Status.ObservedGeneration = tt.clusterGeneration
 			config.Cluster.Status.ReconciledGeneration = tt.reconciledGeneration
@@ -771,6 +774,7 @@ func TestClusterReconcilerReconcileGenerations(t *testing.T) {
 func TestClusterReconcilerReconcileSelfManagedClusterWithExperimentalUpgrades(t *testing.T) {
 	g := NewWithT(t)
 	ctx := context.Background()
+	version := test.DevEksaVersion()
 
 	selfManagedCluster := &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
@@ -780,6 +784,7 @@ func TestClusterReconcilerReconcileSelfManagedClusterWithExperimentalUpgrades(t 
 			BundlesRef: &anywherev1.BundlesRef{
 				Name: "my-bundles-ref",
 			},
+			EksaVersion: &version,
 			ClusterNetwork: anywherev1.ClusterNetwork{
 				CNIConfig: &anywherev1.CNIConfig{
 					Cilium: &anywherev1.CiliumConfig{},
@@ -889,6 +894,7 @@ func TestClusterReconcilerReconcileDeletedSelfManagedCluster(t *testing.T) {
 func TestClusterReconcilerReconcileSelfManagedClusterRegAuthFailNoSecret(t *testing.T) {
 	g := NewWithT(t)
 	ctx := context.Background()
+	version := test.DevEksaVersion()
 
 	selfManagedCluster := &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
@@ -906,6 +912,7 @@ func TestClusterReconcilerReconcileSelfManagedClusterRegAuthFailNoSecret(t *test
 			RegistryMirrorConfiguration: &anywherev1.RegistryMirrorConfiguration{
 				Authenticate: true,
 			},
+			EksaVersion: &version,
 		},
 		Status: anywherev1.ClusterStatus{
 			ReconciledGeneration: 1,
@@ -1104,6 +1111,7 @@ func TestClusterReconcilerDeleteNoCAPIClusterSuccess(t *testing.T) {
 
 func TestClusterReconcilerSkipDontInstallPackagesOnSelfManaged(t *testing.T) {
 	ctx := context.Background()
+	version := test.DevEksaVersion()
 	cluster := &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "my-cluster",
@@ -1123,6 +1131,7 @@ func TestClusterReconcilerSkipDontInstallPackagesOnSelfManaged(t *testing.T) {
 			ManagementCluster: anywherev1.ManagementCluster{
 				Name: "",
 			},
+			EksaVersion: &version,
 		},
 		Status: anywherev1.ClusterStatus{
 			ReconciledGeneration: 1,
@@ -1243,6 +1252,7 @@ func TestClusterReconcilerPackagesDeletion(s *testing.T) {
 }
 
 func TestClusterReconcilerPackagesInstall(s *testing.T) {
+	version := test.DevEksaVersion()
 	newTestCluster := func() *anywherev1.Cluster {
 		return &anywherev1.Cluster{
 			ObjectMeta: metav1.ObjectMeta{
@@ -1263,6 +1273,7 @@ func TestClusterReconcilerPackagesInstall(s *testing.T) {
 				ManagementCluster: anywherev1.ManagementCluster{
 					Name: "my-management-cluster",
 				},
+				EksaVersion: &version,
 			},
 			Status: anywherev1.ClusterStatus{
 				ReconciledGeneration: 1,
@@ -1460,6 +1471,7 @@ func vsphereDataCenter(cluster *anywherev1.Cluster) *anywherev1.VSphereDatacente
 }
 
 func vsphereCluster() *anywherev1.Cluster {
+	version := test.DevEksaVersion()
 	return &anywherev1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      clusterName,
@@ -1497,6 +1509,7 @@ func vsphereCluster() *anywherev1.Cluster {
 					Labels: nil,
 				},
 			},
+			EksaVersion: &version,
 		},
 		Status: anywherev1.ClusterStatus{
 			ReconciledGeneration: 1,
--- a/internal/test/cluster.go
+++ b/internal/test/cluster.go
@@ -25,10 +25,13 @@ type ClusterSpecOpt func(*cluster.Spec)
 var configFS embed.FS
 
 // DevEksaVersion can be used in tests.
-var DevEksaVersion v1alpha1.EksaVersion = "v0.0.0-dev"
+func DevEksaVersion() v1alpha1.EksaVersion {
+	return v1alpha1.EksaVersion("v0.0.0-dev")
+}
 
 func NewClusterSpec(opts ...ClusterSpecOpt) *cluster.Spec {
 	s := &cluster.Spec{}
+	version := DevEksaVersion()
 	s.Config = &cluster.Config{
 		Cluster: &v1alpha1.Cluster{
 			ObjectMeta: metav1.ObjectMeta{
@@ -36,6 +39,7 @@ func NewClusterSpec(opts ...ClusterSpecOpt) *cluster.Spec {
 			},
 			Spec: v1alpha1.ClusterSpec{
 				WorkerNodeGroupConfigurations: []v1alpha1.WorkerNodeGroupConfiguration{{}},
+				EksaVersion:                   &version,
 			},
 		},
 	}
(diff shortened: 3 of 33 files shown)


pkg/api/v1alpha1/eksarelease_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// Reference to a bundle in the cluster
+	BundlesRef BundlesRef `json:"bundlesRef"`
+}
+
+// EKSAReleaseStatus defines the observed state of EKSARelease.
+type EKSAReleaseStatus struct{}
+
+// EKSAReleaseList contains a list of EKSARelease.
+// +kubebuilder:object:root=true
+type EKSAReleaseList struct {
```
g-gaston, line 44: what is this for?

pkg/api/v1alpha1/eksarelease_types.go
```diff
@@ -0,0 +1,52 @@
+package v1alpha1
```
g-gaston, line 1: shouldn't this be in the release package?

pkg/cluster/builder.go
```diff
@@ -117,3 +127,40 @@ func (b FileSpecBuilder) getBundles() (*releasev1.Bundles, error) {
...
 	return bundles.Read(b.reader, bundlesURL)
 }
+
+func (b FileSpecBuilder) getRelease(config *Config) (*releasev1.EksARelease, error) {
+	var opts []manifests.ReaderOpt
+	if b.releasesManifestURL != "" {
+		opts = append(opts, manifests.WithReleasesManifest(b.releasesManifestURL))
+	}
+	manifestReader := manifests.NewReader(b.reader, opts...)
+	return manifestReader.ReadReleaseForVersion(string(*config.Cluster.Spec.EksaVersion))
```
g-gaston, line 137: aren't we doing this twice now? one here and one to get the bundles

pkg/cluster/builder.go
```diff
@@ -117,3 +127,40 @@ func (b FileSpecBuilder) getBundles() (*releasev1.Bundles, error) {
...
+}
+
+func getEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles, cluster *v1alpha1.Cluster) *v1alpha1.EKSARelease {
+	if release == nil {
```
g-gaston, line 141: what case is this handling?

pkg/cluster/builder.go
```diff
@@ -117,3 +127,40 @@ func (b FileSpecBuilder) getBundles() (*releasev1.Bundles, error) {
...
+func (b FileSpecBuilder) getRelease(config *Config) (*releasev1.EksARelease, error) {
+	var opts []manifests.ReaderOpt
+	if b.releasesManifestURL != "" {
+		opts = append(opts, manifests.WithReleasesManifest(b.releasesManifestURL))
+	}
+	manifestReader := manifests.NewReader(b.reader, opts...)
+	return manifestReader.ReadReleaseForVersion(string(*config.Cluster.Spec.EksaVersion))
+}
+
+func getEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles, cluster *v1alpha1.Cluster) *v1alpha1.EKSARelease {
```
g-gaston, line 140: this method signature is a bit weird
what about something like `buildEKSARelease(release, bundles)

pkg/cluster/builder_test.go
```diff
@@ -74,7 +84,7 @@ func TestNewSpecWithBundlesOverrideValid(t *testing.T) {
 	v := version.Info{GitVersion: "v0.0.1"}
 	reader := files.NewReader()
 	b := cluster.NewFileSpecBuilder(reader, v,
-		cluster.WithReleasesManifest("testdata/invalid_release_version.yaml"),
+		// cluster.WithReleasesManifest("testdata/invalid_release_version.yaml"),
```
g-gaston, line 87: typo??

pkg/cluster/fetch.go
```diff
@@ -33,6 +34,20 @@ func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundles
 		return nil, err
 	}
 
+	eksaRelease := v1alpha1.EKSARelease{
```
g-gaston, line 37: why couldn't this method just fetch the real `EKSARelease` object?

pkg/cluster/fetch.go
```diff
@@ -212,6 +227,14 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
 // BuildSpecFromConfig constructs a cluster.Spec for an eks-a cluster config by retrieving all dependencies objects from the cluster using a kubernetes client.
 func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*Spec, error) {
 	bundlesName, bundlesNamespace := bundlesNamespacedKey(config.Cluster)
+	eksaRelease := &v1alpha1.EKSARelease{}
+	if config.Cluster.Spec.EksaVersion != nil {
```
g-gaston, line 231: we should throw an error if eksa version is nil when we get here

pkg/cluster/spec.go
```diff
@@ -227,3 +231,12 @@ func BundlesRefDefaulter() Defaulter {
 		return nil
 	}
 }
+
+// EksaVersionDefaulter sets the default EksaVersion in Cluster.
+func EksaVersionDefaulter() Defaulter {
```
g-gaston, line 236: is this being used?

pkg/clustermanager/cluster_manager.go
```diff
@@ -1145,6 +1148,20 @@ func (c *ClusterManager) ApplyBundles(ctx context.Context, clusterSpec *cluster.
 	return nil
 }
 
+// ApplyReleases applies the EKSARelease objects to the cluster.
+func (c *ClusterManager) ApplyReleases(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error {
```
g-gaston, line 1152: does this need to be exported?

pkg/clustermanager/cluster_manager_test.go
```diff
@@ -1865,7 +1865,7 @@ func TestClusterManagerCreateEKSAResourcesSuccess(t *testing.T) {
 	c, m := newClusterManager(t)
 
 	m.client.EXPECT().ApplyKubeSpecFromBytesForce(ctx, tt.cluster, gomock.Any())
-	m.client.EXPECT().ApplyKubeSpecFromBytes(ctx, tt.cluster, gomock.Any())
+	m.client.EXPECT().ApplyKubeSpecFromBytes(ctx, tt.cluster, gomock.Any()).MaxTimes(2)
```
g-gaston, line 1868: nit: add comment saying why is 2: what "things" we are applying

pkg/clustermanager/cluster_manager.go
tatlat: no

pkg/cluster/builder_test.go
tatlat: "invalid_release_version.yaml" does not exist so the test always fails
I've replaced it with "simple_release.yaml"

controllers/cluster_controller_test.go
```diff
@@ -392,6 +396,40 @@ func TestClusterReconcilerReconcileSelfManagedClusterRegAuthFailNoSecret(t *test
 	g.Expect(err).To(MatchError(ContainSubstring("fetching registry auth secret")))
 }
 
+func TestClusterReconcilerReconcileSelfManagedClusterFailNoEksaVersion(t *testing.T) {
```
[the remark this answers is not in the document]
tatlat: Not sure if it's what you were looking for but I added TestClusterReconcilerSetDefaultEksaVersion in cluster_controller_test_test.go

controllers/cluster_controller.go
```diff
@@ -554,3 +560,15 @@ func (r *ClusterReconciler) setBundlesRef(ctx context.Context, clus *anywherev1.
 	clus.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
 	return nil
 }
+
+func (r *ClusterReconciler) setDefaultEksaVersion(ctx context.Context, clus *anywherev1.Cluster) error {
+	mgmtCluster := &anywherev1.Cluster{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
```
vivek-koppuru, line 566: It looks like we can refactor this call out to make it easier not have to replicate this business logic and instead just call a function and have it return as expected, like `ValidateManagementClusterExists`

pkg/cluster/builder.go
```diff
@@ -92,7 +100,18 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
...
 	}
 
-	return NewSpec(config, bundlesManifest, eksd)
+	release, err := mReader.ReadReleaseForVersion(b.cliVersion.GitVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	if release == nil {
+		return nil, fmt.Errorf("EKS-A version is not a valid release")
```
vivek-koppuru, line 109: We should probably say `EksaVersion is not a valid release` right?

pkg/cluster/fetch.go
```diff
@@ -27,29 +30,15 @@ type AWSIamConfigFetch func(ctx context.Context, name, namespace string) (*v1alp
 // BuildSpecForCluster constructs a cluster.Spec for an eks-a cluster by retrieving all
 // necessary objects using fetch methods
 // This is deprecated in favour of BuildSpec.
-func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, fluxConfigFetch FluxConfigFetch, oidcFetch OIDCFetch, awsIamConfigFetch AWSIamConfigFetch) (*Spec, error) {
-	bundles, err := GetBundlesForCluster(ctx, cluster, bundlesFetch)
+func BuildSpecForCluster(ctx context.Context, cluster *v1alpha1.Cluster, bundlesFetch BundlesFetch, eksdReleaseFetch EksdReleaseFetch, gitOpsFetch GitOpsFetch, fluxConfigFetch FluxConfigFetch, oidcFetch OIDCFetch, awsIamConfigFetch AWSIamConfigFetch, eksaReleaseFetch EKSAReleaseFetch) (*Spec, error) {
```
vivek-koppuru, line 33: Instead of adding a new param, should we pass in a struct called `Fetchers` that sets all of these?

pkg/cluster/fetch.go
```diff
@@ -112,6 +139,20 @@ func bundlesNamespacedKey(cluster *v1alpha1.Cluster) (name, namespace string) {
...
 
+func getEKSAReleaseForCluster(ctx context.Context, cluster *v1alpha1.Cluster, fetch EKSAReleaseFetch) (*v1alpha1release.EKSARelease, error) {
+	if cluster.Spec.EksaVersion == nil {
+		return nil, fmt.Errorf("Cluster's EksaVersion is nil")
+	}
+	version := string(*cluster.Spec.EksaVersion)
+	name := v1alpha1release.GenerateEKSAReleaseName(version)
+	eksaRelease, err := fetch(ctx, name, constants.EksaSystemNamespace)
+	if err != nil {
+		return nil, fmt.Errorf("fetching EKSARelease for cluster: %v", err)
```
vivek-koppuru, line 150: Do we have a different EKSARelease and EksaRelease still?

pkg/executables/kubectl.go
```diff
@@ -1885,6 +1886,23 @@ func (k *Kubectl) GetBundles(ctx context.Context, kubeconfigFile, name, namespac
 	return response, nil
 }
 
+// GetEKSARelease gets an EKSARelease resource from a cluster.
+func (k *Kubectl) GetEKSARelease(ctx context.Context, kubeconfigFile, name, namespace string) (*releasev1alpha1.EKSARelease, error) {
```
g-gaston, line 1890: sorry I think i missed this
Where do we use it?

pkg/cluster/builder.go
```diff
@@ -72,6 +78,8 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 		return nil, err
 	}
 	configManager.RegisterDefaulters(BundlesRefDefaulter())
+	v := v1alpha1.EksaVersion(b.cliVersion.GitVersion)
+	config.Cluster.Spec.EksaVersion = &v
```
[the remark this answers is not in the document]
tatlat: b.cliVersion.GitVersion returns a string
v1alpha1.EksaVersion casts it to an EksaVersion

pkg/cluster/fetch.go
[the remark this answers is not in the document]
tatlat: It is still in use in cluster_manager

controllers/cluster_controller_test.go
```diff
@@ -896,6 +900,40 @@ func TestClusterReconcilerReconcileSelfManagedClusterRegAuthFailNoSecret(t *test
...
+	controller := gomock.NewController(t)
+	providerReconciler := mocks.NewMockProviderClusterReconciler(controller)
+	iam := mocks.NewMockAWSIamConfigReconciler(controller)
+	clusterValidator := mocks.NewMockClusterValidator(controller)
+	registry := newRegistryMock(providerReconciler)
+	c := fake.NewClientBuilder().WithRuntimeObjects(selfManagedCluster).Build()
+
+	r := controllers.NewClusterReconciler(c, registry, iam, clusterValidator, nil)
+	_, err := r.Reconcile(ctx, clusterRequest(selfManagedCluster))
+	g.Expect(err).To(MatchError(ContainSubstring("not found")))
```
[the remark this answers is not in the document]
tatlat: I made a mistake here. Changed the test and expected error string to better reflect what happens.

pkg/cluster/fetch.go
tatlat: yes

pkg/cluster/builder.go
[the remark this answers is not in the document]
tatlat: consider it done

pkg/cluster/fetch.go
```diff
@@ -212,6 +253,17 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
 // BuildSpecFromConfig constructs a cluster.Spec for an eks-a cluster config by retrieving all dependencies objects from the cluster using a kubernetes client.
 func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*Spec, error) {
 	bundlesName, bundlesNamespace := bundlesNamespacedKey(config.Cluster)
```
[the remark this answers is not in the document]
tatlat: Yes we need them. I should only override those values if the cluster's bundlesRef isn't being used.

pkg/cluster/fetch.go
```diff
@@ -212,6 +253,17 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
...
+	eksaRelease := &v1alpha1release.EKSARelease{}
+	if config.Cluster.Spec.EksaVersion == nil {
+		return nil, fmt.Errorf("Cluster's EksaVersion is nil")
+	}
+	version := string(*config.Cluster.Spec.EksaVersion)
```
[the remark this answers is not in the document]
tatlat: GenerateEKSAReleaseName takes a string and does string manipulation. I can't have it take v1alpha1.EksaVersion because that would cause a circular dependency.

pkg/executables/kubectl.go
tatlat: cluster_manager uses it

controllers/cluster_controller_test.go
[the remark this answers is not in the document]
cxbrowne1207: Ah, my bad. I misread this to be testing the validation in `setDefaultEksaVersion` where the "not found" error is being set as the `FailureMessage` and also being returned.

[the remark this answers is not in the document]
tatlat: added a check

pkg/cluster/builder.go
```diff
@@ -104,16 +123,44 @@ func (b FileSpecBuilder) getConfig(clusterConfigURL string) (*Config, error) {
...
-			opts = append(opts, manifests.WithReleasesManifest(b.releasesManifestURL))
-		}
-		manifestReader := manifests.NewReader(b.reader, opts...)
 		return manifestReader.ReadBundlesForVersion(b.cliVersion.GitVersion)
 	}
 
 	return bundles.Read(b.reader, bundlesURL)
 }
+
+func (b FileSpecBuilder) buildEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles) *releasev1.EKSARelease {
```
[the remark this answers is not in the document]
tatlat: i changed it to a function

pkg/cluster/builder.go
```diff
@@ -104,16 +123,44 @@ func (b FileSpecBuilder) getConfig(clusterConfigURL string) (*Config, error) {
...
+	eksaRelease := &releasev1.EKSARelease{
+		TypeMeta: v1.TypeMeta{
+			Kind:       releasev1.EKSAReleaseKind,
+			APIVersion: v1alpha1.SchemeBuilder.GroupVersion.String(),
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Name:      releasev1.GenerateEKSAReleaseName(b.cliVersion.GitVersion),
```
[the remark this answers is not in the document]
tatlat: I changed it to do that but the release struct contains build info in its Version so I had to do some string manipulation before generating the eksarelease name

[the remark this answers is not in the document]
g-gaston: Why can't we keep the build info? it's part of the release version

pkg/cluster/fetch.go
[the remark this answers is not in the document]
g-gaston: can we just use `BuildSpec` in cluster manager?

controllers/cluster_controller.go
```diff
@@ -267,12 +270,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
...
 		if err = r.setBundlesRef(ctx, cluster); err != nil {
 			return ctrl.Result{}, err
 		}
 	}
 
+	if cluster.Spec.EksaVersion == nil && cluster.Spec.BundlesRef == nil {
+		if err = r.setDefaultEksaVersion(ctx, cluster); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
g-gaston, line 283: I think the way this logic operates is wrong
Also we might benefit from combining these two methods so the code looks more coherent
Let's sync offline, it will be easier

controllers/cluster_controller.go
```diff
@@ -544,13 +553,38 @@ func aggregatedGeneration(config *cluster.Config) int64 {
...
+	if mgmtCluster.Spec.EksaVersion == nil {
+		clus.Status.FailureMessage = ptr.String("Management cluster cannot have nil EksaVersion")
+		return fmt.Errorf("mgmt cluster cannot have nil EksaVersion")
+	}
+
+	clus.Spec.EksaVersion = mgmtCluster.Spec.EksaVersion
+	return nil
+}
+
+func validateManagementClusterExists(ctx context.Context, clus *anywherev1.Cluster, client client.Client) (*anywherev1.Cluster, error) {
```
g-gaston, line 580: why is a validate method returning an object?
This signature is a bit confusing

I think we should clean this up a bit

internal/test/cluster.go
```diff
@@ -25,17 +25,22 @@ type ClusterSpecOpt func(*cluster.Spec)
 var configFS embed.FS
 
 // DevEksaVersion can be used in tests.
-var DevEksaVersion v1alpha1.EksaVersion = "v0.0.0-dev"
+func DevEksaVersion() v1alpha1.EksaVersion {
+	version := v1alpha1.EksaVersion("v0.0.0-dev")
+	return version
```
g-gaston, line 30: nit:
```suggestion
	return v1alpha1.EksaVersion("v0.0.0-dev")
```

pkg/cluster/builder.go
```diff
@@ -104,16 +119,45 @@ func (b FileSpecBuilder) getConfig(clusterConfigURL string) (*Config, error) {
...
-		}
-		manifestReader := manifests.NewReader(b.reader, opts...)
 		return manifestReader.ReadBundlesForVersion(b.cliVersion.GitVersion)
 	}
 
 	return bundles.Read(b.reader, bundlesURL)
 }
+
+func buildEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles) *releasev1.EKSARelease {
+	version := strings.Split(release.Version, "+")[0]
```
g-gaston, line 140: is it possible to include the full semver in the name?

pkg/cluster/fetch.go
```diff
@@ -211,7 +262,21 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
 
 // BuildSpecFromConfig constructs a cluster.Spec for an eks-a cluster config by retrieving all dependencies objects from the cluster using a kubernetes client.
 func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*Spec, error) {
-	bundlesName, bundlesNamespace := bundlesNamespacedKey(config.Cluster)
+	eksaRelease := &v1alpha1release.EKSARelease{}
+	if config.Cluster.Spec.BundlesRef == nil {
+		if config.Cluster.Spec.EksaVersion == nil {
+			return nil, fmt.Errorf("cluster's EksaVersion cannot be nil")
```
g-gaston, line 268:
```suggestion
			return nil, fmt.Errorf("either cluster's EksaVersion or BundlesRef need to be set")
```

pkg/cluster/fetch.go
```diff
@@ -211,7 +262,21 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
...
+		version := string(*config.Cluster.Spec.EksaVersion)
+		eksaReleaseName := v1alpha1release.GenerateEKSAReleaseName(version)
+		if err := client.Get(ctx, eksaReleaseName, constants.EksaSystemNamespace, eksaRelease); err != nil {
+			return nil, fmt.Errorf("error getting EKSARelease %s", eksaReleaseName)
+		}
+	} else {
+		eksaRelease = nil
+	}
+
+	bundlesName, bundlesNamespace := bundlesNamespacedKey(config.Cluster, eksaRelease)
```
g-gaston, line 279: how does this work if the eksaRelease is nil?

pkg/clustermanager/cluster_manager.go
```diff
@@ -1121,6 +1125,20 @@ func (c *ClusterManager) ApplyBundles(ctx context.Context, clusterSpec *cluster.
 	return nil
 }
 
+// ApplyReleases applies the EKSARelease manifest.
+func (c *ClusterManager) ApplyReleases(ctx context.Context, clusterSpec *cluster.Spec, cluster *types.Cluster) error {
```
g-gaston, line 1129: does this need to be exported?

tatlat: yes, it's needed in the upgrade workflow

pkg/cluster/builder.go
tatlat: yes, changed it

controllers/cluster_controller.go
tatlat: i fixed the method's name

pkg/cluster/fetch.go
tatlat: eksaRelease isn't needed in bundlesNamespacedKey, but I did remove setting it to nil
bundlesNamespacedKey will fall back to the cluster's BundlesRef if the eksaRelease's BundlesRef is empty

release/api/v1alpha1/eksarelease.go
```diff
@@ -10,5 +10,6 @@ const EKSAReleaseKind = "EKSARelease"
 
 // Generates the naming convention of EKSARelease from a version.
 func GenerateEKSAReleaseName(version string) string {
+	version = strings.ReplaceAll(version, "+", "-")
 	return fmt.Sprintf("eksa-%s", strings.ReplaceAll(version, ".", "-"))
```
g-gaston, line 14: is there a way to reverse this transformation?
i wonder if we should use a diff algorithm so we can go back and forth
let me know if you don't think this is doable

controllers/cluster_controller.go
```diff
@@ -552,39 +546,34 @@ func aggregatedGeneration(config *cluster.Config) int64 {
...
-	if mgmtCluster.Spec.EksaVersion == nil {
-		clus.Status.FailureMessage = ptr.String("Management cluster cannot have nil EksaVersion")
-		return fmt.Errorf("mgmt cluster cannot have nil EksaVersion")
+	if mgmtCluster.Spec.BundlesRef != nil {
+		clus.Spec.BundlesRef = mgmtCluster.Spec.BundlesRef
+		return nil
 	}
 
-	clus.Spec.EksaVersion = mgmtCluster.Spec.EksaVersion
-	return nil
```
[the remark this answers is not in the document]
tatlat: We already do that

release/api/v1alpha1/eksarelease.go
tatlat: There should only ever be a "+" in front of build so to reverse this transformation, we could just turn the last "." into a "+".

g-gaston: Semver can also have a "-" after the three numbers
What do you think about replacing "+" with "-plus-"?

g-gaston (commented):
/lgtm

PR #6041, opened 2023-06-16
Add observedGeneration and ControlPlaneInitialized condition to ClusterStatus

cxbrowne1207:
*Issue #, if available:*
#5628 

*Description of changes:*
Changes to the Cluster status as per the [proposal](https://github.com/aws/eks-anywhere/pull/6005). Added:
- Added ObservedGeneration field
- ControlPlaneInitialized condition

*Testing (if applicable):*
- Unit testing
- Manual testing

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -12,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -35,7 +37,7 @@ import (
 )
 
 const (
-	defaultRequeueTime = time.Minute
+	defaultRequeueTime = 10 * time.Second
 	// ClusterFinalizerName is the finalizer added to clusters to handle deletion.
 	ClusterFinalizerName = "clusters.anywhere.eks.amazonaws.com/finalizer"
 )
@@ -178,6 +180,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 }
 
 // Reconcile reconciles a cluster object.
+// nolint:gocyclo //TODO: Reduce high cycomatic complexity.
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=clusters;snowmachineconfigs;snowippools;vspheredatacenterconfigs;vspheremachineconfigs;dockerdatacenterconfigs;tinkerbellmachineconfigs;tinkerbelldatacenterconfigs;cloudstackdatacenterconfigs;cloudstackmachineconfigs;nutanixdatacenterconfigs;nutanixmachineconfigs;bundles;awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=oidcconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
@@ -192,7 +195,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 // +kubebuilder:rbac:groups="",namespace=eksa-system,resources=secrets,verbs=delete;
 // +kubebuilder:rbac:groups=tinkerbell.org,resources=hardware;hardware/status,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=bmc.tinkerbell.org,resources=machines;machines/status,verbs=get;list;watch
-func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, reterr error) {
 	log := ctrl.LoggerFrom(ctx)
 	// Fetch the Cluster object
 	cluster := &anywherev1.Cluster{}
@@ -211,9 +214,32 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ 
 	}
 
 	defer func() {
-		// Always attempt to patch the object and status after each reconciliation.
-		if err := patchHelper.Patch(ctx, cluster); err != nil {
+		err := r.updateStatus(ctx, log, cluster)
+		if err != nil {
 			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+
+		args := make([]string, 0, len(cluster.Status.Conditions))
+		for _, condition := range cluster.Status.Conditions {
+			args = append(args, fmt.Sprintf("(%s=%s %s)", condition.Type, condition.Status, condition.Reason))
+		}
+
+		log.Info("Current conditions", "conditions", strings.Join(args, ", "))
+
+		// Always attempt to patch the object and status after each reconciliation.
+		patchOpts := []patch.Option{}
+
+		// Patch ObservedGeneration only if the reconciliation completed without error
+		if reterr == nil {
+			patchOpts = append(patchOpts, patch.WithStatusObservedGeneration{})
+		}
+		if err := patchCluster(ctx, patchHelper, cluster, patchOpts...); err != nil {
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+
+		// Only requeue if we are not already re-queueing and the Cluster ready condition is false
+		if reterr == nil && !result.Requeue && result.RequeueAfter <= 0 && conditions.IsFalse(cluster, anywherev1.ReadyCondition) {
+			result = ctrl.Result{RequeueAfter: defaultRequeueTime}
 		}
 	}()
 
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
 	return controller.Result{}, nil
 }
 
+func (r *ClusterReconciler) updateStatus(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
+	// An EKS-A Cluster managed by the CLI is deleted after the cluster finalizer is removed, leaving the CAPI cluster behind. In this case, we do not want to update
+	// the status. If we do, patching will throw a 404 NotFound error when trying to update the status of the already deleted
+	// Cluster object.
+	if !cluster.DeletionTimestamp.IsZero() && metav1.HasAnnotation(cluster.ObjectMeta, anywherev1.ManagedByCLIAnnotation) {
+		log.Info("Cluster managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
+		return nil
+	}
+
+	// ObservedGeneration represents the last observed generation, and consequently represents that the current status is up to date.
+	// Then if observedGeneration is equal to generation AND the Cluster's "Ready" condition is "True", then we can skip another status update
+	if cluster.Status.ObservedGeneration == cluster.Generation && conditions.IsTrue(cluster, anywherev1.ReadyCondition) {
+		log.Info("Generation matches observedGeneration and the cluster is ready, skipping status update")
+		return nil
+	}
+
+	log.Info("Updating cluster status")
+	defer func() {
+		// Always update the readyCondition by summarizing the state of other conditions.
+		conditions.SetSummary(cluster,
+			conditions.WithConditions(
+				anywherev1.ControlPlaneReadyCondition,
+				anywherev1.WorkersReadyConditon,
+			),
+		)
+	}()
+
+	if err := r.updateControlPlaneStatus(ctx, log, cluster); err != nil {
+		return errors.Wrap(err, "updating controlplane status")
+	}
+
+	return nil
+}
+
+func (r *ClusterReconciler) updateControlPlaneStatus(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
+	log.Info("Updating control plane status")
+	kcp, err := controller.GetKubeadmControlPlane(ctx, r.client, cluster)
+	if err != nil {
+		return errors.Wrapf(err, "getting kubeadmcontrolplane")
+	}
+
+	if err = clusters.UpdateControlPlaneInitializedCondition(cluster, kcp); err != nil {
+		return errors.Wrap(err, "updating control plane initialized condition")
+	}
+
+	return nil
+}
+
 func (r *ClusterReconciler) reconcileDelete(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (ctrl.Result, error) {
 	if cluster.IsSelfManaged() {
 		return ctrl.Result{}, errors.New("deleting self-managed clusters is not supported")
@@ -443,6 +517,23 @@ func (r *ClusterReconciler) ensureClusterOwnerReferences(ctx context.Context, cl
 	}
 
 	return nil
+}
+
+func patchCluster(ctx context.Context, patchHelper *patch.Helper, cluster *anywherev1.Cluster, patchOpts ...patch.Option) error {
+	// Patch the object, ignoring conflicts on the conditions owned by this controller.
+	options := append([]patch.Option{
+		patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
+			// Add each condition her that the controller should ignored conflicts for.
+			anywherev1.ReadyCondition,
+			anywherev1.ControlPlaneInitializedCondition,
+			anywherev1.ControlPlaneReadyCondition,
+			anywherev1.WorkersReadyConditon,
+			anywherev1.DefaultCNIConfiguredCondition,
+		}},
+	}, patchOpts...)
+
+	// Always attempt to patch the object and status after each reconciliation.
+	return patchHelper.Patch(ctx, cluster, options...)
 }
 
 // aggregatedGeneration computes the combined generation of the resources linked
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -3,7 +3,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -37,7 +36,7 @@ import (
 )
 
 const (
-	defaultRequeueTime = 10 * time.Second
+	defaultRequeueTime = time.Minute
 	// ClusterFinalizerName is the finalizer added to clusters to handle deletion.
 	ClusterFinalizerName = "clusters.anywhere.eks.amazonaws.com/finalizer"
 )
@@ -180,7 +179,8 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) 
 }
 
 // Reconcile reconciles a cluster object.
-// nolint:gocyclo //TODO: Reduce high cycomatic complexity.
+// nolint:gocyclo
+// TODO: Reduce high cycomatic complexity. https://github.com/aws/eks-anywhere-internal/issues/1449
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=clusters;snowmachineconfigs;snowippools;vspheredatacenterconfigs;vspheremachineconfigs;dockerdatacenterconfigs;tinkerbellmachineconfigs;tinkerbelldatacenterconfigs;cloudstackdatacenterconfigs;cloudstackmachineconfigs;nutanixdatacenterconfigs;nutanixmachineconfigs;bundles;awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=oidcconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=awsiamconfigs,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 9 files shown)


controllers/cluster_controller.go
```diff
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
 	return controller.Result{}, nil
 }
 
+func (r *ClusterReconciler) updateStatus(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
+	// An EKS-A Cluster managed by the CLI is deleted after the cluster finalizer is removed, leaving the CAPI cluster behind. In this case, we do not want to update
+	// the status. If we do, patching will throw a 404 NotFound error when trying to update the status of the already deleted
+	// Cluster object.
+	if !cluster.DeletionTimestamp.IsZero() && metav1.HasAnnotation(cluster.ObjectMeta, anywherev1.ManagedByCLIAnnotation) {
+		log.Info("Cluster managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
```
vivek-koppuru, line 391: If cluster is managed by CLI, why do we want to update status here? Are you trying to say Cluster managed by EKS-A, or what about the CLI makes it so that CAPI cluster remains and we want to update cluster status?

controllers/cluster_controller.go
```diff
@@ -35,7 +37,7 @@ import (
 )
 
 const (
-	defaultRequeueTime = time.Minute
+	defaultRequeueTime = 10 * time.Second
```
g-gaston, line 40: can we not change this? just have a different const for your new usecase

controllers/cluster_controller.go
```diff
@@ -178,6 +180,7 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger)
 }
 
 // Reconcile reconciles a cluster object.
+// nolint:gocyclo //TODO: Reduce high cycomatic complexity.
```
g-gaston, line 183: we might want a ticket for this one
or else we'll never do it

controllers/cluster_controller.go
```diff
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+		log.Info("Generation matches observedGeneration and the cluster is ready, skipping status update")
+		return nil
+	}
+
+	log.Info("Updating cluster status")
+	defer func() {
+		// Always update the readyCondition by summarizing the state of other conditions.
+		conditions.SetSummary(cluster,
+			conditions.WithConditions(
+				anywherev1.ControlPlaneReadyCondition,
```
vivek-koppuru, line 407: We should consider adding the Initialized one mainly because that tells us if we can retrieve kubeconfig which is an important status for Cluster.

controllers/cluster_controller.go
```diff
@@ -211,10 +214,33 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+		if err != nil {
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+
+		args := make([]string, 0, len(cluster.Status.Conditions))
+		for _, condition := range cluster.Status.Conditions {
+			args = append(args, fmt.Sprintf("(%s=%s %s)", condition.Type, condition.Status, condition.Reason))
+		}
+
+		log.Info("Current conditions", "conditions", strings.Join(args, ", "))
```
g-gaston, line 227: IMO we don't need this, we can drop it

controllers/cluster_controller.go
```diff
@@ -211,10 +214,33 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+
 		// Always attempt to patch the object and status after each reconciliation.
-		if err := patchHelper.Patch(ctx, cluster); err != nil {
+		patchOpts := []patch.Option{}
+
+		// Patch ObservedGeneration only if the reconciliation completed without error
```
g-gaston, line 232: I like this
But I would extend the comment to explain the "why", here I'm only getting the "what" (which I could possible get from just reading the code)

controllers/cluster_controller.go
```diff
@@ -211,10 +214,33 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+		if reterr == nil {
+			patchOpts = append(patchOpts, patch.WithStatusObservedGeneration{})
+		}
+		if err := patchCluster(ctx, patchHelper, cluster, patchOpts...); err != nil {
 			reterr = kerrors.NewAggregate([]error{reterr, err})
 		}
+
+		// Only requeue if we are not already re-queueing and the Cluster ready condition is false
```
g-gaston, line 240: same thing, a bit more about the why would be great
Also, maybe mention we do this to be able to update the status continuously until the cluster becomes ready, since there might be changes in state of the world that don't trigger reconciliation requests

controllers/cluster_controller.go
g-gaston, line 391: This log is confusing IMO
what about
"Cluster deletion in progress managed by the CLI, skipping updating cluster status"

controllers/cluster_controller.go
```diff
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+		log.Info("Cluster managed by CLI has been deleted but CAPI cluster may remain, skipping updating cluster status")
+		return nil
+	}
+
+	// ObservedGeneration represents the last observed generation, and consequently represents that the current status is up to date.
+	// Then if observedGeneration is equal to generation AND the Cluster's "Ready" condition is "True", then we can skip another status update
+	if cluster.Status.ObservedGeneration == cluster.Generation && conditions.IsTrue(cluster, anywherev1.ReadyCondition) {
+		log.Info("Generation matches observedGeneration and the cluster is ready, skipping status update")
+		return nil
+	}
```
g-gaston, line 400: I would vote to remove this
This will prevent the status from reflecting a cluster that becomes "not ready" after it has been ready in the past
Which is half of the value that the status provides

```suggestion
```

controllers/cluster_controller.go
```diff
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+
+	log.Info("Updating cluster status")
+	defer func() {
+		// Always update the readyCondition by summarizing the state of other conditions.
+		conditions.SetSummary(cluster,
```
g-gaston, line 405: do we want to do this even when there is an error?
it seems to me that this could just be at the end of the method without a defer

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,63 @@
...
+		conditions.Set(cluster, waitingForCPInitializedCondition())
+		return nil
+	}
+
+	conditions.MarkTrue(cluster, anywherev1.ControlPlaneInitializedCondition)
+	return nil
+}
+
+func waitingForCPInitializedCondition() *anywherev1.Condition {
+	return conditions.FalseCondition(anywherev1.ControlPlaneInitializedCondition, anywherev1.WaitingForControlPlaneInitializedReason, clusterv1.ConditionSeverityInfo, anywherev1.FirstControlPlaneUnavailableMessage)
```
vivek-koppuru, line 47: So this returns true when the CP initialized condition is false?

controllers/cluster_controller.go
```diff
@@ -357,6 +383,54 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	if err != nil {
+		return errors.Wrapf(err, "getting kubeadmcontrolplane")
+	}
+
+	if err = clusters.UpdateControlPlaneInitializedCondition(cluster, kcp); err != nil {
+		return errors.Wrap(err, "updating control plane initialized condition")
+	}
+
+	return nil
+}
```
g-gaston, line 432: what about just moving this whole thing to the clusters package
that way all the "status" logic is there and the controller just orchestrates when it's called

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
+package v1alpha1
+
+// Conditions and condition Reasons for the Cluster object.
+const (
+	// ReadyCondition reports a summary of other conditions, indicating an overall operational
+	// state of the cluster.
```
g-gaston, line 6: I think we need to extend this. Like reflect precessily what it means: all cp nodes in the right version, all worker nodes in the right version, all nodes ready, not old nodes

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
...
+	ReadyCondition ConditionType = "Ready"
+
+	// PendingUpdateReason reports when an update has been scheduled but has yet to happen.
```
g-gaston, line 9: This is a bit "obscure", like an "update" to what?
maybe something in the lines of "OutdatedInformationReason" "reports the system is waiting for stale cluster information to be refreshed"

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
...
+	// ControlPlaneInitializedCondition reports that all the control plane nodes are
+	// in a fully ready and running state.
+	ControlPlaneReadyCondition ConditionType = "ControlPlaneReady"
+
+	// ControlPlaneInitializedCondition reports that the first control plane instance has been initialized
+	// and so the control plane is available and an API server instance is ready for processing requests.
+	ControlPlaneInitializedCondition ConditionType = "ControlPlaneInitialized"
+
+	// WaitingForControlPlaneInitializedReason used when cluster is waiting for control plane to be initialized.
+	WaitingForControlPlaneInitializedReason = "WaitingForControlPlaneInitialized"
```
g-gaston, line 21: what is this reason telling us? really nothing right? what about `ControlPlaneInitializationInProgressReason`

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// FirstControlPlaneUnavailableMessage used when waiting for the first control plane instance to become
+	// initialized and available.
+	FirstControlPlaneUnavailableMessage = "The first control plane instance is not available yet"
+
+	// WaitingForControlPlaneReadyReason used when cluster is waiting for control plane nodes to transition to a
+	// fully ready and running state.
+	WaitingForControlPlaneReadyReason = "WaitingForControlPlaneReady"
```
g-gaston, line 29: same here, this doesn't say anything
what we want is the actual reasons why is not ready:
- init in progress
- scaling up
- scaling down
- upgrade in process
...

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// WorkersReadyConditon used reports the status on the worker nodes and is true when the expected
+	// number of workers are in a fully ready and running state.
+	WorkersReadyConditon ConditionType = "WorkersReady"
+
+	// WaitingForWorkersReadyReason used when cluster is waiting for control plane nodes to transition to a
+	// fully ready and running state.
+	WaitingForWorkersReadyReason = "WaitingForWorkersReady"
```
g-gaston, line 37: same comment as for CP

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// DefaultCNIConfiguredCondition reports the default cni cluster has been configured successfully.
+	DefaultCNIConfiguredCondition ConditionType = "DefaultCNIConfigured"
+
+	// WaitingForDefaultCNIConfiguredReason used when cluster is waiting default cni to be installed.
+	WaitingForDefaultCNIConfiguredReason = "WaitingForDefaultCNIConfigured"
```
g-gaston, line 43: same about "waiting" reasons

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func checkKubeadmControlPlaneError(cluster *anywherev1.Cluster, clusterConditionType anywherev1.ConditionType, kcp *controlplanev1.KubeadmControlPlane, kcpConditionType clusterv1.ConditionType) (*anywherev1.Condition, error) {
+	kcpCondition := conditions.Get(kcp, kcpConditionType)
+
+	if kcpCondition == nil {
+		return nil, fmt.Errorf("unable to read condition %s from kubeadmcontrolplane %s", kcpConditionType, kcp.Name)
```
g-gaston, line 54: this is misleading
conditions don't have to be present always, if they are not, they just have a "false value"

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,63 @@
...
+
+	conditions.MarkTrue(cluster, anywherev1.ControlPlaneInitializedCondition)
+	return nil
+}
+
+func waitingForCPInitializedCondition() *anywherev1.Condition {
+	return conditions.FalseCondition(anywherev1.ControlPlaneInitializedCondition, anywherev1.WaitingForControlPlaneInitializedReason, clusterv1.ConditionSeverityInfo, anywherev1.FirstControlPlaneUnavailableMessage)
+}
+
+func checkKubeadmControlPlaneError(cluster *anywherev1.Cluster, clusterConditionType anywherev1.ConditionType, kcp *controlplanev1.KubeadmControlPlane, kcpConditionType clusterv1.ConditionType) (*anywherev1.Condition, error) {
```
g-gaston, line 50: this method doesn't really tell much about what this does
maybe we can start by defining in a comment what it does and go from there

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,63 @@
...
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+// UpdateControlPlaneInitializedCondition updates the ControlPlaneInitialized condition if it hasn't already been set.
+// This condition should be set only once.
+func UpdateControlPlaneInitializedCondition(cluster *anywherev1.Cluster, kcp *controlplanev1.KubeadmControlPlane) error {
```
g-gaston, line 15: I'm missing checking the generation and observedGeneration in the KCP in this method

controllers/cluster_controller.go
cxbrowne1207: Ah. I'm sorry for the confusion. 

What we're trying to say here is that when the cluster deletion process is being managed by the CLI, we **should not** update the status on the EKS-A cluster object because the reconciler will just remove the finalizer on the EKS-A cluster object during reconcileDelete [here](https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L365). Therefore, we'd have no object to update the status of.

So, it's not that the CAPI cluster remains, but CLI is handling the deletion primarily. I'll update the comment and log here to be more clear.

cxbrowne1207: After talking offline that, we discussed that skipping updating the status when the cluster is marked for deletion AND has no finalizers is actually what we want to do here. The main thing is that we want to skip updating the status, if the cluster object has already been fully deleted because then subsequent patch operations would fail if the status is updated after it is fully deleted.

controllers/cluster_controller.go
cxbrowne1207: Okay, will do

pkg/controller/clusters/status.go
cxbrowne1207: No, this returns `anywherev1.Condition` object with "False" status. This might be clearer if it were a `const` instead.

controllers/cluster_controller.go
cxbrowne1207: Okay, I see your point. I'll move this to the end of the method

pkg/api/v1alpha1/condition_consts.go
cxbrowne1207: Okay, I will address the ones with the CP first, but then remove these and those to do with the CNI to keep the PR focused for now

cxbrowne1207: Okay. I think I will keep this PR focused for now by removing these reasons and include them in subsequent PRs when they're used

pkg/api/v1alpha1/condition_consts.go
cxbrowne1207: Okay. I think I will keep this PR focused for now by removing these reasons then include them in subsequent PRs when they're used

pkg/api/v1alpha1/condition_consts.go
cxbrowne1207: Same as the above

controllers/cluster_controller.go
cxbrowne1207: do we want to wait that long before another update on the status? that's what would happen whenever this is used since the status is updated after every reconcile.

pkg/controller/clusters/status.go
cxbrowne1207: Okay, I'll made changes to make the use of this method more clear.

pkg/controller/clusters/status.go
cxbrowne1207: I thought that it would be a good idea to throw an error if the condition did not exist here, as it was expected. I've removed it now. If it's false, then it will be handled as just being a "false value"

controllers/cluster_controller.go
cxbrowne1207: Created the ticket [here](https://github.com/aws/eks-anywhere-internal/issues/1449)

controllers/cluster_controller.go
cxbrowne1207: Removed

controllers/cluster_controller.go
cxbrowne1207: Agreed. I've moved this to the clusters package

vivek-koppuru (approved):

controllers/cluster_controller.go
```diff
@@ -211,10 +214,33 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+		if err != nil {
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+
 		// Always attempt to patch the object and status after each reconciliation.
-		if err := patchHelper.Patch(ctx, cluster); err != nil {
+		patchOpts := []patch.Option{}
+
+		// We want the observedGeneration to indicate, that the status shown is up-to-date given the desired spec of the same generation.
+		// However, if there is an error while updating the status, we may a partial status update, In this case,
```
g-gaston, line 226:
```suggestion
		// However, if there is an error while updating the status, we may get a partial status update, In this case,
```

controllers/cluster_controller.go
```diff
@@ -357,6 +383,32 @@ func (r *ClusterReconciler) postClusterProviderReconcile(ctx context.Context, lo
...
+	// When EKS-A cluster is fully deleted, we do not need to update the status. Without this check
+	// the subsequent patch operations would fail if the status is updated after it is fully deleted.
+	if !cluster.DeletionTimestamp.IsZero() && len(cluster.GetFinalizers()) == 0 {
+		log.Info("Cluster is fully deleted, skipping cluster status update")
+		return nil
+	}
+
+	log.Info("Updating cluster status")
+
+	if err := clusters.UpdateControlPlaneStatus(ctx, r.client, cluster); err != nil {
```
g-gaston, line 396: this name is a bit confusing
We are not updating the control plane status
we are updating the cluster status part that reflects the state of the cluster control plane
right?

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// ControlPlaneInitializedCondition reports that all the control plane nodes are
+	// in a fully ready and running state.
+	ControlPlaneReadyCondition ConditionType = "ControlPlaneReady"
+
+	// ControlPlaneInitializedCondition reports that the first control plane instance has been initialized
+	// and so the control plane is available and an API server instance is ready for processing requests.
+	ControlPlaneInitializedCondition ConditionType = "ControlPlaneInitialized"
+
+	// ControlPlaneInitializationInProgressReason used when cluster the control plane initilization is in progress.
```
g-gaston, line 21:
```suggestion
	// ControlPlaneInitializationInProgressReason reports that the control plane initilization is in progress.
```

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,34 @@
...
+	ControlPlaneInitializationInProgressReason = "ControlPlaneInitializationInProgress"
+
+	// FirstControlPlaneUnavailableMessage used when waiting for the first control plane instance to become
+	// initialized and available.
+	FirstControlPlaneUnavailableMessage = "The first control plane instance is not available yet"
```
g-gaston, line 26: what's the difference between this one and ControlPlaneInitializationInProgressReason?

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,34 @@
...
+
+	// WorkersReadyConditon used reports the status on the worker nodes and is true when the expected
+	// number of workers are in a fully ready and running state.
```
g-gaston, line 29: I think we might have left overs from the previous phrasing :)

pkg/api/v1alpha1/condition_consts.go
```diff
@@ -0,0 +1,34 @@
...
+	WorkersReadyConditon ConditionType = "WorkersReady"
+
+	// DefaultCNIConfiguredCondition reports the default cni cluster has been configured successfully.
+	DefaultCNIConfiguredCondition ConditionType = "DefaultCNIConfigured"
```
g-gaston, line 33: I wonder if `EKSACNI` is better than "default"
just because our api potentially supports different cni's
and they are all "default"? that's a bit weird

pkg/controller/clusters/status.go
```diff
@@ -0,0 +1,84 @@
...
+	// We make sure to check that the status is up to date before using it
+	if kcp.Status.ObservedGeneration != kcp.ObjectMeta.Generation {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneInitializedCondition, anywherev1.OutdatedInformationReason, clusterv1.ConditionSeverityInfo, "")
+		return
+	}
+
+	// Surface any errors from the KubeadmControlPlane, relating to control plane availability.
+	kcpCondition := getKubeadmControlPlaneConditionIfError(kcp, controlplanev1.AvailableCondition)
+	if kcpCondition != nil {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneInitializedCondition, kcpCondition.Reason, kcpCondition.Severity, kcpCondition.Message)
```
g-gaston, line 51: I'm ok with this for now but we might want to think about it a bit more
reasons are supposed to be machine readable, which means the should be part of our api, so our users know what to look for
here we are mirroring capi blindly, which means our user won't know about these reasons

pkg/api/v1alpha1/condition_consts.go
cxbrowne1207: This is the corresponding message for that reason.

pkg/api/v1alpha1/condition_consts.go
cxbrowne1207: Yeah, we can discussion this a little bit more offline with the team.

pkg/controller/clusters/status.go
cxbrowne1207: Okay, we'll remove this section for now as the goal is currently to track progress of lifecycle operations. We can need to surface errors, make them more specific. We'll may iterate over this to surface errors in the future.

controllers/cluster_controller.go
cxbrowne1207: yes, i've updated the name to reflect that purpose better.

g-gaston (commented):
/lgtm

niceeee

PR #6056, opened 2023-06-20
Apply insecure flag to oras copy

d8660091:
*Issue #, if available:* When `--insecure` is passed `eksctl-anywhere import images` command, curated packages yaml bundles cannot be pushed to registry mirror with self generated cert. The following error will show:

`x509: certificate signed by unknown authority`

*Description of changes:* Insecure flag is applied to oras

*Testing (if applicable):* Manual invoked eksctl-anywhere import images with and without insecure flag.


TerryHowe (approved):
/lgtm

PR #6067, opened 2023-06-21
Add ControlPlaneReady condition for Cluster status

cxbrowne1207:
*Issue #, if available:*
#5628 

*Description of changes:*
Changes to the Cluster status as per the the proposal [here]( https://github.com/aws/eks-anywhere/pull/6005):
-`ControlPlaneReady` condition

*Testing (if applicable):*
- Unit testing
- Manual testing

### DIFF
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -72,9 +72,12 @@ func updateControlPlaneReadyCondition(cluster *anywherev1.Cluster, kcp *controlp
 	}
 
 	readyReplicas := int(kcp.Status.ReadyReplicas)
-	if readyReplicas == expected {
-		conditions.MarkTrue(cluster, anywherev1.ControlPlaneReadyCondition)
+	if readyReplicas != expected {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, anywherev1.NodesNotReadyReason, clusterv1.ConditionSeverityInfo, "Control plane nodes not ready yet, %d expected (%d ready)", expected, readyReplicas)
+		return
 	}
+
+	conditions.MarkTrue(cluster, anywherev1.ControlPlaneReadyCondition)
 }
 
 // updateControlPlaneInitializedCondition updates the ControlPlaneInitialized condition if it hasn't already been set.
--- a/pkg/controller/clusters/status_test.go
+++ b/pkg/controller/clusters/status_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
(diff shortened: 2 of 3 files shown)


pkg/controller/clusters/status.go
```diff
@@ -22,10 +22,61 @@ func UpdateClusterStatusForControlPlane(ctx context.Context, client client.Clien
...
+		return
+	}
+
+	if totalReplicas > expected {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, anywherev1.ScalingDownReason, clusterv1.ConditionSeverityInfo, "Scaling down control plane nodes, %d expected (%d actual)", expected, totalReplicas)
+		return
+	}
+
+	readyReplicas := int(kcp.Status.ReadyReplicas)
+	if readyReplicas == expected {
```
g-gaston, line 75: what happens when this is not true?

pkg/controller/clusters/status_test.go
```diff
@@ -128,11 +293,19 @@ func TestUpdateControlPlaneInitializedCondition(t *testing.T) {
...
+			condition := conditions.Get(cluster, tt.wantCondition.Type)
 			g.Expect(condition).ToNot(BeNil())
 
+			wantMessage := tt.wantCondition.Message
+			message := condition.Message
+
+			// Clear these out for easier comparison
 			condition.LastTransitionTime = v1.Time{}
+			condition.Message = ""
+			tt.wantCondition.Message = ""
```
g-gaston, line 305: why change the test input?

pkg/controller/clusters/status_test.go
```diff
@@ -128,11 +293,19 @@ func TestUpdateControlPlaneInitializedCondition(t *testing.T) {
...
-			condition := conditions.Get(cluster, anywherev1.ControlPlaneInitializedCondition)
+			condition := conditions.Get(cluster, tt.wantCondition.Type)
 			g.Expect(condition).ToNot(BeNil())
 
+			wantMessage := tt.wantCondition.Message
+			message := condition.Message
+
+			// Clear these out for easier comparison
 			condition.LastTransitionTime = v1.Time{}
+			condition.Message = ""
```
g-gaston, line 304: this is a bit hacky
can we just compare the fields we are interested in directly? it's more explicit and easier to understand

pkg/controller/clusters/status_test.go
cxbrowne1207: I was trying to isolate the message so that we could compare them by substring.

pkg/controller/clusters/status_test.go
cxbrowne1207: Yeah, we can do that

pkg/controller/clusters/status.go
cxbrowne1207: If it isn't true, then no update is made. The intention here was to preserve the previous false status until everything is ready.

Though, I suppose we may have a situation where there is no previous status, and the nodes aren't ready yet. I think then we'll need another "reason" here. Something like `ProvisioningInProgress`?

g-gaston: Under what scenarios we might get here?
- no outdated replicas. This means all replicas existing are at the desired version.
- number of existing replicas is the same as desired replicas.

So we are either:
- provisioning the last replica of the CP (machine already created but not healthy yet)
- one or more of the existing machines became not ready

So `ProvisioningInProgress` only covers half of those scenarios. What about "NodesNotReady"? that is true for both scenarios.
And in the future we can always add logic to distinguish between the two scenarios if necessary.

pkg/controller/clusters/status_test.go
g-gaston: is this required if we compare field by field instead of the full condition?

pkg/controller/clusters/status.go
cxbrowne1207: Yeah `NodesNotReady` sounds good to me in that case

pkg/controller/clusters/status_test.go
cxbrowne1207: no, there is no reason that we can't. this has been updated to directly compare the fields instead.

PR #6069, opened 2023-06-22
Set skip ip annotation on cluster create for skip ip check flag

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Set skip ip annotation on cluster create when skip ip check flag is passed to the CLI. This is done for future work where the CLI would be using the controller to create workload clusters and the control plane ip validation needs to be skipped by the controller.

*Testing (if applicable):*
- unit tests
- test manually on vsphere cluster

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -1167,17 +1167,12 @@ func (c *Cluster) DisableControlPlaneIPCheck() {
 	c.Annotations[skipIPCheckAnnotation] = "true"
 }
 
-// SkipControlPlaneIPCheck checks it the `skip-ip-check` annotation is set on the Cluster object.
-func (c *Cluster) SkipControlPlaneIPCheck() bool {
+// ControlPlaneIPCheckDisabled checks it the `skip-ip-check` annotation is set on the Cluster object.
+func (c *Cluster) ControlPlaneIPCheckDisabled() bool {
 	if s, ok := c.Annotations[skipIPCheckAnnotation]; ok {
 		return s == "true"
 	}
 	return false
-}
-
-// SkipIPCheckAnnotation returns the `skip-ip-check` annotation as a string.
-func (c *Cluster) SkipIPCheckAnnotation() string {
-	return skipIPCheckAnnotation
 }
 
 func (c *Cluster) ResourceType() string {
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -33,6 +33,9 @@ const (
 	// etcdAnnotation can be applied to EKS-A machineconfig CR for etcd, to prevent controller from making changes to it.
 	etcdAnnotation = "anywhere.eks.amazonaws.com/etcd"
 
+	// skipIPCheckAnnotation skips the availability control plane IP validation during cluster creation. Use only if your network configuration is conflicting with the default port scan.
+	skipIPCheckAnnotation = "anywhere.eks.amazonaws.com/skip-ip-check"
+
 	// managementAnnotation points to the name of a management cluster
 	// cluster object.
 	managementAnnotation = "anywhere.eks.amazonaws.com/managed-by"
@@ -1154,6 +1157,27 @@ func (c *Cluster) PausedAnnotation() string {
 
 func (c *Cluster) ControlPlaneAnnotation() string {
 	return controlPlaneAnnotation
+}
+
+// DisableControlPlaneIPCheck sets the `skip-ip-check` annotation on the Cluster object.
+func (c *Cluster) DisableControlPlaneIPCheck() {
+	if c.Annotations == nil {
+		c.Annotations = make(map[string]string, 1)
+	}
+	c.Annotations[skipIPCheckAnnotation] = "true"
+}
+
+// SkipControlPlaneIPCheck checks it the `skip-ip-check` annotation is set on the Cluster object.
+func (c *Cluster) SkipControlPlaneIPCheck() bool {
+	if s, ok := c.Annotations[skipIPCheckAnnotation]; ok {
+		return s == "true"
+	}
+	return false
+}
+
+// SkipIPCheckAnnotation returns the `skip-ip-check` annotation as a string.
+func (c *Cluster) SkipIPCheckAnnotation() string {
+	return skipIPCheckAnnotation
 }
 
 func (c *Cluster) ResourceType() string {
--- a/pkg/cli/createclusterdefaulter.go
+++ b/pkg/cli/createclusterdefaulter.go
@@ -1,0 +1,33 @@
+package cli
+
+import (
+	"context"
+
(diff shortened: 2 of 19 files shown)


pkg/cluster/builder.go
```diff
@@ -55,7 +55,7 @@ func NewFileSpecBuilder(reader manifests.FileReader, cliVersion version.Info, op
 
 // Build constructs a new [Spec] by reading the cluster config in yaml from a file and
 // Releases, Bundles and EKS-D manifests from the configured URLs.
-func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
+func (b FileSpecBuilder) Build(clusterConfigURL string, skipIPCheck bool) (*Spec, error) {
```
g-gaston, line 58: we should avoid adding any logic here for "options"

can we have something a like a "defaulter" service?

Maybe it's time we start using this
https://github.com/aws/eks-anywhere/blob/main/designs/validations.md

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -33,6 +33,9 @@ const (
 	// etcdAnnotation can be applied to EKS-A machineconfig CR for etcd, to prevent controller from making changes to it.
 	etcdAnnotation = "anywhere.eks.amazonaws.com/etcd"
 
+	// skipIPCheckAnnotation can be applied to EKS-A cluster, to prevent controller from validating the control plane IP.
```
g-gaston, line 36:
```suggestion
	// SkipIPCheckAnnotation skips the availability control plane IP validation during cluster creation. Use only if your network configuration is conflicting with the default port scan.
```

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1137,6 +1140,27 @@ func (c *Cluster) ControlPlaneAnnotation() string {
 	return controlPlaneAnnotation
 }
 
+// SetSkipIPCheck sets the `skip-ip-check` annotation on the Cluster object.
+func (c *Cluster) SetSkipIPCheck() {
```
g-gaston, line 1144:
```suggestion
func (c *Cluster) DisableControlPlaneIPCheck() {
```

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1137,6 +1140,27 @@ func (c *Cluster) ControlPlaneAnnotation() string {
...
+	if c.Annotations == nil {
+		c.Annotations = make(map[string]string, 1)
+	}
+	c.Annotations[skipIPCheckAnnotation] = "true"
+}
+
+// IsSkipIPCheck checks it the `skip-ip-check` annotation is set on the Cluster object.
+func (c *Cluster) IsSkipIPCheck() bool {
```
g-gaston, line 1152:
```suggestion
func (c *Cluster) SkipControlPlaneIPCheck() bool {
```

pkg/providers/tinkerbell/reconciler/reconciler.go
```diff
@@ -95,8 +95,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *an
 		return controller.Result{}, err
 	}
 
-	return controller.NewPhaseRunner[*Scope]().Register(
-		r.ValidateControlPlaneIP,
```
g-gaston, line 99: Alternatively, you could add this logic inside `ipValidator.ValidateControlPlaneIP` and avoid having to modify all reconcilers
That's the advnatage of having the shared logic in a common place

pkg/config/config.go
```diff
@@ -17,3 +17,8 @@ type CliConfig struct {
 	GitPrivateKeyFile   string
 	GitKnownHostsFile   string
 }
+
+// CreateCliConfig is the config we use for create specific configurations.
```
g-gaston, line 21:
```suggestion
// CreateClusterCLIConfig is the config we use for create cluster specific configurations.
```
we could also put this in the `cli` package

pkg/dependencies/factory.go
```diff
@@ -1004,6 +1007,21 @@ func (f *Factory) WithCliConfig(cliConfig *cliconfig.CliConfig) *Factory {
 	return f
 }
 
+// WithCreateClusterDefaulter builds a create cluster defaulter that builds defaulter dependencies specific to the create cluster command. The defaulter is then run once the factory is built in the create cluster command.
+func (f *Factory) WithCreateClusterDefaulter(createCliConfig *cliconfig.CreateCliConfig) *Factory {
```
g-gaston, line 1011: why a pointer?

pkg/providers/validator/validate.go
```diff
@@ -37,6 +37,10 @@ func NewIPValidator(opts ...IPValidatorOpt) *IPValidator {
 // ValidateControlPlaneIPUniqueness checks whether or not the control plane endpoint defined
 // in the cluster spec is available.
 func (v *IPValidator) ValidateControlPlaneIPUniqueness(cluster *v1alpha1.Cluster) error {
+	if cluster.SkipControlPlaneIPCheck() {
+		return nil
```
g-gaston, line 41: it would be interesting to log this

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1156,6 +1159,27 @@ func (c *Cluster) ControlPlaneAnnotation() string {
...
+// SkipControlPlaneIPCheck checks it the `skip-ip-check` annotation is set on the Cluster object.
+func (c *Cluster) SkipControlPlaneIPCheck() bool {
+	if s, ok := c.Annotations[skipIPCheckAnnotation]; ok {
+		return s == "true"
+	}
+	return false
+}
+
+// SkipIPCheckAnnotation returns the `skip-ip-check` annotation as a string.
+func (c *Cluster) SkipIPCheckAnnotation() string {
```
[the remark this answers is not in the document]
mitalipaygude: This isn't being used in the code anywhere so I can remove it. I added this exported method in case the annotation needs to be accessed outside of the v1alpha1 package.

pkg/providers/validator/validate.go
mitalipaygude: I thought of logging this, however this is also used in the CLI and it would need some refactoring to send the log parameter here for logging.

pkg/cli/createclusterdefaulter.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/aws/eks-anywhere/pkg/defaulting"
+)
+
+// CreateClusterDefaulter defines the cluster defaulter for create cluster command defaults.
+type CreateClusterDefaulter struct {
+	runner *defaulting.Runner[*cluster.Spec]
+}
+
+// Default is the type used for registering various default methods in the defaulting runner.
+type Default func(ctx context.Context, spec *cluster.Spec) (*cluster.Spec, error)
```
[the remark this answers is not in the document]
mitalipaygude: forgot to clean it up, removing

PR #6071, opened 2023-06-22
Add WorkersReady condition to cluster status

cxbrowne1207:
Issue #, if available:
#5628 

Description of changes:
- WorkersReady condition

Testing (if applicable):

- Unit testing
- Manual testing

### DIFF
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -74,6 +74,20 @@ func testKubeadmControlPlaneFromCluster(cluster *anywherev1.Cluster) *controlpla
 	})
 }
 
+func machineDeploymentsFromCluster(cluster *anywherev1.Cluster) []clusterv1.MachineDeployment {
+	return []clusterv1.MachineDeployment{
+		*test.MachineDeployment(func(md *clusterv1.MachineDeployment) {
+			md.ObjectMeta.Name = "md-0"
+			md.ObjectMeta.Labels = map[string]string{
+				clusterv1.ClusterNameLabel: cluster.Name,
+			}
+			md.Status.Replicas = 1
+			md.Status.ReadyReplicas = 1
+			md.Status.UpdatedReplicas = 1
+		}),
+	}
+}
+
 func newVsphereClusterReconcilerTest(t *testing.T, objs ...runtime.Object) *vsphereClusterReconcilerTest {
 	ctrl := gomock.NewController(t)
 	govcClient := vspheremocks.NewMockProviderGovcClient(ctrl)
@@ -248,17 +262,23 @@ func TestClusterReconcilerReconcileGenerations(t *testing.T) {
 				awsIAM.Generation = tt.awsIAMGeneration
 			}
 
+			kcp := testKubeadmControlPlaneFromCluster(config.Cluster)
+			machineDeployments := machineDeploymentsFromCluster(config.Cluster)
+
 			g := NewWithT(t)
 			ctx := context.Background()
 
-			objs := make([]runtime.Object, 0, 7)
+			objs := make([]runtime.Object, 0, 8)
 			objs = append(objs, config.Cluster, bundles)
 			for _, o := range config.ChildObjects() {
 				objs = append(objs, o)
 			}
 
-			kcp := testKubeadmControlPlaneFromCluster(config.Cluster)
 			objs = append(objs, kcp)
+
+			for _, obj := range machineDeployments {
+				objs = append(objs, obj.DeepCopy())
+			}
 
 			client := fake.NewClientBuilder().WithRuntimeObjects(objs...).Build()
 			mockCtrl := gomock.NewController(t)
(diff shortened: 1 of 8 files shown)


controllers/cluster_controller_test.go
```diff
@@ -248,18 +262,24 @@ func TestClusterReconcilerReconcileGenerations(t *testing.T) {
...
 			}
 
+			kcp := testKubeadmControlPlaneFromCluster(config.Cluster)
+			machineDeployments := machineDeploymentsFromCluster(config.Cluster)
+
 			g := NewWithT(t)
 			ctx := context.Background()
 
-			objs := make([]runtime.Object, 0, 7)
+			objs := make([]runtime.Object, 0, 8)
```
g-gaston, line 271:
```suggestion
			objs := make([]runtime.Object, 0, 7+len(machineDeployments))
```

internal/test/testdata.go
```diff
@@ -209,5 +209,5 @@ func MachineDeployment(opts ...MachineDeploymentOpt) *clusterv1.MachineDeploymen
 		opt(md)
 	}
 
-	return md
+	return md.DeepCopy()
```
g-gaston, line 212: why deep copy?

pkg/controller/clusterapi.go
```diff
@@ -91,3 +91,18 @@ func GetMachineDeployment(ctx context.Context, client client.Client, machineDepl
...
 	return machineDeployment, nil
 }
+
+// GetMachineDeployments reads all of cluster-api MachineDeployment for an eks-a cluster using a kube client.
+func GetMachineDeployments(ctx context.Context, c client.Client, cluster *anywherev1.Cluster) ([]clusterv1.MachineDeployment, error) {
+	machineDeployments := &clusterv1.MachineDeploymentList{}
+
+	err := c.List(ctx, machineDeployments, client.MatchingLabels{clusterv1.ClusterNameLabel: cluster.Name}, client.InNamespace(constants.EksaSystemNamespace))
+	if apierrors.IsNotFound(err) {
+		return nil, nil
```
g-gaston, line 101: we should test this since it's part of the main behavior this method proposes

pkg/controller/clusters/status.go
```diff
@@ -110,6 +122,70 @@ func updateControlPlaneInitializedCondition(cluster *anywherev1.Cluster, kcp *co
 	conditions.MarkTrue(cluster, anywherev1.ControlPlaneInitializedCondition)
 }
 
+// updateWorkersReadyCondition updates the WorkersReadyConditon condition after checking the state of the worker node groups
+// in the cluster.
+func updateWorkersReadyCondition(cluster *anywherev1.Cluster, machineDeployments []clusterv1.MachineDeployment) {
+	initializedCondition := conditions.Get(cluster, anywherev1.ControlPlaneInitializedCondition)
+	if initializedCondition.Status != "True" {
+		conditions.MarkFalse(cluster, anywherev1.WorkersReadyConditon, initializedCondition.Reason, initializedCondition.Severity, initializedCondition.Message)
```
g-gaston, line 130: I'm not sure about the reusing the reason here, it might not work well always
What about just using "ControlPlaneNotInitialized" always?
if the user wants know why that is true, they can check the cp condition

pkg/controller/clusters/status.go
```diff
@@ -110,6 +122,70 @@ func updateControlPlaneInitializedCondition(cluster *anywherev1.Cluster, kcp *co
...
+		return
+	}
+
+	// TODO: Here we should update the status based on the DefaultCNIConfigured condition. When it's false and skipUpgrades
+	// is not enabled for the CNI, WorkersReady should be marked false with the corresponding reason from the
+	// DefaultCNIConfigured condition.
```
g-gaston, line 136: I don't necessarily agree with this
This makes an assumption about the order in which conditions are set and the order the clusters are reconciled
Why not just set the reason based on the observed state of the cluster?

pkg/controller/clusters/status.go
```diff
@@ -110,6 +122,70 @@ func updateControlPlaneInitializedCondition(cluster *anywherev1.Cluster, kcp *co
...
+	// First, we need aggregate the number of nodes across worker node groups  to be able to assess the condition of the workers
+	// as a whole.
+	totalReadyReplicas := 0
+	totalUpdatedReplicas := 0
+	totalReplicas := 0
+
+	for _, md := range machineDeployments {
+		// We make sure to check that the status is up to date before using the information from the machine deployment status.
+		if md.Status.ObservedGeneration != md.ObjectMeta.Generation {
+			conditions.MarkFalse(cluster, anywherev1.WorkersReadyConditon, anywherev1.OutdatedInformationReason, clusterv1.ConditionSeverityInfo, "worker node group %s status not up to date yet", md.Name)
```
g-gaston, line 152:
```suggestion
			conditions.MarkFalse(cluster, anywherev1.WorkersReadyConditon, anywherev1.OutdatedInformationReason, clusterv1.ConditionSeverityInfo, "Worker node group %s status not up to date yet", md.Name)
```

internal/test/testdata.go
cxbrowne1207: Ah, I was suppose remove this from here, I was testing something here. Thanks for catching that

pkg/controller/clusters/status.go
cxbrowne1207: I had been assume an order the clusters are reconciled here, but yeah if we don't want that we shouldn't do this. It's also probably not needed at all, since the workers condition should focus on the state of the workers. I'll remove this comment

pkg/controller/clusterapi.go
cxbrowne1207: This one should not be here actually :)

g-gaston (approved):

PR #6085, opened 2023-06-26
Fix side effects E2E tests

taneyland:
*Issue #, if available:*

*Description of changes:*
We are setting OIDCConfig in the workload clusters in this test, so we need to set OIDC env vars in E2E infra for management cluster upgrade side effects tests by mentioning OIDC in test name

*Testing (if applicable):*

### DIFF
--- a/test/e2e/docker_test.go
+++ b/test/e2e/docker_test.go
@@ -843,7 +843,7 @@ func TestDockerKubernetes126to127GithubFluxEnabledUpgradeFromLatestMinorRelease(
 	)
 }
 
-func TestDockerKubernetes126ManagementClusterUpgradeFromLatestSideEffects(t *testing.T) {
+func TestDockerKubernetes126WithOIDCManagementClusterUpgradeFromLatestSideEffects(t *testing.T) {
 	provider := framework.NewDocker(t)
 	runTestManagementClusterUpgradeSideEffects(t, provider, "", v1alpha1.Kube126)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -137,7 +137,7 @@ EKS_A_CROSS_PLATFORMS := $(foreach platform,$(EKS_A_PLATFORMS),eks-a-cross-platf
 E2E_CROSS_PLATFORMS := $(foreach platform,$(EKS_A_PLATFORMS),e2e-cross-platform-$(platform))
 EKS_A_RELEASE_CROSS_PLATFORMS := $(foreach platform,$(EKS_A_PLATFORMS),eks-a-release-cross-platform-$(platform))
 
-DOCKER_E2E_TEST := TestDockerKubernetes125SimpleFlow
+DOCKER_E2E_TEST := TestDockerKubernetes126ManagementClusterUpgradeFromLatestSideEffects
```
cxbrowne1207, line 140: Do we want to keep this change?

cxbrowne1207 (commented):
Why did we need to add OIDC env vars check for this test?

Makefile
taneyland: No I'm just testing. I have the hold label on the PR

taneyland:
> Why did we need to add OIDC env vars check for this test?

We're setting the OIDC config in the workload clusters

PR #6087, opened 2023-06-26
Add DefaultCNIConfigured condition to cluster status

cxbrowne1207:
*Issue #, if available:*
#5628 

*Description of changes:*

*Testing (if applicable):*
- Unit testing

### DIFF
--- a/pkg/networking/cilium/reconciler/reconciler.go
+++ b/pkg/networking/cilium/reconciler/reconciler.go
@@ -7,9 +7,12 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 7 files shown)


pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -103,12 +110,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
...
 			return controller.Result{}, err
-		} else if result.Return() {
+		}
+
+		// When there is no error but a requeue has been returned, then a version upgrade is in progress
+		if !result.Result.IsZero() {
+			conditions.MarkFalse(spec.Cluster, anywherev1.DefaultCNIConfiguredCondition, anywherev1.DefaultCNIUpgradeInProgressReason, clusterv1.ConditionSeverityInfo, "Cilium version upgrade needed")
+		}
+
+		if result.Return() {
```
g-gaston, line 123: this seems redundant
why not keep the code as it was and just add the markFalse just before the return?

cxbrowne1207: If `result.Return()` just means `r.Result != nil`, then that means it can possible that it does not contain a re-queue request. So, here I wanted to make sure that there was a re-queue, without nesting another if statement in that block that checks for `result.Return()`.

PR #6105, opened 2023-06-30
Add FailureReason to Cluster status

cxbrowne1207:
Issue #, if available:
#5628 

Description of changes:

- Added FailureReason

Testing (if applicable):
- Unit testing
- Manual testing

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -1272,7 +1272,7 @@ func (c *Cluster) MachineConfigRefs() []Ref {
 }
 
 // SetFailure sets the failureMessage and failureReason of the Cluster status.
-func (c *Cluster) SetFailure(failureMessage string, failureReason FailureReasonType) {
+func (c *Cluster) SetFailure(failureReason FailureReasonType, failureMessage string) {
 	c.Status.FailureMessage = ptr.String(failureMessage)
 	c.Status.FailureReason = &failureReason
 }
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/semver"
+	"github.com/aws/eks-anywhere/pkg/utils/ptr"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -832,11 +832,48 @@ var validCiliumPolicyEnforcementModes = map[CiliumPolicyEnforcementMode]bool{
...
 // ClusterStatus defines the observed state of Cluster.
 type ClusterStatus struct {
 	// Descriptive message about a fatal problem while reconciling a cluster
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	// Machine readable value about a terminal problem while reconciling the cluster
+	// set at the same time as failureMessage
+	// +optional
+	FailureReason *string `json:"failureReason,omitempty"`
```
g-gaston, line 875: maybe worth making this its own type? it sets intention about these field having a limited number of preset values.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -832,11 +832,48 @@ var validCiliumPolicyEnforcementModes = map[CiliumPolicyEnforcementMode]bool{
...
+	UnavailableControlPlaneIPReason = "UnavailableControlPlaneIP"
+)
+
+// Reasons for the terminal failures while reconciling the Cluster object specific for Tinkerbell.
+const (
+	// HardwareValidationFailedReason reports that the hardware validation has failed.
+	HardwareValidationFailedReason = "HardwareValidationFailed"
+
+	// RufioMachineValidationFailedReason reports that the rufio machine validation has failed.
+	RufioMachineValidationFailedReason = "RufioMachineValidationFailed"
```
g-gaston, line 863: Can we avoid having the Rufio name in our API?
I think that the name of an specific internal component shouldn't bleed into our public API. What if the name of the component changes?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -832,11 +832,48 @@ var validCiliumPolicyEnforcementModes = map[CiliumPolicyEnforcementMode]bool{
...
+const (
+	// MissingDependentObjectsReason reports that the Cluster is missing dependent objects.
+	MissingDependentObjectsReason = "MissingDependentObjects"
+
+	// ManagementClusterRefInvalidReason reports that the Cluster management cluster reference is invalid. This
+	// can whether if it does not exist or the cluster referenced is not a management cluster.
+	ManagementClusterRefInvalidReason = "ManagementClusterRefInvalid"
+
+	// ClusterSpecValidationFailedReason reports that the Cluster spec validation has failed.
+	ClusterSpecValidationFailedReason = "ClusterSpecValidationFailed"
```
g-gaston, line 845: instead of validation failed, can we just say `...Invalid`? same comment for the other reasons

Also I don't think we need to specify `ClusterSpec` right? Can it just be `ClusterInvalid`?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1211,6 +1248,12 @@ func (c *Cluster) MachineConfigRefs() []Ref {
 	return machineConfigRefMap.toSlice()
 }
 
+// SetFailure sets the failureMessage and failureReason of the Cluster status.
+func (c *Cluster) SetFailure(failureMessage *string, failureReason *string) {
```
g-gaston, line 1252: this is a helper, so why wouldn't we just take strings and do the pointer conversion inside?
it guarantees not nil pointers and also it makes using this helper easier

pkg/api/v1alpha1/cluster_types.go
cxbrowne1207: Good idea

pkg/api/v1alpha1/cluster_types.go
cxbrowne1207: Updated to take a string and a `FailureReasonType`. Also added a `ClearFailure` helper for clearing failures.

pkg/providers/snow/reconciler/reconciler_test.go
```diff
@@ -261,6 +273,7 @@ func TestReconcilerReconcileCNIErrorClientRegistry(t *testing.T) {
 
 	tt.Expect(err).To(MatchError(ContainSubstring("building client")))
 	tt.Expect(tt.cluster.Status.FailureMessage).To(BeZero())
+	tt.Expect(tt.cluster.Status.FailureReason).To(BeZero())
```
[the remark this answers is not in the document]
cxbrowne1207: I mainly wanted to be consistent with the previous `FailureMessage` check

PR #6127, opened 2023-07-06
Dynamic hardware reservation for Tinkerbell tests

rahulbabu95:
*Issue #, if available:*

At present, the E2E tests for Tinkerbell provider are split between 6 test runners with each runner being statically allocated hardware pool of 4 baremetal machines. This limits the current E2E runs to just trigger tests that require 4 or less hardware and skip tests that need more than 4 hardware. Also, since multiple tests are split and grouped into a single test runner, we face the SSM timeout issue which has a hard limit of 5 Hours (300 minutes). Multiple Tink E2E tests fail due to timeout without being triggered due to above reason. 

*Description of changes:*

The change introduces a dynamic hardware reservation system for the Tinkerbell E2E tests. The change maps each test with a single instance of test runner (similar to other providers at present), and polls for hardware inventory dynamically enabling in running larger tests that need more than 4 hardware.

NOTE: This change only affects Non-airgapped Tinkerbell tests as the Airgapped Tests have dedicated hardware pool and are a small proportion of Tinkerbell tests. 

*Testing (if applicable):*


test/e2e/tink_hw_count.yaml
```diff
@@ -0,0 +1,235 @@
+TinkTestsHwCount:
```
abhinavmpandey08, line 1: nit: Can we call it `TinkerbellTestsHardwwareCount`

test/e2e/tink_hw_count.yaml
```diff
@@ -0,0 +1,235 @@
...
+- name: TestTinkerbellKubernetes123AWSIamAuth
```
abhinavmpandey08, line 2: nit: Can you sort this list lexicographically so the tests under here are easier to find?

test/e2e/tinkhwcount.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"sigs.k8s.io/yaml"
+)
+
+// TinkTest maps each Tinkbell test with the hardware count needed for the test.
```
abhinavmpandey08, line 11: nit: Let's use Tinkerbell everywhere instead of Tink
```suggestion
// TinkerbellTest maps each Tinkerbell test with the hardware count needed for the test.
```

internal/test/e2e/tinkHwQueue.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"sync"
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+)
+
+type hwQueue struct {
+	hws []*api.Hardware
+	mu  sync.Mutex
```
abhinavmpandey08, line 11: What's the mutex being used for? Is this queue supposed to be thread safe?

internal/test/e2e/run.go
```diff
@@ -482,3 +470,47 @@ func logTestGroups(logger logr.Logger, instancesConf []instanceRunConf) {
...
+
+func reserveTinkHw(conf instanceRunConf, invQueue *hwQueue) {
+	now := time.Now()
+	after := now.Add(hwPollingTimeout)
+	for {
+		if now.After(after) {
+			conf.logger.V(1).Info("TestRunner timed out waiting for hardware", "instanceId", conf.instanceId)
+			break
+		}
+		invQueue.mu.Lock()
```
abhinavmpandey08, line 499: The mutex locking and unlocking should ideally happen at the queue side. The tests should only care about getting the requested hardware from the queue, not handling concurrency for the queue.

internal/test/e2e/run.go
```diff
@@ -482,3 +470,47 @@ func logTestGroups(logger logr.Logger, instancesConf []instanceRunConf) {
...
+		if conf.hardwareCount <= len(invQueue.hws) {
+			conf.logger.V(1).Info("Reserved hardware for TestRunner", "instanceId", conf.instanceId, "hardwarePool", conf.hardware)
+			conf.hardware = invQueue.dequeueHw(conf.hardwareCount)
```
abhinavmpandey08, line 502: Let's re-think on how we want to handle the hardware reservation. I don't think handling enqueuing and dequeuing hardware at the test level is the best idea. If the test thread crashes or fails for some reason, that the hardware it reserved is lost.
I think what you should do from the tests' perspective is to just ask the hardware reservation system to reserve some number of hardware. The hardware reservation system should handle the implementation details on how to it wants to reserve the hardware, for example enqueuing and de-queuing the hardware. Also, there when you dequeue the hardware, you should store a list of hardware that is reserved by a test in a map so that info is not lost.

test/e2e/tink_hw_count.yaml
rahulbabu95: Right now the order is the same as the one in the E2E file. If we are to find a test here, can't we directly search for the test that we are looking for.? Although sorting lexicographically can help a bit, when adding a new test and finding the right place in the file will be a pain point! given the ad-hoc nature in which we add e2e tests i am not sure if it will be benefecial to maintain the order here.

internal/test/e2e/tinkHwQueue.go
rahulbabu95: Yes, its a concurrent queue since multiple Goroutines tries to read and write from the queue it needs to be thread safe!

internal/test/e2e/tinkerbellHardwareQueue.go
```diff
@@ -0,0 +1,58 @@
...
+	now := time.Now()
+	after := now.Add(hwPollingTimeout)
+	for {
+		if now.After(after) {
+			return nil, fmt.Errorf("hardware polling request timed out")
+		}
+		hwQu.mu.Lock()
+		if count <= len(hwQu.hws) {
+			hardwareReserved := hwQu.dequeueHw(count)
+			hwQu.mu.Unlock()
```
pokearu, line 29: nit: do a defer unlock after the lock on line 27.

internal/test/e2e/tinkerbellHardwareQueue.go
```diff
@@ -0,0 +1,58 @@
...
+			return hardwareReserved, nil
+		}
+		hwQu.mu.Unlock()
+		time.Sleep(5 * time.Minute)
```
pokearu, line 33: Maybe reduce the timeout to 1 min and make it a variable thats configurable?

internal/test/e2e/run.go
```diff
@@ -482,3 +471,35 @@ func logTestGroups(logger logr.Logger, instancesConf []instanceRunConf) {
...
+	return hardware, nil
+}
+
+func reserveTinkerbellHardware(conf instanceRunConf, invQueue *hardwareQueue) error {
+	reservedTinkerbellHardware, err := invQueue.reserveHardware(conf.hardwareCount)
+	if err != nil {
+		return fmt.Errorf("TestRunner timed out waiting for hardware - test instanceId: %v", conf.instanceId)
+	}
+	conf.hardware = reservedTinkerbellHardware
+	conf.logger.V(1).Info("Reserved hardware for TestRunner", "instanceId", conf.instanceId, "hardwarePool", conf.hardware)
```
pokearu, line 498: Lets not print the whole conf.hardware list. Lets just print the names of the hw?

internal/test/e2e/run.go
```diff
@@ -150,18 +163,26 @@ type instanceRunConf struct {
...
 }
 
-func RunTests(conf instanceRunConf) (testInstanceID string, testCommandResult *testCommandResult, err error) {
+//nolint:gocyclo, revive // RunTests responsible launching test runner to run tests is complex.
+func RunTests(conf instanceRunConf, hardwareQueue *hardwareQueue) (testInstanceID string, testCommandResult *testCommandResult, err error) {
 	testRunner, err := newTestRunner(conf.testRunnerType, conf.testRunnerConfig)
 	if err != nil {
 		return "", nil, err
 	}
+	if conf.hardwareCount > 0 {
```
pokearu, line 180: I guess this is kinda checking if its a tinkerbell test. But just by looking at this If its hard to know that. Can we have a method that does isTinkerbellTest() and that can internally check this?

internal/test/e2e/run.go
```diff
@@ -195,6 +216,11 @@ func RunTests(conf instanceRunConf) (testInstanceID string, testCommandResult *t
 		return session.instanceId, nil, err
 	}
 
+	// Release hardware back to inventory for Tink Tests
+	if conf.hardwareCount > 0 {
+		releaseTinkerbellHardware(conf, hardwareQueue)
```
pokearu, line 221: can we defer this after reserve so we always release even if there are errors.

internal/test/e2e/tinkerbellHardwareQueue.go
rahulbabu95: defer unlock wouldn't work here as this is a polling function that has to periodically poll to check for hardware availability. We do not want to wait until the timeout to unlock as there could be tests that need lesser number hardware waiting for the lock.?

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+// GetTinkerbellHardwareMap returns a map of tink tests with associated hardware count.
+func GetTinkerbellHardwareMap() (map[string]int, error) {
+	tinkerbellHardwareMap, err := readTinkerbellTestFile()
+	if err != nil {
+		return nil, fmt.Errorf("error getting hardware count for tinkerbell tests: %v", err)
+	}
+	return tinkerbellHardwareMap, nil
+}
+
+func readTinkerbellTestFile() (map[string]int, error) {
```
chrisdoherty4, line 28: Can the body of this function be in `GetTinkerbellHardwareMap()`?

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+type TinkerbellTest struct {
+	Name  string `yaml:"name"`
+	Count int    `yaml:"count"`
+}
+
+// GetTinkerbellHardwareMap returns a map of tink tests with associated hardware count.
+func GetTinkerbellHardwareMap() (map[string]int, error) {
+	tinkerbellHardwareMap, err := readTinkerbellTestFile()
+	if err != nil {
+		return nil, fmt.Errorf("error getting hardware count for tinkerbell tests: %v", err)
```
chrisdoherty4, line 23:
```suggestion
		return nil, fmt.Errorf("get hardware count: %v", err)
```

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return tinkerbellHardwareMap, nil
+}
+
+func readTinkerbellTestFile() (map[string]int, error) {
+	tinkerbellTests := make(map[string][]TinkerbellTest)
+	err := yaml.Unmarshal([]byte(tinkerbellHardwareCountFile), &tinkerbellTests)
+	if err != nil {
+		return nil, fmt.Errorf("unable to Unmarshal %s file: %v", TinkerbellHardwareCountFile, err)
```
chrisdoherty4, line 32: The error is a little misleading because this isn't a file, its embedded data.

```suggestion
		return nil, fmt.Errorf("unable to unmarshal hardware yaml: %v", TinkerbellHardwareCountFile, err)
```

I'm curious why we're embedding this to? It makes things quite rigid as configuration can't live separately from this code (i.e. is required at compile time).

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+	tinkerbellHardwareMap, err := readTinkerbellTestFile()
+	if err != nil {
+		return nil, fmt.Errorf("error getting hardware count for tinkerbell tests: %v", err)
+	}
+	return tinkerbellHardwareMap, nil
+}
+
+func readTinkerbellTestFile() (map[string]int, error) {
+	tinkerbellTests := make(map[string][]TinkerbellTest)
+	err := yaml.Unmarshal([]byte(tinkerbellHardwareCountFile), &tinkerbellTests)
```
chrisdoherty4, line 30: Embedding the file as type `[]byte` will avoid the conversion to `[]byte`.

chrisdoherty4, line 30: Embedding the file as type `[]byte` will avoid the conversion to `[]byte`.

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to Unmarshal %s file: %v", TinkerbellHardwareCountFile, err)
+	}
+	tinkerbellTestsHardwareCount, ok := tinkerbellTests[tinkerbellTestsHardwareCountIdentifier]
+	if !ok {
+		return nil, fmt.Errorf("error in type assertion: %v", tinkerbellTests[tinkerbellTestsHardwareCountIdentifier])
+	}
```
chrisdoherty4, line 37: I think we can format the YAML a little differently so the identifier isn't necessary.

If we want something really simple we could use `map[string]int`. Tests require unique names so we should never have duplicate map keys.

```
TestTinkerbellKubernetes123AWSIamAuth: 2
TestTinkerbellKubernetes127AWSIamAuth: 1
...
```

If we want to extensibility per test then `[]map[string]string`.

```
- name: TestTinkerbellKubernetes123AWSIamAuth
  count: 2
- name: TestTinkerbellKubernetes127AWSIamAuth
  count: 1
...
```

test/e2e/tinkerbellhardwarecount.go
chrisdoherty4, line 1: File names should be underscore separated words: `tinkerbell_hardware_count.go`

internal/test/e2e/tinkerbellHardwareQueue.go
chrisdoherty4, line 1: File names should be underscore separated words: `tinkerbell_hardware_queue.go`

internal/test/e2e/tinkerbellHardwareQueue.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Default timeout for Tink tests to poll for hardware.
+const hwPollingTimeout = 120 * time.Minute
+
+type hardwareQueue struct {
+	hws []*api.Hardware
+	mu  sync.Mutex
+}
+
+func (hwQu *hardwareQueue) reserveHardware(count int) ([]*api.Hardware, error) {
```
chrisdoherty4, line 19: What happens if, in a hypothetical world, hardware never becomes available? How does this gracefully terminate?

chrisdoherty4, line 19: Callers have no way of terminating this process. Here's an alternative:

```go
func (q *hardwareQueue) ReserveHardware(ctx context.Context, count int) []*api.Hardware, error) {
  ctx, cancel := context.WithTimeout(ctx, hwPollingTimeout)
  defer cancel()

  ticker := time.NewTicker(1 * time.Minute)
  defer ticker.Stop()

  for {
    select{
    case <-ctx.Done():
      return nil, ctx.Err()

    case <-ticker.C:
      // No timeout logic, only reservation logic.
    }
  }
}
```

While the ticker approach is OK because the ticks are 1m apart, an alternative approach is to use conditions with a notification being published every time hardware is enqueued. That way, there's no need to wait up-to 1m before returning reserved hardware.

internal/test/e2e/tinkerbellHardwareQueue.go
```diff
@@ -0,0 +1,58 @@
...
+			hardwareReserved := hwQu.dequeueHw(count)
+			hwQu.mu.Unlock()
+			return hardwareReserved, nil
+		}
+		hwQu.mu.Unlock()
+		time.Sleep(1 * time.Minute)
+	}
+}
+
+func (hwQu *hardwareQueue) releaseHardware(hardwareToRelease []*api.Hardware) {
```
chrisdoherty4, line 37:
```suggestion
func (hwQu *hardwareQueue) releaseHardware(hw []*api.Hardware) {
```

What's the rationale for the separate `enqueueHw` and `releaseHardware` methods?

internal/test/e2e/tinkerbellHardwareQueue.go
```diff
@@ -0,0 +1,58 @@
...
+	"sync"
+	"time"
+
+	"github.com/aws/eks-anywhere/internal/pkg/api"
+)
+
+// Default timeout for Tink tests to poll for hardware.
+const hwPollingTimeout = 120 * time.Minute
+
+type hardwareQueue struct {
```
chrisdoherty4, line 14: I'm not sure a queue is the right way to frame this; it feels like it could be a semaphore. The order in which hardware is returned seems irrelevant. What inspired modeling this as a queue?

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+var tinkerbellHardwareCountFile string
+
+// TinkerbellTest maps each Tinkbell test with the hardware count needed for the test.
+type TinkerbellTest struct {
+	Name  string `yaml:"name"`
+	Count int    `yaml:"count"`
+}
+
+// GetTinkerbellHardwareMap returns a map of tink tests with associated hardware count.
+func GetTinkerbellHardwareMap() (map[string]int, error) {
```
chrisdoherty4, line 20:
```suggestion
func GetTinkerbellHardwareRequirements() (map[string]int, error) {
```

test/e2e/tinkerbellhardwarecount.go
```diff
@@ -0,0 +1,43 @@
...
+//go:embed tinkerbell_hardware_count.yaml
+var tinkerbellHardwareCountFile string
+
+// TinkerbellTest maps each Tinkbell test with the hardware count needed for the test.
+type TinkerbellTest struct {
+	Name  string `yaml:"name"`
+	Count int    `yaml:"count"`
+}
+
+// GetTinkerbellHardwareMap returns a map of tink tests with associated hardware count.
```
chrisdoherty4, line 19:
```suggestion
// GetTinkerbellHardwareMap returns a map of Tinkerbell test name => required hardware.
```

internal/test/e2e/run.go
```diff
@@ -150,18 +163,28 @@ type instanceRunConf struct {
...
+	hardwareCount                                                             int
 	bundlesOverride                                                           bool
 	testRunnerType                                                            TestRunnerType
 	testRunnerConfig                                                          TestInfraConfig
 	cleanupVms                                                                bool
 	logger                                                                    logr.Logger
 }
 
-func RunTests(conf instanceRunConf) (testInstanceID string, testCommandResult *testCommandResult, err error) {
+//nolint:gocyclo, revive // RunTests responsible launching test runner to run tests is complex.
```
chrisdoherty4, line 174: Please can you raise an issue for this to make it visible.

internal/test/e2e/run.go
```diff
@@ -482,3 +468,43 @@ func logTestGroups(logger logr.Logger, instancesConf []instanceRunConf) {
...
+	invQueue.releaseHardware(conf.hardware)
+	logTinkerbellTestHardwareInfo(conf, "Released")
+}
+
+func logTinkerbellTestHardwareInfo(conf instanceRunConf, action string) {
+	var hardwareInfo []string
+	for _, hardware := range conf.hardware {
+		hardwareInfo = append(hardwareInfo, hardware.Hostname)
+	}
+	conf.logger.V(1).Info(action+" hardware for TestRunner: ", "instanceId", conf.instanceId, "hardwarePool", strings.Join(hardwareInfo, ", "))
```
chrisdoherty4, line 509: Is something missing after the `:`?

internal/test/e2e/run.go
```diff
@@ -482,3 +468,43 @@ func logTestGroups(logger logr.Logger, instancesConf []instanceRunConf) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Tinkerbell hardware: %v", err)
+	}
+	return hardware, nil
+}
+
+func reserveTinkerbellHardware(conf instanceRunConf, invQueue *hardwareQueue) error {
+	reservedTinkerbellHardware, err := invQueue.reserveHardware(conf.hardwareCount)
+	if err != nil {
+		return fmt.Errorf("TestRunner timed out waiting for hardware - test instanceId: %v", conf.instanceId)
```
chrisdoherty4, line 492: What's `TestRunner`? That might be context this function doesn't have.

chrisdoherty4, line 492: What's `TestRunner`? That might be context this function doesn't have.

```suggestion
		return fmt.Errorf("timed out waiting for hardware: test_instance_id=%v", conf.instanceId)
```

internal/test/e2e/run.go
```diff
@@ -150,18 +163,28 @@ type instanceRunConf struct {
...
 	bundlesOverride                                                           bool
 	testRunnerType                                                            TestRunnerType
 	testRunnerConfig                                                          TestInfraConfig
 	cleanupVms                                                                bool
 	logger                                                                    logr.Logger
 }
 
-func RunTests(conf instanceRunConf) (testInstanceID string, testCommandResult *testCommandResult, err error) {
+//nolint:gocyclo, revive // RunTests responsible launching test runner to run tests is complex.
+func RunTests(conf instanceRunConf, hardwareQueue *hardwareQueue) (testInstanceID string, testCommandResult *testCommandResult, err error) {
```
chrisdoherty4, line 175: Did we consider adding the queue/semaphor to the configuration object?

test/e2e/tinkerbellhardwarecount.go
rahulbabu95: We could. but the idea was that the test runner should only be needing the Hardware configuration (which is the exported function) but not necessarily need to know or worry about the logic for parsing which is done by `readTinkerbellTestFile()` within the e2e package.

test/e2e/tinkerbellhardwarecount.go
rahulbabu95: Yes, I did think about this. I went with this approach at first. But to have some conformance with the existing Yaml like `skipped_tests.yaml`, I decided to add an Identifier in place just to make the yaml more readable.

internal/test/e2e/tinkerbellHardwareQueue.go
rahulbabu95: The queue object in itself, just has two methods enqueue and dequeue which any queue is generally expected to have. Since its a concurrent queue, it also has an inherent lock. The locking and unlocking to make the write or enqueuing (thread safe) is handled by `realeaseHardware` . This is just for modularity and separating out the queue functionality from our reservation logic.

internal/test/e2e/run.go
rahulbabu95: Test runner is the `instanceRunConf` that is being passed as the first parameter to this function. So the context is there but I think it need not be included in the error message.!

internal/test/e2e/run.go
rahulbabu95: Yes we did. But did not think that was a great idea since the configuration object already has the tinkerbell hardware and count fields needed. Since the configuration object should ideally be provider agnostic (but already has many fields particular to Tinkerbell) thought of separating it form the configuration object!

internal/test/e2e/tinkerbellHardwareQueue.go
rahulbabu95: At first I thought of having a context but didn't see it adding any additional value for two reasons. One we are just dealing with a hard timeout (in which contexts can be useful if passed around to multiple functions)  and not any additional conditions. The entire file doesn't have context in any of the function signatures, so it doesn't add any interoperability. Also, the existing function returns the hardware as soon as it is reserved and doesn't wait up to a minute to return it! Since we are not making a request to an external resource like say a DB or server didn't find extra value in introducing contexts here!

test/e2e/tinkerbellhardwarecount.go
rahulbabu95: great question! Our E2E tests, when they are run we just copy the `e2e.test` binary to CodeBuild which means any additional files needed should be passed explicitly. since we are planning to establish this file as a requirement for Tinkerbell tests, embedding this avoid passing the file everytime explicitly (like we do presently for skipped_tests) to Codebuild during test runs.

internal/test/e2e/tinkerbellHardwareQueue.go
rahulbabu95: We needed a in-memory data structure that is simple to implement and could provide a mechanism for hardware reservation. Although order doesn't really matter it could provide us extensibility if in future we need to have some control over the order in which these tests reserve the order! Also, there isn't necessarily any downside to keeping it as a queue for the current use case .?

test/e2e/tinkerbellhardwarecount.go
chrisdoherty4: OK. I'm not sure I subscribe to the rationale but we can change later if needed.

test/e2e/tinkerbellhardwarecount.go
chrisdoherty4: I don't think we need to follow suit with `skipped_tests.yaml`; the configurations serve different purposes.

test/e2e/tinkerbellhardwarecount.go
chrisdoherty4: After talking this through and understanding a little more, it turns out the process is shared between all provider test runs so panicking would cause everything to fail. We'll continue returning the error.

internal/test/e2e/run.go
chrisdoherty4: The config object isn't provider agnostic though, and already contains functional objects like `Session`. Having a separate parameter disassociates the configuration and, from a future maintainer perspective, raises the question "Should I add it to the config object or as a parameter".

rahulbabu95: I agree it isn't but ideally it should be since its responsible for launching all tests. And for the questions, I think it's a valid thought consideration to have and not over populate the configuration object with too many case specific parameters given that the generic functionality of that object to run tests across provider without delving into details for each provider.

internal/test/e2e/tinkerbell_hardware_queue.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+func (hwQu *hardwareCatalogue) releaseHardware(hws []*api.Hardware) {
+	hwQu.mu.Lock()
+	hwQu.acquireHw(hws)
+	hwQu.mu.Unlock()
+}
+
+func (hwQu *hardwareCatalogue) acquireHw(hws []*api.Hardware) {
```
chrisdoherty4, line 44: I think acquire and release are the wrong way round. Acquire should return hardware to you as a consumer and release should put it back.

internal/test/e2e/tinkerbell_hardware_queue.go
```diff
@@ -0,0 +1,59 @@
...
+			hardwareReserved := hwQu.releaseHw(count)
+			hwQu.mu.Unlock()
+			return hardwareReserved, nil
+		}
+		hwQu.mu.Unlock()
+		time.Sleep(1 * time.Minute)
+	}
+}
+
+func (hwQu *hardwareCatalogue) releaseHardware(hws []*api.Hardware) {
```
chrisdoherty4, line 38: With the `release` function, I don't think this is necessary any longer.

internal/test/e2e/tinkerbell_hardware_queue.go
```diff
@@ -0,0 +1,59 @@
...
+	hwQu.mu.Lock()
+	hwQu.acquireHw(hws)
+	hwQu.mu.Unlock()
+}
+
+func (hwQu *hardwareCatalogue) acquireHw(hws []*api.Hardware) {
+	hwQu.hws = append(hwQu.hws, hws...)
+}
+
+func (hwQu *hardwareCatalogue) releaseHw(count int) []*api.Hardware {
```
chrisdoherty4, line 48: Can we combine this with `releaseHardware`?

internal/test/e2e/tinkerbell_hardware_queue.go
rahulbabu95: good catch!

chrisdoherty4 (approved):

pokearu (approved):
/lgtm
/meow

PR #6128, opened 2023-07-06
[WIP] Upgrade vSphere management cluster without creating kind cluster

panktishah26:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/blob/main/designs/kindless-upgrades.md

*Description of changes:*
This PR contains changes of management cluster upgrade without kind cluster operations. The feature is under feature flag. Export EXP_SELF_MANAGED_API_UPGRADE and ExpSelfManagedAPIUpgrade flags to true to enable this feature for vSphere Management cluster. This PR doesn't affect the current flow of workload cluster upgrade process. 

- [ ] All bootstrap cluster operations and adjacent code should be removed. (Since we have not enabled this feature for other providers, the current bootstrap cluster operations are still in the code.)
- [ ] The "upgrade needed" step is not needed and can be removed. (Not included)
- [x] Pausing the EKS-A cluster needs to be removed, however, pausing the Flux controller is still needed.
- [ ] Cilium upgrade needs to be removed from the core components upgrade task, since the controller takes care of this. (Not included - need a few data points)
- [ ] The upgrade task itself needs to be re-implemented. We recommend implementing in outside of the workflow, using the [new workflow tasks](https://github.com/aws/eks-anywhere/blob/main/pkg/workflow/task/workload/create.go) as inspiration. It should follow the following steps:
  - [x] Update EKS-A resources with the user provided user spec using a Kubernetes client. 
  - [x] Watch/pull the cluster status, following the upgrade process. Beware of race conditions/caches and make sure we leverage generation/observedGeneration appropriately.
  - [x] Blocking until the cluster is Ready or timeout is reached.
  - [x] reconcileClusterDefinitions needs to be split and some parts reordered.
  - [x] Creating the new Bundles and EKS-D manifest in the cluster (in case of EKS-A version upgrade) needs to be moved up in the workflow, before core components upgrade. (I haven't moved this up before ore components upgrade, we can discuss this offline)
  - [ ] Updating the EKS-A resources is not needed, since the upgrade task already does it.
  - [x] All the GitOps operations should remain in this task.
- [x] Consider keeping a task to perform a backup of CAPI resources before applying any changes to the EKS-A cluster spec.

*Testing (if applicable):*


pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
```diff
@@ -58,7 +60,7 @@ func (r *CloudStackMachineConfig) ValidateUpdate(old runtime.Object) error {
 		return nil
 	}
 
-	if oldCloudStackMachineConfig.IsManagement() {
+	if oldCloudStackMachineConfig.IsManagement() && !features.IsActive(features.ExperimentalSelfManagedClusterUpgrade()) {
```
g-gaston, line 63: I'm confused by this statement
should it be not negated?

pkg/api/v1alpha1/cluster.go
```diff
@@ -309,6 +310,17 @@ func (c *Cluster) ClearPauseAnnotation() {
 	}
 }
 
+// ExperimentalSelfManagedClusterUpgrade returns true if
+// ExperimentalSelfManagedClusterUpgrade is active, cluster is self managed
+// and the provider is vSphere
+// in future we will remove this method once we have kindless upgrade for all the providers.
+func (c *Cluster) ExperimentalSelfManagedClusterUpgrade() bool {
```
g-gaston, line 317: can we have this method implemented somewhere else? i don't think it belongs here, and even less as cluster method

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -316,20 +316,20 @@ func validateImmutableFieldsCluster(new, old *Cluster) field.ErrorList {
 
 	clusterlog.Info("Cluster config is associated with management cluster", "name", old.Name)
 
-	if !RefSliceEqual(new.Spec.IdentityProviderRefs, old.Spec.IdentityProviderRefs) {
+	if !RefSliceEqual(new.Spec.IdentityProviderRefs, old.Spec.IdentityProviderRefs) || old.ExperimentalSelfManagedClusterUpgrade() {
```
g-gaston, line 319: do we need this here? it seems like the method will return if old.ExperimentalSelfManagedClusterUpgrade() is true on line 314

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
 	return nil
 }
 
+// UpgradeClusterAPI incorporates EKS-A cluster status changes for cluster upgrade process.
+func (c *ClusterManager) UpgradeClusterAPI(ctx context.Context, managementCluster, workloadCluster *types.Cluster, newClusterSpec *cluster.Spec, provider providers.Provider) error {
+	logger.V(3).Info("check generation and observed generation numbers")
```
g-gaston, line 699:
```suggestion
	logger.V(3).Info("Checking generation and observed generation numbers")
```

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
...
+
+	isStatusUpdated := func() error {
+		eksaCluster, err := c.clusterClient.GetEksaCluster(ctx, managementCluster, managementCluster.Name)
+		if err != nil {
+			return fmt.Errorf("getting EKS-A cluster details for getting observed generation during an upgrade %v", err)
+		}
+		observedGeneration := eksaCluster.Status.ObservedGeneration
+		generation := eksaCluster.Generation
```
g-gaston, line 707: why wouldn't we make this part of each condition check?

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
 	return nil
 }
 
+// UpgradeClusterAPI incorporates EKS-A cluster status changes for cluster upgrade process.
+func (c *ClusterManager) UpgradeClusterAPI(ctx context.Context, managementCluster, workloadCluster *types.Cluster, newClusterSpec *cluster.Spec, provider providers.Provider) error {
```
g-gaston, line 698: Could we implement this outside of cluster manager?
I would like to get rid of it eventually, and this new code is gonna stay for a long time.

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
...
+	logger.V(3).Info("check generation and observed generation numbers")
+
+	isStatusUpdated := func() error {
+		eksaCluster, err := c.clusterClient.GetEksaCluster(ctx, managementCluster, managementCluster.Name)
```
g-gaston, line 702: I would prefer if we used a kubernetes.Client here :) with the `.Get` method :)

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
...
+		}
+		observedGeneration := eksaCluster.Status.ObservedGeneration
+		generation := eksaCluster.Generation
+		if observedGeneration != generation {
+			return fmt.Errorf("generation and observed generation fields differ")
+		}
+		return nil
+	}
+
+	timeout := 5 * time.Minute
```
g-gaston, line 714: I think we should respect the no timeouts option here

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
...
+		return fmt.Errorf("retries exhausted waiting for status ObservedGeneration field to be updated: %v", err)
+	}
+
+	logger.V(3).Info("Waiting for control plane to be ready")
+	err := c.clusterClient.WaitForEKSAControlPlaneReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), newClusterSpec.Cluster.Name)
+	if err != nil {
+		return fmt.Errorf("waiting for management cluster control plane to be ready: %v", err)
+	}
+
+	logger.V(3).Info("Waiting for default CNI to be congigured")
```
g-gaston, line 726:
```suggestion
	logger.V(3).Info("Waiting for default CNI to be updated")
```

pkg/clustermanager/cluster_manager.go
```diff
@@ -689,6 +694,56 @@ func (c *ClusterManager) UpgradeCluster(ctx context.Context, managementCluster,
...
+		return fmt.Errorf("waiting for management cluster default CNI to be configured: %v", err)
+	}
+
+	logger.V(3).Info("Waiting for worker nodes to be ready after upgrade")
+	err = c.clusterClient.WaitForEKSAWorkersReady(ctx, managementCluster, c.controlPlaneWaitTimeout.String(), newClusterSpec.Cluster.Name)
+	if err != nil {
+		return fmt.Errorf("waiting for management cluster control plane to be ready: %v", err)
+	}
+
+	logger.V(3).Info("Waiting for cluster to be ready after upgrade")
```
g-gaston, line 738:
```suggestion
	logger.V(3).Info("Waiting for cluster upgrade to be completed")
```

pkg/executables/kubectl.go
```diff
@@ -143,6 +143,31 @@ type capiMachinesResponse struct {
 	Items []clusterv1.Machine
 }
 
+// WaitForEKSAControlPlaneAvailable blocks until the first control plane is available for EKSA cluster.
+func (k *Kubectl) WaitForEKSAControlPlaneAvailable(ctx context.Context, cluster *types.Cluster, timeout string, newClusterName string) error {
+	return k.Wait(ctx, cluster.KubeconfigFile, timeout, "ControlPlaneInitialized", fmt.Sprintf("%s/%s", eksaClusterResourceType, newClusterName), constants.DefaultNamespace)
```
g-gaston, line 148: I would prefer if instead of relying on kubectl, we just implement it ourselves
That way we can check the observed generation on every try

Same for the other conditions
We can just have a `WaitForClusterCondition(ctx, client, cluster, condition)` method

pkg/clustermanager/cluster_manager.go
```diff
@@ -1110,6 +1165,34 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
...
 }
 
+// UpdateEKSAResources applies the eks-a cluster specs (cluster, datacenterconfig, machine configs, etc.), as well as the
+// release bundle to the cluster. For Kindless management cluster upgrade task, we don't need to pause the cluserSpec and
+// datacenterConfig webhook validation.
+func (c *ClusterManager) UpdateEKSAResources(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec,
+	datacenterConfig providers.DatacenterConfig, machineConfigs []providers.MachineConfig,
+) error {
+	if clusterSpec.Cluster.Namespace != "" {
+		if err := c.clusterClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, clusterSpec.Cluster.Namespace); err != nil {
```
g-gaston, line 1175: I don't think we should do this
This should up to the user to do
If you want, we can have a validation that runs before any of this. That said, this is for updates right? so the namespace has to exist already

pkg/clustermanager/cluster_manager.go
```diff
@@ -1110,6 +1165,34 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
...
+	resourcesSpec, err := clustermarshaller.MarshalClusterSpec(clusterSpec, datacenterConfig, machineConfigs)
+	if err != nil {
+		return err
+	}
+	logger.V(4).Info("Applying eksa yaml resources to cluster")
+	logger.V(6).Info(string(resourcesSpec))
+	if err = c.applyResource(ctx, cluster, resourcesSpec); err != nil {
+		return err
+	}
+	return c.ApplyBundles(ctx, clusterSpec, cluster)
```
g-gaston, line 1193: IMO this should be done earlier
Maybe when we do the core components upgrade?

pkg/workflows/upgrade.go
g-gaston, line 1: I thought the idea was to duplicate this workflow and make it exclusive for management clusters, and just branch our at the cmd level based on the flag. Can we do that? That's a way cleaner approach

pkg/api/v1alpha1/cloudstackmachineconfig_webhook.go
panktishah26: If the cluster is a management cluster, it used to return from this line without verification. Thus, I have added one more check if the feature is active it should allow to verify further and not return from here.

pkg/clustermanager/cluster_manager.go
panktishah26: GetEksaCluster used Get method to get the cluster details. https://github.com/aws/eks-anywhere/blob/main/pkg/executables/kubectl.go#L1500C25-L1500C25. Let me know if you want this any different way.

panktishah26:
Closing this PR as this has been reimplemented in several smaller PRs like #6622 #6391 #6388 etc

PR #6135, opened 2023-07-07
Remove runc replace directive in go.mod

chrisdoherty4:
We patched a security vuln in our runc dep using a replace directive because dependencies couldn't build with a proper upgrade.

Our deps have since updated meaning the replace directive isn't required and we can just upgrade.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,10 +96,14 @@ require (
 	github.com/bmc-toolbox/bmclib/v2 v2.0.1-0.20230212165211-bac64498b8ba // indirect
 	github.com/bmc-toolbox/common v0.0.0-20221115135648-0b584f504396 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -210,7 +225,7 @@ replace (
 	// opencontainers/runc patched a CVE in v1.1.5.
 	// Various dependencies both direct and indirect are yet to upgrade runc. All seem to work
 	// with v1.1.5 so we're forcing it with replace.
-	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.5
+	//github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.5
```
g-gaston, line 228: Delete this block?

chrisdoherty4: Oh, thanks. That's what happens when you leave something simple for several days :D.

chrisdoherty4: Fixed.

g-gaston (commented):
/lgtm

PR #6165, opened 2023-07-12
Handle special cases for Cilium upgrades when using API

taneyland:
*Issue #, if available:*

*Description of changes:*
The new Cilium version changes the port number for `envoy-metrics` and `prometheus` but the field names remained the same. This caused "duplicate value" errors when upgrading to the new Cilium version using server-side apply because the merge key for these fields is the port number, not the name. See below:

`"err":"failed to reconcile object /v1, Kind=Service, kube-system/cilium-agent: Service \"cilium-agent\" is invalid: spec.ports[1].name: Duplicate value: \"envoy-metrics\"","errVerbose":"Service \"cilium-agent\" is invalid`

To alleviate this issue, we will use the client.Updatestrategy to update the object to be compatible with the new version of Cilium. We are only doing this for the Cilium Service, DaemonSet, and Deployment since those are the only objects affected. The rest of the objects in the Cilium upgrade manifest will continue to be applied using server-side apply.

*Testing (if applicable):*

Tested with Tilt

### DIFF
--- a/pkg/controller/serverside/reconcile.go
+++ b/pkg/controller/serverside/reconcile.go
@@ -39,3 +39,13 @@ func ReconcileObject(ctx context.Context, c client.Client, obj client.Object) er
 
 	return nil
 }
+
+// UpdateObject updates the existing object during reconciliation.
+// This is intended for special use cases only as the preferred method to reconcile objects is server-side apply.
+func UpdateObject(ctx context.Context, c client.Client, obj client.Object) error {
+	if err := c.Update(ctx, obj, client.FieldOwner(fieldManager)); err != nil {
+		return errors.Wrapf(err, "failed to reconcile object %s, %s/%s", obj.GetObjectKind().GroupVersionKind(), obj.GetNamespace(), obj.GetName())
+	}
+
+	return nil
+}
--- a/pkg/networking/cilium/reconciler/reconciler.go
+++ b/pkg/networking/cilium/reconciler/reconciler.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -25,6 +27,12 @@ import (
 
 const (
 	defaultRequeueTime = time.Second * 10
+)
+
+var (
+	serviceKind    = corev1.SchemeGroupVersion.WithKind("Service").GroupKind()
+	daemonSetKind  = appsv1.SchemeGroupVersion.WithKind("DaemonSet").GroupKind()
+	deploymentKind = appsv1.SchemeGroupVersion.WithKind("Deployment").GroupKind()
 )
 
 type Templater interface {
@@ -203,7 +211,18 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
 	}
 
 	logger.Info("Applying Cilium upgrade manifest")
-	if err := serverside.ReconcileYaml(ctx, client, upgradeManifest); err != nil {
+
+	// When upgrading from Cilium v1.11.x --> Cilium v1.12.x, the port number for a few components changed but the port names remained the same.
+	// This caused "duplicate value" errors when upgrading to the new Cilium version using server-side apply because the merge key for these fields is the port number, not the name.
+	// To alleviate this issue, we will use the client.Update strategy to update the yaml to be compatible with the new version of Cilium.
+	// We are only doing this for the Cilium Service, DaemonSet, and Deployment since those are the only objects affected.
+	// The rest of the objects in the Cilium upgrade manifest will continue to be applied using server-side apply.
+	manifestObjs, err := r.reconcileSpecialCases(ctx, client, upgradeManifest)
+	if err != nil {
+		return controller.Result{}, err
+	}
+
+	if err := serverside.ReconcileObjects(ctx, client, manifestObjs); err != nil {
 		return controller.Result{}, err
 	}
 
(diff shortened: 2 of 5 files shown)


g-gaston:
/hold
we might not need this/it might not solve the problem completely

pkg/controller/serverside/reconcile.go
```diff
@@ -39,3 +39,16 @@ func ReconcileObject(ctx context.Context, c client.Client, obj client.Object) er
 
 	return nil
 }
+
+// ReconcileReplaceObject completely replaces the current object with the new object during reconciliation.
+// This is intended for special use cases only as the preferred method to reconcile objects is server-side apply.
+func ReconcileReplaceObject(ctx context.Context, c client.Client, currentObj, newObj client.Object) error {
+	// MergeFrom creates a Patch that patches using the merge-patch strategy with the given object as base.
+	// When using MergeFrom, existing lists will be completely replaced by new lists.
+	err := c.Patch(ctx, newObj, client.MergeFrom(currentObj))
```
g-gaston, line 48: this is not replacing right? it's doing a merge
I think we just need `client.Update`
and in that case, we probably don't need a helper method

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -203,6 +205,16 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
 	}
 
 	logger.Info("Applying Cilium upgrade manifest")
+
+	// When upgrading from Cilium v1.11.x --> Cilium v1.12.x, the port number for a few components changed but the field names remained the same.
```
g-gaston, line 209:
```suggestion
	// When upgrading from Cilium v1.11.x --> Cilium v1.12.x, the port number for a few components changed but the port name remained the same.
```

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -203,6 +205,16 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
...
+	// This caused "duplicate value" errors when upgrading to the new Cilium version using server-side apply because the merge key for these fields is the port number, not the name.
+	// To alleviate this issue, we will use the client.MergeFrom strategy to completely replace the old object with the new object that is compatible with the new version of Cilium.
```
g-gaston, line 211:
```suggestion
	// To alleviate this issue, we will use the client.Update strategy to completely replace the old object with the new object that is compatible with the new version of Cilium.
```

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -203,6 +205,16 @@ func (r *Reconciler) upgrade(ctx context.Context, logger logr.Logger, client cli
...
+	// We are only doing this for the Cilium Service, DaemonSet, and Deployment since those are the only objects affected.
+	// The rest of the objects in the Cilium upgrade manifest will continue to be applied using server-side apply.
+	if err := r.reconcileSpecialCases(ctx, client, upgradeManifest); err != nil {
+		return controller.Result{}, err
+	}
+
 	if err := serverside.ReconcileYaml(ctx, client, upgradeManifest); err != nil {
 		return controller.Result{}, err
 	}
```
g-gaston, line 220: This is doing the un-marshaling twice
I would first unmarshall, separate the objects into two slices based on the Kind and then reconcile separately.

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -260,3 +272,35 @@ func (r *Reconciler) installPreflight(ctx context.Context, client client.Client,
...
 }
+
+//nolint:gocyclo
+func (r *Reconciler) reconcileSpecialCases(ctx context.Context, c client.Client, yaml []byte) error {
+	objs, err := clientutil.YamlToClientObjects(yaml)
+	if err != nil {
+		return err
+	}
+
+	currentDS, currentDeployment, currentSvc, err := getObjects(ctx, c, cilium.DaemonSetName, cilium.DeploymentName)
```
g-gaston, line 283: do we need this helper?
just trying to simplify, could we just get the 3 objects separately?

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -260,3 +272,35 @@ func (r *Reconciler) installPreflight(ctx context.Context, client client.Client,
...
+	if err != nil {
+		return err
+	}
+
+	for _, o := range objs {
+		if o.GetObjectKind().GroupVersionKind().GroupKind() == corev1.SchemeGroupVersion.WithKind("Service").GroupKind() && o.GetName() == cilium.ServiceName {
```
g-gaston, line 289: should we store this in a var outside the method? `corev1.SchemeGroupVersion.WithKind("Service").GroupKind()`
to avoid recalculating

pkg/networking/cilium/reconciler/reconciler.go
taneyland: We don't need it anymore since we're gonna use Client.Update instead of MergeFrom

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -260,3 +279,25 @@ func (r *Reconciler) installPreflight(ctx context.Context, client client.Client,
 
 	return nil
 }
+
+func (r *Reconciler) reconcileSpecialCases(ctx context.Context, c client.Client, yaml []byte) ([]client.Object, error) {
```
chrisdoherty4, line 283: Presumable this needs removing in a subsequent release: have we an issue to track?

taneyland: No, we don't have one yet but I can create one for us to track

pkg/controller/serverside/reconcile.go
```diff
@@ -39,3 +39,13 @@ func ReconcileObject(ctx context.Context, c client.Client, obj client.Object) er
 
 	return nil
 }
+
+// UpdateObject updates the existing object during reconciliation.
+// This is intended for special use cases only as the preferred method to reconcile objects is server-side apply.
```
chrisdoherty4, line 44: Given this is a special use-case, shouldn't this live in the package with the special use-case?

taneyland: Hmm yeah good point. I put in this file because it serves a similar purpose as the other methods in this file. So even those we only use this method for this one use-case, there isn't anything specific to reconciling Cilium in this logic. Leaving it here would also allow us to use this same method if it's ever needed in other components.

pkg/networking/cilium/reconciler/reconciler.go
chrisdoherty4: Thanks, Taylor. Probably stick a blocker label on it straight away so its clear we need it.

taneyland: For sure! Added the label

PR #6168, opened 2023-07-13
Add DefaultCNI field to Cluster status

cxbrowne1207:
Issue #, if available:
 #5628 

Description of changes:
- Added DefaultCNI field

It also includes a fix for the `DefaultCNIConfigured` condition, where when a workload cluster control plane is upgraded, it is marked "False" with the `ControlPlaneNotReady`. It is not re-evaluated after the control plane becomes ready, as the controller stops reconciling before then. So, this PR addresses the issue by removing the condition update's dependency on `ControlPlaneReady`. We only need to change this status when there is a. change in the state of the CNI detected.

Testing (if applicable):
- Unit testing
- Manual testing

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -893,7 +893,7 @@ type ClusterCNI struct {
 	// +optional
 	Version string `json:"version,omitempty"`
 
-	// Applied, Not Applied
+	// Status reports the current state of the default CNI.
 	Status ClusterCNIStatus `json:"status,omitempty"`
 }
 
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -870,18 +870,18 @@ const (
 	MachineInvalidReason FailureReasonType = "MachineInvalid"
 )
 
-// ClusterCNIStatus is a type for defining statuses for Cluster default cni.
-type ClusterCNIStatus string
+// ClusterCNIState is a type for defining statuses for Cluster default cni.
+type ClusterCNIState string
 
 const (
 	// ClusterCNIInitializing reports that the default CNI is initializing.
-	ClusterCNIInitializing ClusterCNIStatus = "initializing"
+	ClusterCNIInitializing ClusterCNIState = "initializing"
 
 	// ClusterCNIInstalled reports that the default CNI has been installed.
-	ClusterCNIInstalled ClusterCNIStatus = "installed"
+	ClusterCNIInstalled ClusterCNIState = "installed"
 
 	// ClusterCNIUpdating reports that the default CNI is updating.
-	ClusterCNIUpdating ClusterCNIStatus = "updating"
+	ClusterCNIUpdating ClusterCNIState = "updating"
 )
 
 // ClusterCNI represents a CNI cluster component.
@@ -893,8 +893,8 @@ type ClusterCNI struct {
 	// +optional
 	Version string `json:"version,omitempty"`
 
-	// Status reports the current state of the default CNI.
-	Status ClusterCNIStatus `json:"status,omitempty"`
+	// State reports the current state of the default CNI.
+	State ClusterCNIState `json:"state,omitempty"`
 }
 
 // ClusterStatus defines the observed state of Cluster.
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -870,6 +870,33 @@ const (
 	MachineInvalidReason FailureReasonType = "MachineInvalid"
 )
 
+// ClusterCNIStatus is a type for defining statuses for Cluster default cni.
+type ClusterCNIStatus string
+
+const (
+	// ClusterCNIInitializing reports that the default CNI is initializing.
+	ClusterCNIInitializing ClusterCNIStatus = "initializing"
+
+	// ClusterCNIInstalled reports that the default CNI has been installed.
+	ClusterCNIInstalled ClusterCNIStatus = "installed"
+
+	// ClusterCNIUpdating reports that the default CNI is updating.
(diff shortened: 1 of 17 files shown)


pkg/controller/clusters/status.go
```diff
@@ -41,10 +42,11 @@ func UpdateClusterStatusForWorkers(ctx context.Context, client client.Client, cl
 
 // UpdateClusterStatusForCNI updates the Cluster status for the default cni before the control plane is ready. The CNI reconciler
 // handles the rest of the logic for determining the condition and updating the status based on the current state of the cluster.
-func UpdateClusterStatusForCNI(ctx context.Context, cluster *anywherev1.Cluster) {
+func UpdateClusterStatusForCNI(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
```
g-gaston, line 45: There is starting to be too much duplication between this and the cni reconciler. plus we keep bleeding cilium specific logic here.

What do you think about adding a new method to the CNI reconciler interface like  `UpdateClusterStatusForCNI`. And then just call that method from the cluster reconciler where you update the status. That way there is only one way to update the status, regardless of the type of cluster and the scenario.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -870,6 +870,33 @@ const (
 	MachineInvalidReason FailureReasonType = "MachineInvalid"
 )
 
+// ClusterCNIStatus is a type for defining statuses for Cluster default cni.
+type ClusterCNIStatus string
```
g-gaston, line 874:
```suggestion
type ClusterCNIState string
```

nit: to avoid confusion between status objects

pkg/networking/cilium/installation.go
```diff
@@ -25,6 +28,27 @@ type Installation struct {
...
 }
 
+// Version returns the version of the cilium Installation.
+func (i Installation) Version() (string, error) {
+	if i.DaemonSet == nil {
+		return "", nil
+	}
+
+	dsImage := i.DaemonSet.Spec.Template.Spec.Containers[0].Image
+	_, dsImageTag := oci.Split(dsImage)
```
g-gaston, line 38: why not just return dsImageTag here?

pkg/controller/clusters/status.go
```diff
@@ -210,3 +226,12 @@ func updateWorkersReadyCondition(cluster *anywherev1.Cluster, machineDeployments
 func controlPlaneInitializationInProgressCondition() *anywherev1.Condition {
 	return conditions.FalseCondition(anywherev1.ControlPlaneInitializedCondition, anywherev1.ControlPlaneInitializationInProgressReason, clusterv1.ConditionSeverityInfo, "The first control plane instance is not available yet")
 }
+
+// SetClusterDefaultCNI sets the DefaultCNI for the Cluster status.
+func SetClusterDefaultCNI(cluster *anywherev1.Cluster, status anywherev1.ClusterCNIStatus, version string) {
```
g-gaston, line 231: do we need a method for this?

pkg/controller/clusters/status.go
cxbrowne1207: Sounds good. I'll move this method from here and add a new method to the CNI reconciler interface to avoid bleeding cilium specific logic here.

Can you clarify what your thoughts on handling the duplication is exactly? The self managed clusters don't use the cni reconciler and the status updates in the cni reconciler are integrated into the reconcile process. So it would be hard to remove duplication (even after the move), at least without changing that.

pkg/networking/cilium/installation.go
cxbrowne1207: ah, yeah that makes things much simpler.

pkg/controller/clusters/status.go
cxbrowne1207: it can be removed, and we could set the field directly

pkg/controller/clusters/status.go
g-gaston: can't this new method set the information for both workload and management clusters?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -870,6 +870,33 @@ const (
...
+type ClusterCNI struct {
+	// Name of the default CNI
+	Name string `json:"name,omitempty"`
+
+	// Empty if not applied or no version can be inferred.
+	// +optional
+	Version string `json:"version,omitempty"`
+
+	// State reports the current state of the default CNI.
+	State ClusterCNIState `json:"state,omitempty"`
```
g-gaston, line 897: is this really optional?

pkg/networking/cilium/reconciler/reconciler.go
```diff
@@ -128,10 +148,60 @@ func (r *Reconciler) Reconcile(ctx context.Context, logger logr.Logger, client c
...
+		State:   anywherev1.ClusterCNIInstalled,
+	}
 
 	return r.deletePreflightIfExists(ctx, client, spec)
 }
 
+// UpdateClusterStatusForCNI updates the Cluster status for the default cni before the control plane is ready. The CNI reconciler
+// handles the rest of the logic for determining the condition and updating the status based on the current state of the cluster.
+func (r *Reconciler) UpdateClusterStatusForCNI(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
+	if conditions.Get(cluster, anywherev1.DefaultCNIConfiguredCondition) == nil {
```
g-gaston, line 163: can we have a comment to explain why this block does what it does?

pkg/api/v1alpha1/cluster_types.go
cxbrowne1207: it's not supposed to be optional. are you referring to the omitempty tag? that can be removed

cxbrowne1207:
Closing, we've decided to shelf this for now and come back to it at a later date

Issue #6039, opened 2023-06-16
[bare metal] cli panic when machine ref to cp node doesnt match/exist


jacobweinstock:
**What happened**:
The cli panics under the following scenario. The `kind: Cluster` object, `spec.controlPlaneConfiguration.machineGroupRef.name` is set to a value that doesn't exist/doesn't match a `TinkerbellMachineConfig` in the `mycluster.yaml`.

```bash
eksctl anywhere create cluster mycluster -f mycluster.yaml --unhealthy-machine-timeout 10m --tinkerbell-bootstrap-ip <IP> -z hardware.csv
```

```bash
Performing setup and validations
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2bdf668]

goroutine 1 [running]:
github.com/aws/eks-anywhere/pkg/providers/tinkerbell.(*Provider).MachineConfigs(0xc00043e580, 0x3c30d50?)
	github.com/aws/eks-anywhere/pkg/providers/tinkerbell/tinkerbell.go:266 +0x168
github.com/aws/eks-anywhere/pkg/providers/tinkerbell.(*Provider).SetupAndValidateCreateCluster(0xc00043e580, {0x3c30d50, 0xc000052078}, 0xc000b78040)
	github.com/aws/eks-anywhere/pkg/providers/tinkerbell/create.go:124 +0x1c5
github.com/aws/eks-anywhere/pkg/workflows.(*SetAndValidateTask).providerValidation.func1()
	github.com/aws/eks-anywhere/pkg/workflows/create.go:190 +0xb2
github.com/aws/eks-anywhere/pkg/validations.(*Runner).Run(0xc000dcdb00?)
	github.com/aws/eks-anywhere/pkg/validations/runner.go:24 +0x42
github.com/aws/eks-anywhere/pkg/workflows.(*SetAndValidateTask).Run(0xc000d9f2c0?, {0x3c30d50, 0xc000052078}, 0xc000d9d9e0)
	github.com/aws/eks-anywhere/pkg/workflows/create.go:177 +0x365
github.com/aws/eks-anywhere/pkg/task.(*taskRunner).RunTask(0xc000e085a0, {0x3c30d50, 0xc000052078}, 0xc000d9d9e0)
	github.com/aws/eks-anywhere/pkg/task/task.go:159 +0x463
[...]

PR #6171, opened 2023-07-13
check existence of machine config before further operations

HiteshRepo:
*Issue #, if available:*
#6039 

*Description of changes:*
check existence of machine config before further operations


HiteshRepo:
@chrisdoherty4 could you please take a look at this?

chrisdoherty4:
@HiteshRepo Apologies for the slowness in the last few hectic weeks. We'll take a look shortly. 

/cc chrisdoherty4 jacobweinstock

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -263,19 +263,23 @@ func (p *Provider) DatacenterConfig(_ *cluster.Spec) providers.DatacenterConfig
 func (p *Provider) MachineConfigs(_ *cluster.Spec) []providers.MachineConfig {
 	configs := make(map[string]providers.MachineConfig, len(p.machineConfigs))
 	controlPlaneMachineName := p.clusterConfig.Spec.ControlPlaneConfiguration.MachineGroupRef.Name
-	p.machineConfigs[controlPlaneMachineName].Annotations = map[string]string{p.clusterConfig.ControlPlaneAnnotation(): "true"}
-	if p.clusterConfig.IsManaged() {
-		p.machineConfigs[controlPlaneMachineName].SetManagedBy(p.clusterConfig.ManagedBy())
+	if _, ok := p.machineConfigs[controlPlaneMachineName]; ok {
```
chrisdoherty4, line 266: I don't think we want to silently ignore the fact a machine config is missing. I think we can prevent this from ever happening by validating all machine configs are present in `NewProvider()`.

Specifically, on [L117](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/tinkerbell/tinkerbell.go#L117) we `TinkebrellMachineConfig` spec but we don't handle it not being present.

chrisdoherty4 (changes requested):
I don't think we want to silently ignore the fact a machine config is missing. I think we can prevent this from ever happening by validating all machine configs are present in NewProvider().

Specifically, on L117 we TinkebrellMachineConfig spec but we don't handle it not being present.

pkg/providers/tinkerbell/tinkerbell.go
```diff
@@ -263,19 +270,23 @@ func (p *Provider) DatacenterConfig(_ *cluster.Spec) providers.DatacenterConfig
...
```
chrisdoherty4, line 273: I think we can revert this change. If these machine configs are missing, there's a catastrophic problem and silently ignoring the fact may result in side effects with processes later on.

The checks added in `NewProvider` should be sufficient to protect against this issue now.

chrisdoherty4 (changes requested):
@HiteshRepo thanks again. I've added 1 final comment regarding reverting the original changes.

chrisdoherty4:
Looking great @HiteshRepo, thank you for your submissions.

chrisdoherty4 (approved):

PR #6174, opened 2023-07-13
Use sha256 for eks-anywhere-package tag

d8660091:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/release/pkg/helm/helm.go
+++ b/release/pkg/helm/helm.go
@@ -139,12 +139,6 @@ func ModifyAndPushChartYaml(i releasetypes.ImageArtifact, r *releasetypes.Releas
 
 	// If the chart is packages, we find the image tag values and overide them in the values.yaml.
 	if strings.Contains(helmDest, "eks-anywhere-packages") {
-		imageTagMap, err := GetPackagesImageTags(eksArtifacts)
-		fmt.Printf("Overwriting helm values.yaml version to new image tags %v\n", imageTagMap)
-		err = OverWriteChartValuesImageTag(helmDest, imageTagMap)
-		if err != nil {
-			return fmt.Errorf("overwriting the values.yaml version: %w", err)
-		}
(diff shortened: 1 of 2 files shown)


release/pkg/helm/helm.go
```diff
@@ -454,7 +448,9 @@ func OverWriteChartValuesImageSha(filename string, shaMap map[string]anywherev1a
 		return err
 	}
 	values["controller"].(map[string]interface{})["digest"] = shaMap["eks-anywhere-packages"].ImageDigest
+	values["controller"].(map[string]interface{})["tag"] = shaMap["eks-anywhere-packages"].ImageDigest
```
jaxesn, line 451: before modifying this helm chart, can we check if the varlues are even different and if they are not, dont do this at all?

also, if we do push a new helm chart, i think the sha in the bundle should reflect this modified and pushed chart vs what it is now, the original chart.

d8660091: I did one check, versions in bundle-release.yaml and chart's values.yaml are matching, but we cannot induce that all bundle-release.yaml are good. There might be some good reason why this modification exists.

Right, we need to use the new digest in dev-bundle-release.yaml's packages. I can do it in a separate PR.

d8660091: In fact, let's try removing the modification

jaxesn: No way this is for all charts? I don't really follow, all I'm saying is if the digest it's about to set is what's already in the values yaml for the chart sr pushed there is no reason to update

We do need to check with Jonah and try to understand what issue was being solved by making them tags instead of digests, but if we want to support airgapped I think ultimately we need to use digests here. 

I also wonder if we should change the cli to pull the helm chart being sha instead of tag to ensure we are installing the same one referenced in the bundle.

PR #6176, opened 2023-07-14
Configure machine health checks for cluster controller

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
This PR configures machine health checks for cluster controller while creating workload clusters. It also changes to the way we install machine health checks using the CLI. A new section is added to the eksa spec that let's users configure machine health checks. The newly added spec will be used as the source of truth from the upcoming releases. We will support the flags for these timeouts for the next couple releases, but will drop support after that.

```
kind: Cluster
metadata:
  name: docker126
spec:
  clusterNetwork:
  .
  .
  .
  workerNodeGroupConfigurations:
  - count: 1
    name: md-0
  machineHealthChecks:
    nodeStartupTimeout: "10m0s"
    unhealthyMachineTimeout: "5m0s"
```

Expected behavior from the following cases:
- new cluster in cli: If the eksa spec has values configured for the machine health check (MHC) by the users, the cluster will be created with those timeouts. If no values are specified, the defaults will be set  which installing machine health checks which are 10m0s for the nodeStartupTimeout and 5m0s for the unhealthyMachineTimeout.
- existing cluster in upgrade cli - While upgrading a cluster with no machine health check configured by users in the spec, we fetch the MHCs set on the cluster and use those values to populate the eksa cluster spec.
- new cluster in controller
- existing cluster in controller

*Testing (if applicable):*
- test creation on management cluster
- creation for workload clusters using tilt and CLI
- new cluster in cli
- existing cluster in upgrade cli
- new cluster in controller
- existing cluster in controller


pkg/cluster/defaults.go
```diff
@@ -28,3 +31,42 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+// NewMachineHealthCheckDefaulter allows to create a new MachineHealthCheckDefaulter.
+func NewMachineHealthCheckDefaulter(nodeStartupTimeout, unhealthyMachineTimeout string) MachineHealthCheckDefaulter {
+	return MachineHealthCheckDefaulter{
+		nodeStartupTimeout:      nodeStartupTimeout,
+		unhealthyMachineTimeout: unhealthyMachineTimeout,
+	}
+}
+
+// MachineHealthCheckDefault sets the defaults for machine health check timeouts.
+func (d MachineHealthCheckDefaulter) MachineHealthCheckDefault(ctx context.Context, spec *Spec) (*Spec, error) {
```
vivek-koppuru, line 50: We don't need an error here then

pkg/clustermanager/cluster_manager_test.go
```diff
@@ -1827,19 +1835,23 @@ func TestInstallMachineHealthChecksWithTimeoutOverride(t *testing.T) {
 	}
 }
 
-func TestInstallMachineHealthChecksWithNoTimeout(t *testing.T) {
-	tt := newTest(t, clustermanager.WithNoTimeouts())
-	tt.clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations[0].Name = "worker-1"
-	wantMHC := expectedMachineHealthCheck(maxTime, maxTime)
-	tt.mocks.client.EXPECT().ApplyKubeSpecFromBytes(tt.ctx, tt.cluster, wantMHC)
+// func TestInstallMachineHealthChecksWithNoTimeout(t *testing.T) {
```
vivek-koppuru, line 1838: Is this meant to be commented out?

pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -79,3 +80,27 @@ func setCNIConfigDefault(cluster *Cluster) error {
 	cluster.Spec.ClusterNetwork.CNI = ""
 	return nil
 }
+
+func setMachineHealthCheckTimeoutDefaults(cluster *Cluster) error {
+	if cluster.Spec.MachineHealthCheck != nil && len(cluster.Spec.MachineHealthCheck.NodeStartupTimeout) != 0 && len(cluster.Spec.MachineHealthCheck.UnhealthyMachineTimeout) != 0 {
```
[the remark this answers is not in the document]
mitalipaygude: yeah so this check is for validating that if the machine health check and timeouts are set in the cluster spec itself, no defaults need to be set.

pkg/api/v1alpha1/cluster_defaults.go
```diff
@@ -79,3 +80,27 @@ func setCNIConfigDefault(cluster *Cluster) error {
 	cluster.Spec.ClusterNetwork.CNI = ""
 	return nil
 }
+
+func setMachineHealthCheckTimeoutDefaults(cluster *Cluster) error {
```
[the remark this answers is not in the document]
mitalipaygude: so this method needs to have the same signature as the other defaults from line 12, which is why I added it for this one.

pkg/cluster/defaults.go
mitalipaygude: here as well since the defaulter needs the signature that returns an error as well as the spec, I made it compliant.

pkg/clustermanager/cluster_manager_test.go
mitalipaygude: no this one i am working on to figure out the approach.

pkg/api/v1alpha1/cluster_defaults.go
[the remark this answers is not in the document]
vivek-koppuru: Oh right hmm okay

pkg/api/v1alpha1/cluster.go
```diff
@@ -177,6 +177,10 @@ func NewCluster(clusterName string) *Cluster {
 		},
 		Spec: ClusterSpec{
 			KubernetesVersion: Kube119,
```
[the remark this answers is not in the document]
vivek-koppuru: Can we update it now then since we noticed it? I think this would be the only place anyways right? Is this used for unit testing?

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,42 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+
+	if spec.Config.Cluster.Spec.MachineHealthCheck == nil {
+		spec.Config.Cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{}
+	}
+
+	if len(spec.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout) == 0 {
+		if spec.Cluster.Spec.DatacenterRef.Kind == anywherev1.TinkerbellDatacenterKind && d.nodeStartupTimeout == constants.DefaultNodeStartupTimeout.String() {
+			spec.Config.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout = constants.DefaultTinkerbellNodeStartupTimeout.String()
+		} else {
+			spec.Config.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout = d.nodeStartupTimeout
```
vivek-koppuru, line 63: So we only set default for Tinkerbell? What will this be for the other providers, 0?

pkg/clustermanager/cluster_manager.go
```diff
@@ -777,7 +775,17 @@ func getProviderNamespaces(providerDeployments map[string][]string) []string {
...
 func (c *ClusterManager) InstallMachineHealthChecks(ctx context.Context, clusterSpec *cluster.Spec, workloadCluster *types.Cluster) error {
-	mhc, err := templater.ObjectsToYaml(kubernetes.ObjectsToRuntimeObjects(clusterapi.MachineHealthCheckObjects(clusterSpec.Cluster, c.unhealthyMachineTimeout, c.nodeStartupTimeout))...)
+	unhealthyMachineTimeout, err := time.ParseDuration(clusterSpec.Cluster.Spec.MachineHealthCheck.UnhealthyMachineTimeout)
+	if err != nil {
+		return err
+	}
+
+	nodeStartupTimeout, err := time.ParseDuration(clusterSpec.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout)
+	if err != nil {
+		return err
```
vivek-koppuru, line 785: Seems like we missed a test case for the second one?

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+
+		// check if machine health checks have already been installed on the cluster and populate the EKSA spec
+		mhc, err := r.machineHealthChecksExists(ctx, log, cluster)
+		if err != nil {
+			return err
+		}
+		if mhc != nil {
+			cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{
+				NodeStartupTimeout:      mhc.Spec.NodeStartupTimeout.OpenAPISchemaFormat(),
+				UnhealthyMachineTimeout: mhc.Spec.UnhealthyConditions[0].Timeout.String(),
```
vivek-koppuru, line 50: Do we need to have a null check for the array here?

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "fetching machine health check %s", kcpMHCName)
+	}
+
+	return mhc, nil
+}
+
+// setMachineHealthCheckTimeoutDefaults sets default tiemout values for cluster's machine health checks.
+func setMachineHealthCheckTimeoutDefaults(cluster *anywherev1.Cluster) {
+	if len(cluster.Spec.MachineHealthCheck.NodeStartupTimeout) == 0 {
```
vivek-koppuru, line 98: This is another place we are running the defaults? I saw some defaulting above, but not sure if that was for only CLI or something.

pkg/cluster/defaults.go
mitalipaygude: the default for others for node startup is 10m. that gets set for the `d.nodeStartupTimeout`

pkg/machinehealthcheck/reconciler.go
mitalipaygude: yes the cluster defaulter only sets defaults for the CLI. the ones in here set it for the controller.

pkg/machinehealthcheck/reconciler.go
mitalipaygude: we don't really need a null check since the machine health check always populates 2 conditions, but i don't mind adding that in if you have a preference.

pkg/machinehealthcheck/reconciler.go
vivek-koppuru: Can we combine this defaulting logic in one place instead of having it defined in before? That is in case we decide to change the default values in the future?

pkg/machinehealthcheck/reconciler.go
vivek-koppuru: I just want to make sure we don't run into exceptions, if machine health check dynamically populates then we should have the check, but it does make an assumption that the creator does that which can mess things up later when the assumption changes.

pkg/cluster/defaults.go
vivek-koppuru: In that case, any reason we aren't doing `d.nodeStartupTimeout = constants.DefaultTinkerbellNodeStartupTimeout.String()` and then not having an else condition below?

mitalipaygude: The `d.nodeStartupTimeout` is set in the factory based on if the flag is set or not in cluster Options.

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,53 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+
+	SetMachineHealthCheckTimeoutDefaults(spec.Cluster, d.nodeStartupTimeout, d.unhealthyMachineTimeout)
+
+	return spec, nil
+}
+
+// SetMachineHealthCheckTimeoutDefaults sets default tiemout values for cluster's machine health checks.
+func SetMachineHealthCheckTimeoutDefaults(cluster *anywherev1.Cluster, nodeStartupTimeout, unhealthyMachineTimeout string) {
+	if len(cluster.Spec.MachineHealthCheck.NodeStartupTimeout) == 0 {
+		if cluster.Spec.DatacenterRef.Kind == anywherev1.TinkerbellDatacenterKind {
```
vivek-koppuru, line 73: We don't need this anymore then right? We are already doing this above. Otherwise we should remove it above?

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/aws/eks-anywhere/pkg/controller/serverside"
+)
+
+// Reconciler allows to reconcile machine health checks.
+type Reconciler struct {
+	client client.Client
+}
+
+// New returns a new Reconciler.
+func New(client client.Client) *Reconciler {
```
vivek-koppuru, line 29: Can we initialize the defaulter here through the factory? I think that's what we do for cli right? And then we can just use the same defaulter instead of calling the exported method below.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -275,6 +276,14 @@ type ControlPlaneConfiguration struct {
 	SkipLoadBalancerDeployment bool `json:"skipLoadBalancerDeployment,omitempty"`
 }
 
+// MachineHealthCheck allows to configure timeouts for machine health checks.
+type MachineHealthCheck struct {
+	// NodeStartupTimeout overrides the the default node startup timeout in machine health checks.
+	NodeStartupTimeout string `json:"nodeStartupTimeout,omitempty"`
```
g-gaston, line 282: Should we use [`metav1.Duration`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Duration)?
I also, I think these two should probably be pointers, right? Because they are both optional and independent from each other.

pkg/cli/createclusterdefaulter.go
```diff
@@ -13,10 +13,11 @@ type CreateClusterDefaulter struct {
 }
 
 // NewCreateClusterDefaulter to instantiate and register defaults.
-func NewCreateClusterDefaulter(skipIPCheck cluster.ControlPlaneIPCheckAnnotationDefaulter) CreateClusterDefaulter {
+func NewCreateClusterDefaulter(skipIPCheck cluster.ControlPlaneIPCheckAnnotationDefaulter, mhcDefaulter cluster.MachineHealthCheckDefaulter) CreateClusterDefaulter {
```
g-gaston, line 16: nit: no need to do it now, but I suspect the dependencies here will keep growing. Having an "input" struct will make this easier to extend (for example, reduces the amount of changes in unit tests).

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,49 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
 
 	return spec, nil
 }
+
+// MachineHealthCheckDefaulter is the defaulter created to configure the machine health check timeouts.
+type MachineHealthCheckDefaulter struct {
+	nodeStartupTimeout      string
```
g-gaston, line 37: Use duration here?

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,49 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+}
+
+// MachineHealthCheckDefault sets the defaults for machine health check timeouts.
+func (d MachineHealthCheckDefaulter) MachineHealthCheckDefault(ctx context.Context, spec *Spec) (*Spec, error) {
+	if spec.Config.Cluster.Spec.MachineHealthCheck != nil && len(spec.Config.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout) != 0 && len(spec.Config.Cluster.Spec.MachineHealthCheck.UnhealthyMachineTimeout) != 0 {
+		return spec, nil
+	}
+
+	if spec.Config.Cluster.Spec.MachineHealthCheck == nil {
+		spec.Config.Cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{}
```
g-gaston, line 56: we should only do this if any of the timeouts are set in this defaulter right? otherwise it should remain nil

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,49 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+	}
+
+	if len(spec.Cluster.Spec.MachineHealthCheck.NodeStartupTimeout) == 0 {
+		if spec.Cluster.Spec.DatacenterRef.Kind == anywherev1.TinkerbellDatacenterKind && d.nodeStartupTimeout == constants.DefaultNodeStartupTimeout.String() {
```
g-gaston, line 60: sorry if I missed this, but is this new or are we just moving the logic from somewhere else?
Also, I think a comments explaining why this is necessary would be awesome.

pkg/cluster/defaults.go
g-gaston: if we are always going to set a value why just make this field a struct instead of a pointer? Not asking to change, just asking for opinions.

pkg/cluster/defaults.go
```diff
@@ -28,3 +31,49 @@ func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx c
...
+		}
+	}
+
+	setMachineHealthCheckTimeoutDefaults(spec.Cluster, d.nodeStartupTimeout, d.unhealthyMachineTimeout)
+
+	return spec, nil
+}
+
+// setMachineHealthCheckTimeoutDefaults sets default tiemout values for cluster's machine health checks.
+func setMachineHealthCheckTimeoutDefaults(cluster *anywherev1.Cluster, nodeStartupTimeout, unhealthyMachineTimeout string) {
```
g-gaston, line 71: ok so if I'm getting this right, this defaulter will always be initialized with something, either user provided values or our constant defaults. Which means that any cluster being managed from the CLI will end up with values set in these fields, no matter if the user set them or not.

This seems to present a change vs the API managed clusters. It seems like those by default will have empty values and then the controller will use the defaults when creating the MHC, but won't alter the original cluster.

I'm not pushing one way or the other, but I would love if we could make the behavior consistent.

pkg/clustermanager/cluster_manager.go
```diff
@@ -778,7 +776,17 @@ func getProviderNamespaces(providerDeployments map[string][]string) []string {
...
+	}
+
+	mhc, err := templater.ObjectsToYaml(kubernetes.ObjectsToRuntimeObjects(clusterapi.MachineHealthCheckObjects(clusterSpec.Cluster, unhealthyMachineTimeout, nodeStartupTimeout))...)
```
g-gaston, line 789: why wouldn't the `MachineHealthCheckObjects` simply take the timeouts from the cluster?

pkg/config/config.go
```diff
@@ -21,5 +21,7 @@ type CliConfig struct {
 
 // CreateClusterCLIConfig is the config we use for create cluster specific configurations.
 type CreateClusterCLIConfig struct {
-	SkipCPIPCheck bool
+	SkipCPIPCheck           bool
+	NodeStartupTimeout      string
+	UnhealthyMachineTimeout string
```
g-gaston, line 26: Curious, do we currently let users modify the MHC timeouts during cluster upgrades (with the CLI)?

pkg/gitops/flux/files_test.go
```diff
@@ -273,3 +274,22 @@ func TestFileGeneratorWriteFluxSystemFilesWriteFluxPatchesError(t *testing.T) {
 
 	tt.Expect(tt.g.WriteFluxSystemFiles(tt.clusterSpec)).To(MatchError(ContainSubstring("error in write patches")))
 }
+
+func NewCluster(clusterName string) *anywherev1.Cluster {
```
g-gaston, line 278: does this need to be exported?

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+		defaulter: defaulter,
+	}
+}
+
+// Reconcile installs machine health checks for a given cluster.
+// nolint:gocyclo
+func (r *Reconciler) Reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) error {
+	log.Info("Installing machine health checks for workload", "cluster", cluster.Name)
+
+	clusterSpec, err := anywhereCluster.BuildSpec(ctx, clientutil.NewKubeClient(r.client), cluster)
```
g-gaston, line 42: food for thought: if every reconciler ends up building the spec, we might just want to change some/all of them to just take the spec directly and pass it from the cluster controller.

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// getMachineHealthChecks checks if machine health checks already exist on the cluster and returns it.
+func (r *Reconciler) getMachineHealthChecks(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (*clusterv1.MachineHealthCheck, error) {
+	mhc := &clusterv1.MachineHealthCheck{}
+	// we get only kcp machine health check as it had the same timeouts as worker machine health checks
+	kcpMHCName := clusterapi.ControlPlaneMachineHealthCheckName(cluster)
+	err := r.client.Get(ctx, types.NamespacedName{Name: kcpMHCName, Namespace: constants.EksaSystemNamespace}, mhc)
+	if apierrors.IsNotFound(err) {
+		log.Info("Machine health checks don't exist for", "cluster", cluster.Name)
```
g-gaston, line 100: i think we can remove this log before merging

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+				}
+			}
+
+			if len(mhc.Spec.UnhealthyConditions) > 0 {
+				cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{
+					UnhealthyMachineTimeout: mhc.Spec.UnhealthyConditions[0].Timeout.Duration.String(),
+				}
+			}
+		}
+	}
```
g-gaston, line 65: Curious, what is this block for?
You are back filling the cluster spec with the current values of the MHC?

pkg/machinehealthcheck/reconciler.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// machineHealthCheckDefaulter := anywhereCluster.NewMachineHealthCheckDefaulter(constants.DefaultNodeStartupTimeout.String(), constants.DefaultUnhealthyMachineTimeout.String())
+	clusterSpec, err = r.defaulter.MachineHealthCheckDefault(ctx, clusterSpec)
```
g-gaston, line 68: does this need the cluster spec? can it be written in a way so it only needs the cluster?

pkg/machinehealthcheck/reconciler.go
mitalipaygude: this is for clusters that exist before this release, and have machine health checks configured with different values that are not the default ones.

pkg/config/config.go
mitalipaygude: no we don't, looks like that was another bug that i ended up fixing in this PR.

pkg/clustermanager/cluster_manager.go
mitalipaygude: it also uses the name of the cluster for generating the machine health check name.

pkg/cluster/defaults.go
mitalipaygude: moving the logic, we currently do this in the cluster_manager for Tinkerbell. I can add comments yeah

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -165,6 +167,15 @@ func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
 func buildCreateCliConfig(clusterOptions *createClusterOptions) config.CreateClusterCLIConfig {
 	createCliConfig := config.CreateClusterCLIConfig{}
 	createCliConfig.SkipCPIPCheck = clusterOptions.skipIpCheck
+	if clusterOptions.noTimeouts {
+		maxTime := time.Duration(math.MaxInt64)
+		createCliConfig.NodeStartupTimeout = maxTime.String()
+		createCliConfig.UnhealthyMachineTimeout = maxTime.String()
```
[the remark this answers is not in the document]
mitalipaygude: the no timeouts flag also sets the values for machine health checks here https://github.com/aws/eks-anywhere/blob/main/pkg/clustermanager/cluster_manager.go#L268-L269

controllers/cluster_controller.go
```diff
@@ -374,6 +382,12 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 		}
 	}
 
+	if cluster.IsManaged() {
+		if err := r.machineHealthCheck.Reconcile(ctx, log, cluster); err != nil {
```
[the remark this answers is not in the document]
mitalipaygude: yes the machine health checks get deleted if the cluster is deleted. why do we want to wait till the capi cluster is created?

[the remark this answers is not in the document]
vivek-koppuru: I think we had this discussion where mhcs can get in the way of node rollouts, but not sure if that is an issue anymore.

pkg/cluster/defaults.go
mitalipaygude: This also makes the same changes for clusters created with the API. the original cluster object looks like same with MHC initialized in the spec for API and CLI managed clusters.

pkg/clustermanager/cluster_manager.go
vivek-koppuru: I think he means have the `time.ParseDuration` calls inside the `clusterapi.MachineHealthCheckObjects` right?

pkg/clustermanager/cluster_manager.go
```diff
@@ -790,6 +800,43 @@ func (c *ClusterManager) InstallMachineHealthChecks(ctx context.Context, cluster
...
+func machineHealthCheckDiffExists(eksaMachineHealthCheck *v1alpha1.MachineHealthCheck, capiMachineHealthCheck *clusterv1.MachineHealthCheck) bool {
+	if eksaMachineHealthCheck == nil && capiMachineHealthCheck == nil {
+		return false
+	}
+
+	if eksaMachineHealthCheck == nil || capiMachineHealthCheck == nil {
+		return true
+	}
+
+	if len(eksaMachineHealthCheck.NodeStartupTimeout) != len(capiMachineHealthCheck.Spec.NodeStartupTimeout.Duration.String()) {
```
vivek-koppuru, line 829: What if it is 20m00s vs 20m0s? Is that going to be counted as a difference because of this check?

config/manifest/eksa-components.yaml
```diff
@@ -6328,6 +6341,18 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - cluster.x-k8s.io
+  resources:
+  - machinehealthchecks
```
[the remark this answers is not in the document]
mitalipaygude: setting the namespace to eksa-system converts it from a clusterrole to a role and the controller errors saying patch is forbidden.

pkg/cluster/defaults.go
```diff
@@ -21,10 +26,55 @@ func NewControlPlaneIPCheckAnnotationDefaulter(skipIPCheck bool) ControlPlaneIPC
 }
 
 // ControlPlaneIPCheckDefault sets the annotation for control plane skip ip check if the flag is set to true.
-func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx context.Context, spec *Spec) (*Spec, error) {
+func (d ControlPlaneIPCheckAnnotationDefaulter) ControlPlaneIPCheckDefault(ctx context.Context, cluster *anywherev1.Cluster) (*anywherev1.Cluster, error) {
```
vivek-koppuru, line 29: So we want to change these to cluster now?

pkg/clustermanager/cluster_manager.go
```diff
@@ -805,6 +813,36 @@ func (c *ClusterManager) InstallMachineHealthChecks(ctx context.Context, cluster
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func machineHealthCheckEqual(eksaMachineHealthCheck *v1alpha1.MachineHealthCheck, capiMachineHealthCheck *clusterv1.MachineHealthCheck) bool {
+	oldEksaMachineHealthCheck := &v1alpha1.MachineHealthCheck{
+		NodeStartupTimeout: &clusterv1.DefaultNodeStartupTimeout,
```
vivek-koppuru, line 836: Why is this default?

pkg/machinehealthcheck/reconciler/reconciler.go
```diff
@@ -0,0 +1,70 @@
...
+	log.Info("Installing machine health checks for workload", "cluster", cluster.Name)
+
+	if cluster.Spec.MachineHealthCheck == nil {
+		mhc, err := machinehealthcheck.GetMachineHealthChecks(ctx, r.client, cluster)
+		if err != nil {
+			return err
+		}
+		if mhc != nil {
+			if mhc.Spec.NodeStartupTimeout != nil {
+				cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{
```
vivek-koppuru, line 44: When is the cluster object patched on the cluster? I know you mentioned we were but didn't the call have to be somewhere here then?

pkg/machinehealthcheck/reconciler/reconciler_test.go
```diff
@@ -0,0 +1 @@
+package reconciler_test
```
vivek-koppuru, line 1: Will we add tests?

mitalipaygude: yes working on those currently

pkg/cluster/defaults.go
mitalipaygude: yes, since we aren't yet using the cluster spec anywhere, changed it to cluster.

pkg/machinehealthcheck/reconciler/reconciler.go
mitalipaygude: Yes, the call on line 64 that says ReconcileObjects is internally calling a patch operation.

mitalipaygude:
/test eks-anywhere-presubmit
--

cmd/eksctl-anywhere/cmd/options.go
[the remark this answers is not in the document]
g-gaston: @jiayiwang7 is this on purpose?
Not sure is this the expected behavior

config/manifest/eksa-components.yaml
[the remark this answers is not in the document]
g-gaston: Do you know why that happens?
I'm not sure if it's wise to just give the controller more permissions because "there is an error" if we don't understand what error is coming from.

controllers/cluster_controller.go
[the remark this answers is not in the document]
g-gaston: > why do we want to wait till the capi cluster is created?
Because if the capi cluster object doesn't exist, the MHC won't be deleted if the user deletes the eks-a cluster. What's the point of creating MHC for a cluster that we don't know if it will be created (it can fail validations for exmaple).

pkg/workflows/upgrade.go
```diff
@@ -535,6 +533,13 @@ func (s *upgradeWorkloadClusterTask) Run(ctx context.Context, commandContext *ta
 		return &CollectDiagnosticsTask{}
 	}
 
+	logger.V(4).Info("Updating machine health checks on cluster")
+	err = commandContext.ClusterManager.UpdateMachineHealthChecks(ctx, commandContext.ClusterSpec, commandContext.ManagementCluster)
```
g-gaston, line 537: could we let the controller do this for us?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -146,7 +149,7 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 	}
 	defer close(ctx, deps)
 
-	clusterSpec, err = deps.CreateClusterDefaulter.Run(ctx, clusterSpec)
+	clusterSpec.Cluster, err = deps.CreateClusterDefaulter.Run(ctx, clusterSpec.Cluster)
```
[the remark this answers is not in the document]
mitalipaygude: so we don't currently update anything in the defaulter that is not the cluster. since the machine health check defaulter is created for both create and upgrade, i changed the signature post this comment https://github.com/aws/eks-anywhere/pull/6176#discussion_r1271315060

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -166,11 +169,65 @@ func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
...
+		createCliConfig.NodeStartupTimeout.Duration = maxTime
+		createCliConfig.UnhealthyMachineTimeout.Duration = maxTime
 
-	return createCliConfig
+		return createCliConfig, nil
+	}
+
+	unhealthyMachineTimeout, err := time.ParseDuration(clusterOptions.unhealthyMachineTimeout)
+	if err != nil {
+		return createCliConfig, err
```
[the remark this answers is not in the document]
mitalipaygude: since the config is not a pointer, i return an empty struct.

controllers/factory.go
```diff
@@ -525,3 +532,29 @@ func (f *Factory) withPackageControllerClient() *Factory {
...
 	return f
 }
+
+func (f *Factory) withMachineHealthCheckReconciler() *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.machineHealthCheckReconciler != nil {
+			return nil
+		}
+
+		nodeStartupTimeout := &metav1.Duration{
```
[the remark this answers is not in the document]
mitalipaygude: the actual defaulting logic is in the defaulter called from the controller, this is only to initialize the object  with default values.

pkg/workflows/upgrade.go
mitalipaygude: what about reconciling management cluster's mhcs tho?

pkg/machinehealthcheck/reconciler/reconciler.go
```diff
@@ -0,0 +1,70 @@
...
+
+			if len(mhc.Spec.UnhealthyConditions) > 0 {
+				cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{
+					UnhealthyMachineTimeout: &mhc.Spec.UnhealthyConditions[0].Timeout,
+				}
+			}
+		}
+	}
+
+	cluster, err := r.defaulter.MachineHealthCheckDefault(ctx, cluster)
```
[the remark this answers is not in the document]
mitalipaygude: this would set the defaults in case the eksa spec does not contain machine health checks.

config/manifest/eksa-components.yaml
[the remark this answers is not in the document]
mitalipaygude: Afaik this happens because the controller is cluster scoped (via the cluster roles) and can not list based on namespaces.  Looks like this is true for all other resources listed for the cluster controller as well.

pkg/cluster/defaults.go
mitalipaygude: We are always setting the value from release 0.17. However because users can leave it empty, decided to make it a pointer for easier nil checks.

controllers/cluster_controller.go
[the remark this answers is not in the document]
mitalipaygude: I wanted to install them pre cluster reconcile so that while creating the machines for the workload cluster itself the machine health checks could have been leveraged. I agree with the delete use case tho, moved it to post reconcile.

pkg/cli/createclusterdefaulter.go
mitalipaygude: https://github.com/aws/eks-anywhere/issues/6285

pkg/cluster/defaults.go
mitalipaygude: we were going to build the spec from the controller, however since we don't use spec today in any of the defaulters, I decided to not build it unless an actual use case comes up where we want to default something other than the cluster object.

mitalipaygude: since we are using just the cluster now and not updating any other objects in the spec and based on the discussion here https://github.com/aws/eks-anywhere/pull/6176#discussion_r1271315060 , this defaults also needs to be updated since all defaults need to have the same signature.

pkg/machinehealthcheck/reconciler/reconciler.go
```diff
@@ -0,0 +1,85 @@
...
+	log.Info("Installing machine health checks for workload", "cluster", cluster.Name)
+
+	if cluster.Spec.MachineHealthCheck == nil {
+		cluster.Spec.MachineHealthCheck = &anywherev1.MachineHealthCheck{}
+		mhc, err := machinehealthcheck.GetControlPlaneMachineHealthCheck(ctx, r.client, cluster)
+		if err != nil {
+			return err
+		}
+		if mhc != nil {
+			log.Info("Inside if condition")
```
vivek-koppuru, line 45: remove

vivek-koppuru (approved):

cmd/eksctl-anywhere/cmd/createcluster.go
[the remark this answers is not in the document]
g-gaston: Right, but that comment was about not needing the spec to set defaults from the reconciler. Not about changing the signature of the CLI default runner.

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -166,11 +168,57 @@ func buildCliConfig(clusterSpec *cluster.Spec) *config.CliConfig {
...
+	nodeStartupTimeout, err := time.ParseDuration(clusterOptions.nodeStartupTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	upgradeCliConfig.NodeStartupTimeout = nodeStartupTimeout
+	upgradeCliConfig.UnhealthyMachineTimeout = unhealthyMachineTimeout
+
+	return &upgradeCliConfig, nil
 }
```
g-gaston, line 222: this code is mostly copy-paste from the other method. not a blocker, but we should consider reusing the code

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
```diff
@@ -250,6 +250,29 @@ spec:
                 type: array
               kubernetesVersion:
                 type: string
+              machineHealthCheck:
+                description: MachineHealthCheck allows to configure timeouts for machine
+                  health checks. Machine Health Checks are responsible for remediating
+                  unhealthy Machines. Configuring these values will decide how long
+                  CAPI would wait to remediate unhealthy machine or determine health
```
g-gaston, line 257: Can we not mention CAPI? It adds a foreign concept to the API which IMO increases the level of entry to understand it. I think we should be able to define what this does in a CAPI agnostic way.

config/manifest/eksa-components.yaml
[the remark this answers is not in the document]
g-gaston: Let's take this offline, but you are trying to do "create + update" operations, not list. So that doesn't really explain the issue.

pkg/cluster/defaults.go
g-gaston: I think this is the same as https://github.com/aws/eks-anywhere/pull/6176/files#r1273894766
I'll continue here and we can close the other one

IMO we should avoid having to build the spec in the controller if we don't need it. But we shouldn't change the CLI defaulter signature. That took a spec for a reason, it's explained in the validations and defaults design doc.

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -0,0 +1,32 @@
+package machinehealthcheck
```
g-gaston, line 1: Would this package make more sense under `clusterapi`?

pkg/machinehealthcheck/reconciler/reconciler.go
```diff
@@ -0,0 +1,77 @@
...
+
+	err = serverside.ReconcileObjects(ctx, r.client, clientutil.ObjectsToClientObjects(mhcObject))
+	if err != nil {
+		return fmt.Errorf("applying machine health checks: %v", err)
+	}
+
+	return nil
+}
+
+func setNodeStartupTimeout(cluster *anywherev1.Cluster, capiMHC *clusterv1.MachineHealthCheck) {
```
g-gaston, line 65: nit
```suggestion
func copyNodeStartupTimeoutToCluster(cluster *anywherev1.Cluster, capiMHC *clusterv1.MachineHealthCheck) {
```

pkg/machinehealthcheck/reconciler/reconciler.go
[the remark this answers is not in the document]
g-gaston: I'm confused. If the cluster doesn't have anything configured (existing cluster before this release), the block above will read the existing MHC and populate the cluster, that's great.

Otherwise, for all new clusters, shouldn't the webhook set the defaults? This way this reconciler only takes care of old clusters (this is a temporary thing, only required for the transition) and of creating/updating MHC.

cmd/eksctl-anywhere/cmd/options.go
mitalipaygude: yes i have a follow up to combine the timeout structs for create and upgrade defaulters.

config/manifest/eksa-components.yaml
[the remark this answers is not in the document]
mitalipaygude: the exact operation is patch, and I said list as an example of the operation. I believe its for an operation the client tries to perform.

pkg/machinehealthcheck/machinehealthcheck.go
mitalipaygude: sure, I can move it. i saw a similar reconcile for aws iam auth at the pkg level so replicated it.

pkg/machinehealthcheck/reconciler.go
mitalipaygude: https://github.com/aws/eks-anywhere/issues/6301

pkg/machinehealthcheck/reconciler/reconciler.go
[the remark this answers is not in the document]
mitalipaygude: The reason we don't want to move the defaults to the webhook level is because we want to default values if the machine health checks in the spec are empty. If we end up defaulting in the webhooks before getting to the reconciler, there is no way for us to detect if the values were set by the controller defaults or the user.

config/manifest/eksa-components.yaml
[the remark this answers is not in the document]
mitalipaygude: The permissions are using the ClusterRole due to client trying to list this resource from cache.
When trying to list resources not already in cache, it starts an informer for that type using the scope of the cache.
So if the manager is cluster-scoped, the new informers created by the cache will be cluster-scoped. Discussed this with Guillermo offline and came to the conclusion that the permissions need to be cluster scoped.

vivek-koppuru (approved):

PR #6180, opened 2023-07-14
Modular k8s upgrade BuildSpec

tatlat:
*Issue #, if available:*

*Description of changes:*
NewSpec and BuildSpec will now include eksd release and version bundles corresponding to worker node group's kubernetes version.

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/common.go
+++ b/cmd/eksctl-anywhere/cmd/common.go
@@ -6,23 +6,37 @@ import (
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/dependencies"
 	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/files"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/manifests/bundles"
 	"github.com/aws/eks-anywhere/pkg/version"
 	"github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// getImages returns all the images in the Bundle for the cluster kubernetes versions.
+// This is deprecated. It builds a file reader in line, prefer using the dependency factory.
 func getImages(clusterSpecPath, bundlesOverride string) ([]v1alpha1.Image, error) {
 	var specOpts []cluster.FileSpecBuilderOpt
 	if bundlesOverride != "" {
 		specOpts = append(specOpts, cluster.WithOverrideBundlesManifest(bundlesOverride))
 	}
-	clusterSpec, err := readAndValidateClusterSpec(clusterSpecPath, version.Get(), specOpts...)
+	cliVersion := version.Get()
+	spec, err := readClusterSpec(clusterSpecPath, cliVersion, specOpts...)
 	if err != nil {
 		return nil, err
 	}
-	bundle := clusterSpec.VersionsBundle
-	images := append(bundle.Images(), clusterSpec.KubeDistroImages()...)
-	return images, nil
+
+	kubeVersions := spec.Cluster.KubernetesVersions()
+	kubeVersionsFilter := make([]string, 0, len(kubeVersions))
+	for _, version := range kubeVersions {
+		kubeVersionsFilter = append(kubeVersionsFilter, string(version))
+	}
+
+	return bundles.ReadImages(
+		files.NewReader(files.WithEKSAUserAgent("cli", cliVersion.GitVersion)),
+		spec.Bundles,
+		kubeVersionsFilter...,
+	)
 }
 
 // getKubeconfigPath returns an EKS-A kubeconfig path. The return van be overriden using override
--- a/cmd/eksctl-anywhere/cmd/listovas.go
+++ b/cmd/eksctl-anywhere/cmd/listovas.go
@@ -65,8 +65,18 @@ func listOvas(context context.Context, clusterSpecPath, bundlesOverride string) 
 		return err
 	}
 
-	bundle := clusterSpec.VersionsBundle
+	for _, version := range clusterSpec.Cluster.KubernetesVersions() {
+		bundle := clusterSpec.VersionsBundle(version)
+		err := printOvas(bundle)
+		if err != nil {
+			return err
+		}
+	}
 
+	return nil
+}
+
+func printOvas(bundle *cluster.VersionsBundle) error {
 	titler := cases.Title(language.English)
 	for _, ova := range bundle.Ovas() {
 		if strings.Contains(ova.URI, string(eksav1alpha1.Bottlerocket)) {
@@ -85,7 +95,6 @@ func listOvas(context context.Context, clusterSpecPath, bundlesOverride string) 
 		}
 		fmt.Println(yamlIndent(2, string(yamlOutput)))
 	}
-
 	return nil
 }
 
--- a/pkg/cluster/builder.go
+++ b/pkg/cluster/builder.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	eksdv1 "github.com/aws/eks-distro-build-tooling/release/api/v1alpha1"
 	"github.com/pkg/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -82,12 +83,8 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 		return nil, err
 	}
 
-	versionsBundle, err := GetVersionsBundle(config.Cluster, bundlesManifest)
-	if err != nil {
-		return nil, err
-	}
+	eksdReleases, err := getAllEksdReleases(config.Cluster, bundlesManifest, b.reader)
 
-	eksd, err := bundles.ReadEKSD(b.reader, *versionsBundle)
 	if err != nil {
 		return nil, err
 	}
@@ -101,7 +98,34 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 	config.Cluster.Spec.EksaVersion = &releaseVersion
 	eksaRelease := buildEKSARelease(release, bundlesManifest)
 
-	return NewSpec(config, bundlesManifest, eksd, eksaRelease)
+	return NewSpec(config, bundlesManifest, eksdReleases, eksaRelease)
+}
+
+func getAllEksdReleases(cluster *v1alpha1.Cluster, bundlesManifest *releasev1.Bundles, reader bundles.Reader) ([]eksdv1.Release, error) {
+	versions := cluster.KubernetesVersions()
+	m := make([]eksdv1.Release, 0, len(versions))
+	for _, version := range versions {
+		eksd, err := getEksdReleases(version, bundlesManifest, reader)
+		if err != nil {
+			return nil, err
+		}
+		m = append(m, *eksd)
+	}
+	return m, nil
+}
+
+func getEksdReleases(version v1alpha1.KubernetesVersion, bundlesManifest *releasev1.Bundles, reader bundles.Reader) (*eksdv1.Release, error) {
+	versionsBundle, err := GetVersionsBundle(version, bundlesManifest)
+	if err != nil {
+		return nil, err
+	}
+
+	eksd, err := bundles.ReadEKSD(reader, *versionsBundle)
+	if err != nil {
+		return nil, err
+	}
+
+	return eksd, nil
 }
 
 func (b FileSpecBuilder) getConfig(clusterConfigURL string) (*Config, error) {
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -29,8 +29,9 @@ func bundlesNamespacedKey(cluster *v1alpha1.Cluster, release *v1alpha1release.EK
 	return name, namespace
 }
 
-func GetVersionsBundle(clusterConfig *v1alpha1.Cluster, bundles *v1alpha1release.Bundles) (*v1alpha1release.VersionsBundle, error) {
-	return getVersionsBundleForKubernetesVersion(clusterConfig.Spec.KubernetesVersion, bundles)
+// GetVersionsBundle gets the VersionsBundle that corresponds to KubernetesVersion.
+func GetVersionsBundle(version v1alpha1.KubernetesVersion, bundles *v1alpha1release.Bundles) (*v1alpha1release.VersionsBundle, error) {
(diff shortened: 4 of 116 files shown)


pkg/providers/vsphere/defaults.go
```diff
@@ -77,7 +77,11 @@ func (d *Defaulter) setDefaultTemplateIfMissing(ctx context.Context, spec *Spec,
 
 func (d *Defaulter) setupDefaultTemplate(ctx context.Context, spec *Spec, machineConfig *anywherev1.VSphereMachineConfig) error {
```
g-gaston, line 78: this might need to change
we might need a different template depending if the machineConfig is used for CP or for a worker node group

pkg/providers/vsphere/tags.go
```diff
@@ -20,8 +20,12 @@ func requiredTemplateTags(clusterSpec *cluster.Spec, machineConfig *v1alpha1.VSp
 
 func requiredTemplateTagsByCategory(clusterSpec *cluster.Spec, machineConfig *v1alpha1.VSphereMachineConfig) map[string][]string {
 	osFamily := machineConfig.Spec.OSFamily
+	vb, err := clusterSpec.GetCPVersionsBundle()
```
g-gaston, line 23: same here
this depends on the kubernetes version assigned to the machine config (cp vs workers)

pkg/providers/vsphere/template.go
```diff
@@ -61,7 +61,10 @@ func (vs *VsphereTemplateBuilder) GenerateCAPISpecControlPlane(
 }
 
 func (vs *VsphereTemplateBuilder) isCgroupDriverSystemd(clusterSpec *cluster.Spec) (bool, error) {
```
g-gaston, line 63: this probably depends on the node, doesn't it?

g-gaston (commented):
This looks so much better already!
However, I'm missing unit tests for the providers capi spec generation.
We have updated the existing unit tests to work with the new logic, which is great. But we are missing tests that run through the new logic when different k8s versions are specified.

cmd/eksctl-anywhere/cmd/common.go
```diff
@@ -20,7 +20,7 @@ func getImages(clusterSpecPath, bundlesOverride string) ([]v1alpha1.Image, error
 	if err != nil {
 		return nil, err
 	}
-	bundle := clusterSpec.VersionsBundle
```
jonathanmeier5, line 23: eksctl-anywhere isn't production code as far as I'm aware, but have we validated changing these defaults has no side effects?

pkg/providers/vsphere/defaults.go
```diff
@@ -75,9 +76,31 @@ func (d *Defaulter) setDefaultTemplateIfMissing(ctx context.Context, spec *Spec,
...
+	}
+
+	return nil
+}
+
+func getVersionsBundleForMachineConfig(spec *cluster.Spec, machineConfig *anywherev1.VSphereMachineConfig) *cluster.VersionsBundle {
+	if spec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name != machineConfig.Name {
+		wng := getWorkerNodeGroupForMachineConfig(spec.Cluster, machineConfig)
+		if wng != nil {
+			return spec.WorkerNodeGroupVersionsBundle(*wng)
```
jonathanmeier5, line 93: Could we get test coverage for this?

jonathanmeier5 (commented):
Few small questions/nits.

it would be nice if we could squash this branch's commits before merging with a clear commit message indicating what is included, that way we don't end up with a bunch of `fix lint` messages in our git history.

pkg/cluster/builder.go
```diff
@@ -82,17 +84,48 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
...
+	m = append(m, *eksd)
+
+	for _, wng := range cluster.Spec.WorkerNodeGroupConfigurations {
+		if wng.KubernetesVersion != nil {
+			version = *wng.KubernetesVersion
+			eksd, err = getEksdReleases(version, bundlesManifest, reader)
+			if err != nil && eksd == nil {
+				return nil, err
+			}
+			m = append(m, *eksd)
```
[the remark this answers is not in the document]
vivek-koppuru: Would it not get to here due to the `continue` above?

pkg/providers/vsphere/spec.go
```diff
@@ -37,6 +37,21 @@ func (s *Spec) machineConfigs() []*anywherev1.VSphereMachineConfig {
 	return machineConfigs
 }
 
+func (s *Spec) templates() []string {
```
vivek-koppuru, line 40: Is this actually unused?

vivek-koppuru (commented):
If there are only worker node group changes in templates, do we go through the bootstrap cluster process?

pkg/cluster/builder.go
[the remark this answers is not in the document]
g-gaston: that continue wasn't there before :)

pkg/cluster/spec.go
```diff
@@ -113,16 +110,72 @@ func NewSpec(config *Config, bundles *v1alpha1.Bundles, eksdRelease *eksdv1alpha
...
+	vb := &VersionsBundle{
+		VersionsBundle: v,
+		KubeDistro:     kd,
+	}
+
+	return vb, nil
+}
+
+// VersionBundles returns a VersionsBundle if one exists for the provided kubernetes version and nil otherwise.
+func (s *Spec) VersionBundles(version eksav1alpha1.KubernetesVersion) *VersionsBundle {
```
g-gaston, line 159: should this method be named `VersionsBundle`?

pkg/providers/cloudstack/testdata/cluster_main_worker_node_group_kubernetes_version.yaml
```diff
@@ -0,0 +1,127 @@
...
+  datacenterRef:
+    kind: CloudStackDatacenterConfig
+    name: test
+  externalEtcdConfiguration:
+    count: 3
+    machineGroupRef:
+      kind: CloudStackMachineConfig
+      name: test-etcd
+  kubernetesVersion: "1.21"
+  workerNodeGroupConfigurations:
```
g-gaston, line 31: I think we should add another worker node group with no kube version
To all providers
So we test it defaults correctly to the top level one

pkg/providers/snow/snow_test.go
```diff
@@ -863,23 +863,98 @@ func TestGenerateCAPISpecForUpgrade(t *testing.T) {
 	test.AssertContentToFile(t, string(gotMd), "testdata/expected_results_main_md_ubuntu.yaml")
 }
 
+func TestGenerateCAPISpecForUpgradeWorkerVersion(t *testing.T) {
+	tt := newSnowTest(t)
+	workerVersionsBundle := givenVersionsBundle()
+	workerVersionsBundle.KubeDistro.Kubernetes.Tag = "v1.20.2-eks-1-20-7"
+	tt.clusterSpec.VersionsBundles["1.20"] = workerVersionsBundle
+	mt := wantSnowMachineTemplate()
+	tt.kubeUnAuthClient.EXPECT().KubeconfigClient(tt.cluster.KubeconfigFile).Return(tt.kubeconfigClient)
```
g-gaston, line 872: where are we here setting different kube versions for the workers?

pkg/cluster/fetch.go
```diff
@@ -59,7 +60,46 @@ func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*S
...
+}
+
+func fetchAllEksdReleases(ctx context.Context, client Client, cluster *v1alpha1.Cluster, bundles *v1alpha1release.Bundles) ([]eksdv1alpha1.Release, error) {
+	m := make([]eksdv1alpha1.Release, 0)
+	visited := map[string]bool{}
+
+	version := cluster.Spec.KubernetesVersion
+	visited[string(version)] = true
+	eksd, err := getEksdRelease(ctx, client, version, bundles)
+	if err != nil && eksd == nil {
```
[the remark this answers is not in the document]
tatlat: no

pkg/cluster/builder.go
```diff
@@ -105,7 +101,49 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 	config.Cluster.Spec.EksaVersion = &releaseVersion
 	eksaRelease := buildEKSARelease(release, bundlesManifest)
 
-	return NewSpec(config, bundlesManifest, eksd, eksaRelease)
+	return NewSpec(config, bundlesManifest, eksdReleases, eksaRelease)
+}
+
+func getAllEksdReleases(cluster *v1alpha1.Cluster, bundlesManifest *releasev1.Bundles, reader bundles.Reader) ([]eksdv1.Release, error) {
+	m := make([]eksdv1.Release, 0)
```
g-gaston, line 108: this will always have at least size 1
```suggestion
	m := make([]eksdv1.Release, 0, 1)
```

pkg/cluster/builder.go
```diff
@@ -105,7 +101,49 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
...
+
+	for _, wng := range cluster.Spec.WorkerNodeGroupConfigurations {
+		if wng.KubernetesVersion != nil {
+			version = *wng.KubernetesVersion
+			if _, ok := visited[version]; ok {
+				continue
+			}
+			visited[version] = struct{}{}
+			eksd, err = getEksdReleases(version, bundlesManifest, reader)
+			if err != nil && eksd == nil {
```
g-gaston, line 126: Can we just remove the eksd nil check here and in line 113?

cmd/eksctl-anywhere/cmd/listovas.go
```diff
@@ -65,8 +65,17 @@ func listOvas(context context.Context, clusterSpecPath, bundlesOverride string)
 		return err
 	}
 
-	bundle := clusterSpec.VersionsBundle
+	for _, bundle := range clusterSpec.VersionsBundlesMap {
```
g-gaston, line 68: this trusts that the versionsBundlesMap only has releases for the kubernetes versions in the cluster object. I think this is an internal implementation, it's probably a bad idea to make external callers rely on this. Can't we just get the unique kubernetes versions from the cluster and iterate over them?

g-gaston: Or if you think it's something you use frequently enough, write a method in the spec like `ClusterVersionBundles` that returns such filtered collection

pkg/cluster/fetch.go
```diff
@@ -74,7 +75,46 @@ func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*S
...
-	versionsBundle, err := GetVersionsBundle(config.Cluster, bundles)
+	eksdReleases, err := fetchAllEksdReleases(ctx, client, config.Cluster, bundles)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewSpec(config, bundles, eksdReleases, eksaRelease)
+}
+
+func fetchAllEksdReleases(ctx context.Context, client Client, cluster *v1alpha1.Cluster, bundles *v1alpha1release.Bundles) ([]eksdv1alpha1.Release, error) {
```
g-gaston, line 86: I think this whole method can be significatly simplified by using `cluster.KubernetesVersions`
same with the similar method in fetch.go

pkg/cluster/spec.go
```diff
@@ -116,16 +113,72 @@ func NewSpec(config *Config, bundles *v1alpha1.Bundles, eksdRelease *eksdv1alpha
...
+	vb := &VersionsBundle{
+		VersionsBundle: v,
+		KubeDistro:     kd,
+	}
+
+	return vb, nil
+}
+
+// VersionsBundles returns a VersionsBundle if one exists for the provided kubernetes version and nil otherwise.
+func (s *Spec) VersionsBundles(version eksav1alpha1.KubernetesVersion) *VersionsBundle {
```
g-gaston, line 162: ok maybe we need to synç offline, we keep coming back to this name. Sorry I have created any confusion. I think this should be
```suggestion
// VersionsBundle returns a VersionsBundle if one exists for the provided kubernetes version and nil otherwise.
func (s *Spec) VersionsBundle(version eksav1alpha1.KubernetesVersion) *VersionsBundle {
```
Because it returns one single versions bundle

pkg/cluster/spec.go
```diff
@@ -13,27 +13,22 @@ import (
...
-	OIDCConfig        *eksav1alpha1.OIDCConfig
-	AWSIamConfig      *eksav1alpha1.AWSIamConfig
-	ManagementCluster *types.Cluster // TODO(g-gaston): cleanup, this doesn't belong here
-	EKSARelease       *v1alpha1.EKSARelease
+	Bundles            *v1alpha1.Bundles
+	OIDCConfig         *eksav1alpha1.OIDCConfig
+	AWSIamConfig       *eksav1alpha1.AWSIamConfig
+	ManagementCluster  *types.Cluster // TODO(g-gaston): cleanup, this doesn't belong here
+	EKSARelease        *v1alpha1.EKSARelease
+	VersionsBundlesMap map[eksav1alpha1.KubernetesVersion]*VersionsBundle
```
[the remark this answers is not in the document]
tatlat: The collision comes from the function with the same name

pkg/cluster/builder.go
```diff
@@ -105,7 +101,33 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 	config.Cluster.Spec.EksaVersion = &releaseVersion
 	eksaRelease := buildEKSARelease(release, bundlesManifest)
 
-	return NewSpec(config, bundlesManifest, eksd, eksaRelease)
+	return NewSpec(config, bundlesManifest, eksdReleases, eksaRelease)
+}
+
+func getAllEksdReleases(cluster *v1alpha1.Cluster, bundlesManifest *releasev1.Bundles, reader bundles.Reader) ([]eksdv1.Release, error) {
+	m := make([]eksdv1.Release, 0, 1)
+	for _, version := range cluster.KubernetesVersions() {
```
g-gaston, line 109: nit, not a blocker
```suggestion
	versions := cluster.KubernetesVersions()
	m := make([]eksdv1.Release, 0, len(versions))
	for _, version := ranger versions {
```

pkg/cluster/fetch.go
```diff
@@ -74,7 +75,29 @@ func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*S
...
+	m := make([]eksdv1alpha1.Release, 0, 1)
+	for _, version := range cluster.KubernetesVersions() {
```
g-gaston, line 88: same nit here about the length

Issue #5865, opened 2023-05-12
Fail on missing privileges for VSphere user


vivek-koppuru:
Currently, we throw a warning if there are missing privileges for a VSphere user in the form of a `-v=3` log. We started off with a hard fail https://github.com/aws/eks-anywhere/pull/2907 but moved to a warning https://github.com/aws/eks-anywhere/pull/3160 to allow users to continue with the workflow. Now that we have `--skip-validations` as an option https://github.com/aws/eks-anywhere/pull/5697, we should switch back to hard failing but allowing for the checks to be bypassed through this flag so that we better alert users instead of issues happening with creation/upgrade later in the workflow.

PR #6189, opened 2023-07-17
Refactor skipvalidations to include create

pokearu:
*Issue #5865:*

*Description of changes:*
`--skip-validations` flag was initally only done for upgrade preflights. Refactored the code to have skip validations for create cluster.

*Testing:*
Tested creating a cluster
```bash
./eksctl-anywhere create cluster -f ./config.yaml 
```


cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -98,6 +99,14 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
 		return fmt.Errorf("failed to build cluster manager opts: %v", err)
 	}
 
+	var skippedValidations map[string]bool
+	if len(uc.skipValidations) != 0 {
```
jonathanmeier5, line 103: I'd love to get all these validations out of the command function, but it's not something we should worry about for this.

pkg/validations/skipvalidations.go
```diff
@@ -0,0 +1,34 @@
...
+
+// ValidateSkippableValidation validates if provided validations are supported by EKSA to skip for upgrades.
+func ValidateSkippableValidation(skippedValidations []string, skippableValidations []string) (map[string]bool, error) {
+	svMap := validSkippableValidationsMap(skippableValidations)
+
+	for i := range skippedValidations {
+		validationName := skippedValidations[i]
+		_, ok := svMap[validationName]
+		if !ok {
+			return nil, fmt.Errorf("invalid validation name to be skipped. The supported validations that can be skipped using --skip-validations are %s", strings.Join(skippableValidations[:], ","))
```
jonathanmeier5, line 28: It would be nice if we could collect a list of these errors instead of failing on the first bad skipped validation, but I don't think it's worth blocking this PR on.

jonathanmeier5 (approved):
/lgtm

pkg/validations/skipvalidations.go
pokearu: Ah like say what all are invalid. Makes sense! 
So right now we have only 1 for upgrade and I will add one for create, so yes definitely something we can do as a follow.

PR #6193, opened 2023-07-17
Adding skip validation support for vsphere validations

pokearu:
*Issue #5865:*

*Description of changes:*
The PR adds support for `--skip-validations` for create cluster and also pass's it to the provider level validations.
Originally we had it limited to upgrade cluster's preflight level checks alone.

*Testing:*
Created a user with missing permissions and tested validations. Now the error shows as

If we pass --skip-validations, the check is skipped.
```bash
root@ip-10-0-59-81:/home/ubuntu# ./eksctl-anywhere create cluster --help
This command is used to create workload clusters

Usage:
  anywhere create cluster -f <cluster-config-file> [flags]

Flags:
...
      --skip-validations stringArray        Bypass create validations by name. Valid arguments you can pass are --skip-validations=vsphere-user-privilege
...
```

Tested creating a cluster by skipping the check
```bash
./eksctl-anywhere create cluster -f ./config.yaml --skip-validations=vsphere-user-privilege
```

*Documentation added/planned :*
Plan to document the new flag and change in user experience.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/supportbundle.go
+++ b/cmd/eksctl-anywhere/cmd/supportbundle.go
@@ -89,7 +89,7 @@ func (csbo *createSupportBundleOptions) createBundle(ctx context.Context, since,
 	}
 
 	deps, err := dependencies.ForSpec(ctx, clusterSpec).
-		WithProvider(csbo.fileName, clusterSpec.Cluster, cc.skipIpCheck, csbo.hardwareFileName, false, csbo.tinkerbellBootstrapIP).
+		WithProvider(csbo.fileName, clusterSpec.Cluster, cc.skipIpCheck, csbo.hardwareFileName, false, csbo.tinkerbellBootstrapIP, map[string]bool{}).
 		WithDiagnosticBundleFactory().
 		Build(ctx)
 	if err != nil {
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/providers/common"
 	"github.com/aws/eks-anywhere/pkg/retrier"
 	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/validations"
 	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
@@ -76,6 +77,7 @@ type vsphereProvider struct {
 	validator             *Validator
 	defaulter             *Defaulter
 	ipValidator           IPValidator
+	skippedValidations    map[string]bool
 }
 
 type ProviderGovcClient interface {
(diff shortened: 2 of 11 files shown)


cmd/eksctl-anywhere/cmd/supportbundle.go
```diff
@@ -89,7 +89,7 @@ func (csbo *createSupportBundleOptions) createBundle(ctx context.Context, since,
 	}
 
 	deps, err := dependencies.ForSpec(ctx, clusterSpec).
-		WithProvider(csbo.fileName, clusterSpec.Cluster, cc.skipIpCheck, csbo.hardwareFileName, false, csbo.tinkerbellBootstrapIP).
+		WithProvider(csbo.fileName, clusterSpec.Cluster, cc.skipIpCheck, csbo.hardwareFileName, false, csbo.tinkerbellBootstrapIP, map[string]bool{}).
```
jonathanmeier5, line 92: why does support bundle need this information?

pkg/providers/vsphere/vsphere.go
```diff
@@ -140,7 +142,18 @@ type IPValidator interface {
...
+// NewProvider initializes and returns a new vsphereProvider.
+func NewProvider(
+	datacenterConfig *v1alpha1.VSphereDatacenterConfig,
+	clusterConfig *v1alpha1.Cluster,
+	providerGovcClient ProviderGovcClient,
+	providerKubectlClient ProviderKubectlClient,
+	writer filewriter.FileWriter,
+	ipValidator IPValidator,
+	now types.NowFunc,
+	skipIPCheck bool,
```
jonathanmeier5, line 154: Can `skipIPCheck` eventually be refactored factored into the `skippedValidations` map?

pkg/providers/vsphere/vsphere.go
```diff
@@ -380,8 +409,10 @@ func (p *vsphereProvider) SetupAndValidateUpgradeCluster(ctx context.Context, cl
 		return fmt.Errorf("validating vsphere machine configs datastore usage: %v", err)
 	}
 
-	if err := p.validator.validateVsphereUserPrivs(ctx, vSphereClusterSpec); err != nil {
-		return fmt.Errorf("validating vsphere user privileges: %v", err)
+	if !p.skippedValidations[validations.VSphereUserPriv] {
+		if err := p.validator.validateVsphereUserPrivs(ctx, vSphereClusterSpec); err != nil {
+			return fmt.Errorf("validating vsphere user privileges: %v", err)
```
jonathanmeier5, line 414: Do we have an argument for not running a test case that triggers this error? It feels like the test case of verifying the error can be thrown would be an important one. Open to hearing counter arguments, those are just my first thoughts.

pkg/providers/vsphere/vsphere.go
```diff
@@ -339,8 +366,10 @@ func (p *vsphereProvider) SetupAndValidateCreateCluster(ctx context.Context, clu
 		logger.Info("Skipping check for whether control plane ip is in use")
 	}
 
-	if err := p.validator.validateVsphereUserPrivs(ctx, vSphereClusterSpec); err != nil {
-		return fmt.Errorf("validating vsphere user privileges: %v", err)
+	if !p.skippedValidations[validations.VSphereUserPriv] {
+		if err := p.validator.validateVsphereUserPrivs(ctx, vSphereClusterSpec); err != nil {
+			return fmt.Errorf("validating vsphere user privileges: %v", err)
```
jonathanmeier5, line 371: (same as below) Do we have an argument for not running a test case that triggers this error? It feels like the test case of verifying the error can be thrown would be an important one. Open to hearing counter arguments, those are just my first thoughts.

jonathanmeier5 (commented):
Left a few questions.

One higher level question: did we support skipped validations on the create command before this PR? If not, can we call out in the PR description that we are adding skipped validation support to the create command?

cmd/eksctl-anywhere/cmd/supportbundle.go
pokearu: oh no it doesn't. Hence the empty object. Similar to the `false` being passed in. Its mainly because its a shared function.

pokearu:
> Left a few questions.
> 
> One higher level question: did we support skipped validations on the create command before this PR? If not, can we call out in the PR description that we are adding skipped validation support to the create command?

We did support it before. I can add another line 👍

pkg/providers/vsphere/vsphere.go
pokearu: No you are right. Just added the tests to cover it 👍

cmd/eksctl-anywhere/cmd/supportbundle.go
jonathanmeier5: Got it. This feels a bit like a code smell to me if components that don't need a resource have to take it as an argument.

Not anyone's fault and definitely not something to fix for this PR, just something to think about.

jonathanmeier5 (approved):
/lgtm

cmd/eksctl-anywhere/cmd/supportbundle.go
pokearu: Yea i wasn't too happy with it either. I wanted to keep it as optional but didnt find a good way to pass along the map since go does it as array of options and i had only one specific thing to pass.

PR #6197, opened 2023-07-19
K8s modular upgrade validations

tatlat:
*Issue #, if available:*

*Description of changes:*
Validations for modular k8s upgrades

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -510,6 +510,7 @@ func WorkerNodeGroupConfigurationsLabelsMapEqual(a, b []WorkerNodeGroupConfigura
 	return true
 }
 
+// WorkerNodeGroupConfigurationKubeVersionUnchanged checks if a worker node group's k8s version has not changed.
 func WorkerNodeGroupConfigurationKubeVersionUnchanged(o, n *WorkerNodeGroupConfiguration, oldTopVersion, newTopVersion KubernetesVersion) bool {
 	oldVersion := o.KubernetesVersion
 	newVersion := n.KubernetesVersion
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -510,6 +510,20 @@ func WorkerNodeGroupConfigurationsLabelsMapEqual(a, b []WorkerNodeGroupConfigura
 	return true
 }
 
+func WorkerNodeGroupConfigurationKubeVersionUnchanged(o, n *WorkerNodeGroupConfiguration, oldTopVersion, newTopVersion KubernetesVersion) bool {
+	oldVersion := o.KubernetesVersion
+	newVersion := n.KubernetesVersion
+
+	if oldVersion == nil {
+		oldVersion = &oldTopVersion
+	}
+	if newVersion == nil {
+		newVersion = &newTopVersion
+	}
+
+	return newVersion.Equal(oldVersion)
+}
+
 type ClusterNetwork struct {
 	// Comma-separated list of CIDR blocks to use for pod and service subnets.
 	// Defaults to 192.168.0.0/16 for pod subnet.
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -510,16 +510,16 @@ func WorkerNodeGroupConfigurationsLabelsMapEqual(a, b []WorkerNodeGroupConfigura
 	return true
 }
 
-// WorkerNodeGroupConfigurationKubeVersionUnchanged checks if a worker node group's k8s version has not changed.
-func WorkerNodeGroupConfigurationKubeVersionUnchanged(o, n *WorkerNodeGroupConfiguration, oldTopVersion, newTopVersion KubernetesVersion) bool {
+// WorkerNodeGroupConfigurationKubeVersionUnchanged checks if a worker node group's k8s version has not changed. The ClusterVersions are the top level kubernetes version of a cluster.
+func WorkerNodeGroupConfigurationKubeVersionUnchanged(o, n *WorkerNodeGroupConfiguration, oldCluster, newCluster *Cluster) bool {
 	oldVersion := o.KubernetesVersion
 	newVersion := n.KubernetesVersion
 
 	if oldVersion == nil {
-		oldVersion = &oldTopVersion
+		oldVersion = &oldCluster.Spec.KubernetesVersion
 	}
 	if newVersion == nil {
-		newVersion = &newTopVersion
+		newVersion = &newCluster.Spec.KubernetesVersion
 	}
 
 	return newVersion.Equal(oldVersion)
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -98,6 +98,8 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 
 	allErrs = append(allErrs, ValidateKubernetesVersionSkew(r, oldCluster)...)
 
+	allErrs = append(allErrs, ValidateWorkerKubernetesVersionSkew(r, oldCluster)...)
+
 	if len(allErrs) != 0 {
 		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
 	}
@@ -347,18 +349,21 @@ func (r *Cluster) ValidateDelete() error {
 
 // ValidateKubernetesVersionSkew validates Kubernetes version skew between upgrades.
 func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
-	var allErrs field.ErrorList
-
 	path := field.NewPath("spec")
-
 	oldVersion := old.Spec.KubernetesVersion
 	newVersion := new.Spec.KubernetesVersion
+
+	return validateKubeVersionSkew(newVersion, oldVersion, path)
+}
+
+func validateKubeVersionSkew(newVersion, oldVersion KubernetesVersion, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
 
 	parsedOldVersion, err := version.ParseGeneric(string(oldVersion))
 	if err != nil {
 		allErrs = append(
 			allErrs,
-			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing cluster version: %v", err.Error())))
+			field.Invalid(path, oldVersion, fmt.Sprintf("parsing cluster version: %v", err.Error())))
 		return allErrs
 	}
 
@@ -366,7 +371,7 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
 	if err != nil {
 		allErrs = append(
 			allErrs,
-			field.Invalid(path, new.Spec.KubernetesVersion, fmt.Sprintf("parsing comparison version: %v", err.Error())))
+			field.Invalid(path, newVersion, fmt.Sprintf("parsing comparison version: %v", err.Error())))
 		return allErrs
 	}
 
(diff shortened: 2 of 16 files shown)


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -510,6 +510,21 @@ func WorkerNodeGroupConfigurationsLabelsMapEqual(a, b []WorkerNodeGroupConfigura
 	return true
 }
 
+// WorkerNodeGroupConfigurationKubeVersionUnchanged checks if a worker node group's k8s version has not changed.
+func WorkerNodeGroupConfigurationKubeVersionUnchanged(o, n *WorkerNodeGroupConfiguration, oldTopVersion, newTopVersion KubernetesVersion) bool {
```
vivek-koppuru, line 514: Maybe call it `oldClusterVersion` or something showing that it's in the Cluster level?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+	}
+
+	return allErrs
+}
+
+// ValidateWorkerKubernetesVersionSkew validates worker node group Kubernetes version skew between upgrades.
+func ValidateWorkerKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+
+	topKubeVersion := new.Spec.KubernetesVersion
```
vivek-koppuru, line 395: Yea I am curious if we can just stick to replacing all "top" with "cluster"?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+	oldTopKubeVersion := old.Spec.KubernetesVersion
+
+	workerNodeGroupMap := make(map[string]*WorkerNodeGroupConfiguration)
+	for i := range old.Spec.WorkerNodeGroupConfigurations {
+		workerNodeGroupMap[old.Spec.WorkerNodeGroupConfigurations[i].Name] = &old.Spec.WorkerNodeGroupConfigurations[i]
+	}
+	for _, nodeGroupNewSpec := range new.Spec.WorkerNodeGroupConfigurations {
+		if workerNodeGrpOldSpec, ok := workerNodeGroupMap[nodeGroupNewSpec.Name]; ok {
```
vivek-koppuru, line 403: How do we measure skew if a new worker node group is added? We want to make sure it doesn't deviate too much from the cluster version right? Or is that not in this scope?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+}
+
+func performWorkerKubernetesValidations(oldVersion, newVersion *KubernetesVersion, oldTopKubeVersion, topKubeVersion KubernetesVersion) field.ErrorList {
+	var allErrs field.ErrorList
+	path := field.NewPath("spec").Child("WorkerNodeConfiguration.kubernetesVersion")
+
+	if oldVersion != nil && newVersion == nil {
+		allErrs = append(
+			allErrs,
+			validateRemoveWorkerKubernetesVersion(topKubeVersion, oldTopKubeVersion, newVersion, oldVersion)...,
```
vivek-koppuru, line 421: Why are we passing in newVersion to this method if we know it is nil and it is only being call in this case?

pkg/validations/upgradevalidations/preflightvalidations.go
```diff
@@ -77,6 +77,13 @@ func (u *UpgradeValidations) PreflightValidations(ctx context.Context) []validat
 				Err:         ValidateServerVersionSkew(ctx, u.Opts.Spec.Cluster, u.Opts.WorkloadCluster, u.Opts.ManagementCluster, k),
 			}
 		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name:        "upgrade cluster worker node group kubernetes version increment",
+				Remediation: "ensure that the cluster worker node group kubernetes version is incremented by one minor version exactly (e.g. 1.18 -> 1.19) and top level kubernetes version does not exceed worker node group version by one minor version",
```
vivek-koppuru, line 83: Skew is 2 and not 1 right?

pkg/providers/docker/docker.go
```diff
@@ -343,12 +343,14 @@ func NeedsNewControlPlaneTemplate(oldSpec, newSpec *cluster.Spec) bool {
...
+// NeedsNewWorkloadTemplate determines if a new workload template is needed.
+func NeedsNewWorkloadTemplate(oldSpec, newSpec *cluster.Spec, newWorker, oldWorker v1alpha1.WorkerNodeGroupConfiguration) bool {
 	if !v1alpha1.WorkerNodeGroupConfigurationSliceTaintsEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
-		!v1alpha1.WorkerNodeGroupConfigurationsLabelsMapEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) {
+		!v1alpha1.WorkerNodeGroupConfigurationsLabelsMapEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
+		!v1alpha1.WorkerNodeGroupConfigurationKubeVersionUnchanged(&newWorker, &oldWorker, newSpec.Cluster.Spec.KubernetesVersion, oldSpec.Cluster.Spec.KubernetesVersion) {
 		return true
 	}
-	return (oldSpec.Cluster.Spec.KubernetesVersion != newSpec.Cluster.Spec.KubernetesVersion) || (oldSpec.Bundles.Spec.Number != newSpec.Bundles.Spec.Number)
+	return oldSpec.Bundles.Spec.Number != newSpec.Bundles.Spec.Number
```
cxbrowne1207, line 353: Why was this removed? Do we not need to roll out a new workload template if the top level KubernetesVersion changes?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, newWorkerVersion, fmt.Sprintf("could not parse version: %v, %v", newCPVersion, err)))
+		return allErrs
+	}
+
+	if parsedOldVersion.LessThan(parsedNewVersion) {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, newWorkerVersion, fmt.Sprintf("can't simultaneously remove worker kubernetesVersion and upgrade top level kubernetesVersion: %v", newCPVersion)))
```
cxbrowne1207, line 480: What's the reason for not allowing removing and upgrading the top version simultaneous? Can't we treat a removal like an upgrade to the top version?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+
+	parsedNewVersion, err := version.ParseGeneric(string(newCPVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, newWorkerVersion, fmt.Sprintf("could not parse version: %v, %v", newCPVersion, err)))
+		return allErrs
+	}
+
+	if parsedOldVersion.LessThan(parsedNewVersion) {
```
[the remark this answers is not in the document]
tatlat: There is no Equal function and parsedNewVersion can only be greater than or equal to parsedOldVersion (these are both cluster level versions). The validations for the cluster level kubernetes version happens before worker nodes so if there was an issue at the cluster level version, then it would be caught there.

pkg/providers/docker/docker.go
tatlat: I replaced it with WorkerNodeGroupConfigurationKubeVersionUnchanged. I'm pretty sure we only want a new workload template if the worker node's kubernetes version changes. If a worker node has a nil kubernetesVersions, then the aforementioned method takes both old and new top level versions and uses them in place of the nil worker level versions.

pkg/api/v1alpha1/cluster_webhook.go
tatlat: We don't want to skip minor versions. If the old worker node version is at 1.19 and the old cluster version is 1.20 but the new cluster version is 1.21, then the worker node will try to jump from 1.19 -> 1.21 which has a skew of two minor versions. 

It's fine if the CP node is two minor versions above Worker Nodes, but we still want to upgrade nodes one minor version at a time.

pkg/api/v1alpha1/cluster_webhook.go
tatlat: We can use the newVersion's value as oldVersion

pkg/api/v1alpha1/cluster_webhook_test.go
```diff
@@ -2017,3 +2019,117 @@ func baseCluster(opts ...clusterOpt) *v1alpha1.Cluster {
 
 	return c
 }
+
+func TestValidateWorkerVersionSkew(t *testing.T) {
+	kube119 := v1alpha1.KubernetesVersion("1.19")
+	kube120 := v1alpha1.KubernetesVersion("1.20")
+	kube121 := v1alpha1.KubernetesVersion("1.21")
+	kubeBad := v1alpha1.KubernetesVersion("badvalue")
+	// kube123 := v1alpha1.KubernetesVersion("1.23")
```
jonathanmeier5, line 2028: Why is this commented out?

pkg/api/v1alpha1/cluster_webhook.go
vivek-koppuru: I see now okay

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,157 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+
+	return allErrs
+}
+
+func validateRemoveWorkerKubernetesVersion(newCPVersion, oldCPVersion KubernetesVersion, oldWorkerVersion *KubernetesVersion) field.ErrorList {
+	var allErrs field.ErrorList
+	path := field.NewPath("spec").Child("WorkerNodeConfiguration.kubernetesVersion")
+
+	parsedWorkerVersion, err := version.ParseGeneric(string(*oldWorkerVersion))
+	if err != nil {
```
jonathanmeier5, line 454: Is there a way we could trigger/test this error flow? Or alternatively, do we have a justification for not testing it? For instance, do we always control the value of this version param so we never expect an error?

pkg/api/v1alpha1/cluster_webhook.go
vivek-koppuru: However, can we explicitly have a case for it instead of make this assumption if we change anything in the validate method, for example add logging? Then we only need to check skew against CP nodes and not the validation for old vs new unnecessarily.

pkg/validations/upgradevalidations/versions.go
```diff
@@ -23,3 +23,18 @@ func ValidateServerVersionSkew(ctx context.Context, newCluster *anywherev1.Clust
...
+	if !cluster.ExistingManagement {
+		managementCluster = mgmtCluster
+	}
+
+	eksaCluster, err := kubectl.GetEksaCluster(ctx, managementCluster, newCluster.Name)
+	if err != nil {
+		return fmt.Errorf("fetching old cluster: %v", err)
+	}
+
+	return anywherev1.ValidateWorkerKubernetesVersionSkew(newCluster, eksaCluster).ToAggregate()
```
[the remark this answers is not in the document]
tatlat: yes

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
cxbrowne1207: Is this not the [Equal](https://github.com/tatlat/eks-anywhere/blob/49b4ae59fdd898f40d9c29e27be1908014b2570b/pkg/semver/semver.go#L67) function?

But yeah, it does not really matter, I was mainly wondering why `LessThan` over simply using the `Equal` comparison.

pkg/providers/docker/docker.go
cxbrowne1207: Yeah, that makes sense.

cxbrowne1207 (commented):
Are we going to add some e2e tests in another PR?

tatlat:
> Are we going to add some e2e tests in another PR?

yes

pkg/validations/upgradevalidations/preflightvalidations.go
```diff
@@ -77,6 +77,13 @@ func (u *UpgradeValidations) PreflightValidations(ctx context.Context) []validat
 				Err:         ValidateServerVersionSkew(ctx, u.Opts.Spec.Cluster, u.Opts.WorkloadCluster, u.Opts.ManagementCluster, k),
 			}
 		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name:        "upgrade cluster worker node group kubernetes version increment",
+				Remediation: "ensure that the cluster worker node group kubernetes version is incremented by one minor version exactly (e.g. 1.18 -> 1.19) and top level kubernetes version does not exceed worker node group version by two minor versions",
```
vivek-koppuru, line 83:
```suggestion
				Remediation: "ensure that the cluster worker node group kubernetes version is incremented by one minor version exactly (e.g. 1.18 -> 1.19) and cluster level kubernetes version does not exceed worker node group version by two minor versions",
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,168 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, newCPVersion, fmt.Sprintf("could not parse version: %v", err)))
+		return allErrs
+	}
+
+	if parsedOldClusterVersion.LessThan(parsedNewClusterVersion) {
+		allErrs = append(
+			allErrs,
+			field.Invalid(path, oldWorkerVersion, fmt.Sprintf("can't simultaneously remove worker kubernetesVersion and upgrade top level kubernetesVersion: %v", newCPVersion)))
```
vivek-koppuru, line 491:
```suggestion
			field.Invalid(path, oldWorkerVersion, fmt.Sprintf("can't simultaneously remove worker kubernetesVersion and upgrade cluster level kubernetesVersion: %v", newCPVersion)))
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -377,7 +382,168 @@ func ValidateKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
...
+		allErrs = append(
+			allErrs,
+			field.Invalid(workerPath, workerVersion, fmt.Sprintf("could not determine minor version difference: %v", err.Error())))
+		return allErrs
+	}
+
+	if !validSkew {
+		allErrs = append(
+			allErrs,
+			field.Invalid(cpPath, cpVersion, fmt.Sprintf("top level minor version must be within 2 versions greater than worker node group version: %v", workerVersion)))
```
vivek-koppuru, line 546:
```suggestion
			field.Invalid(cpPath, cpVersion, fmt.Sprintf("cluster level minor version must be within 2 versions greater than worker node group version: %v", workerVersion)))
```

vivek-koppuru (approved):

PR #6199, opened 2023-07-19
Eksaversion validations

tatlat:
*Issue #, if available:*

*Description of changes:*
Validations for EKSA Version field

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -27,7 +27,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"github.com/aws/eks-anywhere/pkg/features"
+	"github.com/aws/eks-anywhere/pkg/semver"
 )
+
+const supportedMinorVersionIncrement int64 = 1
 
 // log is for logging in this package.
 var clusterlog = logf.Log.WithName("cluster-resource")
@@ -94,6 +97,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 
 	allErrs = append(allErrs, ValidateKubernetesVersionSkew(r, oldCluster)...)
 
+	allErrs = append(allErrs, validateEksaVersionCluster(r, oldCluster)...)
+
+	allErrs = append(allErrs, ValidateEksaVersionSkew(r, oldCluster)...)
+
 	allErrs = append(allErrs, ValidateWorkerKubernetesVersionSkew(r, oldCluster)...)
 
 	if len(allErrs) != 0 {
@@ -111,14 +118,80 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 	return nil
 }
 
+// ValidateEksaVersionSkew ensures that upgrades are sequential by CLI minor versions.
+func ValidateEksaVersionSkew(new, old *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+	eksaVersionPath := field.NewPath("spec").Child("EksaVersion")
+
+	if new.Spec.EksaVersion == nil || old.Spec.EksaVersion == nil {
+		return nil
+	}
+
+	// allow users to update cluster if old cluster is invalid
+	oldEksaVersion, err := semver.New(string(*old.Spec.EksaVersion))
+	if err != nil {
+		return nil
+	}
+
+	newEksaVersion, err := semver.New(string(*new.Spec.EksaVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "EksaVersion is not a valid semver"))
+		return allErrs
+	}
+
+	majorVersionDifference := int64(newEksaVersion.Major) - int64(oldEksaVersion.Major)
+	minorVersionDifference := int64(newEksaVersion.Minor) - int64(oldEksaVersion.Minor)
+
+	// if major different or upgrade difference greater than one minor version
+	if majorVersionDifference != 0 || minorVersionDifference > supportedMinorVersionIncrement {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, fmt.Sprintf("cannot upgrade to %v from %v: EksaVersion upgrades must have a skew of 1 minor version", newEksaVersion, oldEksaVersion)))
+		return allErrs
+	}
+
+	// Allow "downgrades" if old version is greater than managment cluster. We can't check if a cluster's EksaVersion
+	// is less than or equal to the mgmt cluster in the webhook. Instead we check it in the controller where the
+	// EksaVersion will already be applied. However, the cluster will never begin to reconcile due to the validation.
+	// We should not block users from changing EksaVersion to a lower semver in this scenario.
+	failure := old.Status.FailureReason
+	if failure != nil && *failure == EksaVersionInvalidReason {
+		return nil
+	}
+
+	// don't allow downgrades if old version was valid
+	if minorVersionDifference < 0 {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, fmt.Sprintf("cannot downgrade from %v to %v: EksaVersion upgrades must be incremental", oldEksaVersion, newEksaVersion)))
+	}
+
+	return allErrs
+}
+
 func validateBundlesRefCluster(new, old *Cluster) field.ErrorList {
 	var allErrs field.ErrorList
 	bundlesRefPath := field.NewPath("spec").Child("BundlesRef")
 
-	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef == nil {
+	if old.Spec.BundlesRef != nil && new.Spec.BundlesRef == nil && new.Spec.EksaVersion == nil {
 		allErrs = append(
 			allErrs,
 			field.Invalid(bundlesRefPath, new.Spec.BundlesRef, fmt.Sprintf("field cannot be removed after setting. Previous value %v", old.Spec.BundlesRef)))
+	}
+
+	return allErrs
+}
+
+func validateEksaVersionCluster(new, old *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+	eksaVersionPath := field.NewPath("spec").Child("EksaVersion")
+
+	if old.Spec.EksaVersion != nil && new.Spec.EksaVersion == nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, fmt.Sprintf("field cannot be removed after setting. Previous value %v", old.Spec.EksaVersion)))
 	}
 
 	return allErrs
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -56,21 +56,13 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
 
 // BuildSpecFromConfig constructs a cluster.Spec for an eks-a cluster config by retrieving all dependencies objects from the cluster using a kubernetes client.
 func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*Spec, error) {
-	eksaRelease := &v1alpha1release.EKSARelease{}
-	if config.Cluster.Spec.BundlesRef == nil {
-		if config.Cluster.Spec.EksaVersion == nil {
-			return nil, fmt.Errorf("either cluster's EksaVersion or BundlesRef need to be set")
-		}
-		version := string(*config.Cluster.Spec.EksaVersion)
-		eksaReleaseName := v1alpha1release.GenerateEKSAReleaseName(version)
-		if err := client.Get(ctx, eksaReleaseName, constants.EksaSystemNamespace, eksaRelease); err != nil {
-			return nil, fmt.Errorf("error getting EKSARelease %s", eksaReleaseName)
-		}
+	eksaRelease, err := eksaReleaseForCluster(ctx, client, config.Cluster)
+	if err != nil {
+		return nil, err
 	}
 
-	bundlesName, bundlesNamespace := bundlesNamespacedKey(config.Cluster, eksaRelease)
-	bundles := &v1alpha1release.Bundles{}
-	if err := client.Get(ctx, bundlesName, bundlesNamespace, bundles); err != nil {
+	bundles, err := bundlesForEksaRelease(ctx, client, config.Cluster, eksaRelease)
+	if err != nil {
 		return nil, err
 	}
 
@@ -87,4 +79,40 @@ func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*S
 	}
 
 	return NewSpec(config, bundles, eksdRelease, eksaRelease)
+}
+
+// BundlesForCluster returns a bundles resource for the cluster.
+func BundlesForCluster(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*v1alpha1release.Bundles, error) {
+	release, err := eksaReleaseForCluster(ctx, client, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return bundlesForEksaRelease(ctx, client, cluster, release)
+}
+
+func eksaReleaseForCluster(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*v1alpha1release.EKSARelease, error) {
+	eksaRelease := &v1alpha1release.EKSARelease{}
+	if cluster.Spec.BundlesRef == nil {
+		if cluster.Spec.EksaVersion == nil {
+			return nil, fmt.Errorf("either cluster's EksaVersion or BundlesRef need to be set")
+		}
+		version := string(*cluster.Spec.EksaVersion)
+		eksaReleaseName := v1alpha1release.GenerateEKSAReleaseName(version)
+		if err := client.Get(ctx, eksaReleaseName, constants.EksaSystemNamespace, eksaRelease); err != nil {
+			return nil, fmt.Errorf("error getting EKSARelease %s", eksaReleaseName)
+		}
+	}
+
+	return eksaRelease, nil
+}
+
+func bundlesForEksaRelease(ctx context.Context, client Client, cluster *v1alpha1.Cluster, eksaRelease *v1alpha1release.EKSARelease) (*v1alpha1release.Bundles, error) {
+	bundlesName, bundlesNamespace := bundlesNamespacedKey(cluster, eksaRelease)
+	bundles := &v1alpha1release.Bundles{}
+	if err := client.Get(ctx, bundlesName, bundlesNamespace, bundles); err != nil {
+		return nil, err
+	}
+
+	return bundles, nil
 }
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -18,11 +18,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	packagesv1 "github.com/aws/eks-anywhere-packages/api/v1alpha1"
-	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/registrymirror"
@@ -46,7 +46,7 @@ type PackageControllerClient struct {
 	// chartManager installs and deletes helm charts.
 	chartManager          ChartManager
 	clusterName           string
-	clusterSpec           *v1alpha1.ClusterSpec
+	clusterSpec           *anywherev1.ClusterSpec
 	managementClusterName string
 	kubectl               KubectlRunner
 	eksaAccessKeyID       string
@@ -399,7 +399,7 @@ func formatYamlLine(space, key, value string) string {
 	return space + key + ": " + value + "\n"
 }
 
-func formatImageResource(resource *v1alpha1.ImageResource, name string) (result string) {
+func formatImageResource(resource *anywherev1.ImageResource, name string) (result string) {
 	if resource.CPU != "" || resource.Memory != "" {
 		result = "    " + name + ":\n"
 		result += formatYamlLine("      ", "cpu", resource.CPU)
@@ -408,7 +408,7 @@ func formatImageResource(resource *v1alpha1.ImageResource, name string) (result 
 	return result
 }
 
-func formatCronJob(cronJob *v1alpha1.PackageControllerCronJob) (result string) {
+func formatCronJob(cronJob *anywherev1.PackageControllerCronJob) (result string) {
 	if cronJob != nil {
 		result += "cronjob:\n"
 		result += formatYamlLine("  ", "digest", cronJob.Digest)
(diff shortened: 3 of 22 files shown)


pkg/controller/clusters/validations.go
```diff
@@ -54,3 +57,57 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+
+	if err := v.client.Get(ctx, mgmtClusterKey, mgmtCluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			err := fmt.Errorf("unable to retrieve management cluster %v: %v", cluster.Spec.ManagementCluster.Name, err)
+			log.Error(err, "Invalid cluster configuration")
+			return err
+		}
+	}
+
+	if mgmtCluster.Spec.EksaVersion == nil {
```
g-gaston, line 77: I think this is valid, right?
while the management cluster is upgraded, this will be nil

pkg/controller/clusters/validations.go
```diff
@@ -54,3 +57,57 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+		return err
+	}
+
+	mVersion, err := semver.New(string(*mgmtCluster.Spec.EksaVersion))
+	if err != nil {
+		log.Error(err, "Management cluster has invalid EksaVersion")
+		return err
+	}
+
+	if cluster.Spec.EksaVersion == nil {
```
g-gaston, line 89: why is this an error? what about clusters with bundlesRef

pkg/controller/clusters/validations.go
```diff
@@ -54,3 +57,57 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+	wVersion, err := semver.New(string(*cluster.Spec.EksaVersion))
+	if err != nil {
+		log.Error(err, "Workload cluster has invalid EksaVersion")
+		return err
+	}
+
+	if wVersion.GreaterThan(mVersion) {
+		err := fmt.Errorf("cannot upgrade workload cluster with version %v while management cluster is an older version %v", wVersion, mVersion)
+		log.Error(err, "Invalid cluster configuration")
+		return err
```
g-gaston, line 104: set cluster failure reason and message here?

pkg/controller/clusters/validations.go
```diff
@@ -54,3 +57,57 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+	}
+
+	// reset failure message if old matches this validation
+	oldFailure := cluster.Status.FailureMessage
+	if oldFailure != nil && strings.Contains(*oldFailure, "cannot upgrade workload cluster with version") {
+		cluster.Status.FailureMessage = ptr.String("")
```
g-gaston, line 110: just set to nil?
also failure reason
I think there is a method to clear them

pkg/controller/clusters/validations.go
```diff
@@ -54,3 +57,57 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+
+	if wVersion.GreaterThan(mVersion) {
+		err := fmt.Errorf("cannot upgrade workload cluster with version %v while management cluster is an older version %v", wVersion, mVersion)
+		log.Error(err, "Invalid cluster configuration")
+		return err
+	}
+
+	// reset failure message if old matches this validation
+	oldFailure := cluster.Status.FailureMessage
+	if oldFailure != nil && strings.Contains(*oldFailure, "cannot upgrade workload cluster with version") {
```
g-gaston, line 109: prefer to use failure reson

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -483,10 +483,31 @@ func (pc *PackageControllerClient) Reconcile(ctx context.Context, logger logr.Lo
 // getBundleFromCluster based on the cluster's k8s version.
 func (pc *PackageControllerClient) getBundleFromCluster(ctx context.Context, client client.Client, clusterObj *anywherev1.Cluster) (*releasev1.Image, error) {
 	bundles := &releasev1.Bundles{}
-	nn := types.NamespacedName{
-		Name:      clusterObj.Spec.BundlesRef.Name,
-		Namespace: clusterObj.Spec.BundlesRef.Namespace,
+
+	nn := types.NamespacedName{}
+
+	if clusterObj.Spec.BundlesRef == nil {
```
g-gaston, line 489: can we leverage the code we have in the cluster package for this? it seems this is duplicated

pkg/executables/kubectl.go
```diff
@@ -1517,6 +1522,23 @@ func (k *Kubectl) GetEksaCluster(ctx context.Context, cluster *types.Cluster, cl
 	return response, nil
 }
 
+// GetEksaClusters gets all EKSA clusters.
+func (k *Kubectl) GetEksaClusters(ctx context.Context, cluster *types.Cluster) ([]*v1alpha1.Cluster, error) {
+	params := []string{"get", eksaClusterResourceType, "-A", "--kubeconfig", cluster.KubeconfigFile, "-o", "json"}
```
g-gaston, line 1527: why don't you leverage the List method for this?

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
 
 	return nil
 }
+
+// ValidateEksaVersion ensures that the version matches EKS-A CLI.
+func ValidateEksaVersion(ctx context.Context, k KubectlClient, cliVersion string, workload *cluster.Spec) error {
```
g-gaston, line 123: is kubectl needed here?

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	v := workload.Cluster.Spec.EksaVersion
+
+	if cliVersion == "" {
```
g-gaston, line 126: if this is an argument already, why do this?

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+		cliVersion = version.Get().GitVersion
+	}
+
+	if v != nil {
```
g-gaston, line 130: just return if it's nil

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+		if !parsedVersion.SamePatch(parsedCLIVersion) {
+			return fmt.Errorf("cluster's eksaVersion does not match EKS-A")
+		}
+	}
+	return nil
+}
+
+// ValidateEksaVersionSkew ensures that upgrades are sequential by CLI minor versions.
+func ValidateEksaVersionSkew(ctx context.Context, k KubectlClient, cluster *types.Cluster, workload *cluster.Spec) error {
+	c, err := k.GetEksaCluster(ctx, cluster, cluster.Name)
```
g-gaston, line 150: is this the management cluster? if so, can we use the ref in the workload cluster?

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+
+		if !parsedVersion.SamePatch(parsedCLIVersion) {
+			return fmt.Errorf("cluster's eksaVersion does not match EKS-A")
+		}
+	}
+	return nil
+}
+
+// ValidateEksaVersionSkew ensures that upgrades are sequential by CLI minor versions.
+func ValidateEksaVersionSkew(ctx context.Context, k KubectlClient, cluster *types.Cluster, workload *cluster.Spec) error {
```
g-gaston, line 149: workload -> spec

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	c, err := k.GetEksaCluster(ctx, cluster, cluster.Name)
+	if err != nil {
+		return err
+	}
+
+	v := c.Spec.EksaVersion
+
+	// try getting version from bundle's eksa component's cluster controller image
+	if v == nil {
+		version, err := getVersionFromBundle(ctx, k, c.Spec.BundlesRef, cluster)
```
g-gaston, line 159: i would vote to not do this
if the management cluster doesn't yet have a eksaVersion, then we don't validate this

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	minorVersionDifference := float64(parsedUpgradeVersion.Minor) - float64(parsedClusterVersion.Minor)
+	supportedMinorVersionIncrement := float64(1)
+
+	if majorVersionDifference > 0 || !(minorVersionDifference <= supportedMinorVersionIncrement && minorVersionDifference >= 0) {
+		msg := fmt.Sprintf("WARNING: version difference between upgrade version (%d.%d) and cluster version (%d.%d) do not meet the supported version increment of +%f",
+			parsedUpgradeVersion.Major, parsedUpgradeVersion.Minor, parsedClusterVersion.Major, parsedClusterVersion.Minor, supportedMinorVersionIncrement)
+		return fmt.Errorf(msg)
+	}
+
+	return nil
```
g-gaston, line 191: all this logic is duplicated in the webhook
can we reuse?

pkg/validations/cluster.go
```diff
@@ -114,3 +118,132 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	}
+
+	wVersion, err := semver.New(string(*workload.Cluster.Spec.EksaVersion))
+	if err != nil {
+		return fmt.Errorf("workload cluster eksaVersion is invalid: %w", err)
+	}
+
+	if wVersion.GreaterThan(mVersion) {
+		return fmt.Errorf("cannot upgrade workload cluster with version %v while management cluster is an older version %v", wVersion, mVersion)
+	}
```
g-gaston, line 246: it also seems like this logic is duplicated in the controller
any chance we can reuse the code?

pkg/validations/skipvalidations.go
```diff
@@ -9,6 +9,7 @@ import (
 const (
 	PDB             = "pod-disruption"
 	VSphereUserPriv = "vsphere-user-privilege"
+	EVS             = "eksa-version-skew"
```
g-gaston, line 12: can we just name this properly? no one else is going to know what EVS is

pkg/validations/upgradevalidations/preflightvalidations.go
```diff
@@ -91,18 +91,26 @@ func (u *UpgradeValidations) PreflightValidations(ctx context.Context) []validat
 				Err:         ValidateImmutableFields(ctx, k, targetCluster, u.Opts.Spec, u.Opts.Provider),
 			}
 		},
+		func() *validations.ValidationResult {
+			return &validations.ValidationResult{
+				Name:        "validate eksa version matches cli",
+				Remediation: "",
```
g-gaston, line 97: this one should be easy :)

pkg/validations/upgradevalidations/preflightvalidations_test.go
```diff
@@ -385,7 +389,7 @@ func TestPreflightValidationsTinkerbell(t *testing.T) {
 			k.EXPECT().ValidateNodes(ctx, kubeconfigFilePath).Return(tc.nodeResponse)
 			k.EXPECT().ValidateClustersCRD(ctx, workloadCluster).Return(tc.crdResponse)
 			k.EXPECT().GetClusters(ctx, workloadCluster).Return(tc.getClusterResponse, nil)
-			k.EXPECT().GetEksaCluster(ctx, workloadCluster, clusterSpec.Cluster.Name).Return(existingClusterSpec.Cluster, nil).Times(2)
+			k.EXPECT().GetEksaCluster(ctx, workloadCluster, clusterSpec.Cluster.Name).Return(existingClusterSpec.Cluster, nil).AnyTimes()
```
g-gaston, line 392: why anytimes?

pkg/validations/kubectl.go
```diff
@@ -25,6 +25,7 @@ type KubectlClient interface {
 	Version(ctx context.Context, cluster *types.Cluster) (*executables.VersionResponse, error)
 	GetClusters(ctx context.Context, cluster *types.Cluster) ([]types.CAPICluster, error)
 	GetEksaCluster(ctx context.Context, cluster *types.Cluster, clusterName string) (*v1alpha1.Cluster, error)
+	GetEksaClusters(ctx context.Context, cluster *types.Cluster) ([]*v1alpha1.Cluster, error)
```
g-gaston, line 28: where is this being used? sorry if I missed it

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -113,11 +120,56 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
 }
 
+func validateEksaVersionSkew(new, oldCluster *Cluster) field.ErrorList {
+	var allErrs field.ErrorList
+	eksaVersionPath := field.NewPath("spec").Child("EksaVersion")
+
+	if new.Spec.EksaVersion != nil && oldCluster.Spec.EksaVersion != nil {
+		parsedClusterVersion, err := semver.New(string(*oldCluster.Spec.EksaVersion))
+		if err != nil {
+			return allErrs
```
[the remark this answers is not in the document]
tatlat: If the old version is invalid and they're trying to change it, does it matter? I don't think we should block an upgrade because the old value is invalid.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -113,11 +120,56 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+		parsedUpgradeVersion, err := semver.New(string(*new.Spec.EksaVersion))
+		if err != nil {
+			allErrs = append(
+				allErrs,
+				field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "upgrade cluster is invalid"))
+		}
+
+		majorVersionDifference := math.Abs(float64(parsedUpgradeVersion.Major) - float64(parsedClusterVersion.Major))
+		minorVersionDifference := float64(parsedUpgradeVersion.Minor) - float64(parsedClusterVersion.Minor)
+		supportedMinorVersionIncrement := float64(1)
```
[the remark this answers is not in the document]
tatlat: They are uint64, so minorVersionDifference would never be negative without the type conversion.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -113,11 +120,56 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+
+		// if major different or upgrade difference greater than one minor version
+		if majorVersionDifference > 0 || minorVersionDifference > supportedMinorVersionIncrement {
+			allErrs = append(
+				allErrs,
+				field.Invalid(eksaVersionPath, new.Spec.EksaVersion, fmt.Sprintf("cannot upgrade to %v from %v: EksaVersion upgrades must be sequential in minor versions", parsedUpgradeVersion, parsedClusterVersion)))
+		}
+
+		// allow "downgrades" if old version is greater than managment cluster
```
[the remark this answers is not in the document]
tatlat: If the mgmt cluster has v0.0.0 and you apply a workload with v0.1.0, then the change will get past the webhook, but it will fail at preClusterProviderReconcile because we can't have a workload with version greater than mgmt. 
If you try to correct the version (from 0.1.0 to 0.0.0), then webhook will block it because we don't want downgrades. 
But I think we should allow downgrades in this scenario where a successful upgrade never happened.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -113,11 +120,56 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+		failure := oldCluster.Status.FailureMessage
+		if failure != nil && strings.Contains(*failure, "cannot upgrade workload cluster with version") {
```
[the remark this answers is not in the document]
tatlat: I tested this and it worked for me. However, I understand if this is a bad practice.

pkg/controller/clusters/validations.go
tatlat: So I should return nil here? No error?

pkg/validations/cluster.go
tatlat: No, it's not the managment cluster.

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
tatlat: float is needed for absolute value, but i did change minorVersionDifference to an int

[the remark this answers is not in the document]
g-gaston: ok, what about this
```go
majorVersionDifference := int64(parsedUpgradeVersion.Major) - int64(parsedClusterVersion.Major)
	minorVersionDifference := int64(parsedUpgradeVersion.Minor) - int64(parsedClusterVersion.Minor)
	var supportedMinorVersionIncrement int64 = 1

File: pkg/api/v1alpha1/cluster_webhook.go [hunk 24]
```diff
@@ -115,11 +122,66 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+
+	if new.Spec.EksaVersion == nil || oldCluster.Spec.EksaVersion == nil {
+		return nil
+	}
+
+	oldEksaVersion, err := semver.New(string(*oldCluster.Spec.EksaVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "old cluster is invalid"))
```
g-gaston, line 137:
```suggestion
			field.Invalid(eksaVersionPath, old.Spec.EksaVersion, "old cluster is invalid"))
```

However, should we error out here? If the old cluster is invalid, the user will never be able to update it.

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -115,11 +122,66 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+		return allErrs
+	}
+
+	newEksaVersion, err := semver.New(string(*new.Spec.EksaVersion))
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "upgrade cluster is invalid"))
```
g-gaston, line 145:
```suggestion
			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "EksaVersion is not a valid semver"))
```

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -115,11 +122,66 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, fmt.Sprintf("cannot upgrade to %v from %v: EksaVersion upgrades must have a skew of 1 minor version", newEksaVersion, oldEksaVersion)))
+		return allErrs
+	}
+
+	// Allow "downgrades" if old version is greater than managment cluster. We can't check if a cluster's EksaVersion
+	// is less than or equal to the mgmt cluster in the webhook. Instead we check it in the controller where the
+	// EksaVersion will already be applied. However, the cluster will never begin to reconcile due to the validation.
+	// We should not block users from changing EksaVersion to a lower semver in this scenario.
+	failure := oldCluster.Status.FailureReason
+	if failure != nil && strings.Contains(string(*failure), "cannot upgrade workload cluster with version") {
```
g-gaston, line 166: isn't the failure reason a const? why do we need a contains?

pkg/cluster/fetch.go
```diff
@@ -11,6 +11,20 @@ import (
...
+
+	// EKSAReleaseFetcher is used to fetch a bundles resource.
+	EKSAReleaseFetcher func(string) (*v1alpha1release.EKSARelease, error)
+
+	// Fetchers contains fetch functions for resources.
+	Fetchers struct {
+		BundlesFetcher     BundlesFetcher
+		EKSAReleaseFetcher EKSAReleaseFetcher
+	}
+)
```
g-gaston, line 26: I thought we removed all this stuff?

pkg/api/v1alpha1/cluster_webhook.go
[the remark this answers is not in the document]
tatlat: yes, i can do this

pkg/executables/kubectl.go
```diff
@@ -1167,6 +1167,11 @@ type TinkerbellMachineConfigResponse struct {
 	Items []*v1alpha1.TinkerbellMachineConfig `json:"items,omitempty"`
 }
 
+// ClusterResponse contains list of EKS-A Cluster.
+type ClusterResponse struct {
```
g-gaston, line 1171: is this needed?

pkg/validations/cluster.go
```diff
@@ -114,3 +119,110 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	mgmt, err := k.GetEksaCluster(ctx, mgmtCluster, mgmtCluster.Name)
+	if err != nil {
+		return err
+	}
+
+	return ValidateManagementEksaVersion(ctx, logger.Get(), mgmt, workload.Cluster)
+}
+
+// ValidateManagementEksaVersion ensures a workload cluster's EksaVersion is not greater than a management cluster's version.
+func ValidateManagementEksaVersion(ctx context.Context, log logr.Logger, mgmtCluster, cluster *v1alpha1.Cluster) error {
```
g-gaston, line 169: What do we need the ctx or the logger for?

pkg/validations/cluster.go
```diff
@@ -114,3 +119,110 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+	if wVersion.GreaterThan(mVersion) {
+		message := fmt.Sprintf("cannot upgrade workload cluster with version %v while management cluster is an older version %v", wVersion, mVersion)
+		cluster.Status.FailureMessage = ptr.String(message)
+		reason := v1alpha1.FailureReasonType(message)
+		cluster.Status.FailureReason = &reason
+		return fmt.Errorf(message)
+	}
+
+	// reset failure message if old matches this validation
+	oldFailure := cluster.Status.FailureReason
```
g-gaston, line 188: same question here about the the reason being a const

pkg/validations/kubectl.go
```diff
@@ -25,6 +25,7 @@ type KubectlClient interface {
 	Version(ctx context.Context, cluster *types.Cluster) (*executables.VersionResponse, error)
 	GetClusters(ctx context.Context, cluster *types.Cluster) ([]types.CAPICluster, error)
 	GetEksaCluster(ctx context.Context, cluster *types.Cluster, clusterName string) (*v1alpha1.Cluster, error)
+
```
g-gaston, line 28: undo?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -115,11 +121,67 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.Invalid(eksaVersionPath, new.Spec.EksaVersion, "upgrade cluster is invalid"))
+		return allErrs
+	}
+
+	majorVersionDifference := int64(newEksaVersion.Major) - int64(oldEksaVersion.Major)
+	minorVersionDifference := int64(newEksaVersion.Minor) - int64(oldEksaVersion.Minor)
+	const supportedMinorVersionIncrement int64 = 1
```
g-gaston, line 151: make the const a package level const pls

pkg/api/v1alpha1/cluster_webhook.go
tatlat: I don't think it should error out.

pkg/validations/cluster.go
```diff
@@ -114,3 +116,110 @@ func ValidateManagementClusterBundlesVersion(ctx context.Context, k KubectlClien
...
+		return fmt.Errorf("parsing cluster eksa version: %v", err)
+	}
+
+	parsedCLIVersion, err := semver.New(cliVersion)
+	if err != nil {
+		return fmt.Errorf("parsing eksa cli version: %v", err)
+	}
+
+	if !parsedVersion.SamePatch(parsedCLIVersion) {
+		return fmt.Errorf("cluster's eksaVersion does not match EKS-A")
```
g-gaston, line 139:
```suggestion
		return fmt.Errorf("cluster's eksaVersion does not match EKS-A CLI's version")
```

g-gaston (commented):
/lgtm

PR #6201, opened 2023-07-19
Fix for Default CNI condition in Cluster status

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
When a workload cluster control plane is upgraded, it is marked "False" with the `ControlPlaneNotReady`. It is not re-evaluated after the control plane becomes ready, as the controller stops reconciling before then. So, this PR addresses the issue by removing the condition update's dependency on ControlPlaneReady. We only need to change this status when there is a. change in the state of the CNI detected.

*Testing (if applicable):*
- Unit test
- Manual test
   - Create a workload cluster using the controller, then upgrade the control plane node count. The `DefaultCNIConfiguredCondition` should stay "True".

### DIFF
--- a/pkg/controller/clusters/status_test.go
+++ b/pkg/controller/clusters/status_test.go
@@ -716,7 +716,7 @@ func TestUpdateClusterStatusForCNI(t *testing.T) {
 			},
 		},
 		{
-			name:        "control plane is ready, default cni initialized",
+			name:        "control plane is not ready, default cni initialized",
 			skipUpgrade: ptr.Bool(false),
 			conditions: []anywherev1.Condition{
 				{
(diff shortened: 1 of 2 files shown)


pkg/controller/clusters/status_test.go
```diff
@@ -715,6 +715,24 @@ func TestUpdateClusterStatusForCNI(t *testing.T) {
...
+			name:        "control plane is ready, default cni initialized",
+			skipUpgrade: ptr.Bool(false),
+			conditions: []anywherev1.Condition{
+				{
+					Type:   anywherev1.DefaultCNIConfiguredCondition,
+					Status: "True",
+				},
+				{
+					Type:   anywherev1.ControlPlaneReadyCondition,
+					Status: "True",
```
g-gaston, line 728: I think this should be false if you want to test the new code

cxbrowne1207: ah, you're right. thanks for the catch

g-gaston (commented):
/lgtm

PR #6204, opened 2023-07-19
Fix unnecessary KCP spec update when reconciler ControlPlane

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
When the controller reconciles the control plane for a cluster with an external ETCD configuration, the `KubeadmControlPlane.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.External.Endpoints` field is defaulted to an empty slice. However,  at some point that field in KubeadmControlPlane object is filled and updated by another component 

During subsequent reconciliation loops, the controller applies a spec with an empty slice again, causing unnecessary updates to the `.metadata.generation` field, which is integral for tracking that object's status is up to date through it's `.status.observedGeneration` field. These unnecessary updates causes issues with getting correct timely status updates from the KubeadmControlPlane and is also causes the `.metadata.generation` and `.status.observedGeneration` values to balloon quickly.

This PR cleans up these unnecessary updates by overriding the external etcd endpoints on the `KubeadmControlPlane` object generated by the controller with the existing ones from the object in the cluster before applying it.  

*Testing (if applicable):*
- Unit testing
- Manual testing

### DIFF
--- a/pkg/controller/clusters/controlplane.go
+++ b/pkg/controller/clusters/controlplane.go
@@ -101,6 +101,25 @@ func ReconcileControlPlane(ctx context.Context, c client.Client, cp *ControlPlan
 		clientutil.AddAnnotation(cp.KubeadmControlPlane, clusterv1.PausedAnnotation, "true")
 	}
 
+	// When the controller reconciles the control plane for a cluster with an external etcd configuration
+	// the KubeadmControlPlane.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.External.Endpoints field is
+	// defaulted to an empty slice. However, at some point that field in KubeadmControlPlane object is filled
+	// and updated by another component
+
+	// We do not want to update the field with an empty slice again, so here we check if the endpoints for the
+	// external etcd have already been populated on the KubeadmControlPlane object and override ours before applying it.
+	kcp := &controlplanev1.KubeadmControlPlane{}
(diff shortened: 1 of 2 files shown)


pkg/controller/clusters/controlplane.go
```diff
@@ -101,6 +101,25 @@ func ReconcileControlPlane(ctx context.Context, c client.Client, cp *ControlPlan
...
 
+	// When the controller reconciles the control plane for a cluster with an external etcd configuration
+	// the KubeadmControlPlane.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.External.Endpoints field is
+	// defaulted to an empty slice. However, at some point that field in KubeadmControlPlane object is filled
+	// and updated by another component
+
+	// We do not want to update the field with an empty slice again, so here we check if the endpoints for the
+	// external etcd have already been populated on the KubeadmControlPlane object and override ours before applying it.
+	kcp := &controlplanev1.KubeadmControlPlane{}
+	kcpKey := client.ObjectKeyFromObject(cp.KubeadmControlPlane)
```
g-gaston, line 112: we already have the capi cluster
can we use the reference it has?

cxbrowne1207: yes, we can

g-gaston (commented):
/lgtm

PR #6205, opened 2023-07-19
Update docs development README guidance

chrisdoherty4:
This change improves the developers experience when wanting to detach the container hosting the docs site for local development and provies guidance on how to manage the submodules for the docs site.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,9 +1,17 @@
 DOCS_FOLDER=$(shell git rev-parse --show-toplevel)/docs
 IMAGE_REGISTRY ?= public.ecr.aws/eks-distro-build-tooling
 CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/eks-a-hugo
-CONTAINER_RUN     = docker run --rm --interactive --tty --volume $(CURDIR)/../:/src
 DOCS_ARCHIVE      = public.zip
 HTMLTEST_VERSION?=v0.16.0
+
+# Detach can be used with the container-serve command to run the container in the background.
+DETACH ?= false
+
+# _FOCUS_OPTS describes the options for focusing the serve container (background vs foreground).
+_FOCUS_OPTS := --interactive --tty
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -1,10 +1,18 @@
 DOCS_FOLDER=$(shell git rev-parse --show-toplevel)/docs
 IMAGE_REGISTRY ?= public.ecr.aws/eks-distro-build-tooling
 CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/eks-a-hugo
-CONTAINER_RUN     = docker run --rm --interactive --tty --volume $(CURDIR)/../:/src
 DOCS_ARCHIVE      = public.zip
 HTMLTEST_VERSION?=v0.16.0
 
+# Detach can be used with the container-serve command to
```
drewvanstone, line 7: to what?

drewvanstone (approved):
Looks like one comment was incomplete in the makefile but otherwise looks good.

docs/README.md
```diff
@@ -15,17 +15,14 @@ make container-serve
...
-tmux
 make container-build
-make container-serve
-Ctrl-b d
+make container-serve DETACH=true
 ```
-This runs the commands in a tmux session, then detaches it from the current shell.
-To reopen the `tmux` session later, type `tmux attach`.
+
+You may notice an update to the `/docs/themes/docsy` submodule which is the result of a patch. Before committing your changes, you can reset the submodule using `make submodule-reset`.
```
chrisnegus, line 25: Could you also mention that, in any case, they should not commit changes to the /docs/themes/docsy directory?

chrisnegus (commented):
I have just one comment.

docs/Makefile
chrisnegus: Somethin like this?:  ...detach the container from the current shell and run it in the background.

chrisnegus (approved):
/lgtm

PR #6214, opened 2023-07-20
add helper function for Tinkerbell specific workflow for curated packages harbor flow

rahulbabu95:
*Issue #, if available:*
Curated Packages Harbor flow is failing for Tink CI as provider specific workflow is not addressed by the generic harbor simple flow function. This change adds a helper function to address the provider specific workflow for Tinkerbell which is similar to other packages test for Tinkerbell provider.
*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/test/e2e/harbor.go
+++ b/test/e2e/harbor.go
@@ -29,13 +29,31 @@ func runCuratedPackageHarborInstallSimpleFlowLocalStorageProvisioner(test *frame
 	})
 }
 
+func runCuratedPackageHarborInstall(test *framework.ClusterE2ETest) {
+	test.InstallLocalStorageProvisioner()
+	packagePrefix := "test"
+	installNs := "harbor"
+	test.CreateNamespace(installNs)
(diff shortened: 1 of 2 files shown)


test/e2e/harbor.go
```diff
@@ -28,3 +28,15 @@ func runCuratedPackageHarborInstallSimpleFlowLocalStorageProvisioner(test *frame
...
 	})
 }
+
+func runCuratedPackageHarborInstallTinkerbellSimpleFlow(test *framework.ClusterE2ETest) {
+	test.GenerateClusterConfig()
+	test.GenerateHardwareConfig()
+	test.PowerOnHardware()
+	test.CreateCluster(framework.WithControlPlaneWaitTimeout("20m"))
+	test.ValidateControlPlaneNodes(framework.ValidateControlPlaneNoTaints, framework.ValidateControlPlaneLabels)
+	runCuratedPackageHarborInstallSimpleFlowLocalStorageProvisioner(test)
```
pokearu, line 38: I dont think we can reuse this in this manner. This has a test.WithCluster() which i think internally also does a CreateCluster.

We might have to take what this function does and do it here.

PR #6223, opened 2023-07-20
Add Ubuntu 22.04 E2E tests

abhay-krishna:
* Supporting multiple OS versions in E2E tests with backward compatibility.
* Adding vSphere/Tinkerbell tests for Ubuntu 22.04.


cmd/integration_test/build/buildspecs/tinkerbell-test-eks-a-cli.yml
```diff
@@ -37,6 +37,12 @@ env:
     T_TINKERBELL_IMAGE_UBUNTU_1_25: "tinkerbell_ci:image_ubuntu_1_25"
     T_TINKERBELL_IMAGE_UBUNTU_1_26: "tinkerbell_ci:image_ubuntu_1_26"
     T_TINKERBELL_IMAGE_UBUNTU_1_27: "tinkerbell_ci:image_ubuntu_1_27"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_22: "tinkerbell_ci:image_ubuntu_2204_1_22"
```
abhinavmpandey08, line 40: 1.22 isn't supported anymore. We don't need to include this

cmd/integration_test/build/buildspecs/vsphere-test-eks-a-cli.yml
```diff
@@ -14,6 +14,12 @@ env:
     T_VSPHERE_TEMPLATE_UBUNTU_1_25: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-25"
     T_VSPHERE_TEMPLATE_UBUNTU_1_26: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-26"
     T_VSPHERE_TEMPLATE_UBUNTU_1_27: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-27"
+    T_VSPHERE_TEMPLATE_UBUNTU_2204_1_22: "/SDDC-Datacenter/vm/Templates/ubuntu-2204-kube-v1-22"
```
abhinavmpandey08, line 17: Same as above 1.22 isn't supported anymore. We don't need to include this

cmd/integration_test/build/buildspecs/tinkerbell-test-eks-a-cli.yml
abhay-krishna: Ahh I just copied the existing Ubuntu 2004 vars which included 1-22, I'll remove both.

cmd/integration_test/build/buildspecs/vsphere-test-eks-a-cli.yml
abhay-krishna: Done!

test/e2e/tinkerbell_test.go
```diff
@@ -1516,6 +1516,61 @@ func TestTinkerbellKubernetes127UbuntuSimpleFlow(t *testing.T) {
 	runTinkerbellSimpleFlow(test)
 }
 
+func TestTinkerbellKubernetes123Ubuntu2204SimpleFlow(t *testing.T) {
```
g-gaston, line 1519: Is the idea to convert these into upgrade tests at some point? I feel like that probably adds better coverage with no a lot of overhead.

test/framework/vsphere.go
```diff
@@ -128,230 +128,95 @@ func NewVSphere(t *testing.T, opts ...VSphereOpt) *VSphere {
 	return v
 }
 
-// WithRedHat123VSphere vsphere test with redhat 1.23.
-func WithRedHat123VSphere() VSphereOpt {
+// withVSphereKubeVersionAndOS returns a VSphereOpt that adds API fillers to use a vSphere template for
+// the specified OS family and version (default if not provided), corresponding to a particular
+// Kubernetes version, in addition to configuring all machine configs to use this OS family.
+func withVSphereKubeVersionAndOS(osFamily anywherev1.OSFamily, kubeVersion anywherev1.KubernetesVersion, osVersion ...string) VSphereOpt {
```
g-gaston, line 134: IMO we should move away from these. It has proved to be difficult to work with
Can't we use `WithKubeVersionAndOS` instead?

test/framework/template.go
```diff
@@ -38,12 +38,20 @@ type templateRegistry struct {
...
+func (tc *templateRegistry) templateForRelease(t *testing.T, osFamily anywherev1.OSFamily, release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion, osVersion ...string) string {
 	t.Helper()
 	osFamilyStr := string(osFamily)
 	versionsBundle := readVersionsBundles(t, release, kubeVersion)
 	eksDName := versionsBundle.EksD.Name
-	templateEnvVarName := tc.generator.envVarForTemplate(osFamilyStr, eksDName)
+
+	// If an OS version is not passed in, then we use the default version for this OS family.
+	// Otherwise we use the OS version passed in
+	osVersionStr := defaultOSVersions[osFamilyStr]
```
g-gaston, line 49: Is it possible to be explicit and pass the OS version always? Instead of having to use the variadic arguments and having this defaulting here. It might seems simpler to use, but it's far from obvious to a new reader and this logic is quite burried. Plus it's not documented anywhere.

Either that or just have one argument, like `OS`. And then be able to derive the OSFamily from the OS (which each provider can easily do).

test/framework/nutanix.go
```diff
@@ -119,233 +125,273 @@ func RequiredNutanixEnvVars() []string {
...
+		templateEnvVar = fmt.Sprintf("T_NUTANIX_TEMPLATE_NAME_%s_%s_%s", strings.ToUpper(osFamily), osVersion, strings.ToUpper(strings.ReplaceAll(eksDName, "-", "_")))
+	}
+	return templateEnvVar
+}
+
+// defaultNameForTemplate looks for a template: "{eks-d version}-{osFamily}"
+// eg: kubernetes-1-23-eks-22-ubuntu.
+func (n *Nutanix) defaultNameForTemplate(osFamily, osVersion, eksDName string) string {
+	defaultTemplateName := fmt.Sprintf("%s-%s", strings.ToLower(eksDName), strings.ToLower(osFamily))
+	if osVersion != defaultOSVersions[osFamily] {
```
g-gaston, line 377: could we add a comment to all these methods explaining why we do this? (I'm assuming it's for backwards compat?)
Although we could just change all the current env vars to follow this new pattern and the code will be simpler.

test/framework/template.go
abhay-krishna: Parametrizing the OS version in tests is a new system so I felt it would be better to also support the existing way of writing tests that is less strict and can handle defaults. So the main reason is backwards compatibility. I can document it in the E2E readme if that's okay.

test/framework/vsphere.go
abhay-krishna: The provider.KubeVersionAndOS returns a `ClusterConfigFiller` object whereas this field returns a `<provider>Opt` object. Also the methods used within are specific to each provider. I do agree that in the future, it'd definitely be better to abstract it to a common method and just inject the provider for this method to do the right thing. But I think in the interest of time, we should tackle this as a follow-up. I did refactor each provider framework by adding common methods to reduce duplication, but I can take this as a follow-up.

test/e2e/tinkerbell_test.go
abhay-krishna: Yeah I plan to add upgrade tests of the following format:
* `1.26 Ubuntu 2004` to `1.26 Ubuntu 2204`
* `1.26 Ubuntu 2204` to `1.27 Ubuntu 2204`

cmd/integration_test/build/buildspecs/tinkerbell-test-eks-a-cli.yml
```diff
@@ -31,13 +31,16 @@ env:
...
     T_TINKERBELL_IMAGE_UBUNTU_1_24: "tinkerbell_ci:image_ubuntu_1_24"
     T_TINKERBELL_IMAGE_UBUNTU_1_25: "tinkerbell_ci:image_ubuntu_1_25"
     T_TINKERBELL_IMAGE_UBUNTU_1_26: "tinkerbell_ci:image_ubuntu_1_26"
     T_TINKERBELL_IMAGE_UBUNTU_1_27: "tinkerbell_ci:image_ubuntu_1_27"
-    T_TINKERBELL_IMAGE_REDHAT_1_22: "tinkerbell_ci:image_redhat_1_22"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_23: "tinkerbell_ci:image_ubuntu_2204_1_23"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_24: "tinkerbell_ci:image_ubuntu_2204_1_24"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_25: "tinkerbell_ci:image_ubuntu_2204_1_25"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_26: "tinkerbell_ci:image_ubuntu_2204_1_26"
+    T_TINKERBELL_IMAGE_UBUNTU_2204_1_27: "tinkerbell_ci:image_ubuntu_1_27"
```
abhinavmpandey08, line 43:
```suggestion
    T_TINKERBELL_IMAGE_UBUNTU_2204_1_27: "tinkerbell_ci:image_ubuntu_2204_1_27"
```

test/framework/nutanix.go
abhay-krishna: Will do

test/framework/vsphere.go
g-gaston: maybe I missed something
Is this method covering existing code or did you add it new?
if I'm not making sense, let's move this offline

test/framework/template.go
g-gaston: ah yeah I get that, this is new. What I'm trying to say is I think it's better to make it required. I propose simplicity + explicit vs "easy" to use.

test/e2e/tinkerbell_test.go
g-gaston: ok cool
When you do that, can we remove these simple flow ones? They will probably be redundant

test/framework/nutanix.go
abhay-krishna: Done

test/e2e/tinkerbell_test.go
abhay-krishna: Cool, will do.

test/framework/vsphere.go
abhay-krishna: It's existing code, just changed the function signature and unexported it.

test/framework/template.go
abhay-krishna: I don't believe that making it variadic makes it more complex, it's just allowing us to do the defaulting and for the most part is abstracted from the user. I can change it though, but can I take it as a followup? Otherwise it would require touching every method out there.

abhay-krishna: I also added documentation in the e2e readme for configuring the OS version, so I feel that should help a new reader until we make it explicit.

g-gaston: OK, what about something like this.

Instead of considering osFamily and osVersion two different things, we just make them one. They would only make sense as two arguments if the versions could be mix and matched in between os families, but they can't, they are exclusive for each type of OS.

So I believe we can start thinking about an OS as the combination of its family and its version.

```go
type OS string

const (
	Ubuntu2204    OS = "ubuntu-22-04"
	Ubuntu2004    OS = "ubuntu" // We don't add the version number in the string to facilitate backwards compatibility
	Bottlerocket1 OS = "br"     // We don't add the version number in the string to facilitate backwards compatibility
	RedHat8       OS = "redhat" // We don't add the version number in the string to facilitate backwards compatibility
)
```

Then, the signature of the methods is simplified, and the actual implementation too.

```go
// providerTemplateNameGenerator is an interface for getting template name for each provider.
type providerTemplateNameGenerator interface {
	envVarForTemplate(os OS, eksDName string) string
	defaultNameForTemplate(os OS, eksDName string) string
	searchTemplate(ctx context.Context, template string) (string, error)
	defaultEnvVarForTemplate(os OS, kubeVersion anywherev1.KubernetesVersion) string
}

func (tc *templateRegistry) templateForRelease(t *testing.T, operatingSystem OS, release *releasev1.EksARelease, kubeVersion anywherev1.KubernetesVersion) string {
```

This also simplifies the methods to update cluster specs:
```go
func (v *VSphere) WithKubeVersionAndOS(kubeVersion anywherev1.KubernetesVersion, os OS) api.ClusterConfigFiller {
	return api.JoinClusterConfigFillers(
		api.ClusterToConfigFiller(api.WithKubernetesVersion(kubeVersion)),
		api.VSphereToConfigFiller(
			v.templateForKubeVersionAndOS(os, kubeVersion),
			api.WithOsFamilyForAllMachines(osFamiliesForOS[os]),
		),
	)
}
```

Btw, in order to extract the osFamily from the OS:
```go
var osFamiliesForOS = map[OS]anywherev1.OSFamily{
	Ubuntu2204:    anywherev1.Ubuntu,
	Ubuntu2004:    anywherev1.Ubuntu,
	Bottlerocket1: anywherev1.Bottlerocket,
	RedHat8:       anywherev1.RedHat,
}
```
And this can be done transparently to the test user, so it's one concern less.

And finally, this is how the tests read:

```go
func TestVSphereKubernetes127Ubuntu2204SimpleFlow(t *testing.T) {
	vsphere := framework.NewVSphere(t)
	test := framework.NewClusterE2ETest(t, vsphere)
	test.WithClusterConfig(
		vsphere.WithKubeVersionAndOS(v1alpha1.Kube127, framework.Ubuntu2204),
	)

	runSimpleFlow(test)
}
```

Which allows to parametrize tests even more (not asking to make this change, but just an example of how clean it can read):
```go
func TestVSphereKubernetes127Ubuntu2204SimpleFlow(t *testing.T) {
	runTestWithSimpleFlow(
		t,
		framework.NewVSphere(t),
		v1alpha1.Kube127,
		framework.Ubuntu2204,
	)
}

func runTestWithSimpleFlow(t *testing.T, provider framework.Provider, kubeVersion v1alpha1.KubernetesVersion, os framework.OS) {
	test := framework.NewClusterE2ETest(t, provider)
	test.WithClusterConfig(
		provider.WithKubeVersionAndOS(kubeVersion, os),
	)

	runSimpleFlow(test)
}
```

g-gaston: partial implementation, in case it's easier to read: https://github.com/g-gaston/eks-anywhere/pull/new/ubuntu-22.04

test/e2e/README.md
```diff
@@ -16,10 +16,28 @@ or
...
+
+In each provider-specific file in the [E2E framework folder](../framework), there are helper methods of the form `With<OS family><KubeVersion>()`, for example, `WithUbuntu127()`, which instructs the tests to use an Ubuntu template/image corresponding to Kubernetes 1.27, depending on the provider. These methods take an OS version as an optional function parameter to construct the template/image name or environment variable. If the OS family you are writing tests for has multiple versions, for example, Ubuntu 20.04 and Ubuntu 22.04, and you do not provide an OS version is provided when writing the test, the [default version](../framework/os_versions.go) defined for the OS will be used to configure the template/image. If an OS version is provided, that will be used instead.
+
+For example, here is how you would write vSphere Kubernetes 1.27 simple flow test that uses an Ubuntu 22.04 template.
+```Go
+func TestVSphereKubernetes127Ubuntu2204SimpleFlow(t *testing.T) {
+	test := framework.NewClusterE2ETest(
+		t,
+		framework.NewVSphere(t, framework.WithUbuntu127(framework.Ubuntu2204Version)),
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube127)),
```
g-gaston, line 33: I love we documenting this. However, could we favor using `provider.WithKubeVersionAndOS` instead of provider options?

I would prefer all new tests to start using this as we slowly move towards refactoring the old ones. This pattern is more flexible and it favors parametrized tests.

abhay-krishna: Do you mean changing the function signature of `New<Provider>` methods to not have provider options in it?

test/framework/template.go
abhay-krishna: Thanks for the guidance with examples, super clean! I see that more recent tests are using the above patter for provider instantiation without the provider options, while the older ones are using the options.
```
vsphere := framework.NewVSphere(t)
```
Did you want me to remove the provider options for kubeVersionAndOS in the scope of this PR?

test/e2e/README.md
g-gaston: Ah nono
More like this
```go
func TestVSphereKubernetes127Ubuntu2204SimpleFlow(t *testing.T) {
	vsphere := framework.NewVSphere(t)
	test := framework.NewClusterE2ETest(t, vsphere)
	test.WithClusterConfig(
		vsphere.WithKubeVersionAndOS(v1alpha1.Kube127, framework.Ubuntu2204),
	)

	runSimpleFlow(test)
}
```

test/framework/template.go
g-gaston: > Did you want me to remove the provider options for kubeVersionAndOS

Ah nono, that's a lot. I'm only suggesting writing the new tests this way.
All the other provider options can default to the original "version" for that OS, since that's what all the existing tests expect any way.

That way we only scope this changes to your new tests.

abhay-krishna: Done!

test/e2e/README.md
abhay-krishna: changed! Thank you for the example @g-gaston!

test/e2e/docker_test.go
```diff
@@ -851,10 +851,9 @@ func TestDockerKubernetes126to127GithubFluxEnabledUpgradeFromLatestMinorRelease(
 
 func TestDockerKubernetes126WithOIDCManagementClusterUpgradeFromLatestSideEffects(t *testing.T) {
 	provider := framework.NewDocker(t)
-	runTestManagementClusterUpgradeSideEffects(t, provider, "", v1alpha1.Kube126, api.ClusterToConfigFiller(
+	runTestManagementClusterUpgradeSideEffects(t, provider, framework.OS(""), v1alpha1.Kube126, api.ClusterToConfigFiller(
```
g-gaston, line 854: nit: we could have a `Docker` OS
Which would be ignored when creating the config fillers, same way as it is now, but at least it represents the OS better than just mpety string

test/e2e/tinkerbell_test.go
```diff
@@ -1516,6 +1596,71 @@ func TestTinkerbellKubernetes127UbuntuSimpleFlow(t *testing.T) {
...
+	provider := framework.NewTinkerbell(t)
+	test := framework.NewClusterE2ETest(
+		t,
+		provider,
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube123)),
+		framework.WithControlPlaneHardware(1),
+		framework.WithWorkerHardware(1),
```
g-gaston, line 1606: nit: the first one is not needed

test/e2e/tinkerbell_test.go
```diff
@@ -1516,6 +1596,71 @@ func TestTinkerbellKubernetes127UbuntuSimpleFlow(t *testing.T) {
...
+	)
+	test.WithClusterConfig(provider.WithKubeVersionAndOS(v1alpha1.Kube123, framework.Ubuntu2204, nil))
```
g-gaston, line 1608: Ah I see how passing the release is annoying for tests that just want to use main.
Could we create a ticket to address this? Probably creating some kind of helper that only takes kube version and OS for this kind of tests.

test/framework/cluster.go
```diff
@@ -230,6 +231,16 @@ func PersistentCluster() ClusterE2ETestOpt {
 	}
 }
 
+// WithSkipRegenerateClusterConfig skips generating the cluster config if it finds
+// a cluster config already in the ClusterConfigLocation. Prevents overriding an
+// existing updated cluster config with the default one generated by the CLI
+// `generate clusterconfig` command
+func WithSkipRegenerateClusterConfig() ClusterE2ETestOpt {
+	return func(e *ClusterE2ETest) {
+		e.SkipRegenerateClusterConfig = true
```
g-gaston, line 240: We can have this as a follow up since it's a small thing, but should we just make this the default behavior and avoid having to pass a new option?

g-gaston: Wait I see we already had this as default. So now we are changing the default to be the opposite, it will overwrite by default?

test/framework/nutanix.go
```diff
@@ -120,233 +125,267 @@ func RequiredNutanixEnvVars() []string {
...
 
-// UpdateNutanixUbuntuTemplate123Var returns NutanixFiller by reading the env var and setting machine config's
-// image name parameter in the spec.
-func UpdateNutanixUbuntuTemplate123Var() api.NutanixFiller {
-	return api.WithNutanixStringFromEnvVar(nutanixTemplateNameUbuntu123Var, api.WithNutanixMachineTemplateImageName)
+// templateForKubeVersionAndOS returns a Nutanix filler for the given OS and Kubernetes version.
+func (n *Nutanix) templateForKubeVersionAndOS(kubeVersion anywherev1.KubernetesVersion, os OS, release *releasev1.EksARelease) api.NutanixFiller {
+	var template string
+	if release == nil {
+		template = n.templateForDevRelease(kubeVersion, os)
```
g-gaston, line 310: yeah this ties to my previous comment about passing nil
Let's follow up on this to avoid having to do this default in all providers.

test/framework/template.go
g-gaston: Thanks a lot! I know this was a painful change but love the result!

test/framework/nutanix.go
abhay-krishna: Would you opposed to using variadics for this or perhaps a different method for tests where release is non-nil?

test/framework/cluster.go
abhay-krishna: The reason I had to do this is, for tests calling the `WithClusterConfig` method, a cluster config gets generated by that method with the proper fillers and options. Then when we go to the simple flow method, it again calls on the GenerateClusterConfig method that directly runs the CLI command without the fillers and overrides the previously created correct cluster config.

In case of the Ubuntu 2204 tests, this was causing the correct Ubuntu-based cluster config to be replaced by a `bottlerocket` os family in the machine configs since that's the default generated by the CLI command that runs as part of the simple flow method.

abhay-krishna: Removed this new option and instead created separate functions for the new tests that don't generate the cluster config

test/framework/nutanix.go
abhay-krishna: will add this as a suggestion to the issue

g-gaston (approved):

abhinavmpandey08 (approved):

PR #6227, opened 2023-07-20
Add control plane components healthy check when computing ControlPlaneReady condition

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*

While testing upgrades, we ran into an edge case where all the old control plane machines were removed and all the new control plane machines and nodes were all rolled out. However, an old node still remained in the cluster, which resulted in the   `ControlPlaneReady` condition being marked as ready when it was actually not. We use the status on the `KubeadmControlPlane` to compute this and determine the control plane is ready when there are no outdated CAPI machines, number of replicas  match the expected and all are ready. 

Example KCP status in this case

```
status:
  conditions:
  - lastTransitionTime: "2023-07-11T22:24:33Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2023-07-11T22:01:55Z"
    status: "True"
    type: Available
  - lastTransitionTime: "2023-07-11T22:01:54Z"
    status: "True"
    type: CertificatesAvailable
  - lastTransitionTime: "2023-07-11T22:24:01Z"
    message: Control plane node eksa-test-405f3e3-tzcfc does not have a corresponding
      machine
    reason: ControlPlaneComponentsUnhealthy
    severity: Error
    status: "False"
    type: ControlPlaneComponentsHealthy
  - lastTransitionTime: "2023-07-11T22:01:55Z"
    status: "True"
    type: EtcdClusterHealthy
  - lastTransitionTime: "2023-07-11T22:01:55Z"
    status: "True"
    type: MachinesCreated
  - lastTransitionTime: "2023-07-11T22:24:33Z"
    status: "True"
    type: MachinesReady
  - lastTransitionTime: "2023-07-11T22:24:33Z"
    status: "True"
    type: MachinesSpecUpToDate
  - lastTransitionTime: "2023-07-11T22:24:33Z"
    status: "True"
```

In this edge case, an old node still existed in the cluster, but all the machines had already been deleted. This PR adds a control plane components healthy check when determining the status of the `ControlPlaneReady` condition.

*Testing (if applicable):*
- Unit testing
- Manual testing

### DIFF
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -117,6 +117,13 @@ func updateControlPlaneReadyCondition(cluster *anywherev1.Cluster, kcp *controlp
 		return
 	}
 
+	// We check the condition signifying the overall health of the control plane components. Usually, the control plane should be healthy
+	// at this point but if that is not the case, we report it as an error.
+	if conditions.IsFalse(kcp, controlplanev1.ControlPlaneComponentsHealthyCondition) {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, anywherev1.ControlPlaneComponentsUnhealthyReason, clusterv1.ConditionSeverityError, conditions.GetMessage(kcp, controlplanev1.ControlPlaneComponentsHealthyCondition))
+		return
+	}
+
 	conditions.MarkTrue(cluster, anywherev1.ControlPlaneReadyCondition)
 }
 
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/conditions"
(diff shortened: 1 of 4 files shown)


pkg/controller/clusters/status.go
```diff
@@ -117,6 +117,13 @@ func updateControlPlaneReadyCondition(cluster *anywherev1.Cluster, kcp *controlp
 		return
 	}
 
+	// We check the condition signifying the overall health of the control plane components. Usually, the control plane should be healthy
+	// at this point but if that is not the case, we report it as an error.
+	if conditions.IsFalse(kcp, controlplanev1.ControlPlaneComponentsHealthyCondition) {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, anywherev1.ControlPlaneComponentsUnhealthyReason, clusterv1.ConditionSeverityError, conditions.GetMessage(kcp, controlplanev1.ControlPlaneComponentsHealthyCondition))
```
g-gaston, line 123: nit: can we optimize this by getting the condition once?

pkg/controller/clusters/status.go
```diff
@@ -117,6 +118,19 @@ func updateControlPlaneReadyCondition(cluster *anywherev1.Cluster, kcp *controlp
 		return
 	}
 
+	// We check the condition signifying the overall health of the control plane components. Usually, the control plane should be healthy
+	// at this point but if that is not the case, we report it as an error.
+	kcpControlPlaneHealthyCondition := conditions.Get(kcp, controlplanev1.ControlPlaneComponentsHealthyCondition)
+	if kcpControlPlaneHealthyCondition == nil {
```
g-gaston, line 124: I don't think this is necessary correct
Given we don't know what the logic around this is, if it's not present, I would continue as if it was true. It just seems safer.

g-gaston (commented):
/lgtm

Issue #4255, opened 2022-12-01
Allow users to have specify different templates in `VSphereMachineConfig`


abhinavmpandey08:
**What would you like to be added**:
Currently, for vSphere provider, all the machine configs are required to have the same template config. If you try to provide a different template, the CLI fails with the preflight `all VSphereMachineConfigs must have the same template specified`

**Why is this needed**:
Users may want to configure templates with specific settings for specific node groups which can be propagated down to the cloned VMs. The same template shouldn't be required as long as the K8S version and osFamily of the template are the same.

chrisdoherty4:
Are we able to gather some insights into what kind of settings are being tweaked?

abhinavmpandey08:
> Are we able to gather some insights into what kind of settings are being tweaked?

This specific ask was for `sched.swap.vmxSwapEnabled=FALSE` but there are some other advanced VM level settings that other users might have configure.

nickperry:
This restriction also precludes use of EKS Anywhere within my organization.

abhinavmpandey08:
Hi @nickperry, can you explain the use-case for using different templates in your org if you don't mind?
Just trying to collect all the data points here to determine the priority of this issue. Thanks!

nickperry:
We would like to be able to deliver OVAs with additional pre-pulled container images.

We also just fundamentally don't like having unnecessary restrictions imposed in vendor product abstractions over an underlying more flexible technology (CAPV in this case).

jiayiwang7:
Feature will be included in the upcoming v0.17.0 release. Closing issue.

PR #6230, opened 2023-07-20
vsphere cp/worker/etcd can use diff machine template

ddjjia:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/4255

*Description of changes:*
- remove validation of unified template
- remove cpTemplate`vsphereTemplate` and add `controlPlaneTemplate` and `etcdTemplate` in template value map
- rename workerTemplate `vsphereTemplate` to `workerTemplate`

*Testing (if applicable):*
unit test and manual testing, will add e2e in a separate PR


pkg/providers/vsphere/vsphere_test.go
```diff
@@ -3151,6 +3139,37 @@ func TestProviderGenerateCAPISpecForCreateWithPodIAMConfig(t *testing.T) {
 	test.AssertContentToFile(t, string(cp), "testdata/expected_results_pod_iam_config.yaml")
 }
 
+func TestProviderGenerateCAPISpecForCreateWithMultipleMachineTemplate(t *testing.T) {
```
jiayiwang7, line 3142: instead of creating a new test and adding 2 long test files, can we change an existing test and make them use different template and test this change?

pkg/providers/vsphere/template.go
```diff
@@ -172,7 +172,8 @@ func buildTemplateMapCP(
 		"controlPlaneVsphereResourcePool":      controlPlaneMachineSpec.ResourcePool,
 		"vsphereServer":                        datacenterSpec.Server,
 		"controlPlaneVsphereStoragePolicyName": controlPlaneMachineSpec.StoragePolicyName,
-		"vsphereTemplate":                      controlPlaneMachineSpec.Template,
+		"controlPlaneTemplate":                 controlPlaneMachineSpec.Template,
```
vivek-koppuru, line 175: If we are making these changes, can we change the reference of the worker node vsphereTemplate to `workerTemplate`?

pkg/providers/vsphere/vsphere_test.go
ddjjia: I basically use `cluster_main.yaml`but it's also referenced at other places so think better to have a separate test file for this case

cxbrowne1207 (approved):

PR #6231, opened 2023-07-20
Update Cilium doc references to use shortcode

chrisdoherty4:


chrisnegus (approved):
/lgtm
I tried this and it works for me. I raised a concern about keeping the Cilium version number up to date. But I'm okay merging this while we figure that part out.

PR #6215, opened 2023-07-20
[Docs]Add CNI chaining support with Cilium in EKS-A

amitmavgupta:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

*Documentation added/planned (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #6233, opened 2023-07-20
Revert "[Docs]Add CNI chaining support with Cilium in EKS-A (#6215)"

chrisdoherty4:
This reverts commit 05193320d345543e7e11494a78dd394fb72bca77.

The reference is unnecessary for the alert and is included in the ingress documentation.


csplinter:
lgtm. We can add a CNI Chaining section once we have instructions on how to set it up with EKS-A. Right now this content doesn't make sense to have in an alert on a page for replacing the default CNI, hence the revert.

PR #6234, opened 2023-07-20
Apply DS to disable udp offloading before Cilium upgrades in VSphere

taneyland:
*Issue #, if available:*

*Description of changes:*
This is only needed for EKS-A v0.17 because UDP offloading needs to be disabled for the new Cilium version to work with VSphere clusters. Since our templates that were shipped in v0.16 releases still have UDP offloading enabled in the nodes, we must apply this daemon set to disable UDP offloading in all the nodes before upgrading Cilium in that cluster. We will remove this in EKS-A release v0.18.0.

The new DS is using our kind node image because ethtool is installed there.

*Testing (if applicable):*
Manual testing
LatestMinorReleaseE2E passed
```
    cluster.go:896: Skipping VM cleanup
--- PASS: TestVSphereKubernetes125To126UbuntuUpgradeFromLatestMinorRelease (2114.73s)
PASS
```

### DIFF
--- a/pkg/providers/vsphere/config/ethtool_daemonset.yaml
+++ b/pkg/providers/vsphere/config/ethtool_daemonset.yaml
@@ -1,0 +1,33 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: vsphere-disable-udp-offload
+  namespace: {{.eksaSystemNamespace}}
+spec:
+  selector:
+    matchLabels:
+      name: vsphere-disable-udp-offload
+  template:
+    metadata:
+      labels:
+        name: vsphere-disable-udp-offload
+    spec:
+      containers:
+        - name: vsphere-disable-udp-offload-complete
+          image: {{.kindNodeImage}}
+          imagePullPolicy: IfNotPresent
+          command: [ "/bin/sh" ]
+          args: [ "-c", "sleep infinity" ]
+      initContainers:
+        - name: vsphere-disable-udp-offload
+          image: {{.kindNodeImage}}
+          command:
+            - "/bin/sh"
+            - "-c"
+            - "ethtool -K eth0 tx-udp_tnl-segmentation off && ethtool -K eth0 tx-udp_tnl-csum-segmentation off && echo 'done'"
+          securityContext:
+            privileged: true
+      hostNetwork: true
+      restartPolicy: Always
+      tolerations:
+        - operator: "Exists"
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -12,12 +12,15 @@ import (
 
 	"github.com/Masterminds/sprig"
 	etcdv1 "github.com/aws/etcdadm-controller/api/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/bootstrapper"
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/config"
 	"github.com/aws/eks-anywhere/pkg/constants"
@@ -28,6 +31,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/providers"
 	"github.com/aws/eks-anywhere/pkg/providers/common"
 	"github.com/aws/eks-anywhere/pkg/retrier"
+	"github.com/aws/eks-anywhere/pkg/templater"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
 	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
@@ -48,6 +52,7 @@ const (
 	backOffPeriod            = 5 * time.Second
 	disk1                    = "Hard disk 1"
 	disk2                    = "Hard disk 2"
+	ethtoolDaemonSetName     = "vsphere-disable-udp-offload"
 )
 
 //go:embed config/template-cp.yaml
@@ -58,6 +63,9 @@ var defaultClusterConfigMD string
 
 //go:embed config/secret.yaml
 var defaultSecretObject string
+
+//go:embed config/ethtool_daemonset.yaml
+var ethtoolDaemonSetObject string
 
 var (
 	eksaVSphereDatacenterResourceType = fmt.Sprintf("vspheredatacenterconfigs.%s", v1alpha1.GroupVersion.Group)
@@ -135,6 +143,8 @@ type ProviderKubectlClient interface {
 	DeleteEksaDatacenterConfig(ctx context.Context, vsphereDatacenterResourceType string, vsphereDatacenterConfigName string, kubeconfigFile string, namespace string) error
 	DeleteEksaMachineConfig(ctx context.Context, vsphereMachineResourceType string, vsphereMachineConfigName string, kubeconfigFile string, namespace string) error
 	ApplyTolerationsFromTaintsToDaemonSet(ctx context.Context, oldTaints []corev1.Taint, newTaints []corev1.Taint, dsName string, kubeconfigFile string) error
+	GetDaemonSet(ctx context.Context, name, namespace, kubeconfig string) (*appsv1.DaemonSet, error)
+	Delete(ctx context.Context, resourceType, kubeconfig string, opts ...kubernetes.KubectlDeleteOption) error
 }
 
 // IPValidator is an interface that defines methods to validate the control plane IP.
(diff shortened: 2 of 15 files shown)


vivek-koppuru (approved):
This looks fine but we need to make sure that with @tatlat's version bundle changes that we change the reference here to point to the control plane bundle when he gets his PR merged

https://github.com/aws/eks-anywhere/pull/6180

pkg/providers/vsphere/vsphere.go
```diff
@@ -1152,11 +1156,38 @@ func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context) err
...
+	// This is only needed for EKS-A v0.17 because UDP offloading needs to be disabled for
+	// the new Cilium version to work with VSphere clusters. Since our templates that were shipped
+	// in v0.16 releases still have UDP offloading enabled in the nodes, we must apply the daemon set
+	// to disable UDP offloading in all the nodes before upgrading Cilium.
+	// We will remove this in EKS-A release v0.18.0.
+	logger.V(4).Info("Applying ethtool daemonset to disable udp offloading")
+	if err := p.applyEthtoolDaemonSet(ctx, cluster, clusterSpec); err != nil {
+		return err
+	}
+	return nil
```
chrisdoherty4, line 1174:
```suggestion
	return p.applyEthtoolDaemonSet(ctx, cluster, clusterSpec); err 
```

chrisdoherty4 (approved):
Looks great.

pkg/providers/vsphere/config/ethtool_daemonset.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - name: ethtool
+          image: {{.kindNodeImage}}
+          command: ["/bin/sh"]
+          args: ["-c", "ethtool -K eth0 tx-udp_tnl-segmentation off && ethtool -K eth0 tx-udp_tnl-csum-segmentation off && echo 'done' && sleep infinity"]
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            privileged: true
+      hostNetwork: true
+      restartPolicy: Always
+      tolerations:
```
g-gaston, line 25: do we need tolerations for CP?

pkg/providers/vsphere/config/ethtool_daemonset.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: ethtool
```
g-gaston, line 4: nit: `vsphere-ethtool-segmentation-path` or whatever else, but something a bit more explicit

pkg/providers/vsphere/vsphere.go
```diff
@@ -1152,11 +1156,35 @@ func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context) err
...
+	values := map[string]interface{}{
+		"eksaSystemNamespace": constants.EksaSystemNamespace,
+		"kindNodeImage":       bundle.EksD.KindNode.VersionedImage(),
+	}
+	b, err := templater.Execute(ethtoolDaemonSetObject, values)
+	if err != nil {
+		return err
+	}
+
+	if err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, b); err != nil {
```
g-gaston, line 1185: Shouldn't we wait until all pods are done doing their thing?

pkg/providers/vsphere/vsphere.go
```diff
@@ -1152,11 +1156,35 @@ func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context) err
 	return nil
 }
 
-// PreCoreComponentsUpgrade staisfies the Provider interface.
+// PreCoreComponentsUpgrade satisfies the Provider interface.
 func (p *vsphereProvider) PreCoreComponentsUpgrade(
```
g-gaston, line 1160: Is there a method we can run after new nodes have been rolled out to delete this DS?

pkg/providers/vsphere/vsphere.go
```diff
@@ -1152,11 +1156,35 @@ func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context) err
...
 	cluster *types.Cluster,
 	clusterSpec *cluster.Spec,
 ) error {
+	// This is only needed for EKS-A v0.17 because UDP offloading needs to be disabled for
+	// the new Cilium version to work with VSphere clusters. Since our templates that were shipped
+	// in v0.16 releases still have UDP offloading enabled in the nodes, we must apply the daemon set
+	// to disable UDP offloading in all the nodes before upgrading Cilium.
+	// We will remove this in EKS-A release v0.18.0.
+	logger.V(4).Info("Applying ethtool daemonset to disable udp offloading")
+	return p.applyEthtoolDaemonSet(ctx, cluster, clusterSpec)
```
g-gaston, line 1171: can we add a guard clause and only run it for the osFamily that need it? (I think ubuntu and redhat?)

pkg/providers/vsphere/config/ethtool_daemonset.yaml
```diff
@@ -0,0 +1,26 @@
...
+  template:
+    metadata:
+      labels:
+        name: ethtool
+    spec:
+      containers:
+        - name: ethtool
+          image: {{.kindNodeImage}}
+          command: ["/bin/sh"]
+          args: ["-c", "ethtool -K eth0 tx-udp_tnl-segmentation off && ethtool -K eth0 tx-udp_tnl-csum-segmentation off && echo 'done' && sleep infinity"]
```
g-gaston, line 19: Do we need to a `readinessProbe` to be able to determine when the container has finished? or maybe we can remove the `sleep infinity` and just rely on the pods being on state `Succeeded`?

g-gaston (commented):
Amazing turnaround speed!

pkg/providers/vsphere/vsphere.go
taneyland: Good idea. I can create a method to do this

pkg/providers/vsphere/config/ethtool_daemonset.yaml
taneyland: No i think we would need this. We want to run this on all nodes that exist, so we shouldn't need anything specific for CP

pkg/executables/kubectl.go
```diff
@@ -2345,6 +2345,16 @@ func (k *Kubectl) GetDaemonSet(ctx context.Context, name, namespace, kubeconfig
 	return obj, nil
 }
 
+// DeleteDaemonSet deletes the given daemonset object.
+func (k *Kubectl) DeleteDaemonSet(ctx context.Context, name, namespace string, cluster *types.Cluster) error {
```
g-gaston, line 2349: can't we use use the `Delete` method?

pkg/providers/vsphere/config/ethtool_daemonset.yaml
g-gaston: Don't the CP nodes have taints? Let's sync offline, I might be missing something

pkg/providers/vsphere/vsphere.go
```diff
@@ -1148,15 +1156,68 @@ func (p *vsphereProvider) InstallCustomProviderComponents(ctx context.Context, k
...
+	)
+}
+
+func (p *vsphereProvider) checkEthtoolDaemonSetReady(ctx context.Context, cluster *types.Cluster) error {
+	ethtoolDaemonSet, err := p.providerKubectlClient.GetDaemonSet(ctx, ethtoolDaemonSetName, constants.EksaSystemNamespace, cluster.KubeconfigFile)
+	if err != nil {
+		return err
+	}
+
+	if ethtoolDaemonSet.Status.DesiredNumberScheduled != ethtoolDaemonSet.Status.NumberReady {
```
g-gaston, line 1219: won't the pods be ready before the command finishes?

taneyland: Talked offline. Added a readiness probe to the container

pkg/providers/vsphere/vsphere.go
```diff
@@ -1148,15 +1158,72 @@ func (p *vsphereProvider) InstallCustomProviderComponents(ctx context.Context, k
 	return nil
 }
 
-func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context) error {
+// PostBootstrapDeleteForUpgrade runs any provider-specific operations after bootstrap cluster has been deleted.
+func (p *vsphereProvider) PostBootstrapDeleteForUpgrade(ctx context.Context, cluster *types.Cluster) error {
+	// Delete the daemonset that was used to disable udp offloading in ubuntu/redhat nodes.
+	logger.V(3).Info("Deleting vsphere-disable-udp-offload daemonset")
```
vivek-koppuru, line 1164: We want to delete after the upgrade finishes? Should this be once we know that it was installed after we install Cilium?

taneyland: We need to wait until after the new nodes generated with the updated template roll out since those will definitely have udp offloading disabled.

So we could technically delete the DS a little earlier in the flow (immediately after the new nodes are out), but that would have required a new method in the Provider interface so I wanted to leverage this existing provider-specific method for cleanup.

taneyland:
Passed with the latest changes:
```
    cluster.go:896: Skipping VM cleanup
--- PASS: TestVSphereKubernetes125To126UbuntuUpgradeFromLatestMinorRelease (2117.42s)
PASS
```

pkg/providers/vsphere/vsphere.go
vivek-koppuru: I see got it, I thought it was just once and then be good to go for the rollout.

PR #6236, opened 2023-07-20
Modular k8s e2e

tatlat:
*Issue #, if available:*

*Description of changes:*
Not even finished

*Testing (if applicable):*


pkg/executables/kubectl.go
```diff
@@ -1869,6 +1869,46 @@ func (k *Kubectl) ValidateNodesVersion(ctx context.Context, kubeconfig string, k
 	return nil
 }
 
+// ValidateClusterNodeVersion validates the Control Plane node's kubernetes version in E2E tests.
+func (k *Kubectl) ValidateClusterNodeVersion(ctx context.Context, kubeconfig string, kubeVersion v1alpha1.KubernetesVersion) error {
```
g-gaston, line 1873: I don't think this belongs here. This kubectl should be just a client that reads and writes data to the api server. The validation IMO should be somewhere else.

g-gaston: I would suggest just using `List` for example (which I think it's already implemented). And do the validation in your test code.

It might even be easier if you just use a client.Client instead.

test/framework/cluster.go
```diff
@@ -741,6 +741,43 @@ func (e *ClusterE2ETest) ValidateCluster(kubeVersion v1alpha1.KubernetesVersion)
 	}
 }
 
+// ValidateClusterWorkerNodeVersion validates the nodes on a cluster match kubeVersion if they use the top level kubernetes version or workerKubeVersion for a worker node that has its own version.
+func (e *ClusterE2ETest) ValidateClusterWorkerNodeVersion(kubeVersion v1alpha1.KubernetesVersion, workerKubeVersion *v1alpha1.KubernetesVersion) {
```
g-gaston, line 745: can we use the cluster validator that cavaughn wrote for this?

PR #6256, opened 2023-07-24
Change compareEKSAClusterSpec

tatlat:
*Issue #, if available:*

*Description of changes:*
Compare all versions bundle in cluster spec

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -719,7 +719,9 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, clus *types
 }
 
 func compareEKSAClusterSpec(ctx context.Context, currentClusterSpec, newClusterSpec *cluster.Spec) (bool, error) {
-	for version, newVersionsBundle := range newClusterSpec.VersionsBundles {
+	versions := newClusterSpec.Cluster.KubernetesVersions()
+	for _, version := range versions {
+		newVersionsBundle := newClusterSpec.VersionsBundles[version]
 		currentVersionsBundle, ok := currentClusterSpec.VersionsBundles[version]
 		if !ok {
 			logger.V(3).Info("New versions bundle detected")
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -720,6 +720,12 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, clus *types
 
 func compareEKSAClusterSpec(ctx context.Context, currentClusterSpec, newClusterSpec *cluster.Spec) (bool, error) {
 	versions := newClusterSpec.Cluster.KubernetesVersions()
+	oldVersions := currentClusterSpec.Cluster.KubernetesVersions()
+
+	if len(versions) != len(oldVersions) {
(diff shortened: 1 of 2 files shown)


pkg/clustermanager/cluster_manager.go
```diff
@@ -719,12 +719,17 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, clus *types
 }
 
 func compareEKSAClusterSpec(ctx context.Context, currentClusterSpec, newClusterSpec *cluster.Spec) (bool, error) {
-	currentVersionsBundle := currentClusterSpec.ControlPlaneVersionsBundle()
-	newVersionsBundle := newClusterSpec.ControlPlaneVersionsBundle()
+	for version, newVersionsBundle := range newClusterSpec.VersionsBundles {
```
g-gaston, line 722: Can we use here the `cluster.KubernetesVersions`?
At some point I would like for VersionsBundles to contain all versions bundles and not just the ones used by the cluster

pkg/clustermanager/cluster_manager.go
```diff
@@ -719,12 +719,17 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, clus *types
...
+		currentVersionsBundle, ok := currentClusterSpec.VersionsBundles[version]
+		if !ok {
+			logger.V(3).Info("New versions bundle detected")
+			return true, nil
+		}
 
-	if currentVersionsBundle.EksD.Name != newVersionsBundle.EksD.Name {
-		logger.V(3).Info("New eks-d release detected")
-		return true, nil
+		if currentVersionsBundle.EksD.Name != newVersionsBundle.EksD.Name {
```
g-gaston, line 729: this algorithm won't always work by itself. What if you had CP = 1.22 and worker1=1.21 and then you upgrade to CP=1.22 and worker1=1.22. Your algorithm will return a "no change", just because the old cluster already contained the 1.22 bundle.

I think we need to compare cp version and worker node by worker node versions.

g-gaston (commented):
I would add some unit tests. If you think this is difficult to test with the current cluster manager setup (it might be), let me know and we can figure something out.

pkg/clustermanager/cluster_manager.go
```diff
@@ -720,6 +720,12 @@ func (c *ClusterManager) EKSAClusterSpecChanged(ctx context.Context, clus *types
 
 func compareEKSAClusterSpec(ctx context.Context, currentClusterSpec, newClusterSpec *cluster.Spec) (bool, error) {
 	versions := newClusterSpec.Cluster.KubernetesVersions()
+	oldVersions := currentClusterSpec.Cluster.KubernetesVersions()
+
+	if len(versions) != len(oldVersions) {
```
g-gaston, line 725: this makes it a bit more robust but still doesn't catch all cases
What if you had CP = 1.22 and worker1=1.21 and worker2=1.21 and then you upgrade to CP=1.22, worker1=1.22 and worker2=1.21

I advise to first write the tests so you can see where the wholes are

tatlat: The thing is that that EKSAClusterSpecChanged catches that case by comparing the old and new clusters. 
compareEKSAClusterSpec is only called by that function.

g-gaston (approved):

PR #6278, opened 2023-07-25
Fix test framework for eksa version

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -22,6 +22,11 @@ func UpdateClusterStatusForControlPlane(ctx context.Context, client client.Clien
 		return errors.Wrapf(err, "getting kubeadmcontrolplane")
 	}
 
+	if kcp == nil {
+		conditions.Set(cluster, controlPlaneInitializationInProgressCondition())
+		return nil
+	}
+
 	updateControlPlaneInitializedCondition(cluster, kcp)
 	updateControlPlaneReadyCondition(cluster, kcp)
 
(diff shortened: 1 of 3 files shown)


pkg/controller/clusters/status.go
```diff
@@ -22,6 +22,11 @@ func UpdateClusterStatusForControlPlane(ctx context.Context, client client.Clien
 		return errors.Wrapf(err, "getting kubeadmcontrolplane")
 	}
 
+	if kcp == nil {
```
g-gaston, line 25: why this change? it's difficult to understand the changes without any context

tatlat: it caused a nil pointer dereference when deleting a workload cluster in TestVSphereUpgradeKubernetes123to124UbuntuWorkloadClusterAPI in updateControlPlaneReadyCondition(cluster, kcp)

cxbrowne1207: I'd vote to just return nil here instead of also setting the condition then. The ControlPlaneInitialized condition should not be updated again after being "True". This will update the value back to "False"

g-gaston: is it easier to just add this check to `updateControlPlaneReadyCondition`?

tatlat: yeah we can do that instead

cxbrowne1207: yeah, i think that solution is better

g-gaston:
/lgtm
Next time please add some description to the PR to help the reviewers build context :)

PR #6280, opened 2023-07-25
Only apply DS to vsphere redhat nodes

taneyland:
*Issue #, if available:*

*Description of changes:*
We do not need to apply the daemonset to disable udp offloading to ubuntu nodes, so updating that logic to only run for Redhat OS family on VSphere

*Testing (if applicable):*

We will add a note about this in the v0.17 Changelog


g-gaston (commented):
\lgtm

PR #6296, opened 2023-07-26
Update Docker upgrade from latest minor release tests to use kindnetd

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR updates the upgrade from latest minor release docker tests to use kindnet. It also fixes an a nil pointer error that causes a `CrashBackOff` loop in the controller when updating the Cluster status when not using cilium as the default cni.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/side_effects.go
+++ b/test/e2e/side_effects.go
@@ -13,6 +13,7 @@ import (
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/clusterapi"
 	"github.com/aws/eks-anywhere/pkg/types"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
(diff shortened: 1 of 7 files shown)


test/e2e/side_effects.go
```diff
@@ -205,9 +206,7 @@ func newEKSAPackagedBinaryForLocalBinary(tb testing.TB) eksaPackagedBinary {
 	}
 }
 
-func runTestManagementClusterUpgradeSideEffects(t *testing.T, provider framework.Provider, osFamily anywherev1.OSFamily, kubeVersion anywherev1.KubernetesVersion) {
-	latestRelease := latestMinorRelease(t)
-
+func runTestManagementClusterUpgradeSideEffects(t *testing.T, provider framework.Provider, latestRelease *releasev1.EksARelease, configFillers ...api.ClusterConfigFiller) {
```
g-gaston, line 209: I would prefer if we didn't change the signature this way
the idea is that tests will only depend on provider, kube version and OS
So combinations can be easily parametrized

I'm ok if you need to add the extra fillers to make some stuff custom, but could we keep the rest as it was? that way we don't need to change the other provider tests either

Btw, there is an alternative. You could add the kindnetd filler to the default fillers added by the docker provider. That way it will be added to all docker tests by default

cxbrowne1207: i’d prefer the approach to add the extra fillers on then if that’s acceptable, and keep the rest of the signature as it was.

what we wanted to do was use kindnet for only the upgrade from latest minor release docker tests, instead of them all using kindnet by default

PR #6299, opened 2023-07-27
Fix Airgap EksaVersion

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/builder.go
+++ b/pkg/cluster/builder.go
@@ -164,8 +164,10 @@ func (b FileSpecBuilder) getEksaRelease(mReader *manifests.Reader) (*releasev1.E
 		return release, nil
 	}
 
+	// When using bundles-override or a custom bundle, a fake EksaRelease can be used since using a custom bundle
+	// is like creating a new EKS-A version.
 	return &releasev1.EksARelease{
-		Date:              "2023-07-26 20:02:17.931339039 +0000 UTC",
+		Date:              "",
 		Version:           b.cliVersion.GitVersion,
 		Number:            1,
 		GitCommit:         "",
--- a/pkg/cluster/builder.go
+++ b/pkg/cluster/builder.go
@@ -89,7 +89,7 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 		return nil, err
 	}
 
-	release, err := mReader.ReadReleaseForVersion(b.cliVersion.GitVersion)
+	release, err := b.getEksaRelease(mReader)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/cluster/builder.go
```diff
@@ -154,6 +154,28 @@ func (b FileSpecBuilder) getBundles(manifestReader *manifests.Reader) (*releasev
...
+		// this shouldn't return an error at this point due to getBundles performing similar operations prior to this call
+		release, err := mReader.ReadReleaseForVersion(b.cliVersion.GitVersion)
+		if err != nil {
+			return nil, err
+		}
+		return release, nil
+	}
+
+	return &releasev1.EksARelease{
+		Date:              "2023-07-26 20:02:17.931339039 +0000 UTC",
```
vivek-koppuru, line 168: Why is this date hardcoded?

pkg/cluster/builder.go
```diff
@@ -154,6 +154,28 @@ func (b FileSpecBuilder) getBundles(manifestReader *manifests.Reader) (*releasev
 	return bundles.Read(b.reader, bundlesURL)
 }
 
+func (b FileSpecBuilder) getEksaRelease(mReader *manifests.Reader) (*releasev1.EksARelease, error) {
+	if b.bundlesManifestURL == "" {
```
abhinavmpandey08, line 158: Will this be `""` in airgap environments?

pkg/cluster/builder.go
```diff
@@ -154,6 +154,28 @@ func (b FileSpecBuilder) getBundles(manifestReader *manifests.Reader) (*releasev
...
+
+	return &releasev1.EksARelease{
+		Date:              "2023-07-26 20:02:17.931339039 +0000 UTC",
+		Version:           b.cliVersion.GitVersion,
+		Number:            1,
+		GitCommit:         "",
+		GitTag:            "",
+		BundleManifestUrl: "",
+		EksABinary:        releasev1.BinaryBundle{},
+		EksACLI:           releasev1.PlatformBundle{},
```
abhinavmpandey08, line 175: Also, why we these values just empty? Do they not matter? If so, shouldn't the function just return the version?

abhinavmpandey08 (commented):
I might have

pkg/cluster/builder.go
tatlat: "" is there if bundles-override is NOT set. You can see a similar check happening [here](https://github.com/aws/eks-anywhere/pull/6299/files/7aa3d4094a048ce0eeaf5b7c03c0d9a5bc002f9e#diff-5a573820e4737e0df734c87f7ff19388a4717953c43f7406572b012a52451338R150)

vivek-koppuru (approved):

g-gaston (commented):
/lgtm

g-gaston (approved):

g-gaston:
Forcing the merge here bc it's quite difficult to test just one branch of the new code with the current setup. In practice, the program will mostly never walk that path.

It's not worth exporting private methods to testing or complicate the code to tinker with the failure paths.

PR #6300, opened 2023-07-27
Set default templates for all vsphere machine configs

ahreehong:
*Issue #, if available:*

*Description of changes:*
Currently, default templates are only being set in the worker nodes' `vsphereMachineConfig`. This changes it back to set the default templates for all `vsphereMachineConfig`, including the cp and etcd machine configs. 

*Testing (if applicable):*
Confirmed that  `TestVSphereKubernetes124BottlerocketAutoimport` passed locally. 
Manually ran an upgrade with modular Kubernetes Version upgrades with no templates specified to check that the proper templates get set for different node groups.

### DIFF
--- a/pkg/providers/vsphere/defaults.go
+++ b/pkg/providers/vsphere/defaults.go
@@ -29,7 +29,7 @@ func (d *Defaulter) setDefaultsForMachineConfig(ctx context.Context, spec *Spec)
 	setDefaultsForEtcdMachineConfig(spec.etcdMachineConfig())
 
 	for _, w := range spec.Cluster.Spec.WorkerNodeGroupConfigurations {
-		if err := d.setDefaultTemplateIfMissing(ctx, spec, w); err != nil {
+		if err := d.setWorkerDefaultTemplateIfMissing(ctx, spec, w); err != nil {
 			return err
 		}
 	}
@@ -37,6 +37,10 @@ func (d *Defaulter) setDefaultsForMachineConfig(ctx context.Context, spec *Spec)
 	for _, m := range spec.machineConfigs() {
 		m.SetDefaults()
 		m.SetUserDefaults()
+
+		if err := d.setDefaultTemplateIfMissing(ctx, spec, m); err != nil {
+			return err
+		}
 
 		if err := d.setTemplateFullPath(ctx, spec.VSphereDatacenter, m); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/providers/vsphere/defaults.go
```diff
@@ -69,7 +73,7 @@ func setDefaultsForEtcdMachineConfig(machineConfig *anywherev1.VSphereMachineCon
 	}
 }
 
-func (d *Defaulter) setDefaultTemplateIfMissing(ctx context.Context, spec *Spec, workerNodeGroup anywherev1.WorkerNodeGroupConfiguration) error {
+func (d *Defaulter) setWorkerDefaultTemplateIfMissing(ctx context.Context, spec *Spec, workerNodeGroup anywherev1.WorkerNodeGroupConfiguration) error {
```
abhinavmpandey08, line 76: can you also update the log on line 83 to say worker node instead of control plane?

pkg/providers/vsphere/vsphere_test.go
```diff
@@ -2833,6 +2833,30 @@ func TestSetupAndValidateCreateClusterDefaultTemplate(t *testing.T) {
 	}
 }
 
+func TestSetupAndValidateCreateClusterSetDefaultTemplate(t *testing.T) {
```
vivek-koppuru, line 2836: Don't you need to change the name of this function? We can say CP and the above Worker or something

vivek-koppuru (approved):

pkg/providers/vsphere/vsphere_test.go
```diff
@@ -2833,6 +2833,30 @@ func TestSetupAndValidateCreateClusterDefaultTemplate(t *testing.T) {
...
+	clusterSpec.VersionsBundles["1.19"].EksD.Name = eksd119Release
+	clusterSpec.VersionsBundles["1.19"].EksD.KubeVersion = "v1.19.8"
+	clusterSpec.VersionsBundles["1.19"].KubeVersion = "1.19"
+	clusterSpec.Cluster.Namespace = "test-namespace"
+	provider := givenProvider(t)
+	controlPlaneMachineConfigName := clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name
+	clusterSpec.VSphereMachineConfigs[controlPlaneMachineConfigName].Spec.Template = ""
+	etcdMachineConfigName := clusterSpec.Cluster.Spec.ExternalEtcdConfiguration.MachineGroupRef.Name
+	clusterSpec.VSphereMachineConfigs[etcdMachineConfigName].Spec.Template = ""
+	wantError := fmt.Errorf("failed setting default values for vsphere machine configs: can not import ova for osFamily: ubuntu, please use bottlerocket as osFamily for auto-importing or provide a valid template")
```
g-gaston, line 2851: what are we asserting here?
Shouldn't our test verify that all machine configs have a template set at the end of the test?
Btw, you can test this through the validator directly if that's easier

pkg/providers/vsphere/defaults.go
```diff
@@ -87,6 +91,18 @@ func (d *Defaulter) setDefaultTemplateIfMissing(ctx context.Context, spec *Spec,
 	return nil
 }
 
+func (d *Defaulter) setDefaultTemplateIfMissing(ctx context.Context, spec *Spec, m *anywherev1.VSphereMachineConfig) error {
+	if m.Spec.Template == "" {
+		logger.V(1).Info("VSphereMachineConfig template is not set. Using default template.")
+		versionsBundle := spec.ControlPlaneVersionsBundle()
```
g-gaston, line 97: could we add a comment saying here that we assume we need to use the control plane versions bundle bc at this point all worker node machine configs should already have a template?

Or you can pass the versions bundle as an argument and move the comment up to `setDefaultsForMachineConfig` which is probably cleaner.

abhinavmpandey08 (approved):

pkg/providers/vsphere/vsphere_test.go
g-gaston: We are going to address this separately

pkg/providers/vsphere/defaults.go
g-gaston: We are going to address this separately

g-gaston:
/lgtm
We are going to address the comments separately and merge this now to get CI going

PR #6313, opened 2023-07-27
Fix CAPI Backup and add filter for cluster name

vignesh-goutham:
*Description of changes:*
Set the context to right management cluster when upgrading both mgmt and workload cluster for taking backup. 
Set cluster name filter for backing up the cluster that is getting upgraded.

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -106,7 +106,7 @@ type ClientFactory interface {
 
 type ClusterClient interface {
 	KubernetesClient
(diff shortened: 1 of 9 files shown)


vivek-koppuru:
Spoke with @raymond-zhang00 and will be storing two backups with the first one including all the resources and not just the cluster name so that we can cover the scenario for restoration where mgmt cluster fails to upgrade and customer loses access to both kind cluster and mgmt cluster. This helps make it easier to recover the workload cluster resources into a new cluster if the original mgmt cluster is not recoverable.

PR #6323, opened 2023-07-28
Vsphere multi template e2e

ddjjia:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/1603

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -820,6 +820,19 @@ func (k *Kubectl) GetControlPlaneNodes(ctx context.Context, kubeconfig string) (
 	return response.Items, err
 }
 
+// GetVsphereMachine will return list of vSphere machines.
+func (k *Kubectl) GetVsphereMachine(ctx context.Context, kubeconfig string, selector string) ([]vspherev1.VSphereMachine, error) {
+	params := []string{"get", "vspheremachines", "-o", "json", "--namespace", constants.EksaSystemNamespace, "--kubeconfig", kubeconfig, "--selector=" + selector}
+	stdOut, err := k.Execute(ctx, params...)
+	if err != nil {
+		return nil, fmt.Errorf("getting VSphere machine: %v", err)
+	}
+	response := &vspherev1.VSphereMachineList{}
+	err = json.Unmarshal(stdOut.Bytes(), response)
+
+	return response.Items, err
+}
+
 func (k *Kubectl) ValidateNodes(ctx context.Context, kubeconfig string) error {
 	template := "{{range .items}}{{.metadata.name}}\n{{end}}"
 	params := []string{"get", "nodes", "-o", "go-template", "--template", template, "--kubeconfig", kubeconfig}
--- a/test/framework/multi_template.go
+++ b/test/framework/multi_template.go
@@ -1,0 +1,143 @@
+package framework
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	vspherev1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1beta1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/logger"
+)
+
+const (
+	// VSphereMultiTemplateUbuntu127 is to test multiple vSphere templates.
+	VSphereMultiTemplateUbuntu127 = "T_VSPHERE_TEMPLATE_UBUNTU_2204_1_27"
+	// ControlPlaneMachineLabel is to get control plane vSphere machines from machine list.
+	ControlPlaneMachineLabel = "cluster.x-k8s.io/control-plane"
+	// EtcdMachineLabel is to get etcd vSphere machines from machine list.
+	EtcdMachineLabel = "cluster.x-k8s.io/etcd-cluster"
+	// workerMachineLabel is to get worker vSphere machines from machine list.
+	workerMachineLabel = "cluster.x-k8s.io/deployment-name"
+)
+
+// VSphereMultiTemplateUbuntu127RequiredEnvVars is required for TestVSphereMultipleTemplatesUbuntu127.
+var VSphereMultiTemplateUbuntu127RequiredEnvVars = []string{
+	VSphereMultiTemplateUbuntu127,
+}
+
+// RequiredVsphereMultiTemplateUbuntu127EnvVars return required env vars for TestVSphereMultipleTemplatesUbuntu127.
+func RequiredVsphereMultiTemplateUbuntu127EnvVars() []string {
+	return VSphereMultiTemplateUbuntu127RequiredEnvVars
+}
+
+// CheckVsphereMultiTemplateUbuntu127EnvVars checks is required env vars are present.
+func CheckVsphereMultiTemplateUbuntu127EnvVars(t *testing.T) {
+	checkRequiredEnvVars(t, VSphereMultiTemplateUbuntu127RequiredEnvVars)
+}
+
+// VsphereMachineValidation should return an error if either an error is encountered during execution or the validation logically fails.
+// This validation function will be executed by ValidateVsphereMachine and ValidateWorkerNodeVsphereMachine
+// with a vSphere machine config and a corresponding vSphere machine.
+type VsphereMachineValidation func(machineConfig *v1alpha1.VSphereMachineConfig, machine vspherev1.VSphereMachine) (err error)
+
+// ValidateVsphereMachine deduces the control plane or etcd configuration to machine mapping
+// and for each configuration/machine pair executes the provided validation functions.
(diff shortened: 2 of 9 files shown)


cmd/integration_test/build/buildspecs/vsphere-test-eks-a-cli.yml
```diff
@@ -13,6 +13,7 @@ env:
     T_VSPHERE_TEMPLATE_UBUNTU_1_25: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-25"
     T_VSPHERE_TEMPLATE_UBUNTU_1_26: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-26"
     T_VSPHERE_TEMPLATE_UBUNTU_1_27: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-27"
+    T_VSPHERE_MULTI_TEMPLATE_UBUNTU_1_27: "/SDDC-Datacenter/vm/Templates/ubuntu-kube-v1-27-no-snapshots"
```
vivek-koppuru, line 16: Why do we need to use this specific one?

internal/test/e2e/setup.go
```diff
@@ -168,6 +168,11 @@ func (e *E2ESession) setup(regex string) error {
 		return err
 	}
 
+	err = e.setupVsphereMultiTemplateUbuntu127(regex)
```
vivek-koppuru, line 171: It might be best to avoid having to do this logic based on test name rather than just being able to use a 127 template. We have to use a no snapshots one?

pkg/executables/kubectl.go
```diff
@@ -820,6 +820,19 @@ func (k *Kubectl) GetControlPlaneNodes(ctx context.Context, kubeconfig string) (
 	return response.Items, err
 }
 
+// GetVsphereMachine will return list of vSphere machines.
+func (k *Kubectl) GetVsphereMachine(ctx context.Context, kubeconfig string, selector string) ([]vspherev1.VSphereMachine, error) {
```
vivek-koppuru, line 824: We should unit test this

test/framework/multi_template.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// RequiredVsphereMultiTemplateUbuntu127EnvVars return required env vars for TestVSphereMultipleTemplatesUbuntu127.
+func RequiredVsphereMultiTemplateUbuntu127EnvVars() []string {
+	return VSphereMultiTemplateUbuntu127RequiredEnvVars
+}
+
+// CheckVsphereMultiTemplateUbuntu127EnvVars checks is required env vars are present.
+func CheckVsphereMultiTemplateUbuntu127EnvVars(t *testing.T) {
+	checkRequiredEnvVars(t, VSphereMultiTemplateUbuntu127RequiredEnvVars)
```
vivek-koppuru, line 39: If we are anyways checking it here, we don't need the setup logic right? If a user doesn't specify, it should only fail for this test and not for other tests I think?

cmd/integration_test/build/buildspecs/vsphere-test-eks-a-cli.yml
ddjjia: not specific, just need to use a different one other than `T_VSPHERE_TEMPLATE_UBUNTU_1_27`

internal/test/e2e/setup.go
ddjjia: It doesn't have to be a no snapshot one, but need to be a different one then default 1.27 ubuntu

test/framework/multi_template.go
ddjjia: yes, so it's only called in the test `TestVSphereMultipleTemplatesUbuntu127`

internal/test/e2e/setup.go
vivek-koppuru: So why do we need to call the setup here if we only need to env var for the test?

pkg/executables/kubectl.go
```diff
@@ -820,6 +820,19 @@ func (k *Kubectl) GetControlPlaneNodes(ctx context.Context, kubeconfig string) (
...
+	params := []string{"get", "vspheremachines", "-o", "json", "--namespace", constants.EksaSystemNamespace, "--kubeconfig", kubeconfig, "--selector=" + selector}
+	stdOut, err := k.Execute(ctx, params...)
+	if err != nil {
+		return nil, fmt.Errorf("getting VSphere machine: %v", err)
```
vivek-koppuru, line 828: Can we do an error case here to cover this?

test/framework/multi_template.go
```diff
@@ -0,0 +1,108 @@
...
+		for _, machine := range machines {
+			if machineReg.MatchString(machine.Name) {
+				machineFound = true
+				for _, validation := range validations {
+					err := validation(e.ClusterConfig.VSphereMachineConfigs[wng.MachineGroupRef.Name], machine)
+					if err != nil {
+						e.T.Errorf("VSphere machine %v is not valid: %v", machine.Name, err)
+					}
+				}
+				break
```
tatlat, line 90: nit: Could you reduce code nesting by checking if the regex does not match instead and then `continue`? That way you could move the code in the current "if block" to below the continue statement.

test/framework/multi_template.go
```diff
@@ -0,0 +1,109 @@
...
+	machines, err := e.KubectlClient.GetVsphereMachine(ctx, e.Cluster().KubeconfigFile, workerMachineLabel)
+	if err != nil {
+		e.T.Fatal(err)
+	}
+
+	for _, wng := range e.ClusterConfig.Cluster.Spec.WorkerNodeGroupConfigurations {
+		machineFound := false
+		machineReg := regexp.MustCompile(common.WorkerMachineTemplateBase(e.ClusterName, wng.Name) + `-[0-9]+-[a-z0-9]+`)
+		for _, machine := range machines {
+			if !machineReg.MatchString(machine.Name) {
```
vivek-koppuru, line 82: Are we only doing a set of validations for one machine of the machine template? And not all the machines associated with that machine template?

ddjjia: yeah it only validates one machine from each group, cuz each group use the same template

test/framework/multi_template.go
```diff
@@ -0,0 +1,138 @@
...
+
+	machineConfigToMachine := make(map[*v1alpha1.VSphereMachineConfig]vspherev1.VSphereMachine)
+	for _, machine := range machines {
+		if strings.Contains(machine.GetName(), "control-plane") || strings.Contains(machine.Name, "etcd") {
+			continue
+		}
+		if !strings.HasPrefix(machine.GetName(), e.ClusterName) {
+			continue
+		}
+		wng, err := getWngNameFromMachine(machine.GetName(), e.ClusterName)
```
vivek-koppuru, line 100: To be consistent in naming, we should call it `wngName` instead of `wng` then

test/framework/multi_template.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			e.T.Fatal(err)
+		}
+		machineConfigName := wngNameToWng[wng].MachineGroupRef.Name
+		machineConfig := e.ClusterConfig.VSphereMachineConfigs[machineConfigName]
+		if _, exists := machineConfigToMachine[machineConfig]; !exists {
+			machineConfigToMachine[machineConfig] = machine
```
vivek-koppuru, line 107: Why don't we just have a list of machines per machine config and then just validate all the machines, instead of only validating one machine? Isn't it better to be certain that all the machines have the validations run?

test/framework/multi_template.go
```diff
@@ -0,0 +1,138 @@
...
+	for _, wng := range wngConfigs {
+		wngNameToWng[wng.Name] = wng
+	}
+	return wngNameToWng
+}
+
+func getWngNameFromMachine(machineName string, clusterName string) (string, error) {
+	trimmedMachineName := strings.TrimPrefix(machineName, clusterName+"-")
+	wngBaseParts := strings.Split(trimmedMachineName, "-")
+	if len(wngBaseParts) < 3 {
```
vivek-koppuru, line 124: Add a comment for this function on how we are stripping the string to get the name

test/framework/multi_template.go
```diff
@@ -0,0 +1,138 @@
...
+		return "", fmt.Errorf("invalid machine name %v", machineName)
+	}
+	return strings.Join(wngBaseParts[:len(wngBaseParts)-2], "-"), nil
+}
+
+// ValidateMachineTemplate validates if template configured in machine config matches the vSphere machine.
+func ValidateMachineTemplate(machineConfig *v1alpha1.VSphereMachineConfig, machine vspherev1.VSphereMachine) (err error) {
```
vivek-koppuru, line 131: Is this the only validation run today, or are there other?

ddjjia: That's the only one validation on machine config and machine, but build the framework so we can add others later

vivek-koppuru (approved):

PR #6327, opened 2023-07-28
remove validation block bundlesRef and eksaVersion

tatlat:
*Issue #, if available:*

*Description of changes:*
The validation that blocked users from having both bundlesRef and eksaVersion set was causing issues when upgrading a cluster with gitops enabled from latest minor release. 

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -57,6 +57,10 @@ func BuildSpec(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*
 
 // BuildSpecFromConfig constructs a cluster.Spec for an eks-a cluster config by retrieving all dependencies objects from the cluster using a kubernetes client.
 func BuildSpecFromConfig(ctx context.Context, client Client, config *Config) (*Spec, error) {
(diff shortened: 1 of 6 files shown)


pkg/cluster/fetch.go
```diff
@@ -117,14 +117,14 @@ func BundlesForCluster(ctx context.Context, client Client, cluster *v1alpha1.Clu
 
 func eksaReleaseForCluster(ctx context.Context, client Client, cluster *v1alpha1.Cluster) (*v1alpha1release.EKSARelease, error) {
 	eksaRelease := &v1alpha1release.EKSARelease{}
-	if cluster.Spec.BundlesRef == nil {
-		if cluster.Spec.EksaVersion == nil {
-			return nil, fmt.Errorf("either cluster's EksaVersion or BundlesRef need to be set")
-		}
+	if cluster.Spec.EksaVersion == nil && cluster.Spec.BundlesRef == nil {
```
vivek-koppuru, line 120: This validation should be done at the caller of this function and not here, because this is purely for grabbing the eksa release. We should have this check in both `BundlesForCluster` and `BuildSpecFromConfig`

PR #6332, opened 2023-07-28
best effort backup and bootstrap backup to different directory

raymond-zhang00:
*Issue #, if available:*

*Description of changes:*
Updating first workload cluster backup to be best effort. Will first attempt a backup with no cluster filter, then backup with filter if that fails.

Added a check to only take bootstrap cluster backup if bootstrap cluster reference is not nil. Also updated bootstrap backup to not overwrite workload backup and instead write to a separate folder.

*Testing (if applicable):*
unit tests added

### DIFF
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -277,8 +277,24 @@ func WithNoTimeouts() ClusterManagerOpt {
 	}
 }
 
+func clusterctlMoveWaitForInfrastructureRetryPolicy(totalRetries int, err error) (retry bool, wait time.Duration) {
+	// Retry both network and cluster move errors.
+	if match := (clusterctlNetworkErrorRegex.MatchString(err.Error()) || clusterctlMoveErrorRegex.MatchString(err.Error())); match {
+		return true, clusterctlMoveRetryWaitTime(totalRetries)
+	}
+	return false, 0
+}
+
 func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait time.Duration) {
-	// Exponential backoff on network and cluster move errors.  Retrier built-in backoff is linear, so implementing here.
+	// Retry only network errors.
+	if match := clusterctlNetworkErrorRegex.MatchString(err.Error()); match {
+		return true, clusterctlMoveRetryWaitTime(totalRetries)
+	}
+	return false, 0
+}
+
+func clusterctlMoveRetryWaitTime(totalRetries int) time.Duration {
+	// Exponential backoff on errors.  Retrier built-in backoff is linear, so implementing here.
 
 	// Retrier first calls the policy before retry #1.  We want it zero-based for exponentiation.
 	if totalRetries < 1 {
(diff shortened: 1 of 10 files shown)


pkg/workflows/upgrade.go
```diff
@@ -457,10 +457,13 @@ func (s *installCAPITask) Restore(ctx context.Context, commandContext *task.Comm
 
 func (s *moveManagementToBootstrapTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
 	logger.Info("Backing up workload cluster's management resources before moving to bootstrap cluster")
-	err := commandContext.ClusterManager.BackupCAPI(ctx, commandContext.WorkloadCluster, commandContext.ManagementClusterStateDir, commandContext.WorkloadCluster.Name)
+	err := commandContext.ClusterManager.BackupCAPI(ctx, commandContext.WorkloadCluster, commandContext.ManagementClusterStateDir, "")
```
vivek-koppuru, line 460: Let's add a comment here describing this.

pkg/clustermanager/cluster_manager.go
```diff
@@ -295,7 +295,25 @@ func clusterctlMoveRetryPolicy(totalRetries int, err error) (retry bool, wait ti
...
+
+	// Retrier first calls the policy before retry #1.  We want it zero-based for exponentiation.
+	if totalRetries < 1 {
+		totalRetries = 1
+	}
+
+	const networkFaultBaseRetryTime = 10 * time.Second
+	const backoffFactor = 1.5
+	waitTime := time.Duration(float64(networkFaultBaseRetryTime) * math.Pow(backoffFactor, float64(totalRetries-1)))
+
```
vivek-koppuru, line 309: All of this can be in a helper function then

pkg/workflows/upgrade.go
```diff
@@ -513,10 +518,13 @@ func (s *upgradeWorkloadClusterTask) Run(ctx context.Context, commandContext *ta
...
 	if err != nil {
 		commandContext.SetError(err)
-		logger.Info("Backing up management components from bootstrap cluster")
-		err := commandContext.ClusterManager.BackupCAPI(ctx, commandContext.ManagementCluster, commandContext.ManagementClusterStateDir, commandContext.WorkloadCluster.Name)
-		if err != nil {
-			logger.Info("Bootstrap management component backup failed, use existing workload cluster backup", "error", err)
+		// Take backup of bootstrap cluster capi components
+		if commandContext.BootstrapCluster != nil {
+			logger.Info("Backing up management components from bootstrap cluster")
+			err := commandContext.ClusterManager.BackupCAPIWaitForInfrastructure(ctx, commandContext.BootstrapCluster, fmt.Sprintf("bootstrap-%s", commandContext.ManagementClusterStateDir), "")
```
vivek-koppuru, line 524: We technically only need the cluster being upgraded here right, which is the cluster being upgraded? Since we are already attempting to back up all before, but if you think it's better to get more here as well that's fine with me.

raymond-zhang00: From my understanding we just want to backup the CAPI components on the bootstrap cluster. Are you saying we should have a `--filter-cluster` parameter here with the workload cluster name?

vivek-koppuru: Yes we could keep it with the parameter, but it doesn't make a huge difference to me.

pkg/workflows/upgrade.go
```diff
@@ -456,11 +456,16 @@ func (s *installCAPITask) Restore(ctx context.Context, commandContext *task.Comm
...
 func (s *moveManagementToBootstrapTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	// Take best effort CAPI backup of workload cluster without filter.
+	// If that errors, then take CAPI backup filtering on only workload cluster.
 	logger.Info("Backing up workload cluster's management resources before moving to bootstrap cluster")
-	err := commandContext.ClusterManager.BackupCAPI(ctx, commandContext.WorkloadCluster, commandContext.ManagementClusterStateDir, commandContext.WorkloadCluster.Name)
+	err := commandContext.ClusterManager.BackupCAPI(ctx, commandContext.WorkloadCluster, commandContext.ManagementClusterStateDir, "")
 	if err != nil {
-		commandContext.SetError(err)
-		return &CollectDiagnosticsTask{}
+		err = commandContext.ClusterManager.BackupCAPIWaitForInfrastructure(ctx, commandContext.WorkloadCluster, commandContext.ManagementClusterStateDir, commandContext.WorkloadCluster.Name)
```
vivek-koppuru, line 464: Instead of having a separate function, and also a separate param for the filter name, we should make use of optional parameters for this function to configure. I am okay with having that as a follow up due to time.

vivek-koppuru (approved):

pkg/workflows/upgrade.go
raymond-zhang00: Gotcha, I'll go ahead and keep the parameter in for this.

vivek-koppuru (approved):

PR #6335, opened 2023-07-29
Fix stalling on upgrade from latest minor release with flux

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR adds a `ClusterSpecGenerate` struct when marshalling the cluster spec. We needed this specifically such that we could generate a yaml with bundlesRef null so that it would
not be omitted from the kubeapi-server.

There was an issue when upgrading from the latest minor release with gitps v0.16.2 to v0.17 during the "Applying eksa yaml resources to cluster" step stalled indefinitely because both the newly added eksaVersion field and the bundlesRef were populated in submission to the kubeapi server. There is a webhook validation that does not allow both to be set at the same time, so that was throwing an error in the kube-api-server pod. This is strange because, in the original submission of the Cluster resource yaml, bundlesRef was nil. As a result of using the force apply for this operation, it was attempting to delete the eksa Cluster resource and stalled on that operation.

This solution is a temporary patch for the issue. After addressing, we should clean this up.

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -72,6 +72,34 @@ type ClusterSpec struct {
 	MachineHealthCheck *MachineHealthCheck `json:"machineHealthCheck,omitempty"`
 }
 
+// ClusterSpecGenerate is the same as ClusterSpec except for removing the omitempty tag from BundlesRef.
+// TODO: We needed this specifically such that we could generate a yaml with bundlesRef null so that it would
+// not be omitted from the kubeapi-server. There was an issue where the apply ekas yaml resources step when
+// upgrading from the latest minor release with gitps v0.16.2 to v0.17, where the newly added eksaVersion field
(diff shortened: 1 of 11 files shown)


cxbrowne1207:
/hold 

Running a few e2e tests before merging

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1417,7 +1445,23 @@ func (c *Cluster) ConvertConfigToConfigGenerateStruct() *ClusterGenerate {
 			Annotations: c.Annotations,
 			Namespace:   namespace,
 		},
-		Spec: c.Spec,
+		Spec: ClusterSpecGenerate{
```
vivek-koppuru, line 1448: Can we add a quick test for this function? It should be simple.

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #6339, opened 2023-07-31
Update image-builder docs to include latest enhancements

abhay-krishna:
Update image-builder docs to include the recent enhancements:
* OS version configuration.
* Mounting additional files.

### DIFF
--- a/docs/content/en/docs/osmgmt/artifacts.md
+++ b/docs/content/en/docs/osmgmt/artifacts.md
@@ -17,9 +17,13 @@ EKS Anywhere supports three different node operating systems:
 Bottlerocket OVAs and images are distributed by the EKS Anywhere project.
 To build your own Ubuntu-based or RHEL-based EKS Anywhere node, see [Building node images]({{< relref "#building-node-images">}}).
 
+## Prerequisites
+
+Several code snippets on this page use `curl` and `yq` commands. Refer to the [Tools section]({{< relref "../getting-started/install/#tools" >}}) to learn how to install them.
+
 ## Bare Metal artifacts
 
-Artifacts for EKS Anyware Bare Metal clusters are listed below.
+Artifacts for EKS Anywhere Bare Metal clusters are listed below.
 If you like, you can download these images and serve them locally to speed up cluster creation.
 See descriptions of the [osImageURL]({{< relref "../getting-started/baremetal/bare-spec/#osimageurl" >}}) and [`hookImagesURLPath`]({{< relref "../getting-started/baremetal/bare-spec#hookimagesurlpath" >}}) fields for details.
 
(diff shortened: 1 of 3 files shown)


docs/content/en/docs/osmgmt/artifacts.md
```diff
@@ -929,6 +939,43 @@ These steps use `image-builder` to create a Ubuntu-based image for Nutanix AHV a
...
+### Configuring OS version
+
+`image-builder` supports an `os-version` flag that allows you to configure which version of the OS you wish to build. If no OS version is supplied, it will build the default for that OS, according to the table below
+
+|            | Supported versions | Default version |
+|:----------:|:-------------------:|:--------------:|
+| **Ubuntu** |     2004, 2204      |      2004      |
+|  **RHEL**  |          8          |       8        |
+
+Currently, Ubuntu is the only operating system that supports multiple versions.
```
csplinter, line 952: what about patch versions?

abhay-krishna: Patch versions are determined by the corresponding Packer config file for the Hypervisor/OS combination we choose. For example, for Ubuntu 22.04 OVAs, the source image is an Ubuntu 22.04.1 ISO. https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/ova/ubuntu-2204.json#L13

If upstream image-builder were to update this version and we pull in a new release of image-builder with this change, then future builds using image-builder will source the updated patch version of Ubuntu.

abhay-krishna:
/override eks-anywhere-e2e-validate-tinkerbell-hardware-presubmit

docs/content/en/docs/osmgmt/artifacts.md
```diff
@@ -922,14 +932,75 @@ These steps use `image-builder` to create a Ubuntu-based image for Nutanix AHV a
...
+
+### Configuring OS version
+
+`image-builder` supports an `os-version` flag that allows you to configure which version of the OS you wish to build. If no OS version is supplied, it will build the default for that OS, according to the table below.
+
+<table style>
+    <thead align="center">
+        <tr>
+            <th><center>Operating system</th>
+            <th><center>Supported versions</th>
```
vignesh-goutham, line 952: The supported versions here is tricky. We do support these versions but we should probably also add a line saying image-builder will pick the latest version on that release train by running upgrade, so that users know if they choose --os-version 2004, they dont always get 20.04.1. This is the same for Rhel too. Maybe we can replicate what we did for RHEL there with a `+` and add a note on what that means.

abhay-krishna: Good call! Will add a note and the `+`.

docs/content/en/docs/osmgmt/artifacts.md
```diff
@@ -99,20 +99,30 @@ curl -s $BUNDLE_MANIFEST_URL | yq ".spec.versionsBundles[].eksD.ova.bottlerocket
...
-EKS Distro Release
+**os:** This category represents the OS corresponding to this template. The value for this tag must be `os:bottlerocket`.
 
-1.27 - `eksdRelease:kubernetes-1-27-eks-4`
+**eksdRelease:** This category represents the EKS Distro release corresponding to this template. The value for this tag can be obtained programmatically as follows.
 
-1.26 - `eksdRelease:kubernetes-1-26-eks-5`
+Using the latest EKS Anywhere version
+```bash
+EKSA_RELEASE_VERSION=$(curl -sL https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml | yq ".spec.latestVersion")
```
jaxesn, line 110: Can we find somewhere, maybe on the admin machine or pre-reqs page, where we show how to install yq? or at least call it out as needed?

docs/content/en/docs/osmgmt/artifacts.md
```diff
@@ -922,14 +932,77 @@ These steps use `image-builder` to create a Ubuntu-based image for Nutanix AHV a
...
+            <th><center>Operating system</th>
+            <th><center>Supported versions</th>
+            <th><center>Corresponding <code>os-version</code> value</th>
+            <th><center>Default <code>os-version</code> value</th>
+        </tr>
+    </thead>
+    <tbody align="center">
+        <tr >
+            <td rowspan=2><b>Ubuntu</b></td>
+            <td>20.04.1+<sup>*</sup></td>
```
jaxesn, line 960: We should just say what this is right now, i like your note but we know what versions they are.

20.04.6?
22.04.1?
8.7?

jaxesn (changes requested):
I like this! couple comments and please have @chrisnegus give it a once over.

docs/content/en/docs/osmgmt/artifacts.md
```diff
@@ -922,14 +932,77 @@ These steps use `image-builder` to create a Ubuntu-based image for Nutanix AHV a
...
+            <td>8.4+</td>
+            <td>8</td>
+            <td rowspan=2>8</td>
+        </tr>
+    </tbody>
+</table>
+
+Currently, Ubuntu is the only operating system that supports multiple `os-version` values.
+
+<sup>*</sup> <b>NOTE:</b> The `+` indicates that Packer sources an image with this version but there is a possibility that it will get upgraded during the `dist-upgrade` phase, so the final image could be a different version from the one listed here.
```
chrisnegus, line 979:
```suggestion
>**_NOTE:_** The `+` indicates that Packer sources an image with this version but there is a possibility that it will get upgraded during the `dist-upgrade` phase, so the final image could be a different version from the one listed here.
```
We're been inconsistent about marking notes, but I don't think we need to do HTML. This is how others in the doc set have done casual sorts of notes like this.

chrisnegus (commented):
Just one suggestion.

docs/content/en/docs/osmgmt/artifacts.md
abhay-krishna: Done

docs/content/en/docs/osmgmt/artifacts.md
abhay-krishna: Removed the note altogether since we don't need it.

docs/content/en/docs/osmgmt/artifacts.md
abhay-krishna: Done

docs/content/en/docs/osmgmt/artifacts.md
abhay-krishna: Included patch versions in table.

PR #6344, opened 2023-07-31
block modular upgrades for tinkerbell

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -472,8 +472,9 @@ func ValidateWorkerKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
 	for _, nodeGroupNewSpec := range new.Spec.WorkerNodeGroupConfigurations {
 		newVersion := nodeGroupNewSpec.KubernetesVersion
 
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -470,6 +471,12 @@ func ValidateWorkerKubernetesVersionSkew(new, old *Cluster) field.ErrorList {
 	}
 	for _, nodeGroupNewSpec := range new.Spec.WorkerNodeGroupConfigurations {
 		newVersion := nodeGroupNewSpec.KubernetesVersion
+
+		if newVersion != nil && nodeGroupNewSpec.MachineGroupRef.Kind == TinkerbellMachineConfigKind {
```
vivek-koppuru, line 475: Do we have this block for CLI already or is it not needed there?

tatlat: although it's located in the webhook, this function is used in the cli as well

vivek-koppuru (approved):

pkg/api/v1alpha1/cluster_webhook.go
vivek-koppuru: Let's create an issue to move this out then as a followup.

PR #6359, opened 2023-08-01
[WIP] Fix copy packages command for airgapped environment

d8660091:
*Issue #, if available:* "copy packages" doesn't work as expected

*Description of changes:*

*Testing (if applicable):*


cmd/eksctl-anywhere/cmd/copypackages.go
```diff
@@ -144,3 +181,56 @@ func (c CopyPackagesCommand) copyImages(ctx context.Context, dstRegistry registr
 	}
 	return nil
 }
+
```
TerryHowe, line 184: I had a PR out to add tags to the bundle, that was the goal to have digest and version in both bundles.

jaxesn: is that something we should resurrect?  Im wondering what the lift is there?

jaxesn: talked on slack, we can take that as a follow up since its a bigger change to packages.

TerryHowe (approved):
/lgtm

PR #6361, opened 2023-08-01
Validate all cloudstack machine configs in webhook

g-gaston:
Until now, any machine config belonging to a workload cluster was
skipping all webhook validations. All existing validations apply to both
management and workload cluster machine configs, so there is no need to
make a distinction here.


g-gaston:
/hold
until cloudstack webhook experts take a look

mitalipaygude (approved):

PR #6370, opened 2023-08-01
Fix Tinkerbell preflights to account for bundles changes

abhinavmpandey08:
*Description of changes:*
The bundle changes trigger a rolling upgrades but Tinkerbell currently doesn't account for them during preflights which causes the upgrade to fail if a bundle change happened alongside a scaling operation.
This PR adds a preflight to detect bundle changes and marks those also as a trigger for rolling upgrades.
So if a scaling operation happens along with bundle changes, the CLI will fail out early in preflights.

### DIFF
--- a/pkg/providers/tinkerbell/upgrade.go
+++ b/pkg/providers/tinkerbell/upgrade.go
@@ -146,7 +146,8 @@ func (p *Provider) validateAvailableHardwareForUpgrade(ctx context.Context, curr
 	)
 
 	rollingUpgrade := false
-	if currentSpec.Cluster.Spec.KubernetesVersion != newClusterSpec.Cluster.Spec.KubernetesVersion {
+	if currentSpec.Cluster.Spec.KubernetesVersion != newClusterSpec.Cluster.Spec.KubernetesVersion ||
+		currentSpec.Bundles.Spec.Number != newClusterSpec.Bundles.Spec.Number {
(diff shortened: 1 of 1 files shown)


pkg/providers/tinkerbell/upgrade.go
```diff
@@ -146,7 +146,8 @@ func (p *Provider) validateAvailableHardwareForUpgrade(ctx context.Context, curr
 	)
 
 	rollingUpgrade := false
-	if currentSpec.Cluster.Spec.KubernetesVersion != newClusterSpec.Cluster.Spec.KubernetesVersion {
+	if currentSpec.Cluster.Spec.KubernetesVersion != newClusterSpec.Cluster.Spec.KubernetesVersion ||
+		currentSpec.Bundles.Spec.Number != newClusterSpec.Bundles.Spec.Number {
```
jaxesn, line 150: will there always be a current bundle and new bundle?

vivek-koppuru (approved):

pkg/providers/tinkerbell/upgrade.go
abhinavmpandey08: yeah, this codepath is called only for upgrade flow where the `currentSpec.Bundles` will be the old spec's bundle and the `newClusterSpec.Bundles` will be the new spec's bundle.

pokearu (approved):
/lgtm
/approve

PR #6384, opened 2023-08-02
Remove force-cleanup flag

### LINKED ISSUE #6387: Deprecate --force-cleanup

g-gaston:
Fixes #6387

*Description of changes:*
* Show an error message if the flag is used
* Don't suggest force-cleanup in when create kind cluster fails
* Update docs to talk about the delete cluster case

We should not suggest using `--force-cleanup`, specially without any context. It can be quite a dangerous operation during upgrade failures.


jiayiwang7 (approved):

vivek-koppuru (approved):

docs/content/en/docs/reference/eksctl/anywhere_create_cluster.md
```diff
@@ -22,15 +22,15 @@ anywhere create cluster -f <cluster-config-file> [flags]
...
       --external-etcd-wait-timeout string   Override the default external etcd wait timeout (default "1h0m0s")
   -f, --filename string                     Filename that contains EKS-A cluster configuration
-      --force-cleanup                       Force deletion of previously created bootstrap cluster
   -z, --hardware-csv string                 Path to a CSV file containing hardware data.
   -h, --help                                help for cluster
       --install-packages string             Location of curated packages configuration files to install to the cluster
       --kubeconfig string                   Management cluster kubeconfig file
       --no-timeouts                         Disable timeout for all wait operations
-      --node-startup-timeout string         Override the default node startup timeout (default "10m0s")
+      --node-startup-timeout string         Override the default node startup timeout (Defaults to 20m for Tinkerbell clusters) (default "10m0s")
```
g-gaston, line 30: This is because we didn't regenerate the docs after changing this flag
Same with the others that are not force-cleanup

docs/content/en/docs/troubleshooting/troubleshooting.md
```diff
@@ -188,19 +188,20 @@ It is also useful to start a shell session on the Docker container running the b
...
 ```
 
-Cluster creation fails because a cluster of the same name already exists. If you are sure the cluster is not being used, try running the `eksctl anywhere create cluster` again, adding the `--force-cleanup` option.
-
-If that doesn't work, you can manually delete the old cluster:
+This indicates that the cluster creation or deletion fails because a bootstrap cluster of the same name already exists. If you are sure the cluster is not being used, you can manually delete the old cluster:
 
 ```bash
-kind delete cluster --name cluster-name
+kind delete cluster --name ${CLUSTER_NAME}-eks-a-cluster
```
jaxesn, line 200: customers may not have kind installed, we should probably show the docker commands here...

g-gaston: can we just reference the kind official docs?

chrisdoherty4 (approved):

docs/content/en/docs/troubleshooting/troubleshooting.md
jaxesn: i mean for now, but we can also just say to run `docker rm -rf <>`

jaxesn (approved):

g-gaston:
Forcing merge since the only line not tested is just the returned error, and the only thing that has changed is the error message string.

PR #6400, opened 2023-08-03
Add apply server side to the k8s client interface

g-gaston:
*Description of changes:*
This allows code shared between the CLI (where the a kubectl wrapper is used) and the controller (where the controller-runtime client.Client is used) to leverage server side apply.

Part of the kindless upgrades work.


g-gaston:
Reading the lgtm, I just added a new test that I forgot to include

PR #6414, opened 2023-08-04
Add ability to list clusters across namespaces for validating management cluster name

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Currently we make an assumption in our management cluster name validation for our workload cluster that both clusters exist in the same namespace. However, it is possible that the management cluster and the workload cluster are created in different namespaces. This is also how we get the cluster object through the CLI: https://github.com/aws/eks-anywhere/blob/2347399db3b997eddc9e2e770af1c1ef19bd6ccb/pkg/executables/kubectl.go#L1500

We are keeping the current check of getting the cluster based on the same namespace, but if we don't find it, we will fall back to looking for the cluster using the list api. If we still don't find the cluster or it errors out, we will tell the user that we were unable to find the cluster as we did before. If we find multiple clusters, which should not be possible as the cluster-api cluster resources can't have the same name in the `eksa-system` namespace, we will throw an error too. Once we get past these validations, we will return the cluster in the different namespace. 

Since the list api for controller-runtime needs an index to look off of, we need to add an indexer based on the cluster name to do the lookup, otherwise it would fail with the following error:
```
List on GroupVersionKind anywhere.eks.amazonaws.com/v1alpha1, Kind=Cluster specifies selector on field metadata.name, but no index with name metadata.name has been registered for GroupVersionKind anywhere.eks.amazonaws.com/v1alpha1, Kind=Cluster
```

I followed the approach listed in https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup to set up the indexer. There is also another approach using the cache to set it up listed here https://stackoverflow.com/questions/57083221/list-custom-resources-from-caching-client-with-custom-fieldselector, but I am unsure what the real difference is between both approaches so I went with the former.

*Testing (if applicable):*
unit testing
functional testing

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -367,6 +367,7 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 	}
 	if cluster.IsManaged() {
 		if err := r.clusterValidator.ValidateManagementClusterName(ctx, log, cluster); err != nil {
+			log.Error(err, "Invalid cluster configuration")
 			cluster.SetFailure(anywherev1.ManagementClusterRefInvalidReason, err.Error())
 			return controller.Result{}, err
 		}
@@ -570,12 +571,14 @@ func aggregatedGeneration(config *c.Config) int64 {
 }
 
 func getManagementCluster(ctx context.Context, clus *anywherev1.Cluster, client client.Client) (*anywherev1.Cluster, error) {
-	mgmtCluster := &anywherev1.Cluster{}
-	if err := client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			failureMessage := fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name)
-			clus.SetFailure(anywherev1.ManagementClusterRefInvalidReason, failureMessage)
-		}
+	mgmtCluster, err := clusters.FetchManagementEksaCluster(ctx, client, clus)
+	if apierrors.IsNotFound(err) {
+		clus.SetFailure(
+			anywherev1.ManagementClusterRefInvalidReason,
+			fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name),
+		)
+	}
+	if err != nil {
 		return nil, err
 	}
 
--- a/manager/main.go
+++ b/manager/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/aws/eks-anywhere/controllers"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/clusterapi"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
 	"github.com/aws/eks-anywhere/pkg/features"
 	snowv1 "github.com/aws/eks-anywhere/pkg/providers/snow/api/v1beta1"
 	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
@@ -142,6 +143,12 @@ func main() {
 	setupChecks(setupLog, mgr)
 	//+kubebuilder:scaffold:builder
 
+	// Adding this indexer to allow for listing Cluster objects by name if they are in different namespaces.
+	if err := mgr.GetFieldIndexer().
+		IndexField(ctx, &anywherev1.Cluster{}, "metadata.name", clientutil.ClusterNameIndexer); err != nil {
+		setupLog.Error(err, "unable to create index for Cluster name")
+		os.Exit(1)
+	}
 	setupLog.Info("Starting manager")
 	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
--- a/pkg/controller/clientutil/indexers_test.go
+++ b/pkg/controller/clientutil/indexers_test.go
@@ -1,0 +1,34 @@
+package clientutil_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
+)
+
+func TestClusterNameIndexerValid(t *testing.T) {
+	g := NewWithT(t)
+	c := &v1alpha1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "my-cluster",
+		},
+	}
+	name := clientutil.ClusterNameIndexer(c)
+	g.Expect(name[0]).To(Equal(c.Name))
+}
+
+func TestClusterNameIndexerFail(t *testing.T) {
+	g := NewWithT(t)
+	c := &v1alpha1.VSphereDatacenterConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "my-cluster",
+		},
+	}
+	g.Expect(func() {
+		clientutil.ClusterNameIndexer(c)
+	}).To(Panic())
+}
--- a/pkg/controller/clusters/management.go
+++ b/pkg/controller/clusters/management.go
@@ -1,0 +1,45 @@
+package clusters
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+// FetchManagementEksaCluster returns the management cluster object for a given workload Cluster.
+// If we are unable to find the management cluster using the same namespace as the current cluster, we will attempt
+// to get a list of the clusters with that name across all the namespaces. If we find multiple, which usually should
+// not happen as these clusters get mapped to a cluster-api cluster object in the eksa-system namespace, then we
+// also error on that because it is not possible to have multiple resources with the same name within a namespace.
+func FetchManagementEksaCluster(ctx context.Context, cli client.Client, cluster *v1alpha1.Cluster) (*v1alpha1.Cluster, error) {
+	mgmtCluster := &v1alpha1.Cluster{}
+	mgmtClusterKey := client.ObjectKey{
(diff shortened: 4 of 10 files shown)


manager/main.go
```diff
@@ -142,6 +142,11 @@ func main() {
 	setupChecks(setupLog, mgr)
 	//+kubebuilder:scaffold:builder
 
+	if err := mgr.GetFieldIndexer().
+		IndexField(ctx, &anywherev1.Cluster{}, "metadata.name", anywherev1.ClusterNameIndexer); err != nil {
+		setupLog.Error(err, "unable to start manager")
```
g-gaston, line 147:
```suggestion
		setupLog.Error(err, "unable to create index for Cluster name")
```

manager/main.go
```diff
@@ -142,6 +142,11 @@ func main() {
 	setupChecks(setupLog, mgr)
 	//+kubebuilder:scaffold:builder
 
+	if err := mgr.GetFieldIndexer().
```
g-gaston, line 145: Can we leave a comment here about why we want this index and what is being used for?
It might not be obvious at a first glance.

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -1485,6 +1486,16 @@ type ClusterList struct {
 	Items           []Cluster `json:"items"`
 }
 
+// ClusterNameIndexer is an indexer for controller to list Cluster objects based on name.
+func ClusterNameIndexer(obj client.Object) []string {
```
g-gaston, line 1490: can we put this somewhere else?
I would rather not have controller-runtime as a dependency in the api package
maybe `controller/clientutil`?

pkg/controller/clusters/validations.go
```diff
@@ -39,11 +39,26 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
 		Namespace: cluster.Namespace,
 		Name:      cluster.Spec.ManagementCluster.Name,
 	}
+	// If we are unable to find the management cluster using the same namespace as the current cluster, we will attempt
+	// to get a list of the clusters with that name across all the namespaces. If we find multiple, which usually should
+	// not happen as these clusters get mapped to a cluster-api cluster object in the eksa-system namespace, then we
+	// also error on that because it is not possible to have multiple resources with the same name within a namespace.
 	if err := v.client.Get(ctx, mgmtClusterKey, mgmtCluster); err != nil {
 		if apierrors.IsNotFound(err) {
```
g-gaston, line 47: nit: to avoid this level of nesting, you can have a guard like `if err != nil && !apierrors.IsNotFound(err) return err`

pkg/controller/clusters/validations.go
```diff
@@ -39,11 +39,26 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
-			err := fmt.Errorf("unable to retrieve management cluster %v: %v", cluster.Spec.ManagementCluster.Name, err)
-			log.Error(err, "Invalid cluster configuration")
-			return err
+			log.Info("Unable to find management cluster in same namespace, looking across namespaces")
+			notFoundErr := fmt.Errorf("unable to retrieve management cluster %v: %v", cluster.Spec.ManagementCluster.Name, err)
+			clusterList := &anywherev1.ClusterList{}
+			if err = v.client.List(ctx, clusterList,
+				client.MatchingFields{"metadata.name": cluster.Spec.ManagementCluster.Name}); err != nil || len(clusterList.Items) == 0 {
+				log.Error(notFoundErr, "Invalid cluster configuration")
```
g-gaston, line 53: nit bc this was already here, but I would leave the error logging up to the caller

pkg/controller/clusters/validations.go
```diff
@@ -39,11 +39,26 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
...
+				return notFoundErr
+			}
+			if len(clusterList.Items) > 1 {
+				multipleFoundErr := fmt.Errorf("found multiple clusters with the name %v", cluster.Spec.ManagementCluster.Name)
```
g-gaston, line 57: nit
```suggestion
				multipleFoundErr := fmt.Errorf("found multiple clusters with the name %s", cluster.Spec.ManagementCluster.Name)
```

pkg/controller/clusters/validations.go
```diff
@@ -39,11 +39,26 @@ func (v *ClusterValidator) ValidateManagementClusterName(ctx context.Context, lo
 		Namespace: cluster.Namespace,
 		Name:      cluster.Spec.ManagementCluster.Name,
 	}
+	// If we are unable to find the management cluster using the same namespace as the current cluster, we will attempt
+	// to get a list of the clusters with that name across all the namespaces. If we find multiple, which usually should
+	// not happen as these clusters get mapped to a cluster-api cluster object in the eksa-system namespace, then we
+	// also error on that because it is not possible to have multiple resources with the same name within a namespace.
 	if err := v.client.Get(ctx, mgmtClusterKey, mgmtCluster); err != nil {
```
g-gaston, line 46: I think there are other places where the management cluster is retrieved, at least in the cluster controller there is one more. Maybe we can extract this reading logic to a separate method that takes (ctx, client and cluster) and reuse it here and in the controller.

pkg/api/v1alpha1/cluster_types.go
vivek-koppuru: Yea good idea, I wasn't happy with this location too

controllers/cluster_controller.go
```diff
@@ -570,15 +571,12 @@ func aggregatedGeneration(config *c.Config) int64 {
...
 func getManagementCluster(ctx context.Context, clus *anywherev1.Cluster, client client.Client) (*anywherev1.Cluster, error) {
-	mgmtCluster := &anywherev1.Cluster{}
-	if err := client.Get(ctx, types.NamespacedName{Name: clus.ManagedBy(), Namespace: clus.Namespace}, mgmtCluster); err != nil {
-		if apierrors.IsNotFound(err) {
-			failureMessage := fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name)
-			clus.SetFailure(anywherev1.ManagementClusterRefInvalidReason, failureMessage)
-		}
+	mgmtCluster, err := clusters.FetchManagementEksaCluster(ctx, client, clus)
+	if err != nil {
+		failureMessage := fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name)
```
g-gaston, line 576: this is not necessarily true, right? it could a transient error, not always a not found

pkg/controller/clientutil/indexers_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
+)
+
+func TestClusterNameIndexerValid(t *testing.T) {
```
g-gaston, line 13: test for the panic? I think gomega has something to make it easier to test

pkg/controller/clusters/management.go
```diff
@@ -0,0 +1,37 @@
...
+	mgmtClusterKey := client.ObjectKey{
+		Namespace: cluster.Namespace,
+		Name:      cluster.ManagedBy(),
+	}
+	err := cli.Get(ctx, mgmtClusterKey, mgmtCluster)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	if apierrors.IsNotFound(err) {
+		notFoundErr := fmt.Errorf("unable to retrieve management cluster %v: %v", cluster.Spec.ManagementCluster.Name, err)
```
g-gaston, line 25:
```suggestion
		notFoundErr := fmt.Errorf("unable to retrieve management cluster %s: %v", cluster.Spec.ManagementCluster.Name, err)
```

pkg/controller/clusters/management.go
```diff
@@ -0,0 +1,37 @@
...
+		clusterList := &v1alpha1.ClusterList{}
+		if err = cli.List(ctx, clusterList,
+			client.MatchingFields{"metadata.name": cluster.Spec.ManagementCluster.Name}); err != nil || len(clusterList.Items) == 0 {
```
g-gaston, line 28: I think if err != nil we should return the error itself, not the precomputed notFound error

vivek-koppuru: Well this is a best effort case, so I decided that we take the error from the Get call instead of the error from the List, which is why I decided the save the error from before. However, we can return the error from list instead if we want to prioritize that error over not finding it from the same namespace.

controllers/cluster_controller.go
```diff
@@ -570,16 +571,15 @@ func aggregatedGeneration(config *c.Config) int64 {
...
-			failureMessage := fmt.Sprintf("Management cluster %s does not exist", clus.Spec.ManagementCluster.Name)
-			clus.SetFailure(anywherev1.ManagementClusterRefInvalidReason, failureMessage)
-		}
+	mgmtCluster, err := clusters.FetchManagementEksaCluster(ctx, client, clus)
+	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
-
-	return mgmtCluster, nil
+	if mgmtCluster == nil {
```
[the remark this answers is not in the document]
vivek-koppuru: Yea I think this is outdated

pkg/controller/clientutil/indexers_test.go
vivek-koppuru: I initially didn't find it but I did now, changing

pkg/controller/clusters/management.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+)
+
+// FetchManagementEksaCluster returns the management cluster object for a given workload Cluster.
+// If we are unable to find the management cluster using the same namespace as the current cluster, we will attempt
+// to get a list of the clusters with that name across all the namespaces. If we find multiple, which usually should
+// not happen as these clusters get mapped to a cluster-api cluster object in the eksa-system namespace, then we
+// also error on that because it is not possible to have multiple resources with the same name within a namespace.
+func FetchManagementEksaCluster(ctx context.Context, cli client.Client, cluster *v1alpha1.Cluster) (*v1alpha1.Cluster, error) {
```
g-gaston, line 18: I think we should write a test for this method
I'm ok doing it as a follow up if we need to get this is soon :)

vignesh-goutham (approved):

PR #6438, opened 2023-08-08
Retrieve management cluster namespace instead of assuming workload cluster namespace

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Builds off of #6414, however there are a few places outside of the controller where we make this assumption as well, which the `GetEksaCluster` function solves for.

*Testing (if applicable):*
unit testing

### DIFF
--- a/pkg/providers/tinkerbell/create.go
+++ b/pkg/providers/tinkerbell/create.go
@@ -152,7 +152,7 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 			return err
 		}
 
-		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, clusterSpec.Cluster.Namespace)
+		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, managementClusterSpec.Namespace)
 		if err != nil {
 			return fmt.Errorf("getting TinkerbellIP of management cluster: %s", err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/providers/tinkerbell/create.go
```diff
@@ -152,7 +152,7 @@ func (p *Provider) SetupAndValidateCreateCluster(ctx context.Context, clusterSpe
 			return err
 		}
 
-		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, clusterSpec.Cluster.Namespace)
+		managementDatacenterConfig, err := p.providerKubectlClient.GetEksaTinkerbellDatacenterConfig(ctx, managementClusterSpec.Spec.DatacenterRef.Name, clusterSpec.ManagementCluster.KubeconfigFile, managementClusterSpec.Namespace)
```
ddjjia, line 155: Do we have validation that dc needs to be with the same ns as mgmt cluster? If not, maybe we should add validation there?

vivek-koppuru: Were you able to create a cluster with two different namespaces for dc and cluster? Not sure if we have this validation for other providers, unless we do. I chose not to cover that because that wasn't what we saw the error for, but if we are specifically missing for tinkerbell I can add it. If we are missing for all the providers, might be better as a follow up.

vivek-koppuru: Looks like we are missing the following validations for tinkerbell: https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/vsphere.go#L53-L68

https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/tinkerbell.go#L11

However, I would vote to handle this as a separate PR then as it required adding those similar validations altogether.

g-gaston (commented):
/lgtm

pkg/providers/tinkerbell/create.go
vivek-koppuru: Created an issue here: https://github.com/aws/eks-anywhere/issues/6452

PR #6450, opened 2023-08-09
Add APIs for ETCD Encryption

abhinavmpandey08:
*Description of changes:*
This PR adds APIs to support ETCD Encryption

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -70,6 +70,10 @@ func (r *Cluster) ValidateCreate() error {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec"), r.Spec, err.Error()))
 	}
 
+	if r.Spec.EtcdEncryption != nil {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec"), r.Spec, "etcdEncryption is not supported during cluster creation"))
+	}
+
 	if len(allErrs) != 0 {
 		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
 	}
@@ -102,6 +106,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 	allErrs = append(allErrs, ValidateEksaVersionSkew(r, oldCluster)...)
 
 	allErrs = append(allErrs, ValidateWorkerKubernetesVersionSkew(r, oldCluster)...)
+
+	if err := validateEtcdEncryptionConfig(r.Spec.EtcdEncryption); err != nil {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec"), r.Spec, err.Error()))
+	}
 
 	if len(allErrs) != 0 {
 		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
--- a/pkg/api/v1alpha1/etcdencryption_test.go
+++ b/pkg/api/v1alpha1/etcdencryption_test.go
@@ -1,0 +1,1 @@
+package v1alpha1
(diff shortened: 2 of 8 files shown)


pokearu (approved):
/lgtm

pkg/api/v1alpha1/etcdencryption_types.go
```diff
@@ -0,0 +1,21 @@
...
+	Providers []EtcdEncryptionProvider `json:"providers"`
+	// Resources defines a list of objects and custom resources definitions that should be encrypted.
+	Resources []string `json:"resources"`
+}
+
+// EtcdEncryptionProvider defines the configuration for ETCD encryption providers.
+// Currently only KMS provider is supported.
+type EtcdEncryptionProvider struct {
+	// KMS defines the configuration for KMS Encryption provider.
+	KMS KMS `json:"kms"`
```
g-gaston, line 14: this probably need to be a pointer if we ever going to support different providers
So one can be filled and the other ones can be nill

pkg/api/v1alpha1/etcdencryption_test.go
g-gaston, line 1: remove file if we don't need tests here?

pkg/api/v1alpha1/etcdencryption_types.go
```diff
@@ -0,0 +1,21 @@
+package v1alpha1
+
+// EtcdEncryption defines the configuration for ETCD encryption.
+type EtcdEncryption struct {
+	Providers []EtcdEncryptionProvider `json:"providers"`
+	// Resources defines a list of objects and custom resources definitions that should be encrypted.
+	Resources []string `json:"resources"`
```
g-gaston, line 7: question about this bc I don't know how it works
would all resources we encrypted with all providers?

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -103,6 +107,10 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
 
 	allErrs = append(allErrs, ValidateWorkerKubernetesVersionSkew(r, oldCluster)...)
 
+	if err := validateEtcdEncryptionConfig(r.Spec.EtcdEncryption); err != nil {
```
g-gaston, line 110: do we need these validations in the CLI as well?

pkg/api/v1alpha1/etcdencryption_types.go
abhinavmpandey08: We don't have plans to support another encryption provider besides KMS. But I can it a pointer, shouldn't make too much difference.

pkg/api/v1alpha1/etcdencryption_types.go
abhinavmpandey08: You can have something like this where you configure one provider for some resources and another provider for other resources. 

```yaml
spec:
  etcdEncryption:
  - providers:
    - kms:
        socketListenAddress: unix:///var/run/kmsplugin/socket1-new.sock
    - kms:
        socketListenAddress: unix:///var/run/kmsplugin/socket1-old.sock
    resources:
    - secrets
    - custom-resource-1.anywhere.eks.amazonsaws.com
  - providers:
    - kms:
        socketListenAddress: unix:///var/run/kmsplugin/socket2-new.sock
    - kms:
        socketListenAddress: unix:///var/run/kmsplugin/socket2-old.sock
    resources:
    - configmaps
    - custom-resource-2.anywhere.eks.amazonsaws.com
```

In this example, `providers[0]` will be used to encrypt `secrets` and `custom-resource-1.anywhere.eks.amazonsaws.com` while `providers[1]` will be used to `configmaps` and `custom-resource-2.anywhere.eks.amazonsaws.com`

pkg/api/v1alpha1/cluster_webhook.go
abhinavmpandey08: yeah, I will add in this and more validations the CLI as well. I was thinking I would do it in a follow-up PR when I make other CLI changes. Does that sound good?

pokearu (approved):
/lgtm

PR #6454, opened 2023-08-09
Fix all worker node groups being rolled when labels adjusted

### LINKED ISSUE #6330: Adjusting labels on a single worker node group rolls out all worker node groups
**What Happened?**
When adding, removing or changing labels on a single worker node group it rolls out new nodes for all worker node groups. This is using the CLI; untested with the lifecycle controller.

**What should happen?**
Modifying the label configuration on a worker node group rolls out only the node associated with that worker node group.

**Analysis**
The issue was originally found using the CloudStack provider but the bug exists in the majority (probably all) providers.

The code [iterates over each worker node group checking if it requires a new template](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/cloudstack.go#L659). 
The call path ends in [`NeedsNewWorkloadTemplate`](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/cloudstack.go#L446). `NeedsNewWorkloadTemplate` [calls a function](https://github.com/aws/eks-anywhere/blob/main/pkg/providers/cloudstack/cloudstack.go#L451) to compare all worker node groups labels with their existing representations.

If there is > 1 worker node group, the algorithm will return true if there are changes to any worker node group labels.

chrisdoherty4:
Fixes #6330.

**Summary**
Manipulating labels on a single worker node group resulted in all worker node groups being rolled. This change ensures the worker node group with the adjusted labels is the only worker node group to be rolled out. 

The bug impacted the Docker, Tinkerbell, vSphere, Nutanix and CloudStack providers. It is only present for CLI based clusters.

Additional manual tests were run using the CloudStack provider.

Since the CLI will be changed to leverage the controller for all cases in future, no effort has been made to refactor the code so the fix isn't needed across all providers.

### DIFF
--- a/pkg/providers/cloudstack/cloudstack.go
+++ b/pkg/providers/cloudstack/cloudstack.go
@@ -448,7 +448,7 @@ func NeedsNewWorkloadTemplate(oldSpec, newSpec *cluster.Spec, oldCsdc, newCsdc *
 		return true
 	}
 	if !v1alpha1.WorkerNodeGroupConfigurationSliceTaintsEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
-		!v1alpha1.WorkerNodeGroupConfigurationsLabelsMapEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
+		!v1alpha1.MapEqual(oldWorker.Labels, newWorker.Labels) ||
 		!v1alpha1.WorkerNodeGroupConfigurationKubeVersionUnchanged(oldWorker, newWorker, oldSpec.Cluster, newSpec.Cluster) {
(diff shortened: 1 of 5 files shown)


abhinavmpandey08 (approved):

pkg/providers/cloudstack/cloudstack.go
```diff
@@ -448,7 +448,7 @@ func NeedsNewWorkloadTemplate(oldSpec, newSpec *cluster.Spec, oldCsdc, newCsdc *
 		return true
 	}
 	if !v1alpha1.WorkerNodeGroupConfigurationSliceTaintsEqual(oldSpec.Cluster.Spec.WorkerNodeGroupConfigurations, newSpec.Cluster.Spec.WorkerNodeGroupConfigurations) ||
```
abhinavmpandey08, line 450: Should we do the same thing here as well?

chrisdoherty4: Yes, I'll raise an issue to track independently and fix.

PR #6482, opened 2023-08-14
Fix worker node groups being rolled out when taints are changes

### LINKED ISSUE #6481: Adjusting taints on a single worker node group rolls out all worker node groups
See #6330 which describes the same problem for labels.

chrisdoherty4:
Fixes #6481 

When worker node group taints are adjusted it would roll out all worker node groups. This fix ensures the worker node group that has changed is the only group to be rolled out.


abhinavmpandey08 (approved):
/lgtm

PR #6496, opened 2023-08-15
Separate etcd upgrade changes from kcp changes

g-gaston:
## Description of changes
In order to have more control over the upgrade process and make sure
than in case of error the cleanup is easier, we upgrade etcd separately
from the CP nodes for external etcd topologies. The controller will
apply etcd changes in isolation and wait for etcd to be ready before
moving on with the CP nodes.

These changes allows us to implement [this](https://github.com/aws/eks-anywhere/blob/main/designs/kindless-upgrades.md#separating-etcd-from-cp-changes).

*Testing (if applicable):*
Tested manually with vsphere

### DIFF
--- a/pkg/controller/clusters/controlplane.go
+++ b/pkg/controller/clusters/controlplane.go
@@ -3,13 +3,18 @@ package clusters
 import (
 	"context"
 	"reflect"
+	"time"
 
 	etcdv1 "github.com/aws/etcdadm-controller/api/v1beta1"
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aws/eks-anywhere/pkg/controller"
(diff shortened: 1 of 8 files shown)


pkg/controller/clusters/controlplane.go
```diff
@@ -44,91 +49,180 @@ type ControlPlane struct {
...
-
+	// and updated by the kcp controller.
+	//
 	// We do not want to update the field with an empty slice again, so here we check if the endpoints for the
 	// external etcd have already been populated on the KubeadmControlPlane object and override ours before applying it.
-	kcp := &controlplanev1.KubeadmControlPlane{}
-	kcpKey := client.ObjectKey{
-		Name:      cluster.Spec.ControlPlaneRef.Name,
-		Namespace: cluster.Spec.ControlPlaneRef.Namespace,
+	externalEndpoints := currentKCP.Spec.KubeadmConfigSpec.ClusterConfiguration.Etcd.External.Endpoints
```
panktishah26, line 178: I get that ectdadm controller will remove etcdv1.UpgradeInProgressAnnotation annotation on the etcd cluster after it has been upgraded successfully but just to double check should we check the annotation doesn't exists when we upgrade the control plane?

panktishah26 (commented):
Everything looks great! I just have one question.

pkg/controller/clusters/controlplane.go
g-gaston: Ah interesting. Why do you think that could be necessary? The status contains the same information. The annotation is only there so the KCP/capi cluster controller can read it without knowing the whole etcdadmcluster API. But we do know the full api, so we don't need the validation.

panktishah26: I was just thinking that if we are adding the annotation during etcd upgrade, should we check that the annotation doesn't exist before we start the CP reconciliation. I got what you are saying. Thanks!

g-gaston:
Merging this manually. Codecov is at 75% and all untested code are just error branches coming from the k8s client api requests.

PR #6501, opened 2023-08-16
Error out when management kubeconfig is not present for workload cluster operations

rahulbabu95:
When management cluster kubeconfig is not set during workload cluster create/upgrade, eks-a doesn't fail and incorrectly sets management cluster to same as workload cluster when `clusterspec.ManagementCluster` value is nil which is determined based on kubeconfig file being explicitly passed or set as env variable. This change tries to additionally find the kubeconfig file locally if the env variable is not set for kubeconfig and errors out when not able to find any valid kubeconfig file.  

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -138,8 +138,9 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
 	}
 
 	if clusterSpec.Cluster.IsManaged() && options.managementKubeconfig == "" {
-		if kubeconfig.FromEnvironment() != "" {
-			options.managementKubeconfig = kubeconfig.FromEnvironment()
+		envKubeconfig := kubeconfig.FromEnvironment()
+		if envKubeconfig != "" {
+			options.managementKubeconfig = envKubeconfig
 		} else {
 			// check if kubeconfig for management cluster exists locally
 			managementKubeconfigPath := kubeconfig.FromClusterName(clusterSpec.Cluster.Spec.ManagementCluster.Name)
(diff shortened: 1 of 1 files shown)


cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -137,7 +138,17 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
 	}
 
 	if clusterSpec.Cluster.IsManaged() && options.managementKubeconfig == "" {
-		options.managementKubeconfig = kubeconfig.FromEnvironment()
+		if kubeconfig.FromEnvironment() != "" {
```
vivek-koppuru, line 141: Let's save the value in a variable so we avoid multiple calls

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -137,7 +138,18 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
...
+			managementKubeconfigPath := kubeconfig.FromClusterName(clusterSpec.Cluster.Spec.ManagementCluster.Name)
+			if validations.FileExistsAndIsNotEmpty(managementKubeconfigPath) {
+				options.managementKubeconfig = managementKubeconfigPath
+			} else {
+				return nil, fmt.Errorf("management kubeconfig file not found, must be present for workload cluster operations")
+			}
+		}
 	}
 
 	if options.managementKubeconfig != "" {
```
vivek-koppuru, line 155: If this is still empty, then the management cluster will not be set. Is that okay? Should the below logic go under the above if condition instead as ManagementCluster will only be set for managed workload clusters anyways?

vivek-koppuru (approved):

PR #6533, opened 2023-08-18
Update registrymirror doc

d8660091:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/docs/content/en/docs/getting-started/optional/registrymirror.md
+++ b/docs/content/en/docs/getting-started/optional/registrymirror.md
@@ -139,3 +139,24 @@ For [Mac](https://docs.docker.com/desktop/mac/#add-tls-certificates), you can fo
 {{% alert title="Note" color="primary" %}}
   You may need to restart Docker after adding the certificates.
 {{% /alert %}}
+
+## Registry configurations
+Depending on what registry you decide to use, you will need to create the following projects:
+
+```
+bottlerocket
+eks-anywhere
+eks-distro
+isovalent
+cilium-chart
+```
+
+For example, if a registry is available at `private-registry.local`, then the following projects will have to be created:
+
+```
+https://private-registry.local/bottlerocket
+https://private-registry.local/eks-anywhere
+https://private-registry.local/eks-distro
+https://private-registry.local/isovalent
+https://private-registry.local/cilium-chart
+```
--- a/docs/content/en/docs/getting-started/optional/registrymirror.md
+++ b/docs/content/en/docs/getting-started/optional/registrymirror.md
@@ -98,7 +98,7 @@ export REGISTRY_PASSWORD=<password>
 You can use the `download images` and `import images` commands to pull images from `public.ecr.aws` and push them to your
 private registry.
 The `copy packages` must be used if you want to copy EKS Anywhere Curated Packages to your registry mirror.
-The `download images` command also pulls the cilium chart from `public.ecr.aws` and pushes it to the registry mirror. It requires the registry credentials for performing a login. Set the following environment variables for the login:
+The `download images` command also pulls the Cilium chart from `public.ecr.aws` and pushes it to the registry mirror. It requires the registry credentials for performing a login. Set the following environment variables for the login:
 ```bash
 export REGISTRY_USERNAME=<username>
 export REGISTRY_PASSWORD=<password>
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/getting-started/optional/registrymirror.md
```diff
@@ -140,25 +139,3 @@ For [Mac](https://docs.docker.com/desktop/mac/#add-tls-certificates), you can fo
 {{% alert title="Note" color="primary" %}}
   You may need to restart Docker after adding the certificates.
 {{% /alert %}}
-
-## Registry configurations
```
jaxesn, line 144: shouldnt we keep this?  if the registry they are using requires these to exist ahead of time, don't they still need to do this?

docs/content/en/docs/getting-started/optional/registrymirror.md
```diff
@@ -101,7 +100,6 @@ private registry.
 The `copy packages` must be used if you want to copy EKS Anywhere Curated Packages to your registry mirror.
 The `download images` command also pulls the cilium chart from `public.ecr.aws` and pushes it to the registry mirror. It requires the registry credentials for performing a login. Set the following environment variables for the login:
```
csplinter, line 101: Nit: capitalize "Cilium"

TerryHowe:
```suggestion
The `download images` command also pulls the Cilium chart from `public.ecr.aws` and pushes it to the registry mirror. It requires the registry credentials for performing a login. Set the following environment variables for the login:
```

docs/content/en/docs/getting-started/optional/registrymirror.md
```diff
@@ -50,7 +50,7 @@ spec:
 * __Example__: ```port: 443```
 
 ### __ociNamespaces__ (optional)
-* __Description__: mapping from upstream registries to the locations of the private registry. When specified, the artifacts pulled from an upstream registry will be put in its corresponding location/namespace in the private registry. The target location/namespace must be already existing.
+* __Description__: when you need to mirror multiple registries, you can map the upstream registries to "namespaces" of the mirror.
```
chrisnegus, line 53: Is this a one-to-one mapping of a registry to a namespace? If so, I'd suggest wording like:
```suggestion
* __Description__: when you need to mirror multiple registries, you can map each upstream registry to the "namespace" of its mirror.
```
Let me know if I misunderstand you.

chrisnegus (commented):
I just have one question about the wording of registryMirrorConfiguration.

docs/content/en/docs/getting-started/optional/registrymirror.md
abhinavmpandey08: yeah I think this section is still applicable

d8660091: Is it only required for harbor? I never created those projects for my docker registry server running locally?

docs/content/en/docs/getting-started/optional/registrymirror.md
d8660091: sounds better to me, thanks!

docs/content/en/docs/getting-started/optional/registrymirror.md
abhinavmpandey08: i know it's needed for harbor and i think another one as well, maybe artifactory

jaxesn: yeah we should keep this.

d8660091: added back

docs/content/en/docs/getting-started/optional/registrymirror.md
```diff
@@ -152,8 +151,7 @@ isovalent
 cilium-chart
 ```
 
-For example, if a registry is available at `private-registry.local`, then the following 
-projects will have to be created:
+For example, if a registry is available at `private-registry.local`, then the following projects will have to be created:
```
drewvanstone, line 154: There is nothing following this statement? Is there supposed to be an example here that's missing?

drewvanstone (changes requested):
Looks like we're missing an example or something. Everything else looks good from a writing perspective. Haven't tested it for correctness.

docs/content/en/docs/getting-started/optional/registrymirror.md
d8660091: is it wrapped by github UI?

drewvanstone (approved):
Yep you're right. Github preview was truncating the doc

PR #6541, opened 2023-08-21
Add verify auto generate files make target

cxbrowne1207:
*Issue #, if available:*
This PR adds a make target to verify auto-generated files are up to date by using a git diff after running the commands that generate them. The intention is to use this in a prow job triggered when making PRs to the repository. For example, we could update the [eks-anywhere-mocks-presubmits]( https://github.com/aws/eks-anywhere-prow-jobs/blob/main/jobs/aws/eks-anywhere/eks-anywhere-mocks-presubmits.yaml) to run this make target and check all of these generated files instead of only the mocks.

*Description of changes:*

*Testing (if applicable):*


Makefile
```diff
@@ -692,6 +701,20 @@ generate-core-manifests: $(CONTROLLER_GEN) ## Generate manifests for the core pr
...
+.PHONY: verify-generate-manifests
+verify-generate-manifests: generate-manifests ## Verify if generated manifests files e.g. CRD, RBAC etc. need to be updated
+	$(eval DIFF=$(shell git diff --raw -- '*.go' | wc -c))
+	if [[ $(DIFF) != 0 ]]; then \
+		echo "Detected out of date manifest files. please run 'make generate-manifests'"; \
+		exit 1;\
+	fi
+
+.PHONY: verify-auto-gen-files
+verify-auto-gen-files:
```
chrisdoherty4, line 713: 'auto-gen' is a little misleading. If they're automatically generated, why verify at all, it should be a solved problem? :D

PR #6551, opened 2023-08-22
fix upgrade from latest e2e

tatlat:
*Issue #, if available:*

*Description of changes:*
Preflight validations currently see upgrading from a release build to dev build as a downgrade. This change lets you upgrade with a dev build in this scenario. 

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster_webhook.go
+++ b/pkg/api/v1alpha1/cluster_webhook.go
@@ -32,6 +32,8 @@ import (
 
 const supportedMinorVersionIncrement int64 = 1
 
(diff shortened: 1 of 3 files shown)


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -32,6 +32,8 @@ import (
 
 const supportedMinorVersionIncrement int64 = 1
 
+var devBuild, _ = semver.New("v0.0.0-dev")
```
jaxesn, line 35: You may need to handle the v0.0.0-dev-release-0.17 case as well here. get abhay to confirm the release branch release numbers.

Issue #6452, opened 2023-08-09
Missing cluster level validations for Tinkerbell


vivek-koppuru:
As discussed in https://github.com/aws/eks-anywhere/pull/6438#discussion_r1288054139, we are missing cluster level validations for Tinkerbell that will check thinks like making sure all the subresources are in the same namespace. https://github.com/aws/eks-anywhere/blob/main/pkg/cluster/tinkerbell.go

Missing validations present in vsphere for example: https://github.com/aws/eks-anywhere/blob/848f820e5cf2d807a2d42fd822fcbb8f61d3770b/pkg/cluster/vsphere.go#L38-L71

PR #6556, opened 2023-08-22
cloudstack cluster/dc/mc should use same namespace

ddjjia:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/6452

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/tinkerbell.go
+++ b/pkg/cluster/tinkerbell.go
@@ -2,7 +2,6 @@ package cluster
 
 import (
 	"context"
-
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/cluster/validate_test.go
```diff
@@ -75,3 +76,52 @@ func TestValidateConfigDifferentNamespace(t *testing.T) {
...
+			clusterNS: constants.EksaSystemNamespace,
+			dcNS:      constants.EksaSystemNamespace,
+			mcNS:      "mc-ns",
+			wantErr:   "CloudStackMachineConfig and Cluster objects must have the same namespace specified",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			config := clusterConfigFromFile(t, "testdata/cluster_1_20_cloudstack.yaml")
```
vivek-koppuru, line 114: Since we are only testing for the namespace, can we create the config through code like the above function rather than a file? In general we want to remove the number of references to files that we have

pkg/cluster/tinkerbell.go
```diff
@@ -26,6 +26,24 @@ func tinkerbellEntry() *ConfigManagerEntry {
 			machineConfigsProcessor(processTinkerbellMachineConfig),
 			processTinkerbellTemplateConfigs,
 		},
+		Validations: []Validation{
+			func(c *Config) error {
+				if c.TinkerbellDatacenter != nil {
+					if err := validateSameNamespace(c, c.TinkerbellDatacenter); err != nil {
+						return err
```
vivek-koppuru, line 33: This is so weird that it is not detecting this, do you happen to know why that might be the case? Like if we add a print statement here, do you see the unit test go through this code path?

ddjjia: cuz the test uses cloudstack not tinkerell 😂 fixed now

vivek-koppuru (approved):

Issue #5905, opened 2023-05-23
`--kubeconfig` not honored with `upgrade cluster` command


jacobweinstock:
**What happened**:
When running `upgrade cluster` the `--kubeconfig` flag is not honored and the process fails.

```bash
$ ./eksctl-anywhere upgrade cluster -z hardware.csv -f bottlerocket.yaml --kubeconfig ./management-cluster.kubeconfig -v12
2023-05-23T17:52:49.602Z	V6	Executing command	{"cmd": "/usr/bin/docker version --format {{.Client.Version}}"}
2023-05-23T17:52:49.664Z	V6	Executing command	{"cmd": "/usr/bin/docker info --format '{{json .MemTotal}}'"}
Error: failed to upgrade cluster: common validations failed due to: validating kubeconfig "bottlerocket/bottlerocket-eks-a-cluster.kubeconfig": file does not exist
```

**What you expected to happen**:
The upgrade command should honor the `--kubeconfig` flag.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a management cluster
2. Move the kubeconfig to a new location
3. run a cluster upgrade specifying the location of the management kubeconfig with `--kubeconfig`

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

ddjjia:
- bug fixed by https://github.com/aws/eks-anywhere/pull/6666
- to deprecate `wconfig` and get wl kubeconfig from mgmt cluster https://github.com/aws/eks-anywhere/issues/6598 
- to remove wconfig flag https://github.com/aws/eks-anywhere/issues/6673

PR #6573, opened 2023-08-25
CLI deprecate wl kubeconfig flag

ddjjia:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/5905

*Description of changes:*
- deprecate `wConfig` in update and delete
    - exit when use `wConfig`
- rename `wConfig` to `clusterConfig` to avoid confusion

*Testing (if applicable):*
- manual testing
    - create mgmt cluster 
    - rename cluster dir
    - use --w-config leads to error
    - use --kubeconfig to upgrade
    - use --kubeconfig to delete

### DIFF
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 
 	"github.com/spf13/cobra"
 
@@ -38,6 +39,10 @@ var deleteClusterCmd = &cobra.Command{
 		if err := dc.deleteCluster(cmd.Context()); err != nil {
 			return fmt.Errorf("failed to delete cluster: %v", err)
 		}
+		if uc.wConfig != "" {
+			logger.MarkFail(wConfigDeprecationMessage)
+			return errors.New("please remove the --w-config flag and use --kubeconfig instead")
+		}
 		return nil
 	},
 }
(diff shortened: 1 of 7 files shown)


cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -160,7 +158,7 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
 
 	workloadCluster := &types.Cluster{
 		Name:           clusterSpec.Cluster.Name,
-		KubeconfigFile: getKubeconfigPath(clusterSpec.Cluster.Name, uc.wConfig),
+		KubeconfigFile: getKubeconfigPath(clusterSpec.Cluster.Name, uc.managementKubeconfig),
```
ddjjia, line 161: Im thinking if we should name it `kubeconfig` or `clusterKubeconfig`

vivek-koppuru (commented):
Deprecate technically means to not remove but add a warning message that we will, but there might not be a benefit to that either. Should we say that this isn't used and remove in a minor release, or just remove the flag during a patch release?

cmd/eksctl-anywhere/cmd/deletecluster.go
```diff
@@ -38,6 +39,10 @@ var deleteClusterCmd = &cobra.Command{
 		if err := dc.deleteCluster(cmd.Context()); err != nil {
 			return fmt.Errorf("failed to delete cluster: %v", err)
 		}
+		if uc.wConfig != "" {
+			logger.MarkFail(wConfigDeprecationMessage)
+			return errors.New("please remove the --w-config flag and use --kubeconfig instead")
```
vivek-koppuru, line 44:
```suggestion
			return errors.New("--w-config is deprecated. Use --kubeconfig instead")
```

cmd/eksctl-anywhere/cmd/supportbundle.go
```diff
@@ -97,7 +109,7 @@ func (csbo *createSupportBundleOptions) createBundle(ctx context.Context, since,
 	}
 	defer close(ctx, deps)
 
-	supportBundle, err := deps.DignosticCollectorFactory.DiagnosticBundle(clusterSpec, deps.Provider, getKubeconfigPath(clusterSpec.Cluster.Name, csbo.wConfig), bundleConfig)
+	supportBundle, err := deps.DignosticCollectorFactory.DiagnosticBundle(clusterSpec, deps.Provider, getKubeconfigPath(clusterSpec.Cluster.Name, csbo.clusterKubeconfig), bundleConfig)
```
vivek-koppuru, line 112: What happens if this is passed in as empty? Would it be able to get the right kubeconfig to get the bundle?

ddjjia: https://github.com/aws/eks-anywhere/blob/d975bda33508cae68d1368f66f0e61370d1ae538/cmd/eksctl-anywhere/cmd/common.go#L46
If no kubeconfig is specified in cmd, it use default path to locate kubeconfig file, which is the same path upon creation. The bug was reproduced by move or rename the kubeconfig file/dir. So if customers have to provide kubeconfig file path if they change it, otherwise it only use the default path

vivek-koppuru (approved):

PR #6584, opened 2023-08-28
Fix bare metal upgrade with custom pod CIDR

### LINKED ISSUE #6442: Hegel trusted proxies reverts to 192.168.0.0/16 on upgrade (scale up)
**What happened?**
When upgrading a bare metal cluster that uses a customized pod CIDR network the trusted proxies nodes were not provisioning beyond OS launch (i.e. cloud-init didn't run its usual commands to put SSH keys on the node and run kubeadm).

**What you expected to happen?**
Bare metal clusters using custom pod CIDR networks can be upgraded.

**How to reproduce it (as minimally and precisely as possible)?**
Create a bare metal cluster using a custom pod CIDR network and attempt to scale up. Observe the pod CIDR revert to the default in Hegel's trusted proxy environment variable.

**Environment**
EKS Anywhere v0.16.2

**Analysis**
On upgrade the Tinkerbell stack Helm chart is reapplied. The expectation is that no configuration would have changed in the scale up and therefore nothing changes. However, the trusted proxies does change reverting to the default 192.168.0.0/16 network. This prevents nodes from retrieving user-data because Envoy fronts the Tinkerbell stack so will have an untrusted IP from the customized pod CIDR.

chrisdoherty4:
Closes #6442.

When upgrading bare metal clusters that use a custom pod CIDR Hegel looses the pod CIDR as a trusted proxy. It can be caused by either an upgrade of the management cluster or upgrade of a workload cluster. 

The problem stems from trusted proxies not being included in the values when upgrading the Tinkerbell helm chart and incorrectly identifying the need to upgrade when upgrading workload clusters.

This change has undergone extensive manual testing with both management and workload cluster creations and upgrades (scale ups and downs). 

Additionally, I found a programmer error in our unit tests where we incorrect deserialized YAML into map data structures which resulted in the comparison of 2 empty maps. Correcting the error revealed numerous unit test failures. I reworked some parts of the code to bring about determinism and avoid re-writing tests to accommodate for the determinism due to the time pressure. I'll circle back and rework some of the code to tidy it up in a subsequent PR.

### DIFF
--- a/pkg/providers/tinkerbell/stack/testdata/expected_with_docker_options.yaml
+++ b/pkg/providers/tinkerbell/stack/testdata/expected_with_docker_options.yaml
@@ -4,14 +4,14 @@ boots:
   - -osie-path-override=https://anywhere-assests.eks.amazonaws.com/tinkerbell/hook
   deploy: false
   env:
-  - name: TINKERBELL_GRPC_AUTHORITY
-    value: 1.2.3.4:42113
-  - name: BOOTS_EXTRA_KERNEL_ARGS
-    value: tink_worker_image=public.ecr.aws/eks-anywhere/tink-worker:latest
   - name: DATA_MODEL_VERSION
     value: kubernetes
   - name: TINKERBELL_TLS
     value: "false"
+  - name: TINKERBELL_GRPC_AUTHORITY
+    value: 1.2.3.4:42113
+  - name: BOOTS_EXTRA_KERNEL_ARGS
+    value: tink_worker_image=public.ecr.aws/eks-anywhere/tink-worker:latest
   image: public.ecr.aws/eks-anywhere/boots:latest
 createNamespace: false
 envoy:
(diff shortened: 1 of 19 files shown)


pkg/providers/tinkerbell/stack/testdata/expected_with_docker_options.yaml
```diff
@@ -34,8 +34,7 @@ rufio:
 tinkController:
   image: public.ecr.aws/eks-anywhere/tink-controller:latest
 tinkServer:
-  args:
-  - --tls=false
```
cxbrowne1207, line 38: I noticed that TLS is still set to false in get boots env. Was this never necessary or is this related to something else that change for this to be removed?

chrisdoherty4: It wasn't necessary.

PR #6621, opened 2023-09-01
Add workflow for management kindless upgrades

panktishah26:
This branchs out the code at the command level and uses a different workflow if kindless upgrades are enabled. This facilitates the transition once we are ready to flip the switch and avoids having to maintain multiple codepaths in the existing workflow, which from experience, it has proven to be difficult and prone to errors. The tradeoff is we need to mainetain the same functionality in two workflows until we decide to deprecate the old one.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


panktishah26:
Closing this PR as this is covered in #6622

PR #6622, opened 2023-09-01
Add workflow for management kindless upgrades vSphere

panktishah26:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/blob/main/designs/kindless-upgrades.md

*Description of changes:*
This branch out the code at the command level and uses a different workflow if kindless upgrades are enabled. This facilitates the transition once we are ready to flip the switch and avoids having to maintain multiple codepaths in the existing workflow, which from experience, it has proven to be difficult and prone to errors. The tradeoff is we need to maintain the same functionality in two workflows until we decide to deprecate the old one.

The feature is under feature flag. Export EXP_SELF_MANAGED_API_UPGRADE and ExpSelfManagedAPIUpgrade flags to true to enable this feature for vSphere Management cluster. This PR doesn't affect the current flow of workload cluster upgrade process

*Testing (if applicable):*


pkg/workflows/management/upgrade_cluster.go
```diff
@@ -0,0 +1,72 @@
...
+			logger.Info("Backing up management components from bootstrap cluster")
+			err := commandContext.ClusterManager.BackupCAPIWaitForInfrastructure(ctx, commandContext.BootstrapCluster, fmt.Sprintf("bootstrap-%s", commandContext.ManagementClusterStateDir), commandContext.ManagementCluster.Name)
+			if err != nil {
+				logger.Info("Bootstrap management component backup failed, use existing workload cluster backup", "error", err)
+			}
+		}
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	// TODO: move to its own task
```
g-gaston, line 49: can we tackle this? it seems super easy

pkg/workflows/management/upgrade_cluster.go
```diff
@@ -0,0 +1,72 @@
...
+		if err := commandContext.ClusterManager.ApplyBundles(ctx, commandContext.ClusterSpec, commandContext.ManagementCluster); err != nil {
+			commandContext.SetError(err)
+			return &workflows.CollectMgmtClusterDiagnosticsTask{}
+		}
+
+		if err := commandContext.ClusterManager.ApplyReleases(ctx, commandContext.ClusterSpec, commandContext.ManagementCluster); err != nil {
+			commandContext.SetError(err)
+			return &workflows.CollectMgmtClusterDiagnosticsTask{}
+		}
+	}
```
g-gaston, line 33: Could we move this to a different task as well?
What do you think, I see this different than an "upgrade"
Is more like "install new eks-a version components"

pkg/workflows/management/upgrade_test.go
```diff
@@ -0,0 +1,648 @@
...
+	)
+}
+
+func (c *upgradeManagementTestSetup) expectApplyBundles(err error) {
+	gomock.InOrder(
+		c.clusterManager.EXPECT().ApplyBundles(
+			c.ctx, c.newClusterSpec, c.managementCluster,
+		).Return(err),
+	)
+	// c.clusterManager.EXPECT().ApplyReleases(c.ctx, c.newClusterSpec, c.managementCluster),
```
g-gaston, line 210: remove?

pkg/cluster/wait.go
```diff
@@ -18,23 +18,29 @@ import (
 func WaitForCondition(ctx context.Context, client kubernetes.Reader, cluster *anywherev1.Cluster, retrier *retrier.Retrier, conditionType anywherev1.ConditionType) error {
 	return retrier.Retry(func() error {
 		c := &anywherev1.Cluster{}
-		if err := client.Get(ctx, cluster.Name, cluster.Namespace, c); err != nil {
+
+		namespace := cluster.Namespace
+		if namespace == "" {
+			namespace = "default"
```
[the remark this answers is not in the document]
panktishah26: It used to give me below error as it wasn't considering the namespace if we don't provide one. Thus, I added this block. Let me know if you want me to add this somewhere else. 
`if Name is specified, Namespace is required", "errorVerbose": "if Name is specified, Namespace is required`

pkg/cluster/wait.go
```diff
@@ -18,23 +18,29 @@ import (
...
+		fmt.Println(namespace)
+		if err := client.Get(ctx, cluster.Name, namespace, c); err != nil {
 			return err
 		}
 
 		observedGeneration := c.Status.ObservedGeneration
 		generation := c.Generation
 		if observedGeneration != generation {
-			return errors.Errorf("cluster generation (%d) and observedGeneration (%d) differ", generation, observedGeneration)
+			return fmt.Errorf("cluster generation (%d) and observedGeneration (%d) differ", generation, observedGeneration)
```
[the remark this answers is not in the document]
panktishah26: `errors.Errorf` includes stack trace like below which we would not like to include so I changed it to `fmt.Errorf` which only includes an error message.
`cluster generation (6) and observedGeneration (5) differ", "errorVerbose": "cluster generation (6) and observedGeneration (5) differ\ngithub.com/aws/eks-anywhere/pkg/cluster.WaitForCondition.func1\n\t/Users/xxx/workspace/eks-anywhere/pkg/cluster/wait.go:34\ngithub.com/aws/eks-anywhere/pkg/retrier.(*Retrier).Retry\n\t/Users/xxx/workspace/eks-anywhere/pkg/retrier/retrier.go:84..........`

pkg/workflows/diagnostics.go
```diff
@@ -53,6 +54,11 @@ func (s *CollectWorkloadClusterDiagnosticsTask) Name() string {
 
 func (s *CollectMgmtClusterDiagnosticsTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
 	logger.Info("collecting management cluster diagnostics")
+	if features.IsActive(features.ExperimentalSelfManagedClusterUpgrade()) {
```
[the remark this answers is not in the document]
panktishah26: I agree the toggle can be on during workload cluster but if the toggle is off, in terms of management cluster, wouldn't the kubeconfig of `commandContext.ManagementCluster` and `commandContext.BootstrapCluster` be different? I guess it would not save the logs of bootstrap cluster if we change `commandContext.BootstrapCluster` to a `commandContext.ManagementCluster`.

pkg/workflows/management/gitops.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	logger.Info("Resuming GitOps cluster resources kustomization")
+	err = commandContext.GitOpsManager.ResumeClusterResourcesReconcile(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec, commandContext.Provider)
+	if err != nil {
+		commandContext.SetError(err)
+		return &writeClusterConfig{}
+	}
+	// TODO(pjshah): check where to incorporate this field
+	// if !s.eksaSpecDiff {
```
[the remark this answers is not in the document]
panktishah26: Yeah, I commented this out because I want to evaluate if this is needed or not before removing it from the code. We do use this field for kind upgrade operation.

pkg/workflows/management/upgrade_cluster.go
panktishah26: I think we should keep this under the upgrade process because to me it appears as a part of upgrade process. I am happy to move this around though.

pkg/cluster/wait.go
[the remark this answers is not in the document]
g-gaston: Why wouldn't we want that?
I find that info quite useful
What's your concern?

pkg/cluster/wait.go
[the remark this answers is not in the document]
g-gaston: I understand that, namespace is required 
What I was saying is I thought we were already setting the namespace to default somewhere else so I was wondering how we got this error?

pkg/workflows/diagnostics.go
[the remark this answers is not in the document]
g-gaston: Management cluster field should point to whichever is the cluster with the management resources at any point in time
Maybe we need to verify that but I think that's how it's supposed to work

pkg/workflows/management/gitops.go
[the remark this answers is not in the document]
g-gaston: Sweet. Do you mind writing that down in the comment?
And just say in the description you will deal with it in a separate or

pkg/workflows/management/upgrade_cluster.go
g-gaston: Interesting
Why do think it's part of the upgrade step?
These two operations don't make any changes to the cluster or affect any component running

pkg/workflows/diagnostics.go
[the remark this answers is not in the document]
panktishah26: For the current upgrade workflow, We move management resources to bootstrap cluster after bootstrap cluster creation and it points to a different kubeconfig file. I guess we should keep it to bootstrap cluster for now untill we remove kind upgrade entirely.

pkg/cluster/wait.go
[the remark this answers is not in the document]
panktishah26: Sorry I couldn't find it in the kindless upgrade workflow.

pkg/workflows/management/core_components.go
```diff
@@ -0,0 +1,110 @@
...
+
+type ensureEtcdCAPIComponentsExist struct{}
+
+// Run ensureEtcdCAPIComponentsExist ensures ETCD CAPI providers on the management cluster.
+func (s *ensureEtcdCAPIComponentsExist) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Ensuring etcd CAPI providers exist on management cluster before upgrade")
+	if err := commandContext.CAPIManager.EnsureEtcdProvidersInstallation(ctx, commandContext.ManagementCluster, commandContext.Provider, commandContext.CurrentClusterSpec); err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
```
abhinavmpandey08, line 20: Why do we need a special check just for etcd controller? Why not check the other CAPI controllers?

pkg/workflows/management/gitops.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+)
+
+type pauseGitOpsReconcile struct{}
+
+// Run pauseGitOpsReconcile pause GitOps reconciler before management cluster upgrade.
+func (s *pauseGitOpsReconcile) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Pausing GitOps cluster resources reconcile")
+	err := commandContext.GitOpsManager.PauseClusterResourcesReconcile(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec, commandContext.Provider)
```
abhinavmpandey08, line 16: Should we use current cluster spec here instead so it tries to pause the current objects and not any new ones?
```suggestion
	err := commandContext.GitOpsManager.PauseClusterResourcesReconcile(ctx, commandContext.ManagementCluster, commandContext.CurrentClusterSpec, commandContext.Provider)
```

pkg/workflows/management/core_components.go
```diff
@@ -0,0 +1,110 @@
...
+	if err = commandContext.GitOpsManager.Install(ctx, commandContext.ManagementCluster, commandContext.CurrentClusterSpec, commandContext.ClusterSpec); err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	changeDiff, err = commandContext.GitOpsManager.Upgrade(ctx, commandContext.ManagementCluster, commandContext.CurrentClusterSpec, commandContext.ClusterSpec)
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
```
abhinavmpandey08, line 72: This looks a little weird? Why is GitOps being installed and then upgraded right away?
If it's being installed fresh, shouldn't the latest version be installed already?

pkg/workflows/management/pre_cluster_upgrade.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+)
+
+type preClusterUpgrade struct{}
+
+// Run preClusterUpgrade implements steps to be performed before management cluster's upgrade.
+func (s *preClusterUpgrade) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	// Take best effort CAPI backup of workload cluster without filter.
+	// If that errors, then take CAPI backup filtering on only workload cluster.
```
abhinavmpandey08, line 16: Does this comment mean we initially try to backup all CAPI components and if that fails then we try to backup only the CAPI components for the cluster being upgraded?

If that's what we are trying to do, then this doesn't seem accurate. I think this will create a backup of only the management cluster's CAPI components and not the workload clusters

```
commandContext.ClusterManager.BackupCAPI(ctx, commandContext.ManagementCluster, commandContext.ManagementClusterStateDir, commandContext.ManagementCluster.Name)
```

pkg/workflows/management/upgrade_cluster.go
```diff
@@ -0,0 +1,52 @@
...
+		commandContext.SetError(err)
+		// TODO(@pjshah): check if we need this or not
+		// Take backup of bootstrap cluster capi components
+		// if commandContext.BootstrapCluster != nil {
+		// 	logger.Info("Backing up management components from bootstrap cluster")
+		// 	err := commandContext.ClusterManager.BackupCAPIWaitForInfrastructure(ctx, commandContext.BootstrapCluster, fmt.Sprintf("bootstrap-%s", commandContext.ManagementClusterStateDir), commandContext.ManagementCluster.Name)
+		// 	if err != nil {
+		// 		logger.Info("Bootstrap management component backup failed, use existing workload cluster backup", "error", err)
+		// 	}
+		// }
```
abhinavmpandey08, line 33: I think you already have a task that takes a backup of CAPI components so this shouldn't be needed

pkg/workflows/management/gitops.go
```diff
@@ -0,0 +1,91 @@
...
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	logger.Info("Forcing reconcile Git repo with latest commit")
+	err = commandContext.GitOpsManager.ForceReconcileGitRepo(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec)
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
```
abhinavmpandey08, line 63: These two don't really fit into `resumeGitOps` since it's more reconciling stuff instead of just resuming. Might want to rename this task or split this up

pkg/workflows/management/core_components.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *ensureEtcdCAPIComponentsExist) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
+	return &pauseGitOpsReconcile{}, nil
+}
+
+type upgradeCoreComponents struct {
+	UpgradeChangeDiff *types.ChangeDiff
+}
+
+// Run upgradeCoreComponents upgrades pre cluster upgrade components.
+func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
```
abhinavmpandey08, line 43: Where is it upgrading cilium? That's also part of core components upgrade right? I didn't see it in this PR

cmd/eksctl-anywhere/cmd/upgradecluster.go
```diff
@@ -184,7 +176,36 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command) error {
 
 	upgradeValidations := upgradevalidations.New(validationOpts)
 
-	err = upgradeCluster.Run(ctx, clusterSpec, managementCluster, workloadCluster, upgradeValidations, uc.forceClean)
+	if features.ExperimentalSelfManagedClusterUpgrade().IsActive() && clusterConfig.IsSelfManaged() {
+		logger.Info("Management kindless upgrade")
```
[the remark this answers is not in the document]
panktishah26: Yes, we will remove this along with the feature flag.

pkg/workflows/management/core_components.go
panktishah26: I guess we install flux controllers during core components upgrade if it does not exist before in the cluster spec.  https://github.com/aws/eks-anywhere/pull/2759
It is a part of current workflow https://github.com/aws/eks-anywhere/blob/main/pkg/workflows/upgrade.go#L251-L260 and we haven't changed it.

pkg/workflows/management/upgrade_cluster.go
panktishah26: Yes, just want to double check before removing it.

abhinavmpandey08 (approved):
/lgtm

pkg/workflows/management/gitops.go
panktishah26: Discussed this offline. `ClusterSpec` works with current flow.

pkg/workflows/management/upgrade_cluster.go
panktishah26: Yes, I got your point. I have moved this to a different task.

pkg/workflows/diagnostics.go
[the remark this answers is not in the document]
g-gaston: It seems like we do
https://github.com/aws/eks-anywhere/blob/main/pkg/workflows/upgrade.go#L492

pkg/workflows/management/install_new_components.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+)
+
+type installNewComponents struct{}
+
+// Run installNewComponents performs actions needed to upgrade the management cluster.
+func (s *installNewComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	if commandContext.UpgradeChangeDiff.Changed() {
```
g-gaston, line 14: I think we should do the apply no matter what
Even if `UpgradeChangeDiff` doesn't have any diff, it's possible that we have a new bundle, hence the controller will need it.

g-gaston: So I'm proposing removing this `if`

pkg/workflows/management/upgrade_cluster.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+)
+
+type upgradeCluster struct{}
+
+// Run upgradeCluster performs actions needed to upgrade the management cluster.
+func (s *upgradeCluster) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	// TODO(g-gaston): move this to eks-a installer and eks-d installer
+	err := commandContext.EksdInstaller.InstallEksdManifest(ctx, commandContext.ClusterSpec, commandContext.ManagementCluster)
```
g-gaston, line 16: I think this should be part of `installNewComponents`

PR #6656, opened 2023-09-07
Update API upgrade test to specify use appropriate templates

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This CR updates from latest release API tests to:
1. Update the workload cluster config machines templates before creating the cluster with the old bundle.  

At the step of creating a new workload cluster with the old Eksa version, the test may fail validating template tags (particularly for vSphere)) if the old bundle is using a different eksd tags. So, we need to also update the workload cluster machine templates with a template appropriate for the Kubernetes version and the old release. 

*Testing (if applicable):*


cxbrowne1207:
/hold

Running the tests locally again to ensure they're passing

PR #6667, opened 2023-09-08
Fix side effect of unnecessary workload cluster machine rollout 

jiayiwang7:
... during management cluster upgrade due to package reconcile error

*Issue #, if available:*

E2E test `TestDockerKubernetes126WithOIDCManagementClusterUpgradeFromLatestSideEffects` constantly failing with error:

```sh
2023-09-07T20:58:21.360Z	V0	🎉 Cluster upgraded!
2023-09-07T20:58:21.362Z	V4	Task finished	{"task_name": "delete-kind-cluster", "duration": "3.177917836s"}
2023-09-07T20:58:21.362Z	V4	----------------------------------
2023-09-07T20:58:21.362Z	V4	Tasks completed	{"duration": "18m44.684232162s"}
2023-09-07T20:58:21.364Z	V3	Cleaning up long running container	{"name": "eksa_1694119167170291469"}
    side_effects.go:150: Checking for changes in machines for 10m0s
    side_effects.go:155: Reading CAPI machines for cluster main-i-067a6-ccc8a7c-w-0
    side_effects.go:52: The new controller has had cascading changes: cluster main-i-067a6-ccc8a7c-w-0 has chaged: Different number of machines: before 7, after 8
--- FAIL: TestCloudStackKubernetes123WithOIDCManagementClusterUpgradeFromLatestSideEffects (2322.45s)
FAIL
```

Same error shows up across providers. It indicates that the management cluster upgrade causes side effect of unnecessary workload cluster machine rollout.

Upon investigation, we found out that the EKS-A workload cluster object does not have `childrenReconciledGeneration` and `reconciledGeneration` which is used to determine [whether to reconcile the workload cluster or not](https://github.com/aws/eks-anywhere/blob/82807d01e96e598650e42ab2df1f3660960493ce/controllers/cluster_controller.go#L299-L305). The two generation fields are only get added [here](https://github.com/aws/eks-anywhere/blob/82807d01e96e598650e42ab2df1f3660960493ce/controllers/cluster_controller.go#L353-L354) after cluster provider reconciles are done. In this case, the `postClusterProviderReconcile` returns an error where the [curated packages reconcile](https://github.com/aws/eks-anywhere/blob/82807d01e96e598650e42ab2df1f3660960493ce/controllers/cluster_controller.go#L412-L418) fails at installing packages controller, thus the two generation fields are not added to the cluster status. When user triggers a management cluster upgrade without this workload cluster fully reconciled, it can cause side effect of rolling out workload cluster nodes.

*Description of changes:*

As short term fix, we move the packages reconcile out of the `postClusterProviderReconcile` and after the two generation fields are set - so that a failing packages reconcile will not affect the core cluster reconcile and not trigger side effect of unnecessary workload cluster machine rollout.

Long term, we should properly handle packages reconcile error and not triggering machine upgrade when packages reconcile is still in progress. Similarly, figure out how to handle other post cluster reconciler like IAM auth and MHC reconcile error and make sure it does not cause side effect.

*Testing (if applicable):*

TODO

N/A

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -353,6 +353,19 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, log logr.Logger, clus
 	cluster.Status.ReconciledGeneration = cluster.Generation
 	cluster.Status.ChildrenReconciledGeneration = aggregatedGeneration
 
+	// TODO(eksa-controller-SME): properly handle packages reconcile error and not triggering machine upgrade when
+	// packages reconcile is still in progress.
+	// Moving the packages reconcile after the above two generation fields are set, so that packages reconcile error
+	// does not cause side effect of rolling out of workload cluster machines during management cluster upgrade.
+	reconcileResult, err = r.packagesReconcile(ctx, log, cluster)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if reconcileResult.Return() {
+		return reconcileResult.ToCtrlResult(), nil
+	}
+
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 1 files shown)


abhinavmpandey08 (approved):
/lgtm

PR #6674, opened 2023-09-11
Prevent bare metal machine config references from changing to existing machine configs

### LINKED ISSUE #6633: Bare Metal: Machine configs can be changed to existing machine configs
**Summary**
We block changing the machine config references for control planes and worker node groups on bare metal clusters. However, its possible to change references to existing machine configs. For example, I can change the control plane machine config reference to an existing worker node group machine config.

chrisdoherty4:
Closes #6633.

In bare metal clusters users shouldn't be able to change machine configs. However, EKS-A incorrectly allowed users to change the machine config references to existing machine configs.

### DIFF
--- a/pkg/providers/tinkerbell/upgrade.go
+++ b/pkg/providers/tinkerbell/upgrade.go
@@ -9,6 +9,7 @@ import (
 	rufiov1 "github.com/tinkerbell/rufio/api/v1alpha1"
 	tinkv1alpha1 "github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/cluster"
@@ -244,7 +245,22 @@ func (p *Provider) ValidateNewSpec(ctx context.Context, cluster *types.Cluster, 
 	oSpec := prevDatacenterConfig.Spec
 	nSpec := p.datacenterConfig.Spec
 
-	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
+	currCPMachineRef := prevSpec.Spec.ControlPlaneConfiguration.MachineGroupRef
+	desiCPMachineRef := clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef
+	if !currCPMachineRef.Equal(desiCPMachineRef) {
+		return errors.New("control plane machine config reference is immutable")
+	}
+
+	err = validateWorkerNodeGroupMachineConfigRefsUnchanged(prevSpec.Spec.WorkerNodeGroupConfigurations,
+		clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations)
+	if err != nil {
+		return err
+	}
+
+	// Validate worker node group machine configs are the same.
+	prevMachineConfigRefs := collection.ToMap(prevSpec.MachineConfigRefs(), func(v v1alpha1.Ref) string {
+		return v.Name
+	})
 
 	desiredWorkerNodeGroups := clusterSpec.Cluster.Spec.WorkerNodeGroupConfigurations
 	currentWorkerNodeGroups := collection.ToMap(prevSpec.Spec.WorkerNodeGroupConfigurations,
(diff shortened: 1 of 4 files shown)


pkg/providers/tinkerbell/upgrade.go
```diff
@@ -292,6 +307,41 @@ func (p *Provider) ValidateNewSpec(ctx context.Context, cluster *types.Cluster,
 	return nil
 }
 
+func validateControlPlaneMachineConfigRefUnchanged(current, desired *v1alpha1.Cluster) error {
```
jacobweinstock, line 310: I appreciate that this pattern of naming already exists, but it feels way too long. Thoughts on a shorter name? We could add a function code comment to help clarify it, if needed. Maybe something like,`func validateRef` or `func validateCPRef`, etc?

Also, could this function just take in the `v1alpha1.Ref` struct instead of the whole `v1alpha1.Cluster`?

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -244,7 +245,21 @@ func (p *Provider) ValidateNewSpec(ctx context.Context, cluster *types.Cluster,
 	oSpec := prevDatacenterConfig.Spec
 	nSpec := p.datacenterConfig.Spec
 
-	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
+	err = validateControlPlaneMachineConfigRefUnchanged(prevSpec, clusterSpec.Cluster)
```
jacobweinstock, line 248: what are you thoughts on the `if err := fn(); err != nil { return err }` pattern?

chrisdoherty4: I'd rather stick to 100 chars and extend vertically.

pkg/providers/tinkerbell/upgrade.go
chrisdoherty4: I honestly just prefer the clarity in this instance. If its not part of the function name, its part of a comment on the function that someone has to read anyway.

If we drop this to `validateCPRef(v1alpha1Ref)` then I'll just get rid of the function all together because it'll be superfluous.

chrisdoherty4: I just got rid of this func, it felt superfluous.

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -244,7 +245,22 @@ func (p *Provider) ValidateNewSpec(ctx context.Context, cluster *types.Cluster,
...
 	nSpec := p.datacenterConfig.Spec
 
-	prevMachineConfigRefs := machineRefSliceToMap(prevSpec.MachineConfigRefs())
+	currCPMachineRef := prevSpec.Spec.ControlPlaneConfiguration.MachineGroupRef
+	desiCPMachineRef := clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef
+	if !currCPMachineRef.Equal(desiCPMachineRef) {
+		return errors.New("control plane machine config reference is immutable")
+	}
+
+	err = validateWorkerNodeGroupMachineConfigRefsUnchanged(prevSpec.Spec.WorkerNodeGroupConfigurations,
```
jacobweinstock: As it's not clear to me, I would be interested in a code comment or similar to describe why we might want these to be unchanged.

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -292,6 +308,30 @@ func (p *Provider) ValidateNewSpec(ctx context.Context, cluster *types.Cluster,
 	return nil
 }
 
+func validateWorkerNodeGroupMachineConfigRefsUnchanged(current, desired []v1alpha1.WorkerNodeGroupConfiguration) error {
```
jacobweinstock: Not PR blocking; I want to voice my opinion again on the naming of this function. A shorter name like: `validateRefsUnchanged` provides just as much information/value to understanding this function as the current name but it's significantly shorter. This is valuable for a few reasons. It improves the readability, avoids repetition of input parameter values, it doesn't perpetuate the pattern of very long naming, it more closely adheres to the stdlib naming patterns, and promotes code comments for greater understanding, etc.

While not normative or canonical, this style guide from Google also describes this idea well: https://google.github.io/styleguide/go/best-practices#function-and-method-names

chrisdoherty4: This is probably fine to do now. With a previous iteration of the PR there was a collision between `validateControlPlaneMachineRefsUnchanged` and `validateWorkerNodeGroupMachineRefsUnchanged` (the Google styleguide points this out as an exception). 

Naming in bloated packages is always going to be difficult because there is so much in-scope. While I agree, I get dubious real fast in packages like `tinkerbell`.

pkg/providers/tinkerbell/upgrade.go
chrisdoherty4: If I knew concretely, I would love to add it, sadly I don't. This PR merely adds a missing piece of the logic to correctly enforce the constraint.

chrisdoherty4: If I knew concretely, I would love to add it, sadly I don't. This PR merely adds a missing piece of the logic to correctly enforce the constraint.

pkg/providers/tinkerbell/upgrade.go
chrisdoherty4: Updated

jacobweinstock (approved):
thanks for shortening names :-) I know naming in many instances is very much personal preference.

PR #6681, opened 2023-09-14
Add regional registries

d8660091:
*Issue #, if available:*

*Description of changes:* Add curated packages regional registries for beta stage.

*Testing (if applicable):* 
Manual:
1) Set EKSA_AWS_* credential environment variables which has access to regional curated package registry
2) install packagecontroller `eksctl-anywhere install packagecontroller`
3) verified the regional registry were used for defaultImageRegistry, defaultRegistry fields of PackageBundleController
4) verified the packages controller were running healthy, and all its dependencies are healthy 
5) manually installed hello package
6) verified the hello package was running healthy

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -40,6 +40,8 @@ const (
 
 type PackageControllerClientOpt func(client *PackageControllerClient)
 
+type registryAccessTester func(ctx context.Context, accessKey, secret, registry, region string) error
+
 type PackageControllerClient struct {
 	kubeConfig string
 	chart      *releasev1.Image
@@ -72,6 +74,9 @@ type PackageControllerClient struct {
 
 	// mu provides some thread-safety.
 	mu sync.Mutex
+
+	// registryAccessTester test if the aws credential has access to registry
+	registryAccessTester registryAccessTester
 }
 
 // ClientBuilder returns a k8s client for the specified cluster.
@@ -107,6 +112,7 @@ func NewPackageControllerClientFullLifecycle(logger logr.Logger, chartManager Ch
 		skipWaitForPackageBundle: true,
 		eksaRegion:               eksaDefaultRegion,
 		clientBuilder:            clientBuilder,
+		registryAccessTester:     TestRegistryAccess,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -260,6 +267,16 @@ func (pc *PackageControllerClient) GetCuratedPackagesRegistries() (sourceRegistr
...
 		}
+
+		regionalRegistry := GetRegionalRegistry(defaultRegistry, pc.eksaRegion)
+		if err := pc.registryAccessTester(ctx, pc.eksaAccessKeyID, pc.eksaSecretAccessKey, regionalRegistry, pc.eksaRegion); err == nil {
+			// use regional registry when the above credential is good
+			logger.V(6).Info("Using regional registry")
+			defaultRegistry = regionalRegistry
+			defaultImageRegistry = regionalRegistry
+		} else {
+			logger.V(6).Info("Using non-regional registry", "RegionalRegistryAccessIssue", err)
```
cxbrowne1207, line 278: What does it mean to use a non-regional registry?

Any reason we don't fail if they can't access a regional registry? Curious if we're going to keep or deprecate the fallback registry indefinitely.

```suggestion
			logger.V(6).Info("Using fallback registry", "Registry", defaultRegistry" "RegionalRegistryAccessIssue", err)
```

pkg/curatedpackages/regional_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	req.Header.Add("Authorization", "Basic "+authToken)
+
+	resp2, err := getResponse(req)
+	if err != nil {
+		return err
+	}
+
+	bodyBytes, err := io.ReadAll(resp2.Body)
+	if resp2.StatusCode != 200 && resp2.StatusCode != 404 {
```
cxbrowne1207, line 61: Why aren't we return an error if the StatusCode is 404 here?

d8660091: the function only tests if the customer has the IAM policy, 404 means the customer has it, so 404 is good here

pkg/curatedpackages/packagecontrollerclient.go
d8660091: There will be users still using the old registry, POC or old subscribers. The curated packages would still be installed in this case, because customers can set the AWS secrets after getting subscription.

pkg/curatedpackages/regional_registry.go
cxbrowne1207: Mind adding a comment explaining that? 404 typically means the server cannot find the requested resource, so this confused me a little.

cxbrowne1207 (approved):

PR #6703, opened 2023-09-19
Source Bottlerocket artifacts from new upload location

abhay-krishna:
We upload Bottlerocket OS images to an OS version folder and we need to source it from the same folder during dev and staging releases.


abhay-krishna:
/override eks-anywhere-cli-attribution-presubmit eks-anywhere-e2e-presubmit eks-anywhere-presubmit

PR #6709, opened 2023-09-21
Change timeout type from time.Duration to metav1.Duration

abhinavmpandey08:
*Description of changes:*
Change timeout type from time.Duration to metav1.Duration so users can specify timeout in `3s` format instead of seconds in nanoseconds.


abhay-krishna (approved):
/approve

PR #6721, opened 2023-09-22
Copy Rufio Machine v1alpha1 types from upstream:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This allows us to not have to import Rufio. Rufio uses controller-runtime v0.15.0 (and soon v0.16.2). EKS Anywhere use v0.14.2. Upgrading EKS Anywhere means all other dependent libraries will need upgraded too. This is not feasible at the moment from a time perspective, there are too many to update. For example: capv, capc, capd, abhay-krishna/cluster-api, aws/etcdadm-bootstrap-provider, etc.

*Testing (if applicable):*


pkg/api/v1alpha1/thirdparty/tinkerbell/rufio/machine.go
chrisdoherty4, line 1: More of an organizational thing, but I don't think we'd want third party APIs under the `api` pkg because this seems like an internal concern and not something we're interested in advertising outwardly.

chrisdoherty4: Ignore this. I didn't realize the thirdparty already exists.

chrisdoherty4 (approved):

Issue #5364, opened 2023-03-24
make unit-test should not run integration tests


jacobweinstock:
**What happened**:
The make target `unit-test` runs the integration test tool [envtest](https://book.kubebuilder.io/reference/envtest.html). `envtest` clearly says that they are an integration test framework. `make unit-test` should not use `envtest`. This will require refactoring of our Go unit tests.
Here are just 2 examples of us using `envtest` in unit tests. There are plenty more. 
https://github.com/aws/eks-anywhere/blob/1f2ebecec032881d78001e107e065cb140a7f652/controllers/main_test.go#L13
https://github.com/aws/eks-anywhere/blob/1f2ebecec032881d78001e107e065cb140a7f652/pkg/api/v1alpha1/nutanixdatacenterconfig_webhook_test.go#L122

CC @chrisdoherty4 , @TerryHowe 

**What you expected to happen**:
I expect that `make unit-test` only runs unit test. Ideally, I should be able to run `go test ./...` or `make unit-test` and have them function pretty much the same way. `go test ./...` should not be complaining with `Failed setting up envtest` errors.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release:
- EKS Distro Release:

chrisdoherty4:
I would be in favor of isolating `envtest` tests to a separate integration suite.

TerryHowe:
Since we have pretty comprehensive e2e tests, I don't think energy and time spent on envtest is well spent. It is better to have fast unit tests and good coverage and leave to e2e things that cannot be covered there. I removed envtest from curated packages because of the time and complexity it required.

chrisdoherty4:
I found recently that significant improvements have been made to the controller runtime fake client - it feels great in a unit test context.

PR #6723, opened 2023-09-22
Separate out integration tests:

jacobweinstock:
*Issue #, if available:*
#5364

*Description of changes:*
Without breaking any Make goals, this allows specifying an env var (`INTEGRATION_TESTS_ENABLED`) that will cause any test that uses [`envtest`](https://book.kubebuilder.io/reference/envtest.html) to either run or not. By default, `make unit-test` will run all tests that use `envtest`. i.e. `INTEGRATION_TESTS_ENABLED=true`. To disable tests that use `envtest`, set `INTEGRATION_TESTS_ENABLED=false`.

For example, with `make`, to disable `envtest` tests run:
`INTEGRATION_TESTS_ENABLED=false make unit-test`

When running `go` directly, the default behavior is to not run any `envtest` tests.
`go test ./...` is the same as `INTEGRATION_TESTS_ENABLED=false go test ./...`

To enable `envtest` with `go` directly, you'll need to start `envtest` and then run: `INTEGRATION_TESTS_ENABLED=true go test ./...`

*Testing (if applicable):*

### DIFF
--- a/controllers/cloudstack_datacenter_controller_test.go
+++ b/controllers/cloudstack_datacenter_controller_test.go
@@ -3,6 +3,7 @@ package controllers_test
 import (
 	"context"
 	"errors"
+	"os"
 	"testing"
 	"time"
 
@@ -23,6 +24,9 @@ import (
 )
 
 func TestCloudStackDatacenterReconcilerSetupWithManager(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	client := env.Client()
 	r := controllers.NewCloudStackDatacenterReconciler(client, nil)
 
@@ -31,6 +35,9 @@ func TestCloudStackDatacenterReconcilerSetupWithManager(t *testing.T) {
 }
 
 func TestCloudStackDatacenterReconcilerSuccess(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 
@@ -79,6 +86,9 @@ func TestCloudStackDatacenterReconcilerSuccess(t *testing.T) {
 }
 
 func TestCloudStackDatacenterReconcilerSetDefaultSuccess(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 
@@ -146,6 +156,9 @@ func TestCloudStackDatacenterReconcilerSetDefaultSuccess(t *testing.T) {
 }
 
 func TestCloudstackDatacenterConfigReconcilerDelete(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 
@@ -170,6 +183,9 @@ func TestCloudstackDatacenterConfigReconcilerDelete(t *testing.T) {
 }
 
 func TestCloudstackDatacenterConfigGetValidatorFailure(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 
@@ -198,6 +214,9 @@ func TestCloudstackDatacenterConfigGetValidatorFailure(t *testing.T) {
 }
 
 func TestCloudstackDatacenterConfigGetDatacenterFailure(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 	client := fake.NewClientBuilder().WithScheme(runtime.NewScheme()).Build()
@@ -217,6 +236,9 @@ func TestCloudstackDatacenterConfigGetDatacenterFailure(t *testing.T) {
 }
 
 func TestCloudstackDatacenterConfigGetExecConfigFailure(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 
@@ -243,6 +265,9 @@ func TestCloudstackDatacenterConfigGetExecConfigFailure(t *testing.T) {
 }
 
 func TestCloudstackDatacenterConfigAccountNotPresentFailure(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
+	}
 	g := NewWithT(t)
 	ctx := context.Background()
 	dcConfig := createCloudstackDatacenterConfig()
--- a/controllers/cloudstack_datacenter_controller_test.go
+++ b/controllers/cloudstack_datacenter_controller_test.go
@@ -3,7 +3,6 @@ package controllers_test
 import (
 	"context"
 	"errors"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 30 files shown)


controllers/main_test.go
```diff
@@ -7,8 +7,12 @@ import (
...
 
+const integrationTestEnvVar = "INTEGRATION_TESTS_ENABLED"
+
 var env *envtest.Environment
 
 func TestMain(m *testing.M) {
-	os.Exit(envtest.RunWithEnvironment(m, envtest.WithAssignment(&env)))
+	if os.Getenv(integrationTestEnvVar) == "true" {
+		os.Exit(envtest.RunWithEnvironment(m, envtest.WithAssignment(&env)))
+	}
```
g-gaston, line 17: Doesn't this skip all tests for the package? both the ones that use envtest and the ones that don't.

I could be wrong but I thought if you include a `TestMain` that executed always. And if you don't run `m`, then nothing runs.

g-gaston (commented):
Doing something like
```
if os.Getenv(integrationTestEnvVar) != "true" {
		s.Skipf("set env var '%v=true' to run this test", integrationTestEnvVar)
	}
```

adds extra work for engineers. It's not a lot but it's one more thing to remember. And the worse thing is, if they forget, they will just break `go test ./...` again (which is what we are trying to solve here) but they'll probably won't realize since they'll have `INTEGRATION_TESTS_ENABLED` enabled, which will just work.

Alternatively, what do you think about using go build tags? That would require moving all tests using envtest to a different file, like `*_integration_test.go`. This requires identifying all tests that need envtest, but it seems looking at this PR that you already did all the difficult work :)

The benefit of this is there is no need for extra code in the test functions, just putting them in the right file. And if you don't put them in the right file, the tests won't even compile since it won't have access to `var env *envtest.Environment`. TBH I also personally like the idea of having them in a different file if we are going to start treating them differently.

chrisdoherty4:
I'm indifferent to the approach we use as both have tradeoffs. If we shoot for the env variable approach though, I think it would be useful to add a helper so I don't need to include the boilerplate.

```
func TestFoo(t *testing.T) {
	test.MarkIntegration(t)
}
```

jacobweinstock:
> test.MarkIntegration(t)

Thanks for the feedback. The challenge i have with build tags is that they are easily discoverable. With this current pattern you get that discoverability. `go test` will clearly show the env var(s) that need to be set in order to run the integration tests. [Reference](https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html).

In regard to the extra work; using `envtest` is way more work than adding a one-liner `test.MarkIntegration(t)`. [Reference](https://github.com/aws/eks-anywhere/issues/5364#issuecomment-1483025481).

The breaking of `go test` if an engineer forgets to add `test.MarkIntegration(t)` is actually a good thing. With this pattern they will now see that wrote an integration test without marking it accordingly. Couple that with In the future we would have `make unit-test` only run unit tests and be a part of the github action.

chrisdoherty4 (approved):
This approach seems sound to me.

controllers/main_test.go
jacobweinstock: I think you're right, good catch. i've updated this to also be able to run without envtest.

controllers/cloudstack_datacenter_controller_test.go
```diff
@@ -16,13 +16,15 @@ import (
...
 	"github.com/aws/eks-anywhere/controllers"
+	"github.com/aws/eks-anywhere/internal/test"
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 )
 
 func TestCloudStackDatacenterReconcilerSetupWithManager(t *testing.T) {
+	test.MarkIntegration(t)
```
g-gaston, line 27: this made me think
What if we made engineers **have** to call this/some other method in order to use envtest?

Right now this method just skips the test if the env var is not set. But if the engineer doesn't add this line, the test will compile and try to run.

We could however "require" this method. Right now all tests using envtest need a package level variable called `env`, that is setup from `TestMain`. What if we set a pattern where instead of accessing this variable directly, we call a method to retrieve it? That way engineers will "need" to call this method if they want access to envtest. Of course, they could still look for the package level variable, but if we name it something like `internalEnvtest` and add a decent comment, I doubt this will happen by mistake.

Quick example of what I'm proposing:

```go
func TestCloudStackDatacenterReconcilerSetupWithManager(t *testing.T) {
  env := requireEnvtest(t)
```

The in this package `main_test.go`:
```go
// internalEnvtest allows to use envtest when integration
// tests are enabled.
// DO NOT use it directly, instead use requireEnvtest.
var internalEnvtest *envtest.Environment

func requireEnvtest(tb testing.TB) *envtest.Environment {
  test.MarkIntegration(t)
  return internalEnvtest
}
```

g-gaston:
> Thanks for the feedback, @g-gaston . The challenge i have with build tags is that they are not easily discoverable. With this current pattern you get that discoverability. `go test` will clearly show the env var(s) that need to be set in order to run the integration tests. [Reference](https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html).

I think this hints to our main discrepancy on this topic. Please correct me if I'm making the wrong assumption, but I feel you see `go test` as the top level entry point to our repo tests, while for me, `make` is such entry point.

From where I stand, `make help` (or `make [tab]`) is how we discover the different sets of tests one can run. And running those targets takes care of setting up the environment so they run properly. Instead of discovering "how to configure the tests", you just discover "the available tests".

Not trying to convince anyone here, just pointing where I think the disagreement comes from :)

I still prefer build tags, but if you are firm on using env vars + t.Skip, please take a look to my comment above about a possible iteration of your idea, that I think makes it more obvious/robust.
 
> In regard to the extra work; using `envtest` is way more work than adding a one-liner `test.MarkIntegration(t)`. [Reference](https://github.com/aws/eks-anywhere/issues/5364#issuecomment-1483025481).

There might be something I'm missing. Running tests with envtest should only require running the right `make` target. And if it's about writing tests with envtest, at most, is adding the a `main_test.go` file to the right folder. If we are talking about using envtest with `go test`, it only requires a one time setup and exporting an envvar.

I'm not sure what extra work envtest involves (the reference link seems to be just terry's opinion, which I appreciate but it's just one opinion) but I'm interested in hearing the issues you faced because I would like to fix them (orthogonally to this PR).

jacobweinstock:
> > Thanks for the feedback, @g-gaston . The challenge i have with build tags is that they are not easily discoverable. With this current pattern you get that discoverability. `go test` will clearly show the env var(s) that need to be set in order to run the integration tests. [Reference](https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html).
> 
> I think this hints to our main discrepancy on this topic. Please correct me if I'm making the wrong assumption, but I feel you see `go test` as the top level entry point to our repo tests, while for me, `make` is such entry point.
> 
> From where I stand, `make help` (or `make [tab]`) is how we discover the different sets of tests one can run. And running those targets takes care of setting up the environment so they run properly. Instead of discovering "how to configure the tests", you just discover "the available tests".
> 
> Not trying to convince anyone here, just pointing where I think the disagreement comes from :)
> 
> I still prefer build tags, but if you are firm on using env vars + t.Skip, please take a look to my comment above about a possible iteration of your idea, that I think makes it more obvious/robust.
> 
> > In regard to the extra work; using `envtest` is way more work than adding a one-liner `test.MarkIntegration(t)`. [Reference](https://github.com/aws/eks-anywhere/issues/5364#issuecomment-1483025481).
> 
> There might be something I'm missing. Running tests with envtest should only require running the right `make` target. And if it's about writing tests with envtest, at most, is adding the a `main_test.go` file to the right folder. If we are talking about using envtest with `go test`, it only requires a one time setup and exporting an envvar.
> 
> I'm not sure what extra work envtest involves (the reference link seems to be just terry's opinion, which I appreciate but it's just one opinion) but I'm interested in hearing the issues you faced because I would like to fix them (orthogonally to this PR).

Thanks for the reply @g-gaston. There is a lot going on in this PR but the main goal is to update `make unit-test` to only run unit tests. I think the only outstanding concern is whether to use an env var or build tags, no?

controllers/cloudstack_datacenter_controller_test.go
chrisdoherty4: I'm not sold on the value. `envtest` features in a subset of integration tests. Moreover, developers have the freedom to tweak how envtest gets configured dependent on their usecase.

g-gaston:
> @g-gaston Are we OK with this change?

TBH, I'm not sure the net change here is positive. But if you are both believe it is and are both aligned on the solution, let's try it and see what happens. Just make sure we send a message to the team informing of the new requirement, since this will affect everyone writing tests with envtest.

PR #6735, opened 2023-09-27
CLI changes for enabling modular upgrades for tinkerbell provider

rahulbabu95:
*Issue #, if available:*
Modular Upgrades are not enabled in Tinkerbell Provider currently because, we just have one OSImageUrl field that can be specified in data center config to be used by all machine configs in the cluster.

*Description of changes:*
Enable specifying OSImageUrl field per machine config to help facilitate modular upgrades. For backward compatibility continue supporting OSImageUrl field at data center config too and eventually deprecate. 

*Testing (if applicable):*
Manually tested upgrading CP and worker separately


pkg/providers/tinkerbell/assert.go
```diff
@@ -313,6 +334,26 @@ func (v *ValidatableTinkerbellCAPI) WorkerNodeHardwareGroups() []WorkerNodeHardw
...
 }
 
+// ClusterK8sVersion returns the Kubernetes version in major.minor format for a ValidatableTinkerbellCAPI.
+func (v *ValidatableTinkerbellCAPI) ClusterK8sVersion() *v1alpha1.KubernetesVersion {
+	return v.extractK8sVersion(v.KubeadmControlPlane.Spec.Version)
+}
+
+// WorkerNodeGroupK8sVersion returns each worker node group mapped to Kubernetes version in major.minor format for a ValidatableTinkerbellCAPI.
+func (v *ValidatableTinkerbellCAPI) WorkerNodeGroupK8sVersion() map[string]*v1alpha1.KubernetesVersion {
+	WNGK8sversion := make(map[string]*v1alpha1.KubernetesVersion)
```
chrisdoherty4, line 344: Normally you'd leave this lowercase. `wngK8sVersion`.

pkg/providers/tinkerbell/assert.go
```diff
@@ -313,6 +334,26 @@ func (v *ValidatableTinkerbellCAPI) WorkerNodeHardwareGroups() []WorkerNodeHardw
...
+	for _, wng := range v.WorkerGroups {
+		k8sVersion := v.extractK8sVersion(*wng.MachineDeployment.Spec.Template.Spec.Version)
+		WNGK8sversion[wng.MachineDeployment.Name] = k8sVersion
+	}
+	return WNGK8sversion
+}
+
+func (v *ValidatableTinkerbellCAPI) extractK8sVersion(k8sversion string) *v1alpha1.KubernetesVersion {
```
chrisdoherty4, line 352: What does this do? What's the expected format of `k8sVersion`?

chrisdoherty4, line 352:
```suggestion
func (v *ValidatableTinkerbellCAPI) extractK8sVersion(version string) *v1alpha1.KubernetesVersion {
```

pkg/providers/tinkerbell/assert.go
```diff
@@ -403,41 +444,64 @@ func ExtraHardwareAvailableAssertionForRollingUpgrade(catalogue *hardware.Catalo
...
-		)
-		if err != nil {
+
+		if err := ensureWorkerHardwareAvailability(spec, current, requirements, eksaVersionUpgrade); err != nil {
+			return err
+		}
+
+		if spec.HasExternalEtcd() {
+			return fmt.Errorf("external etcd upgrade is not supported")
+		}
```
chrisdoherty4, line 459: Wsa this check in this func before? If not I don't think its necessary, leave other validations to validate etcd is unspecified.

pkg/providers/tinkerbell/validate.go
```diff
@@ -29,12 +29,32 @@ func validateOsFamily(spec *ClusterSpec) error {
...
+	for _, mc := range machineConfigs {
+		if mc.Spec.OSImageURL != "" && dataCenterOSImageURLfound {
+			return fmt.Errorf("overlapping OSImageURL found, OSImageURL can either be set at datacenter config or for each machine config not both")
+		}
+		if mc.Spec.OSImageURL == "" && !dataCenterOSImageURLfound && mc.Spec.OSFamily != v1alpha1.Bottlerocket {
+			return fmt.Errorf("OSImageURL should be set for each machine config not found for: %s", mc.ObjectMeta.Name)
+		}
+	}
+	return nil
+}
```
chrisdoherty4, line 56: I think this can be combined into a single function: `validateOSImageURL`. The smell is the fact we pass a bool into `validateMachineCfgOSImageURL`. It suggests the 2 funcs are tied together yet we've split it up.

pkg/providers/tinkerbell/assert.go
rahulbabu95: Yes this case was already present in the function i had to just split up the function a bit as the complexity for over 10.!

pkg/providers/tinkerbell/assert.go
rahulbabu95: it is expected to be in the format we specify in our spec which is major.minor i.e. For ex. '1.xx'

pkg/providers/tinkerbell/upgrade.go
```diff
@@ -371,6 +371,39 @@ func (p *Provider) isScaleUpDown(oldCluster *v1alpha1.Cluster, newCluster *v1alp
 	return false
 }
 
+func (p *Provider) isEksaVersionUpgrade(currentSpec, newClusterSpec *cluster.Spec) bool {
+	return currentSpec.Bundles.Spec.Number != newClusterSpec.Bundles.Spec.Number
```
chrisdoherty4, line 375: I think this can be a variable at the site of use given its only used in 1 place. 

E.g. 

```
isEKSAVersionUpgrade := currentSpec.Bundles.Spec.Number != newClusterSpec.Bundles.Spec.Number
```

pkg/providers/tinkerbell/assert.go
```diff
@@ -313,6 +334,26 @@ func (v *ValidatableTinkerbellCAPI) WorkerNodeHardwareGroups() []WorkerNodeHardw
...
+func (v *ValidatableTinkerbellCAPI) WorkerNodeGroupK8sVersion() map[string]*v1alpha1.KubernetesVersion {
+	wngK8sversion := make(map[string]*v1alpha1.KubernetesVersion)
+	for _, wng := range v.WorkerGroups {
+		k8sVersion := v.toK8sVersion(*wng.MachineDeployment.Spec.Template.Spec.Version)
+		wngK8sversion[wng.MachineDeployment.Name] = k8sVersion
+	}
+	return wngK8sversion
+}
+
+func (v *ValidatableTinkerbellCAPI) toK8sVersion(k8sversion string) *v1alpha1.KubernetesVersion {
```
chrisdoherty4, line 352: Why are we using pointers for `KubernetesVersion`? From what I could tell all consumers dereference it anyway?

pkg/providers/tinkerbell/template.go
```diff
@@ -69,9 +69,16 @@ func (tb *TemplateBuilder) GenerateCAPISpecControlPlane(clusterSpec *cluster.Spe
 	if err != nil {
 		return nil, err
 	}
+	var OSImageURL string
+	if clusterSpec.TinkerbellDatacenter.Spec.OSImageURL != "" {
+		OSImageURL = clusterSpec.TinkerbellDatacenter.Spec.OSImageURL
+	} else {
+		OSImageURL = tb.controlPlaneMachineSpec.OSImageURL
+	}
```
chrisdoherty4, line 77: I think we should preference the `OSImageURL` specified on the machine config, not the datacenter config.

pkg/providers/tinkerbell/validate.go
```diff
@@ -29,12 +29,25 @@ func validateOsFamily(spec *ClusterSpec) error {
...
 		return fmt.Errorf("please use bottlerocket as osFamily for auto-importing or provide a valid osImageURL")
 	}
 	return nil
 }
 
+func validateOSImageURLDontOverlap(spec *ClusterSpec) error {
+	dcOSImageURL := spec.DatacenterConfig.Spec.OSImageURL
+	for _, mc := range spec.MachineConfigs {
+		if mc.Spec.OSImageURL != "" && dcOSImageURL != "" {
+			return fmt.Errorf("overlapping OSImageURL found, OSImageURL can either be set at datacenter config or for each machine config not both")
```
chrisdoherty4, line 42:
```suggestion
			return errors.New("cannot specify OSImageURL on both TinkerbellMachineConfig's and TinkerbellDatacenterConfig")
```

pkg/providers/tinkerbell/validate.go
```diff
@@ -29,12 +29,25 @@ func validateOsFamily(spec *ClusterSpec) error {
...
+		}
+		if mc.Spec.OSImageURL == "" && dcOSImageURL == "" && mc.Spec.OSFamily != v1alpha1.Bottlerocket {
+			return fmt.Errorf("OSImageURL should be set for each machine config not found for: %s", mc.ObjectMeta.Name)
```
chrisdoherty4, line 45:
```suggestion
			return fmt.Errorf("missing OSImageURL on TinkerbellMachineConfig '%s'", mc.ObjectMeta.Name)
```

pkg/providers/tinkerbell/validate.go
```diff
@@ -29,12 +29,25 @@ func validateOsFamily(spec *ClusterSpec) error {
...
 	}
 
-	if controlPlaneOsFamily != v1alpha1.Bottlerocket && spec.DatacenterConfig.Spec.OSImageURL == "" {
+	if controlPlaneOsFamily != v1alpha1.Bottlerocket && spec.DatacenterConfig.Spec.OSImageURL == "" && spec.ControlPlaneMachineConfig().Spec.OSImageURL == "" {
 		return fmt.Errorf("please use bottlerocket as osFamily for auto-importing or provide a valid osImageURL")
 	}
 	return nil
 }
 
+func validateOSImageURLDontOverlap(spec *ClusterSpec) error {
```
chrisdoherty4, line 38:
```suggestion
func validateOSImageURL(spec *ClusterSpec) error {
```

pkg/providers/tinkerbell/template.go
rahulbabu95: there's no precedence here actually as we don't allow specifying at both levels. so here we just set the value for generating template based on which level the value is set and it would never be the case that if machine config imageURL is present DC could take preference (as we error out early).!

chrisdoherty4 (approved):

PR #6741, opened 2023-09-28
Rufio options

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This incorporates Rufio options per machine. The only options currently supported are for the Rufio/bmclib RPC provider. All options are configurable via env vars and cli flags.

*Testing (if applicable):*


pkg/dependencies/factory.go
```diff
@@ -381,8 +382,20 @@ func (f *Factory) WithExecutableBuilder() *Factory {
 	return f
 }
 
+// ProviderOptions contains per provider options.
+type ProviderOptions struct {
```
jacobweinstock, line 386: this allows us to add more options in the future without continually breaking the `WithProvider` function signature.

chrisdoherty4:
@jacobweinstock Is this awaiting the rebase since the other PR merged?

jacobweinstock:
> @jacobweinstock Is this awaiting the rebase since the other PR merged?

Sorry, yes it is. Just rebased. Thanks!

pkg/providers/tinkerbell/hardware/catalogue_bmc.go
```diff
@@ -87,21 +90,147 @@ func toRufioMachine(m Machine) *v1alpha1.Machine {
 	// TODO(chrisdoherty4)
```
chrisdoherty4, line 90: Oops 🤦🏻

pkg/providers/tinkerbell/hardware/csv.go
```diff
@@ -18,11 +18,13 @@ import (
...
 type CSVReader struct {
 	reader *csv.Unmarshaller
+	// BMCOptions used in a Machine that do not have a corresponding column in the CSV.
+	BMCOptions *BMCOptions
 }
 
 // NewCSVReader returns a new CSVReader instance that consumes csv data from r. r should return io.EOF when no more
 // records are available.
-func NewCSVReader(r io.Reader) (CSVReader, error) {
+func NewCSVReader(r io.Reader, opts *BMCOptions) (CSVReader, error) {
```
chrisdoherty4, line 27: Are the BMC options not per hardware?

pkg/providers/tinkerbell/hardware/machine.go
```diff
@@ -27,6 +28,74 @@ type Machine struct {
 	BMCUsername  string `csv:"bmc_username, omitempty"`
 	BMCPassword  string `csv:"bmc_password, omitempty"`
 	VLANID       string `csv:"vlan_id, omitempty"`
+
+	// BMCOptions are the options used for Rufio providers.
+	BMCOptions *BMCOptions `csv:"-"`
```
chrisdoherty4, line 33: How does a user specify the opts?

chrisdoherty4:
I left some inline comments but broadly I'm not clear how this is meant to work. Do we expect all hardware to use the same settings? What happens if we want to plumb through different settings per hardware as supported by Rufio?

jacobweinstock:
> I left some inline comments but broadly I'm not clear how this is meant to work. Do we expect all hardware to use the same settings? What happens if we want to plumb through different settings per hardware as supported by Rufio?

Thanks. 
Do we expect all hardware to use the same settings?  Yes.
What happens if we want to plumb through different settings per hardware as supported by Rufio? Not supported right now. Don't have a use case for that yet either.

pkg/providers/tinkerbell/hardware/machine.go
jacobweinstock: Correct, you cannot set these via the hardware csv. They are all set via env vars only.

chrisdoherty4:
Understood. I don't think there's anything in the 'interface' to configuring this that would prevent future expansion: does that sound accurate? Specifically, we're not adding anything to the hardware CSV and we're just offering a bunch of CLI/ENV options for configuring a single rpc endpoint - does that sound right?

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -50,14 +61,42 @@ func init() {
...
 	createClusterCmd.Flags().StringArrayVar(&cc.skipValidations, "skip-validations", []string{}, fmt.Sprintf("Bypass create validations by name. Valid arguments you can pass are --skip-validations=%s", strings.Join(createvalidations.SkippableValidations[:], ",")))
+	tinkerbellFlags(createClusterCmd.Flags(), cc.providerOptions.Tinkerbell.BMCOptions.RPC)
+
+	aflag.MarkRequired(createClusterCmd.Flags(), aflag.ClusterConfig.Name)
+}
 
-	flags.MarkRequired(createClusterCmd.Flags(), flags.ClusterConfig.Name)
+func tinkerbellFlags(fs *pflag.FlagSet, r *hardware.RPCOpts) {
+	aflag.String(aflag.TinkerbellBMCConsumerURL, &r.ConsumerURL, fs)
+	aflag.MarkHidden(fs, aflag.TinkerbellBMCConsumerURL.Name)
```
chrisdoherty4, line 77: What's the reason for hiding these?

Also curious if we thought about including these in the cluster config somehow? Historically there's been a strong preference for using the cluster config and this seems like something that could feature on the datacenter config?

chrisdoherty4 (approved):
LGTM.

The inline comment on cluster config vs use of CLI/env vars should probably be circulated some.

cmd/eksctl-anywhere/cmd/createcluster.go
jacobweinstock: _What's the reason for hiding these?_  These are what i would consider advanced options and not something I would recommend we clutter a cli help message with.

_Curious if we thought about including these in the cluster config somehow?_ Yeah, i did consider it. The concern i have for adding to the cluster config is that i believe we don't understand enough about how/if/when customers will use this functionality. I would prefer to hide it like i did and learn more before adding it to the cluster config where deprecating feels a lot more difficult.

chrisdoherty4: Great points, agreed.

PR #6752, opened 2023-09-29
Add wait in between the retries while waiting for failure message to …

panktishah26:
*Description of changes:*

*Testing (if applicable):*
Add wait in between the retries while waiting for failure message to be empty for upgrade management task. Add support for notimeout flag for failure message check.

Currently, the retry doesn't wait enough for machineConfig template to be updated for the management cluster upgrade task so the upgrade process fails. Increased the time out for the retry function by adding wait in between the retry so it doesn't throttle the API server with the requests. 

### DIFF
--- a/pkg/clustermanager/applier.go
+++ b/pkg/clustermanager/applier.go
@@ -17,10 +17,11 @@ import (
 )
 
 const (
-	applyClusterSpecTimeout        = 2 * time.Minute
-	waitForClusterReconcileTimeout = time.Hour
-	retryBackOff                   = time.Second
-	defaultFieldManager            = "eks-a-cli"
+	applyClusterSpecTimeout           = 2 * time.Minute
+	waitForClusterReconcileTimeout    = time.Hour
+	retryBackOff                      = time.Second
+	waitForFailureMessageErrorTimeout = 20 * time.Second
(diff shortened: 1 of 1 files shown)


pkg/clustermanager/applier.go
```diff
@@ -173,3 +176,10 @@ func (a Applier) retrierForWait(waitStartTime time.Time) *retrier.Retrier {
 		retrier.WithRetryPolicy(retrier.BackOffPolicy(a.retryBackOff)),
 	)
 }
+
+func (a Applier) retrierForFailureMessage() *retrier.Retrier {
+	return retrier.New(
+		a.waitForFailureMessage,
+		retrier.WithRetryPolicy(retrier.BackOffPolicy(5*retryBackOff)),
```
g-gaston, line 183:
```suggestion
		retrier.WithRetryPolicy(retrier.BackOffPolicy(a.retryBackOff)),
```

panktishah26, line 183: `retryBackOff` wait time is of 1 second. To avoid hitting rate limiting api server I have kept the wait time of 5 seconds between each retry. We can change it if needed.

g-gaston (commented):
/lgtm

PR #6756, opened 2023-10-02
Add extra logs and increase verbosity of upgrade command for e2e tests

panktishah26:
*Description of changes:*
Increase the verbosity of upgrade command for e2e tests and added logs for debugging e2e tests failures.

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/wait.go
+++ b/pkg/cluster/wait.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/cluster-api/util/conditions"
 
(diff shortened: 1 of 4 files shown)


pkg/cluster/wait.go
```diff
@@ -51,6 +52,10 @@ func WaitFor(ctx context.Context, client kubernetes.Reader, cluster *anywherev1.
 
 		observedGeneration := c.Status.ObservedGeneration
 		generation := c.Generation
+
+		log := logger.Get()
```
g-gaston, line 56: just inject the logger in the method if we need it, please :)
alternatively, you can pass it in the context

pkg/cluster/wait.go
```diff
@@ -51,6 +52,10 @@ func WaitFor(ctx context.Context, client kubernetes.Reader, cluster *anywherev1.
...
+		log.V(9).Info("cluster generation is %s and observedGeneration is %s", generation, observedGeneration)
```
g-gaston, line 57: Does this work? this is a strctured logger, it takes key values, not a formatting sequence

pkg/clustermanager/applier.go
```diff
@@ -95,7 +95,7 @@ func WithApplierRetryBackOff(backOff time.Duration) ApplierOpt {
 // until the changes are fully reconciled.
 func (a Applier) Run(ctx context.Context, spec *cluster.Spec, managementCluster types.Cluster) error {
 	var client kubernetes.Client
-
+	a.log.V(9).Info("cluster generation is %s before applying specs", spec.Cluster.Generation)
```
g-gaston, line 98: same here about logger and key values

pkg/cluster/wait.go
panktishah26: Done. I did not inject it earlier because I was going to remove it after we resolve the issue.

pkg/cluster/wait.go
panktishah26: sorry I did not test this and completely missed this.

g-gaston (approved):

PR #6761, opened 2023-10-03
Refactor registry tester to support aws config

d8660091:
*Issue #, if available:*

*Description of changes:* When a user provides key/secret that has access to regional ECR, the package controller would automatically switch to the regional ECR. However when a user provides an AWS credential through AWS config file that has access to regional ECR, the curated packages controller would not use region ECR endpoint. This PR fixes this issue.

*Testing (if applicable):* Manuall invoked `eksa install packagecontroller` with config environment variable, and verified the config has been stored, and regional ECR endpoint is used.

### DIFF
--- a/pkg/curatedpackages/packagecontrollerclient.go
+++ b/pkg/curatedpackages/packagecontrollerclient.go
@@ -40,8 +40,6 @@ const (
 
 type PackageControllerClientOpt func(client *PackageControllerClient)
 
-type registryAccessTester func(ctx context.Context, accessKey, secret, registry, region string) error
-
 type PackageControllerClient struct {
 	kubeConfig string
 	chart      *releasev1.Image
@@ -76,7 +74,7 @@ type PackageControllerClient struct {
 	mu sync.Mutex
 
 	// registryAccessTester test if the aws credential has access to registry
-	registryAccessTester registryAccessTester
+	registryAccessTester RegistryAccessTester
 }
 
 // ClientBuilder returns a k8s client for the specified cluster.
@@ -112,7 +110,7 @@ func NewPackageControllerClientFullLifecycle(logger logr.Logger, chartManager Ch
 		skipWaitForPackageBundle: true,
 		eksaRegion:               eksaDefaultRegion,
 		clientBuilder:            clientBuilder,
-		registryAccessTester:     TestRegistryAccess,
+		registryAccessTester:     &DefaultRegistryAccessTester{},
 	}
 }
 
@@ -171,7 +169,7 @@ func NewPackageControllerClient(chartManager ChartManager, kubectl KubectlRunner
 		kubectl:              kubectl,
 		registryMirror:       registryMirror,
 		eksaRegion:           eksaDefaultRegion,
-		registryAccessTester: TestRegistryAccess,
+		registryAccessTester: &DefaultRegistryAccessTester{},
 	}
 
 	for _, o := range options {
@@ -269,7 +267,7 @@ func (pc *PackageControllerClient) GetCuratedPackagesRegistries(ctx context.Cont
 		}
 
 		regionalRegistry := GetRegionalRegistry(defaultRegistry, pc.eksaRegion)
-		if err := pc.registryAccessTester(ctx, pc.eksaAccessKeyID, pc.eksaSecretAccessKey, regionalRegistry, pc.eksaRegion); err == nil {
+		if err := pc.registryAccessTester.Test(ctx, pc.eksaAccessKeyID, pc.eksaSecretAccessKey, pc.eksaRegion, pc.eksaAwsConfig, regionalRegistry); err == nil {
 			// use regional registry when the above credential is good
 			logger.V(6).Info("Using regional registry")
 			defaultRegistry = regionalRegistry
@@ -619,7 +617,7 @@ func WithClusterSpec(clusterSpec *cluster.Spec) func(client *PackageControllerCl
 }
 
 // WithRegistryAccessTester sets the registryTester.
-func WithRegistryAccessTester(registryTester registryAccessTester) func(client *PackageControllerClient) {
+func WithRegistryAccessTester(registryTester RegistryAccessTester) func(client *PackageControllerClient) {
 	return func(config *PackageControllerClient) {
 		config.registryAccessTester = registryTester
 	}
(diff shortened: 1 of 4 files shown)


pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -619,7 +617,7 @@ func WithClusterSpec(clusterSpec *cluster.Spec) func(client *PackageControllerCl
 }
 
 // WithRegistryAccessTester sets the registryTester.
-func WithRegistryAccessTester(registryTester registryAccessTester) func(client *PackageControllerClient) {
+func WithRegistryAccessTester(registryTester RegistryAccessTester) func(client *PackageControllerClient) {
```
jonathanmeier5, line 620: Just confirming, this is only used in tests?

pkg/curatedpackages/regional_registry.go
```diff
@@ -22,28 +24,33 @@ var prodRegionalECRMap = map[string]string{
...
+// DefaultRegistryAccessTester the default implementation of RegistryAccessTester.
+type DefaultRegistryAccessTester struct{}
+
+// Test if the AWS static credential or sharedConfig has valid permission to access an ECR registry.
+func (r *DefaultRegistryAccessTester) Test(ctx context.Context, accessKey, secret, region, awsConfig, registry string) (err error) {
+	authTokenProvider := &DefaultRegistryAuthTokenProvider{}
+
+	var authToken string
+	if len(awsConfig) > 0 {
+		authToken, err = authTokenProvider.GetTokenByAWSConfig(ctx, awsConfig)
```
jonathanmeier5, line 41: Do we need to write the order of precedence here in our docs? Or have we already established that somewhere? Not a blocker, maybe just something to write a ticket for.

pkg/curatedpackages/regional_registry.go
```diff
@@ -22,28 +24,33 @@ var prodRegionalECRMap = map[string]string{
...
 	}
-	authToken := out.AuthorizationData[0].AuthorizationToken
 
-	return TestRegistryWithAuthToken(*authToken, registry, http.DefaultClient.Do)
+	return TestRegistryWithAuthToken(authToken, registry, http.DefaultClient.Do)
 }
 
 // TestRegistryWithAuthToken test if the registry can be acccessed with auth token.
-func TestRegistryWithAuthToken(authToken, registry string, getResponse func(req *http.Request) (*http.Response, error)) error {
+func TestRegistryWithAuthToken(authToken, registry string, do Do) error {
```
jonathanmeier5, line 53: Is "Do" a golang convention or just a convenient interface name?

pkg/curatedpackages/regional_registry.go
```diff
@@ -22,28 +24,33 @@ var prodRegionalECRMap = map[string]string{
...
+}
 
-	ecrClient := ecr.NewFromConfig(cfg)
-	out, err := ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
+// DefaultRegistryAccessTester the default implementation of RegistryAccessTester.
+type DefaultRegistryAccessTester struct{}
+
+// Test if the AWS static credential or sharedConfig has valid permission to access an ECR registry.
+func (r *DefaultRegistryAccessTester) Test(ctx context.Context, accessKey, secret, region, awsConfig, registry string) (err error) {
+	authTokenProvider := &DefaultRegistryAuthTokenProvider{}
```
jonathanmeier5, line 37: what do you think about making this part of the struct? then it's a little easier to do dependency injection for testing?

pkg/curatedpackages/regional_registry_test.go
```diff
@@ -2,41 +2,126 @@ package curatedpackages_test
...
+}
+
+func (m *mockRegistryAuthTokenProvider) GetTokenByAWSKeySecret(ctx context.Context, key, secret, region string) (string, error) {
+	if key == goodAwsKey && secret == goodAwsSecret {
+		return goodAuthToken, nil
 	}
+	return badAuthToken, nil
+}
+
+func TestTestAWSConfigRegistryAccessWithAWSConfig(t *testing.T) {
```
jonathanmeier5, line 73: I don't love the TestTest stutter here, but we've already gone with this as a naming convention so I don't think it's worth refactoring the whole thing. Are these types of Tester objects common in golang? In general it feels funny to me to have Test in a resource name that is not a test.

jonathanmeier5 (approved):
/lgtm

pkg/curatedpackages/regional_registry.go
d8660091: This order is somehow determined by credentialproviderpackages https://github.com/aws/eks-anywhere-packages/blob/main/credentialproviderpackage/pkg/awscred/awscred.go#L48C25-L48C25

pkg/curatedpackages/regional_registry.go
d8660091: it's a method of http.Client, but I used it here for alias of the function type

pkg/curatedpackages/regional_registry.go
```diff
@@ -76,3 +83,80 @@ func GetRegionalRegistry(defaultRegistry, region string) string {
...
+
+// Do is a function type that takes a http request and returns a http response.
+type Do func(req *http.Request) (*http.Response, error)
+
+// TestRegistryAccessWithAWSConfig test if the AWS config has valid permission to access container registry.
+func TestRegistryAccessWithAWSConfig(ctx context.Context, awsConfig, registry string, tokenProvider RegistryAuthTokenProvider, do Do) error {
+	token, err := tokenProvider.GetTokenByAWSConfig(ctx, awsConfig)
+	if err != nil {
+		return err
+	}
```
jonathanmeier5, line 151: Actually, I missed out codecov was quite low. Could we add some tests to get up to 80% coverage?

jonathanmeier5 (changes requested):

pkg/curatedpackages/regional_registry.go
d8660091: The disadvantage is that the field has to be specified clearly when creating the struct. Maybe the advantage doesn't worth the effort

pkg/curatedpackages/regional_registry_test.go
d8660091: Yeah, the name is funny when you point it out. I don't think it's very common wildly.

jonathanmeier5 (approved):
/lgtm

PR #6771, opened 2023-10-05
Allow to configure gitOps during kindless management cluster

panktishah26:
*Description of changes:*
We don't allow to configure gitOps for management cluster upgrade. For kindless management upgrade flow, we use controller to upgrade our cluster so this PR allows configuring gitOps for the management cluster if the kindless feature is enabled.

*Testing (if applicable):*


g-gaston (commented):
/lgtm

PR #6781, opened 2023-10-06
Verify the condition matches certain amount of times kindless upgrade

panktishah26:
*Description of changes:*
With the new kindless upgrade management task we have seen an error message like `kubernetes version v1.25.13-eks-a6e69b0 does not match expected version 1.26`. CLI relies on cluster status object field to decide on cluster upgrade process. This error happens due to cluster status field not being upgraded immediately due to certain race condition or just due to the lag. To fix these issues we have added a check to verify the conditions certain amount of times before moving to the next steps.

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/wait.go
+++ b/pkg/cluster/wait.go
@@ -17,8 +17,9 @@ import (
 // WaitForCondition blocks until either the cluster has this condition as True
 // or the retrier timeouts. If observedGeneration is not equal to generation,
 // the condition is considered false regardless of the status value.
-func WaitForCondition(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, retrier *retrier.Retrier, conditionType anywherev1.ConditionType) error {
-	return WaitFor(ctx, log, client, cluster, retrier, func(c *anywherev1.Cluster) error {
+// total field is to check the total number of times the given condition is met for consistency.
+func WaitForCondition(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, total int, retrier *retrier.Retrier, conditionType anywherev1.ConditionType) error {
+	return WaitFor(ctx, log, client, cluster, total, retrier, func(c *anywherev1.Cluster) error {
 		condition := conditions.Get(c, conditionType)
 		if condition == nil {
 			return fmt.Errorf("cluster doesn't yet have condition %s", conditionType)
@@ -37,7 +38,8 @@ type Matcher func(*anywherev1.Cluster) error
 // WaitFor gets the cluster object from the client
 // checks for generation and observedGeneration condition
 // matches condition and returns error if the condition is not met.
-func WaitFor(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, retrier *retrier.Retrier, matcher Matcher) error {
+func WaitFor(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, total int, retrier *retrier.Retrier, matcher Matcher) error {
+	count := 0
 	return retrier.Retry(func() error {
 		c := &anywherev1.Cluster{}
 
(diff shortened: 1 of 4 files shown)


pkg/cluster/wait.go
```diff
@@ -59,6 +60,15 @@ func WaitFor(ctx context.Context, log logr.Logger, client kubernetes.Reader, clu
...
 
-		return matcher(c)
+		if err := matcher(c); err != nil {
+			count = 0
+			return err
+		}
+
+		if count < total {
+			log.V(9).Info("Cluster has reached to expected condition in %d/%d times", count, total)
+		}
```
g-gaston, line 70: This part is wrong, right?
You are never increasing the count
And also you are never return an error if the count happens to be smaller than total

pkg/cluster/wait.go
```diff
@@ -17,8 +17,8 @@ import (
 // WaitForCondition blocks until either the cluster has this condition as True
 // or the retrier timeouts. If observedGeneration is not equal to generation,
 // the condition is considered false regardless of the status value.
-func WaitForCondition(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, retrier *retrier.Retrier, conditionType anywherev1.ConditionType) error {
-	return WaitFor(ctx, log, client, cluster, retrier, func(c *anywherev1.Cluster) error {
+func WaitForCondition(ctx context.Context, log logr.Logger, client kubernetes.Reader, cluster *anywherev1.Cluster, total int, retrier *retrier.Retrier, conditionType anywherev1.ConditionType) error {
```
g-gaston, line 20: Can we document what total does? For both methods pls

pkg/cluster/wait.go
g-gaston: We are missing a test here, you would have seen that this doesn't work if you wrote one :)

panktishah26: Updated the code here. Thanks for pointing that out. I have added unit test too.

pkg/clustermanager/applier.go
```diff
@@ -20,10 +20,12 @@ const (
 	applyClusterSpecTimeout           = 2 * time.Minute
 	waitForClusterReconcileTimeout    = time.Hour
 	retryBackOff                      = time.Second
-	waitForFailureMessageErrorTimeout = 20 * time.Second
+	waitForFailureMessageErrorTimeout = time.Minute
 	defaultFieldManager               = "eks-a-cli"
 )
 
+var defaultConditionCheckTotalCount = 20
```
panktishah26, line 27: I wanted to make this configurable for unit test.

pkg/clustermanager/applier.go
```diff
@@ -20,10 +20,12 @@ const (
 	applyClusterSpecTimeout           = 2 * time.Minute
 	waitForClusterReconcileTimeout    = time.Hour
 	retryBackOff                      = time.Second
-	waitForFailureMessageErrorTimeout = 20 * time.Second
+	waitForFailureMessageErrorTimeout = time.Minute
```
panktishah26, line 23: I realized we would keep `defaultConditionCheckTotalCount` to at least one third of the wait time. Thus, I have changed `waitForFailureMessageErrorTimeout` to 1 minute.
The main reason is, we make count to 0 inside [waitFor](https://github.com/aws/eks-anywhere/pull/6781/files#diff-29080fc751d87a584f5bf80922c39b3467a34639557e8bdcf1e574a4ae719273R65) method if the condition doesn't match.
In case where the condition doesn't get match for some time and it matches at 15 seconds per say it needs 20 seconds more to check for the condition consistency.

pkg/clustermanager/applier.go
g-gaston: Can you make this a const? You can't update this from unit tests either since it's in a diff package

g-gaston: If you want to make it configurable, pls add another member to the applier struct like with the other params

Issue #6767, opened 2023-10-04
Add backup and restore strategy of failed cluster in doc site


jiayiwang7:
In certain unfortunate circumstances, an EKS Anywhere cluster may find itself in an unrecoverable state due to various factors such as a failed cluster upgrade, underlying infrastructure problems, or network issues, rendering the cluster inaccessible through conventional means. We need documents that outline detailed steps to guide user through the process of restoring a failed cluster from backups in these critical situations.

sharmadinesha:
Hello Joey, we do see this issue created a month back, seems we are also stuck in a similar situation , where we wish to take a backup of a bottlerocket nodes and restore it say in case when the nodes are not reachable nor the cluster.  Is there any update on the documentation for the same !

jiayiwang7:
> Hello Joey, we do see this issue created a month back, seems we are also stuck in a similar situation , where we wish to take a backup of a bottlerocket nodes and restore it say in case when the nodes are not reachable nor the cluster. Is there any update on the documentation for the same !

Hello the doc site is updated with instruction. Please take a look and give it a try: https://anywhere.eks.amazonaws.com/docs/clustermgmt/cluster-backup-restore/

PR #6783, opened 2023-10-06
Add instruction of workload cluster backup and restore

jiayiwang7:
*Issue #, if available:*

Part2 of #6767 

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
+++ b/docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
@@ -161,4 +161,177 @@ If the cluster is no longer accessible in any means, or the infrastructure machi
 
 ## Restore a workload cluster
 
-Restoring a workload cluster is a delicate process. If you have an [EKS Anywhere Enterprise Subscription](https://aws.amazon.com/eks/eks-anywhere/pricing/), please contact AWS support team if you wish to perform such operation.
+### Cluster accessible and the infrastructure components not changed after etcd backup was taken
+
+Similar to the failed management cluster without infrastructure components change situation, follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the workload cluster itself from the backup.
+
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+If the workload cluster is still accessible, but the infrastructure machines are changed after the etcd backup was taken, you can still try restoring the cluster itself from the etcd backup. Although doing so is risky: it can potentially cause the node names, IPs and other infrastructure configurations to revert to a state that is no longer valid. Restoring etcd effectively takes a cluster back in time and all clients will experience a conflicting, parallel history. This can impact the behavior of watching components like Kubernetes controller managers, EKS Anywhere cluster controller manager, and Cluster API controller managers.
+
+If the original workload cluster becomes inaccessible or cannot be restored to a healthy state from an outdated etcd, a new workload cluster needs to be created. This new cluster should be managed by the same management cluster that oversaw the original. You must then restore your workload applications to this new cluster from the etcd backup of the original. This ensures the management cluster retains control, with all data from the old cluster intact. Below is an example of applying the etcd `backup etcd-snapshot-w01.db` from a failed workload cluster `w01` to a new cluster `w02`:
+
+1. Create a new workload cluster to which you will be migrating your workloads and applications from the original failed workload cluster.
+
+    You can define a cluster config similar to your old workload cluster, with a different cluster name (if the old workload cluster still exists), and run cluster creation of the new workload cluster with the **exact same EKS Anywhere version** used to create the old workload cluster.
+
+    ```bash
+    export MGMT_CLUSTER="mgmt"
+    export MGMT_CLUSTER_KUBECONFIG=${MGMT_CLUSTER}/${MGMT_CLUSTER}-eks-a-cluster.kubeconfig
+
+    eksctl anywhere create cluster -f w02.yaml --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+    ```
+
+1. Save the config map objects of the new workload cluster to a file.
+
+    Save a copy of the new workload cluster's `cluster-info`, `kube-proxy` and `kubeadm-config` config map objects before the restore. This is necessary as the etcd restore will override the config maps above with the metadata information (certificates, endpoint, etc.) from the old cluster.
+
+    ```bash
+    export WORKLOAD_CLUSTER_NAME="w02"
+    export WORKLOAD_CLUSTER_KUBECONFIG=${WORKLOAD_CLUSTER_NAME}/${WORKLOAD_CLUSTER_NAME}-eks-a-cluster.kubeconfig
+
+    cat <<EOF >> w02-cm.yaml
+    $(kubectl get -n kube-public cm cluster-info -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    ---
+    $(kubectl get -n kube-system cm kube-proxy -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    ---
+    $(kubectl get -n kube-system cm kubeadm-config -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    EOF
+    ```
+
+    Manually remove the `creationTimestamp`, `resourceVersion`, `uid` from the config map objects, so that later you can run `kubectl apply` against this file without errors.
+
+
+1. Follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the old workload cluster's etcd backup `etcd-snapshot-w01.db` onto the new workload cluster `w02`. Use different restore process based on OS family:
+    * [BottleRocket]({{< relref "../etcd-backup-restore/bottlerocket-etcd-backup/#restore-etcd-from-backup" >}})
+    * [Ubuntu]({{< relref "../etcd-backup-restore/ubuntu-rhel-etcd-backup/#restore" >}})
+
+    {{% alert title="Warning" color="warning" %}}
+
+    Do not unpause the cluster reconcilers as the workload cluster is not completely setup with required configurations yet. Unpausing the cluster reconcilers might cause unexpected machine rollout and leave the cluster in unhealthy state.
+
+    {{% /alert %}}
+
+    You might notice that after restoring the original etcd backup to the new workload cluster `w02`, all the nodes go to `NotReady` state with node names changed to have prefix `w01-*`. This is because restoring etcd effectively applies the node data from the original cluster which causes a conflicting history and can impact the behavior of watching components like Kubelets, Kubernetes controller managers.
+
+    ```bash
+    kubectl get nodes --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+
+    NAME                              STATUS     ROLES           AGE     VERSION
+    w01-bbtdd                         NotReady   control-plane   3d23h   v1.27.3-eks-6f07bbc
+    w01-md-0-66dbcfb56cxng8lc-8ppv5   NotReady   <none>          3d23h   v1.27.3-eks-6f07bbc
+    ```
+
+1. Restart Kubelet of the control plane and worker nodes of the new workload cluster after the restore.
+
+    For some cases, Kubelet on the node will automatically restart and nodes becomes ready. For other cases, you need to manually restart the Kubelet on all the control plane and worker nodes in order to bring back the nodes to ready state. Kubelet registers the node itself with the apisever which then updates etcd with the correct node data of the new workload cluster `w02`.
+
+    {{< tabpane >}}
+    {{< tab header="Ubuntu or RHEL" lang="bash" >}}
+# SSH into the control plane and worker nodes. You must do this for each node.
+ssh -i ${SSH_KEY} ${SSH_USERNAME}@<node IP>
+sudo su
+systemctl restart kubelet
+    {{< /tab >}}
+    {{< tab header="Bottlerocket" lang="bash" >}}
+# SSH into the control plane and worker nodes. You must do this for each node.
+ssh -i ${SSH_KEY} ${SSH_USERNAME}@<node IP>
+apiclient exec admin bash
+sheltie
+systemctl restart kubelet
+    {{< /tab >}}
+    {{< /tabpane >}}
+
+1. Add back `node-role.kubernetes.io/control-plane` label to **all** the control plane nodes.
+
+    ```bash
+    kubectl label nodes <control-plane-node-name> node-role.kubernetes.io/control-plane= --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    ```
+
+1. Remove the lagacy nodes (if any).
+
+    ```bash
+    kubectl get nodes --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+
+    NAME                              STATUS     ROLES           AGE     VERSION
+    w01-bbtdd                         NotReady   control-plane   3d23h   v1.27.3-eks-6f07bbc
+    w01-md-0-66dbcfb56cxng8lc-8ppv5   NotReady   <none>          3d23h   v1.27.3-eks-6f07bbc
+    w02-fcbm2j                        Ready      control-plane   91m     v1.27.3-eks-6f07bbc
+    w02-md-0-b7cc67cd4xd86jf-4c9ktp   Ready      <none>          73m     v1.27.3-eks-6f07bbc
+    ```
+    
+    ```bash
+    kubectl delete node w01-bbtdd --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    kubectl delete node w01-md-0-66dbcfb56cxng8lc-8ppv5 --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    ```
+
+1. Re-apply the original config map objects of the workload cluster.
+
+    Re-apply the `cluster-info`, `kube-proxy` and `kubeadm-config` config map objects we saved in previous step to the workload cluster `w02`.
+
+    ```bash
+    kubectl apply -f w02-cm.yaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    ```
+
+1. Validate the nodes are in ready state.
+
+    ```bash
+    kubectl get nodes --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+
+    NAME                              STATUS   ROLES           AGE     VERSION
+    w02-djshz                         Ready    control-plane   9m7s    v1.27.3-eks-6f07bbc
+    w02-md-0-6bbc8dd6d4xbgcjh-wfmb6   Ready    <none>          3m55s   v1.27.3-eks-6f07bbc
+    ```
+
+1. Restart the system pods to ensure that they use the config maps you re-applied in previous step.
+
+    ```bash
+    kubectl rollout restart ds kube-proxy -n kube-system --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    ```
+
+1. Unpause the cluster reconcilers
+
+    ```bash
+    kubectl annotate clusters.anywhere.eks.amazonaws.com $WORKLOAD_CLUSTER_NAME anywhere.eks.amazonaws.com/paused- --kubeconfig=$MGMT_CLUSTER_KUBECONFIG
+
+    kubectl patch clusters.cluster.x-k8s.io $WORKLOAD_CLUSTER_NAME --type merge -p '{"spec":{"paused": false}}' -n eksa-system --kubeconfig=$MGMT_CLUSTER_KUBECONFIG
+    ```
+
+1. Rollout and restart all the machine objects so that the workload cluster has a clean state.
+
+    ```bash
+    # Substitute the EKS Anywhere release version with whatever CLI version you are using
+    EKSA_RELEASE_VERSION=v0.18.3
+    BUNDLE_MANIFEST_URL=$(curl -s https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml | yq ".spec.releases[] | select(.version==\"$EKSA_RELEASE_VERSION\").bundleManifestUrl")
+    CLI_TOOLS_IMAGE=$(curl -s $BUNDLE_MANIFEST_URL | yq ".spec.versionsBundles[0].eksa.cliTools.uri")
+
+
+    # Rollout restart all the control plane machines
+    docker run -i --network host -w $(pwd) -v $(pwd):/$(pwd) --entrypoint clusterctl ${CLI_TOOLS_IMAGE} alpha rollout restart kubeadmcontrolplane/${WORKLOAD_CLUSTER_NAME} -n eksa-system --kubeconfig ${MGMT_CLUSTER_KUBECONFIG}
+
+    # Rollout restart all the worker machines
+    # You need to repeat below command for each worker node group
+    docker run -i --network host -w $(pwd) -v $(pwd):/$(pwd) --entrypoint clusterctl ${CLI_TOOLS_IMAGE} alpha rollout restart machinedeployment/${WORKLOAD_CLUSTER_NAME}-md-0 -n eksa-system --kubeconfig ${MGMT_CLUSTER_KUBECONFIG}
+    ```
+
+1. Validate the new workload cluster is in the desired state.
+
+    ```bash
+    kubectl get clusters -n default -o custom-columns="NAME:.metadata.name,READY:.status.conditions[?(@.type=='Ready')].status" --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+
+    NAME   READY
+    mgmt   True
+    w02    True
+
+    kubectl get clusters.cluster.x-k8s.io -n eksa-system --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+
+    NAME   PHASE         AGE
+    mgmt   Provisioned   11h   
+    w02    Provisioned   11h 
+
+    kubectl get kcp -n eksa-system  --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+
+    NAME   CLUSTER    INITIALIZED   API SERVER AVAILABLE   REPLICAS   READY   UPDATED   UNAVAILABLE   AGE   VERSION
+    mgmt   mgmt       true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    w02    w02        true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    ```
--- a/docs/content/en/docs/clustermgmt/cluster-rebootnode.md
+++ b/docs/content/en/docs/clustermgmt/cluster-rebootnode.md
@@ -42,17 +42,8 @@ If this cannot be verified, do not proceed any further
     - add the paused annotation to the EKSA clusters and CAPI clusters: 
     ```bash
     kubectl annotate clusters.anywhere.eks.amazonaws.com $CLUSTER_NAME anywhere.eks.amazonaws.com/paused=true --kubeconfig=$MGMT_KUBECONFIG
-    ```
 
-    **NOTE**: If you are using vSphere provider, it is also necessary to set `cluster.spec.paused` to true. For example:
-    ```bash
-    kubectl edit clusters.cluster.x-k8s.io -n eksa-system $CLUSTER_NAME --kubeconfig=$MGMT_KUBECONFIG
-    ```
-    add the `paused: true` line under the spec section:
-    ```bash
-    ...
-    spec:
-      paused: true
+    kubectl patch clusters.cluster.x-k8s.io $CLUSTER_NAME --type merge -p '{"spec":{"paused": true}}' -n eksa-system --kubeconfig=$MGMT_KUBECONFIG
     ```
 
 1. For all of the nodes in the cluster, perform the following steps in this order: worker nodes, control plane nodes, and etcd nodes.
@@ -83,15 +74,6 @@ If this cannot be verified, do not proceed any further
 1. Remove the paused annotations from EKS Anywhere cluster.
     ```bash
     kubectl annotate clusters.anywhere.eks.amazonaws.com $CLUSTER_NAME anywhere.eks.amazonaws.com/paused- --kubeconfig=$MGMT_KUBECONFIG
-    ```
 
-    **NOTE**: If you are using vSphere provider, it is also necessary to set `cluster.spec.paused` to false
-    ```bash
-    kubectl edit clusters.cluster.x-k8s.io -n eksa-system $CLUSTER_NAME --kubeconfig=$MGMT_KUBECONFIG
-    ```
-    set paused in the spec section to false:
-    ```bash
-    ...
-    spec:
-      paused: false
+    kubectl patch clusters.cluster.x-k8s.io $CLUSTER_NAME --type merge -p '{"spec":{"paused": false}}' -n eksa-system --kubeconfig=$MGMT_KUBECONFIG
     ```
--- a/docs/content/en/docs/clustermgmt/etcd-backup-restore/bottlerocket-etcd-backup.md
+++ b/docs/content/en/docs/clustermgmt/etcd-backup-restore/bottlerocket-etcd-backup.md
@@ -116,11 +116,9 @@ Make sure to setup the [admin environment variables]({{< relref "#admin-machine-
 
     Before starting the process of restoring etcd, you have to pause some cluster reconciliation objects so EKS Anywhere doesn't try to perform any operations on the cluster while you restore the etcd snapshot.
     ```bash
-    # Pause control plane reconcilation
-    kubectl --kubeconfig=${MANAGEMENT_KUBECONFIG} -n eksa-system annotate machinehealthchecks ${CLUSTER_NAME}-kcp-unhealthy cluster.x-k8s.io/paused=true
+    kubectl annotate clusters.anywhere.eks.amazonaws.com $CLUSTER_NAME anywhere.eks.amazonaws.com/paused=true --kubeconfig=$MANAGEMENT_KUBECONFIG
 
-    # Pause etcd reconcilation
-    kubectl --kubeconfig=${MANAGEMENT_KUBECONFIG} -n eksa-system annotate etcdadmclusters ${CLUSTER_NAME}-etcd cluster.x-k8s.io/paused=true
+    kubectl patch clusters.cluster.x-k8s.io $CLUSTER_NAME --type merge -p '{"spec":{"paused": true}}' -n eksa-system --kubeconfig=$MANAGEMENT_KUBECONFIG
     ```
 
 2. Stop control plane core components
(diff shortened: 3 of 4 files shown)


docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,146 @@
...
+    eksctl anywhere create cluster -f mgmt-new.yaml
+    ```
+
+1. Move the custom resources of all the workload clusters to the new management cluster created above.
+
+    Using vSphere provider as an example, we are moving the Cluster API custom resources, such as `vpsherevms`, `vspheremachines` and `machines` of the **workload clusters**, from the old management cluster to the new management cluster created in above step. Using the `--filter-cluster` flag with `clusterctl move` command so that we are only targeting the custom resources from the workload clusters.
+
+
+    ```bash
+    MGMT_CLUSTER_OLD="mgmt-old"
```
vivek-koppuru, line 43: Maybe call out that this is an example?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,146 @@
...
+
+Depending on whether the infrastructure components are changed or not after the etcd backup was take - e.g. machines are rolled out and recreated, new IP address assigned to the machines - different strategy needs to be applied in order to restore the management cluster.
+
+### Infrastructure components not changed after etcd backup was taken
+
+If the underlying infrastructure machines are not changed after the etcd backup was taken, simply follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the management cluster from the backup.
+
+TODO: what if cluster is not accessible?
+
+### Infrastructure components changed after etcd backup was taken
```
vivek-koppuru, line 25: So this will mainly cover replacing the management cluster only?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Prerequisite
+
+Always backup your EKS Anywhere cluster. Refer to the [Backup cluster]({{< relref "./backup-cluster" >}}) and make sure you have the updated etcd and Cluster API backup at hand.
+
+## Restore a management cluster
+
+As EKS Anywhere management cluster contains the management components of itself plus all the workload clusters it manages, the restoring process can be more complicated than just restoring all the objects from the etcd backup. To be more specific, all the core EKS Anywhere and Cluster API custom resources, managing the lifecycle (provisioning, upgrading, operating, etc.) of the management and its workload clusters, are stored in the management cluster. This includes all the supporting infrastructure, like virtual machines, networks and load balancers. For example, after a failed cluster upgrade, the infrastructure components can change after the etcd backup was taken. Since the backup does not contain the new state of the half upgraded cluster, simply restoring it can create the virtual machines UUID and IP mismatches, rendering EKS Anywhere incapable of healing the cluster.
```
abhay-krishna, line 19:
```suggestion
As EKS Anywhere management cluster contains the management components of itself plus all the workload clusters it manages, the restoration process can be more complicated than just restoring all the objects from the etcd backup. To be more specific, all the core EKS Anywhere and Cluster API custom resources, managing the lifecycle (provisioning, upgrading, operating, etc.) of the management and its workload clusters, are stored in the management cluster. This includes all the supporting infrastructure, like virtual machines, networks and load balancers. For example, after a failed cluster upgrade, the infrastructure components can change after the etcd backup was taken. Since the backup does not contain the new state of the half upgraded cluster, simply restoring it can create the virtual machines UUID and IP mismatches, rendering EKS Anywhere incapable of healing the cluster.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/backup-cluster.md
```diff
@@ -0,0 +1,39 @@
...
+## Etcd backup
+
+For optimal cluster maintenance, it is crucial to perform regular etcd backups on all your EKS Anywhere management and workload clusters. In the event of a cluster upgrade failure, make sure to **always** execute the etcd backup beforehand. To create an etcd backup for your cluster, follow the guidelines provided in the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) section.
+
+
+## Cluster API backup
+
+Since cluster failures primarily occur following unsuccessful cluster upgrades, EKS Anywhere takes the proactive step of automatically creating backups for the Cluster API objects that capture the states of both the management cluster and its workload clusters. These backups are stored within the management cluster folder, where the upgrade command is initiated from the Admin machine, and are generated before each upgrade process.
+
+Although the likelihood of a cluster failure occurring without any associated cluster upgrade operation is relatively low, it is still recommended to manually back up these Cluster API objects on a routine basis. For example, to create a Cluster API backup of a cluster:
```
vivek-koppuru, line 22:
```suggestion
Although the likelihood of a cluster failure occurring without any associated cluster upgrade operation is relatively low, it is still recommended to manually back up these Cluster API objects after every upgrade. For example, to create a Cluster API backup of a cluster:
```

Also, we do a backup before every upgrade right, regardless of success or failure? Should we just be using that to backup instead of having the user run this command? My above suggestion is if we change these instructions to backup the restore folder that gets created.

docs/content/en/docs/clustermgmt/cluster-backup-restore/backup-cluster.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Etcd backup
+
+For optimal cluster maintenance, it is crucial to perform regular etcd backups on all your EKS Anywhere management and workload clusters. In the event of a cluster upgrade failure, make sure to **always** execute the etcd backup beforehand. To create an etcd backup for your cluster, follow the guidelines provided in the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) section.
+
+
+## Cluster API backup
+
+Since cluster failures primarily occur following unsuccessful cluster upgrades, EKS Anywhere takes the proactive step of automatically creating backups for the Cluster API objects that capture the states of both the management cluster and its workload clusters. These backups are stored within the management cluster folder, where the upgrade command is initiated from the Admin machine, and are generated before each upgrade process.
```
abhinavmpandey08, line 20: Just want to confirm, does the workload cluster upgrade also take a backup of all other clusters including management?
Also, I think adding the folder layout here might be beneficial. Something like this:
```
my-mgmt-cluster/
├── cluster-state-backup-2023-10-11T02_55_56        <------ Folder with a backup of CAPI objects
├── my-mgmt-cluster-eks-a-cluster.kubeconfig
├── my-mgmt-cluster-eks-a-cluster.yaml
└── generated
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/backup-cluster.md
```diff
@@ -0,0 +1,39 @@
...
+Although the likelihood of a cluster failure occurring without any associated cluster upgrade operation is relatively low, it is still recommended to manually back up these Cluster API objects on a routine basis. For example, to create a Cluster API backup of a cluster:
+
+
+```bash
+MGMT_CLUSTER="mgmt"
+MGMT_CLUSTER_KUBECONFIG=${MGMT_CLUSTER}/${MGMT_CLUSTER}-eks-a-cluster.kubeconfig
+BACKUP_DIRECTORY=backup-mgmt
+
+# Substitute the container version with whatever EKS Anywhere CLI version you are using
+CONTAINER=public.ecr.aws/eks-anywhere/cli-tools:v0.16.2-eks-a-41
```
abhinavmpandey08, line 31: Can we provide a cmd here to set the `CONTAINER`? Most users won't have an easy way of deducing this image.
@abhay-krishna can you provide a fancy cmd for this?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
abhinavmpandey08, line 19: nit
```suggestion
As EKS Anywhere management cluster contains the management components of itself plus all the workload clusters it manages, the restoring process can be more complicated than just restoring all the objects from the etcd backup. To be more specific, all the core EKS Anywhere and Cluster API custom resources, that manage the lifecycle (provisioning, upgrading, operating, etc.) of the management and its workload clusters, are stored in the management cluster. This includes all the supporting infrastructure, like virtual machines, networks and load balancers. For example, after a failed cluster upgrade, the infrastructure components can change after the etcd backup was taken. Since the backup does not contain the new state of the half upgraded cluster, simply restoring it can create virtual machines UUID and IP mismatches, rendering EKS Anywhere incapable of healing the cluster.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,148 @@
...
+    # Substitute the workspace path with the workspace you are using
+    WORKSPACE_PATH="/home/ec2-user/eks-a"
+    CLUSTER_STATE_BACKUP_LATEST=$(ls -Art ${WORKSPACE_PATH}/${MGMT_CLUSTER_OLD} | grep 'cluster-state-backup' | tail -1)
+    CLUSTER_STATE_BACKUP_LATEST_PATH=${WORKSPACE_PATH}/${MGMT_CLUSTER_OLD}/${CLUSTER_STATE_BACKUP_LATEST}/
+
+    # Substitute the container version with whatever EKS Anywhere CLI version you are using
+    CONTAINER=public.ecr.aws/eks-anywhere/cli-tools:v0.16.2-eks-a-41
+
+
+    # The clusterctl move command needs to be executed on each workload cluster you have. It will will only move the workload cluster resources from the EKS Anywhere backup to the new management cluster. If you have multiple workload clusters, you have to run the command for each cluster as shown below.
```
abhinavmpandey08, line 60: nit: can we break this up into multiple lines so it's easier to read when rendererd?
```suggestion
    # The clusterctl move command needs to be executed for each workload cluster. 
    # It will only move the workload cluster resources from the EKS Anywhere backup to the new management cluster. 
    # If you have multiple workload clusters, you have to run the command for each cluster as shown below.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,148 @@
...
+
+    # Move workload cluster w01 resources to the new management cluster mgmt-new
+    docker run -i --network host -w $(pwd) -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd):/$(pwd) --entrypoint clusterctl ${CONTAINER} move \
```
abhinavmpandey08, line 63: do we need to mount the docker sock? This container is only running clusterctl so it shouldn't need docker.sock?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,148 @@
...
+
+    NAME       CLUSTER    INITIALIZED   API SERVER AVAILABLE   REPLICAS   READY   UPDATED   UNAVAILABLE   AGE   VERSION
+    mgmt-new   mgmt-new   true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    w01        w01        true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    w02        w02        true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    ```
+
+## Restore a workload cluster 
+
+Similar to failed management cluster without infrastructure components change, follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the workload cluster itself from the backup.
```
abhinavmpandey08, line 148: Are you also planning on adding a section here for restoring a workload cluster if infrastructure changes do happen? 
For example: during a workload cluster upgrade, if all the CP got rolled out but the upgrade failed during worker upgrade, a simple ETCD restore won't work, since doing a restore would cause the node names, IPs and potentially other things to revert to a state that is longer valid.
Etcd restore is really only a good idea if users lose only their etcd cluster entirely and want to recover their data, or they want to revert their own deployments to the previous state and nothing else in the infra (nodes specifically) has changed.

docs/content/en/docs/clustermgmt/cluster-backup-restore/backup-cluster.md
jiayiwang7: we do for upgrade, but not for create. I'm including manual backup instruction in case a cluster just failed without eksa upgrade.

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: yea i was kinda avoid talking about this until we have a solution for the case where workload cluster is completely inaccessible / infrastructure changed. But you are right, let me add these notes for clarification.

abhinavmpandey08 (approved):
/lgtm
/woof

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+    mgmt-new   mgmt-new   true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    w01        w01        true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    w02        w02        true          true                   2          2       2                       11h   v1.27.1-eks-1-27-4
+    ```
+
+## Restore a workload cluster 
+
+### Cluster accessible and the infrastructure components not changed after etcd backup was taken
+
+Similar to failed management cluster without infrastructure components change, follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the workload cluster itself from the backup.
```
chrisnegus, line 159:
```suggestion
Similar to the failed management cluster without infrastructure components change situation, follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the workload cluster itself from the backup.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+{{% alert title="Warning" color="warning" %}}
+
+Do not apply the etcd restore unless you are very sure that the infrastructure layer is not changed after the etcd backup was taken - i.e. the node, machines, VMs and their assigned IPs need to be exactly the same as when the backup was taken.
```
chrisnegus, line 163:
```suggestion
Do not apply the etcd restore unless you are very sure that the infrastructure layer is not changed after the etcd backup was taken. In other words, the nodes, machines, VMs and their assigned IPs need to be exactly the same as when the backup was taken.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+{{% /alert %}}
+
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+During a workload cluster upgrade, if all the control plane nodes get rolled out but the upgrade fails during worker nodes upgrade, a simple etcd restore will not work, since doing a restore would cause the node names, IPs and potentially other infrastructure configurations to revert to a state that is no longer valid. Similarly, when the workload cluster is completely inaccessible, restoring etcd in a newly created workload cluster will not work due to the mismatch between the new and old clusters' node spec. 
+
+Restoring etcd effectively takes a cluster back in time and all clients will experience a conflicting, parallel history. This can impact the behavior of watching components like Kubernetes controller managers, EKS Anywhere cluster controller manager, Cluster API controller managers. Etcd restore is only suitable if you lose only your etcd cluster and want to recover your data, or revert your own deployments to the previous state and nothing else in the infrastructure layer (nodes specifically) has changed.
```
chrisnegus, line 171:
```suggestion
Restoring etcd effectively takes a cluster back in time and all clients will experience a conflicting, parallel history. This can impact the behavior of watching components like Kubernetes controller managers, EKS Anywhere cluster controller manager, and Cluster API controller managers. Etcd restore is only suitable if you lose only your etcd cluster and want to recover your data, or revert your own deployments to the previous state and nothing else in the infrastructure layer (nodes specifically) has changed.
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+Therefore under this extreme circumstance, you may need to manually update the CAPI infrastructure objects, such as the infra VMs and machines to use the existing or latest configurations in order to bring the workload cluster back to healthy state.
```
chrisnegus, line 173:
```suggestion
Therefore under this extreme circumstance, you may need to manually update the CAPI infrastructure objects, such as the infra VMs and machines to use the existing or latest configurations in order to bring the workload cluster back to a healthy state. 
```

chrisnegus (commented):
This reads really well @jiayiwang7 . I just have a few small wording suggestions.

docs/content/en/docs/clustermgmt/cluster-backup-restore/backup-cluster.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Etcd backup
+
+For optimal cluster maintenance, it is crucial to perform regular etcd backups on all your EKS Anywhere management and workload clusters. **Always** take an etcd backup before performing an upgrade so it can be used to restore the cluster to a previous state in the event of a cluster upgrade failure. To create an etcd backup for your cluster, follow the guidelines provided in the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) section.
+
+
+## Cluster API backup
+
+Since cluster failures primarily occur following unsuccessful cluster upgrades, EKS Anywhere takes the proactive step of automatically creating backups for the Cluster API objects that capture the states of both the management cluster and its workload clusters. These backups are stored within the management cluster folder, where the upgrade command is initiated from the Admin machine, and are generated before each management cluster upgrade process. For example, after executing a cluster upgrade command on `mgmt-cluster`, a backup folder is generated with the naming convention of `cluster-state-backup-${timestamp}`:
```
g-gaston, line 20: If I'm not mistaken, there is a chance we don't backup all clusters if one of them is not ready and just fallback to only backup the management cluster. Maybe we should point this out?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+{{% /alert %}}
+
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+If the cluster is no longer accessible in any means, or the infrastructure machines are changed after the etcd backup was taken, restoring this management cluster itself from the outdated etcd backup will not work. Instead, you need to create a new management cluster, and migrate all the EKS Anywhere resources of the old workload clusters to the new one, so that the new management cluster can maintain the new ownership of managing the existing workload clusters. Below is an example of migrating a failed management cluster `mgmt-old` with its workload clusters `w01` and `w02` to a new management cluster `mgmt-new`:
+
+1. Create a new management cluster to which you will be migrating your workload clusters later.
+
+    You can define a cluster config similar to your old management cluster, with a different cluster name, and run cluster creation of the new management cluster.
```
g-gaston, line 39: > with a different cluster name

is this necessary? why?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+    MGMT_CLUSTER_OLD="mgmt-old"
+    MGMT_CLUSTER_NEW="mgmt-new"
+    MGMT_CLUSTER_NEW_KUBECONFIG=${MGMT_CLUSTER_NEW}/${MGMT_CLUSTER_NEW}-eks-a-cluster.kubeconfig
+    WORKLOAD_CLUSTER_1="w01"
+    WORKLOAD_CLUSTER_2="w02"
+
+    # Substitute the workspace path with the workspace you are using
+    WORKSPACE_PATH="/home/ec2-user/eks-a"
+    CLUSTER_STATE_BACKUP_LATEST=$(ls -Art ${WORKSPACE_PATH}/${MGMT_CLUSTER_OLD} | grep 'cluster-state-backup' | tail -1)
+    CLUSTER_STATE_BACKUP_LATEST_PATH=${WORKSPACE_PATH}/${MGMT_CLUSTER_OLD}/${CLUSTER_STATE_BACKUP_LATEST}/
```
g-gaston, line 60: It's cool to have this automated here but I think we should explain what this is in a way that allows the user to understand the right value and set it themselves if they need to. IMO understanding all these steps and what they do is more important that providing a script that does everything automatically.

In a lot cases, something will not work and will require some manual tinkering. If we don't have enough "text" explanation, users will need to deeply read the code to understand what it does. And reading bash is not the easiest thing to do :)

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+    # Substitute the EKS Anywhere release version with whatever CLI version you are using
```
g-gaston, line 62:
```suggestion
    # Substitute the EKS Anywhere release version with the EKS-A version of the original management cluster
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+If the cluster is no longer accessible in any means, or the infrastructure machines are changed after the etcd backup was taken, restoring this management cluster itself from the outdated etcd backup will not work. Instead, you need to create a new management cluster, and migrate all the EKS Anywhere resources of the old workload clusters to the new one, so that the new management cluster can maintain the new ownership of managing the existing workload clusters. Below is an example of migrating a failed management cluster `mgmt-old` with its workload clusters `w01` and `w02` to a new management cluster `mgmt-new`:
+
+1. Create a new management cluster to which you will be migrating your workload clusters later.
+
+    You can define a cluster config similar to your old management cluster, with a different cluster name, and run cluster creation of the new management cluster.
+
+    ```sh
+    eksctl anywhere create cluster -f mgmt-new.yaml
```
g-gaston, line 42: I think we should make it very clear that they need to use the exact same eks-a version their old management cluster was at.

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+{{% alert title="Warning" color="warning" %}}
+
+Do not apply the etcd restore unless you are very sure that the infrastructure layer is not changed after the etcd backup was taken. In other words, the nodes, machines, VMs, and their assigned IPs need to be exactly the same as when the backup was taken.
+
+{{% /alert %}}
+
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+During a workload cluster upgrade, if all the control plane nodes get rolled out but the upgrade fails during worker nodes upgrade, a simple etcd restore will not work, since doing a restore would cause the node names, IPs and potentially other infrastructure configurations to revert to a state that is no longer valid. Similarly, when the workload cluster is completely inaccessible, restoring etcd in a newly created workload cluster will not work due to the mismatch between the new and old clusters' node spec.
```
g-gaston, line 169: This is interesting. I'm assuming we have tested this. Do we have the error messages of the different components?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,173 @@
...
+
+Restoring etcd effectively takes a cluster back in time and all clients will experience a conflicting, parallel history. This can impact the behavior of watching components like Kubernetes controller managers, EKS Anywhere cluster controller manager, and Cluster API controller managers. Etcd restore is only suitable if you lose only your etcd cluster and want to recover your data, or revert your own deployments to the previous state and nothing else in the infrastructure layer (nodes specifically) has changed.
+
+Therefore under this extreme circumstance, you may need to manually update the CAPI infrastructure objects, such as the infra VMs and machines to use the existing or latest configurations in order to bring the workload cluster back to a healthy state.
```
g-gaston, line 173: I don't understand this part. What does need to be updated? Why?

I understand if we haven't had time to test this yet, but to me this line is confusing and doesn't really tell the user what to do. If we don't know/can't really provide the right instructions, shouldn't we just advice to create a new workload cluster and move the workloads?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: it is not necessary, just more clear. Esp when old cluster still exists/accessible with the folder name

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: etcd backup includes backing up the node itself. So the new workload cluster w02 will have original cluster w01 node spec after etcd backup. And they will all be `notReady` state

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: I just don't have an exact/concrete plan for this situation yet.

> create a new workload cluster and move the workloads?

move the workloads how? with third party backup solution like velero?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
g-gaston: Ah I see I see. Can the user just go and delete the old node objects with kubectl? For some providers they will be deleted automatically, but there is no harm in doing it manually.

Is this the only issue with the outdated backup? If it is, it seems like deleting the node objects is a fair tradeoff vs having to tinker with capi objects or recreating the cluster.

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
g-gaston: that depends on the user setup
But I'm assuming everyone has at least the yaml manifests checked in a git repo. Even if they don't have everything automated, everyone should be able to redeploy all those manifests in the new cluster.

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+    MGMT_CLUSTER_NEW_KUBECONFIG=${MGMT_CLUSTER_NEW}/${MGMT_CLUSTER_NEW}-eks-a-cluster.kubeconfig
+
+    kubectl apply -f w01/w01-eks-a-cluster.yaml --kubeconfig ${MGMT_CLUSTER_NEW_KUBECONFIG}
+    kubectl apply -f w02/w02-eks-a-cluster.yaml --kubeconfig ${MGMT_CLUSTER_NEW_KUBECONFIG}
+    ```
+
+1. Validate all clusters are in the desired state.
+
+    ```bash
+    kubectl get clusters -n default --kubeconfig ${MGMT_CLUSTER_NEW}/${MGMT_CLUSTER_NEW}-eks-a-cluster.kubeconfig
```
g-gaston, line 140: you could check here that the eks-a clusters are also ready through their status
like with a custom column in this command

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+
+Similar to the failed management cluster without infrastructure components change situation, follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the workload cluster itself from the backup.
+
+### Cluster not accessible or infrastructure components changed after etcd backup was taken
+
+If the workload cluster is still accessible, but the infrastructure machines are changed after the etcd backup was taken, you can still try restoring the cluster itself from the etcd backup. Although doing so is risky: it can potentially cause the node names, IPs and other infrastructure configurations to revert to a state that is no longer valid. Restoring etcd effectively takes a cluster back in time and all clients will experience a conflicting, parallel history. This can impact the behavior of watching components like Kubernetes controller managers, EKS Anywhere cluster controller manager, and Cluster API controller managers. You may need to manually update the CAPI infrastructure objects, such as the infra VMs and machines to use the existing or latest configurations in order to bring the workload cluster back to a healthy state.
```
g-gaston, line 170: > You may need to manually update the CAPI infrastructure objects, such as the infra VMs and machines to use the existing or latest configurations in order to bring the workload cluster back to a healthy state.

I think I'm missing the scenario where this is needed. Shouldn't this work in the same way as the process you outline below (restoring backup in a new workload cluster).

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+    eksctl anywhere create cluster -f w02.yaml --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+    ```
+
+1. Follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the old workload cluster's etcd backup `etcd-snapshot-w01.db` onto the new workload cluster `w02`.
+
+    Use different restore process based on OS family:
+    * [BottleRocket]({{< relref "../etcd-backup-restore/bottlerocket-etcd-backup/#restore-etcd-from-backup" >}})
+    * [Ubuntu]({{< relref "../etcd-backup-restore/ubuntu-rhel-etcd-backup/#restore" >}})
+
+    You might notice that after restoring the original etcd backup to the new workload cluster `w02`, all the node names have prefix `w01-*` and the nodes go to `NotReady` state. This is because restoring etcd effectively applies the node data from the original cluster which causes a conflicting history and can impact the behavior of watching components like Kubelets, Kubernetes controller managers.
```
g-gaston, line 189: Should we add a step to delete the old node entries?
I think some of the providers will leave them hanging and will need manual cleanup

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: yea it does. we can remove this for less confusion

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+
+If the original workload cluster is no longer accessible in any means, or restoring the workload cluster itself from the outdated etcd does not bring the cluster back to healthy state, then you need to create a new workload cluster based of the same management cluster that managed the original workload cluster, and restore all your workload applications from the etcd backup of the original workload cluster to the new one, so that the same management cluster can maintain the ownership of managing the new workload cluster, with all the same data from the old workload cluster. Below is an example of applying the etcd backup `etcd-snapshot-w01.db` of a failed workload cluster `w01` to a new workload cluster `w02`:
```
vivek-koppuru, line 172:
```suggestion
If the original workload cluster is no longer accessible in any means, or restoring the workload cluster itself from the outdated etcd does not bring the cluster back to healthy state, then you need to create a new workload cluster based on the same management cluster that managed the original workload cluster, and restore all your workload applications from the etcd backup of the original workload cluster to the new one, so that the same management cluster can maintain the ownership of managing the new workload cluster, with all the same data from the old workload cluster. Below is an example of applying the etcd backup `etcd-snapshot-w01.db` of a failed workload cluster `w01` to a new workload cluster `w02`:
```

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+
+    ```bash
+    eksctl anywhere create cluster -f w02.yaml --kubeconfig $MGMT_CLUSTER_KUBECONFIG
+    ```
+
+1. Follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the old workload cluster's etcd backup `etcd-snapshot-w01.db` onto the new workload cluster `w02`.
+
+    Use different restore process based on OS family:
+    * [BottleRocket]({{< relref "../etcd-backup-restore/bottlerocket-etcd-backup/#restore-etcd-from-backup" >}})
+    * [Ubuntu]({{< relref "../etcd-backup-restore/ubuntu-rhel-etcd-backup/#restore" >}})
```
vivek-koppuru, line 187:
```suggestion
1. Follow the [External etcd backup and restore]({{< relref "../etcd-backup-restore/etcdbackup" >}}) to restore the old workload cluster's etcd backup `etcd-snapshot-w01.db` onto the new workload cluster `w02`. Use different restore process based on OS family:
    * [BottleRocket]({{< relref "../etcd-backup-restore/bottlerocket-etcd-backup/#restore-etcd-from-backup" >}})
    * [Ubuntu]({{< relref "../etcd-backup-restore/ubuntu-rhel-etcd-backup/#restore" >}})
```
Thought this flows better but if you would like to keep the previous you should

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -0,0 +1,245 @@
...
+# SSH into the control plane and worker nodes. You must do this for each node.
+ssh -i ${SSH_KEY} ${SSH_USERNAME}@<node IP>
+sudo su
+systemctl restart kubelet
+    {{< /tab >}}
+    {{< tab header="Bottlerocket" lang="bash" >}}
+# SSH into the control plane and worker nodes. You must do this for each node.
+ssh -i ${SSH_KEY} ${SSH_USERNAME}@<node IP>
+apiclient exec admin bash
+sheltie
```
vivek-koppuru, line 233:
```suggestion
sudo sheltie
```
Or is the sudo not required here? I think if you ssh to the node and directly run sudo sheltie, that is also enough not requiring you to run the apiclient command right?

jiayiwang7: when you are in admin container, sudo is not required

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -161,4 +161,179 @@ If the cluster is no longer accessible in any means, or the infrastructure machi
...
+    cat <<EOF >> w02-cm.yaml
+    $(kubectl get -n kube-public cm cluster-info -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    ---
+    $(kubectl get -n kube-system cm kube-proxy -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    ---
+    $(kubectl get -n kube-system cm kubeadm-config -oyaml --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG)
+    EOF
+    ```
+
+    Remove the metadata's `creationTimestamp`, `resourceVersion`, `uid` from the config map.
```
jiayiwang7, line 202: is there a better way? a kubectl cmd?

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -161,4 +161,179 @@ If the cluster is no longer accessible in any means, or the infrastructure machi
...
+
+If the original workload cluster is no longer accessible in any means, or restoring the workload cluster itself from the outdated etcd does not bring the cluster back to healthy state, then you need to create a new workload cluster based on the same management cluster that managed the original workload cluster, and restore all your workload applications from the etcd backup of the original workload cluster to the new one, so that the same management cluster can maintain the ownership of managing the new workload cluster, with all the same data from the old workload cluster. Below is an example of applying the etcd backup `etcd-snapshot-w01.db` of a failed workload cluster `w01` to a new workload cluster `w02`:
```
g-gaston, line 172: nit: this sentence is very long. the content is good but maybe we can split it into multiple sentences

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
```diff
@@ -161,4 +161,179 @@ If the cluster is no longer accessible in any means, or the infrastructure machi
...
+
+    ```bash
+    kubectl rollout restart ds kube-proxy -n kube-system --kubeconfig $WORKLOAD_CLUSTER_KUBECONFIG
+    ```
+
+1. Unpause the cluster reconcilers
+
+    ```bash
+    # unpause control plane reconcilation
+    kubectl --kubeconfig=${MGMT_CLUSTER_KUBECONFIG} -n eksa-system annotate machinehealthchecks ${WORKLOAD_CLUSTER_NAME}-kcp-unhealthy cluster.x-k8s.io/paused-
```
g-gaston, line 296: why `${WORKLOAD_CLUSTER_NAME}-kcp-unhealthy`?

g-gaston: also
```suggestion
    kubectl --kubeconfig=${MGMT_CLUSTER_KUBECONFIG} -n eksa-system annotate kcp ${WORKLOAD_CLUSTER_NAME}-kcp-unhealthy cluster.x-k8s.io/paused-
```

g-gaston (commented):
looks good
just nits and typos

docs/content/en/docs/clustermgmt/cluster-backup-restore/restore-cluster.md
jiayiwang7: hmmm any recommendation to update this? it reads ok to me L)

g-gaston:
```suggestion
If the original workload cluster becomes inaccessible or cannot be restored to a healthy state from an outdated etcd, a new workload cluster needs to be created. This new cluster should be managed by the same management cluster that oversaw the original. You must then restore your workload applications to this new cluster from the etcd backup of the original. This ensures the management cluster retains control, with all data from the old cluster intact. Below is an example of applying the etcd `backup etcd-snapshot-w01.db` from a failed workload cluster `w01` to a new cluster `w02`:
```

g-gaston (commented):
/lgtm

PR #6799, opened 2023-10-11
Strip comments from ssh key from the controller using mutating webhook

panktishah26:
*Description of changes:*
We srtip the comments for tinkerbell cluster create and upgrade operations using CLI. If the `sshAuthorizedKey` is something like `ssh AAAAC3NzaC1lZD abc@xyz.com` we strip the comment portion and make it `ssh AAAAC3NzaC1lZD`. We did not implement this feature for eksa controller. This PR adds the same feature for controller using mutating webhook. 

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
@@ -45,6 +45,8 @@ var _ webhook.Defaulter = &TinkerbellMachineConfig{}
 func (r *TinkerbellMachineConfig) Default() {
 	tinkerbellmachineconfiglog.Info("Setting up Tinkerbell Machine Config defaults", klog.KObj(r))
 	r.SetDefaults()
+	tinkerbellmachineconfiglog.Info("Normalize SSHKeys by removing comments from the keys", klog.KObj(r))
+	normalizeSSHKeys(r)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/api/v1alpha1/tinkerbellmachineconfig_webhook.go
```diff
@@ -45,6 +45,8 @@ var _ webhook.Defaulter = &TinkerbellMachineConfig{}
 func (r *TinkerbellMachineConfig) Default() {
 	tinkerbellmachineconfiglog.Info("Setting up Tinkerbell Machine Config defaults", klog.KObj(r))
 	r.SetDefaults()
+	tinkerbellmachineconfiglog.Info("Mutating Tinkerbell Machine Config resources", klog.KObj(r))
```
chrisdoherty4, line 48: While I appreciate the log message, its a little generic. I would suggest calling `mutateTinkerbellMachineConfigFields(c)` from here and perhaps changing it to `normalizeSSHKeys` then change the log message to be specific. If we end up needing to do this sort of thing again we can rework as needed but I suspect we won't.

chrisdoherty4 (commented):
1 small simplification and clarification with logging and we're good to go.

chrisdoherty4 (approved):

PR #6806, opened 2023-10-12
Added an additional check to make sure etcdadm and capi clusters are available

panktishah26:
*Description of changes:*
We are having an issue in creating workload cluster using the controller if external Etcd is configured on the workload cluster. It used to fail with `failed calling webhook "mutation.vspheremachineconfig.anywhere.amazonaws.com` error. This issue happens because reconciler tries to get the CAPI cluster before the CAPI cluster becomes available for the workload cluster.
Added additional checks to make sure CAPI cluster are Etcdadm cluster are available before moving to check if the etcd machines are available or not.

*Testing (if applicable):*

### DIFF
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -17,6 +17,8 @@ import (
 
 // UpdateClusterStatusForControlPlane checks the current state of the Cluster's control plane and updates the
 // Cluster status information.
+// There is a posibility that UpdateClusterStatusForControlPlane does not update the
+// controleplane status specially in case where it still waiting for cluster objects to be created.
 func UpdateClusterStatusForControlPlane(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
 	kcp, err := controller.GetKubeadmControlPlane(ctx, client, cluster)
 	if err != nil {
@@ -29,10 +31,11 @@ func UpdateClusterStatusForControlPlane(ctx context.Context, client client.Clien
 		if err != nil {
 			return errors.Wrap(err, "getting capi cluster")
 		}
-
-		etcdadmCluster, err = getEtcdadmCluster(ctx, client, capiCluster)
-		if err != nil {
-			return errors.Wrap(err, "reading etcdadm cluster")
+		if capiCluster != nil {
+			etcdadmCluster, err = getEtcdadmCluster(ctx, client, capiCluster)
+			if err != nil {
+				return errors.Wrap(err, "reading etcdadm cluster")
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/clusters/status.go
```diff
@@ -85,6 +86,13 @@ func UpdateClusterStatusForCNI(ctx context.Context, cluster *anywherev1.Cluster)
 // updateConditionsForEtcdAndControlPlane updates the ControlPlaneReady condition if etcdadm cluster is not ready.
 func updateConditionsForEtcdAndControlPlane(cluster *anywherev1.Cluster, kcp *controlplanev1.KubeadmControlPlane, etcdadmCluster *etcdv1.EtcdadmCluster) {
 	// Make sure etcd cluster is ready before marking ControlPlaneReady status to true
+	// This condition happens while creating a workload cluster from the management cluster using controller
+	// where it tries to get the etcdadm cluster for the first time before it generates the resources.
+	if cluster.Spec.ExternalEtcdConfiguration != nil && etcdadmCluster == nil {
+		conditions.MarkFalse(cluster, anywherev1.ControlPlaneReadyCondition, anywherev1.RollingUpgradeInProgress, clusterv1.ConditionSeverityInfo, "Etcd cluster is not available yet")
```
cxbrowne1207, line 92: This could this not happen on create as well right? Is RollingUpgradeInProgress the right reason here?


Maybe we should create a new reason for these etcd scenarios. E.g. `anywherev1.ExternalEtcdNotReady`

chrisdoherty4 (changes requested):
I don't understand how this is possible? What are the circumstances for this happening? It feels peculiar to conditionally update conditions based on the existence of the objects we're responsible for creating all in the same algorithm.

tatlat:
> I don't understand how this is possible. What are the circumstances for this happening? It feels peculiar to conditionally update conditions based on the existence of objects we're responsible for creating all in the same algorithm.

When we apply the workload cluster config in the API tests, the controller would attempt to update the status of the workload cluster before the CAPI cluster could be created. In UpdateClusterStatusForControlPlane, controller.GetCAPICluster would return nil and the controller would get stuck in a a cycle of panicking and restarting. Since the controller was always restarting, the webhooks were unreachable.

chrisdoherty4:
> the controller would attempt to update the status of the workload cluster before the CAPI cluster could be created

Do you mean for the cluster to _actually_ be created or for the Cluster API Cluster object to be submitted to the cluster?

cxbrowne1207 (approved):

chrisdoherty4 (approved):

PR #6821, opened 2023-10-13
Remove patch for bundlesRef and ClusterSpecGenerate struct

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR removes a patch that substitutes the ClusterSpec with a ClusterSpecGenerate so that the bundlesRef field on the would not use `omitempty` on it's JSON tag. This patch is no longer needed, the following was why it was added before.

There was an issue when upgrading from the latest minor release with gitops to v0.17 during the "Applying eksa yaml resources to cluster" step stalled indefinitely. This seems to happen when there is a web-hook validation error, triggering the `k --force apply` to delete the Cluster resource.

In this case, ghough the Cluster object that was force applied was valid with the `eksaVersion` specified and no  `bundlesRef` set, it was trying to do that 3 way strategic merge and added the old `bundlesRef` to the new spec as well. However, there is a web-hook validation that does not allow both to be set at the same time, so that was throwing an error in the `kube-apiserver` pod. Normally, the `k --force apply` would delete the resource and just recreate it, but instead it the CLI gets stuck. We patched this issue by creating `ClusterSpecGenerate` so that we could generate a yaml where `bundlesRef` null so that it would not be omitted when submitted to the `kube-apiserver`.

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -49,7 +49,7 @@ func NewClusterGenerate(clusterName string, opts ...ClusterGenerateOpt) *Cluster
 		ObjectMeta: ObjectMeta{
 			Name: clusterName,
 		},
-		Spec: ClusterSpecGenerate{
+		Spec: ClusterSpec{
 			KubernetesVersion: GetClusterDefaultKubernetesVersion(),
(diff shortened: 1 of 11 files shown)


abhinavmpandey08 (approved):
/lgtm
/woof

PR #6728, opened 2023-09-25
Kindless upgrade changes for vSphere & CloudStack provider

panktishah26:
*Issue #, if available:*

*Description of changes:*
This PR includes fixes for below 3 issues,
1. Validates Cloudstack template name and kubernetes version from the controller side and waits for both to match for further reconciliation.
2. Since we have seperated ETCD and CP reconciliation, it waits for ETCD to be ready before marking ControlPlaneReady to true.
3. From the CLI side during management cluster upgrade, waits for Cluster FailureMessage to be nil before waiting for other condition to met.

*Testing (if applicable):*

*Documentation added/planned (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #6828, opened 2023-10-16
Validate worker node kubernetes version with the template name

panktishah26:
*Description of changes:*
We need to validate worker node specific kubernetes version (if present) in case of modular upgrade with the template name field. The template name field should contain the kubernetes version specified in the workernode configuration if the upgrade type is modular upgrade.

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/cloudstack/validator.go
+++ b/pkg/providers/cloudstack/validator.go
@@ -118,6 +118,11 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, clusterSp
 		return fmt.Errorf("cannot find CloudStackMachineConfig %v for control plane", clusterSpec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name)
 	}
 
+	// validate template field of each CloudStackMachineConfigs with the cluster spec kubernetes version.
+	if err := v.validateTemplateMatchesKubernetesVersion(ctx, controlPlaneMachineConfig.Spec.Template.Name, string(clusterSpec.Cluster.Spec.KubernetesVersion)); err != nil {
+		return fmt.Errorf("machine config %s validation failed: %v", controlPlaneMachineConfig.Name, err)
+	}
+
 	if clusterSpec.Cluster.Spec.ExternalEtcdConfiguration != nil {
 		etcdMachineConfig := etcdMachineConfig(clusterSpec)
 		if etcdMachineConfig == nil {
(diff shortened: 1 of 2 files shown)


pkg/providers/cloudstack/validator.go
```diff
@@ -130,13 +130,21 @@ func (v *Validator) ValidateClusterMachineConfigs(ctx context.Context, clusterSp
 		if !ok {
 			return fmt.Errorf("cannot find CloudStackMachineConfig %v for worker nodes", workerNodeGroupConfiguration.MachineGroupRef.Name)
 		}
+		// validate template field of worker group spec with the kubernetes version of each workerNodeGroup - in case of modular upgrade.
```
chrisdoherty4, line 133: I don't think the algorithm here is correct. We need to validate the following:

1. The control plane's machine config template specifies the top level Kubernetes version. 
2. For each worker node group, if the worker node group specifies its own Kube version, that version exists in the machine configs template, _else_ the top level Kube version eixsts in the machine configs template.

L142-150 are checking each machine config has the root Kube version, but that's not necessarily what we expect due to modular upgrade (point 2 above).

The way this algorithm is structured makes the validation challenging to follow and the resulting error messages are unnecessarily long and contain detail that ought to live on the API documentation. Given we're doing some work in this area can we take a moment to refactor that also?

See https://github.com/aws/eks-anywhere/pull/6805/files#diff-d794b88ef59dfa162a093bf0492363f2f03fb35c96e442ca1644ef4dea859b5eR76-R97 which contains the Tinkerbell check that's fundamentally the same.

chrisdoherty4 (commented):
See the inline comment. I don't think we're doing quite the right checks with these changes and I think the code could benefit from a small refactor to make it clear what our expectations are.

pkg/providers/cloudstack/validator.go
panktishah26: Makes sense. I have reimplemented the algorithm. It satisfies both 1 and 2 points. Regarding the error message, since this is a new ask from the user, we wanted to make it more obvious to understand the behavior here. We discussed this in this PR comment [here](https://github.com/aws/eks-anywhere/pull/6728#discussion_r1336417088) and [here](https://github.com/aws/eks-anywhere/pull/6728#discussion_r1337799633). I am open to any other suggestions.

chrisdoherty4: I empathize with the concern and what we're trying to do, but I don't agree with the approach. There are a few reasons we shouldn't include all the current detail: 

1. Errors should concisely state the problem and leave describing valid inputs and examples to documentation. 
2. It can be challenging to adequately describe valid input for all inputs. For example, perhaps some inputs require a more elaborate explanation and examples that can't be added to errors. We start generating inconsistent errors when we offer examples and explanations in some cases but not others.
3. Documentation and code fall out of sync (I [raised a bug](https://github.com/aws/eks-anywhere/issues/6836) to that effect recently). If we change expectations as we have on a handful of occasions we need to regurgitate the API documentation in this error message; its a maintenance overhead.

If we want to provide remedial recommendations in code I think we need a more comprehensive approach.

chrisdoherty4 (approved):

PR #6857, opened 2023-10-20
Automate envtest setup from tests if not configured

g-gaston:
*Description of changes:*
Before this, running a test that required envtest without setting the KUBEBUILDER_ASSETS env var would fail. Now, if the env var is not set, the envtest setup will download the necessary assets and configure the tests to use them.

This means that `go test` (on individual tests or whole packages) now works out of the box.


jonathanmeier5 (approved):
/lgtm

this is great

PR #6871, opened 2023-10-23
Add template name and uuid name check for Nutanix kindless feature

panktishah26:
*Description of changes:*
Add template name and uuid name check for Nutanix kindless management cluster upgrade feature. This is needed for kubernetes version upgrade where the controller waits until the template name matches the kubernetes version before reconciling the objects. 

We rely on the eksa-controller from now onwards to upgrade the management cluster using CLI. Sometimes it happens that during the kubernetes version upgrade, when we apply the cluster spec changes to the cluster, the cluster object gets upgraded and starts reconciling immediately before machineconfig gets upgraded and it might upgrade the cluster with previous machine config's image UUID or name. To avoid this issue, we have made it mandatory for the image name and image we get from the UUID to have the cluster kubernetes version and/or worker node kubernetes version(modular upgrade) kubernetes version in the name. 

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/nutanix/validator.go
+++ b/pkg/providers/nutanix/validator.go
@@ -66,6 +66,41 @@ func (v *Validator) ValidateClusterSpec(ctx context.Context, spec *cluster.Spec,
 		}
 	}
 
+	return v.checkImageNameMatchesKubernetesVersion(ctx, spec, client)
+}
+
+func (v *Validator) checkImageNameMatchesKubernetesVersion(ctx context.Context, spec *cluster.Spec, client Client) error {
+	controlPlaneMachineConfig := spec.NutanixMachineConfigs[spec.Cluster.Spec.ControlPlaneConfiguration.MachineGroupRef.Name]
+	if controlPlaneMachineConfig == nil {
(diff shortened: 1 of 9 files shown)


pkg/providers/nutanix/validator.go
```diff
@@ -264,6 +299,34 @@ func (v *Validator) validateImageConfig(ctx context.Context, client Client, iden
...
 
+func (v *Validator) validateTemplateMatchesKubernetesVersion(ctx context.Context, identifier anywherev1.NutanixResourceIdentifier, client Client, kubernetesVersionName string) error {
+	var templateName string
+	if identifier.Type == anywherev1.NutanixIdentifierUUID {
+		imageUUID := *identifier.UUID
+		imageDetails, err := client.GetImage(ctx, imageUUID)
+		if err != nil {
+			return fmt.Errorf("failed to find image with uuid %s: %v", imageUUID, err)
+		}
+		templateName = *imageDetails.Spec.Name
```
abhinavmpandey08, line 310: `imageDetails.Spec` and `imageDetails.Spec.Name` are both pointers. Let's validate they aren't nil before we dereference them

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #6899, opened 2023-10-26
Add actions for vulnerability scans

g-gaston:
*Description of changes:*
This should prevent vulnerable deps from being added and also will scan our code to narrow down if we are affecting by new CEV or not. This will run on every PR, every commit to `main` and release branches and periodically every night.


drewvanstone (approved):
Nice, should we add license reviews in a future PR?

PR #6921, opened 2023-10-27
Add SANs configurability for API server

### LINKED ISSUE #6863: FQDN SAN for API Server TLS certificate
As a user I want to specify an FQDN to be added as a [Subject Alternative Name (SAN)](https://en.wikipedia.org/wiki/Subject_Alternative_Name) in API server TLS certificates so I can use FQDNs for cluster interactions.

chrisdoherty4:
Closes #6863.

Add support for specifying API Server Certificate Subject Alternative Names. This allows users to make requests via domain names or alternate IPs.


tatlat (changes requested):
Why are we changing this?

chrisdoherty4:
@tatlat Its been modified to include `sprig` functions instead of copying Helm and Sprig implementations. I wanted some additional functions for the templates.

tatlat (approved):

PR #7048, opened 2023-11-13
Patch incorrect conversion between uint64 and int64

chrisdoherty4:
The semver package unnecessarily uses uint64. This alters it to use int64 alleviating the problem.

Uses of uint should only be used when needed, typically when performing bitwise operations.

### DIFF
--- a/pkg/semver/semver.go
+++ b/pkg/semver/semver.go
@@ -11,7 +11,7 @@ const semverRegex = `^v?(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>
 var semverRegexp = regexp.MustCompile(semverRegex)
 
 type Version struct {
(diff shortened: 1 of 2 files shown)


jacobweinstock:
Hey @chrisdoherty4 , where is behavior being experienced? I think what you describe is what I'm seeing with `generate tinkerbelltemplateconfig` command. Is this the issue to which you're referring? Is it happening anywhere else?

```bash
$ eksctl-anywhere generate tinkerbelltemplateconfig -f testing.yaml                            
Error: unable to get cluster config from file: getting Bundles file: invalid eksa version: invalid major version in semver : strconv.ParseUint: parsing "": invalid syntax
```

chrisdoherty4:
@jacobweinstock This was flagged as a security problem. I wasn't seeing a particular error.

jacobweinstock (approved):

PR #7054, opened 2023-11-15
Add scaffolding for EKS-A in-place upgrade APIs

abhinavmpandey08:
*Description of changes:*
Add Scaffolding for in-place upgrade in EKS-A including APIs/CRDs for in-place as well as bare bones controllers

*Testing (if applicable):*


pkg/api/v1alpha1/controlplaneupgrade_types.go
```diff
@@ -0,0 +1,52 @@
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
chrisdoherty4, line 8: We can delete these, its well known.

pkg/api/v1alpha1/controlplaneupgrade_types.go
```diff
@@ -0,0 +1,52 @@
...
+
+// ControlPlaneUpgradeSpec defines the desired state of ControlPlaneUpgrade.
+type ControlPlaneUpgradeSpec struct {
```
chrisdoherty4, line 11: Is everything required? If not, can we annotate with `+optional`. Additionally if htere's any CEL validation we can include we should add it.

chrisdoherty4, line 11: I think we need documentation for every field on expected formats, expected objects that are being referenced, constraints etc.

controllers/controlplaneupgrade_controller.go
```diff
@@ -0,0 +1,62 @@
...
+type ControlPlaneUpgradeSpecReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=controlplaneupgrades,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=controlplaneupgrades/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=controlplaneupgrades/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
```
chrisdoherty4, line 40: Was this generated? I think we can drop most/all of it.

pkg/api/v1alpha1/controlplaneupgrade_types.go
abhinavmpandey08: This PR is mainly just adding scaffolding so we can have a place to go and start adding code.
We'll have follow up PRs where we are going to add validations/optional tags.

abhinavmpandey08: Same as above. 
This PR is mainly just adding scaffolding so we can have a place to go and start adding code.
We'll have follow up PRs where we are going to add field descriptions and stuff.
Does that sound good?

controllers/controlplaneupgrade_controller.go
abhinavmpandey08: done

pokearu (approved):
/lgtm

PR #7055, opened 2023-11-15
Fix eksa installer unit test with real manifest

tatlat:
*Issue #, if available:*

*Description of changes:*
We have to manually update the object count in our manifest every time we add CRDs. This change counts the number of objects in our components manifest and sets the correct count in the test, so we won't have to manually update it.

*Testing (if applicable):*


pkg/clustermanager/eksa_installer_test.go
```diff
@@ -69,8 +72,15 @@ func newInstallerTest(t *testing.T, opts ...clustermanager.EKSAInstallerOpt) *in
 func TestEKSAInstallerInstallSuccessWithRealManifest(t *testing.T) {
 	tt := newInstallerTest(t)
 	tt.newSpec.VersionsBundles["1.19"].Eksa.Components.URI = "../../config/manifest/eksa-components.yaml"
+	file, err := os.ReadFile("../../config/manifest/eksa-components.yaml")
+	if err != nil {
+		t.Fatalf("could not read eksa-components")
+	}
+	manifest := string(file)
+	expectedObjectCount := strings.Count(manifest, "---")
+	fmt.Println(expectedObjectCount)
```
abhinavmpandey08, line 81: what is this?S???????

PR #7058, opened 2023-11-17
Remove vSphere DHCP customization page

chrisdoherty4:
We include an item on the vSphere pre-req that stipulates a DHCP server is required on the network. The vSphere DHCP doc goes a step further and provides configurations for custom DHCP solutions which we don't support.


chrisnegus (commented):
I have no particular reason to keep this page. But if you do remove it, please also deal with the reference to this page here: https://anywhere.eks.amazonaws.com/docs/troubleshooting/troubleshooting/#no-ip-assigned-to-a-vm

PR #7061, opened 2023-11-19
Add implementation for NodeUpgradeController

abhinavmpandey08:
*Description of changes:*
The implementation is based on this design https://github.com/aws/eks-anywhere/pull/6893

The goal of this PR is to add the implementation for the NodeUpgradeController.
The purpose of this controller is to handle in-place upgrade of nodes by provisioning upgrader pods on the individual nodes that go in and upgrade the components on the node to the desired version.
These components include:
- containerd
- cni-plugins
- kubeadm
- kubelet
- kubectl


config/rbac/role.yaml
```diff
@@ -27,6 +27,16 @@ rules:
   - nodes
   verbs:
   - list
+- apiGroups:
+  - ""
+  resources:
+  - pods
```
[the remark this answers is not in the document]
abhinavmpandey08: No the configmaps workaround stuff doesn't happen through the controller. For that thing, the script just uses the admin config on the control plane nodes which has full permissions

controllers/nodeupgrade_controller.go
```diff
@@ -2,42 +2,390 @@ package controllers
...
+		return ctrl.Result{}, nil
+	}
+
+	if err := namespaceOrCreate(ctx, remoteClient, log, constants.EksaSystemNamespace); err != nil {
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("Upgrading node", "Node", node.Name)
+	upgraderPod := &corev1.Pod{}
+	if conditions.IsTrue(nodeUpgrade, anywherev1.UpgraderPodCreated) || upgraderPodExists(ctx, remoteClient, node.Name) {
```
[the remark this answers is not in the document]
abhinavmpandey08: The first check is just seeing if the condition was added to the NodeUpgrade object but if for some reason the patching fails and the condition is not set, we should still verify if the actual pod exists.

pkg/api/v1alpha1/nodeupgrade_types.go
```diff
@@ -1,28 +1,62 @@
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-// NodeUpgradeKind stores the Kind for NodeUpgrade.
-const NodeUpgradeKind = "NodeUpgrade"
+const (
+	// NodeUpgradeKind stores the Kind for NodeUpgrade.
+	NodeUpgradeKind = "NodeUpgrade"
```
mitalipaygude, line 11: should the value be `NodeUpgradeKind` like all the below constants?

controllers/nodeupgrade_controller.go
```diff
@@ -2,42 +2,390 @@ package controllers
...
+				completed = false
+			}
+		}
+	}
+	nodeUpgrade.Status.Completed = completed
+}
+
+func getInitContainerStatus(pod *corev1.Pod, containerName string) (*corev1.ContainerStatus, error) {
+	for _, status := range pod.Status.InitContainerStatuses {
+		if status.Name == containerName {
```
vivek-koppuru, line 334: We may want a mapping in the future to make it more readable.

pkg/nodeupgrader/upgrader.go
```diff
@@ -0,0 +1,144 @@
...
+
+// UpgradeFirstControlPlanePod returns an upgrader pod that should be deployed on the first control plane node.
+func UpgradeFirstControlPlanePod(nodeName, image, kubernetesVersion, etcdVersion string) *corev1.Pod {
+	p := upgraderPod(nodeName, image)
+	p.Spec.InitContainers = containersForUpgrade(image, nodeName, "kubeadm_in_first_cp", kubernetesVersion, etcdVersion)
+	return p
+}
+
+// UpgradeRestControlPlanePod returns an upgrader pod that can be deployed on the remaining control plane nodes.
+func UpgradeRestControlPlanePod(nodeName, image string) *corev1.Pod {
```
vivek-koppuru, line 48: Instead of Rest, we can maybe say Secondary, Other, Support, etc.

pkg/nodeupgrader/upgrader.go
```diff
@@ -0,0 +1,144 @@
...
+	p := upgraderPod(nodeName, image)
+	p.Spec.InitContainers = containersForUpgrade(image, nodeName, "kubeadm_in_rest_cp")
+	return p
+}
+
+// UpgradeWorkerPod returns an upgrader pod that can be deployed on worker nodes.
+func UpgradeWorkerPod(nodeName, image string) *corev1.Pod {
```
vivek-koppuru, line 55: These feel like they should be member methods

vivek-koppuru (approved):

pkg/nodeupgrader/upgrader.go
abhinavmpandey08: done

controllers/nodeupgrade_controller.go
abhinavmpandey08: yeah I can add that in a follow up

pkg/nodeupgrader/upgrader.go
abhinavmpandey08: hmm, let's discuss this offline and see if this is something we can acknowledge in a follow up

pokearu (approved):
/lgtm

mitalipaygude (approved):

config/rbac/role.yaml
g-gaston, line 33: we need to move this to a role instead of cluster role
this gives way too many privileges to the controller

pkg/api/v1alpha1/nodeupgrade_types.go
```diff
@@ -1,28 +1,62 @@
...
+
+	// UpgraderPodCreated reports whether the upgrader pod has been created for the node upgrade.
+	UpgraderPodCreated ConditionType = "UpgraderPodCreated"
```
g-gaston, line 14: I don't know if it's wise making the phases of an upgrade tied in this way to the API. Your phases might change more frequently than you want your API to change. Also, what if some phases become optional? How are client going to know that?

config/rbac/role.yaml
abhinavmpandey08: I don't think this is needed anymore since the controller is building a remote client from the kubeconfig stored on the cluster which has admin on workload clusters. I'll do some testing and remove it

pkg/api/v1alpha1/nodeupgrade_types.go
abhinavmpandey08: >I don't know if it's wise making the phases of an upgrade tied in this way to the API. Your phases might change more frequently than you want your API to change.

Can you elaborate a little on this? This is just adding container statuses on the node upgrade object.
The controller will check the container statuses at the end of every reconcile loop and update the node upgrade status based on that. And once all the containers have finished, it will mark the upgrade as completed and end the reconcile loop.
I extrapolated  this logic from the EKS-A controller which does something similar where it monitors the KCP/MD/CAPI cluster conditions and uses it to update EKS-A Cluster conditions. And when all conditions are met, the reconcile is marked complete.

>Also, what if some phases become optional? How are client going to know that?

This depends on how we decide to implement it. We were thinking we won't make the phases optional from NodeUpgradeController's perspective. Rather the pod will check if the step it's performing needs to be done or not. If it doesn't need to be performed, the upgrader will just not do it and return a success. 
For example, if the containerd version on the node is already the latest one, then the upgrader can just skip copying over the containerd binary and return without errors.

PR #7110, opened 2023-11-30
Create management cluster using controller behind feature flag

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Create management cluster using controller behind feature flag.

*Testing (if applicable):*

- Manual testing for docker, vSphere, and Cloudstack.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/createcluster.go
+++ b/cmd/eksctl-anywhere/cmd/createcluster.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/validations/createvalidations"
 	"github.com/aws/eks-anywhere/pkg/workflow/management"
 	"github.com/aws/eks-anywhere/pkg/workflows"
+	newManagementWorkflow "github.com/aws/eks-anywhere/pkg/workflows/management"
 	"github.com/aws/eks-anywhere/pkg/workflows/workload"
 )
 
@@ -269,6 +270,22 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		)
 		err = createWorkloadCluster.Run(ctx, clusterSpec, createValidations)
 
+	} else if clusterSpec.Cluster.IsSelfManaged() && features.UseControllerViaCLIWorkflow().IsActive() {
+
+		logger.Info("Using the new workflow using the controller for management cluster create")
+
+		createMgmtCluster := newManagementWorkflow.NewCreate(
+			deps.Bootstrapper,
+			deps.Provider,
+			deps.ClusterManager,
+			deps.GitOpsFlux,
+			deps.Writer,
+			deps.EksdInstaller,
+			deps.PackageInstaller,
+			deps.ClusterApplier,
+		)
+
+		err = createMgmtCluster.Run(ctx, clusterSpec, createValidations)
 	} else {
 		err = createCluster.Run(ctx, clusterSpec, createValidations, cc.forceClean)
 	}
--- a/pkg/clustermanager/cluster_manager.go
+++ b/pkg/clustermanager/cluster_manager.go
@@ -450,6 +450,46 @@ func (c *ClusterManager) CreateWorkloadCluster(ctx context.Context, managementCl
 	return workloadCluster, nil
 }
 
+// GetWorkloadCluster gets workload cluster.
+func (c *ClusterManager) GetWorkloadCluster(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec, provider providers.Provider) (*types.Cluster, error) {
+	clusterName := clusterSpec.Cluster.Name
+
+	workloadCluster := &types.Cluster{
+		Name: clusterName,
+	}
+
+	logger.V(3).Info("Waiting for workload kubeconfig generation", "cluster", clusterName)
+
+	// Use a buffer to cache the kubeconfig.
+	var buf bytes.Buffer
+
+	if err := c.getWorkloadClusterKubeconfig(ctx, clusterName, managementCluster, &buf); err != nil {
+		return nil, fmt.Errorf("waiting for workload kubeconfig: %v", err)
+	}
+
+	rawKubeconfig := buf.Bytes()
+
+	// The Docker provider wants to update the kubeconfig to patch the server address before
+	// we write it to disk. This is to ensure we can communicate with the cluster even when
+	// hosted inside a Docker Desktop VM.
+	if err := provider.UpdateKubeConfig(&rawKubeconfig, clusterName); err != nil {
+		return nil, err
+	}
+
+	kubeconfigFile, err := c.writer.Write(
+		kubeconfig.FormatWorkloadClusterKubeconfigFilename(clusterName),
+		rawKubeconfig,
+		filewriter.PersistentFile,
+		filewriter.Permission0600,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("writing workload kubeconfig: %v", err)
+	}
+	workloadCluster.KubeconfigFile = kubeconfigFile
+
+	return workloadCluster, nil
+}
+
 func (c *ClusterManager) waitUntilControlPlaneAvailable(
 	ctx context.Context,
 	clusterSpec *cluster.Spec,
@@ -1161,6 +1201,22 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
 		return err
 	}
 	if err = c.ApplyBundles(ctx, clusterSpec, cluster); err != nil {
+		return err
+	}
+	return c.ApplyReleases(ctx, clusterSpec, cluster)
+}
+
+// CreateEKSAReleaseBundle applies the eks-a release bundle to the cluster.
+func (c *ClusterManager) CreateEKSAReleaseBundle(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
+	if clusterSpec.Cluster.Namespace != "" {
+		if err := c.clusterClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, clusterSpec.Cluster.Namespace); err != nil {
+			return err
+		}
+	}
+
+	clusterSpec.Cluster.AddManagedByCLIAnnotation()
+
+	if err := c.ApplyBundles(ctx, clusterSpec, cluster); err != nil {
 		return err
 	}
 	return c.ApplyReleases(ctx, clusterSpec, cluster)
--- a/pkg/workflows/management/create.go
+++ b/pkg/workflows/management/create.go
@@ -1,0 +1,60 @@
+package management
+
+import (
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
+)
+
+// Create is a schema for create cluster.
+type Create struct {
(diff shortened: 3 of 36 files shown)


pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+	commandContext.BootstrapCluster = bootstrapCluster
+
+	logger.Info("Provider specific pre-capi-install-setup on bootstrap cluster")
+	if err = commandContext.Provider.PreCAPIInstallOnBootstrap(ctx, bootstrapCluster, commandContext.ClusterSpec); err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	logger.Info("Installing cluster-api providers on bootstrap cluster")
+	if err = commandContext.ClusterManager.InstallCAPI(ctx, commandContext.ClusterSpec, bootstrapCluster, commandContext.Provider); err != nil {
```
g-gaston, line 57: why is capi in the create bootstrap cluster task? and if we want capi here, why isn't eksa here as well?
trying to understand the reasoning here

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+		return &workflows.CollectDiagnosticsTask{}
+	}
+	err = commandContext.EksdInstaller.InstallEksdManifest(ctx, commandContext.ClusterSpec, commandContext.BootstrapCluster)
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectDiagnosticsTask{}
+	}
+
+	logger.Info("Applying cluster spec to bootstrap cluster")
+	if err := commandContext.ClusterUpgrader.Run(ctx, commandContext.ClusterSpec, *commandContext.BootstrapCluster); err != nil {
```
g-gaston, line 120: why is this here? shouldn't it be in the create cluster task?

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+
+func (s *installEksaComponentsTask) Checkpoint() *task.CompletedTask {
+	return nil
+}
+
+// createWorkloadClusterTask implementation
+
+func (s *createWorkloadClusterTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Creating new workload cluster")
+	workloadCluster, err := commandContext.ClusterManager.GetWorkloadCluster(ctx, commandContext.BootstrapCluster, commandContext.ClusterSpec, commandContext.Provider)
```
g-gaston, line 144: i have my doubts about this method

first i would prefer if we don't implement it in cluster manager
that's a super overloaded entity and tbh, not a good abstraction (I am the one to be blamed there)
It can still be in the clustermanager package if that makes it a bit easier for you, but ideally it should not depend on the clustermanager object.

Second, I wonder if we can just group it with the applier to create a higher level abstraction. Like a `ClusterCreator` or `CreateCluster` (horrible name, feel free to choose a better one). This entity would take the cluster spec and the kubeconfig for the "management cluster" (bootstrap in this occasion) and return the `types.Cluster` object for the newly created workload cluster. It can use the applier internally. It's a similar concept to what you can find here https://github.com/aws/eks-anywhere/blob/main/pkg/workflow/task/workload/create.go (minus the part that ties to the `Task` interface. Alternatively you could have a `CreateCluster` with two methods, `Create` (or `CreateSync`) and `WriteKubeconfig`.

I believe this would also be reusable for the workload cluster flow.

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectDiagnosticsTask{}
+	}
+	commandContext.WorkloadCluster = workloadCluster
+
+	if !commandContext.BootstrapCluster.ExistingManagement {
```
g-gaston, line 151: needed?

pkg/clustermanager/cluster_manager.go
```diff
@@ -1167,6 +1208,37 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
 	return c.ApplyReleases(ctx, clusterSpec, cluster)
 }
 
+// CreateEKSAReleaseBundle applies the eks-a release bundle to the cluster.
+func (c *ClusterManager) CreateEKSAReleaseBundle(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
```
g-gaston, line 1212: do we really need the wrapper?

pkg/clustermanager/cluster_manager.go
```diff
@@ -1167,6 +1208,37 @@ func (c *ClusterManager) CreateEKSAResources(ctx context.Context, cluster *types
...
+	}
+
+	if err := c.ApplyBundles(ctx, clusterSpec, cluster); err != nil {
+		return err
+	}
+	return c.ApplyReleases(ctx, clusterSpec, cluster)
+}
+
+// ApplyEKSASpec applies the eks-a cluster specs (cluster, datacenterconfig, machine configs, etc.).
+func (c *ClusterManager) ApplyEKSASpec(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec,
```
g-gaston, line 1226: do we need this? can't we just use the applier?

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+func (s *installResourcesOnManagementTask) Checkpoint() *task.CompletedTask {
+	return nil
+}
+
+func (s *moveClusterManagementTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	if commandContext.BootstrapCluster.ExistingManagement {
+		return &installEksaComponentsOnWorkloadTask{}
+	}
+	logger.Info("Moving cluster management from bootstrap to workload cluster")
+	err := commandContext.ClusterManager.MoveCAPI(ctx, commandContext.BootstrapCluster, commandContext.WorkloadCluster, commandContext.WorkloadCluster.Name, commandContext.ClusterSpec, types.WithNodeRef())
```
g-gaston, line 219: do we need to pause the eks-a reconciliation before we do this?
if not, you take the risk of the eks-a controller trying to create a new capi cluster

basically do the same thing that clusterctl does to move resources

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+
+type installGitOpsManagerTask struct{}
+
+type moveClusterManagementTask struct{}
+
+type writeClusterConfigTask struct{}
+
+type deleteBootstrapClusterTask struct{}
+
+type installCuratedPackagesTask struct{}
```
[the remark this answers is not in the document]
mitalipaygude: Yes I definitely intend to do that, this one is more like a POC code to agree on the flow.

.gitignore
```diff
@@ -26,3 +26,4 @@ docs/.hugo_build.lock
 *.log
 pkg/executables/TestDeployTemplate*
 pkg/files/config
+*.yaml
```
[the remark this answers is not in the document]
mitalipaygude: this was a local change, didn't mean to add this here.

also, wouldn't this mean that yamls can't be added at the root level? i think we can still add them in other directories inside

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/eks-anywhere/pkg/validations/createvalidations"
 	"github.com/aws/eks-anywhere/pkg/workflow/management"
 	"github.com/aws/eks-anywhere/pkg/workflows"
+	m "github.com/aws/eks-anywhere/pkg/workflows/management"
```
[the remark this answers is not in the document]
mitalipaygude: yeah i will rename the alias to a meaningful one

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -252,6 +253,30 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 		}
 
 		err = wflw.Run(ctx)
+	} else if clusterSpec.Cluster.IsSelfManaged() && features.UseControllerViaCLIWorkflow().IsActive() {
+
+		fmt.Println("Using the new workflow using the controller for management cluster create")
```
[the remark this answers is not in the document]
mitalipaygude: Another POC level log, would mostly remove it in the real flow

pkg/workflows/management/create.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
+)
+
+// CreateManagement is a schema for create cluster.
+type CreateManagement struct {
```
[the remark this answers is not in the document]
mitalipaygude: Yes this will be Create, I named it like this for clear segregation under the feature flag for myself.

pkg/workflows/management/create_cluster.go
```diff
@@ -0,0 +1,379 @@
...
+func (s *createBootStrapClusterTask) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
+	return nil, nil
+}
+
+func (s *createBootStrapClusterTask) Checkpoint() *task.CompletedTask {
+	return nil
+}
+
+func (s *installEksaComponentsTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	if !commandContext.BootstrapCluster.ExistingManagement {
```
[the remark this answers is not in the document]
mitalipaygude: yes I agree, this is the clean up i was referring to actually, but let me remove it right away.

pkg/workflows/management/create_cluster.go
mitalipaygude: So CAPI was always in the create bootstrap cluster task, so I left it there itself. I created a new one below for installing the eksa components tasks, but if you have a strong preference I can combine them all into the create bootstrap cluster task.

pkg/workflows/management/create_cluster.go
mitalipaygude: We can add the pause reconcile annotation here. However, since the order of tasks is that we move CAPI first, and then install EKSA controller, the capi cluster will already exist by the time the controller starts reconciliation, so would it make a difference?

pkg/workflows/management/create_cluster.go
mitalipaygude: https://github.com/orgs/aws/projects/162/views/1?pane=issue&itemId=46266515

pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -61,10 +61,6 @@ func (r *Cluster) ValidateCreate() error {
 
 	var allErrs field.ErrorList
 
-	if !r.IsReconcilePaused() && r.IsSelfManaged() {
```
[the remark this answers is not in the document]
mitalipaygude: https://github.com/aws/eks-anywhere-internal/issues/2097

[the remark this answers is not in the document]
mitalipaygude: https://github.com/aws/eks-anywhere/pull/7210

pkg/clustermanager/cluster_manager.go
mitalipaygude: I ended up using this wrapper because while creating the cluster's namespace we use the cluster Client that is a private Retrier Client used by the cluster manager.

Issue #7114, opened 2023-12-01
Packages issues when using registry mirror


tatlat:
**Issues**:
Registry mirror CA Certificate used in curated package does not update after cluster upgrade with registryMirrorConfig change.
Cannot enable/install curated packages on workload clusters when using a registry mirror

**What happened**:

When updating the caCertContent of a registry mirror configuration of a management cluster, the new certificate was not properly propagated to the packages controller, which led to errors when installing packages due to it using the old certificate. Additionally, packages were not being enabled on newly created workload clusters with registry mirror configs due to the same certificate propagation issue. 

The packages controller gets the CA certificate from 2 secrets: registry-mirror-secret and registry-mirror-cred. The latter gets its data from the former secret. For management clusters, the registry-mirror-secret secret is created in the eksa-packages namespace whereas it is created in `eksa-packages-<clusterName>` for workload clusters. The latter secret is in the eksa-packages namespace and contains the CA certificates for each cluster as `<clusterName>_ca.crt`. After creating clusters named mgmt and worker, the `registry-mirror-cred` secret contained `mgmt_ca.crt` but did not contain `worker_ca.crt`. The `registry-mirror-secret` secrets for both clusters contained the default values as seen below.
```
data:
    CACERTCONTENT: ""
    ENDPOINT: ""
    INSECURE: ZmFsc2UK
    PASSWORD: ""
    USERNAME: ""
```

It seems that registry-mirror-secret receives its data from a helm chart. The EKS-Anywhere controller overrides the default values of the helm chart when enabling packages, which is usually triggered by the CLI when creating a cluster or by the EKS-Anywhere controller as a post reconcile task. However, a new eks-anywhere-packages pod was being rolled out shortly after enabling packages without re-overriding the default values. As a result, `registry-mirror-secret` briefly contains the correct registry data, but loses that data when a new packages pod rolls out. For workload clusters, the `registry-mirror-secret` secret never contains the correct registry data at any point.

Further investigation of the eks-anywhere-packages repo suggests that registry-mirror-secret is only used to fill registry-mirror-cred. 

**What you expected to happen**:

Upgrading a management or workload's cluster CA certificate would also update the certificate used by curated packages. 
To be able to create workload clusters that use registry mirrors and still use curated packages.

**How to reproduce it (as minimally and precisely as possible)**:

Create a registry mirror.
Successfully create a mgmt cluster with correct registry mirror configuration and packages enabled.
[...]

PR #7115, opened 2023-12-01
Fix registry mirror issues in packages

tatlat:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/7114

*Description of changes:*
Changes the registry-mirror-cred secret in eksa-packages namespace directly from EKS-Anywhere controller.

*Testing (if applicable):*


config/rbac_default/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
...
+- patch: |-
+    - op: replace
+      path: /metadata/name
+      value: eksa-packages-role
+  target:
+    group: rbac.authorization.k8s.io
+    kind: Role
+    name: eksa-manager-role
+    namespace: eksa-packages
+    version: v1
```
g-gaston, line 38: Does this mean that every time we add a role for a different namespace we would need to add a couple entries here?

controllers/cluster_controller.go
```diff
@@ -386,6 +388,10 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 		}
 	}
 
+	if err := r.packagesClient.UpdateSecrets(ctx, r.client, cluster); err != nil {
```
g-gaston, line 391: could this be part of the `packagesReconcile` method?

controllers/cluster_controller_test.go
```diff
@@ -367,6 +368,7 @@ func TestClusterReconcilerReconcileConditions(t *testing.T) {
 			log := testr.New(t)
 			logCtx := ctrl.LoggerInto(ctx, log)
 
+			mockPkgs.EXPECT().UpdateSecrets(logCtx, testClient, sameName(config.Cluster))
```
g-gaston, line 371: maybe we should think about adding a helper to group all expectations? most tests apparently don't care about the expectations, just about the result

pkg/clustermanager/cluster_manager.go
```diff
@@ -1137,6 +1137,11 @@ func (c *ClusterManager) CreateEKSANamespace(ctx context.Context, cluster *types
 	return c.clusterClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, constants.EksaSystemNamespace)
 }
 
+// CreatePackagesNamespace creates the eksa-packages namespace on the cluster if it doesn't already exist.
+func (c *ClusterManager) CreatePackagesNamespace(ctx context.Context, cluster *types.Cluster) error {
```
g-gaston, line 1141: curious, why is this needed now?
doesn't the "install packages" functionality (through helm) create this ns?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -236,6 +248,70 @@ func (pc *PackageControllerClient) Enable(ctx context.Context) error {
 	return nil
 }
 
+// UpdateSecrets is used to update the registry-mirror-cred secret used by the packages controller.
+func (pc *PackageControllerClient) UpdateSecrets(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
+	secretName := "registry-mirror-cred"
```
g-gaston, line 253: const?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -236,6 +248,70 @@ func (pc *PackageControllerClient) Enable(ctx context.Context) error {
...
+	secretKey := types.NamespacedName{
+		Namespace: constants.EksaPackagesName,
+		Name:      secretName,
+	}
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: constants.EksaPackagesName,
+		},
```
g-gaston, line 262: why fill the secret if you are just gonna read with the client?

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -236,6 +248,70 @@ func (pc *PackageControllerClient) Enable(ctx context.Context) error {
...
+	}
+	credErr := client.Get(ctx, secretKey, secret)
+	err := fillRegistrySecret(cluster.Name, cluster.Spec.RegistryMirrorConfiguration, secret)
+	if err != nil {
+		return err
+	}
+
+	if apierrors.IsNotFound(credErr) {
```
g-gaston, line 270: if you just want to do an "upsert", you don't need to do a get. Just do the upgrade and if the object doesn't exist, it will return a IsNotFound. Then you can call create

However, if you do a `Get` first, at least check if the object needs to be updated or not. That way we avoid an extra write call. Remember that this will be called on every reconcile loop.

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -236,6 +248,70 @@ func (pc *PackageControllerClient) Enable(ctx context.Context) error {
...
+		return client.Create(ctx, secret)
+	} else if credErr == nil {
+		return client.Update(ctx, secret)
+	}
+	return credErr
+}
+
+func fillRegistrySecret(clusterName string, registry *anywherev1.RegistryMirrorConfiguration, secret *corev1.Secret) error {
```
g-gaston, line 278: how was this secret created before?
I was expecting to have already code that did this. If it's helm, why don't we use helm to update the values as well? this would be duplicated logic in that case
It's possible that I'm not getting this right

pkg/curatedpackages/packagecontrollerclient.go
```diff
@@ -236,6 +248,70 @@ func (pc *PackageControllerClient) Enable(ctx context.Context) error {
...
+
+	dconfig := &dockerConfig{Auths: make(map[string]*dockerAuth)}
+	configData, ok := secret.Data[configName]
+	if ok {
+		err := json.Unmarshal(configData, dconfig)
+		if err != nil {
+			return err
+		}
+	}
+	username, password, err := config.ReadCredentials()
```
g-gaston, line 296: why are we reading from env vars here? The controller should avoid at all cost passing data through env vars

pkg/workflows/create.go
```diff
@@ -252,6 +252,12 @@ func (s *CreateWorkloadClusterTask) Run(ctx context.Context, commandContext *tas
 			return &CollectDiagnosticsTask{}
 		}
 
+		err = commandContext.ClusterManager.CreatePackagesNamespace(ctx, workloadCluster)
```
g-gaston, line 255: why wasn't this needed before?

config/rbac_default/kustomization.yaml
tatlat: Yes, we would need to patch the name of the role for other namespaces since they never get created if there is already another role with the same name in eksa-system.

controllers/cluster_controller.go
tatlat: So one of the issues I found was that the controller's `reconcile` method sometimes returns before it reaches `packagesReconcile` which is why I put it here to ensure the secret gets updated.

pkg/clustermanager/cluster_manager.go
tatlat: We need the namespace before installing packages or else we can't add role/rolebindings in that ns

controllers/cluster_controller.go
g-gaston: I see. And what would be the problem on waiting until reconcile doesn't return early?

pkg/curatedpackages/packagecontrollerclient.go
jiayiwang7: > it will use the image found in packages bundle instead where it rolls out a new pod. Once that pod rolls out with the new image, the secrets are reset to their default values.

why would the secret be reset to default after pod rolls out with image in package bundle? shouldn't we try solving this issue instead?

jiayiwang7:
mind follow up with a PR to remove `registry-mirror-secret` in package controller if it's no longer used anywhere?

PR #7137, opened 2023-12-06
Fix check for existing PRs

abhay-krishna:
Fix check for existing PRs so that older PRs are not missed due to the `gh pr list` command's pagination.


pokearu (commented):
/lgtm
/approve

PR #7139, opened 2023-12-06
Fix controller generating cilium manifests with registry mirror

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
The CNI reconciler fails to generate cilium manifests in an airgapped environment because it tries to fetch the image from public.ecr.aws instead of the registry mirror. This is because the registry mirror for the [helm](https://github.com/aws/eks-anywhere/blob/c4d01fed9ae8030a948a939315f8747d7d53e429/pkg/executables/helm.go#L20) executable reference in the [cilium templater](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/networking/cilium/templater.go#L33) is never set when managing a cluster using FLC, so when generating the manifests helm , [the logic that replaces the host](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/registrymirror/registrymirror.go#L85) in the image uri is skipped

This PR addresses the issue by enabling the controller to handle helm charts when reconciling the CNI is to construct the `Helm` instance in the controller. Instead of depending directly on the Helm, we can inject a HelmFactory to the [cilium templater](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/networking/cilium/templater.go#L33).  We can then use the `HelmFactory` to create an instance of `Helm` with a registry mirror configuration if one is defined on the `Cluster` object.

*Testing (if applicable):*

### DIFF
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -771,7 +771,7 @@ func (f *Factory) WithHelm(opts ...executables.HelmOpt) *Factory {
 	return f
 }
 
-func (f *Factory) WithHelmFactory(opts ...executables.HelmOpt) *Factory {
+func (f *Factory) WithHelmFactory() *Factory {
 	f.WithExecutableBuilder()
 
 	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
--- a/pkg/dependencies/factory.go
+++ b/pkg/dependencies/factory.go
@@ -107,6 +107,7 @@ type Dependencies struct {
 	SnowValidator               *snow.Validator
 	IPValidator                 *validator.IPValidator
 	UnAuthKubectlClient         KubeClients
+	HelmFactory                 *HelmFactory
 	CreateClusterDefaulter      cli.CreateClusterDefaulter
 	UpgradeClusterDefaulter     cli.UpgradeClusterDefaulter
 }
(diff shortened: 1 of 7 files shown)


pkg/networking/cilium/templater.go
```diff
@@ -112,6 +127,20 @@ func WithPolicyAllowedNamespaces(namespaces []string) ManifestOpt {
 	}
 }
 
+func (t *Templater) registryLogin(ctx context.Context, helm Helm, spec *cluster.Spec) error {
```
d8660091, line 130: looks like this "t" is not used, it can be a function instead of a method of Templater

pkg/networking/cilium/templater.go
```diff
@@ -62,7 +68,16 @@ func (t *Templater) GenerateUpgradePreflightManifest(ctx context.Context, spec *
 		return nil, err
 	}
 
-	manifest, err := t.helm.Template(ctx, uri, version, namespace, v, kubeVersion)
+	r := registrymirror.FromCluster(spec.Cluster)
+	helm := t.helmFactory.GetInstance(executables.WithRegistryMirror(r))
+
+	if spec.Cluster.Spec.RegistryMirrorConfiguration != nil {
```
d8660091, line 74: This if check is duplicated in registryLogin function

pkg/dependencies/helm.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func NewHelmFactory(builder ExecutableBuilder) *HelmFactory {
+	return &HelmFactory{
+		builder: builder,
+	}
+}
+
+func (f *HelmFactory) GetInstance(opts ...executables.HelmOpt) *executables.Helm {
+	f.mu.Lock()
```
d8660091, line 51: Curious, do we really need a lock here?

pkg/dependencies/factory.go
```diff
@@ -770,6 +771,24 @@ func (f *Factory) WithHelm(opts ...executables.HelmOpt) *Factory {
 	return f
 }
 
+func (f *Factory) WithHelmFactory(opts ...executables.HelmOpt) *Factory {
+	f.WithExecutableBuilder()
+
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
```
d8660091, line 777: Overall, factory is not really a preferred abstract in golang. As we have adopted it, we should try not making it more complicated.

This method introduces something feels like "nested factory", which makes the abstract harder to understand. I believe we can achieve the target without this "helmFactory"

PR #7144, opened 2023-12-07
Add CP upgrade reconciler logic

rahulbabu95:
Add CP uprgader controller logic that is responsible for scheduling node upgrade objects for all control plane nodes. Tested the  controller by manually applying the CP upgrade object.

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/controllers/controlplaneupgrade_controller.go
+++ b/controllers/controlplaneupgrade_controller.go
@@ -146,9 +146,8 @@ func (r *ControlPlaneUpgradeReconciler) reconcile(ctx context.Context, log logr.
 					return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
 				}
 				return ctrl.Result{}, nil
-			} else {
-				return ctrl.Result{}, fmt.Errorf("getting node upgrader for machine %s: %v", machineRef.Name, err)
 			}
+			return ctrl.Result{}, fmt.Errorf("getting node upgrader for machine %s: %v", machineRef.Name, err)
 		}
 		if !nodeUpgrade.Status.Completed {
 			return ctrl.Result{}, nil
--- a/controllers/controlplaneupgrade_controller.go
+++ b/controllers/controlplaneupgrade_controller.go
@@ -18,23 +18,43 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"time"
 
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+// controlPlaneUpgradeFinalizerName is the finalizer added to NodeUpgrade objects to handle deletion.
+const (
+	controlPlaneUpgradeFinalizerName = "controlplaneupgrades.anywhere.eks.amazonaws.com/finalizer"
+	// TODO(in-place): Fetch these versions dynamically from the bundle instead of using the hardcoded one.
+	kubernetesVersion = "v1.28.3-eks-1-28-9"
+	etcdVersion       = "v3.5.9-eks-1-28-9"
 )
 
 // ControlPlaneUpgradeReconciler reconciles a ControlPlaneUpgradeReconciler object.
 type ControlPlaneUpgradeReconciler struct {
 	client client.Client
+	log    logr.Logger
 }
 
 // NewControlPlaneUpgradeReconciler returns a new instance of ControlPlaneUpgradeReconciler.
 func NewControlPlaneUpgradeReconciler(client client.Client) *ControlPlaneUpgradeReconciler {
 	return &ControlPlaneUpgradeReconciler{
 		client: client,
+		log:    ctrl.Log.WithName("ControlPlaneUpgradeController"),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


controllers/controlplaneupgrade_controller_test.go
```diff
@@ -0,0 +1,179 @@
...
+	r := controllers.NewControlPlaneUpgradeReconciler(client)
+	req := cpUpgradeRequest(cpUpgrade)
+	_, err := r.Reconcile(ctx, req)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	cpu := &anywherev1.ControlPlaneUpgrade{}
+	err = client.Get(ctx, types.NamespacedName{Name: cpUpgrade.Name, Namespace: "eksa-system"}, cpu)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	err = client.Delete(ctx, cpUpgrade)
```
taneyland, line 89: I think we should set the DeletionTimestamp instead of calling delete here

controllers/controlplaneupgrade_controller.go
```diff
@@ -18,42 +18,220 @@ package controllers
...
+			},
+			KubernetesVersion:     kubernetesVersion,
+			EtcdVersion:           &etcdVersion,
+			FirstNodeToBeUpgraded: firstControlPlane,
+		},
+	}
+}
 
-	// TODO(user): your logic here
+func (r *ControlPlaneUpgradeReconciler) reconcileDelete(ctx context.Context, log logr.Logger, cpUpgrade *anywherev1.ControlPlaneUpgrade) (ctrl.Result, error) {
```
taneyland, line 186: Should we actually be deleting the CPUpgrade object here? Instead of Node Upgrader?

controllers/controlplaneupgrade_controller.go
```diff
@@ -18,42 +18,220 @@ package controllers
...
+
+func (r *ControlPlaneUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, cpUpgrade *anywherev1.ControlPlaneUpgrade) (ctrl.Result, error) {
+	// return early if controlplane upgrade is already complete
+	if cpUpgrade.Status.Ready {
+		log.Info("All Control Plane nodes are upgraded")
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("Upgrading all Control Plane nodes")
+	var firstControlPlane bool
```
vivek-koppuru, line 137: put `var` at the top of the func or use `firstControlPlane := false`

controllers/controlplaneupgrade_controller.go
```diff
@@ -18,42 +18,220 @@ package controllers
...
+	for idx, machineRef := range cpUpgrade.Spec.MachinesRequireUpgrade {
+		firstControlPlane = idx == 0
+		nodeUpgrade := nodeUpgrader(machineRef, kubernetesVersion, etcdVersion, firstControlPlane)
+		if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+		}
+		for {
```
vivek-koppuru, line 144: Why do we have a nested for loop here? I think the way to wait is to create the objects once, and then return the controller, and then have a check above to see if the objects were already created before attempting to create.

controllers/controlplaneupgrade_controller.go
```diff
@@ -18,42 +18,217 @@ package controllers
...
+	if !cpUpgrade.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, log, cpUpgrade)
+	}
+	controllerutil.AddFinalizer(cpUpgrade, controlPlaneUpgradeFinalizerName)
+	return r.reconcile(ctx, log, cpUpgrade)
+}
+
+func (r *ControlPlaneUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, cpUpgrade *anywherev1.ControlPlaneUpgrade) (ctrl.Result, error) {
+	// return early if controlplane upgrade is already complete
+	if cpUpgrade.Status.Ready {
```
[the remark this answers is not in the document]
rahulbabu95: isn't this more like for everytime it enters the reconcile loop check if the status is ready then exit early and stop doing more.?

controllers/controlplaneupgrade_controller.go
rahulbabu95: we need the nested for because we want to create the node upgrade one by one... Although it looks like a nested for it doesn't do anything that actually wait for each of the nodeupgrader object to run through it's lifecycle.

controllers/controlplaneupgrade_controller.go
```diff
@@ -57,3 +125,114 @@ func (r *ControlPlaneUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	log.Info("Upgrading all Control Plane nodes")
+
+	for idx, machineRef := range cpUpgrade.Spec.MachinesRequireUpgrade {
+		firstControlPlane = idx == 0
+		nodeUpgrade := nodeUpgrader(machineRef, kubernetesVersion, etcdVersion, firstControlPlane)
+		if err := r.client.Get(ctx, GetNamespacedNameType(nodeUpgraderName(machineRef.Name), constants.EksaSystemNamespace), nodeUpgrade); err != nil {
+			if apierrors.IsNotFound(err) {
+				if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+				}
```
vivek-koppuru, line 146: I think this is still missing a return when the create is successful, to force a requeue right away

controllers/controlplaneupgrade_controller.go
```diff
@@ -57,3 +125,114 @@ func (r *ControlPlaneUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+			}
+		}
+		if !nodeUpgrade.Status.Completed {
+			return ctrl.Result{}, nil
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func patchControlPlaneUpgrade(ctx context.Context, log logr.Logger, patchHelper *patch.Helper, cpUpgrade anywherev1.ControlPlaneUpgrade, patchOpts ...patch.Option) error {
```
vivek-koppuru, line 158: Why is log passed in here? Also, any reason you moved this to it's own function if it only contains one line?

rahulbabu95: right, We don't need it here.!

controllers/controlplaneupgrade_controller.go
```diff
@@ -57,3 +126,111 @@ func (r *ControlPlaneUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	for idx, machineRef := range cpUpgrade.Spec.MachinesRequireUpgrade {
+		firstControlPlane = idx == 0
+		nodeUpgrade := nodeUpgrader(machineRef, kubernetesVersion, etcdVersion, firstControlPlane)
+		if err := r.client.Get(ctx, GetNamespacedNameType(nodeUpgraderName(machineRef.Name), constants.EksaSystemNamespace), nodeUpgrade); err != nil {
+			if apierrors.IsNotFound(err) {
+				if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+				} else {
+					return ctrl.Result{}, nil
+				}
```
vivek-koppuru, line 149:
```suggestion
				}
			        return ctrl.Result{}, nil
```

vivek-koppuru (approved):

controllers/controlplaneupgrade_controller.go
```diff
@@ -57,3 +126,110 @@ func (r *ControlPlaneUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		nodeUpgrade := nodeUpgrader(machineRef, kubernetesVersion, etcdVersion, firstControlPlane)
+		if err := r.client.Get(ctx, GetNamespacedNameType(nodeUpgraderName(machineRef.Name), constants.EksaSystemNamespace), nodeUpgrade); err != nil {
+			if apierrors.IsNotFound(err) {
+				if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+				}
+				return ctrl.Result{}, nil
+			} else {
+				return ctrl.Result{}, fmt.Errorf("getting node upgrader for machine %s: %v", machineRef.Name, err)
+			}
```
vivek-koppuru, line 151:
```suggestion
			}
			return ctrl.Result{}, fmt.Errorf("getting node upgrader for machine %s: %v", machineRef.Name, err)
```

vivek-koppuru (approved):

PR #7145, opened 2023-12-07
Move iam logic to a separate controller

g-gaston:
*Description of changes:*
The cluster controller is becoming quite complex, with a lot of responsibilities all mixed up in the same "sequential chunk".

There are multiple things we can try, this is just one of them: here I'm exploring the idea of having separate controllers that although the all reconcile Cluster objects, the have different responsibilities, leaving the cluster controller responsible of reconciling the actual k8s cluster.

This at the same time goes in favor and against of the idea of "one controller per resource": we have multiple controller reconciling the `Cluster` resource but in fact what they all care about a different resource: one of the cluster child child objects (in this case the `AWSIamConfig`.

Let me what you think.

### DIFF
--- a/controllers/iamconfig_reconciler.go
+++ b/controllers/iamconfig_reconciler.go
@@ -1,0 +1,112 @@
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/cluster-api/util/patch"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/controller"
+	"github.com/aws/eks-anywhere/pkg/controller/handlers"
+)
+
+const IAMConfigClusterFinalizerName = "clusters.anywhere.eks.amazonaws.com/iam-finalizer"
+
+// IAMConfigClusterReconciler reconciles the IAM installation for Clusters.
+type IAMConfigClusterReconciler struct {
+	client     client.Client
+	awsIamAuth AWSIamConfigReconciler
+}
+
+// AWSIamConfigReconciler manages aws-iam-authenticator installation and configuration for an eks-a cluster.
+type AWSIamConfigReconciler interface {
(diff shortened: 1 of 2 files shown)


controllers/iamconfig_reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	defer func() {
+		if err := patchCluster(ctx, patchHelper, cluster); err != nil {
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+	}()
+
+	// AddFinalizer	is idempotent
+	controllerutil.AddFinalizer(cluster, ClusterFinalizerName)
```
d8660091, line 77: should this be IAMConfigClusterFinalizerName?

g-gaston: Sure thing :)

controllers/iamconfig_reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	log := ctrl.LoggerFrom(ctx)
+	cluster := &anywherev1.Cluster{}
+	if err := r.client.Get(ctx, req.NamespacedName, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if cluster.HasAWSIamConfig() {
```
vivek-koppuru, line 61: Shouldn't this condition be the opposite? We want to exit the reconcile if it doesn't have it right?

vivek-koppuru (commented):
I do like this, it's cleaner

controllers/iamconfig_reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	if conditions.IsFalse(cluster, anywherev1.ControlPlaneReadyCondition) {
+		return ctrl.Result{}, nil
+	}
+
+	result, err := r.awsIamAuth.Reconcile(ctx, log, cluster)
+
+	return result.ToCtrlResult(), err
+}
+
+func (r *IAMConfigClusterReconciler) reconcileDelete(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (ctrl.Result, error) {
```
jonathanmeier5, line 104: Is this intentionally reconcileDelete instead of ReconcileDelete? the interface declares a public ReconcileDelete method I believe?

jonathanmeier5:
I like this approach. I've wondered for a little while if it might be time for us to start breaking out different pieces of functionality from the single controller reconciliation control loop.

controllers/iamconfig_reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (r *IAMConfigClusterReconciler) reconcile(ctx context.Context, log logr.Logger, cluster *anywherev1.Cluster) (ctrl.Result, error) {
+	if result, err := r.awsIamAuth.EnsureCASecret(ctx, log, cluster); err != nil {
+		return ctrl.Result{}, err
+	} else if result.Return() {
+		return result.ToCtrlResult(), nil
+	}
+
+	if conditions.IsFalse(cluster, anywherev1.ControlPlaneReadyCondition) {
```
cxbrowne1207, line 95: So with this separation of responsibility between different controllers, are we going to want to add a status to the `Cluster` resource to track the state of these? It doesn't seem necessary, but after it sounds like we'd have separate controller logs to monitor the status on these resources.

PR #7163, opened 2023-12-12
Cleanup E2E Let's Encrypt TXT Records

jonathanmeier5:
*Issue #, if available:*

N/A

*Description of changes:*

Prior to this PR, cert-manager package end to end tests were failing with the following error:

```
Exceeded limit on number of multi-record answer RRSets _acme-challenge.<domain>

```
This PR should permanently fix the issue they were running into.

For context, when cert-manager provisions a certificate signed by Let's Encrypt it needs a DNS provider
to create a TXT record to verify control of the domain.

We use Route53 for this purpose.

However, Route53 has a service limit of 100 records per-domain within a zone.

The cert-manager packages tests were not cleaning up their Route53 entries, and so it was filling up and hitting the service quota.

By deleting the let's encrypt cert and allowing the cert-manager controller to cleanup related resources, we should avoid this problem in future.

*Testing (if applicable):*

Invoked E2E test.

None.

### DIFF
--- a/test/e2e/certmanager.go
+++ b/test/e2e/certmanager.go
@@ -29,6 +29,7 @@ func runCertManagerRemoteClusterInstallSimpleFlow(test *framework.MulticlusterE2
 		packageFile := e.BuildPackageConfigFile(packageName, packagePrefix, EksaPackagesNamespace)
 		test.ManagementCluster.InstallCuratedPackageFile(packageFile, kubeconfig.FromClusterName(test.ManagementCluster.ClusterName))
 		e.VerifyCertManagerPackageInstalled(packagePrefix, EksaPackagesNamespace, cmPackageName, withMgmtClusterSetup(test.ManagementCluster))
+		e.CleanupCerts(withMgmtClusterSetup(test.ManagementCluster))
 		e.DeleteCluster()
 	})
 	time.Sleep(5 * time.Minute)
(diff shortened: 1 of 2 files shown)


test/e2e/certmanager.go
```diff
@@ -29,6 +29,7 @@ func runCertManagerRemoteClusterInstallSimpleFlow(test *framework.MulticlusterE2
 		packageFile := e.BuildPackageConfigFile(packageName, packagePrefix, EksaPackagesNamespace)
 		test.ManagementCluster.InstallCuratedPackageFile(packageFile, kubeconfig.FromClusterName(test.ManagementCluster.ClusterName))
 		e.VerifyCertManagerPackageInstalled(packagePrefix, EksaPackagesNamespace, cmPackageName, withMgmtClusterSetup(test.ManagementCluster))
+		e.CleanupCerts(withMgmtClusterSetup(test.ManagementCluster))
```
cxbrowne1207, line 32: Would the controller be guaranteed to have enough time to clean up the certs before the delete operation? 
Perhaps we need a wait operation before deleting?

jonathanmeier5: we're using a kubectl client ([here](https://github.com/aws/eks-anywhere/blob/35e85705aa594680bb2fde2caeb92b6f2c0b16b1/pkg/executables/kubectl.go#L2287-L2287)), so I think delete should hang until the resource is removed: https://github.com/kubernetes/kubernetes/issues/67479#issuecomment-413873372

Thoughts?

cxbrowne1207: I think the client hangs when there's a finalizer on the resource?

If the controller removes the finalizer only after cleaning up the related resources you mentioned, then yeah this looks fine to me

PR #6766, opened 2023-10-04
Add a RoutingMode option to cniConfig

meyskens:
Fixes  #3182

*Description of changes:* This adds two routing modes for CIlium to use in the CNI configuration. Tunneled keeps the default geneve tunnel mode.
Direct adds a no tunnel mode with native as routing mode

*Documentation added/planned (if applicable):* Added a part explaining the option and linking to Cilium documentation to the `cni.md` file.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #7170, opened 2023-12-13
Fix controller generating cilium manifests with registry mirror

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
The CNI reconciler fails to generate cilium manifests in an airgapped environment because it tries to fetch the image from public.ecr.aws instead of the registry mirror. This is because the registry mirror for the [helm](https://github.com/aws/eks-anywhere/blob/c4d01fed9ae8030a948a939315f8747d7d53e429/pkg/executables/helm.go#L20) executable reference in the [cilium templater](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/networking/cilium/templater.go#L33) is never set when managing a cluster using FLC, so when generating the manifests helm , [the logic that replaces the host](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/registrymirror/registrymirror.go#L85) in the image uri is skipped

This PR addresses the issue by enabling the controller to handle helm charts when reconciling the CNI is to construct the Helm client in the controller. Instead of depending directly on the Helm executable client, we inject a `HelmFactory` to the [cilium templater](https://github.com/aws/eks-anywhere/blob/6f5168b9121b1e8fbb3527119f0ffae83a4ef8bb/pkg/networking/cilium/templater.go#L33). We can then use the `HelmFactory` to create an instance of a `HelmClient` configured for the registry mirror of the cluster.

*Testing (if applicable):*
- Created a new airgap upgrade from latest minor release e2e test for Docker which will test this flow during kindless upgrades. This is only necessarily "airgapped" before the upgrade process. 
- Test manually by creating a docker cluster in a simulated airgapped environment and upgrading from latest minor release where cilium is updated.
- Ran a couple of e2e test with the following scenarios generally testing the change:
   -  Upgrade from latest minor release (CLI and FLC)
   -  Creating Management cluster, creating workload cluster and upgrading workload cluster
   -  Create Management cluster, creating workload cluster with API, and upgrading workload cluster through FLC


pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for _, o := range opts {
+		o(hf)
+	}
+
+	return hf
+}
+
+// GetClient returns a new Helm executeble.
+func (f *HelmFactory) GetClient(opts ...executables.HelmOpt) cluster.Helm {
```
g-gaston, line 72: What is this method for?

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+	var rUsername, rPassword string
+	var err error
+
+	rUsername, rPassword, err = configcli.ReadCredentialsFromSecret(ctx, f.client)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return "", "", err
+	}
+
+	if apierrors.IsNotFound(err) {
+		rUsername, rPassword, err = configcli.ReadCredentials()
```
g-gaston, line 131: i don't get this, sorry
what is this trying to solve?

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+
+	if managmentCluster.RegistryAuth() {
+		rUsername, rPassword, err = f.getClusterRegistryCredentrials(ctx, managmentCluster)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	r := registrymirror.FromCluster(managmentCluster)
+	helm := f.GetClient(executables.WithRegistryMirror(r))
```
g-gaston, line 112: does this work if there is no registry mirror config?

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// HelmFactoryOpt configures a HelmFactory.
+type HelmFactoryOpt func(*HelmFactory)
+
+// HelmFactory is responsible for creating and owning instances of Helm client.
+type HelmFactory struct {
+	client         cluster.Client
+	helmClient     *cluster.HelmClient
+	mu             sync.Mutex
```
g-gaston, line 28: it doesn't seem like you are using this

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+	builder        ExecutableBuilder
+	registryMirror *registrymirror.RegistryMirror
```
g-gaston, line 30: are you using this?

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+	BuildHelmExecutable(...executables.HelmOpt) *executables.Helm
+}
+
+// HelmFactoryOpt configures a HelmFactory.
+type HelmFactoryOpt func(*HelmFactory)
+
+// HelmFactory is responsible for creating and owning instances of Helm client.
+type HelmFactory struct {
+	client         cluster.Client
+	helmClient     *cluster.HelmClient
```
g-gaston, line 27: are you using this?

pkg/networking/cilium/templater.go
```diff
@@ -24,18 +22,19 @@ const (
 	defaultBackOffPeriod = 5 * time.Second
 )
 
-type Helm interface {
-	Template(ctx context.Context, ociURI, version, namespace string, values interface{}, kubeVersion string) ([]byte, error)
-	RegistryLogin(ctx context.Context, registry, username, password string) error
+// HelmFactory builds a helmClient specifically using the managment cluster's registry mirror configuration.
+type HelmFactory interface {
```
g-gaston, line 26: you are describing the implementation here instead of the abstractionm that this interface represents
```suggestion
// HelmClientFactory provides a helm client for a cluster.
type HelmClientFactory interface {
```

you could even call it HelmClientRegistry

pkg/networking/cilium/templater.go
```diff
@@ -24,18 +22,19 @@ const (
...
+	GetClientForCluster(ctx context.Context, clus *anywherev1.Cluster) (*cluster.HelmClient, error)
```
g-gaston, line 27:
```suggestion
	Get(ctx context.Context, clus *anywherev1.Cluster) (*cluster.HelmClient, error)
```

pkg/networking/cilium/templater.go
```diff
@@ -61,8 +60,16 @@ func (t *Templater) GenerateUpgradePreflightManifest(ctx context.Context, spec *
 	if err != nil {
 		return nil, err
 	}
+	helm, err := t.helmFactory.GetClientForCluster(ctx, spec.Cluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get helm client for cluster %s: %v", spec.Cluster.ManagedBy(), err)
```
g-gaston, line 65:
```suggestion
		return nil, fmt.Errorf("getting helm client for cluster %s: %v", spec.Cluster.Name, err)
```

the abstraction should hide the fact that the client is for the management cluster

pkg/networking/cilium/templater.go
```diff
@@ -61,8 +60,16 @@ func (t *Templater) GenerateUpgradePreflightManifest(ctx context.Context, spec *
...
+	}
+
+	if err := helm.RegistryLoginIfNeeded(ctx); err != nil {
```
g-gaston, line 68: could the client factory do this?
i expect it to return a client that is usable, and it should already be able to determine when a login is needed (eg. after building a client with auth)

pkg/networking/cilium/templater.go
```diff
@@ -131,21 +138,17 @@ func (t *Templater) GenerateManifest(ctx context.Context, spec *cluster.Spec, op
...
-				return nil, err
-			}
-			endpoint := net.JoinHostPort(spec.Cluster.Spec.RegistryMirrorConfiguration.Endpoint, spec.Cluster.Spec.RegistryMirrorConfiguration.Port)
-			if err := t.helm.RegistryLogin(ctx, endpoint, username, password); err != nil {
-				return nil, err
-			}
-		}
+	helm, err := t.helmFactory.GetClientForCluster(ctx, spec.Cluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get helm client for cluster %s: %v", spec.Cluster.ManagedBy(), err)
```
g-gaston, line 143: same comment here

test/e2e/airgap.go
```diff
@@ -192,3 +194,30 @@ func runDockerAirgapConfigFlow(test *framework.ClusterE2ETest) {
 	test.DeleteCluster(framework.WithBundlesOverride(bundleReleasePathFromArtifacts))
 	test.ChangeInstanceSecurityGroup(os.Getenv(framework.RegistryMirrorDefaultSecurityGroup))
 }
+
+func runDockerAirgapUpgradeFromReleaseFlow(test *framework.ClusterE2ETest, latestRelease *releasev1.EksARelease, wantVersion anywherev1.KubernetesVersion, clusterOpts ...framework.ClusterE2ETestOpt) {
```
g-gaston, line 198: is there anything specific to docker in this method?

test/e2e/docker_test.go
```diff
@@ -590,6 +590,22 @@ func TestDockerKubernetes127AirgappedRegistryMirrorAndCert(t *testing.T) {
...
+		framework.NewDocker(t),
+		framework.WithClusterFiller(api.WithExternalEtcdTopology(1)),
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube127)),
+		framework.WithAuthenticatedRegistryMirror(constants.DockerProviderName),
+	)
+	runDockerAirgapUpgradeFromReleaseFlow(
+		test,
+		release,
+		v1alpha1.Kube127,
+	)
```
g-gaston, line 606: what kind of upgrade is this running? it doesn't seem to pass any changes

pkg/config/registry.go
```diff
@@ -31,10 +30,9 @@ func ReadCredentials() (username, password string, err error) {
 
 // ReadCredentialsFromSecret reads from Kubernetes secret registry-credentials.
 // Returns the username and password, or error.
-func ReadCredentialsFromSecret(ctx context.Context, client client.Client) (username, password string, err error) {
+func ReadCredentialsFromSecret(ctx context.Context, client cluster.Client) (username, password string, err error) {
```
g-gaston, line 33: what was this change for?

pkg/cluster/helm.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Helm is underlying executable used to perform Helm operations.
+type Helm interface {
+	Template(ctx context.Context, ociURI, version, namespace string, values interface{}, kubeVersion string) ([]byte, error)
+	RegistryLogin(ctx context.Context, registry, username, password string) error
+}
+
+// HelmClient is responsible for perfoming helm registry operations for a cluster.
+type HelmClient struct {
```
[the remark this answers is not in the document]
cxbrowne1207: The `HelmClient` here is a layer around the Helm executable which wraps the necessary information from the `Cluster` and registry credentials required for interacting with it's registry. It's essentially a registry client for a cluster.

The `HelmFactory` creates this and stores it. To do that, it fetches the management cluster and the registry credentials, it encapsulates that information and makes calls to the registry.

Did that help?

pkg/dependencies/factory.go
```diff
@@ -813,7 +840,8 @@ func (f *Factory) WithCNIInstaller(spec *cluster.Spec, provider providers.Provid
 	if spec.Cluster.Spec.ClusterNetwork.CNIConfig.Kindnetd != nil {
 		f.WithKubectl().WithFileReader()
 	} else {
-		f.WithKubectl().WithCiliumTemplater()
+		managmentClusterName := kubeconfig.FromClusterName(spec.Cluster.ManagedBy())
+		f.WithKubectl().WithKubeClientFromKubeconfig(managmentClusterName).WithCiliumTemplater()
```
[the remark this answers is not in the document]
cxbrowne1207: I also mentioned this above, but the CLI also currently uses the `cilium.Templater`. That depends on the `HelmFactory` which now depends on a client. So, this changes creates a `kubernetes.Client` that can be passed when building the `HelmFactory`

Thoughts? maybe there is a better way to get a client in the case of the CLI that I'm missing or alternatives

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
...
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	configcli "github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/executables"
+	"github.com/aws/eks-anywhere/pkg/registrymirror"
+)
+
+// ExecutableBuilder builds the Helm executeble and returns it.
+type ExecutableBuilder interface {
+	BuildHelmExecutable(...executables.HelmOpt) *executables.Helm
```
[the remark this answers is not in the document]
cxbrowne1207: Okay, gotcha. I'll attempt to decouple this.

curious. Why do you think it's not ideal to depend on the executables package? This will make consumers have to depend on it, yes, but when are these dependencies acceptable in your opinion.

pkg/dependencies/helmfactory.go
cxbrowne1207: This is actually suppose to be a helper that builds and returns the `Helm` executeable object. I had forgotten to re-adjust this after some changes. I plan to change this `buildClient`

pkg/dependencies/helmfactory.go
cxbrowne1207: This is related to the client being passed uses the kubeconfig at `<clusterName>/<clusterName>-eks-a-cluster.kubeconfig`. During cluster creation for the CLI, the objects are still in the bootstrap cluster.

pkg/dependencies/helmfactory.go
cxbrowne1207: Ah, yes. My bad, forgot to re-include the use of these after some refactoring.

pkg/dependencies/helmfactory.go
cxbrowne1207: Yes, `registrymirror.FromCluster` would return nil and then the option would simply set it to nil on the `Helm` instance.

pkg/dependencies/helmfactory.go
cxbrowne1207: Yes, this option is set when building in the dependencies factory, mainly for the CLI flow and the reasons give [here](https://github.com/aws/eks-anywhere/pull/7170/files/6a9ada464d5672a4de7598e026c7ca40ce97a83d#r1425706153)

pkg/networking/cilium/templater.go
cxbrowne1207: Okay, thanks I'll refactor this

pkg/networking/cilium/templater.go
cxbrowne1207: Okay, I'll adjust these

pkg/networking/cilium/templater.go
cxbrowne1207: Yeah, I was thinking all the Factory does is return instances of the `HelmClient`. We could certainly do this in the `HelmFactory` as well though. Then I don't think we'd need the `HelmClient` abstraction.

pkg/dependencies/helmfactory.go
cxbrowne1207: This is the instance of the `HelmClient` being returned to the caller of `GetClientForCluster`.

test/e2e/airgap.go
cxbrowne1207: It changes the security group of the EC2 instances it runs in to "airgap" it, these are only run for the docker airgapped tests.

test/e2e/docker_test.go
cxbrowne1207: It's only upgrading from the latest minor release of EKS-A to the current branch. I could add a kubernetes version change or something, but I think this is also enough to test the bundle upgrades.

pkg/cluster/helm.go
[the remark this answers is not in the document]
g-gaston: I would suggest putting this in the `helm` package
The struct can be maybe `ClientForCluster`
Although from what I can see, it doesn't seem like you need the cluster. Only the auth

So maybe you can ditch the cluster field and name this `AuthenticatedClient`

pkg/dependencies/factory.go
```diff
@@ -770,6 +772,30 @@ func (f *Factory) WithHelm(opts ...executables.HelmOpt) *Factory {
...
+func (f *Factory) WithHelmFactory(opts ...HelmFactoryOpt) *Factory {
+	f.WithExecutableBuilder()
+
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.dependencies.HelmFactory != nil {
+			return nil
+		}
+
+		if f.registryMirror != nil {
+			opts = append(opts, WithRegistryMirror(f.registryMirror))
```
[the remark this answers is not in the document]
g-gaston: It seems to me than in the CLI you don't need to build the client in runtime, so you don't really need a client. I think here is where we make use of dependency injection and inject a different `HelmClientFactory` implementation. In this case, your factory can just a dumb pass through that always returns the same client.

That way you keep the templater code common to both cli and controller and you don't pollute the factory interface with options that are only required in one case and don't make much sense in the client factory abstraction.

pkg/dependencies/factory.go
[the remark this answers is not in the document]
g-gaston: got it, i think my answer above can help with this As well

pkg/dependencies/helmfactory.go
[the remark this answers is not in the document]
g-gaston: in general, nowhere
`executables` is in general a implementation detail. It wraps around binaries to offer a go interface to access their functionality. Hence, no other package should depend on it, because no other package needs to know about the fact that the clients are implemented by wrapping binary executions.

pkg/dependencies/helmfactory.go
g-gaston: got it, let's juts make it unexported then

pkg/dependencies/helmfactory.go
g-gaston: ok same as comment above, we prob want to use a different factory for the CLI

pkg/dependencies/helmfactory.go
g-gaston: ok
maybe it's worth doing an if
since this assumes that creating a client with a nil registr5y mirror is valid
but kind of nit

pkg/dependencies/helmfactory.go
g-gaston: yeah, let's finish the discussion there and then come back if necessary

pkg/dependencies/helmfactory.go
g-gaston: yeah but you are setting it and never reading it

test/e2e/airgap.go
g-gaston: ahh i see
does this mean you couldn't execute this method in another test using a different provider?

test/e2e/docker_test.go
g-gaston: ahhh got you no that works
if we are not passing spec changes, then let's remove the options in the runDockerAirgapUpgradeFromReleaseFlow method

test/e2e/airgap.go
cxbrowne1207: You probably can assuming the other providers use an EC2 instance as an admin machine. Though, unlike the other providers the Docker clusters are created on the admin machine, so this makes sense only for this provider

pkg/dependencies/factory.go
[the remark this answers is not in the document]
cxbrowne1207: Okay got it. I like that better.

Though, I think in the different `HelmClientFactory` implementation will have to know how to log in as well (since we're also we moving the registry login call to the factory). So, CLI would still be reading the environment for registry mirror details.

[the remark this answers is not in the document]
cxbrowne1207: > Okay got it. I like that better.
> 
> Though, it seems to me that `HelmClientFactory` implementation will have to know how to log in as well (since we're also we moving the registry login call to the factory). So, CLI would still be reading the environment for registry mirror details.

pkg/cluster/helm.go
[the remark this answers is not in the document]
cxbrowne1207: Chose to get rid of this abstraction altogether. It became very minimal and was no longer needed.

pkg/config/registry.go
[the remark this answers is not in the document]
cxbrowne1207: I've reverted this change since adding the different `HelmClientFactory` implementation.

I originally changed this because I needed the type in the `HelmClientFactory` that would eventually use this to support multiple client types`client.Client` and a `kubernetes.Client`.

pkg/dependencies/factory.go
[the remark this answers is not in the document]
cxbrowne1207: This was removed now with the addition of a new implementation for `cilium.HelmClientFactory`

pkg/dependencies/helmfactory.go
cxbrowne1207: This was removed after addressing the issues mentioned in the linked thread.

pkg/dependencies/helmfactory.go
```diff
@@ -0,0 +1,138 @@
+package dependencies
```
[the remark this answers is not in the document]
cxbrowne1207: Done.

pkg/dependencies/helmfactory.go
cxbrowne1207: Done

pkg/dependencies/helmfactory.go
cxbrowne1207: Moved the login to the `HelmClientFactory`, and now it uses it before returning it to the consumer.

pkg/networking/cilium/templater.go
cxbrowne1207: Done

pkg/config/registry.go
cxbrowne1207: This was reverted.

It was for allowing this method to be compatible with `client.Client` (for the controller) and `kubernetes.Client` (for the CLI) that was being passed to the HelmClientFactory. However, we no longer have this problem after adding the difference `HelmClientFactory` implementation for the CLI flow.

pkg/dependencies/helmfactory.go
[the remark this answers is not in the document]
cxbrowne1207: I've decoupled the `executables` dependency from this package by reversing the dependency and changing the `BuildHelmExecutable` method to implement an interface from the helm package:

```
func (b *ExecutablesBuilder) BuildHelmExecutable(opts ...HelmOpt) *Helm
```
to

```
func (b *ExecutablesBuilder) BuildHelmExecutable(opts ...helm.Opt) helm.ExecuteableClient
```

It was a little messy and had to tough other parts of the code to make the change

controllers/factory.go
```diff
@@ -465,8 +465,16 @@ func (f *Factory) withCloudStackValidatorRegistry() *Factory {
 	return f
 }
 
+func (f *Factory) withCiliumTemplater() *Factory {
+	f.dependencyFactory.
+		WithHelmClientFactory(f.manager.GetClient()).
+		WithCiliumTemplater()
```
g-gaston, line 471: i would prefer if we didn't trust the execution order to overwrite the default client factory

what we can do here is to have WithHelmClientFactory and WithCiliumTemplater defined in this controllers factory, so we can have our own logic

pkg/executables/helm.go
```diff
@@ -18,49 +18,28 @@ const (
...
 type Helm struct {
-	executable     Executable
-	registryMirror *registrymirror.RegistryMirror
-	env            map[string]string
-	insecure       bool
+	executable  Executable
+	helmOptions *helm.Config // Embed HelmOptions in Helm struct
 }
 
 type HelmOpt func(*Helm)
```
g-gaston, line 25: delete?

pkg/executables/helm.go
```diff
@@ -76,7 +55,7 @@ func (h *Helm) Template(ctx context.Context, ociURI, version, namespace string,
 	params = h.addInsecureFlagIfProvided(params)
 	params = append(params, "-f", "-")
 
-	result, err := h.executable.Command(ctx, params...).WithStdIn(valuesYaml).WithEnvVars(h.env).Run()
+	result, err := h.executable.Command(ctx, params...).WithStdIn(valuesYaml).WithEnvVars(h.helmOptions.Env).Run()
```
g-gaston, line 58: imo env is an option that should be part of this helm struct and not the helm.Options, since it only applies to this particular client implementation

g-gaston: Is this even being used?

pkg/executables/helm.go
```diff
@@ -235,9 +215,9 @@ func (h *Helm) UpgradeChartWithValuesFile(ctx context.Context, chart, ociURI, ve
 		"--wait",
 	}
 	for _, opt := range opts {
-		opt(h)
+		opt(h.helmOptions)
```
g-gaston, line 218: wow is this updating the receiver? it seems like this is how it used to be before, which is not great. Do you mind creating a ticket or something to track this and come back and change it? this is not thread safe

pkg/helm/envclient_factory.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Get returns the helm registry client.
+// The parameters here are not used and it only returns the client that is initialized using Init.
+func (f *EnvClientFactory) Get(_ context.Context, _ *anywherev1.Cluster) (RegistryClient, error) {
+	return f.helmClient, nil
+}
+
+// Init builds the helm registry client once using the registry mirror information from the cluster information.
+// It should be called at least once first, before trying to retrieving and using the client using Get.
+// It only builds the helm registry client once.
```
g-gaston, line 37: Let's add a comment saying this is not thread safe and the caller should guarantee that it doesn't get called from multiple threads

pkg/helm/envclient_factory.go
```diff
@@ -0,0 +1,63 @@
...
+	opts = append(opts, WithRegistryMirror(r))
+
+	helmClient := f.buildClient(opts...)
+
+	if r == nil || r != nil && !r.Auth {
+		f.helmClient = helmClient
+		return nil
+	}
+
+	rUsername, rPassword, err := configcli.ReadCredentials()
```
g-gaston, line 52: let's add a comment saying these should be injected on construction or passed to this method as arguments

pkg/helm/factory.go
```diff
@@ -0,0 +1,125 @@
...
+	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	configcli "github.com/aws/eks-anywhere/pkg/config"
+	"github.com/aws/eks-anywhere/pkg/controller/clusters"
+	"github.com/aws/eks-anywhere/pkg/registrymirror"
+)
+
+// Config contains configuration options for Helm.
+type Config struct {
+	RegistryMirror *registrymirror.RegistryMirror
+	Env            map[string]string
```
g-gaston, line 18: yeah as I said before this probably belongs to the helm executable

pkg/helm/factory.go
```diff
@@ -0,0 +1,125 @@
...
+	Delete(ctx context.Context, kubeconfigFilePath, installName, namespace string) error
+	UpgradeChartWithValuesFile(ctx context.Context, chart, ociURI, version, kubeconfigFilePath, valuesFilePath string, opts ...Opt) error
+	InstallChartWithValuesFile(ctx context.Context, chart, ociURI, version, kubeconfigFilePath, valuesFilePath string) error
+	InstallChart(ctx context.Context, chart, ociURI, version, kubeconfigFilePath, namespace, valueFilePath string, skipCRDs bool, values []string) error
+	Template(ctx context.Context, ociURI, version, namespace string, values interface{}, kubeVersion string) ([]byte, error)
+	RegistryLogin(ctx context.Context, registry, username, password string) error
+}
+
+// RegistryClient represents a Helm client.
+type RegistryClient interface {
```
g-gaston, line 40: confused what this is for?

pkg/helm/factory.go
```diff
@@ -0,0 +1,125 @@
...
+// ClientFactory provides a helm client for a cluster.
+type ClientFactory struct {
+	client     client.Client
+	helmClient RegistryClient
+	mu         sync.Mutex
+	builder    ExecutableBuilder
+}
+
+// NewClientFactory returns a new helm ClientFactory.
+func NewClientFactory(client client.Client, builder ExecutableBuilder) *ClientFactory {
```
g-gaston, line 83:
```suggestion
func NewClientForClusterFactory(client client.Client, builder ExecutableBuilder) *ClientFactory {
```
just to maker the diff with the other factory clear?

pkg/helm/factory.go
```diff
@@ -0,0 +1,125 @@
...
+		builder: builder,
+		mu:      sync.Mutex{},
+	}
+	return hf
+}
+
+// Get returns a new Helm client configured using information from the provided cluster's management cluster.
+func (f *ClientFactory) Get(ctx context.Context, clus *anywherev1.Cluster) (RegistryClient, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
g-gaston, line 95: what is this for?
I'm not sure what we gain by storing the client in the factory if you always generate a new one on every call no matter what

pkg/helm/factory.go
```diff
@@ -0,0 +1,125 @@
...
+	RegistryMirror *registrymirror.RegistryMirror
+	Env            map[string]string
+	Insecure       bool
+}
+
+// Opt is a functional option for configuring Helm behavior.
+type Opt func(*Config)
+
+// ExecuteableClient represents a Helm client.
+type ExecuteableClient interface {
```
g-gaston, line 26:
```suggestion
// Client is a Helm client.
type Client interface {
```

pkg/networking/cilium/templater.go
```diff
@@ -24,18 +23,19 @@ const (
 	defaultBackOffPeriod = 5 * time.Second
 )
 
-type Helm interface {
-	Template(ctx context.Context, ociURI, version, namespace string, values interface{}, kubeVersion string) ([]byte, error)
-	RegistryLogin(ctx context.Context, registry, username, password string) error
+// HelmClientFactory provides a helm client for a cluster.
+type HelmClientFactory interface {
+	Get(ctx context.Context, clus *anywherev1.Cluster) (helm.RegistryClient, error)
```
g-gaston, line 28: why can't we just return `helm.Client` here and avoid the second interface?

pkg/executables/helm.go
cxbrowne1207: Yeah, I'll create the ticket and address it right after

pkg/helm/factory.go
cxbrowne1207: I mainly created this interface to scope down the methods available to the client returned by the factory, as only `RegistryLogin` and `Template` is called on it at some point.

`ExecuteableClient` is more general interface that has more methods due to the other consumers of `BuildHelmExecutable` 

If we don't need it though, I can remove.

pkg/networking/cilium/templater.go
cxbrowne1207: We can. I mentioned this above as well, so I mainly created this interface to scope down the methods available to the client returned by the factory, as only `RegistryLogin` and `Template` is called on it at some point.

ExecuteableClient is more general interface that has more methods due to the other consumers of BuildHelmExecutable

Though, we can remove it if it's not necessary

pkg/helm/factory.go
g-gaston: Talked offline, we are going with one interface for now

pkg/executables/helm.go
cxbrowne1207: Done

https://github.com/aws/eks-anywhere/issues/7176

pkg/helm/factory.go
cxbrowne1207: Yeah, we probably don't need / want to rebuild the executaeble every timer. That was the intention for this. 
So,, I’ll add a base registry comparison here

pkg/executables/helm.go
cxbrowne1207: Discussed offline. We'll make it a bit more implementation agnostic by using a `helm.WithProxyConfig` and then so the executable will turn the struct into `env`

pkg/executables/helm.go
```diff
@@ -18,54 +18,45 @@ const (
...
+		env:        joinEnv(defaultEnv, helmConfig.ProxyConfig),
 	}
 
 	return h
 }
 
+// joinEnv joins the default and the provided maps together, then return the
+// new map.
+func joinEnv(oldEnv, newEnv map[string]string) map[string]string {
+	env := make(map[string]string)
```
g-gaston, line 50: nit: why do we need a new map? could we reuse one of them and have a "mergeMap"?

pkg/helm/envclient_factory.go
```diff
@@ -0,0 +1,66 @@
...
+
+	helmClient := f.buildClient(opts...)
+
+	if r == nil || r != nil && !r.Auth {
+		f.helmClient = helmClient
+		return nil
+	}
+
+	// The registry credentials should be injected on construction through environment variabless REGISTRY_USERNAME
+	// and REGISTRY_PASSWORD, or passed to this method as arguments.
```
g-gaston, line 54:
```suggestion
	// TODO(cxbrowne1207): The registry credentials should be injected on construction through environment variabless REGISTRY_USERNAME
	// and REGISTRY_PASSWORD, or passed to this method as arguments.
```

maybe add link to issue?

pkg/helm/factory.go
```diff
@@ -0,0 +1,126 @@
...
+	RegistryMirror *registrymirror.RegistryMirror
+	ProxyConfig    map[string]string
+	Insecure       bool
+}
+
+// Opt is a functional option for configuring Helm behavior.
+type Opt func(*Config)
+
+// Client represents a Helm client.
+type Client interface {
```
g-gaston, line 27: nit: can we make this file smaller by moving stuff out?
we have polluted a bit the responsability
maybe have a client.go and a config.go

pkg/helm/factory.go
```diff
@@ -0,0 +1,126 @@
...
+// WithInsecure configures helm to skip validating TLS certificates when
+// communicating with the Kubernetes API.
+func WithInsecure() Opt {
+	return func(c *Config) {
+		c.Insecure = true
+	}
+}
+
+// WithProxyConfig sets the proxy configurations on the Config.
+func WithProxyConfig(proxyConfig map[string]string) Opt {
```
g-gaston, line 56: why isn't proxyConfig a struct?

pkg/helm/factory.go
```diff
@@ -0,0 +1,126 @@
...
+	return func(c *Config) {
+		c.ProxyConfig = proxyConfig
+	}
+}
+
+// ExecutableBuilder builds the Helm executable and returns it.
+type ExecutableBuilder interface {
+	BuildHelmExecutable(...Opt) Client
+}
```
g-gaston, line 65:
```suggestion
// ClientBuilder builds a Helm Client.
type ExecutableBuilder interface {
	Build(...Opt) Client
}
```

pkg/helm/factory.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// There is no need to rebuild the client everytime,
+	// here we check if the registry has changed and rebuild the helm client in that case
+	// Rebuild the Helm client only if the registry has changed
+	if f.helmClient == nil || !reflect.DeepEqual(f.registryMirror, r) {
+		f.helmClient = f.builder.BuildHelmExecutable(WithRegistryMirror(r), WithInsecure())
+		f.registryMirror = r
+	}
+
+	if r != nil && managmentCluster.RegistryAuth() {
```
g-gaston, line 119: can we avoid the extra logging if the client was already valid?

TBH, i'm not sure if all this complexity is worth it
With the current code, the only thing you avoid is to build a new struct. You still have to read a cluster object, a secret and perform the login call. Is it worth all the stuff to make it thread safe? Or should we simplify this for now and always build a new client? You should be able to optimize this in the future pretty easily since you have the logic pretty well encapsulated in the factory
You could even add a new factory that provides caching.

pkg/helm/factory.go
```diff
@@ -0,0 +1,126 @@
...
+			return nil, err
+		}
+	}
+
+	r := registrymirror.FromCluster(managmentCluster)
+
+	// There is no need to rebuild the client everytime,
+	// here we check if the registry has changed and rebuild the helm client in that case
+	// Rebuild the Helm client only if the registry has changed
+	if f.helmClient == nil || !reflect.DeepEqual(f.registryMirror, r) {
```
g-gaston, line 114: i advise against using deepequal in actual code, for tests is ok
is it difficult to write an `Equal` method for `registryMirror`?

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
```diff
@@ -66,6 +66,8 @@ spec:
                       in the cluster
                     properties:
                       cilium:
+                        description: CiliumConfig contains configuration specific
```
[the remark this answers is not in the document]
cxbrowne1207: This was updated in [this PR](https://github.com/aws/eks-anywhere/pull/6766). It seems like the generate-files-presubmit didn't catch it. 

Added an issue to track it [here](https://github.com/aws/eks-anywhere-internal/issues/2114)

pkg/helm/factory.go
cxbrowne1207: In the factory, proxyConfig is fetched from the env and handled as a map.
Did we want to change it so the factory passes a struct to it instead?

pkg/helm/factory.go
cxbrowne1207: We could avoid by moving the login into the if, but you have a good point. It may not be worth it since we still have to read the cluster object anyway

We can probably revisit if necessary for optimization. Does simplifying and removing the locking mechanism sound good to you?

pkg/helm/factory.go
cxbrowne1207: It's not difficult to write an equal method for `registryMirror`.

However, this is related to the thread [here](https://github.com/aws/eks-anywhere/pull/7170#discussion_r1428182609), so it may be removed

pkg/helm/factory.go
cxbrowne1207: `Build` is not a method defined by the `ExecutablesBuilder` [implementation](https://github.com/aws/eks-anywhere/blob/2a15b1bfcee63552454d5906b7fd625e224b34e2/pkg/executables/builder.go#L70) used in the dependency factory.

pkg/helm/factory.go
cxbrowne1207: Agreed to remove it offline

pkg/helm/envclient_factory.go
cxbrowne1207: Done

pkg/helm/factory.go
cxbrowne1207: Discussed offline. We'll add a comment describing the valid keys for `WithProxyConfig`.

pkg/helm/factory.go
cxbrowne1207: Discussed offline, we'll call it `BuildHelm`  instead
and add a new method that just forwards to the existing one.

pkg/helm/factory.go
cxbrowne1207: We decided to remove this along with the client caching and locking mechanism.

pkg/executables/helm.go
cxbrowne1207: I think it's better to avoid modifying the inputs directly here and instead return a new one. I think this method does what a `mergeMap` would do.

pkg/dependencies/factory.go
```diff
@@ -107,6 +108,8 @@ type Dependencies struct {
 	SnowValidator               *snow.Validator
 	IPValidator                 *validator.IPValidator
 	UnAuthKubectlClient         KubeClients
+	HelmClientFactory           cilium.HelmClientFactory
+	HelmExecutableBuilder       helm.ClientBuilder
```
g-gaston, line 112: nit: ideally these field would reference the implementation, not the interface

pkg/dependencies/factory.go
```diff
@@ -73,7 +74,7 @@ type Dependencies struct {
 	Clusterctl                  *executables.Clusterctl
 	Flux                        *executables.Flux
 	Troubleshoot                *executables.Troubleshoot
-	Helm                        *executables.Helm
+	Helm                        helm.Client
```
g-gaston, line 77: same here referencing the implementation

pkg/helm/factory.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ClientBuilder builds a Helm Client.
+type ClientBuilder interface {
+	BuildHelm(...Opt) Client
+}
+
+// ClientFactory provides a helm client for a cluster.
+type ClientFactory struct {
+	client     client.Client
+	helmClient Client
```
g-gaston, line 22: we don't need this anymore

pkg/executables/helm.go
g-gaston: not sure what the benefit is if you end map discarding both inputs in the caller
but ok, then at least allocate the memory in advance?

config/crd/bases/anywhere.eks.amazonaws.com_clusters.yaml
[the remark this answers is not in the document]
g-gaston: got you
can you add to the ticket something like: we should add a comment to the Cilium field so the descriptions matches the field name

g-gaston (commented):
/lgtm

pkg/executables/helm.go
cxbrowne1207: Okay, I've considered it and after thinking about the additional code required to allocate the memory in advance, I decided to go with the original suggestion. Previously, I was mainly concerned that someone using this could have unexpected results due to the side effects on the input. Since this an internal function, there doesn't seem to be that risk.

PR #7023, opened 2023-11-10
Add bare metal OS requirements

chrisdoherty4:
Also update the NPM arg so we can run `make server`.

PR #7178, opened 2023-12-15
Advocate using `etcdutl` instead of `etcdctl` for backup and restore

abhay-krishna:
The v3.5.0 release of etcd saw the introduction of a new command-line tool called `etcdutl` that took over the snapshot/restore/backup functionalities of `etcdctl`. Hence, from etcd version v3.5.0, using `etcdctl` for creating and restoring snapshots is deprecated and using this command displays a deprecation warning which suggests to use `etcdutl` tool instead. This PR updates the docs to do the same based on etcd version. The `etcdutl` snapshot and restore commands are essentially drop-in replacements for their `etcdctl` counterparts.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -50,7 +50,7 @@ deploy: zip ## Deploy docs to Amplify
 	bash ./deploy-docs.sh public.zip
 
 serve: submodule build ## Boot the development server.
-	hugo server --buildFuture --baseUrl http://127.0.0.1
+	hugo server --buildFuture --baseURL http://127.0.0.1
 
 server: serve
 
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -50,7 +50,7 @@ deploy: zip ## Deploy docs to Amplify
 	bash ./deploy-docs.sh public.zip
 
 serve: submodule build ## Boot the development server.
-	hugo server --buildFuture --baseUrl http://127.0.0.1
+	hugo server --buildFuture --baseURL http://127.0.0.1
```
chrisnegus, line 53: This change is in chrisdoherty4's PR: https://github.com/aws/eks-anywhere/pull/7023/files. Probably doesn't matter. And you did find it twice.

chrisnegus (commented):
I just had one unnecessary comment. The tabs you used seem to work fine. As long as everyone agrees with the change of tools, it looks good to me.

jiayiwang7:
nice, there are a few other places in troubleshooting/cert renew doc that uses etcdctl too. Mind changing them as well?

docs/Makefile
abhay-krishna: Yeah this had been bothering me for a while, I thought I'd include the change in my next docs PR and here it is.

PR #7186, opened 2023-12-15
Parallelize I/O bound operations in the release process

abhay-krishna:
The release process performs a bunch of operations such as generating the artifacts list to be "released", downloading the artifacts, renaming them and uploading them to their destination. Each of these operations is performed in a for-loop that iterates over the artifacts list and "processes" (downloads or renames or uploads) them sequentially. But since the individual artifacts don't have dependencies among themselves, they can be processed in parallel within the scope of a single operation. This PR changes the loops to use goroutines to handle these operations in parallel. Since these are mostly I/O bound operations, running them concurrently can save a lot of time. These changes yielded a **60% reduction in job time (15 min to 6 min)** during testing.

Implementation changes:
* An individual `errgroup.Group` has been used for each operation to manage the goroutines. This is quite similar to a WaitGroup ([this article](https://medium.com/@yardenlaif/go-sync-or-go-home-errgroup-f91a0ee72d3f) explores the slight differences with examples) but is better for graceful error handling. Looping over the artifact list, the errgroup creates goroutines for processing each individual artifact and waits for all of them to finish and fast-fails if any of them errors out. Errors ,if any, are propagated to the main goroutine, which preserves the current behavior of the sequential release process.
* The artifacts table and image digests table, which were originally implemented as a map of project name to artifacts and image URIs to image digest respectively, have now been implemented as `sync.Map`, a concurrent map implementation. This is because these tables get used across the code by several goroutines for both read and write operations. `sync.Map` compromises type safety (which is why several files have type assertions on the data returned from the map) for concurrency, and since the release process is deterministic when it comes to assigning values and how each artifact is processed, we can assert the types as and when required, and reap the benefits of concurrent map operations.

⚠️ Note: It is to be noted that the operations (download, upload, etc) themselves don't run in parallel due to ordering (i.e., an artifact cannot be uploaded to destination before it is downloaded from source)


release/cli/pkg/helm/helm.go
```diff
@@ -466,19 +466,17 @@ func OverWriteChartValuesImageSha(filename string, shaMap map[string]anywherev1a
...
-				if artifact.Image.AssetName == "eks-anywhere-packages" || artifact.Image.AssetName == "ecr-token-refresher" {
-					m[artifact.Image.AssetName] = artifact.Image.ReleaseImageURI
-				}
+				m[artifact.Image.AssetName] = artifact.Image.ReleaseImageURI
 			}
 		}
 	}
 	if len(m) == 0 {
-		return nil, fmt.Errorf("No assets found for eks-anywhere-packages, or ecr-token-refresher in eksArtifacts")
+		return nil, fmt.Errorf("No assets found for eks-anywhere-packages, or ecr-token-refresher in packagesArtifacts")
```
cxbrowne1207, line 479: Should this just be `eks-anywhere-packages`? Assuming I'm understanding that ecr-token-refresher was removed from the artifacts earlier.

release/cli/pkg/operations/bundle_release.go
```diff
@@ -146,3 +149,22 @@ func GenerateBundleSpec(r *releasetypes.ReleaseConfig, bundle *anywherev1alpha1.
 	fmt.Printf("%s Successfully generated bundle manifest spec\n", constants.SuccessIcon)
 	return nil
 }
+
+func getImageDigest(_ context.Context, r *releasetypes.ReleaseConfig, artifact releasetypes.Artifact) (string, error) {
```
cxbrowne1207, line 153: Is the `context.Context` param here needed?

release/cli/pkg/operations/download.go
```diff
@@ -44,139 +47,162 @@ func DownloadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]
...
-							objectShasumFileKey = filepath.Join(latestSourceS3PrefixFromMain, objectShasumFileName)
-						} else {
-							return fmt.Errorf("retries exhausted waiting for checksum file to be uploaded to source location: %v", err)
-						}
-					}
+	return nil
+}
 
-					err = s3.DownloadFile(objectShasumFileLocalFilePath, r.SourceBucket, objectShasumFileKey)
+func handleArchiveDownload(_ context.Context, r *releasetypes.ReleaseConfig, artifact releasetypes.Artifact, s3Retrier *retrier.Retrier) error {
```
cxbrowne1207, line 80: Is the `context.Context` param here being used?

release/cli/pkg/bundles/bottlerocket.go
```diff
@@ -24,7 +25,7 @@ import (
 	releasetypes "github.com/aws/eks-anywhere/release/cli/pkg/types"
 )
 
-func GetBottlerocketHostContainersBundle(r *releasetypes.ReleaseConfig, eksDReleaseChannel string, imageDigests map[string]string) (anywherev1alpha1.BottlerocketHostContainersBundle, error) {
+func GetBottlerocketHostContainersBundle(r *releasetypes.ReleaseConfig, eksDReleaseChannel string, imageDigests sync.Map) (anywherev1alpha1.BottlerocketHostContainersBundle, error) {
```
cxbrowne1207, line 28: Can we pass an interface with the `Load` method here instead of `sync.Map`? I think we can create a struct that implements it, and the type assertion can be done there.

release/cli/pkg/bundles/bottlerocket.go
```diff
@@ -61,23 +67,30 @@ func bottlerocketDefaultArtifact(r *releasetypes.ReleaseConfig, metadataFile, im
 	}, nil
 }
 
-func bottlerocketKubeadmBootstrapArtifact(r *releasetypes.ReleaseConfig, eksDReleaseChannel string, imageDigests map[string]string) anywherev1alpha1.Image {
-	artifacts := r.BundleArtifactsTable[fmt.Sprintf("bottlerocket-bootstrap-%s", eksDReleaseChannel)]
+func bottlerocketKubeadmBootstrapArtifact(r *releasetypes.ReleaseConfig, eksDReleaseChannel string, imageDigests sync.Map) (anywherev1alpha1.Image, error) {
```
cxbrowne1207, line 70: Same comment here (and the others as well) about passing an interface instead of `sync.Map`

release/cli/pkg/operations/download.go
```diff
@@ -44,139 +47,162 @@ func DownloadArtifacts(r *releasetypes.ReleaseConfig, eksArtifacts map[string][]
...
+				return errors.Cause(err)
 			}
 		}
 	}
-	fmt.Printf("%s Successfully downloaded artifacts\n", constants.SuccessIcon)
+
+	return nil
+}
+
+func handleManifestDownload(_ context.Context, r *releasetypes.ReleaseConfig, artifact releasetypes.Artifact, s3Retrier *retrier.Retrier) error {
```
cxbrowne1207, line 174: Is the context.Context param here being used? I'm wondering if these can be removed.

release/cli/pkg/assets/manifests/manifests.go
```diff
@@ -59,7 +59,7 @@ func GetManifestAssets(rc *releasetypes.ReleaseConfig, manifestComponent *assett
 	manifestArtifact := &releasetypes.ManifestArtifact{
 		SourceS3Key:       manifestFile,
 		SourceS3Prefix:    sourceS3Prefix,
-		ArtifactPath:      filepath.Join(rc.ArtifactDir, fmt.Sprintf("%s-manifests", projectName), rc.BuildRepoHead),
+		ArtifactPath:      filepath.Join(rc.ArtifactDir, fmt.Sprintf("%s-manifests", componentName), rc.BuildRepoHead),
```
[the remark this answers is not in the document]
abhay-krishna: This is just the local path where the manifests get downloaded on disk. Using the project name as the name of the manifests folder caused issues with CAPI as there are 3 sets of manifests (core-cluster-api, kubeadm-bootstrap and kubeadm-control-plane) which were essentially being stored in the same folder. Doing this sequentially didnt cause issues but with the parallel execution changes introduced in this PR, some files ended up getting overwritten.  Changing to component name ensures they all get stored in different directories and don't get overwritten.

release/cli/pkg/assets/assets.go
```diff
@@ -30,10 +34,9 @@ import (
 	"github.com/aws/eks-anywhere/release/cli/pkg/filereader"
 	releasetypes "github.com/aws/eks-anywhere/release/cli/pkg/types"
 	bundleutils "github.com/aws/eks-anywhere/release/cli/pkg/util/bundles"
-	sliceutils "github.com/aws/eks-anywhere/release/cli/pkg/util/slices"
 )
 
-func getAssetsFromConfig(ac *assettypes.AssetConfig, rc *releasetypes.ReleaseConfig, eksDReleaseChannel, eksDReleaseNumber, kubeVersion string) ([]releasetypes.Artifact, error) {
+func getAssetsFromConfig(_ context.Context, ac *assettypes.AssetConfig, rc *releasetypes.ReleaseConfig, eksDReleaseChannel, eksDReleaseNumber, kubeVersion string) ([]releasetypes.Artifact, error) {
```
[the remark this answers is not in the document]
abhay-krishna: The context param is not being used but it helps with cancellations. In case of `errgroup.WithContext`, the context is cancelled the first time this function returns an error or after all the goroutines have successfully run, which is indicated by a non-nill error from `errgroup.Wait()` method.

release/cli/pkg/bundles/package-controller.go
```diff
@@ -31,13 +32,17 @@ import (
 	"github.com/aws/eks-anywhere/release/cli/pkg/version"
 )
 
-func GetPackagesBundle(r *releasetypes.ReleaseConfig, imageDigests map[string]string) (anywherev1alpha1.PackageBundle, error) {
-	artifacts := map[string][]releasetypes.Artifact{
-		"eks-anywhere-packages":       r.BundleArtifactsTable["eks-anywhere-packages"],
-		"ecr-token-refresher":         r.BundleArtifactsTable["ecr-token-refresher"],
-		"credential-provider-package": r.BundleArtifactsTable["credential-provider-package"],
```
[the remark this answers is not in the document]
abhay-krishna: The artifacts are mapped based on the project name, so everything should just be mapped to `eks-anywhere-packages`. That means `r.BundleArtifactsTable["ecr-token-refresher"]` and `r.BundleArtifactsTable["credential-provider-package"]` are essentially empty and weren't even needed in the first place, so I removed them.

release/cli/pkg/helm/helm.go
abhay-krishna: The check was
```go
if artifact.Image.AssetName == "eks-anywhere-packages" || artifact.Image.AssetName == "ecr-token-refresher"
```
I only removed the mapping of `ecr-token-refresher` project name to a list of artifacts, since we don't have a project like that. The image artifact for ecr-token-refresher very much exists, inside the artifacts list mapped to project name `eks-anywhere-packages`, and that's what we're checking for here.

release/cli/pkg/operations/download.go
abhay-krishna: Responded [above](https://github.com/aws/eks-anywhere/pull/7186#discussion_r1433300878).

release/cli/pkg/operations/download.go
abhay-krishna: Responded [above](https://github.com/aws/eks-anywhere/pull/7186#discussion_r1433300878)

release/cli/pkg/operations/bundle_release.go
abhay-krishna: Responded [above](https://github.com/aws/eks-anywhere/pull/7186#discussion_r1433300878)

PR #7201, opened 2023-12-18
Add reconcile logic for MD Upgrade controller

taneyland:
*Issue #, if available:*

*Description of changes:*
Adding logic to reconcile MachineDeploymentUpgrade objects. This performs upgrades on the nodes in place but using the upgrade script provided in the node upgrader controller image.

*Testing (if applicable):*
Tilt with upgrader image. I observed the workers upgrade from 1.27 to 1.28

### DIFF
--- a/controllers/machinedeploymentupgrade_controller.go
+++ b/controllers/machinedeploymentupgrade_controller.go
@@ -18,23 +18,40 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"time"
 
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/constants"
+)
+
+const (
+	// mdUpgradeFinalizerName is the finalizer added to MachineDeploymentUpgrade objects to handle deletion.
+	mdUpgradeFinalizerName = "machinedeploymentupgrades.anywhere.eks.amazonaws.com/finalizer"
 )
 
 // MachineDeploymentUpgradeReconciler reconciles a MachineDeploymentUpgrade object.
 type MachineDeploymentUpgradeReconciler struct {
 	client client.Client
+	log    logr.Logger
 }
 
 // NewMachineDeploymentUpgradeReconciler returns a new instance of MachineDeploymentUpgradeReconciler.
 func NewMachineDeploymentUpgradeReconciler(client client.Client) *MachineDeploymentUpgradeReconciler {
 	return &MachineDeploymentUpgradeReconciler{
 		client: client,
+		log:    ctrl.Log.WithName("MachineDeploymentUpgradeController"),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +118,92 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
...
+		if err := r.client.Get(ctx, GetNamespacedNameType(nodeUpgrade.ObjectMeta.Name, nodeUpgrade.ObjectMeta.Namespace), nodeUpgrade); err == nil {
+			if err := r.client.Update(ctx, nodeUpgrade); err != nil {
+				return ctrl.Result{}, err
+			}
+		} else if err := r.client.Create(ctx, nodeUpgrade); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+		}
+	}
+	total := len(mdUpgrade.Spec.MachinesRequireUpgrade)
+	mdUpgrade.Status.RequireUpgrade = int64(total)
```
vivek-koppuru, line 141: I think we should keep all status related code in an `updateStatus` function so that it can be separated from the initiation of the creation of these objects.

controllers/nodeupgrade_controller.go
```diff
@@ -384,8 +385,32 @@ func upgraderPodExists(ctx context.Context, remoteClient client.Client, nodeName
...
+		},
+		Spec: anywherev1.NodeUpgradeSpec{
+			Machine: corev1.ObjectReference{
+				Kind:      machineRef.Kind,
+				Namespace: constants.EksaSystemNamespace,
+				Name:      machineRef.Name,
+			},
+			KubernetesVersion:     kubernetesVersion,
+			EtcdVersion:           &etcdVersion,
+			FirstNodeToBeUpgraded: firstControlPlane,
```
vivek-koppuru, line 413: Wondering if we want to separate cp and worker node as a field here instead of saying it's not the first node to be upgraded, but that also means the upgrader pod needs to take in that information.

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +118,92 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&anywherev1.MachineDeploymentUpgrade{}).
 		Complete(r)
 }
+
+func (r *MachineDeploymentUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, mdUpgrade *anywherev1.MachineDeploymentUpgrade) (ctrl.Result, error) {
+	// return early if machine deployment upgrade is already complete
+	if mdUpgrade.Status.Ready {
```
vivek-koppuru, line 124: When does this get triggered to not ready if that will be determined by the status of the nodes? I guess we are assuming that when we create the objects, this field will be empty right?

taneyland: I removed this check because we should just allow this to be idempotent

controllers/nodeupgrade_controller.go
taneyland: Yeah these fields are ignored in the md upgrader pods and aren't required technically so I could separate them

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -43,12 +56,63 @@ func NewMachineDeploymentUpgradeReconciler(client client.Client) *MachineDeploym
 //+kubebuilder:rbac:groups=anywhere.eks.amazonaws.com,resources=machinedeploymentupgrades/finalizers,verbs=update
 
 // Reconcile reconciles a MachineDeploymentUpgrade object.
-func (r *MachineDeploymentUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+// nolint:gocyclo
```
vivek-koppuru, line 59: What errors do we see if we remove this?

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +121,73 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&anywherev1.MachineDeploymentUpgrade{}).
 		Complete(r)
 }
+
+func (r *MachineDeploymentUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, mdUpgrade *anywherev1.MachineDeploymentUpgrade) (ctrl.Result, error) {
+	log.Info("Upgrading all worker nodes")
+	for _, machineRef := range mdUpgrade.Spec.MachinesRequireUpgrade {
+		nodeUpgrade := nodeUpgrader(machineRef, "v1.28.3-eks-1-28-9", "v3.5.9-eks-1-28-9", false)
```
vivek-koppuru, line 128: define these as constants for now at top so we know to replace them later?

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +121,73 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
...
+		if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create node upgrader for machine %s:  %v", machineRef.Name, err)
+		}
+	}
+
+	err := r.updateStatus(ctx, log, mdUpgrade)
```
vivek-koppuru, line 134: So this is needed again here? Defeats the purpose of the defer func() right?

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +121,73 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
...
+	// When MachineDeploymentUpgrade is fully deleted, we do not need to update the status. Without this check
+	// the subsequent patch operations would fail if the status is updated after it is fully deleted.
+	if !mdUpgrade.DeletionTimestamp.IsZero() && len(mdUpgrade.GetFinalizers()) == 0 {
+		log.Info("MachineDeploymentUpgrade is deleted, skipping status update")
+		return nil
+	}
+
+	log.Info("Updating MachineDeploymentUpgrade status")
+
+	var ready int64
```
vivek-koppuru, line 170: Let's have all `var` declarations at the top of the function, otherwise do something like `ready := int64(0)`, tho I don't remember if that's how the syntax is.

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +121,73 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
...
+	total := len(mdUpgrade.Spec.MachinesRequireUpgrade)
+	mdUpgrade.Status.RequireUpgrade = int64(total)
+
+	for _, machine := range mdUpgrade.Spec.MachinesRequireUpgrade {
+		nodeUpgrade, err := getNodeUpgrade(ctx, r.client, nodeUpgraderName(machine.Name))
+		if err != nil {
+			return err
+		}
+		if nodeUpgrade.Status.Completed {
+			ready++
```
vivek-koppuru, line 180: Do we need to decrement `RequireUpgrade` as well then before leaving this function? Or is that just the total count? If it's just count, then the name should be different?

controllers/nodeupgrade_controller.go
```diff
@@ -384,8 +385,40 @@ func upgraderPodExists(ctx context.Context, remoteClient client.Client, nodeName
...
+	n := &anywherev1.NodeUpgrade{}
+	if err := remoteClient.Get(ctx, GetNamespacedNameType(nodeUpgradeName, constants.EksaSystemNamespace), n); err != nil {
+		return nil, err
+	}
+	return n, nil
+}
+
+// nodeUpgradeName returns the name of the node upgrade object based on the machine reference.
+func nodeUpgraderName(machineRefName string) string {
+	return fmt.Sprintf("%s-node-upgrade", machineRefName)
```
vivek-koppuru, line 404:
```suggestion
	return fmt.Sprintf("%s-node-upgrader", machineRefName)
```

controllers/nodeupgrade_controller.go
vivek-koppuru: Can we at least make this field optional if we pass it in from md controller? But if it's too much refactor, then I say just leave it for now until we figure it out later.

controllers/machinedeploymentupgrade_controller.go
taneyland: Things have moved around so we don't need this anymore. Removed it

taneyland: `cyclomatic complexity 12 of func `(*MachineDeploymentUpgradeReconciler).Reconcile` is high `

vivek-koppuru: Can we link the GitHub issue like we do here? But this doesn't have to block the merge of the PR: https://github.com/aws/eks-anywhere/blob/main/controllers/cluster_controller.go#L205

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +128,92 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&anywherev1.MachineDeploymentUpgrade{}).
 		Complete(r)
 }
+
+func (r *MachineDeploymentUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, mdUpgrade *anywherev1.MachineDeploymentUpgrade) (ctrl.Result, error) {
+	log.Info("Upgrading all worker nodes")
+	for _, machineRef := range mdUpgrade.Spec.MachinesRequireUpgrade {
+		nodeUpgrade := mdNodeUpgrader(machineRef, kubeVersion)
+		if err := r.client.Create(ctx, nodeUpgrade); client.IgnoreAlreadyExists(err) != nil {
```
vivek-koppuru, line 136: Are we planning to schedule this one at a time, waiting for the previous node to finish first? Spoke with @rahulbabu95 on a potential approach for control plane.

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -57,3 +126,95 @@ func (r *MachineDeploymentUpgradeReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&anywherev1.MachineDeploymentUpgrade{}).
 		Complete(r)
 }
+
+func (r *MachineDeploymentUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger, mdUpgrade *anywherev1.MachineDeploymentUpgrade) (ctrl.Result, error) {
+	log.Info("Upgrading all worker nodes")
+	for _, machineRef := range mdUpgrade.Spec.MachinesRequireUpgrade {
+		nodeUpgrade := mdNodeUpgrader(machineRef, kubernetesVersion)
```
rahulbabu95, line 133: can we get the node Upgrade object first to see if it exists and handle the condition accordingly instead of attempting to create the object in every reconcile ?

PR #7217, opened 2023-12-20
Skip kube vip load balancer disabler

alikhan1621:
*Issue #, if available:*

*Description of changes:*
Suggested skipping kube vip disable. Not needed for most users. 

*Testing (if applicable):*


ndeksa (commented):
@alikhan1621  Your changes dont make much difference to the existing step. If you are suggesting that this step doesn't clarify in what condition an user can skip the step, then, you may want to rephrase it,  something like - Skip this step if you dont want to use the default load balancer kub-vip.  If you want to use a different load balancer, you can disable default Kube-vip as follows:"

PR #7223, opened 2023-12-20
add doc page for using GPUs with EKS-A

csplinter:

### DIFF
--- a/docs/content/en/docs/workloadmgmt/using-gpus.md
+++ b/docs/content/en/docs/workloadmgmt/using-gpus.md
@@ -7,13 +7,13 @@ description: >
   How to use the NVIDIA GPU Operator with EKS Anywhere on bare metal
 ---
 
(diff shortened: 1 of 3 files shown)


docs/content/en/docs/workloadmgmt/using-gpus.md
```diff
@@ -0,0 +1,165 @@
...
+See the [bare metal configuration]({{< relref "../getting-started/baremetal/bare-spec">}}) page and the [bare metal preparation]({{< relref "../getting-started/baremetal/bare-preparation/#prepare-hardware-inventory">}})  page for details. If the cluster spec sample below is used, your hardware inventory definition must have `type=cp` for the `labels` field in the hardware inventory for your server.
+
+<details>
+  <summary>Expand for a sample cluster spec</summary>
+  <br /> 
+  {{% content "gpu-sample-cluster.md" %}}
+</details>
+
+
+### Step 2: Create a single node EKS Anywhere cluster
```
chrisnegus, line 25:
```suggestion
### Step 2: Create a single-node EKS Anywhere cluster
```

chrisnegus (commented):
Just a few suggestions. Also, I generally only use the "Step 1:", "Step 2:", and so on, format for procedures that have substeps. This procedure could be done with simple "1.", "2." and so on, numbering. Not a big deal, but its something we could think about for consistency.

PR #7228, opened 2023-12-21
Upgrade workload using controller behind flag

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -177,7 +177,7 @@ func clustersHaveEksaVersion(mgmtCluster, cluster *v1alpha1.Cluster) bool {
 		return false
 	}
 
-	if cluster.Spec.EksaVersion == nil && mgmtCluster.Spec.EksaVersion == nil {
+	if cluster.Spec.EksaVersion == nil || mgmtCluster.Spec.EksaVersion == nil {
 		return false
 	}
 
--- a/pkg/validations/cluster.go
+++ b/pkg/validations/cluster.go
@@ -177,7 +177,7 @@ func clustersHaveEksaVersion(mgmtCluster, cluster *v1alpha1.Cluster) bool {
 		return false
 	}
 
-	if cluster.Spec.EksaVersion == nil || mgmtCluster.Spec.EksaVersion == nil {
+	if cluster.Spec.EksaVersion == nil && mgmtCluster.Spec.EksaVersion == nil {
 		return false
 	}
 
--- a/pkg/workflows/workload/upgrade.go
+++ b/pkg/workflows/workload/upgrade.go
@@ -1,0 +1,67 @@
+package workload
+
+import (
+	"context"
+
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
+)
+
+// Upgrade is a schema for upgrade cluster.
+type Upgrade struct {
+	provider         providers.Provider
+	clusterManager   interfaces.ClusterManager
+	gitOpsManager    interfaces.GitOpsManager
+	writer           filewriter.FileWriter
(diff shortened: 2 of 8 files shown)


pkg/validations/cluster.go
```diff
@@ -177,7 +177,7 @@ func clustersHaveEksaVersion(mgmtCluster, cluster *v1alpha1.Cluster) bool {
 		return false
 	}
 
-	if cluster.Spec.EksaVersion == nil && mgmtCluster.Spec.EksaVersion == nil {
+	if cluster.Spec.EksaVersion == nil || mgmtCluster.Spec.EksaVersion == nil {
```
g-gaston, line 180: curious why this change? If it was a bug, could we add a test?

pkg/workflows/workload/upgrade.go
```diff
@@ -0,0 +1,62 @@
...
+		writer:           writer,
+		eksdInstaller:    eksdInstaller,
+		ClusterUpgrader:  clusterUpgrade,
+		packageInstaller: packageInstaller,
+	}
+}
+
+// Run Upgrade implements upgrade functionality for workload cluster's upgrade operation.
+func (c *Upgrade) Run(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec, validator interfaces.Validator) error {
+	logger.Info("POC New workflow creating workload cluster using the controller")
```
g-gaston, line 47: we already a similar log at the cmd level
if you wanna keep both just make sure we clean them up when removing the feature toggle

pkg/workflows/workload/upgradecluster.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+)
+
+type upgradeCluster struct{}
+
+// Run upgradeCluster performs actions needed to upgrade the management cluster.
+func (s *upgradeCluster) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("Creating workload cluster")
```
g-gaston, line 15:
```suggestion
	logger.Info("Upgrading workload cluster components")
```

pkg/workflows/workload/validate.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/validations"
+)
+
+type setAndValidateWorkloadTask struct{}
+
+// SetAndValidateTask implementation
+
+// Run upgradeCluster performs actions needed to upgrade the workload cluster.
```
g-gaston, line 16: fix comments

pkg/workflows/workload/validate.go
```diff
@@ -0,0 +1,59 @@
...
+func (s *setAndValidateWorkloadTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	logger.Info("POC Performing setup and validations")
```
g-gaston, line 18: now i'm just curious
why the same similar log everywhere? also tbh i'm a bit confused by the term poc
not big deal if you want to keep this while you test but it seems like extra work to clean later

pkg/workflows/workload/writeclusterconfig.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		commandContext.SetError(err)
+	}
+
+	successMsg := ""
+	if commandContext.CurrentClusterSpec != nil {
+		successMsg = "Cluster upgraded!"
+	} else {
+		successMsg = "Cluster created!"
+	}
```
g-gaston, line 26: the abstractions presented by `pkg/workflows` don't favor reusing tasks
the fact that you are tying to do this here seems to indicate maybe we should look into `pkg/workflow` instead. You don't need to do it, we can move forward with this PR as is. But maybe worth thinking about it

pkg/validations/cluster.go
tatlat: This change isn't needed. I just made it when I was looking into having the controller set eksaVersion.

g-gaston (commented):
/lgtm

pkg/workflows/workload/writeclusterconfig.go
tatlat: Yeah, I can fix that in a future PR.

PR #7231, opened 2023-12-22
Verify eksarelease exists on mgmt cluster

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -373,6 +373,10 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 		if err := validations.ValidateManagementEksaVersion(mgmt, cluster); err != nil {
 			return controller.Result{}, err
 		}
+
+		if err := validateEksaRelease(ctx, r.client, cluster); err != nil {
+			return controller.Result{}, err
+		}
 	}
 
 	if cluster.RegistryAuth() {
@@ -610,4 +614,18 @@ func (r *ClusterReconciler) setDefaultBundlesRefOrEksaVersion(ctx context.Contex
 
 	clus.Status.FailureMessage = ptr.String("Management cluster must have either EksaVersion or BundlesRef")
 	return fmt.Errorf("could not set default values")
+}
+
+func validateEksaRelease(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
+	if cluster.Spec.EksaVersion == nil {
+		return nil
+	}
+	if err := validations.ValidateEksaReleaseExistOnManagement(ctx, clientutil.NewKubeClient(client), cluster); err != nil {
+		errMsg := fmt.Sprintf("eksarelease %v could not be found on the management cluster", *cluster.Spec.EksaVersion)
+		reason := anywherev1.EksaVersionInvalidReason
+		cluster.Status.FailureMessage = ptr.String(errMsg)
+		cluster.Status.FailureReason = &reason
+		return err
+	}
+	return nil
 }
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -373,10 +373,10 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 		if err := validations.ValidateManagementEksaVersion(mgmt, cluster); err != nil {
 			return controller.Result{}, err
 		}
+	}
 
-		if err := validateEksaRelease(ctx, r.client, cluster); err != nil {
-			return controller.Result{}, err
-		}
+	if err := validateEksaRelease(ctx, r.client, cluster); err != nil {
+		return controller.Result{}, err
 	}
 
 	if cluster.RegistryAuth() {
@@ -620,11 +620,14 @@ func validateEksaRelease(ctx context.Context, client client.Client, cluster *any
 	if cluster.Spec.EksaVersion == nil {
 		return nil
 	}
-	if err := validations.ValidateEksaReleaseExistOnManagement(ctx, clientutil.NewKubeClient(client), cluster); err != nil {
+	err := validations.ValidateEksaReleaseExistOnManagement(ctx, clientutil.NewKubeClient(client), cluster)
+	if apierrors.IsNotFound(err) {
 		errMsg := fmt.Sprintf("eksarelease %v could not be found on the management cluster", *cluster.Spec.EksaVersion)
 		reason := anywherev1.EksaVersionInvalidReason
 		cluster.Status.FailureMessage = ptr.String(errMsg)
 		cluster.Status.FailureReason = &reason
+		return err
+	} else if err != nil {
 		return err
 	}
 	return nil
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -180,7 +180,7 @@ func TestClusterReconcilerReconcileSelfManagedCluster(t *testing.T) {
 
 	clusterValidator := mocks.NewMockClusterValidator(controller)
 	registry := newRegistryMock(providerReconciler)
-	c := fake.NewClientBuilder().WithRuntimeObjects(selfManagedCluster, kcp).Build()
+	c := fake.NewClientBuilder().WithRuntimeObjects(selfManagedCluster, kcp, test.EKSARelease()).Build()
 	mockPkgs := mocks.NewMockPackagesClient(controller)
 	providerReconciler.EXPECT().Reconcile(ctx, gomock.AssignableToTypeOf(logr.Logger{}), sameName(selfManagedCluster))
 	mhcReconciler.EXPECT().Reconcile(ctx, gomock.AssignableToTypeOf(logr.Logger{}), sameName(selfManagedCluster)).Return(nil)
@@ -707,7 +707,7 @@ func TestClusterReconcilerReconcileSelfManagedClusterConditions(t *testing.T) {
 				md.Status = tt.machineDeploymentStatus
 			})
 
-			objs = append(objs, config.Cluster, bundles, kcp, md1)
+			objs = append(objs, config.Cluster, bundles, kcp, md1, test.EKSARelease())
 			for _, o := range config.ChildObjects() {
 				objs = append(objs, o)
 			}
@@ -873,7 +873,7 @@ func TestClusterReconcilerReconcileGenerations(t *testing.T) {
 			ctx := context.Background()
 
 			objs := make([]runtime.Object, 0, 7+len(machineDeployments))
-			objs = append(objs, config.Cluster, bundles)
+			objs = append(objs, config.Cluster, bundles, test.EKSARelease())
 			for _, o := range config.ChildObjects() {
 				objs = append(objs, o)
 			}
@@ -1039,7 +1039,7 @@ func TestClusterReconcilerReconcileSelfManagedClusterRegAuthFailNoSecret(t *test
 	mhcReconciler := mocks.NewMockMachineHealthCheckReconciler(controller)
 
 	registry := newRegistryMock(providerReconciler)
-	c := fake.NewClientBuilder().WithRuntimeObjects(selfManagedCluster).Build()
+	c := fake.NewClientBuilder().WithRuntimeObjects(selfManagedCluster, test.EKSARelease()).Build()
 
 	r := controllers.NewClusterReconciler(c, registry, iam, clusterValidator, nil, mhcReconciler)
 	_, err := r.Reconcile(ctx, clusterRequest(selfManagedCluster))
@@ -1255,7 +1255,7 @@ func TestClusterReconcilerSkipDontInstallPackagesOnSelfManaged(t *testing.T) {
 			ReconciledGeneration: 1,
 		},
 	}
-	objs := []runtime.Object{cluster}
+	objs := []runtime.Object{cluster, test.EKSARelease()}
 	cb := fake.NewClientBuilder()
 	mockClient := cb.WithRuntimeObjects(objs...).Build()
 	nullRegistry := newRegistryForDummyProviderReconciler()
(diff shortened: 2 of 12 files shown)


pkg/cluster/builder.go
```diff
@@ -125,7 +126,9 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 	}
 
 	releaseVersion := v1alpha1.EksaVersion(release.Version)
-	config.Cluster.Spec.EksaVersion = &releaseVersion
+	if config.Cluster.IsSelfManaged() || (config.Cluster.IsManaged() && !features.OverrideEksaVersion().IsActive()) {
```
g-gaston, line 129: why are we using a feature flag here? I probably didn't understand what we are trying to do here

pkg/executables/kubectl.go
```diff
@@ -2535,3 +2536,20 @@ func (k *Kubectl) DeleteCRD(ctx context.Context, crd, kubeconfig string) error {
 
 	return nil
 }
+
+// GetEKSARelease returns an eksareleases CR on the cluster.
+func (k *Kubectl) GetEKSARelease(ctx context.Context, releaseName string, kubeconfigFile string) (*releasev1alpha1.EKSARelease, error) {
```
g-gaston, line 2541: what do you think about instead of adding this method here, injecting a `kubernetes.Client` in the validations code, so you can just use `Get`?

pkg/validations/cluster.go
```diff
@@ -215,3 +217,12 @@ func ValidateK8s129Support(clusterSpec *cluster.Spec) error {
 	}
 	return nil
 }
+
+// ValidateEksaReleaseExistOnManagement checks if there is a corresponding eksareleases CR for workload's eksaVersion on the mgmt cluster.
+func ValidateEksaReleaseExistOnManagement(ctx context.Context, k KubectlClient, mgmtCluster *types.Cluster, workload *cluster.Spec) error {
+	v := workload.Cluster.Spec.EksaVersion
+	if release, err := k.GetEKSARelease(ctx, string(*v), mgmtCluster.KubeconfigFile); err != nil || release == nil {
```
g-gaston, line 224: these seem like they should be two different errors imo (error reading and not found error)

pkg/validations/createvalidations/preflightvalidations.go
```diff
@@ -105,6 +105,13 @@ func (v *CreateValidations) PreflightValidations(ctx context.Context) []validati
 					Err:         validations.ValidateManagementClusterEksaVersion(ctx, k, v.Opts.ManagementCluster, v.Opts.Spec),
 				}
 			},
+			func() *validations.ValidationResult {
+				return &validations.ValidationResult{
+					Name:        "validate eksa release components exist on management cluster",
+					Remediation: fmt.Sprintf("check eksaVersion field for workload cluster %s matches release previously used to upgrade maangement cluster %s", v.Opts.WorkloadCluster.Name, v.Opts.ManagementCluster.Name),
```
g-gaston, line 111: nit: this message is a bit confusing IMO

What do you think about saying something like "make sure you use one of the available versions and providing the kubectl command to read the list of eksa versions?

pkg/executables/kubectl.go
tatlat: I don't have any strong opinions about it. Is there a particular reason we don't want to add this method? 
For the time being, I can remove it and replace it with GetObject instead.

pkg/validations/createvalidations/preflightvalidations.go
tatlat: I agree. Let me know if the new remediation message is still confusing.

pkg/cluster/builder.go
tatlat: We don't need the feature flag. I was just using it to test what would happen if we let the user pass an eksaVersion value where it was previously set by the CLI and vice versa.

pkg/executables/kubectl.go
g-gaston: This pattern where we add accessor methods (get, list) and writer methods (create, update) for every single resource doesn't scale. Just look at the sheer size of `kubectl.go`. Not only requires custom code on the implementation side but it also requires updating the receiving interfaces (and their mocks) every time we add/change one of these methods. Moreover, testing code that uses this type of client can be quite cumbersome, needing custom mocks and specific EXPECT calls for every single read or write. Instead, testing using a FakeClient is way easier.

AFAIK GetObject doesn't behave the same way as the client, since it requires to pass the resource type.

pkg/cluster/builder.go
```diff
@@ -125,7 +125,9 @@ func (b FileSpecBuilder) Build(clusterConfigURL string) (*Spec, error) {
 	}
 
 	releaseVersion := v1alpha1.EksaVersion(release.Version)
-	config.Cluster.Spec.EksaVersion = &releaseVersion
+	if config.Cluster.IsSelfManaged() || (config.Cluster.IsManaged() && config.Cluster.Spec.EksaVersion == nil) {
```
g-gaston, line 128: Why is the logic different for management and workload clusters?

Sorry if I'm resurfacing something we talked about already, but I would love if we could avoid this complexity. It's not the code that concerns me, but the user experience.

This seems like some nuanced logic that we need to document and the user has to take into account on every upgrade. And although I'm sure you have a good reason to do this, I don't thin it's very apparent, which in my experience ends up with the user forgetting.

Could we just either always overwrite it or never overwrite if already set?

pkg/validations/upgradevalidations/preflightvalidations.go
```diff
@@ -134,6 +134,13 @@ func (u *UpgradeValidations) PreflightValidations(ctx context.Context) []validat
 					Err:         validations.ValidateManagementClusterEksaVersion(ctx, k, u.Opts.ManagementCluster, u.Opts.Spec),
 				}
 			},
+			func() *validations.ValidationResult {
+				return &validations.ValidationResult{
+					Name:        "validate eksa release components exist on management cluster",
+					Remediation: fmt.Sprintf("check eksaVersion field for workload cluster %s matches release previously used to upgrade maangement cluster %s", u.Opts.WorkloadCluster.Name, u.Opts.ManagementCluster.Name),
```
g-gaston, line 140: can we use the same message as for create here?

pkg/validations/createvalidations/preflightvalidations.go
```diff
@@ -105,6 +106,13 @@ func (v *CreateValidations) PreflightValidations(ctx context.Context) []validati
 					Err:         validations.ValidateManagementClusterEksaVersion(ctx, k, v.Opts.ManagementCluster, v.Opts.Spec),
 				}
 			},
+			func() *validations.ValidationResult {
+				return &validations.ValidationResult{
+					Name:        "validate eksa release components exist on management cluster",
+					Remediation: fmt.Sprintf("ensure eksaVersion is in the correct format (vMajor.Minor.Patch) and matches any of the available releases on the management cluster: kubectl get eksareleases -n %s --kubeconfig %s", constants.EksaSystemNamespace, v.Opts.ManagementCluster.KubeconfigFile),
```
g-gaston, line 112: nit

```suggestion
					Remediation: fmt.Sprintf("ensure eksaVersion is in the correct format (vMajor.Minor.Patch) and matches one of the available releases on the management cluster: kubectl get eksareleases -n %s --kubeconfig %s", constants.EksaSystemNamespace, v.Opts.ManagementCluster.KubeconfigFile),
```

pkg/cluster/builder.go
tatlat: Yes, we can continue to always overwrite it.

controllers/cluster_controller.go
```diff
@@ -373,6 +373,10 @@ func (r *ClusterReconciler) preClusterProviderReconcile(ctx context.Context, log
 		if err := validations.ValidateManagementEksaVersion(mgmt, cluster); err != nil {
 			return controller.Result{}, err
 		}
+
+		if err := validateEksaRelease(ctx, r.client, cluster); err != nil {
```
g-gaston, line 377: why is this only validated for workload clusters? shouldn't this validation be for all clusters?

controllers/cluster_controller.go
```diff
@@ -611,3 +615,17 @@ func (r *ClusterReconciler) setDefaultBundlesRefOrEksaVersion(ctx context.Contex
 	clus.Status.FailureMessage = ptr.String("Management cluster must have either EksaVersion or BundlesRef")
 	return fmt.Errorf("could not set default values")
 }
+
+func validateEksaRelease(ctx context.Context, client client.Client, cluster *anywherev1.Cluster) error {
+	if cluster.Spec.EksaVersion == nil {
+		return nil
+	}
+	if err := validations.ValidateEksaReleaseExistOnManagement(ctx, clientutil.NewKubeClient(client), cluster); err != nil {
```
g-gaston, line 623: I think you need to distinguish between error reading and the release not existing. You don't want to set an error in the status for the first case.

controllers/cluster_controller_test.go
```diff
@@ -1498,6 +1498,51 @@ func TestClusterReconcilerValidateManagementEksaVersionFail(t *testing.T) {
 	g.Expect(err).To(HaveOccurred())
 }
 
+func TestClusterReconcilerValidateManagementEksaReleaseFail(t *testing.T) {
```
g-gaston, line 1501:
```suggestion
func TestClusterReconcilerNotAvailableEksaVersion(t *testing.T) {
```

controllers/cluster_controller_test.go
```diff
@@ -1498,6 +1498,51 @@ func TestClusterReconcilerValidateManagementEksaVersionFail(t *testing.T) {
...
+	logCtx := ctrl.LoggerInto(ctx, log)
+
+	iam.EXPECT().EnsureCASecret(logCtx, gomock.AssignableToTypeOf(logr.Logger{}), sameName(config.Cluster)).Return(controller.Result{}, nil)
+	clusterValidator.EXPECT().ValidateManagementClusterName(logCtx, gomock.AssignableToTypeOf(logr.Logger{}), sameName(config.Cluster)).Return(nil)
+
+	r := controllers.NewClusterReconciler(testClient, registry, iam, clusterValidator, mockPkgs, nil)
+
+	_, err := r.Reconcile(logCtx, clusterRequest(config.Cluster))
+
+	g.Expect(err).To(HaveOccurred())
```
g-gaston, line 1543: assert the status error?

pkg/validations/cluster.go
```diff
@@ -6,16 +6,22 @@ import (
...
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers"
 	"github.com/aws/eks-anywhere/pkg/semver"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/utils/ptr"
+	releasev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// ReleaseType is the type of the eksareleases custom resource.
+var ReleaseType = fmt.Sprintf("eksareleases.%s", releasev1alpha1.GroupVersion.Group)
```
g-gaston, line 23: is this needed now?

pkg/validations/cluster.go
```diff
@@ -215,3 +221,14 @@ func ValidateK8s129Support(clusterSpec *cluster.Spec) error {
...
+
+// ValidateEksaReleaseExistOnManagement checks if there is a corresponding eksareleases CR for workload's eksaVersion on the mgmt cluster.
+func ValidateEksaReleaseExistOnManagement(ctx context.Context, k kubernetes.Client, workload *v1alpha1.Cluster) error {
+	v := workload.Spec.EksaVersion
+	release := &releasev1alpha1.EKSARelease{}
+	err := k.Get(ctx, releasev1alpha1.GenerateEKSAReleaseName(string(*v)), constants.EksaSystemNamespace, release)
+	if err != nil {
+		return err
+	}
+	return nil
```
g-gaston, line 233:
```suggestion
	return k.Get(ctx, releasev1alpha1.GenerateEKSAReleaseName(string(*v)), constants.EksaSystemNamespace, &releasev1alpha1.EKSARelease{})
```
simpler?

controllers/cluster_controller.go
tatlat: Good catch. I forgot users can upgrade management clusters through FLC now. I'll change it so we validate for all clusters.

pkg/validations/cluster.go
tatlat: no

controllers/cluster_controller_test.go
```diff
@@ -1498,6 +1498,60 @@ func TestClusterReconcilerValidateManagementEksaVersionFail(t *testing.T) {
...
+	req := clusterRequest(config.Cluster)
+	_, err := r.Reconcile(logCtx, req)
+
+	g.Expect(err).To(HaveOccurred())
+	eksaCluster := &anywherev1.Cluster{}
+
+	err = testClient.Get(context.TODO(), req.NamespacedName, eksaCluster)
+	if apierrors.IsNotFound(err) {
+		t.Fatal("expected to find cluster")
+	}
```
g-gaston, line 1550:
```suggestion
	g.Expect(testClient.Get(ctx, req.NamespacedName, eksaCluster)).To(Succeed())
```

controllers/cluster_controller_test.go
```diff
@@ -1498,6 +1498,60 @@ func TestClusterReconcilerValidateManagementEksaVersionFail(t *testing.T) {
...
+	g.Expect(eksaCluster.Status.FailureReason).ToNot(BeNil())
+	g.Expect(string(*eksaCluster.Status.FailureReason)).To(ContainSubstring(string(anywherev1.EksaVersionInvalidReason)))
```
g-gaston, line 1552: nit: I think gomega has a `HaveValue` or something like that that allows you to assert pointer values. So you can avoid the not nil assert and the casting

Also, can we assert the error message as well?

PR #7238, opened 2023-12-29
Add upgrade management components cmd

d8660091:
*Issue #, if available:*

*Description of changes:* Add "upgrade management-components" subcommand to allow users to upgrade their management components in management cluster.
 
*Testing (if applicable):* Manual test against docker and cloudstack. 1) upgrade management-components 2) upgrade the field of eksaVersion of a cluster object and observed that the clusters have been upgraded correctly. Notes, there is an issue with cluster-api 1.6.0 upgrade that requires restarting cluster-api controller to make the upgrade successful.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
+++ b/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
@@ -1,0 +1,96 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/aws/eks-anywhere/cmd/eksctl-anywhere/cmd/aflag"
+	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/workflows/management"
+)
+
+type upgradeManagementComponentsOptions struct {
+	clusterOptions
+}
+
+var umco = &upgradeManagementComponentsOptions{}
+
+func init() {
+	flagSet := upgradeManagementComponentsCmd.Flags()
+	aflag.String(aflag.ClusterConfig, &umco.fileName, flagSet)
+	aflag.String(aflag.BundleOverride, &umco.bundlesOverride, flagSet)
+}
+
+var upgradeManagementComponentsCmd = &cobra.Command{
+	Use:          "management-components",
+	Short:        "Upgrade management components in a management cluster",
+	Long:         "The term 'management components' encompasses all Kubernetes controllers and their CRDs present in the management cluster that are responsible for reconciling your EKS Anywhere (EKS-A) cluster. This command is specifically designed to facilitate the upgrade of these management components. Post this upgrade, the cluster itself can be upgraded by updating the 'eksaRelease' field in your eksa cluster object.",
+	PreRunE:      bindFlagsToViper,
+	SilenceUsage: true,
+	Args:         cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := cmd.Context()
+
+		clusterSpec, err := newClusterSpec(clusterOptions{
+			fileName: umco.fileName,
+		})
+		if err != nil {
+			return err
+		}
+		if !clusterSpec.Cluster.IsSelfManaged() {
+			return fmt.Errorf("cluster %s doesn't contain management components to be upgraded", clusterSpec.Cluster.Name)
+		}
+
+		cliConfig := buildCliConfig(clusterSpec)
+		dirs, err := uc.directoriesToMount(clusterSpec, cliConfig)
+		if err != nil {
+			return err
+		}
+
+		factory := dependencies.ForSpec(clusterSpec).WithExecutableMountDirs(dirs...).
+			WithBootstrapper().
+			WithCliConfig(cliConfig).
+			WithClusterManager(clusterSpec.Cluster, nil).
+			WithClusterApplier().
+			WithProvider(umco.fileName, clusterSpec.Cluster, false, "", false, "", nil, nil).
+			WithGitOpsFlux(clusterSpec.Cluster, clusterSpec.FluxConfig, cliConfig).
+			WithWriter().
+			WithCAPIManager().
+			WithEksdUpgrader().
+			WithEksdInstaller().
+			WithKubectl().
+			WithValidatorClients()
+
+		deps, err := factory.Build(ctx)
+		if err != nil {
+			return err
+		}
+		defer close(cmd.Context(), deps)
+
+		runner := management.NewUpgradeManagementComponentsRunner(
+			deps.Provider,
+			deps.CAPIManager,
+			deps.ClusterManager,
+			deps.GitOpsFlux,
+			deps.Writer,
+			deps.EksdUpgrader,
+			deps.EksdInstaller,
+		)
+
+		managementCluster := &types.Cluster{
+			Name:               clusterSpec.Cluster.Name,
+			KubeconfigFile:     kubeconfig.FromClusterName(clusterSpec.Cluster.Name),
+			ExistingManagement: clusterSpec.Cluster.IsSelfManaged(),
+		}
+
+		validator := management.NewUMCValidator(managementCluster, deps.Kubectl)
+		return runner.Run(ctx, clusterSpec, managementCluster, validator)
+	},
+}
+
+func init() {
+	upgradeCmd.AddCommand(upgradeManagementComponentsCmd)
+}
--- a/pkg/workflows/management/core_components.go
+++ b/pkg/workflows/management/core_components.go
@@ -31,7 +31,7 @@ func (s *ensureEtcdCAPIComponentsExist) Checkpoint() *task.CompletedTask {
 	}
 }
 
-func (s *ensureEtcdCAPIComponentsExist) Restore(ctx context.Context, commandContext *task.CommandContext, completedTask *task.CompletedTask) (task.Task, error) {
+func (s *ensureEtcdCAPIComponentsExist) Restore(_ context.Context, _ *task.CommandContext, _ *task.CompletedTask) (task.Task, error) {
 	return &pauseGitOpsReconcile{}, nil
 }
 
@@ -39,8 +39,7 @@ type upgradeCoreComponents struct {
 	UpgradeChangeDiff *types.ChangeDiff
 }
 
-// Run upgradeCoreComponents upgrades pre cluster upgrade components.
-func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func runUpgradeCoreComponents(ctx context.Context, commandContext *task.CommandContext) error {
 	logger.Info("Upgrading core components")
 
 	err := commandContext.Provider.PreCoreComponentsUpgrade(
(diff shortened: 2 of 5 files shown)


cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
```diff
@@ -0,0 +1,96 @@
...
+func init() {
+	flagSet := upgradeManagementComponentsCmd.Flags()
+	aflag.String(aflag.ClusterConfig, &umco.fileName, flagSet)
+	aflag.String(aflag.BundleOverride, &umco.bundlesOverride, flagSet)
+}
+
+var upgradeManagementComponentsCmd = &cobra.Command{
+	Use:          "management-components",
+	Short:        "Upgrade management components in a management cluster",
+	Long:         "This command is used to upgrade management components in a management cluster",
```
jacobweinstock, line 29: This long description would be a great place to link to or answer questions like: What are the names of the management components? Generally speaking, what will be involved in this process? What is the expected impact to management and workload clusters during this process? Are there any things to take into consideration for when i should or shouldnt run this upgrade process? etc

I don't think all of these questions should or need to be answered in this long description but there is a real opportunity to provide actual value and confidence to customers in this long description.

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+type upgradeCoreComponentsMC struct {
+	upgradeCoreComponents
+}
+
+func (u *upgradeCoreComponentsMC) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	ntask := u.upgradeCoreComponents.Run(ctx, commandContext)
+	if _, ok := ntask.(*workflows.CollectMgmtClusterDiagnosticsTask); !ok {
```
g-gaston, line 147: I would recommend to not trying to reuse tasks across workflows. This package wasn't designed to do so. Instead, just extract the common functionality to a method and call the method from two different tasks that just act as a wrapper.

If you want to reuse tasks, look into `pkg/workflow` which was better designed to do so (and it's better in general).

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,163 @@
...
+	return ntask
+}
+
+type installNewComponentsMC struct {
+	installNewComponents
+}
+
+func (i *installNewComponentsMC) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	ntask := i.installNewComponents.Run(ctx, commandContext)
+	if _, ok := ntask.(*workflows.CollectMgmtClusterDiagnosticsTask); !ok {
```
g-gaston, line 159: same here

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,163 @@
...
+// UpgradeManagementComponentsRunner is a schema for upgrade management components.
+type UpgradeManagementComponentsRunner struct {
+	provider        providers.Provider
+	clusterManager  interfaces.ClusterManager
+	gitOpsManager   interfaces.GitOpsManager
+	writer          filewriter.FileWriter
+	capiManager     interfaces.CAPIManager
+	eksdInstaller   interfaces.EksdInstaller
+	eksdUpgrader    interfaces.EksdUpgrader
+	clusterUpgrader interfaces.ClusterUpgrader
```
g-gaston, line 28: do we need all these things? for example, `ClusterUpgrader` doesn't seem to fit here.

g-gaston: removing them from here also allows us to not have to build them in the factory

cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
```diff
@@ -0,0 +1,96 @@
...
+	PreRunE:      bindFlagsToViper,
```
g-gaston, line 30: nit: are we using viper?

pkg/workflows/management/upgrade_management_components.go
d8660091: Yes, I tried removing each dependency and any missing would cause error.

cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
d8660091: It's there because consistency with other commands and we need "viper.AutomaticEnv()" inside "bindFlagsToViper".

pkg/workflows/management/upgrade_management_components.go
d8660091: I did another check:
clusterUpgrade is removed
eksdinstaller is kept because it run InstallEksdManifest in task "install-new-eksa-version-components"
writer is kept because it writes "checkpoint" file when the task run failed

pkg/workflows/management/core_components.go
```diff
@@ -39,8 +39,7 @@ type upgradeCoreComponents struct {
 	UpgradeChangeDiff *types.ChangeDiff
 }
 
-// Run upgradeCoreComponents upgrades pre cluster upgrade components.
-func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func runUpgradeCoreComponents(ctx context.Context, commandContext *task.CommandContext) error {
```
g-gaston, line 42: nit: we can drop `run` since it doesn't add any new info
```suggestion
func upgradeCoreComponents(ctx context.Context, commandContext *task.CommandContext) error {
```

pkg/workflows/management/install_new_components.go
```diff
@@ -9,21 +9,28 @@ import (
 
 type installNewComponents struct{}
 
-// Run installNewComponents performs actions needed to upgrade the management cluster.
-func (s *installNewComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func runInstallNewComponents(ctx context.Context, commandContext *task.CommandContext) error {
```
g-gaston, line 12: nit: same here
```suggestion
func installNewComponents(ctx context.Context, commandContext *task.CommandContext) error {
```

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/task"
+	"github.com/aws/eks-anywhere/pkg/types"
+	"github.com/aws/eks-anywhere/pkg/validations"
+	"github.com/aws/eks-anywhere/pkg/workflows"
+	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
+)
+
+// UpgradeManagementComponentsRunner is a schema for upgrade management components.
+type UpgradeManagementComponentsRunner struct {
```
g-gaston, line 20:
```suggestion
type UpgradeManagementComponentsWorkflow struct {
```

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,204 @@
...
+		EksdInstaller:     umc.eksdInstaller,
+	}
+
+	return task.NewTaskRunner(&setupAndValidateMC{}, umc.writer).RunTask(ctx, commandContext)
+}
+
+type setupAndValidateMC struct{}
+
+// Run setupAndValidate validates management cluster before upgrade process starts.
+func (s *setupAndValidateMC) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
```
g-gaston, line 109: nit: i would prefer if we had one file per task as in the other workflows we have been writing lately

g-gaston (commented):
/lgtm

pkg/workflows/management/core_components.go
d8660091: I tried it, but upgradeCoreComponents function name is conflicting with the struct name "upgradeCoreComponents"

pkg/workflows/management/install_new_components.go
d8660091: same, the installNewComponents name has been take by the struct

pkg/workflows/management/core_components.go
g-gaston: got you
maybe rename the task to `upgradeCoreComponentsTask`?

pkg/workflows/management/install_new_components.go
jacobweinstock: for what are you optimizing by having this function at all? Is it redundant to `func (s *installNewComponents) Run`?

pkg/workflows/management/core_components.go
jacobweinstock: for what are you optimizing by having this function at all? Is it redundant to `func (s *upgradeCoreComponents) Run`?

d8660091: sounds good to me

pkg/workflows/management/core_components.go
```diff
@@ -31,16 +31,15 @@ func (s *ensureEtcdCAPIComponentsExist) Checkpoint() *task.CompletedTask {
...
 	return &pauseGitOpsReconcile{}, nil
 }
 
 type upgradeCoreComponents struct {
 	UpgradeChangeDiff *types.ChangeDiff
 }
 
-// Run upgradeCoreComponents upgrades pre cluster upgrade components.
-func (s *upgradeCoreComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func upgradeCoreComponentsTask(ctx context.Context, commandContext *task.CommandContext) error {
```
jacobweinstock, line 42: For what are you optimizing with this func? What value do we gain in maintainability or understandability by having it instead of putting its logic directly in `func (s *upgradeCoreComponents) Run`?

pkg/workflows/management/install_new_components.go
```diff
@@ -9,21 +9,28 @@ import (
 
 type installNewComponents struct{}
 
-// Run installNewComponents performs actions needed to upgrade the management cluster.
-func (s *installNewComponents) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+func installNewComponentsTask(ctx context.Context, commandContext *task.CommandContext) error {
```
jacobweinstock, line 12: For what are you optimizing with this func? What value do we gain in maintainability or understandability by having it instead of putting its logic directly in `func (s *installNewComponents) Run`?

jacobweinstock (changes requested):

pkg/workflows/management/core_components.go
d8660091: As discussed with Guillermo, we want to reuse this functionality in the task of upgradeCoreComponentsMC, although it will introduce some code duplicity.

pkg/workflows/management/install_new_components.go
d8660091: Same as above

pkg/workflows/management/core_components.go
jacobweinstock: This code can still be shared across funcs. `upgradeCoreComponentsMC` can create a `upgradeCoreComponents` struct and call its `Run` method. Im still having trouble understanding the value and for what is being optimized.

jacobweinstock (dismissed):

pkg/workflows/management/core_components.go
d8660091: Yeah, it's a little controversial, https://github.com/aws/eks-anywhere/pull/7238#discussion_r1439359609. Since I didn't have a strong opinion on it, I went to the approach proposed in that comment.

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,204 @@
...
+		Checkpoint: nil,
+	}
+}
+
+type upgradeCoreComponentsMC struct {
+	UpgradeChangeDiff *types.ChangeDiff
+}
+
+func (s *upgradeCoreComponentsMC) Name() string {
+	return "upgrade-core-components"
```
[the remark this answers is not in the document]
d8660091: Modified

pkg/workflows/management/upgrade_management_components.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (s *installNewComponentsMC) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	if err := installNewComponentsTask(ctx, commandContext); err != nil {
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+	return nil
+}
+
+func (s *installNewComponentsMC) Name() string {
+	return "install-new-eksa-version-components"
```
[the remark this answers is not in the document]
d8660091: Modified

pkg/workflows/management/core_components.go
jacobweinstock: i see. thanks. This is not clear from the code and will be lost in the code once this is merged. It also doesnt seem like the same thing here. I'm not talking about type asserting anything like this "`ntask.(*workflows.CollectMgmtClusterDiagnosticsTask)`"

I'm talking about reusing `upgradeCoreComponents`. something like:
```Go
upgradeCoreComponents{}.Run()
```

Issue #3182, opened 2022-08-26
Allow customization of cilium CNI


jacobweinstock:
**What would you like to be added**: I would like to be able to customize the Cilium CNI deployment. Specifically, I would like to be able to use Cilium's [native routing](https://docs.cilium.io/en/v1.10/concepts/networking/routing/#native-routing) functionality.

**Why is this needed**: By default cilium uses encapsulation for network traffic and that adds overhead and latency. For latency sensitive workloads this is required.

chrisdoherty4:
Reopened as docs are outstanding.

PR #7239, opened 2023-12-29
Add docs for routing mode

chrisdoherty4:
Direct routing was added as part of #3182. This adds the API documentation to providers as well as other missing documentation.


chrisnegus (commented):
/lgtm
I think it's a great idea to look for more opportunities to reuse small sections of text between the providers when appropriate.

PR #7245, opened 2024-01-02
 Add cleanup to reset security group on aigrapped registry mirror docker tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
Currently, when a docker air-gapped e2e test fails, it does not reset the security group to default so outgoing commands fail. We have a few SSM commands run during [post test processing](https://github.com/aws/eks-anywhere/blob/b84f94b08010b5469927593473b36b4223071bbe/internal/test/e2e/run.go#L279), that timeout because the airgapped security group does not allow the outgoing connection. This increases the duration of the resulting codebuild  by at least 30 minutes (10 minute timeout per ssm command for`uploadJUnitReportFromInstance`, `uploadGeneratedFilesFromInstance` and `uploadDiagnosticArchiveFromInstance`). This PR adds a cleanup to docker air-gapped registry mirror e2e test, that resets the security group back to the default. This instead of relying on a successful test run to do so, this also captures in case of panic or some other kind of failure.

*Testing (if applicable):*
- Ran `TestDockerKubernetes127AirgappedUpgradeFromLatestRegistryMirrorAndCert` locally, observed the security group reset post test run.

### DIFF
--- a/test/e2e/airgap.go
+++ b/test/e2e/airgap.go
@@ -189,13 +189,14 @@ func runDockerAirgapConfigFlow(test *framework.ClusterE2ETest) {
 	test.ExtractDownloadedArtifacts()
 	test.DownloadImages()
 	test.ChangeInstanceSecurityGroup(os.Getenv(framework.RegistryMirrorAirgappedSecurityGroup))
+	test.SetRegistryMirrorDefaultInstanceSecurityGroupOnCleanup()
 	test.ImportImages()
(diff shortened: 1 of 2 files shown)


d8660091:
/LGTM, but keep a close eye on it as this might delay the network recovery, there is a chance that the successful test may not be able to upload the test logs if the admin machine is destroyed shortly after network recovery.

PR #7250, opened 2024-01-03
Add EKS-A management version annotation for cluster object

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -41,6 +41,9 @@ const (
 	// cluster object.
 	managementAnnotation = "anywhere.eks.amazonaws.com/managed-by"
 
+	// EKSAManagementComponentVersion is an annotation applied to an EKS-A management cluster pointing to the current version of EKS management components.
+	EKSAManagementComponentVersion = "anywhere.eks.amazonaws.com/eksa-management-component-version"
```
d8660091, line 45: looking at other field names, is it more consistent to name it "ManagementComponentsVersionAnnotation"

pkg/clustermanager/cluster_manager.go
```diff
@@ -1395,3 +1395,19 @@ func (c *ClusterManager) buildSpecForCluster(ctx context.Context, clus *types.Cl
 func (c *ClusterManager) DeletePackageResources(ctx context.Context, managementCluster *types.Cluster, clusterName string) error {
 	return c.clusterClient.DeletePackageResources(ctx, managementCluster, clusterName)
 }
+
+// UpdateEKSAManagementComponentVersion updates the eksa-management-component-version annotation for the managment cluster.
+func (c *ClusterManager) UpdateEKSAManagementComponentVersion(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec) error {
```
d8660091, line 1400: Update*Annotation

release/api/v1alpha1/bundle_test.go
```diff
@@ -0,0 +1,30 @@
...
+	tests := map[string]struct {
+		version string
+		want    string
+	}{
+		"DevBuildBundleVersion": {
+			version: "v0.0.0-dev+build.8033+04134fa",
+			want:    "v0.0.0-dev+build.8033",
+		},
+		"ProdBundleVersion": {
+			version: "v0.18.3+cc70180",
```
d8660091, line 19: Q: Have you figured out what this hash means?

cxbrowne1207: Yes, spoke to @abhay-krishna it's a hash of all the manifest contents in that bundle and we probably don't need it in the annotation.

pkg/clustermanager/cluster_manager.go
cxbrowne1207: Do you mean `UpdateEKSAManagementComponentVersionAnnotation`?

pkg/api/v1alpha1/cluster_types.go
cxbrowne1207: Ah, good catch. I had meant to have the`Annotation` suffix there

pkg/clustermanager/cluster_manager.go
d8660091: UpdateManagementComponent**s**VersionAnnotation, if you changed the const

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -41,6 +41,9 @@ const (
 	// cluster object.
 	managementAnnotation = "anywhere.eks.amazonaws.com/managed-by"
 
+	// ManagementComponentsVersionAnnotation is an annotation applied to an EKS-A management cluster pointing to the current version of EKS management components.
```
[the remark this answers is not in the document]
cxbrowne1207: This was intended to store the version of the management components referred to in the `EKS-A bundle`, and using that to search for the correct `bundles` object from the cluster.

I see that it would make more sense to use the `EKSARelease` object name used for the management components here though.

release/api/v1alpha1/bundle.go
```diff
@@ -0,0 +1,16 @@
+package v1alpha1
+
+import (
+	"strings"
+)
+
+// ParseVersionFromBundleVersion returns only the version part of the bundle version string without the manifest hash at the end.
+// The bundle component version can be in the format: <version>+<manifest-content-hash> or
+// <version>+<build>+<manifest-content-hash> for the components in the dev build bundles.
```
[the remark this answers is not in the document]
cxbrowne1207: Do we use the same <version> for a EKS-A component across bundles? We had opted to use just the version because it seems that we don't.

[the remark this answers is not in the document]
g-gaston: what's the benefit then?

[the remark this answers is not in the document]
cxbrowne1207: if only the version part is important, this makes that point clear so we wouldn't have to worry about the hash part when doing version comparisons. Although, it does add the complexity of having to parse it to store it.

Seeing that we should store the version from the EKSARelease object instead though, we no longer need this.

pkg/clustermanager/cluster_manager.go
```diff
@@ -1395,3 +1395,19 @@ func (c *ClusterManager) buildSpecForCluster(ctx context.Context, clus *types.Cl
 func (c *ClusterManager) DeletePackageResources(ctx context.Context, managementCluster *types.Cluster, clusterName string) error {
 	return c.clusterClient.DeletePackageResources(ctx, managementCluster, clusterName)
 }
+
+// UpdateManagementComponentsVersionAnnotation updates the eksa-management-component-version annotation for the managment cluster.
+func (c *ClusterManager) UpdateManagementComponentsVersionAnnotation(ctx context.Context, managementCluster *types.Cluster, clusterSpec *cluster.Spec) error {
```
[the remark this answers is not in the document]
cxbrowne1207: Discussed offline. I chose to expose a k8s client, that would be used to apply the changes in the workflow instead of this.

pkg/clustermanager/cluster_manager.go
```diff
@@ -1395,3 +1395,19 @@ func (c *ClusterManager) buildSpecForCluster(ctx context.Context, clus *types.Cl
...
+	if clusterSpec.Cluster.IsManaged() {
+		return nil
+	}
+
+	eksaManagementComponentVersion := releasev1alpha1.ParseVersionFromBundleVersion(clusterSpec.RootVersionsBundle().Eksa.Version)
+	eksaManagementComponentsVersionAnnotation := map[string]string{v1alpha1.ManagementComponentsVersionAnnotation: eksaManagementComponentVersion}
```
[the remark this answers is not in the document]
cxbrowne1207: This has been updated it to use the version from the `EKSARelease` object instead.

pkg/api/v1alpha1/cluster_types.go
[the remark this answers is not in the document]
cxbrowne1207: Some clarification, did you mean the EKSARelease object name or the semantic version [here](https://github.com/aws/eks-anywhere/pull/7250#discussion_r1444635572) object name?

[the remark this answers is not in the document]
cxbrowne1207: Discussed offline, we'll go with the version.

pkg/workflows/interfaces/interfaces.go
```diff
@@ -3,14 +3,22 @@ package interfaces
...
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/providers"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
+// KubernetesClient allows to interact with the k8s api server.
+type KubernetesClient interface {
+	Apply(ctx context.Context, kubeconfigPath string, obj runtime.Object, opts ...kubernetes.KubectlApplyOption) error
```
g-gaston, line 19: why can't we use `kubernetes.Client`?

pkg/api/v1alpha1/cluster_types_test.go
```diff
@@ -3230,3 +3230,29 @@ func TestValidateCluster(t *testing.T) {
...
+	managmentClusterName := "mgmt-cluster"
+	c.SetManagedBy(managmentClusterName)
+
+	expectedManagmentComponentVersion := "v0.0.0-dev+build.0000"
+	c.SetManagmentComponentsVersion(expectedManagmentComponentVersion)
+
+	expectedAnnotations := map[string]string{"anywhere.eks.amazonaws.com/managed-by": managmentClusterName}
+
+	g.Expect(c.Annotations).To(Equal(expectedAnnotations))
+}
```
g-gaston, line 3258: nit: use table tests?

pkg/workflows/create.go
```diff
@@ -372,6 +376,13 @@ func (s *InstallEksaComponentsTask) Run(ctx context.Context, commandContext *tas
 		commandContext.SetError(err)
 		return &CollectDiagnosticsTask{}
 	}
+
+	commandContext.ClusterSpec.Cluster.SetManagmentComponentsVersion(commandContext.ClusterSpec.EKSARelease.Spec.Version)
+	if err := commandContext.Client.Apply(ctx, targetCluster.KubeconfigFile, commandContext.ClusterSpec.Cluster); err != nil {
```
g-gaston, line 381: we should preferable use ServerSideApply so we specify the field owner

pkg/workflows/management/install_new_components.go
```diff
@@ -25,6 +25,12 @@ func runInstallNewComponents(ctx context.Context, commandContext *task.CommandCo
 		commandContext.SetError(err)
 		return err
 	}
+
+	commandContext.ClusterSpec.Cluster.SetManagmentComponentsVersion(commandContext.ClusterSpec.EKSARelease.Spec.Version)
+	if err := commandContext.Client.Apply(ctx, commandContext.ManagementCluster.KubeconfigFile, commandContext.ClusterSpec.Cluster); err != nil {
```
g-gaston, line 30: same as server side

g-gaston, line 30: Why wouldn't we do this in the "upgrade management components" task?
It seems that would be a more obvious place. Since we made this change only after updating those components.

cxbrowne1207: If I understand this correctly, I think you're referring to the `upgradeCoreComponents` task. I've considered this a bit more and I agree with this point, so I've moved it there.

pkg/workflows/interfaces/interfaces.go
cxbrowne1207: we can use `kubernetes.Client` instead. I think I was running into an import cycle at some point when I tried to use it at first, however, that must have been related to something else.

pkg/workflows/create.go
```diff
@@ -16,7 +17,12 @@ import (
 	"github.com/aws/eks-anywhere/pkg/workflows/interfaces"
 )
 
+const (
+	eksaFieldManager = "eks-a-cli"
```
g-gaston, line 21: maybe worth adding it to `constants`?

pkg/api/v1alpha1/cluster_types_test.go
```diff
@@ -3230,3 +3230,37 @@ func TestValidateCluster(t *testing.T) {
...
+			cluster: baseCluster(func(c *v1alpha1.Cluster) {
+				c.SetManagmentComponentsVersion(expectedManagmentComponentVersion)
+			}),
+			wantAnnotations: map[string]string{"anywhere.eks.amazonaws.com/eksa-management-components-version": expectedManagmentComponentVersion},
+		},
+		{
+			name: "managed cluster",
+			cluster: baseCluster(func(c *v1alpha1.Cluster) {
+				c.SetManagedBy(managmentClusterName)
+				c.SetManagmentComponentsVersion(expectedManagmentComponentVersion)
```
g-gaston, line 3254: I think the test would be more clear if you called this in the body of the test instead of the test case
it makes easier to understand what you are testing
The test case specifies input and out output
The test body passed the input through a method and compares the real output with the expected output.

pkg/workflows/create.go
cxbrowne1207: yeah, sounds good. I'll move it there.

pkg/api/v1alpha1/cluster_types_test.go
cxbrowne1207: Updated

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -41,6 +41,11 @@ const (
 	// cluster object.
 	managementAnnotation = "anywhere.eks.amazonaws.com/managed-by"
 
+	// managementComponentsVersionAnnotation is an annotation applied to an EKS-A management cluster pointing to the current version of the management components.
+	// The value for this annotation is expected to correspond to an EKSARelease object version, following semantic version convention: e.g. v0.18.3
+	// This is an internal EKS-A managed annotation, not meant to be updated manually.
+	managementComponentsVersionAnnotation = "anywhere.eks.amazonaws.com/eksa-management-components-version"
```
g-gaston, line 47: i think there was a comment about dropping eksa?
```suggestion
	managementComponentsVersionAnnotation = "anywhere.eks.amazonaws.com/management-components-version"
	
no strong opinion here just making sure we had an agreement there
```

pkg/constants/constants.go
```diff
@@ -99,3 +99,6 @@ const (
 	Upgrade Operation = 1
 	Delete  Operation = 2
 )
+
+// EKSAFieldManager is the owner name for fields applied by the EKS-A CLI.
+const EKSAFieldManager = "eks-a-cli"
```
g-gaston, line 104: nit
```suggestion
const EKSACLIFieldManager = "eks-a-cli"
```

pkg/workflows/create.go
```diff
@@ -372,6 +377,23 @@ func (s *InstallEksaComponentsTask) Run(ctx context.Context, commandContext *tas
...
+
+	commandContext.ClusterSpec.Cluster.SetManagementComponentsVersion(commandContext.ClusterSpec.EKSARelease.Spec.Version)
+	if err := client.ApplyServerSide(ctx,
+		constants.EKSAFieldManager,
+		commandContext.ClusterSpec.Cluster,
+		kubernetes.ApplyServerSideOptions{ForceOwnership: true},
+	); err != nil {
+		commandContext.SetError(err)
+		return &CollectDiagnosticsTask{}
+	}
```
g-gaston, line 395: @mitalipaygude this overlaps with your PR, just putting you two in contact to make sure we don't miss this

g-gaston (commented):
just some minor comments to make sure we don't miss stuff but overall lgtm

pkg/api/v1alpha1/cluster_types.go
cxbrowne1207: Ah yes, agreed. It's redundant

cxbrowne1207: Oh, @d8660091 had mentioned something similar as well, but I thought he was mainly referring to the variable name. Anyway, I've dropped the `eksa` here, which follows convention followed by most of the other annotations as well.

pkg/workflows/management/core_components.go
```diff
@@ -92,10 +92,13 @@ func runUpgradeCoreComponents(ctx context.Context, commandContext *task.CommandC
...
+
+	// ManagedFields must be nil to perform the server-side apply or a BadRequest error is thrown.
+	commandContext.CurrentClusterSpec.Cluster.ManagedFields = nil
+	commandContext.CurrentClusterSpec.Cluster.SetManagementComponentsVersion(commandContext.CurrentClusterSpec.EKSARelease.Spec.Version)
 	if err := client.ApplyServerSide(ctx,
-		constants.EKSAFieldManager,
-		commandContext.ClusterSpec.Cluster,
+		constants.EKSACLIFieldManager,
+		commandContext.CurrentClusterSpec.Cluster,
 		kubernetes.ApplyServerSideOptions{ForceOwnership: true},
```
cxbrowne1207, line 102: Updated this. It should have been using the current cluster spec, we don't want to apply any other cluster updates here.

pkg/workflows/management/core_components.go
```diff
@@ -92,10 +92,13 @@ func runUpgradeCoreComponents(ctx context.Context, commandContext *task.CommandC
 		commandContext.SetError(err)
 		return err
 	}
-	commandContext.ClusterSpec.Cluster.SetManagementComponentsVersion(commandContext.ClusterSpec.EKSARelease.Spec.Version)
+
+	// ManagedFields must be nil to perform the server-side apply or a BadRequest error is thrown.
+	commandContext.CurrentClusterSpec.Cluster.ManagedFields = nil
```
cxbrowne1207, line 97: Unsure how i feel about this one, as it could be a problem if `ManagedFields` is used later on in the workflow for something. However, since it's typically used for kubernetes internal house keeping and server side apply may update this anyway (meaning it would have to be read again), I think this does not pose an issue.

pkg/workflows/management/core_components.go
```diff
@@ -85,6 +90,40 @@ func runUpgradeCoreComponents(ctx context.Context, commandContext *task.CommandC
...
+
+	eksaCluster.SetManagementComponentsVersion(commandContext.ClusterSpec.EKSARelease.Spec.Version)
+
+	resourcesSpec, err := yaml.Marshal(eksaCluster)
+	if err != nil {
+		commandContext.SetError(err)
+		return err
+	}
+
+	logger.V(6).Info(string(resourcesSpec))
```
g-gaston, line 114: do we need this? or was it for debugging

cxbrowne1207: thanks for catching, yeah it was for debugging

g-gaston (commented):
/lgtm

PR #7251, opened 2024-01-04
Add e2e test for upgrading management-components

d8660091:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):* Manually invoked the new e2e test, and it passed.

### DIFF
--- a/test/e2e/docker_test.go
+++ b/test/e2e/docker_test.go
@@ -1334,3 +1334,15 @@ func TestDockerKubernetesRegionalCuratedPackages(t *testing.T) {
 		}
 	})
 }
+
+func TestDockerKubernetesUpgradeManagementComponents(t *testing.T) {
+	release := latestMinorRelease(t)
+	test := framework.NewClusterE2ETest(t, framework.NewDocker(t))
+
+	test.GenerateClusterConfig()
+	// create cluster with old eksa
+	test.CreateCluster(framework.ExecuteWithEksaRelease(release))
+	// upgrade management-components with new eksa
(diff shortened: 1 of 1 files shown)


test/e2e/docker_test.go
```diff
@@ -1334,3 +1334,15 @@ func TestDockerKubernetesRegionalCuratedPackages(t *testing.T) {
...
+
+func TestDockerKubernetesUpgradeManagementComponents(t *testing.T) {
+	release := latestMinorRelease(t)
+	test := framework.NewClusterE2ETest(t, framework.NewDocker(t))
+
+	test.GenerateClusterConfig()
+	// create cluster with old eksa
+	test.CreateCluster(framework.ExecuteWithEksaRelease(release))
+	// upgrade management-components with new eksa
+	test.RunEKSA([]string{"upgrade", "management-components", "-f", test.ClusterConfigLocation, "-v", "99"})
```
cxbrowne1207, line 1346: nit: I think that we should we create a `test.UpgradeManagmentComponents` function and hide these details.

cxbrowne1207 (commented):
Do we also need this to test completing the rest of the upgrade after running upgrade management components?

d8660091:
> Do we also need this to test completing the rest of the upgrade after running upgrade management components?

Because the followup procedure of upgrading the cluster itself will fail due to the bug: https://github.com/aws/eks-anywhere-internal/issues/2126. And I think right now we need a test case to cover the most essential flow of the new subcommand more than we need a comprehensive one that verify the flow of eksa controller reconciliation process. Having said that, we can do this in another PR when the bug is fixed

PR #7262, opened 2024-01-05
Remove ExistingManagement field from Cluster type

g-gaston:
*Description of changes:*
The function of this field was unclear and its name was confusing to say
the least. It seems like we now have a better way to achieve the same
thing and the resulting code is simpler and more expressive.

### DIFF
--- a/pkg/gitops/flux/flux.go
+++ b/pkg/gitops/flux/flux.go
@@ -126,7 +126,7 @@ func (f *Flux) Bootstrap(ctx context.Context, cluster *types.Cluster, clusterSpe
 }
 
 func (f *Flux) BootstrapGithub(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
-	if cluster.ExistingManagement || clusterSpec.FluxConfig.Spec.Github == nil {
+	if clusterSpec.Cluster.IsManaged() || clusterSpec.FluxConfig.Spec.Github == nil {
 		return nil
 	}
 
@@ -134,7 +134,7 @@ func (f *Flux) BootstrapGithub(ctx context.Context, cluster *types.Cluster, clus
 }
 
 func (f *Flux) BootstrapGit(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
-	if cluster.ExistingManagement || clusterSpec.FluxConfig.Spec.Git == nil {
+	if clusterSpec.Cluster.IsManaged() || clusterSpec.FluxConfig.Spec.Git == nil {
 		return nil
 	}
 
--- a/cmd/eksctl-anywhere/cmd/deletecluster.go
+++ b/cmd/eksctl-anywhere/cmd/deletecluster.go
@@ -134,13 +134,11 @@ func (dc *deleteClusterOptions) deleteCluster(ctx context.Context) error {
 		cluster = &types.Cluster{
 			Name:               clusterSpec.Cluster.Name,
 			KubeconfigFile:     kubeconfig.FromClusterName(clusterSpec.Cluster.Name),
-			ExistingManagement: false,
 		}
 	} else {
 		cluster = &types.Cluster{
 			Name:               clusterSpec.Cluster.Name,
 			KubeconfigFile:     clusterSpec.ManagementCluster.KubeconfigFile,
-			ExistingManagement: true,
 		}
 	}
 
--- a/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
+++ b/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
@@ -81,9 +81,8 @@ var upgradeManagementComponentsCmd = &cobra.Command{
 		)
 
 		managementCluster := &types.Cluster{
-			Name:               clusterSpec.Cluster.Name,
-			KubeconfigFile:     kubeconfig.FromClusterName(clusterSpec.Cluster.Name),
-			ExistingManagement: clusterSpec.Cluster.IsSelfManaged(),
+			Name:           clusterSpec.Cluster.Name,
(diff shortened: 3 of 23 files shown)


pkg/gitops/flux/flux.go
```diff
@@ -126,15 +126,15 @@ func (f *Flux) Bootstrap(ctx context.Context, cluster *types.Cluster, clusterSpe
 }
 
 func (f *Flux) BootstrapGithub(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
-	if cluster.ExistingManagement || clusterSpec.FluxConfig.Spec.Github == nil {
+	if clusterSpec.FluxConfig.Spec.Github == nil {
```
chrisdoherty4, line 129: I'm not entirely sure what `ExistingManagement` ever represented (it always felt hyperloaded) but I would've thought it needs replacing with a new thing here as opposed to removed?

g-gaston: From what I extracted from the code, `ExistingManagement` was set to true in management clusters once the management components were installed.

Since I couldn't a find a test that stressed this condition, I assumed it was unnecessary.

That said, you are right and maybe I shouldn't have assumed that fast. Unfortunately I lack context about this code and I can't seem to fully understand how this is orchestrated by just looking at the code.

maybe @jiayiwang7 you would know more? Was this necessary? And if so, what was trying to do?

jiayiwang7: I believe initially `ExistingManagement bool // true is the cluster has EKS Anywhere management components` and we use this field to determine if this cluster is a management cluster. We only bootstrap/install Flux controllers on the management cluster, not workload cluster. This condition is required to ensure that (if the `ExistingManagement` is still being used to determine if its a management cluster)

g-gaston: Does that mean that we can swap it for `cluster.IsSelfManaged()`?

What trips me about that is that this code was returning early if `ExistingManagement` was true. Which means we would never bootstrap management clusters. And that is the opposite of what you are describing.

chrisdoherty4: Side note: this conversation demonstrates why the overall change here is good - this field is confusing.

@vivek-koppuru @tatlat I see you added some code that used `ExistingManagement` in the `upgradevalidations/versions.go` for version skew handling.

How did you interpret `true` and `false` of `ExistingManagement`?

jiayiwang7: @g-gaston yes we can swap it for cluster.IsSelfManaged()

chrisdoherty4: Just a callout, there are 2 other logic changes in this change that probably need double checking.

tatlat: I interpreted  `ExistingManagement` as a way to check if a cluster was a management cluster based on its comment.  It's fine to replace the validations logic with `cluster.IsSelfManaged()`.

vivek-koppuru: Based on what @jiayiwang7 is saying and what the condition seems to show here, https://github.com/aws/eks-anywhere/blob/ed36457011ea3fdece63e295299ecd8c532a837a/cmd/eksctl-anywhere/cmd/deletecluster.go#L137, it seems as if `ExisitingManagement` means if there is a management cluster associated with the cluster in question, hence we are returning nil if we see that as true right? So in reality, the condition to replace with should be `!cluster.IsSelfManaged()`? But we need to change this because this is very confusing. The comment next to the field sounds misleading based on these conditions too.

chrisdoherty4: Is there any reason not to include `IsManagement()` function on `Cluster` (or a helper if we need to do something dynamic)?

`IsSelfManaged()` is a fantastic improvement on `ExistingManagement` but when developing I tend to find myself wanting to know if the `Cluster` is a management cluster and `IsSelfManaged` still feels like a convoluted way of asking that.

g-gaston: Ok i figured out. The meaning (or the use) of `ExistingManagement` changed based on what the target cluster is.

If we are creating a management cluster `ExistingManagement` is false.
If we are creating a workload cluster, `ExistingManagement` is true for the management cluster and false for the workload cluster.

If we are upgrading a management cluster `ExistingManagement` is false.
If we are creating a workload cluster, `ExistingManagement` is true for the management cluster and false for the workload cluster.

Based on that, @jiayiwang7 and I figured that the right logic here is to just skip the bootstrap operation for workload clusters. For management clusters, we want to run bootstrap no matter if it's an upgrade or a create (bootstrap is idempotent).

I adjusted the code and added a couple unit tests to protect this behavior.

g-gaston: > What do we think about including an IsManagement() method on Cluster (or a helper function if we need to do something dynamic)?

Technically `IsManagement` can't be determined thought the cluster spec. Because the cluster spec doesn't say if a cluster is managing other clusters (and a cluster needs to be managing other cluster in order to be a management). So `IsSelfManaged` is more accurate in this situation.

That said, in practice, we use these two concepts interchangeably. In fact, although not accurate, "management cluster" is a more common term, regardless if the cluster is managing workload clusters or not.

I'm in favor of trading correctness for expresiveness/understability. So if we consider that `IsSelfManaged` is convoluted, I would suggest just renaming the method as opposed to have two methods to do the same thing. I would prefer if this was addressed in a separate PR though, if y'all are ok with that.

jiayiwang7 (approved):

chrisdoherty4 (approved):

g-gaston:
Overriding codepatch manually: path coverage is at 76% percent but the only file not well covered is the upgrade workflow, which will be removed in the release cycle in favor of separate workflows for management and workload clusters. In fact this PR increases the overall coverage thanks to the flux tests.

PR #7267, opened 2024-01-05
Separate kubeconfig operations from cluster manager

tatlat:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/pull/7110#discussion_r1410769500

*Description of changes:*
These changes separate some functionality from ClusterManager into a new component which is solely responsible for writing the kubeconfig file. Unlike ClusterManager which used clusterctl to get the kubeconfig, the new component gets the kubeconfig content from the cluster's kubeconfig secret. Another higher level abstraction called ClusterCreator was also added which groups the Applier with the kubeconfig writer. 

*Testing (if applicable):*
Manual testing with docker and cloudstack. 


pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,54 @@
...
+func (cc ClusterCreator) CreateSync(ctx context.Context, spec *cluster.Spec, managementCluster *types.Cluster, provider providers.Provider) (*types.Cluster, error) {
+	err := cc.Applier.Run(ctx, spec, *managementCluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return cc.getWorkloadCluster(ctx, spec.Cluster.Name, managementCluster, provider)
+}
+
+func (cc ClusterCreator) getWorkloadCluster(ctx context.Context, clusterName string, management *types.Cluster, provider providers.Provider) (*types.Cluster, error) {
```
g-gaston, line 28:
```suggestion
func (cc ClusterCreator) buildClusterAccess(ctx context.Context, clusterName string, management *types.Cluster, provider providers.Provider) (*types.Cluster, error) {
```

pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,54 @@
...
+	workloadCluster := &types.Cluster{
```
g-gaston, line 29:
```suggestion
	cluster := &types.Cluster{
```

pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/types"
+)
+
+type ClusterCreator struct {
+	Applier Applier
+	KM      KubeconfigManager
```
g-gaston, line 15: in general, i advise to stay away from acronyms unless they refer to well known things
I don't think KM is a well known thing

pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,54 @@
...
+	return cc.getWorkloadCluster(ctx, spec.Cluster.Name, managementCluster, provider)
+}
+
+func (cc ClusterCreator) getWorkloadCluster(ctx context.Context, clusterName string, management *types.Cluster, provider providers.Provider) (*types.Cluster, error) {
+	workloadCluster := &types.Cluster{
+		Name:               clusterName,
+		ExistingManagement: management.ExistingManagement,
+	}
+
+	fh, path, err := cc.KM.FS.Create(
```
g-gaston, line 34: why is FS defined in KM if KM doesn't depend on it?
It seems to me that ClusterCreator is the one who is concerned about writing to disk. KM is only concerned about getting the kubeconfig and passing it to a writer.

pkg/clustermanager/kubeconfig_manager.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type KubeconfigManager struct {
+	ClientFactory ClientFactory
+	// FS is a file system abstraction providing file creation and write capabilities
+	FS filewriter.FileWriter
+}
+
+// UpdateKubeconfig retrieves the contents of the specified cluster's kubeconfig from a secret and copies it to an io.Writer.
+func (km KubeconfigManager) UpdateKubeconfig(ctx context.Context, clusterName, kubeconfig string, provider providers.Provider, w io.Writer) error {
```
g-gaston, line 25:
```suggestion
func (km KubeconfigManager) WriteKubeconfig(ctx context.Context, clusterName, kubeconfig string, provider providers.Provider, w io.Writer) error {
```

g-gaston: I would love if we could find a way to not pollute this signature with the provider

The only reason the provider is needed here is because docker needs to make some changes on the fly to the kubeconfig content

What if we just had a different `KubeconfigManager` implementation depending on the provider? That way the "selection" is made at dep build time and we don't need to pass the provider around.

pkg/clustermanager/kubeconfig_manager.go
```diff
@@ -0,0 +1,61 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/filewriter"
+	"github.com/aws/eks-anywhere/pkg/providers"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+)
+
+type KubeconfigManager struct {
```
g-gaston, line 18: Could we find a different name? manager is an overloaded term
plus it doesn't tell us a lot

maybe `KubeconfigReader`?

pkg/clustermanager/kubeconfig_manager.go
```diff
@@ -0,0 +1,61 @@
...
+	ClientFactory ClientFactory
+	// FS is a file system abstraction providing file creation and write capabilities
+	FS filewriter.FileWriter
```
g-gaston, line 21: why are these fields exported?

pkg/clustermanager/kubeconfig_manager.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// UpdateKubeconfig retrieves the contents of the specified cluster's kubeconfig from a secret and copies it to an io.Writer.
+func (km KubeconfigManager) UpdateKubeconfig(ctx context.Context, clusterName, kubeconfig string, provider providers.Provider, w io.Writer) error {
+	kubeconfigSecret := &corev1.Secret{}
+
+	err := retrier.New(
+		time.Minute,
+		retrier.WithRetryPolicy(retrier.BackOffPolicy(time.Second)),
```
g-gaston, line 30: nit: this should probably be configurable through the constructor

pkg/workflows/workload/create.go
```diff
@@ -17,17 +18,17 @@ type Create struct {
 	gitOpsManager    interfaces.GitOpsManager
 	writer           filewriter.FileWriter
 	eksdInstaller    interfaces.EksdInstaller
-	clusterCreator   interfaces.ClusterCreator
+	clusterCreator   clustermanager.ClusterCreator
```
g-gaston, line 21: why are we changing this?

pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = cc.KM.UpdateKubeconfig(ctx, clusterName, management.KubeconfigFile, provider, fh)
+	if err != nil {
+		return nil, err
+	}
+
+	defer fh.Close()
```
g-gaston, line 49: this should be called immediately after FS.Create

pkg/workflows/workload/createcluster.go
```diff
@@ -13,7 +13,7 @@ type createCluster struct{}
 // Run createCluster performs actions needed to create the management cluster.
 func (c *createCluster) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
 	logger.Info("Creating workload cluster")
-	if err := commandContext.ClusterCreator.Run(ctx, commandContext.ClusterSpec, *commandContext.ManagementCluster); err != nil {
+	if err := commandContext.ClusterCreator.Applier.Run(ctx, commandContext.ClusterSpec, *commandContext.ManagementCluster); err != nil {
```
g-gaston, line 16: I would prefer if you exposed Run instead of exposing Applier

pkg/clustermanager/kubeconfig_manager.go
tatlat: they don't need to be

pkg/workflows/workload/create.go
tatlat: tbh I don't remember why I changed it. I was probably just testing something. changed it back.

pkg/clustermanager/cluster_creator.go
```diff
@@ -0,0 +1,69 @@
...
+		applier:          applier,
+		kubeconfigReader: kubeconfigReader,
+		fs:               fs,
+	}
+}
+
+// CreateSync creates a workload cluster using the EKS-A controller and returns the types.Cluster object for that cluster.
+func (cc ClusterCreator) CreateSync(ctx context.Context, spec *cluster.Spec, managementCluster *types.Cluster) (*types.Cluster, error) {
+	err := cc.applier.Run(ctx, spec, *managementCluster)
+	if err != nil {
```
g-gaston, line 32:
```suggestion
	if err := cc.applier.Run(ctx, spec, *managementCluster); err != nil {
```

pkg/dependencies/factory.go
```diff
@@ -580,6 +581,27 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+		f.dependencies.KubeconfigReader = genericKubeconfigReader
+	case v1alpha1.DockerDatacenterKind:
+		f.dependencies.KubeconfigReader = docker.BuildKubeconfigReader(f.dependencies.Provider, f.dependencies.UnAuthKubeClient, time.Minute, time.Second)
+	case v1alpha1.TinkerbellDatacenterKind:
+		f.dependencies.KubeconfigReader = genericKubeconfigReader
+	case v1alpha1.SnowDatacenterKind:
+		f.dependencies.KubeconfigReader = genericKubeconfigReader
+	case v1alpha1.NutanixDatacenterKind:
+		f.dependencies.KubeconfigReader = genericKubeconfigReader
+	}
```
g-gaston, line 600: Can we simplify this to
```suggestion
	switch clusterConfig.Spec.DatacenterRef.Kind {
	case v1alpha1.DockerDatacenterKind:
		f.dependencies.KubeconfigReader = docker.BuildKubeconfigReader(f.dependencies.Provider, f.dependencies.UnAuthKubeClient, time.Minute, time.Second)
	default:
		f.dependencies.KubeconfigReader = genericKubeconfigReader
	}
```

pkg/providers/docker/docker.go
```diff
@@ -52,6 +58,14 @@ type provider struct {
 	templateBuilder       *DockerTemplateBuilder
 }
 
+// KubeconfigReader reads the kubeconfig secret on a cluster and copies the contents to a writer.
+type KubeconfigReader struct {
+	provider providers.Provider
```
g-gaston, line 63: Do we actually need to depend on the provider?
`UpdateKubeconfig` is a simple method
Can we just move/copy it here

If I'm not mistaken these changes you propose here will render UpdateKubeconfig in the provider useless. That's actually a great thing bc we can just remove that method from all providers

pkg/providers/docker/docker.go
```diff
@@ -52,6 +58,14 @@ type provider struct {
...
+	client   *kubernetes.UnAuthClient
```
g-gaston, line 64: note for final version: this should be an interface

pkg/providers/docker/docker.go
```diff
@@ -531,6 +545,56 @@ func (p *provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
 	}
 }
 
+// BuildKubeconfigReader creates a KubeconfigReader.
+func BuildKubeconfigReader(provider providers.Provider, unauthClient *kubernetes.UnAuthClient, timeout time.Duration, backoff time.Duration) providers.KubeconfigReader {
+	return KubeconfigReader{
+		client:   unauthClient,
+		timeout:  timeout,
+		backoff:  backoff,
+		provider: provider,
```
g-gaston, line 554: can we make these two optional and have default values?

pkg/providers/docker/docker.go
```diff
@@ -531,6 +545,56 @@ func (p *provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
...
+		}
+
+		err = client.Get(ctx, fmt.Sprintf("%s-kubeconfig", clusterName), constants.EksaSystemNamespace, kubeconfigSecret)
+		if err != nil {
+			return err
+		}
+
+		if err != nil {
+			return err
+		}
```
g-gaston, line 578:
```suggestion
		if err != nil {
			return err
		}
```

pkg/providers/docker/docker.go
```diff
@@ -531,6 +545,56 @@ func (p *provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
...
+	err := retrier.New(
+		kr.timeout,
+		retrier.WithRetryPolicy(retrier.BackOffPolicy(kr.backoff)),
+	).Retry(func() error {
+		client, err := kr.client.BuildClientFromKubeconfig(kubeconfigPath)
+		if err != nil {
+			return err
+		}
+
+		err = client.Get(ctx, fmt.Sprintf("%s-kubeconfig", clusterName), constants.EksaSystemNamespace, kubeconfigSecret)
```
g-gaston, line 571: We can probably create a method in the clusterapi package, like `KubeconfigSecretName` so there is only one place that knows how to generate this name from the cluster name

pkg/providers/kubeconfig_reader.go
```diff
@@ -0,0 +1,66 @@
...
+	timeout time.Duration
+	backoff time.Duration
+}
+
+// BuildKubeconfigReader creates a KubeconfigReader.
+func BuildKubeconfigReader(unauthClient *kubernetes.UnAuthClient, timeout time.Duration, backoff time.Duration) KubeconfigReader {
+	return GenericKubeconfigReader{
+		client:  unauthClient,
+		timeout: timeout,
+		backoff: backoff,
```
g-gaston, line 29: same comment about making these two optional

pkg/providers/kubeconfig_reader.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		err = client.Get(ctx, fmt.Sprintf("%s-kubeconfig", clusterName), constants.EksaSystemNamespace, kubeconfigSecret)
+		if err != nil {
+			return err
+		}
+
+		if err != nil {
+			return err
+		}
```
g-gaston, line 53: duplicated

pkg/providers/provider.go
```diff
@@ -66,3 +67,8 @@ type MachineConfig interface {
 	GetNamespace() string
 	GetName() string
 }
+
+// KubeconfigReader reads the kubeconfig secret on a cluster and copies the contents to a writer.
+type KubeconfigReader interface {
```
g-gaston, line 72: Why do we define this in the provider package?
Ideally we would only define it in the consumer of this interface
If for some reason we need a shared definition, then the kubeconfig package sounds like the right place

pkg/providers/provider.go
```diff
@@ -66,3 +67,8 @@ type MachineConfig interface {
...
+	WriteKubeconfig(ctx context.Context, clusterName, kubeconfig string, w io.Writer) error
```
g-gaston, line 73: just realized the struct is named reader but its only method is a write

pkg/providers/provider.go
tatlat: moved it to kubeconfig

pkg/providers/provider.go
tatlat: renamed to KubeconfigWriter

pkg/providers/docker/docker.go
tatlat: We don't need the entire provider. I changed it to depend on docker's `ProviderClient` instead.

cmd/eksctl-anywhere/cmd/createcluster.go
```diff
@@ -232,6 +233,8 @@ func (cc *createClusterOptions) createCluster(cmd *cobra.Command, _ []string) er
 	}
 	createValidations := createvalidations.New(validationOpts)
 
+	clusCreator := clustermanager.NewClusterCreator(deps.ClusterApplier, deps.KubeconfigReader, deps.Writer)
```
g-gaston, line 236: add this to the factory?

pkg/clustermanager/cluster_creator_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/aws/eks-anywhere/pkg/clustermanager"
+	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/controller/clientutil"
+	mockswriter "github.com/aws/eks-anywhere/pkg/filewriter/mocks"
+	"github.com/aws/eks-anywhere/pkg/kubeconfig"
+)
+
+type clusterCreatorTest struct {
+	*WithT
+	*applierTest
```
g-gaston, line 26: why don't you decouple the creator from the applier?
That way you don't need use this applier test helpers. Just mock the Applier.Run call.

pkg/dependencies/factory_test.go
```diff
@@ -86,6 +89,44 @@ func newTest(t *testing.T, p provider) *factoryTest {
...
+	}{
+		{
+			provider: docker,
+		},
+		{
+			provider: vsphere,
+		},
+		{
+			provider: docker,
+			exists:   true,
```
g-gaston, line 105: nit: this makes me think we want two different tests

pkg/kubeconfig/kubeconfig_writer.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// ClientFactory builds Kubernetes clients.
+type ClientFactory interface {
+	// BuildClientFromKubeconfig builds a Kubernetes client from a kubeconfig file.
+	BuildClientFromKubeconfig(kubeconfigPath string) (kubernetes.Client, error)
+}
+
+// GenericKubeconfigWriter reads the kubeconfig secret on a cluster and copies the contents to a writer.
+type GenericKubeconfigWriter struct {
```
g-gaston, line 26: This name is not very informative. Generic doesn't provide any new info and it's a term that in software is used for more than one thing, so it's ambiguous.

What about `ClusterAPIKubeconfigSecretWriter`?

pkg/kubeconfig/kubeconfig_writer.go
```diff
@@ -0,0 +1,113 @@
...
+		writer.backoff = &backoff
+	}
+}
+
+func defaultKubeconfigWriterOpts() []WriterOpt {
+	return []WriterOpt{WithTimeout(time.Minute), WithBackoff(time.Second)}
+}
+
+// BuildKubeconfigWriter creates a KubeconfigWriter.
+func BuildKubeconfigWriter(unauthClient ClientFactory, opts ...WriterOpt) GenericKubeconfigWriter {
```
g-gaston, line 54: for consistency with other constructors
```suggestion
func NewGenericKubeconfigWriter(unauthClient ClientFactory, opts ...WriterOpt) GenericKubeconfigWriter {
```

pkg/kubeconfig/kubeconfig_writer.go
```diff
@@ -0,0 +1,113 @@
...
+	kr := &GenericKubeconfigWriter{
+		client: unauthClient,
+	}
+
+	if len(opts) == 0 {
+		opts = defaultKubeconfigWriterOpts()
+	}
```
g-gaston, line 61: Why not set the default values directly on the struct and let the options overwrite those values?

pkg/providers/docker/docker.go
```diff
@@ -52,6 +55,12 @@ type provider struct {
 	templateBuilder       *DockerTemplateBuilder
 }
 
+// ProviderKubeconfigWriter reads the kubeconfig secret on a docker cluster and copies the contents to a writer.
+type ProviderKubeconfigWriter struct {
+	docker ProviderClient
+	kubeconfig.GenericKubeconfigWriter
```
g-gaston, line 61: we probably want an interface here

pkg/kubeconfig/kubeconfig_writer.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if _, err := io.Copy(w, bytes.NewReader(rawKubeconfig)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetKubeconfigSecretValue gets the cluster's kubeconfig from the secret.
+func (kr GenericKubeconfigWriter) GetKubeconfigSecretValue(ctx context.Context, clusterName, kubeconfigPath string) ([]byte, error) {
```
g-gaston, line 85:
```suggestion
func (kr GenericKubeconfigWriter) GetClusterKubeconfig(ctx context.Context, clusterName, kubeconfigPath string) ([]byte, error) {
```

pkg/providers/docker/docker.go
```diff
@@ -531,6 +540,35 @@ func (p *provider) UpdateKubeConfig(content *[]byte, clusterName string) error {
...
+	if err != nil {
+		return err
+	}
+
+	port, err := kr.docker.GetDockerLBPort(ctx, clusterName)
+	if err != nil {
+		return err
+	}
+
+	getUpdatedKubeConfigContent(&rawkubeconfig, port)
```
g-gaston, line 563: nit: let's rename this method. `get` makes no sense

pkg/clustermanager/cluster_creator_test.go
```diff
@@ -0,0 +1,69 @@
...
+func (cct *clusterCreatorTest) expectFileCreate() {
+	cct.writer.EXPECT().Create(gomock.Any(), gomock.Any(), gomock.Any()).Return(os.NewFile(uintptr(*pointer.Uint(1)), "test"), "", nil)
+}
+
+func (cct *clusterCreatorTest) expectWriteKubeconfig() {
+	cct.kubeconfigWriter.EXPECT().WriteKubeconfig(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+}
+
+func (cct *clusterCreatorTest) expectApplierRun() {
+	cct.applier.EXPECT().Run(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
g-gaston, line 59: is there a way we can expect a bit more than any?
there is a bit more logic in this code than just calling those methods
The methods need to be called with the right arguments

pkg/dependencies/factory.go
```diff
@@ -580,6 +582,39 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
 	return f
 }
 
+// WithKubeconfigWriter adds the KubeconfigReader dependency depending on the provider.
+func (f *Factory) WithKubeconfigWriter(clusterConfig *v1alpha1.Cluster) *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.dependencies.KubeconfigWriter != nil {
+			return nil
+		}
+		writer := kubeconfig.NewClusterAPIKubeconfigSecretWriter(f.dependencies.UnAuthKubeClient)
```
g-gaston, line 591: shouldn't unauthclient be defined as a dep?

pkg/dependencies/factory.go
```diff
@@ -580,6 +582,39 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+		switch clusterConfig.Spec.DatacenterRef.Kind {
+		case v1alpha1.DockerDatacenterKind:
+			f.dependencies.KubeconfigWriter = docker.NewKubeconfigWriter(f.dependencies.DockerClient, writer)
```
g-gaston, line 594: shouldn't DockerClient be defined as a conditional dependency?

pkg/dependencies/factory.go
```diff
@@ -580,6 +582,39 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+}
+
+// WithClusterCreator adds the ClusterCreator dependency.
+func (f *Factory) WithClusterCreator(clusterConfig *v1alpha1.Cluster) *Factory {
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.dependencies.ClusterCreator != nil {
+			return nil
+		}
+
+		f.dependencies.ClusterCreator = clustermanager.NewClusterCreator(f.dependencies.ClusterApplier, f.dependencies.KubeconfigWriter, f.dependencies.Writer)
```
g-gaston, line 611: these 3 need to be defined as dependencies

pkg/kubeconfig/kubeconfig.go
```diff
@@ -12,6 +14,12 @@ import (
 	"github.com/aws/eks-anywhere/pkg/validations"
 )
 
+// Writer reads the kubeconfig secret on a cluster and copies the contents to a writer.
+type Writer interface {
+	WriteKubeconfig(ctx context.Context, clusterName, kubeconfig string, w io.Writer) error
+	GetClusterKubeconfig(ctx context.Context, clusterName, kubeconfigPath string) ([]byte, error)
+}
```
g-gaston, line 21: Where is this interface required?
If I'm not mistaken the `WriteKubecongig` is required by a few consumers and the `GetClusterKubeconfig` by other

But there is no consumer that needs both. That's why it's better to define the interface where they are consumed when you can (which is most of the time)

pkg/providers/docker/docker.go
```diff
@@ -52,6 +55,12 @@ type provider struct {
 	templateBuilder       *DockerTemplateBuilder
 }
 
+// ProviderKubeconfigWriter reads the kubeconfig secret on a docker cluster and copies the contents to a writer.
+type ProviderKubeconfigWriter struct {
```
g-gaston, line 59:
```suggestion
type KubeconfigWriter struct {
```

The package name-spacing already adds this context

pkg/providers/docker/docker.go
```diff
@@ -52,6 +55,12 @@ type provider struct {
...
+	docker ProviderClient
+	kubeconfig.Writer
```
g-gaston, line 61: if the 	kubeconfig.Writer wasn't broader than it needs you wouldn't need to embed this here and you could keep it as a unexported member

pkg/clustermanager/cluster_creator_test.go
tatlat: I'm not sure how to expect something specific for second and third arguments for filewriter's Create method, but I changed everything else.

mitalipaygude:
Can we update the description and squash the commits?

pkg/dependencies/factory.go
```diff
@@ -580,6 +582,42 @@ func (f *Factory) WithProvider(clusterConfigFile string, clusterConfig *v1alpha1
...
+func (f *Factory) WithKubeconfigWriter(clusterConfig *v1alpha1.Cluster) *Factory {
+	f.WithUnAuthKubeClient()
+	f.buildSteps = append(f.buildSteps, func(ctx context.Context) error {
+		if f.dependencies.KubeconfigWriter != nil {
+			return nil
+		}
+		writer := kubeconfig.NewClusterAPIKubeconfigSecretWriter(f.dependencies.UnAuthKubeClient)
+		switch clusterConfig.Spec.DatacenterRef.Kind {
+		case v1alpha1.DockerDatacenterKind:
+			f.WithDocker()
```
g-gaston, line 595: you need to call `WithDocker` from outside of the build step

the factory more or less works like this

1. Define deps
1.1 Your Deps define their own deps
2. Traverse the tree that has been declared by the dependencies, building them one by one. This uses DFS, which guarantees an object dependencies are built before itself

If you define the deps during the build process is like not declaring them at all

Let me know if this requires a lengthier explanation and we can chat offline. This can be tricky and it's important to get it right.

pkg/clustermanager/cluster_creator_test.go
```diff
@@ -0,0 +1,75 @@
...
+	cct.applier.EXPECT().Run(ctx, spec, managementCluster).Return(nil)
+}
+
+func TestClusterCreatorCreateSync(t *testing.T) {
+	clusCreator, tt := newClusterCreator(t)
+	clusterName := tt.spec.Cluster.Name
+	path := "testpath"
+	writer := os.NewFile(uintptr(*pointer.Uint(0)), "test")
+	tt.expectApplierRun(tt.ctx, tt.spec, *tt.mgmtCluster)
+	tt.expectWriteKubeconfig(tt.ctx, clusterName, tt.mgmtCluster.KubeconfigFile, writer)
```
g-gaston, line 71: nit: it's weird to pass members of the receiver as arguments to that receivers method

pkg/clustermanager/cluster_creator_test.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+	}
+
+	cc := clustermanager.NewClusterCreator(cct.applier, cct.kubeconfigWriter, cct.writer)
+
+	return cc, cct
+}
+
+func (cct *clusterCreatorTest) expectFileCreate(clusterName, path string, w io.WriteCloser) {
+	cct.writer.EXPECT().Create(kubeconfig.FormatWorkloadClusterKubeconfigFilename(clusterName), gomock.AssignableToTypeOf([]filewriter.FileOptionsFunc{})).Return(w, path, nil)
```
g-gaston, line 54: nit: i think it's better to hardcode here the value that you expect, considering that we control the cluster name, we should be able to do so

g-gaston (commented):
/lgtm

PR #7280, opened 2024-01-10
Update in-place upgrade design

abhinavmpandey08:
*Description of changes:*
This PR aims to update the in-place upgrade design doc to reflect the changes in the current design.
Here are some of the main changes in the designs that we made and their justification:

- Relying on annotations on `KCP` and `MD` objects instead of CAPI runtime extensions to trigger in-place upgrades
  - This was changed because the runtime extensions needed for this feature aren't defined in upstream CAPI yet but it will be added when in-place is officially supported upstream. Until then we would have to maintain a patch in our CAPI to add these runtime extensions which may or may not match what upstream CAPI would have. 
    Therefore the initial investment in adding runtime extensions doesn't seem justified considering we can achieve similar functionality by using annotations, which is a 2 way door decision. Once the runtime extension support for in-place upgrade is officially added in upstream CAPI, we can migrate to it without a lot of throwaway code.

- Another decision that was changed was to remove unnecessary fields from the `ControlPlaneUpgrade`, `MachineDeploymentUpgrade` and the `NodeUpgrade` APIs.
  - This was done because we can gather all the information that we need from the KCP/MD ref including the machines that need upgrade, the new Kubernetes version that the machines need to be upgraded to, and versions of other components that need to be upgraded.

- Lastly, the other major change was `NodeUpgrade` controller deciding what commands to execute on a node based on the node type (cp vs worker) instead of being passed a list of commands.
  - This was done because most of the commands were generic between control plane node and worker nodes. The only exception was the `kubeadm upgrade`. So to account for this, the `NodeUpgrade` object will accept a `NodeType` field to signify whether the node to be upgraded should be a cp or a worker. Then the `ControlPlaneUpgrade` and the `MachineDeploymentUpgrade` controller can just pass the `NodeType` down without having to worry about what the exact implementation of what upgrading a control plane node or a worker node would entail, as those details are abstracted away.

### DIFF
--- a/designs/in-place-upgrades-eksa.md
+++ b/designs/in-place-upgrades-eksa.md
@@ -2,7 +2,7 @@
 ## Introduction
 At present, the only supported upgrade strategy in EKS-A is rolling update. However, for certain use cases (such as Single-Node Clusters with no spare capacity, Multi-Node Clusters with VM/OS customizations, etc.), upgrading a cluster via a Rolling Update strategy could either be not feasible or a costly operation (requiring to add new hardware, re-apply customizations...).
 
-In-place upgrades aims to solve this problem by allowing users to perform Kubernetes node upgrades without replacing the underlying machines.
+In-place upgrades aim to solve this problem by allowing users to perform Kubernetes node upgrades without replacing the underlying machines.
 
 In a [previous proposal](in-place-upgrades-capi.md) we defined how a pluggable upgrade strategy allows to implement in-place upgrades with Cluster API. In this doc we'll describe how we will leverage and implement such architecture to offer in-place upgrades in EKS-A.
 
@@ -21,7 +21,7 @@ The goal is to bring enough clarity to the table that a general solution can be 
 * External etcd
 
 ## Overview of the Solution
-**TLDR**: the eks-a controller manager will host the two webhook servers and implement 3 new controllers that will orchestrate the upgrade. These controllers will schedule privilege pods on each node to be upgraded, that will execute the upgrade logic as a sequence of containers.
+**TLDR**: the eks-a controller manager will watch KCP and MD objects for "in-place-upgrade-needed" annotation and implement 3 new controllers that will orchestrate the upgrade. These controllers will schedule privilege pods on each node to be upgraded, that will execute the upgrade logic as a sequence of containers.
 
 ### High level view
 Following the CAPI external upgrade strategy idea, we can start with the following diagram.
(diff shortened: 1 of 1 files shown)


designs/in-place-upgrades-eksa.md
```diff
@@ -79,14 +83,51 @@ There are a few ways to go about this:
...
+	RequireUpgrade int64          `json:"requireUpgrade,omitempty"`
+	Upgraded       int64          `json:"upgraded,omitempty"`
+	Ready          bool           `json:"ready,omitempty"`
+	MachineState   []MachineState `json:"machineState,omitempty"`
+}
+```
 
 #### Upgrading nodes
-We will have `NodeKubeadmUpgrade` CRD and implement a controller to reconcile it. This controller will be responsible from scheduling a pod on the specified workload cluster node with the specified containers as `initContainers`. It will track their progress and bubble up any error/success to the `NodeKubeadmUpgrade` status. The status should also allow to track the progress of the different upgrade "steps".
+We will have `NodeUpgrade` CRD and implement a controller to reconcile it. This controller will be responsible from scheduling a pod on the specified workload cluster node with `initContainers` with different commands depending on whether the node is a control plane node or a worker node. It will track their progress and bubble up any error/success to the `NodeUpgrade` status.
```
vivek-koppuru, line 107:
```suggestion
We will have `NodeUpgrade` CRD and implement a controller to reconcile it. This controller will be responsible from scheduling a pod on the specified workload cluster node with `initContainers` with different commands to the upgrader pod script depending on whether the node is a control plane node or a worker node. It will track their progress and bubble up any error/success to the `NodeUpgrade` status.
```

g-gaston:
Could we add in the PR description the justification for all the changes?

abhinavmpandey08:
> I'm assuming we decided to make these changes to the design because we found reasons to do it and then we implemented that "new design". Could we add in the PR description the justification for all the changes?

I have added more details about the changes made in the design along with the justifications. Lmk if you have any more questions.

g-gaston:
> This was changed because the runtime extensions needed for this feature aren't defined in upstream CAPI yet but it will be added when in-place is officially supported upstream. Until then we would have to maintain a patch in our CAPI to add these runtime extensions which may or may not match what upstream CAPI would have.
Therefore the initial investment in adding runtime extensions doesn't seem justified considering we can achieve similar functionality by using annotations, which is a 2 way door decision. Once the runtime extension support for in-place upgrade is officially added in upstream CAPI, we can migrate to it without a lot of throwaway code.

Then I would suggest we also avoid patching the capi API. If we are concerned about implementing something that doesn't align with the final CAPI proposal (fair enough), the API is the most difficult thing to "backtrack", so we should make sure we don't change it until CAPI accepts the changes.

I would also suggest also that you somehow still present the original idea in the doc and just add more text explaining what the intermediate solution is that you are implementing and how it will be iterated towards the CAPI approved solution. That tell the reader what the final goal is and how our implementation will evolve over time.

> This was done because we can gather all the information that we need from the KCP/MD ref including the machines that need upgrade, the new Kubernetes version that the machines need to be upgraded to, and versions of other components that need to be upgraded.

This explains how the alternative solution works, not why it's better than the original proposal.

> This was done because most of the commands were generic between control plane node and worker nodes. The only exception was the kubeadm upgrade. So to account for this, the NodeUpgrade object will accept a NodeType field to signify whether the node to be upgraded should be a cp or a worker. Then the ControlPlaneUpgrade and the MachineDeploymentUpgrade controller can just pass the NodeType down without having to worry about what the exact implementation of what upgrading a control plane node or a worker node would entail, as those details are abstracted away.

Same here. Why is this better? I'm missing the reasoning behind the change.

Also I would love to see the tradeoffs. For example, you are leaking the "CP vs worker" concept to the node upgrader. You achieve a more compact API by paying the price on coupling. Now your node upgrader api and controller have multiple reasons to change and some of them are shared with the other controllers, which means some in situations you might have to change all the controllers at once.

taneyland (approved):

designs/in-place-upgrades-eksa.md
```diff
@@ -98,15 +149,15 @@ The node upgrade process we need to perform, although different depending on the
...
 
 Draining and uncordoning the node could run in either the container or the host namespace. However, we will run it in the container namespace to be able to leverage the injected credentials for the `ServiceAccount`. This way we don't depend on having a kubeconfig in the host disk. This not only allows us to easily limit the rbac permissions that the `kubectl` command will use, but it's specially useful for worker nodes, since these don't have a kubeconfig with enough permissions to perform these actions (CP nodes have an admin kubeconfig).
 
-In order to codify the logic of each step (the ones that require logic, like the kubeadm upgrade), we will build a single go binary with multiple commands (one per step). The `ControlPlaneKubeadmUpgrade` and `WorkersKubeadmUpgrade` will just reference these commands when building the init containers spec.
+In order to codify the logic of each step (the ones that require logic, like the kubeadm upgrade), we will build a single go binary with multiple commands (one per step). The `ControlPlaneUpgrade` and `MachineDeploymentUpgrade` will just reference these commands when building the init containers spec.
```
abhinavmpandey08, line 160: update: add a line about script temporarily instead of binary.
Update the last sentence to say `NodeUpgrade` instead of CPUpgrade and MDUpgrade

taneyland (commented):
Look great! Couple nits

taneyland (approved):

PR #7283, opened 2024-01-10
Add methods to fetch upgrader image from bundle

taneyland:
*Issue #, if available:*

*Description of changes:*
The eks-a controller will be responsible for creating the custom CP/MD upgrade resources for in-place upgrades. We need a way to fetch the upgrader image from the bundle so this adds methods that will be used in the controller to pull the bundle and fetch the images. The images will be stored in a config map. This upgrader config map will then be fetched by the node upgrader when the image is needed.

ConfigMap:
```
apiVersion: v1
data:
  v1.25.16-eks-1-25-29: public.ecr.aws/l0g8r8j6/aws/upgrader:v1-25-29-eks-a-v0.0.0-dev-build.8103
  v1.26.11-eks-1-26-25: public.ecr.aws/l0g8r8j6/aws/upgrader:v1-26-25-eks-a-v0.0.0-dev-build.8103
  v1.27.8-eks-1-27-19: public.ecr.aws/l0g8r8j6/aws/upgrader:v1-27-19-eks-a-v0.0.0-dev-build.8103
  v1.28.4-eks-1-28-12: public.ecr.aws/l0g8r8j6/aws/upgrader:v1-28-12-eks-a-v0.0.0-dev-build.8103
  v1.29.0-eks-1-29-1: public.ecr.aws/l0g8r8j6/aws/upgrader:v1-29-1-eks-a-v0.0.0-dev-build.8103
kind: ConfigMap

```

*Testing (if applicable):*
performed in-place upgrade with image from bundle

### DIFF
--- a/controllers/nodeupgrade_controller.go
+++ b/controllers/nodeupgrade_controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -25,10 +26,8 @@ import (
 )
 
 const (
-	// TODO(in-place): Get this image from the bundle instead of using the hardcoded one.
-	defaultUpgraderImage = "public.ecr.aws/t2p4l7v3/upgrader:eksdbase"
-	controlPlaneLabel    = "node-role.kubernetes.io/control-plane"
-	podDNEMessage        = "Upgrader pod does not exist"
+	controlPlaneLabel = "node-role.kubernetes.io/control-plane"
+	podDNEMessage     = "Upgrader pod does not exist"
 
 	// nodeUpgradeFinalizerName is the finalizer added to NodeUpgrade objects to handle deletion.
 	nodeUpgradeFinalizerName = "nodeupgrades.anywhere.eks.amazonaws.com/finalizer"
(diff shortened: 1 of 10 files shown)


controllers/nodeupgrade_controller.go
```diff
@@ -168,14 +166,20 @@ func (r *NodeUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger,
 		return ctrl.Result{}, nil
 	}
 
+	configMap := &corev1.ConfigMap{}
+	if err := remoteClient.Get(ctx, types.NamespacedName{Name: constants.UpgraderConfigMapName, Namespace: constants.EksaSystemNamespace}, configMap); err != nil {
```
vivek-koppuru, line 170: So we are planning to retrieve the config map and pass the image directly to NodeUpgrader object? What's stopping us from doing this Get operation in the nodeupgrade controller then?

Other option is so pass in the config map from here to node upgrade object, just so that if node upgrader is run on it's own for any reason, we can pass in any config map. If we decide to mount the config map to the upgrader pod directly, then that gives us a different amount of flexibility where we can deploy the pod directly with a different config map mounted. Just some thoughts that I have to consider.

pkg/constants/constants.go
```diff
@@ -46,6 +46,8 @@ const (
 	NutanixCredentialsName = "nutanix-credentials"
 	EksaLicenseName        = "eksa-license"
 	EksaPackagesName       = "eksa-packages"
+	// UpgraderConfigMapName is the name of config map that stores the upgrader images.
+	UpgraderConfigMapName = "upgrader-config-map"
```
abhinavmpandey08, line 50: `config-map` in the name is a bit redundant since it's of type configmap. How about this instead?
```suggestion
	UpgraderConfigMapName = "in-place-upgrade"
```

pkg/clustermanager/cluster_manager.go
```diff
@@ -1176,6 +1178,16 @@ func (c *ClusterManager) ApplyBundles(ctx context.Context, clusterSpec *cluster.
...
 		return fmt.Errorf("applying bundle spec: %v", err)
 	}
+
+	// We need to update this config map with the new upgrader images whenever we
+	// apply a new Bundles object to the cluster in order to support in-place upgrades.
+	cm, err := getUpgraderImagesFromBundle(clusterSpec)
+	if err != nil {
+		return fmt.Errorf("getting upgrader images from bundle: %v", err)
+	}
+	if err = c.clusterClient.Apply(ctx, cluster.KubeconfigFile, cm); err != nil {
```
abhinavmpandey08, line 1188: Does this apply replace the configmap with new contents or does it merge with the current configmap if present on the cluster?

If it's doing a replace, we might have to fetch the cm in the code and append to it ourselves (if it exists) and then apply it back.

taneyland: Yeah we did decide to append it. Updated this to fetch the existing one

controllers/nodeupgrade_controller.go
```diff
@@ -168,14 +167,23 @@ func (r *NodeUpgradeReconciler) reconcile(ctx context.Context, log logr.Logger,
...
+		return ctrl.Result{}, err
+	}
+	if configMap.Data == nil {
+		return ctrl.Result{}, errors.New("upgrader config map is empty")
+	}
+	upgraderImage := configMap.Data[nodeUpgrade.Spec.KubernetesVersion]
```
abhinavmpandey08, line 177: i meant something like 
```suggestion
	upgraderImage, ok := configMap.Data[nodeUpgrade.Spec.KubernetesVersion]
	if !ok {
	    return ctrl.Result{}, fmt.Errorf("upgrader image corresponding to EKS Distro version %s not found in the configmp", nodeUpgrade.Spec.KubernetesVersion)
	}
```

pkg/clustermanager/cluster_manager.go
```diff
@@ -1394,3 +1407,44 @@ func (c *ClusterManager) buildSpecForCluster(ctx context.Context, clus *types.Cl
...
+		if executables.IsKubectlNotFoundError(err) {
+			return newUpgraderConfigMap(upgraderImages), nil
+		}
+		return nil, err
+	}
+
+	for version, image := range upgraderImages {
+		if _, ok := upgraderConfigMap.Data[version]; !ok {
+			upgraderConfigMap.Data[version] = image
+		}
```
abhinavmpandey08, line 1432: I think for this case, when a version is already in the configmap, we should override it instead of preserving the older version
```suggestion
		upgraderConfigMap.Data[version] = image
```

controllers/nodeupgrade_controller.go
taneyland: Ah yes. I totally agree. Thanks!

abhinavmpandey08 (approved):

PR #7295, opened 2024-01-12
Set etcd url in cloud-config when external etcd is enabled

vignesh-goutham:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/2116

*Description of changes:*
These changes will set a release-url for etcdadm to download the etcd tarball if the right version is not cached already in the OS during build time. This is especially useful, when user is creating a cluster with a really old OS template but with a new EKS-A version. This change will also only be in effect for clusters created/upgraded with 0.19.0 or above, to avoid node rollouts.

### DIFF
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -86,9 +86,9 @@ const (
 	// DevBuildVersion is the version string for the dev build of EKS-A.
 	DevBuildVersion = "v0.0.0-dev"
 
-	// EtcdUrlEksAVersion is the version from which the etcd url will be set
-	// for etcdadm to pull the etcd tarball if that binary isnt cached
-	EtcdUrlEksAVersion = "v0.19.0"
+	// MinEksAVersionWithEtcdURL is the version from which the etcd url will be set
+	// for etcdadm to pull the etcd tarball if that binary isnt cached.
+	MinEksAVersionWithEtcdURL = "v0.19.0"
 )
 
 // Equal checks if two EksaVersions are equal.
--- a/pkg/api/v1alpha1/cluster_types.go
+++ b/pkg/api/v1alpha1/cluster_types.go
@@ -82,8 +82,14 @@ type ClusterSpec struct {
 // EksaVersion is the semver identifying the release of eks-a used to populate the cluster components.
 type EksaVersion string
 
-// DevBuildVersion is the version string for the dev build of EKS-A.
-const DevBuildVersion = "v0.0.0-dev"
+const (
+	// DevBuildVersion is the version string for the dev build of EKS-A.
+	DevBuildVersion = "v0.0.0-dev"
+
+	// EtcdUrlEksAVersion is the version from which the etcd url will be set
+	// for etcdadm to pull the etcd tarball if that binary isnt cached
+	EtcdUrlEksAVersion = "v0.19.0"
+)
 
 // Equal checks if two EksaVersions are equal.
 func (n *EksaVersion) Equal(o *EksaVersion) bool {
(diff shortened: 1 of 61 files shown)


pkg/providers/common/common.go
```diff
@@ -147,3 +149,24 @@ func GetCAPIBottlerocketSettingsConfig(config *v1alpha1.BottlerocketConfiguratio
 
 	return strings.Trim(string(marshaledConfig), "\n"), nil
 }
+
+func GetExternalEtcdReleaseUrl(clusterVersion string, versionBundle *cluster.VersionsBundle) (string, error) {
+	clusterVersionSemVer, err := semver.New(clusterVersion)
```
jiayiwang7, line 154: i don't think we need this as it is validated here: https://github.com/aws/eks-anywhere/blob/9a356ed3aac6f284b5d81eaa4aaf5dbd59c9811d/pkg/api/v1alpha1/cluster.go#L884 (this function is called in both cli and controller i think)

pkg/clusterapi/etcd.go
```diff
@@ -11,15 +11,20 @@ import (
...
 // SetUbuntuConfigInEtcdCluster sets up the etcd config in EtcdadmCluster.
-func SetUbuntuConfigInEtcdCluster(etcd *etcdv1.EtcdadmCluster, version string) {
+func SetUbuntuConfigInEtcdCluster(etcd *etcdv1.EtcdadmCluster, versionsBundle *cluster.VersionsBundle, clusterSpec *cluster.Spec) {
 	etcd.Spec.EtcdadmConfigSpec.Format = etcdbootstrapv1.Format("cloud-config")
 	etcd.Spec.EtcdadmConfigSpec.CloudInitConfig = &etcdbootstrapv1.CloudInitConfig{
-		Version:    version,
+		Version:    versionsBundle.KubeDistro.EtcdVersion,
 		InstallDir: "/usr/bin",
 	}
+	etcdUrl, _ := common.GetExternalEtcdReleaseUrl(string(*clusterSpec.Cluster.Spec.EksaVersion), versionsBundle)
```
jiayiwang7, line 24: shouldn't we handle the error?

pkg/providers/cloudstack/template.go
```diff
@@ -224,6 +225,10 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec) (map[string]interface{}, erro
 		values["externalEtcd"] = true
 		values["externalEtcdReplicas"] = clusterSpec.Cluster.Spec.ExternalEtcdConfiguration.Count
 		values["etcdSshUsername"] = etcdMachineSpec.Users[0].Name
+		etcdUrl, _ := common.GetExternalEtcdReleaseUrl(string(*clusterSpec.Cluster.Spec.EksaVersion), versionsBundle)
```
jiayiwang7, line 228: same for all the providers, shouldn't we handle the error?

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -77,8 +77,14 @@ type ClusterSpec struct {
...
 
-// DevBuildVersion is the version string for the dev build of EKS-A.
-const DevBuildVersion = "v0.0.0-dev"
+const (
+	// DevBuildVersion is the version string for the dev build of EKS-A.
+	DevBuildVersion = "v0.0.0-dev"
+
+	// EtcdUrlEksAVersion is the version from which the etcd url will be set
+	// for etcdadm to pull the etcd tarball if that binary isnt cached
+	EtcdUrlEksAVersion = "v0.19.0"
```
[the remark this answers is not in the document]
vignesh-goutham: This is the version from which we will set the etcd url. So this is pretty much static and wont change. We will use this to compare with the cluster version to check if we need to set etcd url or not.

jiayiwang7 (approved):

Issue #6719, opened 2023-09-22
Support maxSurge and maxUnavailable across providers


jiayiwang7:
Currently only bare metal provider supports `maxSurge` and `maxUnavailable` in rolling upgrade strategy: https://anywhere.eks.amazonaws.com/docs/clustermgmt/cluster-upgrades/baremetal-upgrades/#upgraderolloutstrategyrollingupdatemaxsurge

We should support it across providers so that user can rollout multiple machines at the same time during upgrade.

PR #7315, opened 2024-01-17
adding docker upgrade rollout strategy

raymond-zhang00:
Issue #, if available:
https://github.com/aws/eks-anywhere/issues/6719

Description of changes:
Adding UpgradeRolloutStrategy knobs for workers and control plane to docker provider

Testing (if applicable):
Updated unit tests.

Documentation added/planned (if applicable):
Planning on adding docs on knobs for other providers following the [Tinkerbell upgrade strategy format](https://anywhere.eks.amazonaws.com/docs/clustermgmt/cluster-upgrades/baremetal-upgrades/#upgraderolloutstrategy).


pkg/providers/docker/docker_test.go
```diff
@@ -327,6 +327,38 @@ func TestProviderGenerateDeploymentFileSuccessUpdateMachineTemplate(t *testing.T
 			wantCPFile: "testdata/valid_deployment_cp_expected.yaml",
 			wantMDFile: "testdata/valid_autoscaler_deployment_md_expected.yaml",
 		},
+		{
+			testName: "valid config with rollout",
```
g-gaston, line 331: I'll leave it up to you but I would prefer if we tested these things through `ControlPlaneSpec` and `WorkersSpec`

That decouples the tests from the fact that these spec builders return yaml (which they should never have)

I understand you are just following whatever was in here, which is fair. But It would be better if we could start writing new tests following a better pattern.

This applies to all providers

raymond-zhang00: That makes sense, swapping over to ControlPlaneSpec and WorkersSpec tests instead for all providers

g-gaston (commented):
/lgtm

PR #7319, opened 2024-01-17
Create management cluster using controller bootstrap task

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Create management cluster using controller bootstrap task
The original https://github.com/aws/eks-anywhere/pull/7110 is being split into smaller PRs. This PR is the first one.

*Testing (if applicable):*
- The whole workflow has been tested end to end for Docker, vSphere, and Cloudstack providers.
- Unit tests


pkg/clustermanager/cluster_manager.go
```diff
@@ -523,6 +523,22 @@ func (c *ClusterManager) getWorkloadClusterKubeconfig(ctx context.Context, clust
...
 
+// CreateEKSAReleaseBundle applies the eks-a release bundle to the cluster.
+func (c *ClusterManager) CreateEKSAReleaseBundle(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
+	if clusterSpec.Cluster.Namespace != "" {
+		if err := c.clusterClient.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, clusterSpec.Cluster.Namespace); err != nil {
+			return err
+		}
+	}
+
+	clusterSpec.Cluster.AddManagedByCLIAnnotation()
```
g-gaston, line 534: what is this for?

pkg/clustermanager/cluster_manager.go
```diff
@@ -523,6 +523,22 @@ func (c *ClusterManager) getWorkloadClusterKubeconfig(ctx context.Context, clust
 	return nil
 }
 
+// CreateEKSAReleaseBundle applies the eks-a release bundle to the cluster.
+func (c *ClusterManager) CreateEKSAReleaseBundle(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
```
g-gaston, line 527: can this be done from the eks-a installer?

pkg/clustermanager/cluster_manager.go
mitalipaygude: This condition let's the webhook create management clusters using the eksa controller.

pkg/clustermanager/cluster_manager.go
mitalipaygude: I ended up using this wrapper because while creating the cluster's namespace we use the cluster Client that is a private Retrier Client field used by the cluster manager. I won't be able to use the clusterctl client if I were to move this code.

pkg/clustermanager/cluster_manager.go
g-gaston: ah good point
then could we set it before applying the cluster spec? so in a different task, whichever uses the cluster creator

Also maybe add a comment since not even me remember what this was for 🤦🏻

pkg/clustermanager/cluster_manager.go
g-gaston: why do you need "the clusterctl client"? we should only need a k8s client wrapped by a retrier
which the installer should already have?

mitalipaygude: My bad, I meant the cluster client.

pkg/clustermanager/cluster_manager.go
mitalipaygude: In that case I'll add it before the create workload cluster task and remove it from this PR.

pkg/clustermanager/eksa_installer.go
```diff
@@ -108,6 +109,47 @@ func (i *EKSAInstaller) Upgrade(ctx context.Context, log logr.Logger, c *types.C
...
+		return fmt.Errorf("applying release spec: %v", err)
+	}
+
+	return nil
+}
+
+// CreateNamespaceIfNotPresent creates the cluster namespace if it doesn not exist.
+func (i *EKSAInstaller) CreateNamespaceIfNotPresent(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error {
+	if spec.Cluster.Namespace != "" {
+		if err := i.client.CreateNamespaceIfNotPresent(ctx, cluster.KubeconfigFile, spec.Cluster.Namespace); err != nil {
```
g-gaston, line 145: IMO we should not do this
It should be the use the one creating the namespace before calling eks-a if they want to create a cluster in it

pkg/dependencies/factory.go
```diff
@@ -1103,6 +1104,39 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster, timeoutOpt
...
+		var r *retrier.Retrier
+		if f.config.noTimeouts {
+			r = retrier.NewWithNoTimeout()
+		} else {
+			r = clustermanager.DefaultRetrier()
+		}
+
+		client := clustermanager.NewRetrierClient(
+			&clusterManagerClient{
+				f.dependencies.Clusterctl,
```
g-gaston, line 1125: why do we need clusterctl for the eksa installer? shouldn't it be enough with a k8s client?

pkg/dependencies/factory.go
```diff
@@ -1103,6 +1104,39 @@ func (f *Factory) WithClusterManager(clusterConfig *v1alpha1.Cluster, timeoutOpt
 	return f
 }
 
+// WithEKSAInstaller builds a cluster manager based on the cluster config and timeout options.
+func (f *Factory) WithEKSAInstaller(clusterConfig *v1alpha1.Cluster, timeoutOpts *ClusterManagerTimeoutOptions) *Factory {
```
g-gaston, line 1108: nit: can we update the clustermanager build method to make use of this one?

pkg/workflows/interfaces/interfaces.go
```diff
@@ -92,3 +94,10 @@ type ClusterCreator interface {
 	Run(ctx context.Context, spec *cluster.Spec, managementCluster types.Cluster) error
 	CreateSync(ctx context.Context, spec *cluster.Spec, managementCluster *types.Cluster) (*types.Cluster, error)
 }
+
+// EksaInstaller exposes the EKSA installer methods.
+type EksaInstaller interface {
+	ApplyBundles(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error
+	ApplyReleases(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error
+	CreateNamespaceIfNotPresent(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error
+}
```
g-gaston, line 103: ideally we should have
```suggestion
type EksaInstaller interface {
	Install(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error
}
```

if we can't do that for some reason, let's talk about why and try to figure out a solution

pkg/workflows/management/create_test.go
```diff
@@ -105,13 +110,272 @@ func (c *createTestSetup) expectPreflightValidationsToPass() {
...
+
+		c.clusterManager.EXPECT().InstallCAPI(
+			c.ctx, c.clusterSpec, c.bootstrapCluster, c.provider).Return(err2),
+
+		c.provider.EXPECT().PostBootstrapSetup(
+			c.ctx, c.clusterSpec.Cluster, c.bootstrapCluster).Return(err3),
+	)
+}
+
+// func (c *createTestSetup) expectInstallEksaComponentsBootstrap(err1, err2, err3, err4 error) {
```
g-gaston, line 139: is this not needed? if so, let's remove it

pkg/workflows/management/create_test.go
```diff
@@ -105,13 +110,272 @@ func (c *createTestSetup) expectPreflightValidationsToPass() {
...
+
+	c.writer.EXPECT().Write(fmt.Sprintf("%s-checkpoint.yaml", c.clusterSpec.Cluster.Name), gomock.Any())
+
+	err := c.run()
+	if err == nil {
+		t.Fatalf("Create.Run() expected to return an error %v", err)
+	}
+}
+
+// func TestCreateEKSAReleaseBundleFailure(t *testing.T) {
```
g-gaston, line 334: same

pkg/clustermanager/eksa_installer.go
```diff
@@ -108,6 +109,47 @@ func (i *EKSAInstaller) Upgrade(ctx context.Context, log logr.Logger, c *types.C
 	return changeDiff, nil
 }
 
+// ApplyBundles applies the bundles to the cluster.
+func (i *EKSAInstaller) ApplyBundles(ctx context.Context, log logr.Logger, cluster *types.Cluster, spec *cluster.Spec) error {
```
g-gaston, line 113: Can all these methods be part of `Install`? (you can make them unexported and then call them from `Install`)

PR #7322, opened 2024-01-17
Add Cilium kube-proxy replacement support

chrisdoherty4:
Cilium can take ownership of all kube-proxy responsibilities. In Cilium v1.13 there are partial and strict modes, however the API changes in v1.14 to a binary setting so this support is for v1.13 strict mode only.

### DIFF
--- a/pkg/providers/cloudstack/config/template-cp.yaml
+++ b/pkg/providers/cloudstack/config/template-cp.yaml
@@ -307,6 +307,12 @@ spec:
 {{- end }}
 {{- end }}
 {{- end }}
+{{- with .kubeadmSkipPhases }}
+      skipPhases:
+      {{- range . }}
+      - {{ . }}
+      {{- end }}
+{{- end }}
     joinConfiguration:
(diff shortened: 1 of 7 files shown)


pkg/providers/cloudstack/config/template-cp.yaml
```diff
@@ -306,6 +306,12 @@ spec:
           timeAdded: {{ .TimeAdded }}
 {{- end }}
 {{- end }}
+{{- end }}
+{{- with .kubeadmSkipPhases }}
+      skipPhases:
+      {{- range . }}
+      - {{ . }}
+      {{- end }}
```
cxbrowne1207, line 314: Curious. Why was this needed?

chrisdoherty4: Why is `skipPhases` needed or why is `end`?

cxbrowne1207: `skipPhases`

PR #7328, opened 2024-01-18
Add in place rollout strategy option

taneyland:
*Issue #, if available:*

*Description of changes:*
This adds the option for InPlace upgrade rollout strategy type in EKS-A spec. We will update the KCP/MD objects in a separate PR

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -825,19 +825,20 @@ func validatePodIAMConfig(clusterConfig *Cluster) error {
 }
 
 func validateCPUpgradeRolloutStrategy(clusterConfig *Cluster) error {
-	if clusterConfig.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy == nil {
+	cpUpgradeRolloutStrategy := clusterConfig.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy
+	if cpUpgradeRolloutStrategy == nil {
 		return nil
 	}
 
-	if clusterConfig.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy.Type != "RollingUpdate" {
-		return fmt.Errorf("ControlPlaneConfiguration: only 'RollingUpdate' supported for upgrade rollout strategy type")
+	if cpUpgradeRolloutStrategy.Type != "RollingUpdate" && cpUpgradeRolloutStrategy.Type != "InPlace" {
+		return fmt.Errorf("ControlPlaneConfiguration: only 'RollingUpdate' and 'InPlace' are supported for upgrade rollout strategy type")
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -849,8 +850,8 @@ func validateMDUpgradeRolloutStrategy(w *WorkerNodeGroupConfiguration) error {
 		return nil
 	}
 
-	if w.UpgradeRolloutStrategy.Type != "RollingUpdate" {
-		return fmt.Errorf("WorkerNodeGroupConfiguration: only 'RollingUpdate' supported for upgrade rollout strategy type")
+	if w.UpgradeRolloutStrategy.Type != "RollingUpdate" && w.UpgradeRolloutStrategy.Type != "InPlace" {
+		return fmt.Errorf("WorkerNodeGroupConfiguration: only 'RollingUpdate' and 'InPlace' are supported for upgrade rollout strategy type")
```
rahulbabu95, line 854: This looks like a provider agnostic check. So looks like we might need to restrict "InPlace"  at other providers once the rolloutStrategy support PR for other providers get merged in.? Another option is to check if the type is inplace and `if cpUpgradeRolloutStrategy.Type == "InPlace" && cluster.Spec.DatacenterRef.Kind != TinkerbellDatacenterKind` and error out for this case here.

pkg/providers/tinkerbell/validate.go
```diff
@@ -27,11 +27,22 @@ func validateOsFamily(spec *ClusterSpec) error {
...
+			}
+		}
 	}
 
 	if controlPlaneOsFamily != v1alpha1.Bottlerocket && spec.DatacenterConfig.Spec.OSImageURL == "" && spec.ControlPlaneMachineConfig().Spec.OSImageURL == "" {
 		return fmt.Errorf("please use bottlerocket as osFamily for auto-importing or provide a valid osImageURL")
 	}
+
+	if spec.ControlPlaneConfiguration().UpgradeRolloutStrategy != nil {
+		if controlPlaneOsFamily != v1alpha1.DefaultOSFamily && spec.ControlPlaneConfiguration().UpgradeRolloutStrategy.Type == "InPlace" {
```
rahulbabu95, line 42: While we check for the CP and Worker rollout strategy types separately, do we also want to check if one is specified as `InPlace` other cannot be rolling.? I don't think our controllers are designed in such a way but this will prevent unnecessary complications.

pkg/api/v1alpha1/cluster.go
taneyland: Yes, I'll include the logic to restrict this feature for the other providers in the validations PR.

pkg/providers/tinkerbell/validate.go
taneyland: Sure that makes sense. There is a separate task for fully validating this feature, so I can include this additional check there

PR #7332, opened 2024-01-19
Fix ADOT flaky test

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Fix ADOT flaky test by checking the curl output instead of looking for logs that are causing race conditions.
*Testing (if applicable):*

*Documentation added/planned (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #7334, opened 2024-01-19
Allow adding extra curl args while keeping default behavior

abhay-krishna:
We use `curl` commands to validate the curated packages installations in E2E tests. The default command used is just `curl <endpoint>`, but this was changed to `curl -I <endpoint>` (which gets only curl headers) in #7332 for fixing ADOT tests, which caused other packages to fail since they expect output present in the full `curl` output. This PR allows to pass in custom  `curl` options to the `ValidateCurlEndpoint` method so that each package can customize how the `curl` command is run.

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -1455,8 +1455,8 @@ func (e *ClusterE2ETest) VerifyAdotPackageInstalled(packageName, targetNamespace
 	}
 	podFullIPAddress := strings.Trim(podIPAddress, `'"`) + ":8888/metrics"
 	e.T.Log("Validate content at endpoint", podFullIPAddress)
-	expectedLogs = "otelcol_exporter"
-	e.ValidateEndpointContent(podFullIPAddress, targetNamespace, expectedLogs)
+	expectedLogs = "HTTP/1.1 200 OK"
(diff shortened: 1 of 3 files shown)


mitalipaygude (approved):

test/framework/cluster.go
```diff
@@ -2092,13 +2092,13 @@ func (e *ClusterE2ETest) ApplyPackageFile(packageName, targetNamespace string, P
...
 // and returns the created pod name.
-func (e *ClusterE2ETest) CurlEndpoint(endpoint, namespace string) string {
+func (e *ClusterE2ETest) CurlEndpoint(endpoint, namespace string, extraCurlArgs ...string) string {
 	ctx := context.Background()
 
 	e.T.Log("Launching pod to curl endpoint", endpoint)
 	randomname := fmt.Sprintf("%s-%s", "curl-test", utilrand.String(7))
 	curlPodName, err := e.KubectlClient.RunCurlPod(context.TODO(),
-		namespace, randomname, e.KubeconfigFilePath(), []string{"curl", endpoint})
+		namespace, randomname, e.KubeconfigFilePath(), append([]string{"curl", endpoint}, extraCurlArgs...))
```
mitalipaygude, line 2101: shouldn't this be inside the string array?

PR #7353, opened 2024-01-23
Compare new cluster spec version bundles to management components version bundles in upgrade plan

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR changes the `upgrade plan cluster` logic to compare the management components version bundles, fetched using the management components version annotation, to the new cluster specs version bundles for core component upgrades.

The changes include deciding to select the first Bundles.Spec.VersionsBundles[0] for the upgrade plan comparisons and the actual upgrade itself. This is because for decoupled component upgrades, the management components can be upgraded to the new EKS-A version separately from the Cluster. So, after we have the management components bundles for that new version, but, there are still multiple Kubernetes versions to choose from within the bundle to get the components information. However, because management component images are the same for every Kubernetes version within the same bundle manifest, it's OK to use the first bundle. If there are is differences between the management components on this first versions bundle, and the new cluster specs first versions bundle,
that indicates an upgrade is required.

*Testing (if applicable):*
- Created a management and workload cluster with v0.18.3, checked upgrade plans before upgrading management components.
- Ran the `upgrade management-components` subcommand, checked upgrade plans after.

**Before running upgrade management components**

*Management cluster*
```
./bin/eksctl-anywhere upgrade plan cluster -f mgmt/mgmt-eks-a-cluster.yaml
...
NAME                 CURRENT VERSION       NEXT VERSION
EKS-A Management     v0.18.3+cc70180       v0.0.0-dev+build.8203+c1a3710
cert-manager         v1.13.0+68bec33       v1.13.2+a34c207
cluster-api          v1.5.2+b14378d        v1.6.0+04c07bc
kubeadm              v1.5.2+5762149        v1.6.0+5bf0931
vsphere              v1.7.4+6ecf386        v1.8.5+650acfa
kubeadm              v1.5.2+b8987c8        v1.6.0+7a4ce68
etcdadm-bootstrap    v1.0.10+c9a5a8a       v1.0.10+1ceb898
etcdadm-controller   v1.0.16+0ed68e6       v1.0.17+5e33062
cilium               v1.12.15-eksa.1       v1.13.9-eksa.1
kubernetes           v1.27.8-eks-1-27-17   v1.27.8-eks-1-27-20
```

*Workload Cluster*
```
./bin/eksctl-anywhere upgrade plan cluster -f w01/w01-eks-a-cluster.yaml 
...
NAME         CURRENT VERSION       NEXT VERSION
cilium       v1.12.15-eksa.1       v1.13.9-eksa.1
[...]

PR #7376, opened 2024-01-25
  Extend docker provider api surface to use management components

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR extends the docker provider api surface to use management component to gradually introduce the changes outlined in [this PR](https://github.com/aws/eks-anywhere/pull/7353), split out into multiple. This is the first one.

*Testing (if applicable):*

### DIFF
--- a/pkg/cluster/fetch.go
+++ b/pkg/cluster/fetch.go
@@ -11,6 +11,59 @@ import (
 	v1alpha1release "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// ManagementComponents bundles the resource definitions of all EKS-A management components.
+type ManagementComponents struct {
+	EksD                   v1alpha1release.EksDRelease
+	CertManager            v1alpha1release.CertManagerBundle
+	ClusterAPI             v1alpha1release.CoreClusterAPI
+	Bootstrap              v1alpha1release.KubeadmBootstrapBundle
+	ControlPlane           v1alpha1release.KubeadmControlPlaneBundle
+	VSphere                v1alpha1release.VSphereBundle
+	CloudStack             v1alpha1release.CloudStackBundle
+	Docker                 v1alpha1release.DockerBundle
+	Eksa                   v1alpha1release.EksaBundle
+	Flux                   v1alpha1release.FluxBundle
+	ExternalEtcdBootstrap  v1alpha1release.EtcdadmBootstrapBundle
+	ExternalEtcdController v1alpha1release.EtcdadmControllerBundle
+	Tinkerbell             v1alpha1release.TinkerbellBundle
+	Snow                   v1alpha1release.SnowBundle
+	Nutanix                v1alpha1release.NutanixBundle
+}
+
+// ManagementComponentsFromBundles returns ManagementComponents built from a VersionsBundle.
+//
+// For decoupled component upgrades, the management components can be upgraded to the new EKS-A version
+// separately from the Cluster. So, here we have the management components bundles for that new version,
+// but, there are still multiple Kubernetes versions to choose from within the bundle to get the
+// components information. However, because management component images are the same for every Kubernetes
+// version within the same bundle manifest, it's OK to use the first bundle. If there are is differences between
+// the management components on this first versions bundle, and the new cluster specs first versions bundle,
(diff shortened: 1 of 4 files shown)


pkg/cluster/fetch.go
```diff
@@ -11,6 +11,30 @@ import (
 	v1alpha1release "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// ManagementComponents contain the versions bundle for a managment cluster's core components.
+type ManagementComponents struct {
+	*v1alpha1release.VersionsBundle
+}
```
g-gaston, line 17: I would just copy over whatever are management components and have fields for it
This doesn't decouple upgraders from the format of a versions bundle

I mean something like

```suggestion
type ManagementComponents struct {
    Version string
	CoreCapi releasev1.Whatever
	DockerProvider releasev1.WhateverProvider
	...
	EKSAComponents releasev1.WhateverManifest
}
```

pkg/providers/docker/docker.go
```diff
@@ -77,7 +77,8 @@ type ProviderKubectlClient interface {
 	UpdateAnnotation(ctx context.Context, resourceType, objectName string, annotations map[string]string, opts ...executables.KubectlOpt) error
 }
 
-func NewProvider(providerConfig *v1alpha1.DockerDatacenterConfig, docker ProviderClient, providerKubectlClient ProviderKubectlClient, now types.NowFunc) providers.Provider {
+// NewProvider returns a new providers.Provider.
+func NewProvider(providerConfig *v1alpha1.DockerDatacenterConfig, docker ProviderClient, providerKubectlClient ProviderKubectlClient, now types.NowFunc) providers.TemporaryProvider {
```
g-gaston, line 81: does this need to return an interface?
can't we just return the actual type?

pkg/providers/provider.go
```diff
@@ -46,6 +46,15 @@ type Provider interface {
 	PreCoreComponentsUpgrade(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 }
 
+// TemporaryProvider is a Provider but expanded to take on different interfaces for a few methods. This is a temporary change
+// to facilitate refactoring efforts in smaller changes.
+type TemporaryProvider interface {
```
g-gaston, line 51: do we need a new interface? i think if we return the types, we don't
also, we can still return the old interface even if we don't return the type

what is this interface for? no one is consuming it

cxbrowne1207: This was added to add the new methods without changing the original interface first. They would not be accessible in the tests through `provider.Provider`. Perhaps, we can export the docker `provider` struct that implements `provider.Provider` and reference that in the tests instead?

pkg/providers/docker/docker.go
cxbrowne1207: We can make it return a type

pkg/cluster/fetch.go
```diff
@@ -11,6 +11,58 @@ import (
 	v1alpha1release "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
+// ManagementComponents contains the versions bundle for a management cluster's core components.
```
g-gaston, line 14:
```suggestion
// ManagementComponents bundles the resource definitions of all EKS-A management components.
```

maybe?

pkg/providers/docker/docker.go
```diff
@@ -77,8 +79,9 @@ type ProviderKubectlClient interface {
 	UpdateAnnotation(ctx context.Context, resourceType, objectName string, annotations map[string]string, opts ...executables.KubectlOpt) error
 }
 
-func NewProvider(providerConfig *v1alpha1.DockerDatacenterConfig, docker ProviderClient, providerKubectlClient ProviderKubectlClient, now types.NowFunc) providers.Provider {
-	return &provider{
+// NewProvider returns a new providers.Provider.
```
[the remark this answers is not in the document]
cxbrowne1207: Oh, I forgot to change this one after I update it to return the concrete type. Will change

pkg/cluster/fetch.go
cxbrowne1207: Sounds better. Thanks!

PR #7377, opened 2024-01-25
Add controlplane spec field in CPUpgrade

abhinavmpandey08:
*Description of changes:*
Add `ControlPlaneSpec` field in CPUpgrade which will store the base64 encoded KCP.Spec
This field is needed to update the CAPI objects to the desired state so CAPI and KCP controllers can determine that the in-place upgrade is complete.
This PR also uses this field to update the control plane machine annotation `controlplane.cluster.x-k8s.io/kubeadm-cluster-configuration` with the updated kubeadmClusterConfig.

This new field is needed because if we try to fetch the kcp.spec from the controlPlane ref, we get a dynamic copy of the the spec. It's possible that the KCP object was modified after the `ControlPlaneUpgrade` object was created. In cases like this, we need a static copy of KCP.Spec so we can update the relevant objects with the KCP.Spec values that were there when the `ControlPlaneUpgrade` object was created.

### DIFF
--- a/config/crd/bases/anywhere.eks.amazonaws.com_controlplaneupgrades.yaml
+++ b/config/crd/bases/anywhere.eks.amazonaws.com_controlplaneupgrades.yaml
@@ -74,6 +74,14 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+              controlPlaneSpecData:
+                description: ControlPlaneSpecData contains base64 encoded KCP spec
+                  that's used to update the statuses of CAPI objects once the control
+                  plane upgrade is done. This field is needed so that we have a static
+                  copy of the control plane spec in case it gets modified after the
+                  ControlPlaneUpgrade was created, as ControlPlane is a reference
+                  to the object in real time.
+                type: string
               etcdVersion:
                 description: EtcdVersion refers to the version of ETCD to upgrade
                   to.
@@ -148,6 +156,7 @@ spec:
                 type: array
             required:
             - controlPlane
+            - controlPlaneSpecData
             - etcdVersion
             - kubernetesVersion
             - machinesRequireUpgrade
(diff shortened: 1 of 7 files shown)


config/crd/bases/anywhere.eks.amazonaws.com_controlplaneupgrades.yaml
```diff
@@ -74,6 +74,11 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
+              controlPlaneSpec:
```
vivek-koppuru, line 77: Is this the entire kcp object? Also, I am forgetting why we need controlPlane and controlPlaneSpec separately and if we can somehow explain that in both of the descriptions.

controllers/controlplaneupgrade_controller.go
```diff
@@ -223,8 +228,21 @@ func (r *ControlPlaneUpgradeReconciler) updateStatus(ctx context.Context, log lo
...
+			if err != nil {
+				return fmt.Errorf("decoding cpUpgrade.Spec.ControlPlaneSpec: %v", err)
+			}
+			if err := json.Unmarshal(decodedKcpSpec, kcpSpec); err != nil {
+				return fmt.Errorf("unmarshaling cpUpgrade.Spec.ControlPlaneSpec: %v", err)
+			}
+			clusterConfig, err := json.Marshal(kcpSpec.KubeadmConfigSpec.ClusterConfiguration)
+			if err != nil {
+				return fmt.Errorf("marshaling KCP cluster configuration: %v", err)
+			}
```
vivek-koppuru, line 244: Nit: Can we move this all to a separate function that just returns either the err or the string?

vivek-koppuru (approved):
/approve
/trickortreat

vivek-koppuru (approved):
/trick-or-treat

config/crd/bases/anywhere.eks.amazonaws.com_controlplaneupgrades.yaml
abhinavmpandey08: done

PR #7379, opened 2024-01-26
Update Template in MachineSet once InPlace upgrade is completed

rahulbabu95:
*Issue #, if available:*
Unlike KCP, machineDeployments use `machinesets` constructs which then are responsible for scaling the machines. During the regular flow for rolling upgrades, CAPI generates the updated machinesets based on the new spec, which then scales up new machines in the rolling fashion. Since for InPlace we do not want any new machines to roll over, we need to update the `machineset.Spec.Template.Spec`  to the current machine spec once the InPlace upgrade has been completed. This will ensure that the `machinesets` then syncs with the machines and updates the `Machine.Spec` field to reflect the current value on the `machinesets`. 

*Description of changes:*

*Testing (if applicable):*
Tested manually using a custom `capi-controller` image built to handle InPlace and verified that the machineset was updated with the new template.

### DIFF
--- a/controllers/machinedeploymentupgrade_controller.go
+++ b/controllers/machinedeploymentupgrade_controller.go
@@ -18,14 +18,18 @@ package controllers
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -39,6 +43,7 @@ import (
 const (
 	// mdUpgradeFinalizerName is the finalizer added to MachineDeploymentUpgrade objects to handle deletion.
 	mdUpgradeFinalizerName = "machinedeploymentupgrades.anywhere.eks.amazonaws.com/finalizer"
+	mdLabelIdentifier      = "cluster.x-k8s.io/deployment-name"
 )
 
 // MachineDeploymentUpgradeReconciler reconciles a MachineDeploymentUpgrade object.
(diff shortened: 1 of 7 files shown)


controllers/machinedeploymentupgrade_controller.go
```diff
@@ -74,13 +80,28 @@ func (r *MachineDeploymentUpgradeReconciler) Reconcile(ctx context.Context, req
...
+	if err := r.client.Get(ctx, types.NamespacedName{Name: mdUpgrade.Spec.MachineDeployment.Name, Namespace: mdUpgrade.Spec.MachineDeployment.Namespace}, md); err != nil {
+		return ctrl.Result{}, fmt.Errorf("getting MachineDeployment %s: %v", mdUpgrade.Spec.MachineDeployment.Name, err)
+	}
+
+	ms, err := r.getCurrentMachineSet(ctx, md)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	msPatchHelper, err := patch.NewHelper(ms, r.client)
```
vivek-koppuru, line 98: Shouldn't this happen only if the mdupgrade is marked as done?

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -206,12 +230,50 @@ func (r *MachineDeploymentUpgradeReconciler) updateStatus(ctx context.Context, l
...
+func (r *MachineDeploymentUpgradeReconciler) getCurrentMachineSet(ctx context.Context, md *clusterv1.MachineDeployment) (*clusterv1.MachineSet, error) {
+	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: map[string]string{mdLabelIdentifier: md.Name}})
+	if err != nil {
+		return nil, err
+	}
+	machineSetsList := &clusterv1.MachineSetList{}
+	if err := r.client.List(ctx, machineSetsList, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("getting machine sets for %v: %v", md.Name, err)
+	}
+	revision, ok := md.Annotations[clusterv1.RevisionAnnotation]
```
vivek-koppuru, line 257: Does field selector work for annotations or is that specifically for fields under spec? That might help avoid the for loop below right?

pkg/api/v1alpha1/machinedeploymentupgrade_types.go
```diff
@@ -18,6 +18,9 @@ type MachineDeploymentUpgradeSpec struct {
 
 	// KubernetesVersion refers to the Kubernetes version to upgrade the control planes to.
 	KubernetesVersion string `json:"kubernetesVersion"`
+
+	// MachineSpec is a base64 encoded json string value of the machineDeplopyment.Spec.Template.Spec field that's specification of the desired behavior of the machine.
+	MachineSpec string `json:"machineSpec"`
```
vivek-koppuru, line 23: Let's call this `MachineSpecData` to be inline with the changes we decided for cpupgrade

controllers/machinedeploymentupgrade_controller.go
```diff
@@ -206,12 +222,64 @@ func (r *MachineDeploymentUpgradeReconciler) updateStatus(ctx context.Context, l
 	mdUpgrade.Status.Upgraded = int64(nodesUpgradeCompleted)
 	mdUpgrade.Status.RequireUpgrade = int64(nodesUpgradeRequired)
 	mdUpgrade.Status.Ready = nodesUpgradeRequired == 0
+	if mdUpgrade.Status.Ready {
```
vivek-koppuru, line 225: Why are we not retrieving the MachineSet here? Do we need it earlier?

vivek-koppuru (approved):

PR #7386, opened 2024-01-26
Update KubeadmConfig for cp machines when upgrade is complete

abhinavmpandey08:
*Description of changes:*
KCP controller expects `KubeadmConfig` corresponding to each control plane machine to match `KCP.Spec.KubeadmConfig` for a machine to be considered "UpToDate".

This PR updates the `KubeadmConfig` for each control plane machine as the in-place upgrade is done for that machine.

### DIFF
--- a/controllers/controlplaneupgrade_controller.go
+++ b/controllers/controlplaneupgrade_controller.go
@@ -27,8 +27,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/patch"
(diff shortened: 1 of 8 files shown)


controllers/controlplaneupgrade_controller.go
```diff
@@ -239,42 +242,84 @@ func (r *ControlPlaneUpgradeReconciler) updateMachine(ctx context.Context, log l
...
+	patchHelper, err := patch.NewHelper(kc, r.client)
 	if err != nil {
-		return "", fmt.Errorf("decoding cpUpgrade.Spec.ControlPlaneSpec: %v", err)
+		return fmt.Errorf("initializing patch helper for KubeadmConfig %s: %v", kc.Name, err)
 	}
-	if err := json.Unmarshal(decodedKcpSpec, kcpSpec); err != nil {
-		return "", fmt.Errorf("unmarshaling cpUpgrade.Spec.ControlPlaneSpec: %v", err)
+
+	kcSpec := kcpSpec.KubeadmConfigSpec.DeepCopy()
+	if kc.Spec.InitConfiguration == nil {
```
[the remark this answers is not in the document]
abhinavmpandey08: no, it's because `kcp.Spec. KubeadmConfigSpec` config has both the init and join configs but the `KubeadmConfig` only has either init or join depending whether it's the first control plane node or not.
So, we should only update the `KubeadmConfig`'s init or join config section based on whichever one was set

controllers/controlplaneupgrade_controller.go
```diff
@@ -234,47 +237,89 @@ func (r *ControlPlaneUpgradeReconciler) updateStatus(ctx context.Context, log lo
...
+		kcSpec.InitConfiguration = nil
+	}
+
+	if kc.Spec.JoinConfiguration == nil {
+		kcSpec.JoinConfiguration = nil
+	}
+
+	kc.Spec = *kcSpec
```
vivek-koppuru, line 304: And we know that kcSpec won't be nil right?

controllers/controlplaneupgrade_controller.go
```diff
@@ -234,47 +237,89 @@ func (r *ControlPlaneUpgradeReconciler) updateStatus(ctx context.Context, log lo
...
+	decodedKcpSpec, err := base64.StdEncoding.DecodeString(kcpSpecData)
 	if err != nil {
-		return "", fmt.Errorf("marshaling KCP cluster configuration: %v", err)
+		return nil, fmt.Errorf("decoding cpUpgrade.Spec.ControlPlaneSpec: %v", err)
+	}
+	if err := json.Unmarshal(decodedKcpSpec, kcpSpec); err != nil {
+		return nil, fmt.Errorf("unmarshaling cpUpgrade.Spec.ControlPlaneSpec: %v", err)
 	}
-	return string(clusterConfig), nil
+	return kcpSpec, err
```
vivek-koppuru, line 322: Why are we returning err if we know it has to be nil?
```suggestion
	return kcpSpec, nil
```

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #7368, opened 2024-01-25
Use super-admin Kubeconfig for kube-vip from Kubernetes v1.29 onwards

abhay-krishna:
Before Kubernetes v1.29, the default Kubeconfig file `admin.conf` generated by `kubeadm` was bound to `system:masters` Group. In Kubernetes v1.29, a [change](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.29.md#no-really-you-must-read-this-before-you-upgrade) was made to `kubeadm` to provision a separate Kubeconfig file called `super-admin.conf` which is a break-glass Group that can bypass RBAC, and in addition, scoping down the `admin.conf` User's access to the permissions provided by the `cluster-admin` `ClusterRole`. It is to be noted that in a multi-cluster setup, the `super-admin.conf` is only created on the primary control plane node, which runs `kubeadm init`. This causes a bootstrap issue with EKS-A's `kube-vip` deployment because the static pod manifest is hardcoded to use `admin.conf` but `kube-vip` requires elevated permissions to access the API server during cluster initialization, which are available only when using the `super-admin.conf`. 

This [issue](https://github.com/kube-vip/kube-vip/issues/684) was first reported by CAPV maintainers in Kubernetes Slack `kube-vip` channel ([Initial issue report](https://kubernetes.slack.com/archives/C01RXPHDATB/p1703061589086269), [Workarounds discussion](https://kubernetes.slack.com/archives/C01RXPHDATB/p1704643142829679)) when they were trying to add Kubernetes v1.29 support. Since then, the CAPV maintainers have found a [workaround](https://github.com/kube-vip/kube-vip/issues/684#issuecomment-1864855405) that involves using the `super-admin.conf` only for `kubeadm init` and then migrating to using the `admin.conf` after the cluster has been initialized. This PR basically ports the workaround to our cluster templates for EKS-A's `kube-vip` deployment to work with Kubernetes v1.29.

This is similar to the [logic](https://github.com/aws/eks-anywhere-build-tooling/pull/2863) we added to the Bottlerocket kubeadm-bootstrap host container to fix the same issue for clusters using Bottlerocket nodes.

```console
$ kubectl --kubeconfig vsphere-1-29-test-ubuntu/vsphere-1-29-test-ubuntu-eks-a-cluster.kubeconfig get nodes -o wide       
[...]

PR #7396, opened 2024-01-27
Set kubeconfig volume path for kube-vip static pod based on OS

abhay-krishna:
For Bottlerocket, the admin kubeconfig volume path is `/var/lib/kubeadm/admin.conf`, but in our templates, this field is set as `/etc/kubernetes/admin.conf` for all OS families. During cluster creation, the CAPI kubeadm bootstrap controller patches the kube-vip static pod manifest to change it to the correct hostPath value for Bottlerocket. Instead of doing that, we could set it to the correct path ourselves since we know the OS at the time of cluster creation itself. This migration will help when we make the kubeconfig path dynamic based on the Kubernetes version in #7368.


pkg/providers/tinkerbell/config/template-cp.yaml
```diff
@@ -296,7 +296,11 @@ spec:
             hostNetwork: true
             volumes:
             - hostPath:
+{{- if (eq .format "bottlerocket") }}
```
g-gaston, line 299: I thought we weren't doing br in baremetal anymore?

abhay-krishna: This is just for the existing Kubernetes versions. I am going to remove [this hardcoded value](https://github.com/aws/eks-anywhere-build-tooling/blob/07f3f36b77ca72faddf07b86d1fc0a58d7b93406/projects/kubernetes-sigs/cluster-api/patches/0001-Adding-capi-support-for-Bottlerocket.patch#L890) so that we can control the path from EKS-A template. So we need this change across all providers to maintain the current behavior.

abhinavmpandey08 (approved):
/lgtm
/woof
/approve

PR #7417, opened 2024-01-29
Unify delete workload

tatlat:
*Issue #, if available:*

*Description of changes:*
Added a delete workflow for workload clusters that uses the controller.

*Testing (if applicable):*


pkg/workflows/workload/setup_delete.go
mitalipaygude, line 1: We can move the contents of this file [here](https://github.com/aws/eks-anywhere/blob/main/pkg/workflows/workload/validate.go) since we are doing create and upgrade validations in there too.

mitalipaygude:
Looks pretty good to me!

g-gaston (commented):
/lgtm
/hold

PR #7428, opened 2024-01-30
Add kubernetes entry to upgrade plan

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This changes here are split out from #7353 with some changes, passing the `ManagementComponents` object down to where it's needed to make comparisons.

Adds "kubernete" change entry as a cluster component to the upgrade plan:
- Change messaging  and interface of the EKSD upgrader to less confusing as to what it does: updates eks-d crds, not upgrading eks-d from one version to another.
- Remove types.ChangeDiff as a return from the EKSDUpgrader interface because it's unneeded.

*Testing (if applicable):*
- Created management cluster with v0.18.4 and check upgrade plan output
- Upgrade management-components and check the output upgrade plan

Before upgrade management-components
```
NAME                 CURRENT VERSION       NEXT VERSION
...
cilium               v1.12.15-eksa.1       v1.13.9-eksa.1
kubernetes           v1.27.8-eks-1-27-19   v1.27.8-eks-1-27-21

```

After upgrade management-components (kubernetes and cilium are still there because cluster has not been upgraded yet)
```
NAME                 CURRENT VERSION       NEXT VERSION
...
cilium               v1.12.15-eksa.1       v1.13.9-eksa.1
kubernetes           v1.27.8-eks-1-27-19   v1.27.8-eks-1-27-21
```

After upgrade cluster
```
NAME      CURRENT VERSION   NEXT VERSION
```

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgradeplancluster.go
+++ b/cmd/eksctl-anywhere/cmd/upgradeplancluster.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/aws/eks-anywhere/pkg/dependencies"
+	"github.com/aws/eks-anywhere/pkg/eksd"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networking/cilium"
 	"github.com/aws/eks-anywhere/pkg/types"
@@ -96,6 +97,7 @@ func (uc *upgradeClusterOptions) upgradePlanCluster(ctx context.Context) error {
 	}
 
 	componentChangeDiffs.Append(cilium.ChangeDiff(currentSpec, newClusterSpec))
+	componentChangeDiffs.Append(eksd.ChangeDiff(currentSpec, newClusterSpec))
 
 	serializedDiff, err := serialize(componentChangeDiffs, output)
 	if err != nil {
--- a/pkg/eksd/upgrader.go
+++ b/pkg/eksd/upgrader.go
@@ -3,10 +3,12 @@ package eksd
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/types"
+	releavev1alpha1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
 type Upgrader struct {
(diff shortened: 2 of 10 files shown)


cmd/eksctl-anywhere/cmd/upgradeplancluster.go
```diff
@@ -111,7 +112,9 @@ func (uc *upgradeClusterOptions) upgradePlanCluster(ctx context.Context) error {
 	componentChangeDiffs.Append(fluxupgrader.FluxChangeDiff(currentSpec, newClusterSpec))
 	componentChangeDiffs.Append(capiupgrader.CapiChangeDiff(currentSpec, newClusterSpec, deps.Provider))
 	componentChangeDiffs.Append(cilium.ChangeDiff(currentSpec, newClusterSpec))
-
+	componentChangeDiffs.Append(
+		eksdupgrader.ChangeDiff(cluster.NewManagementComponents(currentSpec.RootVersionsBundle().VersionsBundle), cluster.NewManagementComponents(newClusterSpec.RootVersionsBundle().VersionsBundle)),
```
g-gaston, line 116: don't hate me
but does it make sense to use management components for an eksd diff?
eksd is not a management component. Is a cluster component

pkg/eksd/upgrader.go
```diff
@@ -23,34 +25,45 @@ func NewUpgrader(client EksdInstallerClient, reader Reader, opts ...UpgraderOpt)
 	}
 }
 
-func (u *Upgrader) Upgrade(ctx context.Context, cluster *types.Cluster, currentSpec, newSpec *cluster.Spec) (*types.ChangeDiff, error) {
+// Upgrade checks whether upgrading EKS-D components is necessary and, if so, installs the new EKS-D CRDs.
+func (u *Upgrader) Upgrade(ctx context.Context, cluster *types.Cluster, currentManagementComponents, newManagementComponents *cluster.ManagementComponents, newSpec *cluster.Spec) (*types.ChangeDiff, error) {
```
g-gaston, line 29: Maybe we need to think a bit more here about this
This updates the eks-d CRDs in case new fields or versions have been added.
But that's not "upgrading eks-d", it's not actually changing any cluster component.

I'm fine with the method. But the way we build the change diff I think it's misleading. I think it indicates we are upgrading from one eks-d version to another and we are not

TBH I don't have a solution, it's the first time I think about this confusion. Maybe we can sync offline and see if you have some ideas

cmd/eksctl-anywhere/cmd/upgradeplancluster.go
cxbrowne1207: Fair point. Yeah, this confusion arises in the messages we show in the diff as well

Maybe, we should just keep this one as is referencing the cluster spec

pkg/eksd/upgrader.go
cxbrowne1207: Yeah, we'll continue this offline

g-gaston (commented):
lgtm

cmd/eksctl-anywhere/cmd/upgradeplancluster.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/aws/eks-anywhere/pkg/dependencies"
+	eksdupgrader "github.com/aws/eks-anywhere/pkg/eksd"
```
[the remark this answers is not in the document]
cxbrowne1207: This was to stay consistent with how the other upgraders are imported [here](https://github.com/aws/eks-anywhere/blob/489d6720f63fc46a4073fabba2c398cebf762f76/cmd/eksctl-anywhere/cmd/upgradeplanmanagementcomponents.go#L11). It's not necessary exactly

pkg/eksd/upgrader.go
cxbrowne1207: Discussed offline, we'll change the messaging in the logs, but also change this method to only return an error. This will make it more clear that it is not actually upgrading eks-d from one version to another, but updating the crds when necessary.

g-gaston (commented):
/lgtm

PR #7429, opened 2024-01-30
Add additional validations for in place upgrades

taneyland:
*Issue #, if available:*

*Description of changes:*
This PR adds more validations for in place upgrades to ensure that:
- InPlace can only be specified for Tinkerbell provider with the Ubuntu OS family
- Cp and workers must have the same upgrade rollout strategy specified.

*Testing (if applicable):*
Unit tests

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -482,7 +482,7 @@ func validateWorkerNodeGroups(clusterConfig *Cluster) error {
 			return fmt.Errorf("validating autoscaling configuration: %v", err)
 		}
 
-		if err := validateMDUpgradeRolloutStrategy(&workerNodeGroupConfig); err != nil {
+		if err := validateMDUpgradeRolloutStrategy(&workerNodeGroupConfig, clusterConfig.Spec.DatacenterRef.Kind); err != nil {
 			return fmt.Errorf("validating upgrade rollout strategy configuration: %v", err)
 		}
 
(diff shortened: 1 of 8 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -32,6 +32,8 @@ const (
 	ClusterKind              = "Cluster"
 	RegistryMirrorCAKey      = "EKSA_REGISTRY_MIRROR_CA"
 	podSubnetNodeMaskMaxDiff = 16
+	rollingUpdate            = "RollingUpdate"
+	inPlaceUpgrade           = "InPlace"
```
vivek-koppuru, line 36: These should be in cluster_types.go somehow as enum right? That way we get the validation from that type checking?

pkg/api/v1alpha1/cluster.go
```diff
@@ -832,36 +834,45 @@ func validateCPUpgradeRolloutStrategy(clusterConfig *Cluster) error {
...
+	case rollingUpdate:
+		if cpUpgradeRolloutStrategy.RollingUpdate.MaxSurge < 0 {
+			return fmt.Errorf("ControlPlaneConfiguration: maxSurge for control plane cannot be a negative value")
+		}
+		if cpUpgradeRolloutStrategy.RollingUpdate.MaxSurge > 1 {
+			return fmt.Errorf("ControlPlaneConfiguration: maxSurge for control plane must be 0 or 1")
+		}
+	case inPlaceUpgrade:
+		if clusterConfig.Spec.DatacenterRef.Kind != TinkerbellDatacenterKind {
+			return fmt.Errorf("ControlPlaneConfiguration: 'InPlace' upgrade rollout strategy type is only supported on the Tinkerbell provider")
```
vivek-koppuru, line 847: Do we say Tinkerbell provider officially or BareMetal provider?

pkg/providers/tinkerbell/validate.go
```diff
@@ -27,22 +27,40 @@ func validateOsFamily(spec *ClusterSpec) error {
...
+func validateUpgradeRolloutStrategy(spec *ClusterSpec) error {
+	var cpUpgradeRolloutStrategyType string
+
 	if spec.ControlPlaneConfiguration().UpgradeRolloutStrategy != nil {
-		if controlPlaneOsFamily != v1alpha1.DefaultOSFamily && spec.ControlPlaneConfiguration().UpgradeRolloutStrategy.Type == "InPlace" {
+		cpUpgradeRolloutStrategyType = spec.ControlPlaneConfiguration().UpgradeRolloutStrategy.Type
+		controlPlaneRef := spec.ControlPlaneConfiguration().MachineGroupRef
+		controlPlaneOsFamily := spec.MachineConfigs[controlPlaneRef.Name].OSFamily()
+
+		if controlPlaneOsFamily != v1alpha1.DefaultOSFamily && cpUpgradeRolloutStrategyType == inPlaceUpgradeRolloutStrategy {
```
vivek-koppuru, line 47: What if this DefaultOSFamily changes? Should we just explicitly check for Ubuntu?

pkg/api/v1alpha1/cluster.go
taneyland: It looks like we do say "Bare Metal". I'll update!

pkg/providers/tinkerbell/validate.go
```diff
@@ -27,22 +27,40 @@ func validateOsFamily(spec *ClusterSpec) error {
...
+
+		if controlPlaneOsFamily != v1alpha1.Ubuntu && cpUpgradeRolloutStrategyType == v1alpha1.InPlaceStrategyType {
 			return fmt.Errorf("InPlace upgrades are only supported on the Ubuntu OS family")
 		}
 	}
+
+	for _, group := range spec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		groupRef := group.MachineGroupRef
+
+		if group.UpgradeRolloutStrategy != nil {
```
vivek-koppuru, line 55: Can the control plane strategy be nil while this is not? Or vice versa? Should we have another check for that or you think that is not needed?

taneyland: I think that is technically allowed because it defaults to `RollingUpdate` in CAPI if it isn't specified. But let's say they want to customize the maxSurge only on a worker node group, we don't have anything stopping that today.

Maybe here, I can set it to "RollingUpdate" at the beginning for this method since that's the default:
`cpUpgradeRolloutStrategyType := RollingUpdateStrategyType`
 
then if it is specified, we will set it to that instead. But if it isn't specified, this will still allow the worker nodes to pass the validation & potentially add more customization if desired. Thoughts?

vivek-koppuru (approved):

PR #7444, opened 2024-01-31
Compute docs checksum and upload it to artifacts bucket

sp1999:
*Issue #, if available:*
[#1931](https://github.com/aws/eks-anywhere-internal/issues/1931)

*Description of changes:*
Added a make target to compute the eks anywhere docs checksum from `public/index.html` and upload it to artifacts bucket with public access so that the docs canary can verify with the latest version

*Testing (if applicable):*
Tested locally in personal isengard account

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -81,4 +81,4 @@ compute-checksum: ## Compute docs checksum
 	sha256sum public/index.html > DOCS_CHECKSUM
 
 upload-checksum: compute-checksum ## Upload the checksum to artifacts bucket
-	aws s3 cp DOCS_CHECKSUM s3://artifactsstack-1532887287-artifactsbucket2aac5544-exgxrmntr4id --acl public-read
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -76,3 +76,9 @@ container-server: container-serve
 
 clean: ## Delete generated content. (requires sudo if using container-serve)
 	rm -rf public resources node_modules public.zip
+
+compute-checksum: ## Compute docs checksum
+	sha256sum public/index.html > DOCS_CHECKSUM
+
+upload-checksum: compute-checksum ## Upload the checksum to artifacts bucket
+	aws s3 cp DOCS_CHECKSUM s3://artifactsstack-1532887287-artifactsbucket2aac5544-exgxrmntr4id --acl public-read
```
abhay-krishna, line 84: We shouldn't hardcode this here. We should set it as an env var from CDK.

abhay-krishna (changes requested):

docs/Makefile
sp1999: Updated

PR #7446, opened 2024-01-31
Add feature flag for in place upgrades on vSphere

taneyland:
*Issue #, if available:*

*Description of changes:*
Adding a feature flag because we are using this provider to test in place upgrade changes, but it will only be supported on bare metal when the feature is initially released.

This PR also updates the vsphere capi specs to specify inplace if it is specified in the clusterspec

*Testing (if applicable):*
Unit test
Manual test for the validations step

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -22,6 +22,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/aws/eks-anywhere/pkg/constants"
+	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/networkutils"
 	"github.com/aws/eks-anywhere/pkg/semver"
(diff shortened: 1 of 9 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -841,6 +842,13 @@ func validateCPUpgradeRolloutStrategy(clusterConfig *Cluster) error {
 			return fmt.Errorf("ControlPlaneConfiguration: maxSurge for control plane must be 0 or 1")
 		}
 	case InPlaceStrategyType:
+		if clusterConfig.Spec.DatacenterRef.Kind == VSphereDatacenterKind {
+			if features.IsActive(features.VSphereInPlaceUpgradeEnabled()) {
+				return nil
+			}
+			return fmt.Errorf("in place upgrades are not supported on vSphere. For dev testing, please set the env variable %s", features.VSphereInPlaceEnvVar)
```
abhinavmpandey08, line 849: I would say let's not even mention the feature flag. We don't want people messing around with it and screwing something up
```suggestion
			return errors.New("in place upgrades are not supported on vSphere")
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -866,6 +874,12 @@ func validateMDUpgradeRolloutStrategy(w *WorkerNodeGroupConfiguration, datacente
 			return fmt.Errorf("WorkerNodeGroupConfiguration: maxSurge and maxUnavailable not specified or are 0. maxSurge and maxUnavailable cannot both be 0")
 		}
 	case InPlaceStrategyType:
+		if datacenterRefKind == VSphereDatacenterKind {
+			if features.IsActive(features.VSphereInPlaceUpgradeEnabled()) {
+				return nil
+			}
+			return fmt.Errorf("in place upgrades are not supported on vSphere. For dev testing, please set the env variable %s", features.VSphereInPlaceEnvVar)
```
abhinavmpandey08, line 881: same as above

pkg/api/v1alpha1/cluster.go
taneyland: good callout! updated

abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #7456, opened 2024-02-01
Fix e2e tests to search the kubeconfig consistently

g-gaston:
## Description of changes
The reading code was making assumptions about what the writing code did. This worked when the workload clusters were setup in a very specific way, but if custom folders were used, the assumptions were broken and the test can't find the kubeconfig file.

---

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -1046,7 +1046,7 @@ func (e *ClusterE2ETest) managementCluster() *types.Cluster {
 
 // KubeconfigFilePath retrieves the Kubeconfig path used for the workload cluster.
 func (e *ClusterE2ETest) KubeconfigFilePath() string {
-	return filepath.Join(e.ClusterName, fmt.Sprintf("%s-eks-a-cluster.kubeconfig", e.ClusterName))
+	return filepath.Join(e.ClusterConfigFolder, fmt.Sprintf("%s-eks-a-cluster.kubeconfig", e.ClusterName))
 }
 
 // BuildWorkloadClusterClient creates a client for the workload cluster created by e.
--- a/test/framework/multicluster.go
+++ b/test/framework/multicluster.go
@@ -32,7 +32,8 @@ func NewMulticlusterE2ETest(t *testing.T, managementCluster *ClusterE2ETest, wor
 	for _, c := range workloadClusters {
 		c.clusterFillers = append(c.clusterFillers, api.WithManagementCluster(managementCluster.ClusterName))
 		c.ClusterName = m.NewWorkloadClusterName()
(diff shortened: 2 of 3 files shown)


test/framework/cluster.go
```diff
@@ -1046,7 +1046,7 @@ func (e *ClusterE2ETest) managementCluster() *types.Cluster {
 
 // KubeconfigFilePath retrieves the Kubeconfig path used for the workload cluster.
 func (e *ClusterE2ETest) KubeconfigFilePath() string {
-	return filepath.Join(e.ClusterName, fmt.Sprintf("%s-eks-a-cluster.kubeconfig", e.ClusterName))
+	return filepath.Join(e.ClusterConfigFolder, fmt.Sprintf("%s-eks-a-cluster.kubeconfig", e.ClusterName))
```
g-gaston, line 1049: In most cases the folder is the same as the cluster name, but it doesn't have to be
This removes the assumption and writes all files to the same folder consistently

test/framework/multicluster.go
```diff
@@ -32,7 +32,8 @@ func NewMulticlusterE2ETest(t *testing.T, managementCluster *ClusterE2ETest, wor
 	for _, c := range workloadClusters {
 		c.clusterFillers = append(c.clusterFillers, api.WithManagementCluster(managementCluster.ClusterName))
 		c.ClusterName = m.NewWorkloadClusterName()
-		c.ClusterConfigLocation = filepath.Join(managementCluster.ClusterConfigFolder, c.ClusterName+"-eks-a.yaml")
+		c.ClusterConfigFolder = c.ClusterName
+		c.ClusterConfigLocation = filepath.Join(c.ClusterConfigFolder, c.ClusterName+"-eks-a.yaml")
```
g-gaston, line 36: This was not necessary since with the other fixes you can use a custom config folder
However, for consistency with other multicluster tests that set the workload cluster name through the test constructor, I change the config folder here as well to match the same pattern.

test/framework/workload.go
```diff
@@ -123,7 +123,7 @@ func (w *WorkloadCluster) writeKubeconfigToDisk(ctx context.Context, secretName
 	if err := w.Provider.UpdateKubeConfig(&kubeconfig, w.ClusterName); err != nil {
 		return fmt.Errorf("failed to update kubeconfig for cluster: %s", err)
 	}
-	writer, err := filewriter.NewWriter(w.ClusterConfigFolder)
+	writer, err := filewriter.NewWriter(filepath.Dir(filePath))
```
g-gaston, line 126: This was the assumption that broke everything
the output of `KubeconfigFilePath()` was being passed here as `filePath`
That output could have an arbitrary value depending on the cluster name. However, this code assumes the parent folder was `ClusterConfigFolder`.

But that was only true if `ClusterConfigFolder` == `ClusterName`. Which is not always true.

drewvanstone (approved):
lgtm

PR #7475, opened 2024-02-02
Block autoscaler configuration with Inplace upgrade strategy

rahulbabu95:
*Issue #, if available:*
InPlace upgrades do not roll-over the nodes during an upgrade. In most cases customer might not even have a spare hardware for the node to roll-over in case of baremetal. Cluster autoscaler internally uses Machinesets constructs to scale the nodes based on demand. During or after an inplace upgrade is completed, machinesets might not reflect the true source of truth of the capi machine as we only patch the `machineset.Spec.Template.Spec` field on the machineset after inplace upgrade is done.  Moreover if autoscaler decides to scale the node during when an inplace upgrade is in progress, it can lead to undesirable side-effects. This change blocks Autoscaler configuration along with InPlace upgrades.  

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/providers/tinkerbell/validate.go
+++ b/pkg/providers/tinkerbell/validate.go
@@ -1,6 +1,7 @@
 package tinkerbell
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/providers/tinkerbell/validate.go
```diff
@@ -64,6 +64,21 @@ func validateUpgradeRolloutStrategy(spec *ClusterSpec) error {
...
+func validateAutoScalerDisabledForInPlace(spec *ClusterSpec) error {
+	cpUpgradeRolloutStrategyType := spec.ControlPlaneConfiguration().UpgradeRolloutStrategy
+	// We do not support different strategy types for Inplace between CP and worker nodes so it is okay to check only CP
+	if cpUpgradeRolloutStrategyType == nil || cpUpgradeRolloutStrategyType.Type != v1alpha1.InPlaceStrategyType {
+		return nil
+	}
+
+	for _, wng := range spec.Cluster.Spec.WorkerNodeGroupConfigurations {
+		if wng.AutoScalingConfiguration != nil {
+			return fmt.Errorf("austoscaler configuration not supported with InPlace upgrades")
```
abhinavmpandey08, line 76:
```suggestion
			return errors.New("autoscaler configuration not supported with InPlace upgrades")
```

pkg/providers/tinkerbell/validate.go
```diff
@@ -64,6 +64,21 @@ func validateUpgradeRolloutStrategy(spec *ClusterSpec) error {
 	return nil
 }
 
+func validateAutoScalerDisabledForInPlace(spec *ClusterSpec) error {
+	cpUpgradeRolloutStrategyType := spec.ControlPlaneConfiguration().UpgradeRolloutStrategy
+	// We do not support different strategy types for Inplace between CP and worker nodes so it is okay to check only CP
```
abhinavmpandey08, line 69: Do we already have a validation somewhere that ensures upgrade strategy is the same for CP and all wng?

rahulbabu95: Yes, `validateUpgradeRolloutStrategy` takes care of that validation!

abhinavmpandey08 (approved):
/woof

PR #7497, opened 2024-02-06
Use new dev versions in local and E2E test builds

g-gaston:
## Description of changes
Builds on top of #7463 

This changes the build process for local dev builds and for E2E test builds to use the new versioning schema.

Copy paste of the logic followed from the original PR:

### E2E tests
When a CLI is built for dev E2E tests, it's given the latest available EKS-A dev version. This pins that particular binary to one EKS-A version and hence one Bundles manifest. This is important in order to guarantee that the same Bundles is used for the whole execution of the test pipelines and avoids a race condition where a new EKS-A dev release and Bundles are published during an E2E test run.

### Locally building the CLI
When writing and testing code for the CLI/Controller, most of the time we don't care about particular releases and we just want to use the latest available Bundles that contains the latest available set of components. This verifies that our changes are compatible with the current state of EKS-A dependencies.

To avoid having to rebuild the CLI every time we want to refresh the pulled Bundles or even having to care about fetching the latest version, this PR introduces a special build metadata identifier: `+latest`. This instructs the CLI to not look for an exact match with an EKS-A version, but select the newest one that matches our pre-release. For example: if the release manifest has two releases [`v0.19.0-dev+build.1234`, `v0.19.0-dev+build.1233`], then if the CLI has version `v0.19.0-dev+latest`, then the release `v0.19.0-dev+build.1234` will be selected.

This is the default behavior when building a CLI locally: the Makefile will calculate the appropriate major.minor.patch based on the current HEAD and its closest branch ancestor (either `main` or a `release-*` branch). If you wish to pin your local CLI to a particular version, pass the `DEV_GIT_VERSION` to the make target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,18 +30,21 @@ UNIT_TEST_PACKAGES ?= $$($(GO) list ./... | grep -vE "$(UNIT_TEST_PACKAGE_EXCLUS
 
 ## ensure local execution uses the 'main' branch bundle
 BRANCH_NAME?=main
+# DEV_GIT_VERSION should be something like v0.19.0-dev+latest, depending on the base branch
+# https://github.com/aws/eks-anywhere/blob/main/docs/developer/manifests.md#locally-building-the-cli
 ifneq ($(PULL_BASE_REF),) # PULL_BASE_REF originates from prow
 	BRANCH_NAME=$(PULL_BASE_REF)
+	DEV_GIT_VERSION ?= $(shell source ./scripts/eksa_version.sh && eksa-version::get_next_eksa_version_for_ancestor $(PULL_BASE_REF))-dev+latest
+else
+	DEV_GIT_VERSION ?= $(shell source ./scripts/eksa_version.sh && eksa-version::get_next_eksa_version)-dev+latest
 endif
 ifeq (,$(findstring $(BRANCH_NAME),main))
 ## use the branch-specific bundle manifest if the branch is not 'main'
-DEV_GIT_VERSION:=v0.0.0-dev-${BRANCH_NAME}
 BUNDLE_MANIFEST_URL?=https://dev-release-assets.eks-anywhere.model-rocket.aws.dev/${BRANCH_NAME}/bundle-release.yaml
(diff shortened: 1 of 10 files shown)


g-gaston:
/unhold
@abhay-krishna I tested this locally with the new releases, it's ready to go

Makefile
```diff
@@ -32,16 +32,17 @@ UNIT_TEST_PACKAGES ?= $$($(GO) list ./... | grep -vE "$(UNIT_TEST_PACKAGE_EXCLUS
 BRANCH_NAME?=main
 ifneq ($(PULL_BASE_REF),) # PULL_BASE_REF originates from prow
 	BRANCH_NAME=$(PULL_BASE_REF)
+	DEV_GIT_VERSION ?= $(shell source ./scripts/eksa_version.sh && eksa-version::get_next_eksa_version_for_ancestor $(PULL_BASE_REF))-dev+latest
```
cxbrowne1207, line 35: It may be helpful if you left a comment of an example value for this variable.

pkg/manifests/releases/read.go
```diff
@@ -77,20 +77,38 @@ func ReleaseForVersion(releases *releasev1.Release, version string) (*releasev1.
...
+	// We will look for an exact match at the pre-release level and then compare the build metadata.
+	// This allows a locally built CLI to get the latest dev build without needing to know the exact build number.
+	wantLatestPreRelease := semVer.Buildmetadata == "latest"
+
+	var latestPreRelease *releasev1.EksARelease
+	var latestPreReleaseVersion *semver.Version
 	for _, r := range releases.Spec.Releases {
-		if r.Version == version {
-			return &r, nil
+		release := r
```
cxbrowne1207, line 88: Why do we need to assign `r` to a new variable here?

Makefile
g-gaston: added

cxbrowne1207: Thanks!

abhay-krishna (approved):

PR #7504, opened 2024-02-07
Skip extra hardware validation for InPlace Upgrades

rahulbabu95:
*Issue #, if available:*
As part of our pre-flight assertions currently we check for extra hardware was available during an upgrade for baremetal. With InPlace upgrades we don't need to perform this check when the Upgrade Strategy is configured for InPlace.

*Description of changes:*
This change skips the extra hardware availability validation for upgrades when the rollout strategy is configured to InPlace. 

*Testing (if applicable):*
Tested this manually.

### DIFF
--- a/pkg/providers/tinkerbell/upgrade.go
+++ b/pkg/providers/tinkerbell/upgrade.go
@@ -113,8 +113,12 @@ func (p *Provider) SetupAndValidateUpgradeCluster(ctx context.Context, cluster *
 		return err
 	}
 
-	if err := p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec); err != nil {
-		return err
+	upgradeStrategy := currentClusterSpec.Cluster.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy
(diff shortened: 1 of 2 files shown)


pkg/providers/tinkerbell/upgrade.go
```diff
@@ -113,8 +113,12 @@ func (p *Provider) SetupAndValidateUpgradeCluster(ctx context.Context, cluster *
 		return err
 	}
 
-	if err := p.validateAvailableHardwareForUpgrade(ctx, currentClusterSpec, clusterSpec); err != nil {
-		return err
+	upgradeStrategy := currentClusterSpec.Cluster.Spec.ControlPlaneConfiguration.UpgradeRolloutStrategy
```
taneyland, line 116: I guess this assumes that both CP/worker in place strategy will be the same, which aligns with how we currently have it configured. I know we discussed potentially changing that limitation, so adding a note to keep this validation in mind if we do change it.

taneyland (approved):

abhinavmpandey08 (approved):

PR #7526, opened 2024-02-08
Read release manifest from CLI manifest URL

g-gaston:
*Description of changes:*
This guarantees that the CLI and the tests are always reading the dev release manifest from the same place.


jonathanmeier5 (commented):
/lgtm

PR #7535, opened 2024-02-09
Fix InPlace e2e validation  to check for actual values in kcp/md.spec.Replicas  

rahulbabu95:
*Issue #, if available:*
In our e2e validations for InPlace we check for kcp/md.Spec.Replicas to be equal to the updated and ready replicas count. There was a bug in the validation logic wherein we checked for the addresses of these fields instead of dereferencing and checking for the exact value. 

*Description of changes:*
This change dereferences the kcp/md.spec.replicas to check for the actual value instead of the addresses as it is a pointer.

*Testing (if applicable):*

### DIFF
--- a/test/framework/cluster/validations/cluster.go
+++ b/test/framework/cluster/validations/cluster.go
@@ -87,7 +87,7 @@ func validateKCP(ctx context.Context, vc clusterf.StateValidationConfig) error {
 	}
 	if conditions.IsFalse(kcp, v1beta1.ReadyCondition) {
 		return errors.New("kcp ready condition is not true")
-	} else if kcp.Status.UpdatedReplicas != kcp.Status.ReadyReplicas || kcp.Spec.Replicas != &kcp.Status.UpdatedReplicas {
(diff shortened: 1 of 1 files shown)


test/framework/cluster/validations/cluster.go
```diff
@@ -87,7 +87,7 @@ func validateKCP(ctx context.Context, vc clusterf.StateValidationConfig) error {
 	}
 	if conditions.IsFalse(kcp, v1beta1.ReadyCondition) {
 		return errors.New("kcp ready condition is not true")
-	} else if kcp.Status.UpdatedReplicas != kcp.Status.ReadyReplicas || kcp.Spec.Replicas != &kcp.Status.UpdatedReplicas {
+	} else if kcp.Status.UpdatedReplicas != kcp.Status.ReadyReplicas || *kcp.Spec.Replicas != kcp.Status.UpdatedReplicas {
```
vivek-koppuru, line 90: Can we add a nil check if we are dereferencing then?

vivek-koppuru (approved):

PR #7536, opened 2024-02-09
Move kube-vip manifest modification step to pre-kubeadm commands for Nutanix

abhay-krishna:
Because this command was being placed in `postKubeadmCommands` section of the `preKubeadmCommands`, the control plane nodes were unable to get properly initialized since the `kube-vip` pod wouldn't have the necessary permissions to acquire the lease it needs to for managing the VIP.


chrisdoherty4 (approved):

abhay-krishna:
Irrelevant but stuck in retesting churn

PR #7539, opened 2024-02-09
Pause machine health checks during inplace upgrades

taneyland:
*Issue #, if available:*

*Description of changes:*
Nodes will be down temporarily during inplace upgrades. We don't want CAPI to rollout new nodes because of this, so we will pause machine health checks during inplace upgrades by adding the `cluster.x-k8s.io/paused` annotation

*Testing (if applicable):*
unit tests

### DIFF
--- a/controllers/kubeadmcontrolplane_controller.go
+++ b/controllers/kubeadmcontrolplane_controller.go
@@ -31,6 +31,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/collections"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


controllers/kubeadmcontrolplane_controller.go
```diff
@@ -115,6 +117,13 @@ func (r *KubeadmControlPlaneReconciler) reconcile(ctx context.Context, log logr.
 		return ctrl.Result{}, err
 	}
 
+	log.Info("Pausing control plane machine health check", "MachineHealthCheck", cpMachineHealthCheckName(kcp.Name))
+	mhc := &clusterv1.MachineHealthCheck{}
+	result, err := r.pauseMachineHealthCheck(ctx, mhc, kcp.Name)
+	if err != nil {
+		return result, err
+	}
```
abhinavmpandey08, line 125: Instead of pausing and unpausing every reconcile loop, can we pause whenever we are creating the CPU (right above line 156) and also unpause after deleting the CPU after line 133 (you already have it after line 173 which is good)?
Same thing for MDs

controllers/kubeadmcontrolplane_controller.go
```diff
@@ -201,6 +213,17 @@ func (r *KubeadmControlPlaneReconciler) validateStackedEtcd(kcp *controlplanev1.
 	return nil
 }
 
+func (r *KubeadmControlPlaneReconciler) pauseMachineHealthCheck(ctx context.Context, mhc *clusterv1.MachineHealthCheck, kcpName string) (ctrl.Result, error) {
+	if err := r.client.Get(ctx, GetNamespacedNameType(cpMachineHealthCheckName(kcpName), constants.EksaSystemNamespace), mhc); err != nil {
+		if apierrors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		return ctrl.Result{}, err
```
abhinavmpandey08, line 221: nit: 
```suggestion
		return ctrl.Result{}, fmt.Errorf("getting MachineHealthCheck %s: %v", cpMachineHealthCheckName(kcpName), err)
```

controllers/kubeadmcontrolplane_controller.go
```diff
@@ -201,6 +213,17 @@ func (r *KubeadmControlPlaneReconciler) validateStackedEtcd(kcp *controlplanev1.
...
+	}
+	annotations.AddAnnotations(mhc, map[string]string{machineHealthCheckPausedAnnotation: "true"})
```
abhinavmpandey08, line 223: This is just adding the annotation to the local mhc object right? Don't you have to write the mhc back to cluster to ensure CAPI recognizes it as paused as well?

controllers/kubeadmcontrolplane_controller.go
```diff
@@ -161,6 +182,12 @@ func (r *KubeadmControlPlaneReconciler) reconcile(ctx context.Context, log logr.
 		return ctrl.Result{}, fmt.Errorf("deleting ControlPlaneUpgrade object: %v", err)
 	}
 
+	log.Info("Resuming control plane machine health check", "MachineHealthCheck", cpMachineHealthCheckName(kcp.Name))
+	delete(mhc.Annotations, clusterv1.PausedAnnotation)
+	if err := mhcPatchHelper.Patch(ctx, mhc); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating annotations for machine health check: %v", err)
+	}
```
abhinavmpandey08, line 189: can you add this logic after line 127 as well? Maybe move this to a function and just call it at both places

controllers/machinedeployment_controller.go
```diff
@@ -158,6 +178,12 @@ func (r *MachineDeploymentReconciler) reconcile(ctx context.Context, log logr.Lo
 		return ctrl.Result{}, fmt.Errorf("deleting MachineDeploymentUpgrade object: %v", err)
 	}
 
+	log.Info("Resuming machine deployment machine health check", "MachineHealthCheck", mdMachineHealthCheckName(md.Name))
+	delete(mhc.Annotations, clusterv1.PausedAnnotation)
+	if err := mhcPatchHelper.Patch(ctx, mhc); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating annotations for machine health check: %v", err)
+	}
```
abhinavmpandey08, line 185: Same as for KCP, need this logic after line 125

abhinavmpandey08 (approved):
/approve
/lgtm

PR #7541, opened 2024-02-09
Add validation for stacked etcd with in place on vsphere

taneyland:
*Issue #, if available:*

*Description of changes:*
We must use local etcd with in place upgrades. Bare metal currently only supports local etcd, but for dev testing with vsphere, I'm adding this validation. Without this, the error would happen later in the flow & it would be hidden in the controller logs. 

*Testing (if applicable):*


abhinavmpandey08 (approved):
/lgtm
/woof

PR #7553, opened 2024-02-12
Allow to set tests to run in quick e2e buildspec

g-gaston:
*Description of changes:*
This allows to manually run a pipeline with this buildspec and running only one or a particular subset of tests.

It introduces the value "QUICK" that allows to run all configured tests for the quick pipeline without knowing the tests. This allows a manual pipeline to run all quick tests without passing an empty variable, which is not allowed.

### DIFF
--- a/cmd/integration_test/build/buildspecs/quick-test-eks-a-cli.yml
+++ b/cmd/integration_test/build/buildspecs/quick-test-eks-a-cli.yml
@@ -130,7 +130,10 @@ phases:
           BUNDLES_OVERRIDE=true
         fi
       - SKIPPED_TESTS=$(yq e ".skipped_tests | @csv" ${CODEBUILD_SRC_DIR}/test/e2e/SKIPPED_TESTS.yaml)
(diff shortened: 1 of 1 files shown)


cmd/integration_test/build/buildspecs/quick-test-eks-a-cli.yml
```diff
@@ -130,7 +130,11 @@ phases:
           BUNDLES_OVERRIDE=true
         fi
       - SKIPPED_TESTS=$(yq e ".skipped_tests | @csv" ${CODEBUILD_SRC_DIR}/test/e2e/SKIPPED_TESTS.yaml)
-      - QUICK_TESTS="($(yq e '.quick_tests | join("|")' ./test/e2e/QUICK_TESTS.yaml))"
+      - TESTS="${TESTS:-$(yq e '.quick_tests | join("|")' ./test/e2e/QUICK_TESTS.yaml)}"
```
d8660091, line 133: Great idea modify test case through environment variable! Looks like this line can be removed?

g-gaston: good catch!
changed

PR #7568, opened 2024-02-13
Generate snow provider environment variable map using management components

cxbrowne1207:
*Issue #, if available:*

*Context:*

The newly introduced upgrade `management-components` commands is currently building a new cluster spec and running validations against it. We do not want to do this because the users cluster spec updates are not applied while upgrading management components. Specifically, one problem  that arises is that  if your cluster is currently running a deprecated version of Kubernetes in EKS Anywhere (right now Kubernetes 1.24), the CLI will fail upgrading the management components  when [building the new cluster spec](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go#L37). 

```
```

This is because Kubernetes 1.24 is not supported now and does not exist in the new bundles .

 As a solution to this, we want to remove the need to build the new full `cluster.Spec` in the upgrade management-components workflow. For this to be possible, we need to remove the need for the `cluster.Spec` when accessing  management component information during the upgrade.

*Description of changes:*
This PR changes the snow provider to generate the environment variable map information using management components. It still uses the *cluster.Spec (presumably the current cluster spec) for any addition information, for snow the  credentials secret

*Testing (if applicable):*

### DIFF
--- a/pkg/executables/clusterctl.go
+++ b/pkg/executables/clusterctl.go
@@ -234,7 +234,7 @@ func (c *Clusterctl) InitInfrastructure(ctx context.Context, managementComponent
 		params = append(params, "--kubeconfig", cluster.KubeconfigFile)
 	}
 
-	envMap, err := provider.EnvMap(clusterSpec)
+	envMap, err := provider.EnvMap(managementComponents, clusterSpec)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


g-gaston (commented):
/lgtm

PR #7569, opened 2024-02-13
Update PreCoreComponentsUpgrade to use management components

cxbrowne1207:
*Issue #, if available:*
*Context:*

The newly introduced upgrade `management-components` commands is currently building a new cluster spec and running validations against it. We do not want to do this because the users cluster spec updates are not applied while upgrading management components. Specifically, one problem  that arises is that  if your cluster is currently running a deprecated version of Kubernetes in EKS Anywhere (right now Kubernetes 1.24), the CLI will fail upgrading the management components  when [building the new cluster spec](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go#L37). 

```
```

This is because Kubernetes 1.24 is not supported now and does not exist in the new bundles .

 As a solution to this, we want to remove the need to build the new full `cluster.Spec` in the upgrade management-components workflow. For this to be possible, we need to remove the need for the `cluster.Spec` when accessing  management component information during the upgrade.

*Description of changes:*
This PR updates the preCoreComponentsUpgrade for the providers to use a provided `cluster.Management` to perform the step. It still uses the *cluster.Spec (presumably the current cluster spec) for any addition information. For example, the the `Tinkerbell` provider needs it to run the[ stack installer upgrade](https://github.com/aws/eks-anywhere/blob/91339efb6dc85bf296c2289dbe2eb4849cd25ef7/pkg/providers/tinkerbell/upgrade.go#L464) check `workerNodeConfigurations` and the `TinkerbellDatacenter` to decide whether or not to enable the load balancer.

*Testing (if applicable):*
- Unit tests
- Upgraded management components from v0.18.6

### DIFF
--- a/pkg/providers/docker/docker.go
+++ b/pkg/providers/docker/docker.go
@@ -727,6 +727,7 @@ func getHAProxyImageRepo(haProxyImage releasev1alpha1.Image) string {
 func (p *Provider) PreCoreComponentsUpgrade(
 	ctx context.Context,
 	cluster *types.Cluster,
+	managementComponents *cluster.ManagementComponents,
 	clusterSpec *cluster.Spec,
 ) error {
 	return nil
--- a/pkg/providers/cloudstack/cloudstack.go
+++ b/pkg/providers/cloudstack/cloudstack.go
@@ -980,6 +980,7 @@ func machineDeploymentName(clusterName, nodeGroupName string) string {
 func (p *cloudstackProvider) PreCoreComponentsUpgrade(
 	ctx context.Context,
 	cluster *types.Cluster,
+	managementComponents *cluster.ManagementComponents,
 	clusterSpec *cluster.Spec,
 ) error {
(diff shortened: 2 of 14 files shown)


pkg/providers/docker/docker.go
```diff
@@ -727,6 +727,7 @@ func getHAProxyImageRepo(haProxyImage releasev1alpha1.Image) string {
 func (p *Provider) PreCoreComponentsUpgrade(
 	ctx context.Context,
 	cluster *types.Cluster,
+	managementComponents *cluster.ManagementComponents,
```
g-gaston, line 730: nit:
```suggestion
	_ *cluster.ManagementComponents,
```

g-gaston (commented):
/lgtm

PR #7574, opened 2024-02-14
Fix tinkerbell IP in hegel urls controller

mitalipaygude:
*Issue #, if available:*

*Description of changes:*
Fix tinkerbell IP in hegel urls controller 

*Testing (if applicable):*
- manual test
- unit tests

### DIFF
--- a/pkg/api/v1alpha1/cluster.go
+++ b/pkg/api/v1alpha1/cluster.go
@@ -311,6 +311,29 @@ func (c *Cluster) ClearManagedByCLIAnnotation() {
 	}
 }
 
+// AddTinkerbellIPAnnotation adds the managed-by-cli annotation to the cluster.
+func (c *Cluster) AddTinkerbellIPAnnotation(tinkerbellIP string) {
+	if c.Annotations == nil {
+		c.Annotations = map[string]string{}
+	}
+	c.Annotations[tinkerbellIPAnnotation] = tinkerbellIP
+}
+
+// ClearTinkerbellIPAnnotation removes the managed-by-cli annotation from the cluster.
+func (c *Cluster) ClearTinkerbellIPAnnotation() {
+	if c.Annotations != nil {
+		delete(c.Annotations, tinkerbellIPAnnotation)
+	}
+}
+
+// HasTinkerbellIPAnnotation returns the tinkerbell IP value if the annotation exists.
+func (c *Cluster) HasTinkerbellIPAnnotation() (string, bool) {
+	if tinkerbellIP, ok := c.Annotations[tinkerbellIPAnnotation]; ok {
+		return tinkerbellIP, true
+	}
+	return "", false
+}
+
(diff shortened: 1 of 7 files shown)


pkg/api/v1alpha1/cluster.go
```diff
@@ -311,6 +311,29 @@ func (c *Cluster) ClearManagedByCLIAnnotation() {
...
+
+// ClearTinkerbellIPAnnotation removes the managed-by-cli annotation from the cluster.
+func (c *Cluster) ClearTinkerbellIPAnnotation() {
+	if c.Annotations != nil {
+		delete(c.Annotations, tinkerbellIPAnnotation)
+	}
+}
+
+// HasTinkerbellIPAnnotation returns the tinkerbell IP value if the annotation exists.
+func (c *Cluster) HasTinkerbellIPAnnotation() (string, bool) {
```
jacobweinstock, line 330: You might consider not using the bool in the return. The alternative is for consumers to check if the returned string is empty or not. The bool adds extra checking that doesn't feel necessary. If you do go with this idea, you might then rethink the name of the method. Maybe just remove the `Has`, so its just `TinkerbellIPAnnotation`.

pkg/workflows/management/create_install_eksa.go
```diff
@@ -38,6 +39,11 @@ func (s *installEksaComponentsOnBootstrapTask) Checkpoint() *task.CompletedTask
 type installEksaComponentsOnWorkloadTask struct{}
 
 func (s *installEksaComponentsOnWorkloadTask) Run(ctx context.Context, commandContext *task.CommandContext) task.Task {
+	if commandContext.ClusterSpec.Cluster.Spec.DatacenterRef.Kind == v1alpha1.TinkerbellDatacenterKind {
```
jacobweinstock, line 42: a code comment here around why this is the only place we remove the annotation would be very helpful to future us.

jacobweinstock (commented):
Thanks, @mitalipaygude . Couple of comments, thanks.

pkg/providers/tinkerbell/create.go
```diff
@@ -28,6 +28,8 @@ func (p *Provider) BootstrapClusterOpts(_ *cluster.Spec) ([]bootstrapper.Bootstr
 func (p *Provider) PreCAPIInstallOnBootstrap(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error {
 	logger.V(4).Info("Installing Tinkerbell stack on bootstrap cluster")
 
+	logger.V(4).Info("Adding annotation for tinkerbell ip on bootstrap cluster")
+	clusterSpec.Cluster.AddTinkerbellIPAnnotation(p.tinkerbellIP)
```
jacobweinstock, line 32: a code comment here are why we need to add this annotation will be very helpful for future us.

PR #7578, opened 2024-02-14
Remove unnecessary validations for upgrade management components

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*
- Manual testing
- Running E2E tests
```
--- PASS: TestVSphereKubernetes128UpgradeManagementComponents (747.19s)
PASS
```
```
--- PASS: TestDockerKubernetesUpgradeManagementComponents (615.07s)
PASS
```
```
--- PASS: TestCloudStackKubernetes128UpgradeManagementComponents (854.62s)
PASS
```


pkg/workflows/management/upgrade_management_components.go
```diff
@@ -122,8 +122,8 @@ func (s *setupAndValidateMC) Run(ctx context.Context, commandContext *task.Comma
 	runner.Register(
 		func() *validations.ValidationResult {
 			return &validations.ValidationResult{
-				Name: fmt.Sprintf("%s provider validation", commandContext.Provider.Name()),
-				Err:  commandContext.Provider.SetupAndValidateUpgradeCluster(ctx, commandContext.ManagementCluster, commandContext.ClusterSpec, commandContext.CurrentClusterSpec),
+				Name: fmt.Sprintf("%s provider setup", commandContext.Provider.Name()),
+				Err:  commandContext.Provider.SetupUpgradeManagementComponents(ctx, commandContext.ClusterSpec),
```
d8660091, line 126: It should be "SetupAndValidateUpgradeManagementComponents"

pkg/providers/provider.go
```diff
@@ -14,6 +14,7 @@ type Provider interface {
 	SetupAndValidateCreateCluster(ctx context.Context, clusterSpec *cluster.Spec) error
 	SetupAndValidateDeleteCluster(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec) error
 	SetupAndValidateUpgradeCluster(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec, currentSpec *cluster.Spec) error
+	SetupUpgradeManagementComponents(ctx context.Context, clusterSpec *cluster.Spec) error
```
g-gaston, line 17: Why is this "setup"?
If the goal is to validate before `EnvMap` is called by clusterctl, why not just call this `ValidateManagementComponentsConfig`

cxbrowne1207: We also sets environment variables, so I consider this to be a "setup" as well
E.g. `VSPHERE_USERNAME` is set if `EKSA_VSPHERE_USERNAME` exists.

g-gaston: Why do we do that? Those variables are already passed to clusterctl
Setting env vars like that is not very great
If they are needed, it represents a hidden dependency

cxbrowne1207: I'm not sure why we do that exactly. It seems as though we have a set of environment variables that we expose to the user for input, but another set that we use when passing them to clusterctl.

We could perhaps change this behavior to have `providers.EnvMap` read the input variables directly instead if it's a blocker.

cxbrowne1207: Address through slack. We'll create a ticket to address this and come back to it.

A link to the issue has been added under one of the affected providers (vsphere).

g-gaston (commented):
/lgtm

PR #7595, opened 2024-02-15
Update kube-vip for InPlace upgrades.

rahulbabu95:
*Issue #, if available:*
Kube-vip is deployed as static pod which means during any eksd/k8s version upgrade InPlace, we need to also update the kube-vip manifest with the new image for InPlace upgrades. This change parses the udpated Kube-vip manifest from `KubeadmControlPlaneSpec.KubeadmConfigSpec` and uses a config map mounted to the upgrader pod to facilitate kube-vip update for InPlace.  The upgrader [script](https://github.com/aws/eks-anywhere-build-tooling/pull/2926) then uses the volume mount to update the kube-vip manifest onto the host. 
 
*Description of changes:*

*Testing (if applicable):*
Tested the changes locally. 

### DIFF
--- a/controllers/controlplaneupgrade_controller.go
+++ b/controllers/controlplaneupgrade_controller.go
@@ -49,6 +49,7 @@ const (
 	controlPlaneUpgradeFinalizerName      = "controlplaneupgrades.anywhere.eks.amazonaws.com/finalizer"
 	kubeadmClusterConfigurationAnnotation = "controlplane.cluster.x-k8s.io/kubeadm-cluster-configuration"
 	cloneFromNameAnnotationInfraMachine   = "cluster.x-k8s.io/cloned-from-name"
+	kubeVipStaticPodPath                  = "/etc/kubernetes/manifests/kube-vip.yaml"
 )
 
 // ControlPlaneUpgradeReconciler reconciles a ControlPlaneUpgradeReconciler object.
@@ -141,9 +142,16 @@ func (r *ControlPlaneUpgradeReconciler) reconcile(ctx context.Context, log logr.
 	// return early if controlplane upgrade is already complete
 	if cpUpgrade.Status.Ready {
 		log.Info("All Control Plane nodes are upgraded")
+		// check if kube-vip config map exists and clean it up
+		if err := cleanupKubeVipCM(ctx, log, r.client); err != nil {
+			return ctrl.Result{}, err
+		}
 		return ctrl.Result{}, nil
 	}
 
+	if err := createKubeVipCMIfNotExist(ctx, r.client, cpUpgrade); err != nil {
+		return ctrl.Result{}, err
+	}
 	log.Info("Upgrading all Control Plane nodes")
 
 	for idx, machineRef := range cpUpgrade.Spec.MachinesRequireUpgrade {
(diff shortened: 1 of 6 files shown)


controllers/controlplaneupgrade_controller.go
```diff
@@ -363,3 +367,52 @@ func getCapiMachine(ctx context.Context, client client.Client, nodeUpgrade *anyw
...
+	}
+	return &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.KubeVipConfigMapName,
+			Namespace: constants.EksaSystemNamespace,
+			OwnerReferences: []metav1.OwnerReference{{
```
vivek-koppuru, line 409: This is good but this doesn't ensure it gets cleaned up right away right? Meaning, if a new CPUpgrade needs to get created with the same name due to some subsequent upgrade, we will never get the new CM for this if it never gets deleted?

rahulbabu95: That's a good point. I will clean it up once all the nodes are upgraded for a given cpu. This way we ensure we always clean it after a successful upgrade.

abhinavmpandey08: I think when you have `BlockOwnerDeletion` set on the child object, it blocks the parent object deletion. @rahulbabu95 you can test it out just to be sure

controllers/controlplaneupgrade_controller.go
```diff
@@ -363,3 +371,71 @@ func getCapiMachine(ctx context.Context, client client.Client, nodeUpgrade *anyw
 	}
 	return machine, nil
 }
+
+func cleanupKubeVipCM(ctx context.Context, log logr.Logger, client client.Client) error {
+	cm := &corev1.ConfigMap{}
+	if err := client.Get(ctx, GetNamespacedNameType(constants.KubeVipConfigMapName, constants.EksaSystemNamespace), cm); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("config map %s  not found, skipping deletion", "ConfigMap", constants.KubeVipConfigMapName, "Namespace", constants.EksaSystemNamespace)
```
abhinavmpandey08, line 379: nit
```suggestion
			log.Info("config map %s not found, skipping deletion", "ConfigMap", constants.KubeVipConfigMapName, "Namespace", constants.EksaSystemNamespace)
```

controllers/controlplaneupgrade_controller.go
```diff
@@ -363,3 +371,71 @@ func getCapiMachine(ctx context.Context, client client.Client, nodeUpgrade *anyw
...
+	if err != nil {
+		return nil, err
+	}
+	var kubeVipConfig string
+	for _, file := range kcpSpec.KubeadmConfigSpec.Files {
+		if file.Path == kubeVipStaticPodPath {
+			kubeVipConfig = file.Content
+			break
+		}
+	}
```
abhinavmpandey08, line 421: can you add a check here to see `kubeVipConfig` is not empty after the for loop? incase the kube-vip is not found in files, just error out

pkg/constants/constants.go
```diff
@@ -48,6 +48,9 @@ const (
 	EksaPackagesName       = "eksa-packages"
 	// UpgraderConfigMapName is the name of config map that stores the upgrader images.
 	UpgraderConfigMapName = "in-place-upgrade"
+	// KubeVipConfigMapName is the name of config map that stores the kube-vip config.
+	KubeVipConfigMapName = "kube-vip-in-place-upgrade"
+	KubeVipManifestName  = "kube-vip.yaml"
```
abhinavmpandey08, line 53: nit: add comment for `KubeVipManifestName`

abhinavmpandey08 (approved):

vivek-koppuru (approved):

PR #7598, opened 2024-02-15
Remove Kubernetes v1.29 feature flag and make it the default cluster version

abhay-krishna:
This PR removes the feature flag for Kubernetes v1.29 support and make it the default cluster version.

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -35,7 +35,6 @@ import (
 	"github.com/aws/eks-anywhere/pkg/cluster"
 	"github.com/aws/eks-anywhere/pkg/constants"
 	"github.com/aws/eks-anywhere/pkg/executables"
-	"github.com/aws/eks-anywhere/pkg/features"
 	"github.com/aws/eks-anywhere/pkg/filewriter"
 	"github.com/aws/eks-anywhere/pkg/git"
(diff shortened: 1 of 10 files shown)


test/framework/cluster.go
```diff
@@ -2177,18 +2174,6 @@ func dumpFile(description, path string, t T) {
 	t.Logf("%s:\n%s\n", description, string(b))
 }
 
-func (e *ClusterE2ETest) setFeatureFlagForUnreleasedKubernetesVersion(version v1alpha1.KubernetesVersion) {
-	// Update this variable to equal the feature flagged k8s version when applicable.
-	// For example, if k8s 1.26 is under a feature flag, we would set this to v1alpha1.Kube126
-	unreleasedK8sVersion := v1alpha1.Kube129
```
taneyland, line 2183: I think we usually just set this to an empty version instead of deleting the entire method

abhay-krishna: I did whatever [this PR](https://github.com/aws/eks-anywhere/pull/6830/files#diff-4438221d080eeaeb7e58673f2b24e69bf482eb366de82660ae94342614f9ef7dL2172) was doing for 1.28, but I actually like this better. Fewer additions and deletions.

taneyland (approved):

PR #7601, opened 2024-02-15
Validate immutable Tinkerbell Datacenter fields in web-hook when managing with the Full Lifecycle Controller API

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
After `upgrade management-components` operation, EKSA allows the user to upgrade the cluster one of two ways: through the CLI and through the API. Today, the `upgrade cluster` command will upgrade the Tinkerbell stack, but upgrading the cluster through the API does not. So, any fields changed in the spec after upgrading management components would not be respected if upgrade through the API, as there is no logic in the controller that upgrades the Tinkerbell stack. Overall, there is a discrepancy between the behavior of the CLI `upgrade cluster` command and `kubectl apply -f <cluster.yaml>` when there should not be.

This PR adds some immutability validation in the tinkerbelldatacenterconfig webhook on fields that should be reconciled with trigger a Tinkerbell stack upgrade to address this in short term. ([reference to a stackInstaller.Upgrade in the code](https://github.com/aws/eks-anywhere/blob/899930ea28be6d709e96f2f113913f87861578f2/pkg/providers/tinkerbell/upgrade.go#L462))

- The `TinkerbellDatacenterConfig` fields are now immutable when managed through the API
  - `SkipLoadBalancerDeployment`
  - `HookImagesURLPath`

*Testing (if applicable):*
- Unit tests
- Manual testing, creating cluster and seeing the managed by cli annotation added and removed during upgrade.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgradecluster.go
+++ b/cmd/eksctl-anywhere/cmd/upgradecluster.go
@@ -216,6 +216,7 @@ func (uc *upgradeClusterOptions) upgradeCluster(cmd *cobra.Command, args []strin
 
 	} else {
 		upgradeWorkloadCluster := workload.NewUpgrade(
+			deps.UnAuthKubeClient,
 			deps.Provider,
 			deps.ClusterManager,
 			deps.GitOpsFlux,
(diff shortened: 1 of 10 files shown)


abhay-krishna (approved):
/approve

PR #7605, opened 2024-02-15
Refactoring installing of new components in workflows to decouple from full cluster.Spec

cxbrowne1207:
*Issue #, if available:*

*Context:*
The newly introduced sub command upgrade management-components currently has an issue where:

If you have an EKS Anywhere 1.24. You cannot update the management components if the cluster is using Kubernetes version 1.24, because the CLI runs into the following error: while [building the new cluster spec](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go#L37) 

```
```
It occurs while building the new cluster spec, when trying to [pick the correct versions bundle](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/pkg/cluster/fetch.go#L117) for the cluster’s kubernetes version because Kubernetes 1.24 is not supported v0.19.0, and therefore, does not have an entry in the bundles.Spec.VersionsBundles where it’s selecting from.
 
To address this, we do not need to use the full cluster.Spec in the upgrade management components workflow.

*Description of changes:*
This PR refactors `ClusterManager.ApplyBundles` and `ClusterManager.ApplyReleases` used in the workflow to not need the full `cluster.Spec`

*Testing (if applicable):*
```
--- PASS: TestVSphereKubernetes128UpgradeManagementComponents (753.53s)
PASS
```

```
--- PASS: TestVSphereKubernetes128UbuntuTo129Upgrade (2140.91s)
PASS
```

### DIFF
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -18,7 +18,6 @@ import (
 	"github.com/aws/eks-anywhere/pkg/dependencies"
 	"github.com/aws/eks-anywhere/pkg/files"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
-	"github.com/aws/eks-anywhere/pkg/logger"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
@@ -127,11 +126,17 @@ func readAndValidateClusterSpec(clusterConfigPath string, cliVersion version.Inf
 	return clusterSpec, nil
 }
 
-func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
+func fileSpecBuilderOptsFromClusterOptions(options clusterOptions) []cluster.FileSpecBuilderOpt {
 	var opts []cluster.FileSpecBuilderOpt
 	if options.bundlesOverride != "" {
 		opts = append(opts, cluster.WithOverrideBundlesManifest(options.bundlesOverride))
 	}
+
+	return opts
+}
+
+func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
+	opts := fileSpecBuilderOptsFromClusterOptions(options)
 
 	clusterSpec, err := readAndValidateClusterSpec(options.fileName, version.Get(), opts...)
 	if err != nil {
@@ -157,10 +162,7 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
 }
 
 func newEKSARelease(bundles *releasev1.Bundles, options clusterOptions) (*releasev1.EKSARelease, error) {
-	var opts []cluster.FileSpecBuilderOpt
-	if options.bundlesOverride != "" {
-		opts = append(opts, cluster.WithOverrideBundlesManifest(options.bundlesOverride))
-	}
+	opts := fileSpecBuilderOptsFromClusterOptions(options)
 
 	cliVersion := version.Get()
 	b := cluster.NewFileSpecBuilder(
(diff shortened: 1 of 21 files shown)


cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
```diff
@@ -85,7 +96,7 @@ var upgradeManagementComponentsCmd = &cobra.Command{
 		}
 
 		validator := management.NewUMCValidator(managementCluster, deps.Kubectl)
-		return runner.Run(ctx, clusterSpec, managementCluster, validator)
+		return runner.Run(ctx, bundles, eksaRelease, clusterSpec, managementCluster, validator)
```
d8660091, line 99: It feels strange that bundles, eksaRelease are passed here while clusterSpec contains them, can we not changing the function's parameters but modify the bundles and eksaRelease in clusterSpec?

And, did you verify the fix e2e besides unit testing?

cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -155,10 +161,30 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
...
+
+	cliVersion := version.Get()
+	b := cluster.NewFileSpecBuilder(
+		files.NewReader(files.WithEKSAUserAgent("cli", cliVersion.GitVersion)),
+		cliVersion,
+		opts...,
+	)
+
+	return cluster.GetBundlesManifest(b)
 }
```
g-gaston, line 188: I would advise to build the spec builder with one method and reuse it from the cmd to get the two objects you want

cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
g-gaston, line 99: I would suggest maybe grouping bundles and release in one struct. the method signature is getting quite big
Like `cluster.Release`?

pkg/cluster/builder.go
```diff
@@ -225,3 +226,27 @@ func buildEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles)
 	}
 	return eksaRelease
 }
+
+// GetBundlesManifest builds a Bundles from a FileSpecBuilder.
+func GetBundlesManifest(b FileSpecBuilder) (*releasev1.Bundles, error) {
```
g-gaston, line 231: why do we use a `FileSpecBuilder` to read a bundles?
`FileSpecBuilder` is to build a cluster.Spec
Do we need any of the logic inside the builder? or can we use the `manifests` package directly

pkg/cluster/builder.go
```diff
@@ -225,3 +226,27 @@ func buildEKSARelease(release *releasev1.EksARelease, bundle *releasev1.Bundles)
...
+	bundlesManifest, err := b.getBundles(mReader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting Bundles manifest")
+	}
+
+	return bundlesManifest, nil
+}
+
+// BuildEKSARelease builds an EKSARelease from a FileSpecBuilder and bundles.
+func BuildEKSARelease(b FileSpecBuilder, bundles *releasev1.Bundles) (*releasev1.EKSARelease, error) {
```
g-gaston, line 243: why does "build release" take a bundles?
Each release it already tied to a bundles by definition

cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go
cxbrowne1207: The goal here is to eventually remove the reliance or need to pass the new clusterSpec because building it will throw an error if the customer's cluster is using an unsupported version in the new version of the CLI.

So, I chose to pass them individually with that in mind.

cxbrowne1207:
Closing this stale PR

PR #7607, opened 2024-02-15
Fix check for getting etcd url

abhinavmpandey08:
*Description of changes:*
`GetExternalEtcdReleaseURL` doesn't work for dev builds currently cuz it only checks if the current cluster version is either equal to or greater than the bundle version. The equality check doesn't work for dev bundles because they are of the form `v0.19.0-dev+latest` which doesn't equal the bundle version `v0.19.0`

This PR updates this check to ensure that the cluster version is not less than bundle version.

### DIFF
--- a/internal/test/cluster.go
+++ b/internal/test/cluster.go
@@ -27,7 +27,7 @@ var configFS embed.FS
 
 // DevEksaVersion can be used in tests.
 func DevEksaVersion() v1alpha1.EksaVersion {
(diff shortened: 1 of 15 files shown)


g-gaston (commented):
/lgtm

g-gaston (commented):
/lgtm

mitalipaygude (approved):

PR #7612, opened 2024-02-15
Fix management component upgrades and upgrade plan with Cluster using unsupported Kubernetes versions

cxbrowne1207:
*Issue #, if available:*

*Context:*
The newly introduced sub command upgrade management-components currently has an issue where:

If you have an EKS Anywhere 1.24. You cannot update the management components if the cluster is using Kubernetes version 1.24, because the CLI runs into the following error: while [building the new cluster spec](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/cmd/eksctl-anywhere/cmd/upgrademanagementcomponents.go#L37)

```
It occurs while building the new cluster spec, when trying to [pick the correct versions bundle](https://github.com/aws/eks-anywhere/blob/4de7cda5ac1e6b571d769df82de2391634b750b8/pkg/cluster/fetch.go#L117) for the cluster’s kubernetes version because Kubernetes 1.24 is not supported v0.19.0, and therefore, does not have an entry in the bundles.Spec.VersionsBundles where it’s selecting from.
```

*Description of changes:*
This PR addresses the issue adding a short term building the new cluster spec for the management component upgrades flow without the `VersionsBundles` field. When building the full cluster.Spec definition, we fetch the eksdReleases for the KubernetesVersions for all unique k8s versions specified in the Cluster for both CP and workers.
If the Cluster object is using an unsupported version of Kubernetes, an error thrown
because it does not exist in the Bundles file. This method allows to build a `cluster.Spec` without
encountering this problem when performing only a management component upgrade.

*Testing (if applicable):*
- Manually created a vSphere cluster on Kubernetes version 1.24 with EKS-A v0.18.6
   - Upgrade management-components
   - Checked upgrade plan management-components
```
   [ec2-user@ip-172-31-61-197 vsphere-cluster]$ eksctl-anywhere upgrade management-components -f mgmt04/mgmt04-eks-a-cluster.yaml 
Performing setup and validations
✅ Vsphere provider setup and validation
✅ Control plane ready
✅ Cluster CRDs ready
Upgrading core components
🎉 Management components upgraded!
```

```
[ec2-user@ip-172-31-61-197 vsphere-cluster]$ eksctl-anywhere upgrade plan management-components -f mgmt04/mgmt04-eks-a-cluster.yaml
mgmt04/mgmt04-eks-a-cluster.yaml
Checking new release availability...
All the components are up to date with the latest versions
```
 
- Ran the following E2E tests:
```
--- PASS: TestVSphereKubernetes128UpgradeManagementComponents (753.53s)
PASS
```

```
--- PASS: TestVSphereKubernetes128UbuntuTo129Upgrade (2140.91s)
PASS
```

### DIFF
--- a/cmd/eksctl-anywhere/cmd/options.go
+++ b/cmd/eksctl-anywhere/cmd/options.go
@@ -8,7 +8,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
@@ -19,11 +21,13 @@ import (
 	"github.com/aws/eks-anywhere/pkg/dependencies"
 	"github.com/aws/eks-anywhere/pkg/files"
 	"github.com/aws/eks-anywhere/pkg/kubeconfig"
-	"github.com/aws/eks-anywhere/pkg/logger"
+	"github.com/aws/eks-anywhere/pkg/manifests"
+	"github.com/aws/eks-anywhere/pkg/manifests/bundles"
 	"github.com/aws/eks-anywhere/pkg/providers/cloudstack/decoder"
 	"github.com/aws/eks-anywhere/pkg/types"
 	"github.com/aws/eks-anywhere/pkg/validations"
 	"github.com/aws/eks-anywhere/pkg/version"
+	releasev1 "github.com/aws/eks-anywhere/release/api/v1alpha1"
 )
 
 const defaultTinkerbellNodeStartupTimeout = 20 * time.Minute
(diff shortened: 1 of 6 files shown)


cmd/eksctl-anywhere/cmd/options.go
```diff
@@ -156,10 +160,120 @@ func newClusterSpec(options clusterOptions) (*cluster.Spec, error) {
...
+	s.Bundles = bundles
+	s.Config = config
+
+	s.EKSARelease = eksaRelease
+	return s
+}
+
+func readBasicClusterSpec(clusterConfigPath string, cliVersion version.Info, options clusterOptions) (*cluster.Spec, error) {
+	bundle, err := getBundles(cliVersion, options.bundlesOverride)
+	bundle.Namespace = constants.EksaSystemNamespace
```
g-gaston, line 219: need to do this after the err check

cxbrowne1207: good catch.

g-gaston (commented):
/lgtm

PR #7617, opened 2024-02-16
Revert Tinkerbell upgrade management components e2e test to use kube 1.28

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
The test is trying to make a 1.29 cluster using the newest minor release, but that's not allowed yet. We can only do that after 1.29 is out, with version 0.19.0. Until then, the test needs to be set up for 1.28 instead. This goes for any test trying to set up a 1.29 cluster from the latest minor release.

This PR reverts it to use kube 1.28 instead

*Testing (if applicable):*


test/e2e/tinkerbell_test.go
```diff
@@ -1623,7 +1623,7 @@ func TestTinkerbellSingleNode128To129UbuntuManagementCPUpgradeAPI(t *testing.T)
 	)
 }
 
-func TestTinkerbellKubernetes129UpgradeManagementComponents(t *testing.T) {
+func TestTinkerbellKubernetes128UpgradeManagementComponents(t *testing.T) {
 	release := latestMinorRelease(t)
 	provider := framework.NewTinkerbell(t, framework.WithUbuntu129Tinkerbell())
```
abhay-krishna, line 1628: This needs to be 128 too right?

abhay-krishna (changes requested):

test/e2e/tinkerbell_test.go
cxbrowne1207: You are correct! Thanks

abhay-krishna (approved):

PR #7624, opened 2024-02-17
Fix workload cluster kube-vip update

rahulbabu95:
*Issue #, if available:*
When updating kube-vip manifest for workload cluster, we need the remote client to be able to create the config map with kube-vip manifest data. This change uses remote client instead of management client to create the kube-vip config map for InPlace upgrades. Additionally the upgrader pod is also set to restart on failure as a retry mechanism. 

*Description of changes:*

*Testing (if applicable):*
Tested locally for workload cluster upgrade using InPlace strategy


abhinavmpandey08 (approved):
/approve
/lgtm

PR #7632, opened 2024-02-19
Fix script to get next eksa version

g-gaston:
*Description of changes:*
Just before we release a new minor version, we cut a release branch that we use to run tests and stabilize the release. At this point, there isn't yet any tag for the minor release. The old code was assuming there was one at thus picking the wrong tag. This solves the issue by running the same logic as for main where there are no tags that match the release branch minor version.

*Testing (if applicable):*
Tested manually


abhay-krishna (approved):
/approve

PR #7656, opened 2024-02-21
Fix InPlace upgrade objects cleanup race conditions

abhinavmpandey08:
*Description of changes:*
We noticed that after the `ControlPlaneUpgrade` (CPU) was complete and the corresponding CPU object was deleted, there was a race condition where if the KCP object statuses weren't updated on time, our KCP controller would create another CPU object and try to perform the upgrade again. 

This PR fixes this race condition issue by only deleting the CPU after the KCP object has reached the expected state.
The same fix is applied for in the MachineDeployment controller for cleaning up `MachineDeploymentUpgrade` objects.

### DIFF
--- a/controllers/kubeadmcontrolplane_controller.go
+++ b/controllers/kubeadmcontrolplane_controller.go
@@ -116,9 +116,6 @@ func (r *KubeadmControlPlaneReconciler) reconcile(ctx context.Context, log logr.
 		return ctrl.Result{}, err
 	}
 
-	cpUpgrade := &anywherev1.ControlPlaneUpgrade{}
-	cpuGetErr := r.client.Get(ctx, GetNamespacedNameType(cpUpgradeName(kcp.Name), constants.EksaSystemNamespace), cpUpgrade)
-
 	mhc := &clusterv1.MachineHealthCheck{}
 	if err := r.client.Get(ctx, GetNamespacedNameType(cpMachineHealthCheckName(kcp.Name), constants.EksaSystemNamespace), mhc); err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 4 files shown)


controllers/kubeadmcontrolplane_controller.go
```diff
@@ -131,8 +128,10 @@ func (r *KubeadmControlPlaneReconciler) reconcile(ctx context.Context, log logr.
 		return ctrl.Result{}, err
 	}
 
-	if kcp.Spec.Replicas != nil && (*kcp.Spec.Replicas == kcp.Status.UpdatedReplicas) {
-		if cpuGetErr == nil && cpUpgrade.Status.Ready {
+	cpUpgrade := &anywherev1.ControlPlaneUpgrade{}
+	cpuGetErr := r.client.Get(ctx, GetNamespacedNameType(cpUpgradeName(kcp.Name), constants.EksaSystemNamespace), cpUpgrade)
```
vivek-koppuru, line 132: We aren't actually handling the case where cpuGetErr is nil and it is not the apierrors.IsNotFound error right?

abhinavmpandey08: cpuGetErr == nil is being handled on line 133

controllers/machinedeployment_controller.go
```diff
@@ -139,67 +150,48 @@ func (r *MachineDeploymentReconciler) reconcile(ctx context.Context, log logr.Lo
...
 
-	return ctrl.Result{}, nil
 }
 
 func (r *MachineDeploymentReconciler) inPlaceUpgradeNeeded(md *clusterv1.MachineDeployment) bool {
 	return strings.ToLower(md.Annotations[mdInPlaceUpgradeNeededAnnotation]) == "true"
 }
 
-func (r *MachineDeploymentReconciler) machinesToUpgrade(ctx context.Context, md *clusterv1.MachineDeployment) ([]corev1.ObjectReference, error) {
+func (r *MachineDeploymentReconciler) machinesToUpgrade(ctx context.Context, md *clusterv1.MachineDeployment) ([]*clusterv1.Machine, []corev1.ObjectReference, error) {
```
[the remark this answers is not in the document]
abhinavmpandey08: yeah we are only using the func here so just decided to return both directly

controllers/kubeadmcontrolplane_controller.go
vivek-koppuru: Yea that's what I meant, and ohh I didn't notice that. Should be fine though as even if we explicitly state, it would never really go that far.

vivek-koppuru (approved):

controllers/machinedeployment_controller.go
```diff
@@ -129,8 +126,22 @@ func (r *MachineDeploymentReconciler) reconcile(ctx context.Context, log logr.Lo
...
+	}
+
+	mdUpgrade := &anywherev1.MachineDeploymentUpgrade{}
+	mduGetErr := r.client.Get(ctx, GetNamespacedNameType(mdUpgradeName(md.Name), constants.EksaSystemNamespace), mdUpgrade)
+	if mduGetErr == nil {
+		machinesUpgraded := true
+		for i := range machineList {
+			m := machineList[i]
+			if m.Spec.Version == nil || *m.Spec.Version != mdUpgrade.Spec.KubernetesVersion {
+				machinesUpgraded = false
```
rahulbabu95, line 141: nit: can we break after this.?

PR #7669, opened 2024-02-21
Disable ETCD Learner Mode

jonathanmeier5:
ETCD Learner Mode went to Beta in k8s 1.29 and is now default enabled (see [here](https://kubernetes.io/blog/2023/09/25/kubeadm-use-etcd-learner-mode/), [here](https://github.com/kubernetes/kubeadm/issues/1793#issuecomment-1380585530)).

When a new stacked etcd instance comes up, it joins the cluster in learner mode.

The API Server cannot perform rpc calls against its etcd instance, and fails to come online.

In theory the etcd instance should be promoted to a full member as done in kubeadm [here](https://github.com/kubernetes/kubernetes/blob/fc7325fbec0fe379694e58b82e09f99f166c5e6d/cmd/kubeadm/app/phases/etcd/local.go#L164).

This is not happening because our bottlerocket bootstrap container is not allowing [etcd phase to complete](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/aws/bottlerocket-bootstrap/pkg/kubeadm/controlplane_join.go#L134).

We are disabling this feature for 0.19 and will likely enable it for 0.20 when we also incorporate new CAPI work that allows feature gates to [be mutable](https://github.com/kubernetes-sigs/cluster-api/pull/10154).

The feature gate will [likely](https://github.com/kubernetes/kubeadm/issues/1793#issuecomment-1914077181) go GA in upstream 1.31.

*Issue #, if available:*
This was causing failures on *129StackedEtcdUpgrade end to end tests.

*Testing (if applicable):*
Ran TestVSphereKubernetes128BottlerocketTo129StackedEtcdUpgrade with a custom built controller image that set this EtcdLearnerEnabled feature flag to false.

### DIFF
--- a/pkg/providers/cloudstack/config/template-cp.yaml
+++ b/pkg/providers/cloudstack/config/template-cp.yaml
@@ -67,6 +67,10 @@ spec:
       name: {{.controlPlaneTemplateName}}
   kubeadmConfigSpec:
     clusterConfiguration:
+{{- if (ge (atoi $kube_minor_version) 29) }}
+      featureGates:
+        EtcdLearnerMode: false
+{{- end }}
       imageRepository: {{.kubernetesRepository}}
       etcd:
 {{- if .externalEtcd }}
--- a/pkg/providers/cloudstack/testdata/expected_results_encryption_config_cp_1_29.yaml
+++ b/pkg/providers/cloudstack/testdata/expected_results_encryption_config_cp_1_29.yaml
@@ -59,6 +59,8 @@ spec:
       name: test-control-plane-template-1234567890000
   kubeadmConfigSpec:
     clusterConfiguration:
+      featureGates:
+        EtcdLearnerMode: false
       imageRepository: public.ecr.aws/eks-distro/kubernetes
       etcd:
         local:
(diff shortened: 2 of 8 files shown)


pkg/clustermanager/testdata/testClusterSpec.yaml
```diff
@@ -54,6 +54,9 @@ spec:
     name: docker-test-control-plane
     namespace: default
   kubeadmConfigSpec:
+    clusterConfiguration:
```
jonathanmeier5, line 57: Will fix this dupe

pkg/providers/cloudstack/testdata/expected_results_availability_zones_cp.yaml
```diff
@@ -83,6 +83,9 @@ spec:
       kind: CloudStackMachineTemplate
       name: test-control-plane-template-1234567890000
   kubeadmConfigSpec:
+    clusterConfiguration:
```
jonathanmeier5, line 86: Fix this dupe

pkg/clustermanager/testdata/testClusterSpec.yaml
abhay-krishna, line 57: shouldn't this be added to the `clusterConfiguration` field below?

PR #7672, opened 2024-02-21
Configure SSM delivery

abhinavmpandey08:
*Description of changes:*
Currently, we don't set the delivery timeout for SSM commands. This defaults the delivery timeout to 3600s or 1 hour which is a really long time to wait for the command to just be delivered. If the SSM agent is unhealthy on the test runner, then waiting 1 hour doesn't make sense.

This PR configures the delivery timeout to 5 min. The tests also have another timeout called `Execution Timeout` which specifies how long the command has to finish execution, which is set separately and varies by command. 

### DIFF
--- a/internal/pkg/ssm/command.go
+++ b/internal/pkg/ssm/command.go
@@ -15,7 +15,10 @@ import (
 	"github.com/aws/eks-anywhere/pkg/retrier"
 )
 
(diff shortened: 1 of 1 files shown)


abhay-krishna (approved):
/approve
/woof

abhay-krishna (approved):
/approve

PR #7729, opened 2024-02-28
Improve validation error messages

chrisdoherty4:
* Add information to error messages regarding the total unavailable replicas
* Provide a more meaningful explanation when the API server isn't ready
* Lowercase the word 'Ready' which makes it seem like 'Ready' has special meaning in the error message.
* Removed redundant check for ready control plane nodes.

### DIFF
--- a/pkg/executables/kubectl.go
+++ b/pkg/executables/kubectl.go
@@ -888,20 +888,18 @@ func (k *Kubectl) ValidateControlPlaneNodes(ctx context.Context, cluster *types.
 	observedGeneration := cp.Status.ObservedGeneration
 	generation := cp.Generation
 	if observedGeneration != generation {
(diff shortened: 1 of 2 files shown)


pkg/executables/kubectl.go
```diff
@@ -888,20 +888,18 @@ func (k *Kubectl) ValidateControlPlaneNodes(ctx context.Context, cluster *types.
...
+		return errors.New("api server is not ready")
 	}
 
 	if cp.Status.UnavailableReplicas != 0 {
-		return fmt.Errorf("%v control plane replicas are unavailable", cp.Status.UnavailableReplicas)
+		return fmt.Errorf("%v/%v control plane replicas are unavailable",
+			cp.Status.UnavailableReplicas, cp.Status.Replicas)
 	}
 
-	if cp.Status.ReadyReplicas != cp.Status.Replicas {
```
chrisdoherty4, line 902: Redundant. UnavailableReplicas covers nodes not ready or not created.

jacobweinstock (approved):

PR #7755, opened 2024-02-29
Add api server extra args map to cluster spec

sp1999:
*Issue #, if available:*
[#2091](https://github.com/aws/eks-anywhere-internal/issues/2091)

*Description of changes:* 
Design doc: #7853 

These changes allow for custom configuration of the api server flags in the cluster spec through `spec.controlPlaneConfiguration.APIServerExtraArgs`. If specified, this will be added to apiServerExtraArgs for the kubeadm cluster configuration in kcp object. A feature flag needs to be enabled in order to configure the api server flags. No validations are done for any of the flags configured. It is expected from the users that if they are enabling the feature flag, they know what values to configure for the flags.

*Testing (if applicable):*
- Tested locally for create and upgrade operations with the api-server flags
- Unit tests are added for APIServerExtraArgs and other functions
- E2E tests for creating and upgrading cluster with ApiServerExtraArgs are added for vSphere and CloudStack providers

EKS-A docs need to be updated to add `spec.controlPlaneConfiguration.apiServerExtraArgs` as an optional configuration for the cluster spec


pkg/api/v1alpha1/cluster_types.go
```diff
@@ -420,6 +422,21 @@ func (n *Endpoint) CloudStackEqual(o *Endpoint) bool {
 	return n.Host == o.Host
 }
 
+type ApiServerExtraArgs struct {
```
vivek-koppuru, line 425: I thought we were doing a map and not a struct right?

sp1999: Yes, forgot to update the branch

pkg/api/v1alpha1/cluster.go
```diff
@@ -494,6 +495,16 @@ func validateControlPlaneCertSANs(cfg *Cluster) error {
 	return nil
 }
 
+func validateControlPlaneAPIServerExtraArgs(clusterConfig *Cluster) error {
+	allowedFlags := map[string]string{"service-account-issuer": "", "service-account-jwks-uri": ""}
```
vivek-koppuru, line 499: Let's add a feature flag to disable this validation, should be a simple change as we discussed in doc right?

pkg/providers/vsphere/template.go
```diff
@@ -145,6 +145,7 @@ func buildTemplateMapCP(
 		Append(clusterapi.AwsIamAuthExtraArgs(clusterSpec.AWSIamConfig)).
 		Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig)).
 		Append(clusterapi.EtcdEncryptionExtraArgs(clusterSpec.Cluster.Spec.EtcdEncryption)).
+		Append(clusterapi.APIServerExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration.APIServerExtraArgs)).
```
vivek-koppuru, line 148: Did we skip snow provider?

test/e2e/apiserverextraargs.go
```diff
@@ -0,0 +1,18 @@
...
+package e2e
+
+import (
+	"github.com/aws/eks-anywhere/test/framework"
+)
+
+func runApiServerExtraArgsUpgradeFlow(test *framework.ClusterE2ETest, clusterOpts ...[]framework.ClusterE2ETestOpt) {
+	test.GenerateClusterConfig()
+	test.CreateCluster()
+	for _, opts := range clusterOpts {
```
vivek-koppuru, line 13: Should we validate that the kcp object indeed has these flags set and not set?

test/e2e/cloudstack_test.go
```diff
@@ -17,6 +17,46 @@ import (
...
+		removeApiServerExtraArgsclusterOpts,
+		framework.WithClusterUpgrade(
+			api.RemoveAllAPIServerExtraArgs(),
+		),
+	)
+	runApiServerExtraArgsUpgradeFlow(
+		test,
+		addApiServerExtraArgsclusterOpts,
+		removeApiServerExtraArgsclusterOpts,
+	)
```
vivek-koppuru, line 57: Good design of specifying multiple upgrades 💯

test/e2e/cloudstack_test.go
```diff
@@ -17,6 +17,46 @@ import (
 	"github.com/aws/eks-anywhere/test/framework"
 )
 
+// ApiServerExtraArgs
+func TestCloudStackKubernetes129RedHat8ApiServerExtraArgsSimpleFlow(t *testing.T) {
+	test := framework.NewClusterE2ETest(
+		t,
+		framework.NewCloudStack(t, framework.WithCloudStackRedhat129()),
+		framework.WithClusterFiller(api.WithKubernetesVersion(v1alpha1.Kube129)),
+		framework.WithClusterFiller(api.WithControlPlaneAPIServerExtraArgs("service-account-issuer", "https://my-custom-issuer-url")),
```
vivek-koppuru, line 26: This is a fake issuer right? It doesn't cause any issues?

internal/pkg/api/cluster.go
```diff
@@ -126,6 +126,25 @@ func WithControlPlaneLabel(key string, val string) ClusterFiller {
 	}
 }
 
+// WithControlPlaneAPIServerExtraArgs adds the APIServerExtraArgs to the cluster spec.
+func WithControlPlaneAPIServerExtraArgs(key string, val string) ClusterFiller {
+	return func(c *anywherev1.Cluster) {
+		if c.Spec.ControlPlaneConfiguration.APIServerExtraArgs == nil {
+			c.Spec.ControlPlaneConfiguration.APIServerExtraArgs = map[string]string{}
+		}
+		c.Spec.ControlPlaneConfiguration.APIServerExtraArgs[key] = val
```
vivek-koppuru, line 135: Did we figure out what happens if we specify both this and pod iam config?

pkg/api/v1alpha1/cluster.go
chrisdoherty4: Do you mean a feature flag or an addition to skip validation?

pkg/api/v1alpha1/cluster.go
```diff
@@ -494,6 +495,16 @@ func validateControlPlaneCertSANs(cfg *Cluster) error {
...
+	for k := range clusterConfig.Spec.ControlPlaneConfiguration.APIServerExtraArgs {
+		if _, ok := allowedFlags[k]; !ok {
+			return fmt.Errorf("invalid ControlPlaneConfiguration.APIServerExtraArgs; only 'service-account-issuer' and 'service-account-jwks-uri' flags are accepted currently")
```
chrisdoherty4, line 502: Can we collect errors and use an aggregate so the user doesn't have to repeatedly run the CLI? 

https://github.com/chrisdoherty4/aws-eks-anywhere/blob/ad52e5bf1ac9ffff2b05c79844bb1534cc230c6e/pkg/errors/aggregate.go#L15

pkg/providers/vsphere/template.go
sp1999: Added snow

test/e2e/apiserverextraargs.go
sp1999: Yes, we should. Added the validation in the flow

test/e2e/cloudstack_test.go
sp1999: Yep, there are no issues in configuring any value for the flags. There are no checks for that

test/e2e/cloudstack_test.go
sp1999: Thank you

internal/pkg/api/cluster.go
sp1999: `service-account-issuer` flag in `spec.controlPlaneConfiguration.apiServerExtraArgs` overrides the one in `spec.podIamConfig` based on the order in which we are appending the values. I am yet to check if we can append those values as a comma-separated string so that both can be enabled

pkg/api/v1alpha1/cluster.go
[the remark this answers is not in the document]
sp1999: Updated

sp1999: Good suggestion. Will update that in a separate PR

test/e2e/apiserverextraargs.go
chrisdoherty4, line 1: `apiserverextraargs.go` => `api_server_extra_args.go`

pkg/api/v1alpha1/cluster.go
sp1999: Feature flag

internal/pkg/api/cluster.go
sp1999: Added support to enable both config. PodIamConfig will be appended after the existing controlPlaneConfiguration `service-account-issuer` as a comma-separated string

pkg/api/v1alpha1/cluster_types_test.go
```diff
@@ -1460,6 +1460,10 @@ func TestControlPlaneConfigurationEqual(t *testing.T) {
 	taints1DiffOrder := []corev1.Taint{taint2, taint1}
 	taints2 := []corev1.Taint{taint1}
 
+	var emptyapiServerExtraArgs map[string]string
```
vivek-koppuru, line 1463:
```suggestion
	var emptyApiServerExtraArgs map[string]string
```

pkg/clusterapi/identity.go
```diff
@@ -103,19 +103,34 @@ func configureOIDCInKubeadmControlPlane(kcp *controlplanev1.KubeadmControlPlane,
...
 func configurePodIamAuthInKubeadmControlPlane(kcp *controlplanev1.KubeadmControlPlane, podIamConfig *v1alpha1.PodIAMConfig) {
 	if podIamConfig == nil {
 		return
 	}
 
 	apiServerExtraArgs := kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.APIServer.ExtraArgs
-	for k, v := range PodIAMAuthExtraArgs(podIamConfig) {
-		apiServerExtraArgs[k] = v
+	if _, ok := apiServerExtraArgs["service-account-issuer"]; ok && podIamConfig.ServiceAccountIssuer != "" {
+		apiServerExtraArgs["service-account-issuer"] += "," + podIamConfig.ServiceAccountIssuer
```
vivek-koppuru, line 123: I see the concern here, this seems like a mix of assuming there is only one field and then a for loop below right? I think we should try to be consistent, either remove the for loop below or do this check within that for loop where if service-account-issuer already exists, then append.

pkg/providers/cloudstack/template.go
```diff
@@ -121,10 +121,14 @@ func buildTemplateMapCP(clusterSpec *cluster.Spec) (map[string]interface{}, erro
...
 		Append(clusterapi.AwsIamAuthExtraArgs(clusterSpec.AWSIamConfig)).
-		Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig)).
+		Append(clusterapi.APIServerExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration.APIServerExtraArgs)).
 		Append(clusterapi.EtcdEncryptionExtraArgs(clusterSpec.Cluster.Spec.EtcdEncryption)).
 		Append(sharedExtraArgs)
-
+	if _, ok := apiServerExtraArgs["service-account-issuer"]; ok && clusterSpec.Cluster.Spec.PodIAMConfig.ServiceAccountIssuer != "" {
+		apiServerExtraArgs["service-account-issuer"] += "," + clusterSpec.Cluster.Spec.PodIAMConfig.ServiceAccountIssuer
+	} else {
+		apiServerExtraArgs.Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig))
```
vivek-koppuru, line 130: And looks like we don't have a for loop here, this is confusing compared to the above comment I made.

test/framework/cluster/validations/cluster.go
```diff
@@ -360,3 +364,22 @@ func getWorkerNodeMachineSets(ctx context.Context, vc clusterf.StateValidationCo
...
+	}
+	if kcp == nil {
+		return errors.New("KubeadmControlPlane object not found")
+	}
+	apiServerExtraArgsKCP := kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.APIServer.ExtraArgs
+	apiServerExtraArgsSpec := vc.ClusterSpec.Cluster.Spec.ControlPlaneConfiguration.APIServerExtraArgs
+	if apiServerExtraArgsKCP == nil {
+		return fmt.Errorf("kcp object APIServerExtraArgs is nil expected: %v", apiServerExtraArgsSpec)
+	}
+	if !reflect.DeepEqual(apiServerExtraArgsKCP, apiServerExtraArgsSpec) {
```
vivek-koppuru, line 381: Does this test actually pass? What we specify in our Cluster spec is not the full set of apiserver flags that KCP will return right?

test/e2e/apiserverextraargs.go
sp1999: Updated

pkg/clusterapi/identity.go
sp1999: There is a for loop here as well. It's hidden within the `Append()` function

pkg/providers/cloudstack/template.go
sp1999: There is a for loop here as well. It's hidden within the `Append()` function

pkg/api/v1alpha1/cluster.go
chrisdoherty4: Feature flags should be short lived and have a rough criteria defining when they can be removed.

Looking at the change set, a feature flag isn't justified here. Instead, I think you want to be able to disable the validation, which is a permanent feature.

chrisdoherty4: I reviewed the design doc. Sorry I didn't catch this when reviewing: I don't think we meet the criteria for a feature falg. This is a feature that we may deprecate in the future.

internal/pkg/api/cluster.go
chrisdoherty4: So `podIamConfig` will superseed the service-account-issuer flag? 

That detail needs to be included in the documentation for both `podIamConfig` and the `apiServerExtraArgs`. I'm open to docs being done in a separate PR, but we can't forget to add it in _both places_.

sp1999: No, `podIamConfig` won't supersede the service-account-issuer flag. If the `service-account-issuer` is already configured with `apiServerExtraArgs`, then `podIamConfig` will just append to it, making sure that both issuer urls are accepted for the flag

I was planning to create a separate PR for the docs. I will include this detail there

pkg/clusterapi/identity.go
sp1999: Added the condition within the for-loop for all providers

pkg/api/v1alpha1/cluster_types_test.go
sp1999: Updated

pkg/api/v1alpha1/cluster_types.go
```diff
@@ -307,6 +307,8 @@ type ControlPlaneConfiguration struct {
 	CertSANs []string `json:"certSans,omitempty"`
 	// MachineHealthCheck is a control-plane level override for the timeouts and maxUnhealthy specified in the top-level MHC configuration. If not configured, the defaults in the top-level MHC configuration are used.
 	MachineHealthCheck *MachineHealthCheck `json:"machineHealthCheck,omitempty"`
+	// APIServerExtraArgs defines the flags to configure for the API server.
+	APIServerExtraArgs map[string]string `json:"apiServerExtraArgs,omitempty"`
```
vivek-koppuru, line 311:
```suggestion
	ApiServerExtraArgs map[string]string `json:"apiServerExtraArgs,omitempty"`
```

Should we change to this casing everywhere in the PR?

pkg/clusterapi/identity.go
```diff
@@ -103,19 +103,36 @@ func configureOIDCInKubeadmControlPlane(kcp *controlplanev1.KubeadmControlPlane,
...
 	if podIamConfig == nil {
 		return
 	}
 
 	apiServerExtraArgs := kcp.Spec.KubeadmConfigSpec.ClusterConfiguration.APIServer.ExtraArgs
 	for k, v := range PodIAMAuthExtraArgs(podIamConfig) {
+		if k == "service-account-issuer" {
+			if _, ok := apiServerExtraArgs[k]; ok {
+				apiServerExtraArgs[k] += "," + v
+				continue
```
vivek-koppuru, line 126: Nit: Instead of continue here, could we just set the string to v initially, and then append to that string in this case so that it will still be assigned through the below line?

pkg/features/features.go
```diff
@@ -63,3 +64,11 @@ func VSphereInPlaceUpgradeEnabled() Feature {
 		IsActive: globalFeatures.isActiveForEnvVar(VSphereInPlaceEnvVar),
 	}
 }
+
+// APIServerExtraArgsValidationDisabled is the feature flag for disabling the API server extra args validation.
+func APIServerExtraArgsValidationDisabled() Feature {
```
vivek-koppuru, line 69: We will check out the skip validation flag?

pkg/providers/nutanix/template.go
```diff
@@ -161,8 +161,17 @@ func buildTemplateMapCP(
 	format := "cloud-config"
 	apiServerExtraArgs := clusterapi.OIDCToExtraArgs(clusterSpec.OIDCConfig).
 		Append(clusterapi.AwsIamAuthExtraArgs(clusterSpec.AWSIamConfig)).
-		Append(clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig)).
+		Append(clusterapi.APIServerExtraArgs(clusterSpec.Cluster.Spec.ControlPlaneConfiguration.APIServerExtraArgs)).
 		Append(clusterapi.EtcdEncryptionExtraArgs(clusterSpec.Cluster.Spec.EtcdEncryption))
+	for k, v := range clusterapi.PodIAMAuthExtraArgs(clusterSpec.Cluster.Spec.PodIAMConfig) {
```
vivek-koppuru, line 166: This for loop can be extracted out to a separate file right? How much scope is it to leverage the configure method we use for snow I think?

pkg/features/features.go
sp1999: Yes

pkg/providers/nutanix/template.go
sp1999: Created new function in `extraargs.go` which runs this for loop and sets the `podIamAuthExtraArgs`

pkg/clusterapi/identity.go
sp1999: Updated

pkg/api/v1alpha1/cluster_types.go
sp1999: `golangci-lint` check suggests all capital letters for [initialisms](https://go.dev/wiki/CodeReviewComments#:~:text=return%0A%7D%0A//%20use%20x-,Initialisms,-Words%20in%20names). I can update it to API everywhere for consistency

pkg/api/v1alpha1/cluster.go
sp1999: Moved this validation to preflight validations to allow skipping it through `--skip-validations` option

pkg/api/v1alpha1/cluster.go
sp1999: Removed feature flag and added api server extra args validation to the `skip-validations` option

chrisdoherty4 (commented):
These changes look good overall. Thanks!

internal/pkg/api/cluster.go
```diff
@@ -126,6 +126,26 @@ func WithControlPlaneLabel(key string, val string) ClusterFiller {
 	}
 }
 
+// WithControlPlaneAPIServerExtraArgs adds the APIServerExtraArgs to the cluster spec.
+func WithControlPlaneAPIServerExtraArgs() ClusterFiller {
```
chrisdoherty4, line 130: I didn't notice this before. Using this function name makes me think this should be far more flexible than only supporting JWKS args. We can either figure out a way to make it more flexible, or rename it to be specific to handling JWKS.

pkg/features/features.go
```diff
@@ -7,6 +7,7 @@ const (
 	UseNewWorkflowsEnvVar           = "USE_NEW_WORKFLOWS"
 	UseControllerForCli             = "USE_CONTROLLER_FOR_CLI"
 	VSphereInPlaceEnvVar            = "VSPHERE_IN_PLACE_UPGRADE"
+	APIServerExtraArgsEnabledEnvVar = "API_SERVER_EXTRA_ARGS_ENABLED"
```
chrisdoherty4, line 10: Given our conversation, what's the need for this feature flag?

internal/pkg/api/cluster.go
sp1999: You are right about that. We updated the design to allow configuring any of the flags in the map through a feature flag

pkg/clusterapi/extraargs.go
```diff
@@ -136,6 +145,19 @@ func (e ExtraArgs) Append(args ExtraArgs) ExtraArgs {
...
+func SetPodIAMAuthExtraArgs(podIamConfig *v1alpha1.PodIAMConfig, apiServerExtraArgs map[string]string) {
+	for k, v := range PodIAMAuthExtraArgs(podIamConfig) {
+		args := v
+		if k == "service-account-issuer" {
+			if _, ok := apiServerExtraArgs[k]; ok {
+				args = apiServerExtraArgs[k] + "," + v
+			}
+		}
+		apiServerExtraArgs[k] = args
+	}
```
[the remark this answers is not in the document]
sp1999: Updated

pkg/api/v1alpha1/cluster.go
```diff
@@ -494,6 +496,37 @@ func validateControlPlaneCertSANs(cfg *Cluster) error {
 	return nil
 }
 
+func validateControlPlaneAPIServerExtraArgs(clusterConfig *Cluster) error {
+	if clusterConfig.Spec.ControlPlaneConfiguration.APIServerExtraArgs != nil && !features.IsActive(features.APIServerExtraArgsEnabled()) {
+		return errors.New("please enable feature flag to configure APIServerExtraArgs")
```
vivek-koppuru, line 501:
```suggestion
		return fmt.Errorf("configuring APIServerExtraArgs is not supported. Set env var %v to enable", features.APIServerExtraArgsEnabledEnvVar)
```

test/framework/cluster/validations/cluster.go
```diff
@@ -360,3 +363,24 @@ func getWorkerNodeMachineSets(ctx context.Context, vc clusterf.StateValidationCo
 	}
 	return ms.Items, nil
 }
+
+func validateKCPobjectForAPIServerExtraArgs(ctx context.Context, vc clusterf.StateValidationConfig) error {
```
vivek-koppuru, line 367:
```suggestion
func validateKCPForAPIServerExtraArgs(ctx context.Context, vc clusterf.StateValidationConfig) error {
```

pkg/api/v1alpha1/cluster.go
```diff
@@ -494,6 +496,37 @@ func validateControlPlaneCertSANs(cfg *Cluster) error {
...
+		"oidc-required-claim",
+		"oidc-username-claim",
+		"oidc-username-prefix",
+	}
+	if clusterConfig.Spec.IdentityProviderRefs != nil {
+		for _, ref := range clusterConfig.Spec.IdentityProviderRefs {
+			if ref.Kind == OIDCConfigKind {
+				for _, flag := range oidcFlags {
+					if _, has := clusterConfig.Spec.ControlPlaneConfiguration.APIServerExtraArgs[flag]; has {
+						return fmt.Errorf("%s flag is already configured in OIDCConfig. please remove it from apiServerExtraArgs", flag)
```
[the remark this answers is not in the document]
sp1999: The idea behind that was if someone is referencing the OIDCConfig CRD, then they must have configured atleast some of the oidc flags and we don't want them to configure the other flags through the map. Either they configure it completely through the map or they do it only through OIDCConfig but not both. But I do get your point. I will update the error message to make it more explicit

internal/pkg/api/cluster.go
```diff
@@ -126,6 +126,26 @@ func WithControlPlaneLabel(key string, val string) ClusterFiller {
...
+	return func(c *anywherev1.Cluster) {
+		if c.Spec.ControlPlaneConfiguration.APIServerExtraArgs == nil {
+			c.Spec.ControlPlaneConfiguration.APIServerExtraArgs = map[string]string{}
+		}
+		issuerURL := "https://" + c.Spec.ControlPlaneConfiguration.Endpoint.Host
+		c.Spec.ControlPlaneConfiguration.APIServerExtraArgs["service-account-jwks-uri"] = issuerURL + "/openid/v1/jwks"
```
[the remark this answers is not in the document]
sp1999: Yes

pkg/api/v1alpha1/cluster.go
sp1999: Updated

test/framework/cluster/validations/cluster.go
sp1999: Updated

vivek-koppuru (approved):

PR #7769, opened 2024-03-02
Update error msgs for OSImageURL validation:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
As the error message is provided to Users, including the expected Kubernetes version format to the error message allows to User to clearly understand how to resolve the validation error.

*Testing (if applicable):*


pkg/providers/tinkerbell/validate.go
```diff
@@ -162,14 +162,82 @@ func validateK8sVersionInOSImageURLs(spec *ClusterSpec) error {
...
+
+	if len(seps) > 1 {
+		result[0] = result[0] + ","
+	}
+
+	return strings.Join(result, " ")
+}
+
+// deduplicate returns a new slice with duplicates values removed.
+func deduplicate(s []string) []string {
```
rahulbabu95, line 213: wondering if this is needed, since we have the control over the input and this is not coming from user?

pkg/providers/tinkerbell/validate.go
```diff
@@ -162,14 +162,82 @@ func validateK8sVersionInOSImageURLs(spec *ClusterSpec) error {
...
+// the result will be "1.29, 1-29, 1_29, or 129".
+func permutations(v string, separators []string) string {
+	result := []string{v}
+	split := strings.Split(v, ".")
+	if len(split) == 1 {
+		return v
+	}
+
+	seps := remove(".", deduplicate(separators))
+	for idx, sep := range seps {
```
rahulbabu95, line 187: can we simplify this logic to just parsing in the version and concatenating the seps to list out all accepted versions in a single for loop
```suggestion
        version := strings.Split(v, ".")
        result := []string{v}
        major := version[0]
        minor := version[1]
	for idx, sep := range seps {
	       elem := fmt.Sprintf("%s%s%s",major,sep,minor)

	       if idx == len(seps)-1 {
	       result = append(result, "or")
	       } else if idx != len(seps)-2 {
			elem = elem + ","
		}
		result = append(result, elem)
	}
	return strings.Join(result, " ")
}
```

pkg/providers/tinkerbell/validate.go
```diff
@@ -162,14 +162,82 @@ func validateK8sVersionInOSImageURLs(spec *ClusterSpec) error {
...
+
+// deduplicate returns a new slice with duplicates values removed.
+func deduplicate(s []string) []string {
+	if len(s) <= 1 {
+		return s
+	}
+	result := []string{}
+	seen := make(map[string]struct{})
+	for _, val := range s {
+		val := strings.ToLower(val)
```
rahulbabu95, line 220: since these are all symbols we might not need this too!

pkg/providers/tinkerbell/validate_test.go
```diff
@@ -0,0 +1,31 @@
...
+		initial    string
+		separators []string
+		want       string
+	}{
+		{"125", []string{"-", "_", ""}, "125"},
+		{"1.26", []string{"-", "_", ""}, "1.26, 1-26, 1_26 or 126"},
+		{"1.27", []string{"."}, "1.27"},
+		{"1.28", []string{".", "."}, "1.28"},
+		{"1.29", []string{"-", "-", ""}, "1.29, 1-29 or 129"},
+		{"1.29.1", []string{"-", "_", "_", ""}, "1.29.1, 1-29-1, 1_29_1 or 1291"},
```
rahulbabu95, line 20: we don't support patch version on our spec, so we don't have to be covering this. hence a simpler approach might work?

pkg/providers/tinkerbell/validate.go
jacobweinstock: the control we have over the input is "business" level logic, this function is the "application" level logic. Coupling those together creates for us future issues when business logic changes. I don't recommend we couple them.

pkg/providers/tinkerbell/validate_test.go
jacobweinstock: what versions we support is business level logic. Coupling this with application logic will cause us future maintainability issues in the future. I don't recommend we couple them.

pkg/providers/tinkerbell/validate.go
jacobweinstock: this changes the function to only support a single `.` as a separator. In my opinion, this more generic implementation is basic, flexible, and simple enough to be future proof. I'm not convinced of the value of the trade off of a few less lines of code.

pkg/providers/tinkerbell/validate.go
jacobweinstock: this function doesnt limit the type of strings in the slice. the values could be characters. coupling the inputs to our business logic creates a rigid implementation.

pkg/providers/tinkerbell/validate.go
```diff
@@ -150,8 +150,8 @@ func validateK8sVersionInOSImageURLs(spec *ClusterSpec) error {
 		}
 
 		if !containsK8sVersion(spec.ControlPlaneMachineConfig().Spec.OSImageURL, string(spec.Cluster.Spec.KubernetesVersion)) {
-			return fmt.Errorf("missing kube version from control plane machine config OSImageURL: url=%v, version=%v",
-				spec.ControlPlaneMachineConfig().Spec.OSImageURL, spec.Cluster.Spec.KubernetesVersion)
+			return fmt.Errorf("missing kube version from control plane machine config OSImageURL, the image url must include %v: url=%v, version=%v",
```
mitalipaygude, line 153:
```suggestion
			return fmt.Errorf("missing kube version from control plane machine config OSImageURL, the image url must include %s: url=%v, version=%v",
```

PR #7779, opened 2024-03-05
Use an uncached direct client to read the kcp and md to bypass delay in cache

rahulbabu95:
*Issue #, if available:*
Some of the InPlace tests fail intermittently after a successful upgrade during the validation phase as some of the InPlace CRs are not cleaned up. We use an `in-place-upgrade-needed` on the kcp and md objects to identify the InPlace flow and remove those annotations once the upgrade is successful. In some cases, after the InPlace upgrade is successful the cache used to read the state of objects doesn't get updated quick enough. This leads to our controllers creating the InPlace custom resources again which leads to another upgrade on the node. Moreover since the annotation is removed, cache does get updated eventually leaving these objects in an untracked state as the kcp and md controller do not reconcile for InPlace flow anymore after the removal of annotation. 

*Description of changes:*
Use an uncached direct client to read kcp and md, so that we track these objects on realtime. This way we can rely on annotations on the objects to reconcile for the InPlace flow.

*Testing (if applicable):*
Ran all InPlace tests locally.

### DIFF
--- a/controllers/factory.go
+++ b/controllers/factory.go
@@ -602,6 +602,7 @@ func (f *Factory) WithKubeadmControlPlaneReconciler() *Factory {
 
 		f.reconcilers.KubeadmControlPlaneReconciler = NewKubeadmControlPlaneReconciler(
 			f.manager.GetClient(),
+			f.manager.GetAPIReader(),
 		)
 
 		return nil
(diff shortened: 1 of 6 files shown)


abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #7788, opened 2024-03-05
Bump controller-runtime to v0.16.5

g-gaston:
## Description of changes
Waaaaaay too many changes in this PR. In general, controller-runtime updates are painful. More over, by pushing this change for so long, we got into a situation where a lot of things need to change at once.

### `controller-runtime` v0.16.5
- Fake client breaking changes:
  - Start using `WithStatusSubresource` when we care about status being treated differently through `Update` calls. This is basically all controllers.
  - Add a finalizer when creating an object directly with a deletion timestamp. This make sense since any controller interested in the delete flow should be already adding finalizers.
- Signature of `Watches` for creating a controller has been simplified and doesn't need the `source.Kind` wrapper.
- Manager options signature has changed. Now metrics and webhook fields are nested in its own second level struct.
- Validation webhooks now return a `(admission.Warnings, error)` instead of just `error`. We are not using this functionality for now, but we had to update all our webhooks and test to follow the new signature.
- `handler.MapFunc` has changed, now they take a context as well. We don't use the context but need to conform to the new signature.

### `cluster-api` v1.6.2
- When setting up cluster tracker, the variable `DefaultIndexes` is removed, now we use `[]Index{NodeProviderIDIndex}`

### CAPC
We are still waiting for capc to be updated to capi v1.5, which would enable us to move to controller-runtime v0.15. It's taking quite long and we got to the point where this is preventing us from updating moduled (like helm) that we need to update ASAP.

I opted for "vendoring" just the api structs we need from capc. In order to contain changes and allow for an easy "revert" to use the original module (whenever we can), I created a dummy module so I can just use a `replace` in our go.mod and point to our vendored folder.

I put this in `internal/thirdparty/capc`. You don't need to review that code, I just copied it.

### Assertions in gomock
I had to change some of the assertions in the cluster manager tests. Now they are less strict since they use type instead of content assertion. TBH, they were wrong from the beginning since they expected current and new spec to be the same, so if so, this makes the test slightly better.

I checked and I believe all this is already dead code because of the workflow refactors we did in v0.19.0. It's just that the code hasn't been deleted yet. I'll push to get all this crap removed.

### CAPV
They just change the package where they have the api structs from `api/[version]` to `apis/[version]`.

## Testing
I ran this through all the docker 1.28 tests. All green.


pkg/api/v1alpha1/cluster_webhook.go
```diff
@@ -118,18 +116,24 @@ func (r *Cluster) ValidateUpdate(old runtime.Object) error {
...
 	if len(allErrs) != 0 {
-		return apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
+		return nil, apierrors.NewInvalid(GroupVersion.WithKind(ClusterKind).GroupKind(), r.Name, allErrs)
 	}
 
-	return nil
+	return nil, nil
+}
+
+func supportsEtcdEncryption(cluster *Cluster) bool {
```
g-gaston, line 133: nothing changed here, just minor refactor to reduce cyclomatic complexity

drewvanstone (approved):
Lot of changes! We'll just fix-forward if we run into anything

g-gaston:
I checked codecov.
There is no file losing coverage, it's just coverage that was already missing.
Most of them are webhook.

I'm merging as the e2e tests gave me good confidence on the change.

PR #7790, opened 2024-03-06
Add Cluster Mover

tatlat:
*Issue #, if available:*

*Description of changes:*
Currently for creation of management clusters, we apply EKS-Anywhere resources and waiting for conditions twice (once on bootstrap and again on workload). This causes the create command to take longer than necessary. This PR adds a way to move resources between clusters, which should shorten the time it takes to create management clusters.

*Testing (if applicable):*

### DIFF
--- a/pkg/clustermanager/eksa_mover.go
+++ b/pkg/clustermanager/eksa_mover.go
@@ -1,0 +1,137 @@
+package clustermanager
+
+import (
+	"context"
+	"math"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clients/kubernetes"
+	"github.com/aws/eks-anywhere/pkg/cluster"
+	"github.com/aws/eks-anywhere/pkg/retrier"
+)
+
+// MoverOpt allows to customize a Mover on construction.
+type MoverOpt func(*Mover)
+
+// Mover applies the cluster spec to the management cluster and waits
+// until the changes are fully reconciled.
+type Mover struct {
+	log                logr.Logger
+	clientFactory      ClientFactory
+	moveClusterTimeout time.Duration
+	retryBackOff       time.Duration
+}
+
+// NewMover builds an Mover.
+func NewMover(log logr.Logger, clientFactory ClientFactory, opts ...MoverOpt) *Mover {
+	m := &Mover{
+		log:                log,
+		clientFactory:      clientFactory,
+		moveClusterTimeout: applyClusterSpecTimeout,
+		retryBackOff:       retryBackOff,
+	}
+
+	for _, opt := range opts {
+		opt(m)
+	}
+
+	return m
+}
+
+// WithMoverNoTimeouts disables the timeout for all the waits and retries in management upgrader.
+func WithMoverNoTimeouts() MoverOpt {
+	return func(a *Mover) {
+		maxTime := time.Duration(math.MaxInt64)
+		a.moveClusterTimeout = maxTime
+	}
+}
+
+// WithMoverApplyClusterTimeout allows to configure how long the mover retries
+// to apply the objects in case of failure.
+// Generally only used in tests.
+func WithMoverApplyClusterTimeout(timeout time.Duration) MoverOpt {
+	return func(m *Mover) {
+		m.moveClusterTimeout = timeout
+	}
+}
+
+// WithMoverRetryBackOff allows to configure how long the mover waits between requests
+// to update the cluster spec objects and check the status of the Cluster.
+// Generally only used in tests.
+func WithMoverRetryBackOff(backOff time.Duration) MoverOpt {
+	return func(m *Mover) {
+		m.retryBackOff = backOff
+	}
+}
+
+// Move applies the cluster's namespace and spec without checking for reconcile conditions.
+func (m *Mover) Move(ctx context.Context, spec *cluster.Spec, fromClient, toClient kubernetes.Client) error {
+	m.log.V(3).Info("Moving the cluster object")
+	err := retrier.New(
+		m.moveClusterTimeout,
+		retrier.WithRetryPolicy(retrier.BackOffPolicy(m.retryBackOff)),
+	).Retry(func() error {
+		// read the cluster from bootstrap
+		cluster := &v1alpha1.Cluster{}
+		if err := fromClient.Get(ctx, spec.Cluster.Name, spec.Cluster.Namespace, cluster); err != nil {
+			return errors.Wrapf(err, "reading cluster from source")
+		}
+
+		// pause cluster on bootstrap
+		cluster.PauseReconcile()
+		if err := fromClient.Update(ctx, cluster); err != nil {
+			return errors.Wrapf(err, "updating cluster on source")
+		}
+
+		if err := moveClusterResource(ctx, cluster, toClient); err != nil {
+			return err
+		}
+
+		if err := moveChildObjects(ctx, spec, fromClient, toClient); err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	return err
+}
+
+func moveClusterResource(ctx context.Context, cluster *v1alpha1.Cluster, client kubernetes.Client) error {
+	cluster.ResourceVersion = ""
+	cluster.UID = ""
+
+	// move eksa cluster
+	if err := client.Create(ctx, cluster); err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, "moving cluster %s", cluster.Name)
+	}
+
+	return nil
+}
+
+func moveChildObjects(ctx context.Context, spec *cluster.Spec, fromClient, toClient kubernetes.Client) error {
+	// read and move child objects
+	for _, child := range spec.ChildObjects() {
+		obj := &unstructured.Unstructured{}
+		obj.SetGroupVersionKind(child.GetObjectKind().GroupVersionKind())
+		if err := fromClient.Get(ctx, child.GetName(), child.GetNamespace(), obj); err != nil {
+			return errors.Wrapf(err, "reading child object %s %s", child.GetObjectKind().GroupVersionKind().Kind, child.GetName())
+		}
+
+		obj.SetResourceVersion("")
+		obj.SetUID("")
+		obj.SetOwnerReferences(nil)
+
+		if err := toClient.Create(ctx, obj); err != nil && !apierrors.IsAlreadyExists(err) {
+			return errors.Wrapf(err, "moving child object %s %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
+		}
+	}
+
+	return nil
+}
--- a/pkg/clustermanager/eksa_mover_test.go
+++ b/pkg/clustermanager/eksa_mover_test.go
@@ -1,0 +1,126 @@
+package clustermanager_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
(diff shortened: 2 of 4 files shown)


pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,108 @@
...
+func WithMoverRetryBackOff(backOff time.Duration) MoverOpt {
+	return func(m *Mover) {
+		m.retryBackOff = backOff
+	}
+}
+
+// Move applies the cluster's namespace and spec without checking for reconcile conditions.
+func (m Mover) Move(ctx context.Context, spec *cluster.Spec, targetCluster *types.Cluster) error {
+	var client kubernetes.Client
+	m.log.V(3).Info("Applying cluster spec")
```
mitalipaygude, line 79: Can we change this to something different from the applier log, like `moving the cluster object` or something?

mitalipaygude:
Can we read the eksa objects from the kind cluster, pause the cluster and then apply those objects to the "management" cluster?

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+
+		// read the cluster from bootstrap
+		cluster := &v1alpha1.Cluster{}
+		if err := fromClient.Get(ctx, spec.Cluster.Name, spec.Cluster.Namespace, cluster); err != nil {
+			return errors.Wrapf(err, "getting cluster from %s", fromCluster.Name)
+		}
+
+		// pause cluster on bootstrap
+		cluster.PauseReconcile()
+		if err := fromClient.ApplyServerSide(ctx, defaultFieldManager, cluster, kubernetes.ApplyServerSideOptions{ForceOwnership: true}); err != nil {
```
g-gaston, line 103: I would use `Update` here (assuming it's implemented by this client)
IMO this usecase is a classic read+update with optimistic locking (using resource version)

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+	// create namespace on mgmt
+	ns := &v1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cluster.Namespace,
+		},
+	}
```
g-gaston, line 131: IMO this should happen somewhere else
IMO a mover has no business creating a ns
I think a higher level orchestrator, like one of the tasks in the workflow, would be a better place

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if err := client.Create(ctx, ns); err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, "creating cluster namespace")
+	}
+
+	cluster.ResourceVersion = ""
+	cluster.UID = ""
+
+	// move eksa cluster
+	if err := client.ApplyServerSide(ctx, defaultFieldManager, cluster, kubernetes.ApplyServerSideOptions{ForceOwnership: true}); err != nil {
```
g-gaston, line 141: can we use `Create` here? it ensures we don't mistakenly update the cluster object after creation in case of a retry

and just check for `apierrors.IsAlreadyExists` if you want to make it idempotent

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+		return errors.Wrapf(err, "moving cluster %s", cluster.Name)
+	}
+
+	return nil
+}
+
+func moveChildObjects(ctx context.Context, spec *cluster.Spec, fromClient, toClient kubernetes.Client) error {
+	// read and move child objects
+	for _, child := range spec.ChildObjects() {
+		if err := fromClient.Get(ctx, child.GetName(), child.GetNamespace(), child); err != nil {
```
g-gaston, line 151: i don't love that we are updating the content of the cluster spec here
can we just use a fresh new struct?

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+			return errors.Wrapf(err, "reading child object %s %s", child.GetObjectKind().GroupVersionKind().Kind, child.GetName())
+		}
+
+		child.SetResourceVersion("")
+		child.SetUID("")
+		child.SetOwnerReferences(nil)
+
+		if err := toClient.ApplyServerSide(ctx, defaultFieldManager, child, kubernetes.ApplyServerSideOptions{ForceOwnership: true}); err != nil {
```
g-gaston, line 159: same comment about using `Create` here :)

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,165 @@
...
+
+// MoverOpt allows to customize a Mover on construction.
+type MoverOpt func(*Mover)
+
+// Mover applies the cluster spec to the management cluster and waits
+// until the changes are fully reconciled.
+type Mover struct {
+	log                                        logr.Logger
+	clientFactory                              ClientFactory
+	applyClusterTimeout, waitForFailureMessage time.Duration
```
g-gaston, line 29: do we need `waitForFailureMessage`?

also can we rename `applyClusterTimeout` to something more appropiate for this usecase? these names look like they make more sense in the applier

pkg/clustermanager/eksa_mover_test.go
```diff
@@ -0,0 +1,126 @@
...
+	m := clustermanager.NewMover(tt.log, tt.clientFactory,
+		clustermanager.WithMoverRetryBackOff(time.Millisecond),
+		clustermanager.WithMoverApplyClusterTimeout(time.Millisecond),
+	)
+	err := m.Move(tt.ctx, tt.spec, tt.bootstrap, tt.mgmtCluster)
+
+	tt.Expect(err).To(gomega.HaveOccurred())
+}
+
+func TestMoverGetChildren(t *testing.T) {
```
g-gaston, line 115: Is this supposed to be a fail or success test?

pkg/clustermanager/eksa_mover_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestMoverSuccess(t *testing.T) {
+	tt := newMoverTest(t)
+	objs := tt.spec.ClusterAndChildren()
+	tt.buildClient(nil, nil, objs...)
+	m := clustermanager.NewMover(tt.log, tt.clientFactory,
+		clustermanager.WithMoverRetryBackOff(time.Millisecond),
+		clustermanager.WithMoverNoTimeouts(),
+	)
+
+	tt.Expect(m.Move(tt.ctx, tt.spec, tt.bootstrap, tt.mgmtCluster)).To(gomega.Succeed())
```
g-gaston, line 76: You are missing assertions that check that the objects have actually been created

pkg/dependencies/factory.go
```diff
@@ -1216,6 +1217,26 @@ func (f *Factory) WithClusterDeleter() *Factory {
...
 }
 
+// WithClusterMover builds a cluster mover.
+func (f *Factory) WithClusterMover() *Factory {
+	f.WithLogger().WithUnAuthKubeClient().WithLogger()
+
+	f.buildSteps = append(f.buildSteps, func(_ context.Context) error {
+		var opts []clustermanager.MoverOpt
+		if f.config.noTimeouts {
+			opts = append(opts, clustermanager.WithMoverApplyClusterTimeout(time.Hour))
```
g-gaston, line 1227: shouldn't this be
```suggestion
			opts = append(opts, clustermanager.WithMoverNoTimeouts())
```

pkg/clustermanager/eksa_mover.go
mitalipaygude: Are you saying this because the spec is a pointer? if we create the new child object in the loop, won't it create a new memory for it?

g-gaston: I'm suggesting allocating new memory, yeah
Is that what you are asking? Maybe I didn't understand the question

mitalipaygude: I'm saying the code would actually create a new memory, because that's how the range function works afaik.

g-gaston: I see. Have you tested it? :)

mitalipaygude: yes I tested it on a go playground. I also have seen this in some previous code I was writing.

g-gaston: Oh cool! do you have a link to the playground by any chance?

mitalipaygude: i actually tried it last week, don't have the link to the playground unfortunately

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,146 @@
...
+// to update the cluster spec objects and check the status of the Cluster.
+// Generally only used in tests.
+func WithMoverRetryBackOff(backOff time.Duration) MoverOpt {
+	return func(m *Mover) {
+		m.retryBackOff = backOff
+	}
+}
+
+// Move applies the cluster's namespace and spec without checking for reconcile conditions.
+func (m *Mover) Move(ctx context.Context, spec *cluster.Spec, fromCluster, toCluster *types.Cluster) error {
```
g-gaston, line 74: what would you think about passing two clients instead of the cluster type?

I know this is an existing pattern but hear me out:
The only reason we pass these `type.Cluster` (afaik) is to use them to drive the configuration of a k8s client. This means that on every piece of code where we receive this and need a client, we need to inject a client factory and build the client in runntime. This leads to the same code being written again and again and having to test it again and again. I'm not concerned about the runtime performance implications but about the development overhead this adds.

Wouldn't this code (and all other similar service objects) be simpler if we just passed clients instead? And it would be simpler to test.

Opinions?

pkg/clustermanager/eksa_mover.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// move eksa cluster
+	if err := client.Create(ctx, cluster); err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, "moving cluster %s", cluster.Name)
+	}
+
+	return nil
+}
+
+func moveChildObjects(ctx context.Context, spec cluster.Spec, fromClient, toClient kubernetes.Client) error {
```
g-gaston, line 129: why pass the `cluster.Spec` instead of a pointer? Curious what we are gaining there

pkg/clustermanager/eksa_mover.go
tatlat: Yeah I agree, we should do this.

pkg/clustermanager/eksa_mover.go
tatlat: We should pass the pointer. I removed the pointer when testing if range allocated new memory or not, but now it's back.

pkg/clustermanager/eksa_mover_test.go
```diff
@@ -50,81 +51,76 @@ func newMoverTest(tb testing.TB) *moverTest {
...
+	tt.Expect(m.Move(tt.ctx, tt.spec, tt.fromClient, tt.toClient)).To(gomega.Succeed())
 
 	for _, obj := range objs {
-		tt.Expect(tt.toClient.Get(tt.ctx, obj.GetName(), obj.GetNamespace(), obj)).To(gomega.Succeed())
+		u := &unstructured.Unstructured{}
+		u.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+		tt.Expect(tt.toClient.Get(tt.ctx, obj.GetName(), obj.GetNamespace(), u)).To(gomega.Succeed())
+		original, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+		tt.Expect(err).To(gomega.Succeed())
+		tt.Expect(u.Object["spec"]).To(gomega.Equal(original["spec"]))
```
g-gaston, line 76: nit: this will give you a better diff error if the test fails
```suggestion
		tt.Expect(u.Object["spec"]).To(gomega.BeComparable(original["spec"]))
```

PR #7797, opened 2024-03-06
Bump helm to v3.14.2

g-gaston:
## Description of changes

This started as just a helm update.

However, a few commands in the release/cli module seem to not be working, they appeared to not be handling correctly the fact that the module is `cli` and not in `release`.

I fixed those. However, the unit tests for the release cli were then failing. This wasn't because of the helm or any other module update. The tests were actually broken.

I suspect because of the incorrect command, the tests haven't been running for a while, hence we never detected the issue.

The fix was easy, just a matter of changing the expected results. I believe when we changed the versioning logic for dev releases, these test broke. But since they didn't run, we didn't fix them.


release/Makefile
```diff
@@ -106,7 +106,7 @@ test: manifests generate fmt vet ## Run tests.
 
 ##@ Build
 
-build: clean fmt vet ## Build release binary.
+build: clean fmt ## Build release binary.
```
g-gaston, line 109: I don't think this was ever vetting correctly
Now that I fixed the vet command, it spits a bunch of errors
I'm leaving that to be fixed later in the future. I suspect those errors have been there for a while

PR #7798, opened 2024-03-06
Fix release go.mod with go mod tidy

g-gaston:


g-gaston:
I think this got to a point where i need to update helm in order to make all these deps play well together, so closing this in favor of #7797 
/close

PR #7815, opened 2024-03-08
Delete old command workflows and dead code

g-gaston:
*Description of changes:*
There is probably more if we dig deeper, but I think this is good enough for a first pass.


pkg/clustermanager/shim.go
```diff
@@ -16,32 +13,22 @@ import (
...
 func (s CreateClusterShim) CreateAsync(ctx context.Context, management *types.Cluster) error {
-	if err := s.manager.applyProviderManifests(ctx, s.spec, management, s.provider); err != nil {
-		return fmt.Errorf("installing cluster creation manifests: %v", err)
-	}
-
-	if err := s.manager.InstallMachineHealthChecks(ctx, s.spec, management); err != nil {
-		return fmt.Errorf("installing machine health checks: %v", err)
-	}
-
+	// TODO: implement reusing the apply logic from clustermanager.Applier
```
cxbrowne1207, line 31: Do we still need these? This seems to be used in the "new workflows" that seemed to have never made it out of the feature flag

Issue #7623, opened 2024-02-16
Backslashes in Vsphere passwords are interpreted as escape characters in generated YAML


jaredcookaws:
**What happened**:

EKS Anywhere installation on vSphere failed with error:

`"error: error parsing STDIN: error converting YAML to JSON: yaml: line 11: found unexpected end of stream\n"}`

I found that in the generated YAML config, the vSphere password is being added as a value in between double quotes.  When the backslash is at the end, it escapes the last double quote.  I presume that when the backslash is not at the end, the password is just incorrect when converted to JSON.

**What you expected to happen**:

Expected the EKS Anywhere installer to begin creating cluster nodes in vSphere.

**How to reproduce it (as minimally and precisely as possible)**:

Use a vSphere password with a backslash, exported to the shell with:

`export EKSA_VSPHERE_PASSWOR='mypassword\'`

**Anything else we need to know?**:

**Environment**:
- EKS Anywhere Release: 0.16.5
- EKS Distro Release: 1.27

PR #7829, opened 2024-03-11
Encode username and password in Kubernetes secrets

abhay-krishna:
We need to encode the basic auth credentials to base64 when using them in templates to ensure that passwords containing special characters like escape characters and line separators are properly handled, and don't cause errors when unmarshaling or during `kubectl apply`.

Fixes: #7623

### DIFF
--- a/pkg/providers/docker/config/template-cp.yaml
+++ b/pkg/providers/docker/config/template-cp.yaml
@@ -318,8 +318,8 @@ metadata:
   namespace: {{.eksaSystemNamespace}}
   labels:
     clusterctl.cluster.x-k8s.io/move: "true"
-stringData:
-  username: "{{.registryUsername}}"
-  password: "{{.registryPassword}}"
+data:
+  username: {{.registryUsername | b64enc}}
+  password: {{.registryPassword | b64enc}}
 ---
 {{- end }}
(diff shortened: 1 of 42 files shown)


pkg/providers/vsphere/testdata/expected_cluster_api_server_cert_san_domain_name.yaml
```diff
@@ -418,9 +418,9 @@ stringData:
     metadata:
       name: cloud-provider-vsphere-credentials
       namespace: kube-system
-    stringData:
-      vsphere_server.password: ""
-      vsphere_server.username: ""
+    data:
+      vsphere_server.password:
```
abhinavmpandey08, line 422: Can we add values for these?

abhay-krishna: Good catch! Will push a fix.

pkg/providers/docker/config/template-cp.yaml
```diff
@@ -318,8 +318,8 @@ metadata:
   namespace: {{.eksaSystemNamespace}}
   labels:
     clusterctl.cluster.x-k8s.io/move: "true"
-stringData:
-  username: "{{.registryUsername}}"
-  password: "{{.registryPassword}}"
+data:
+  username: {{.registryUsername | b64enc}}
+  password: {{.registryPassword | b64enc}}
```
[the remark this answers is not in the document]
abhay-krishna: That encoded credentials are used by the controller to create secrets directly through client-go. In the template the plaintext creds are used with the StringData type

Issue #6468, opened 2023-08-11
Properly support updating vSphere credentials during upgrade


jiayiwang7:
Currently, if you export the new vsphere credentials envs `EKSA_VSPHERE_USERNAME` and `EKSA_VSPHERE_PASSWORD` and run `eksctl anywhere upgrade` command, it fails at the preflight validation:
```
the VSphere credentials derived from VSPHERE_PASSWORD and VSPHERE_USERNAME are immutable; please use the same credentials for the upgraded cluster
```

In order to conduct the full credentials upgrade, user needs to manually update the `vsphere_credentials` in the cluster, and potentially update the cluster spec in order to trigger a full EKS-A CAPI cluster upgrade in order to apply the new credentials to all related secrets (e.g. `cloud-provider-vsphere-credentials`, `CLUSTER_NAME-vsphere-credentials`).

We need to have a better user experience around the credential updates through both CLI and FLC/controller, which might require redesigning some of the vsphere credentials logic in the codebase. 

The goal is to improve the UX and be able to update those credentials in a simple and clear way. In order to achieve that, we need to review how EKS-A does the vSphere credential upgrade today, and re-design a better credential upgrade workflow with consideration of all the existing behavior across the board (management vs workload clusters, default credentials vs cloud provider credentials, upgrade through CLI vs FLC, etc.).

jplewes:
If anyone can provide the steps necessary to manually update the vsphere_credentials in a cluster that would be very helpful.  I have a need to change credentials and there is no information on how to do so.

ndeksa:
- [ ] Add exact steps to upgrade vSphere creds ( update public docs)
- [ ] Draft a design proposal to add a command/script
- [ ] Implement command/script  to upgrade vSphere creds

2ez4szliu:
Script for vSphere credential update has been added to doc https://github.com/aws/eks-anywhere/pull/9189

PR #7877, opened 2024-03-20
Add support for updating vSphere credentials

abhay-krishna:
Fixes https://github.com/aws/eks-anywhere/issues/6468

### DIFF
--- a/pkg/providers/vsphere/vsphere.go
+++ b/pkg/providers/vsphere/vsphere.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/base64"
 	"fmt"
 	"os"
 	"reflect"
@@ -13,8 +14,10 @@ import (
 	"github.com/Masterminds/sprig"
 	etcdv1 "github.com/aws/etcdadm-controller/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	"sigs.k8s.io/yaml"
 
 	"github.com/aws/eks-anywhere/pkg/api/v1alpha1"
 	"github.com/aws/eks-anywhere/pkg/bootstrapper"
@@ -34,7 +37,6 @@ import (
 )
 
 const (
-	CredentialsObjectName    = "vsphere-credentials"
 	eksaLicense              = "EKSA_LICENSE"
 	vSphereUsernameKey       = "VSPHERE_USERNAME"
 	vSpherePasswordKey       = "VSPHERE_PASSWORD"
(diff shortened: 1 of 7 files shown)


pkg/providers/vsphere/vsphere.go
```diff
@@ -946,16 +956,16 @@ func (p *vsphereProvider) GenerateCAPISpecForCreate(ctx context.Context, _ *type
 	return controlPlaneSpec, workersSpec, nil
 }
 
-func (p *vsphereProvider) createSecret(ctx context.Context, cluster *types.Cluster, contents *bytes.Buffer) error {
+func (p *vsphereProvider) updateAllVsphereSecrets(ctx context.Context, cluster *types.Cluster, clusterSpec *cluster.Spec, contents *bytes.Buffer) error {
```
g-gaston, line 959: this method uses 3 different patterns for k8s resource generation: go templates, go string formatting and go struct yaml marshaling.

Can we just pick one and use it for all secrets? I'm in favor of only using go struct marshaling.

pkg/providers/vsphere/vsphere.go
```diff
@@ -967,11 +977,69 @@ func (p *vsphereProvider) createSecret(ctx context.Context, cluster *types.Clust
...
+		if err != nil {
+			return fmt.Errorf("marshalling secret object: %v", err)
+		}
+
+		err = p.providerKubectlClient.ApplyKubeSpecFromBytes(ctx, cluster, objBytes)
+		if err != nil {
+			return fmt.Errorf("applying secret object: %v", err)
+		}
+	}
+
```
g-gaston, line 1037: Tests for this new code?

g-gaston (commented):
Looking at the issue, it's mentioned that we want a better user experience and that how we handle credentials in vsphere might need a "redesign". I understand this PR doesn't do that and just adds code to automatically update secrets when the user passes new values to the CLI. Is there a long term vision for this feature?

Regarding the change itself, I think there might be some gotchas/nuances to update secrets that are shared across clusters. Could we add in the PR description more information about how this is supposed to behave for different scenarios (management vs workload clusters)? Another way to do this would be to just write down the expected steps a customer needs to take to update the credentials across all clusters.

Btw I'm not saying the change is a bad one, just making sure we build enough context around it. Right now it requires a lot of previous knowledge and inference from the code to understand the implications.

abhay-krishna:
> Looking at the issue, it's mentioned that we want a better user experience and that how we handle credentials in vsphere might need a "redesign". I understand this PR doesn't do that and just adds code to automatically update secrets when the user passes new values to the CLI. Is there a long term vision for this feature?
> 
> Regarding the change itself, I think there might be some gotchas/nuances to update secrets that are shared across clusters. Could we add in the PR description more information about how this is supposed to behave for different scenarios (management vs workload clusters)? Another way to do this would be to just write down the expected steps a customer needs to take to update the credentials across all clusters.
> 
> Btw I'm not saying the change is a bad one, just making sure we build enough context around it. Right now it requires a lot of previous knowledge and inference from the code to understand the implications.

In the long term, I had envisioned adding a command to the CLI that could update the vSphere credentials. We already have the `eksctl anywhere exp vsphere` subcommand for experimental features such as setting up vSphere users. I was thinking about adding a utility command such as `update-password` to the CLI, which customers can leverage to update their vSphere credentials.

I will update the description with the user experience and add documentation in the vSphere docs for this feature.

PR #7881, opened 2024-03-21
Upgrade registry certificate in packages

tatlat:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/7114

*Description of changes:*

*Testing (if applicable):*


pkg/dependencies/factory.go
```diff
@@ -1317,6 +1317,29 @@ func (f *Factory) WithPackageInstaller(spec *cluster.Spec, packagesLocation, kub
 	return f
 }
 
+// WithPackageInstallerWithoutWait builds a package installer that doesn't wait for active bundles.
+func (f *Factory) WithPackageInstallerWithoutWait(spec *cluster.Spec, packagesLocation, kubeConfig string) *Factory {
+	f.WithKubectl().WithPackageControllerClientWithoutWait(spec, kubeConfig).WithPackageClient()
+	f.buildSteps = append(f.buildSteps, func(_ context.Context) error {
```
g-gaston, line 1323: Can we avoid having this repeated code? can we just make use of the original builder method from here? like defining it as a dependency

pkg/dependencies/factory.go
```diff
@@ -1371,6 +1394,62 @@ func (f *Factory) WithPackageControllerClient(spec *cluster.Spec, kubeConfig str
 	return f
 }
 
+// WithPackageControllerClientWithoutWait builds a package controller client that doesn't wait for active bundles.
+func (f *Factory) WithPackageControllerClientWithoutWait(spec *cluster.Spec, kubeConfig string) *Factory {
```
g-gaston, line 1398: Can we just add a variadic argument to the original constructor method?

Maybe that won't work, but we should avoid having all this logic copy/paste if we can

pkg/workflows/management/post_cluster_upgrade.go
```diff
@@ -28,6 +28,10 @@ func (s *postClusterUpgrade) Run(ctx context.Context, commandContext *task.Comma
 		logger.Info(fmt.Sprintf("management cluster CAPI backup file not found: %v", err))
 	}
 
+	if commandContext.CurrentClusterSpec.Cluster.Spec.RegistryMirrorConfiguration != commandContext.ClusterSpec.Cluster.Spec.RegistryMirrorConfiguration {
```
g-gaston, line 31: I'm not sure if this will work very well since this field is a pointer

Regardless, there is an `Equal` method, I would vote to use that

pkg/workflows/management/post_cluster_upgrade.go
```diff
@@ -28,6 +28,10 @@ func (s *postClusterUpgrade) Run(ctx context.Context, commandContext *task.Comma
...
+		return &upgradeCuratedPackagesTask{}
```
g-gaston, line 32: I would prefer if the sequence of tasks wasn't decided during runtime
Could we just always return the upgrade task and then inside thta task skip the logic if the registry mirrors are the same?

pkg/workflows/management/upgrade_curated_packages.go
```diff
@@ -0,0 +1,64 @@
...
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	packagesManager.Spec.Replicas = replicas
+
+	packagesManager.SetUID("")
+	packagesManager.SetResourceVersion("")
+
+	err = client.Update(ctx, packagesManager)
```
g-gaston, line 45: Since we only want to update once field, could we just use a `Patch`?

Doing an update without a resource version can lead to race condition where two clients ovewrite each other changes. It's not a big deal since this is not happening in a loop, but I feel a patch would be better.

pkg/dependencies/factory.go
```diff
@@ -1317,10 +1317,18 @@ func (f *Factory) WithPackageInstaller(spec *cluster.Spec, packagesLocation, kub
 	return f
 }
 
-func (f *Factory) WithPackageControllerClient(spec *cluster.Spec, kubeConfig string) *Factory {
+// WithPackageInstallerWithoutWait builds a package installer that doesn't wait for active bundles.
+func (f *Factory) WithPackageInstallerWithoutWait(spec *cluster.Spec, packagesLocation, kubeConfig string) *Factory {
+	f.WithPackageInstaller(spec, packagesLocation, kubeConfig)
+	f.WithPackageControllerClient(spec, kubeConfig, curatedpackages.WithSkipWait())
```
g-gaston, line 1323: Shouldn't this be in the opposite order? You want your special package controller client to be built before the default one

```suggestion
	f.WithPackageControllerClient(spec, kubeConfig, curatedpackages.WithSkipWait()).
	    WithPackageInstaller(spec, packagesLocation, kubeConfig)
	
```

pkg/workflows/management/upgrade_curated_packages.go
```diff
@@ -0,0 +1,69 @@
...
+
+	err = client.Get(ctx, "eks-anywhere-packages", constants.EksaPackagesName, packagesManager)
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	replicas := packagesManager.Spec.Replicas
+	packagesManager.Spec.Replicas = ptr.Int32(0)
+	err = client.ApplyServerSide(ctx, constants.DefaultCLIFieldManager, packagesManager, kubernetes.ApplyServerSideOptions{ForceOwnership: true})
```
g-gaston, line 39: Can't we use a patch instead of server side? Server side apply can still overwrite changes if the resource is written in between our read and apply

Patch guarantees we only update the field we want to update (replicas)

pkg/workflows/management/upgrade_curated_packages.go
```diff
@@ -0,0 +1,69 @@
...
+	commandContext.PackageInstaller.InstallCuratedPackages(ctx)
+
+	client, err := commandContext.ClientFactory.BuildClientFromKubeconfig(commandContext.ManagementCluster.KubeconfigFile)
+	if err != nil {
+		commandContext.SetError(err)
+	}
+
+	packagesManager := &appsv1.Deployment{}
+
+	err = client.Get(ctx, "eks-anywhere-packages", constants.EksaPackagesName, packagesManager)
```
g-gaston, line 31: What if `InstallCuratedPackages` fails?
What if the package controller is disabled and not installed?

I don't think we can safely execute this on every case.

pkg/workflows/management/upgrade_curated_packages.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		commandContext.SetError(err)
+		return &workflows.CollectMgmtClusterDiagnosticsTask{}
+	}
+
+	return nil
+}
+
+func (s *upgradeCuratedPackagesTask) Name() string {
+	return "install-curated-packages"
```
g-gaston, line 60: update?

pkg/curatedpackages/packageinstaller.go
```diff
@@ -67,6 +67,26 @@ func (pi *Installer) InstallCuratedPackages(ctx context.Context) {
 	}
 }
 
+// UpgradeCuratedPackages upgrades curated packages as part of the cluster upgrade.
+func (pi *Installer) UpgradeCuratedPackages(ctx context.Context) {
+	if IsPackageControllerDisabled(pi.spec.Cluster) {
+		logger.Info("  Package controller disabled")
```
g-gaston, line 73: why the space?
```suggestion
		logger.Info("Package controller disabled")
```

pkg/curatedpackages/packageinstaller.go
```diff
@@ -67,6 +67,26 @@ func (pi *Installer) InstallCuratedPackages(ctx context.Context) {
...
+		return
+	}
+	PrintLicense()
+	err := pi.installPackagesController(ctx)
+
+	if err != nil {
```
g-gaston, line 79: nit
```suggestion
	if 	err := pi.installPackagesController(ctx); err != nil {
```

pkg/dependencies/factory.go
```diff
@@ -1347,13 +1355,8 @@ func (f *Factory) WithPackageControllerClient(spec *cluster.Spec, kubeConfig str
...
 		}
-		f.dependencies.PackageControllerClient = curatedpackages.NewPackageControllerClient(
-			f.dependencies.Helm,
-			f.dependencies.Kubectl,
-			spec.Cluster.Name,
-			mgmtKubeConfig,
-			&bundle.PackageController.HelmChart,
-			f.registryMirror,
+
+		opts = append(opts,
```
g-gaston, line 1359: nit: I would expect to be able to override these defaults with the new variadic args, but this makes it impossible

pkg/workflows/interfaces/interfaces.go
```diff
@@ -76,6 +76,7 @@ type EksdUpgrader interface {
 
 type PackageInstaller interface {
 	InstallCuratedPackages(ctx context.Context)
+	UpgradeCuratedPackages(ctx context.Context)
```
g-gaston, line 79: could we have a different interface? or change the name of the interface
I prefer a different interface, but up to you

pkg/dependencies/factory.go
tatlat: fixed

pkg/workflows/interfaces/interfaces.go
tatlat: i changed the name :)

pkg/dependencies/factory.go
```diff
@@ -1324,14 +1324,14 @@ func (f *Factory) WithGitOpsFlux(clusterConfig *v1alpha1.Cluster, fluxConfig *v1
 
 func (f *Factory) WithPackageInstaller(spec *cluster.Spec, packagesLocation, kubeConfig string) *Factory {
```
vivek-koppuru, line 1325: Might be worth changing the method names as well if you are changing the interface

vivek-koppuru (approved):

vivek-koppuru (approved):

PR #7884, opened 2024-03-22
Updated information about static IP requirement

jaradtke-aws:
*Issue #, if available:*
The existing verbiage was a bit confusing - and depending on interpretation it was not accurate.
There is only 1 "static IP" needed per EKS Anywhere cluster, and that is for the management endpoint (API)

*Description of changes:*
I removed the confusing phrasing and added an update after reviewing with peers.

*Testing (if applicable):*
N/A - just a doc update


rahulbabu95:
@jaradtke-aws thanks for your change and apologies for getting to it only now!

PR #7888, opened 2024-03-22
Update CLI generate clusterconfig description for provider flag

sp1999:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/cmd/eksctl-anywhere/cmd/generateclusterconfig.go
+++ b/cmd/eksctl-anywhere/cmd/generateclusterconfig.go
@@ -49,7 +49,7 @@ func preRunGenerateClusterConfig(cmd *cobra.Command, args []string) {
 
 func init() {
 	generateCmd.AddCommand(generateClusterConfigCmd)
-	generateClusterConfigCmd.Flags().StringP("provider", "p", "", "Provider to use (vsphere or tinkerbell or docker)")
+	generateClusterConfigCmd.Flags().StringP("provider", "p", "", fmt.Sprintf("Provider to use (%s)", strings.Join(constants.SupportedProviders, " or ")))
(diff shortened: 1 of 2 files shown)


cmd/eksctl-anywhere/cmd/generateclusterconfig.go
```diff
@@ -49,7 +49,7 @@ func preRunGenerateClusterConfig(cmd *cobra.Command, args []string) {
 
 func init() {
 	generateCmd.AddCommand(generateClusterConfigCmd)
-	generateClusterConfigCmd.Flags().StringP("provider", "p", "", "Provider to use (vsphere or tinkerbell or docker)")
+	generateClusterConfigCmd.Flags().StringP("provider", "p", "", "Provider to use (vsphere or tinkerbell or docker or cloudstack or nutanix or snow)")
```
abhay-krishna, line 52: This should be dynamic based on the set of providers we support. 
```suggestion
	generateClusterConfigCmd.Flags().StringP("provider", "p", "", "Provider to use (strings.Join(constants.SupportedProviders, " or "))")
```
We don't have the supported providers list, but we can just add it in the constants file
```
var SupportedProviders = []string{"cloudstack", "docker", "nutanix", "snow", "tinkerbell", "vsphere"}
```

sp1999: Updated

abhay-krishna: Also it's probably best to use the constant names for the providers array, i.e., `CloudStackProviderName` instead of `"cloudstack"`. My example above was just a reference.

PR #7913, opened 2024-03-29
Fix proxy configuration for airgapped environments

ahreehong:
*Description of changes:*
The CNI reconciler fails to generate cilium manifests in an airgapped environment because it tries to fetch the image from `public.ecr.aws` instead of through the proxy. This is due to the proxy configuration not being passed when constructing an instance of the `HelmClient` in the controller.

This PR adds the proxyConfiguration as an option when building the `HelmClient`

*Testing (if applicable):*
- Tested manually by creating a vsphere cluster in a simulated airgapped environment and creating a cluster using proxy configuration
- Created new e2e tests for vsphere and cloudstack that tests a create flow using proxy in an airgapped environment.


test/e2e/airgap.go
```diff
@@ -39,6 +39,21 @@ func runAirgapConfigFlow(test *framework.ClusterE2ETest, localCIDRs string) {
 	test.DeleteCluster()
 }
 
+// runAirgapConfigProxyFlow run airgap deployment but allow bootstrap cluster to access local peers.
```
jiayiwang7, line 42:
```suggestion
// runAirgapConfigProxyFlow runs air-gap deployment workflow with allowing bootstrap cluster to access local peers.
```

jiayiwang7: lets update the function comment for both  `runAirgapConfigProxyFlow` and `runAirgapConfigFlow`, clarifying what the flow does in details for each case.

jiayiwang7 (approved):

PR #7927, opened 2024-04-02
Node upgrader changes to consume upgrader go binary instead of script

rahulbabu95:
*Issue #, if available:*
Consume Upgrader go binary for InPlace upgrades to upgrade nodes instead of using the upgrader script.

*Description of changes:*

*Testing (if applicable):*
Manually tested K8s version upgrades from 1.25 all the way until 1.29 with the binary embedded in the upgrader image instead of invoking the upgrader script. 

### DIFF
--- a/pkg/nodeupgrader/upgrader.go
+++ b/pkg/nodeupgrader/upgrader.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	upgradeScript = "/foo/eksa-upgrades/scripts/upgrade.sh"
+	upgradeBin = "/foo/eksa-upgrades/tools/upgrader"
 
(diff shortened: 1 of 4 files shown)


pkg/nodeupgrader/upgrader.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	upgradeScript = "/foo/eksa-upgrades/scripts/upgrade.sh"
+	upgradeBin = "/foo/eksa-upgrades/tools/upgrader"
```
vivek-koppuru, line 14: We never changed it from foo? Are we able to put this in another location?

vivek-koppuru (approved):

PR #7965, opened 2024-04-11
Improve cluster autoscaler e2e test

abhinavmpandey08:
*Description of changes:*
Currently, cluster autoscaler test creates a busybox deployment with [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) to simulate load to trigger autoscaling.
The problem with this approach is that we expect HPA to scale up the pods to a point where they are un-schedulable which should cause autoscaler to scale up the worker nodes, but there's no guarantee that it'll do that. Plus, HPA can take time to scale the pods as well.
So instead, we replace the uncertainty with HPA and just try to schedule 111 pods which is over the limit of number of pods supported per Kubernetes node. This will ensure that there's atleast 1 pod that's un-schedulable and cause the autoscaler to scale up the MD.

There is a bug in cluster autoscaler currently where it's not able to autoscale the cluster initially because of missing permissions on infrastructure machine template. Cluster Autoscaler does restart after ~10 min after which it starts functioning normally.
Instead of waiting for ~10 min, we'll also force triggering a restart so the e2e doesn't have to wait.
This can be removed once the following issue is resolve upstream https://github.com/kubernetes/autoscaler/issues/6490


pokearu (approved):
/lgtm
/approve

Issue #7875, opened 2024-03-20
Introduce Metrics Serving in EKS Anywhere


jiayiwang7:
**What would you like to be added**:

Introduce options to secure serving metrics on K8s system and EKS-A management components.

**Why is this needed**:

As an EKS Anywhere cluster administrator, I would like to scrape metrics from the K8s system and EKS-A management components in a simple but secure way. Those metrics are useful for building dashboard and alerts, monitoring the healthy state of a cluster.

Currently in EKS-A, metrics of some system components are already exposed by default (e.g. `coredns`, `kube-api-server`). Other system and management components such as `kube-controller-manager` are configured with the default `--bind-address=127.0.0.1` or equivalent, so that these servers are only listening on `localhost`. The goal is to expose those metrics in a secure fashion so that external monitoring services such as Prometheus can consume them properly.

**Details**

There are three types of system/management components we would like to serve metrics from:
1. K8s system components, such as `kube-controller-manager`, `kube-scheduler`, `kube-proxy`.
1. EKS-A management components, such as `eksa-cluster-controller`, `eks-anywhere-packages`.
1. CAPI components, such as `capi-controller`, `capi-kubeadm-control-plane`, `capv-controller` (provider specific), `etcdadm-controller`, `etcdadm-bootstrap-provider`

In the list above, scraping metrics on the secure port of the K8s system components are already introduced as default in Kubernetes with native K8s authentication and authorization workflow: https://github.com/kubernetes/kubernetes/pull/72491. So the controller-manager / scheduler secure metrics should already be enabled by default with `--authentication-kubeconfig` and `authorization-kubeconfig` flags. Regarding how they can emit metrics with RBAC, we need more investigation (whether the above core components can all be exposed from the /metrics endpoint via authentication (user/group/SA) and authorization (via RBAC verb: get, nonResourceURLs: /metrics)).

[...]

Issue #4405, opened 2022-12-14
Expose kube-scheduler and kube-controller-manager metrics


Cajga:
**What would you like to be added**:
We would like to be able to scrape the kube-scheduler and the kube-controller-manager metrics which are important to understand the state of the cluster. Both of these services are exporting Prometheus metrics but by default, we cannot scrape them in EKS Anywhere as these services are only listening on localhost (at least on Bare Metal installations... I do not have EKS-A on other platform to verify this).

A simple way would be to modify the `/etc/kubernetes/manifests/kube-scheduler.yaml` and the `/etc/kubernetes/manifests/kube-controller-manager.yaml` files to have `--bind-address=0.0.0.0` in them (manually modifying these files on the masters is my workaround at the moment).

**Why is this needed**:
As mentioned above, these are important metrics to understand the state of the cluster and set up proper alerting/visualisation in a production environment.

g-gaston:
Thanks for the suggestion @Cajga, we will look into it!

Issue #4299, opened 2022-12-06
Expose kube-proxy prometheus metrics


Cajga:
**What would you like to be added**:
Please expose kube-proxy metrics so, they can be scraped with (ideally, also from an external) prometheus. Currently, in EKS-A, kube-proxy metrics are only available from the node itself:
```
# netstat -natupl|grep 10249
tcp        0      0 127.0.0.1:10249         0.0.0.0:*               LISTEN      1428/kube-proxy
# curl -L -s http://127.0.0.1:10249/metrics |grep -v ^#|head -n 5
apiserver_audit_event_total 0
apiserver_audit_requests_rejected_total 0
go_gc_cycles_automatic_gc_cycles_total 1008
go_gc_cycles_forced_gc_cycles_total 0
go_gc_cycles_total_gc_cycles_total 1008
```

One simple solution would be to modify the default configmap of kube-proxy and add `0.0.0.0:10249` for `metricsBindAddress`:
```
$ kubectl describe configmaps -n kube-system kube-proxy |grep -i address
bindAddress: 0.0.0.0
bindAddressHardFail: false
healthzBindAddress: ""
metricsBindAddress: ""
nodePortAddresses: null
```

According to the [official documentation](https://kubernetes.io/docs/reference/config-api/kube-proxy-config.v1alpha1/) of kube-proxy config, without a value, it defaults to `127.0.0.1:10249`

**Why is this needed**:
To be able to scrape kube-proxy metrics with prometheus which are helpful to understand the state of the cluster (and if kube-proxy is down :) ).

Cajga:
@panktishah26, thanks ahead.

Just wanted to highlight that `metricsBindAddress: 0.0.0.0:10249` is the default in EKS clusters. See [ticket](https://github.com/aws/containers-roadmap/issues/657) about the same discussion there.

ndeksa:
Closing as documentation has been updated to include detailed steps on how to scrape kube-proxy and other components metrics beyond localhost. You can find the instructions here -
https://anywhere.eks.amazonaws.com/docs/clustermgmt/observability/expose-metrics/

PR #7970, opened 2024-04-11
Document steps to expose metrics for EKS-A components

sp1999:
*Issue #, if available:*
#7875 
#4299
#4405  

*Description of changes:*
- Added documentation on exposing metrics for EKS-A components by configuring a proxy as a Daemonset on the host network. New section is added in observability within cluster management in the EKS-A docs.
- Added design doc for exposing metrics of all EKS-components securely


docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+            app: metrics-proxy
+        spec:
+          tolerations:
+          - key: node-role.kubernetes.io/control-plane
+            operator: Exists
+            effect: NoSchedule
+          hostNetwork: true
+          containers:
+            - name: haproxy
+              image: haproxy:2.9
```
abhay-krishna, line 115: We should suggest the ECR public mirror as it's a Docker official image - https://gallery.ecr.aws/docker/library/haproxy

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+            app: metrics-proxy
+        spec:
+          tolerations:
+          - key: node-role.kubernetes.io/control-plane
+            operator: Exists
+            effect: NoSchedule
+          hostNetwork: true
+          containers:
+            - name: haproxy
+              image: public.ecr.aws/docker/library/haproxy:latest
```
jiayiwang7, line 115: lets use the eksa built image: `public.ecr.aws/eks-anywhere/kubernetes-sigs/kind/haproxy:v0.20.0-eks-a-54`

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "Expose metrics for EKS-A components"
+linkTitle: "Expose metrics"
+weight: 100
+date: 2024-04-06
+description: >
+  Expose metrics for EKS-A components
+---
+
+Some kubernetes system components like kube-controller-manager, kube-scheduler and kube-proxy expose metrics only on the localhost by default. In order to expose metrics for these components so that other monitoring systems like Prometheus can scrape them, you need to deploy some proxy as a Daemonset on the host network. The proxy pods also need to be configured with control plane tolerations so that they can be scheduled on the control plane nodes.
```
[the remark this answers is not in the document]
sp1999: For kube proxy, daemonset also needs to be deployed on worker nodes as kube-proxy is itself a daemonset deployed on all nodes

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+date: 2024-04-06
+description: >
+  Expose metrics for EKS-A components
+---
+
+Some Kubernetes system components like kube-controller-manager, kube-scheduler and kube-proxy expose metrics only on the localhost by default. In order to expose metrics for these components so that other monitoring systems like Prometheus can scrape them, you can deploy a proxy as a Daemonset on the host network of the nodes. The proxy pods also need to be configured with control plane tolerations so that they can be scheduled on the control plane nodes.
+
+To configure a proxy for exposing metrics on an EKS-A cluster, you can perform the following steps:
+
+1.  Create a new cluster role for the client to access the metrics endpoint of the components.
```
g-gaston, line 14: Why do we need this? I understand the user needs it, but why is it needed to deploy the proxy?

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+          http-request deny if !{ path /metrics }
+          default_backend kube-scheduler
+        backend kube-scheduler
+          server kube-scheduler 127.0.0.1:10259 ssl verify none check
+    EOF
+    ```
+
+4.  Create a daemonset for the proxy and mount the config map volume onto the proxy pods.
+    
+    Below is an example configuration for the HAProxy daemonset.
```
g-gaston, line 92: I would omit this sentence. Or just add all of this under a subsection (##HAProxy)

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+            - name: haproxy
+              image: public.ecr.aws/eks-anywhere/kubernetes-sigs/kind/haproxy:v0.20.0-eks-a-54
+              env:
+                - name: NODE_IP
+                  valueFrom:
+                    fieldRef:
+                      apiVersion: v1
+                      fieldPath: status.hostIP
+                - name: TOKEN
+                  value: "$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"
```
g-gaston, line 123: Could you help me understand a bit more how this works? Why do we need a service account token in the haproxy pod? What service account is this supposed to be?

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+                name: metrics-proxy
+              name: haproxy-config
+    EOF
+    ```
+
+5.  Verify that the metrics are exposed to the client pods by running the following commands:
+    ```bash
+    kubectl exec -it {client-pod-name} -n {client-pod-namespace} -- bash
+    export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+    curl -H "Authorization: Bearer ${TOKEN}" http://{node-IP}:{component-port}/metrics
```
g-gaston, line 145: nit: it would be better if you gave an example that worked out of the box, using an ephemeral container with `kubectl debug` or simply a `kubectl run`.

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+      - nonResourceURLs:
+          - "/metrics"
+        verbs:
+          - get
+    EOF
+    ```
+
+2.  Create a new cluster role binding to bind the above cluster role to the client pod's service account.
+    
+    Below is an example cluster role binding if you use prometheus as the client to scrape metrics.
```
g-gaston, line 31: same here. Do we need to include this in our doc? This should be part of the "scraper" docs, right?

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+* Snow provider (capas) and docker provider (capd)
+* ECR-credential-provider
+
+**Future Scope:**
+
+* Securely expose metrics for all other components (kube-vip, coredns, cilium, cert-manager)
+
+## Current State of EKS-A components
+
+![table](images/expose-metrics.png)
```
g-gaston, line 62: nit: can we make this a md table instead of an image? updating the image is a pain

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
sp1999: It's not required to deploy the proxy. I will move it to another section

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
sp1999: You are right. This token is not required for the proxy

designs/expose-metrics.md
sp1999: Agree with that but then in this case, the md table won't be properly formatted in the way it looks in the image. Considering that I would prefer keeping the image as we won't be updating the image frequently.

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+These components already perform bearer token authentication and RBAC authorization for client requests and they are already configured to allow listening on all interfaces and IP address families. No further action needs to be taken to expose metrics for these components securely.
+
+#### **Kube-Controller-Manager, Kube-Scheduler and Kube-Proxy:**
+
+Kube-controller-manager and kube-scheduler already perform bearer token authentication and RBAC authorization whereas kube-proxy does not but all three components listen only on the localhost (127.0.0.1) for client requests. We can document the steps to be followed to configure some proxy as a Daemonset on the cluster which forwards the client requests to the metrics endpoint for each component. The proxy pods must run in the `hostNetwork` so that they can access the loopback interfaces of the corresponding pods. There will be no changes done in EKS-A to configure these for the customers.
+
+**Documentation Steps:**
+
+1. Create a cluster role object which gives permissions to get the metrics
+
```
chrisnegus, line 84: Not necessary, but if you indent the yaml, you can number each step "1." instead of 2., 3., etc. It will cause them to number sequentially and save you the trouble of renumbering everything if you ever insert a new step in the middle.

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+#### **Kube-Apiserver and Kubelet:**
+
+These components already perform bearer token authentication and RBAC authorization for client requests and they are already configured to allow listening on all interfaces and IP address families. No further action needs to be taken to expose metrics for these components securely.
+
+#### **Kube-Controller-Manager, Kube-Scheduler and Kube-Proxy:**
+
+Kube-controller-manager and kube-scheduler already perform bearer token authentication and RBAC authorization whereas kube-proxy does not but all three components listen only on the localhost (127.0.0.1) for client requests. We can document the steps to be followed to configure some proxy as a Daemonset on the cluster which forwards the client requests to the metrics endpoint for each component. The proxy pods must run in the `hostNetwork` so that they can access the loopback interfaces of the corresponding pods. There will be no changes done in EKS-A to configure these for the customers.
```
chrisnegus, line 79:
```suggestion
Kube-controller-manager and kube-scheduler already perform bearer token authentication and RBAC authorization, whereas kube-proxy does not. But all three components listen only on the localhost (127.0.0.1) for client requests. We can document the steps to be followed to configure some proxy as a Daemonset on the cluster, which forwards the client requests to the metrics endpoint for each component. The proxy pods must run in the `hostNetwork` so that they can access the loopback interfaces of the corresponding pods. There will be no changes done in EKS Anywhere to configure these for the customers.
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+      name: prometheus-server
+      namespace: observability
+    roleRef:
+      kind: ClusterRole
+      name: metrics-reader
+      apiGroup: rbac.authorization.k8s.io
+    EOF
+    ```
+
+3.  Create a config map to store the proxy configuration.
```
chrisnegus, line 49:
```suggestion
3.  Create a ConfigMap to store the proxy configuration.
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+    
+    Below is an example config map if you use HAProxy as the proxy server.
```
chrisnegus, line 51:
```suggestion
    Below is an example ConfigMap if you use HAProxy as the proxy server.
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+        frontend kube-scheduler
+          bind ${NODE_IP}:10259
+          http-request deny if !{ path /metrics }
+          default_backend kube-scheduler
+        backend kube-scheduler
+          server kube-scheduler 127.0.0.1:10259 ssl verify none check
+    EOF
+    ```
+
+4.  Create a daemonset for the proxy and mount the config map volume onto the proxy pods.
```
chrisnegus, line 90:
```suggestion
4.  Create a Daemonset for the proxy and mount the ConfigMap volume onto the proxy pods.
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+Following components are not considered for exposing metrics securely:
+
+* Snow provider (capas) and docker provider (capd)
+* ECR-credential-provider
+
+**Future Scope:**
+
+* Securely expose metrics for all other components (kube-vip, coredns, cilium, cert-manager)
+
+## Current State of EKS-A components
```
chrisnegus, line 60:
```suggestion
## Current State of EKS Anywhere components
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+Another advantage is that it provides flexibility to the customers in choosing any proxy based on their preference. Some of the most popular proxies include [nginx](https://github.com/nginx-proxy/nginx-proxy), [envoy](https://github.com/envoyproxy/envoy), [haproxy](https://github.com/haproxy/haproxy), [traefik](https://github.com/traefik/traefik), etc. We will document an example config for HAProxy and customers can configure it similarly for other proxies.
+
+One drawback is that it provides a bad User Experience to the customers as they need to configure these additional objects for each of their clusters. But as the number of users which require this feature is not large enough to justify supporting it in EKS-A, this is a good workaround solution to document it to the users. Even kubeadm doesn't support it for the same reason. For more details, check out this [issue](https://github.com/kubernetes/kubeadm/issues/2388#issuecomment-776073834)
```
chrisnegus, line 217:
```suggestion
One drawback is that it provides a bad User Experience to the customers as they need to configure these additional objects for each of their clusters. But as the number of users which require this feature is not large enough to justify supporting it in EKS Anywhere, this is a good workaround solution to document it to the users. Even kubeadm doesn't support it for the same reason. For more details, check out this [issue](https://github.com/kubernetes/kubeadm/issues/2388#issuecomment-776073834)
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+A disadvantage is that prometheus associates metrics with pod names but since the pods are behind a proxy, the proxy's pod name will be used instead for the metrics. (see [issue](https://github.com/prometheus-operator/kube-prometheus/issues/718#issuecomment-776360908) for more details)
+
+#### **EKS-A Controller Manager and EKS-A Packages:**
```
chrisnegus, line 221:
```suggestion
#### **EKS Anywhere Controller Manager and EKS Anywhere Packages:**
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+#### **Etcdadm Bootstrap Provider Controller Manager and Etcdadm Controller Controller Manager:**
+
+These components also do not perform any kind of authentication or authorization for client requests and listen only on the localhost by default. We will implement the diagnostics feature to expose metrics securely on all interfaces.
+
+#### **Capi-Controller-Manager, capi-kubeadm-bootstrap-controller-manager and capi-kubeadm-control-plane-controller-manager:**
+
+These components already implement the diagnostics feature to expose metrics securely on all interfaces. No further action needs to be taken to expose metrics for these components securely.
+
+#### **EKS-A supported CAPI providers (capv, capx, capt, capc):**
```
chrisnegus, line 233:
```suggestion
#### **EKS Anywhere supported CAPI providers (capv, capx, capt, capc):**
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+For capv, the diagnostics feature has already been implemented in the [latest](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/releases/tag/v1.9.2) release but our `eks-anywhere-build-tooling` repo points to an [older release](https://github.com/aws/eks-anywhere-build-tooling/blob/main/projects/kubernetes-sigs/cluster-api-provider-vsphere/GIT_TAG) version which does not support the diagnostics feature and defaults to `127.0.0.1` for `--metrics-bind-addr` flag. We would just need to bump the capv version to the latest release version.
+
+For capx, the diagnostics feature has already been implemented and we also point to the latest release in the `eks-anywhere-build-tooling` repo. No further action needs to be taken to securely expose metrics for capx.
+
+For capt, it does not perform any kind of authentication or authorization for client requests and listens only on the localhost by default. We can implement the diagnostics feature to expose metrics securely on all interfaces.
+
+## Implementation Details
+
+**Diagnostics Feature for all EKS-A, CAPI etcd and provider-specific controllers:**
```
chrisnegus, line 245:
```suggestion
**Diagnostics Feature for all EKS Anywhere, CAPI etcd, and provider-specific controllers:**
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+Add the diagnostics options capi flags and rbac [markers](https://book.kubebuilder.io/reference/markers/rbac) to the controller binary in `main.go` 
+
+## Testing
+
+* E2E tests for the diagnostics feature
+* Unit tests for any additional utility functions implemented
+
+## Documentation
+
+* Add necessary steps on configuring the metrics proxy daemonset in the cluster to the eks-a docs
```
chrisnegus, line 303:
```suggestion
* Add necessary steps on configuring the metrics proxy Daemonset in the cluster to the EKS Anywhere docs
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+* We can reference the [CAPI](https://main.cluster-api.sigs.k8s.io/tasks/diagnostics#scraping-metrics) documentation on diagnostics feature for scraping metrics from the controllers
+
+## Alternate Solutions Considered
+
+### Using kube-rbac-proxy for all components
+
+Using kube-rbac-proxy for all EKS-A components to expose metrics securely by enforcing RBAC policies and authentication. In this approach, metrics requests are routed through the Kube-rbac-proxy, which sits between the client and the API server. Kube-rbac-proxy authenticates the client using various authentication mechanisms such as bearer token, client TLS certificates, request header authentication, etc. It then verifies the client's RBAC permissions and only allows access to the metrics endpoint if the client has the necessary privileges. For more details, check out option 2 [here](https://quip-amazon.com/II8XAy90Pq2v/Expose-metrics-of-EKS-A-components#temp:C:fRf4452d35522194e5bb535f4d14)
```
chrisnegus, line 310:
```suggestion
Using kube-rbac-proxy for all EKS Anywhere components to expose metrics securely by enforcing RBAC policies and authentication. In this approach, metrics requests are routed through the kube-rbac-proxy, which sits between the client and the API server. Kube-rbac-proxy authenticates the client using various authentication mechanisms such as bearer token, client TLS certificates, request header authentication, etc. It then verifies the client's RBAC permissions and only allows access to the metrics endpoint if the client has the necessary privileges. For more details, check out option 2 [here](https://quip-amazon.com/II8XAy90Pq2v/Expose-metrics-of-EKS-A-components#temp:C:fRf4452d35522194e5bb535f4d14)
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+This approach enables authentication and RBAC authorization for all the components but requires maintaining an additional upstream dependency. Some k8s system components already do have authn/authz enabled by default and capi also introduced the diagnostics feature recently which is the preferred way of enabling authn/authz for the controller-runtime based components. So using kube-rbac-proxy as an additional layer of security for these components is not necessary.
```
chrisnegus, line 312:
```suggestion
This approach enables authentication and RBAC authorization for all the components but requires maintaining an additional upstream dependency. Some Kubernetes system components already do have authn/authz enabled by default and capi also introduced the diagnostics feature recently which is the preferred way of enabling authn/authz for the controller-runtime based components. So using kube-rbac-proxy as an additional layer of security for these components is not necessary.
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+Another thing to note is that kube-rbac-proxy project is in alpha stage and may have significant changes in the future. So, it's better to not have a dependency on it in the future if there are better alternatives available.
+
+### Configurable bind-address for kube-scheduler and kube-controller-manager
+
+We can allow customers to configure the `--bind-address`  flag for these components through the cluster spec to allow listening on all interfaces and IP address families so that prometheus or any other component with appropriate RBAC permissions can scrape the metrics endpoint for these components securely.
+
+Currently, we cannot make the `--metrics-bind-address`  flag configurable for kube-proxy through the cluster spec as CAPI doesn't support configuring kube-proxy configuration. In the future, we can either patch CAPI to enable support for configuring kube-proxy or [disable](https://github.com/kubernetes-sigs/cluster-api/issues/4512#issuecomment-1267092583) installing kube-proxy from CAPI and install it in EKS-A instead.
```
chrisnegus, line 320:
```suggestion
Currently, we cannot make the `--metrics-bind-address`  flag configurable for kube-proxy through the cluster spec as CAPI doesn't support configuring kube-proxy configuration. In the future, we can either patch CAPI to enable support for configuring kube-proxy or [disable](https://github.com/kubernetes-sigs/cluster-api/issues/4512#issuecomment-1267092583) installing kube-proxy from CAPI and install it in EKS Anywhere instead.
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+    ...
+    controllerManagerExtraArgs:
+      bindAddress: "0.0.0.0"
+    schedulerExtraArgs:
+      bindAddress: "0.0.0.0"
+```
+
+The main concern with making bind-address configurable is that configuring it to bind on all interfaces might expose the metrics publicly over the internet on a node which has any interface exposed to the internet. In a cluster with single control plane node, binding it to the control plane node ip address would solve the issue but it wouldn't work for HA clusters with multiple control plane nodes which is usually the case in a production environment. Another solution would be to apply [firewall rules](https://github.com/kubernetes/kubeadm/issues/2244#issuecomment-763533964) on every node before binding to 0.0.0.0 but this is not a good idea either.
+
+Another thing to note is that it is tough to validate the range of IP addresses that bind address can be allowed to be configure with. Even kubernetes does not do any such validation for these components. Only validation that can be done is that the address is in a proper IPv4/IPv6 format. If a user configures some unreachable address, it would be hard to debug the issue with the component.
```
chrisnegus, line 341:
```suggestion
Another thing to note is that it is tough to validate the range of IP addresses that bind address can be allowed to be configure with. Even Kubernetes does not do any such validation for these components. Only validation that can be done is that the address is in a proper IPv4/IPv6 format. If a user configures some unreachable address, it would be hard to debug the issue with the component.
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+  "bind-address": "0.0.0.0"
+schedulerExtraArgs:
+  "bind-address": "0.0.0.0"
+```
+
+These flags will be fetched from the cluster spec and added to the [controllerManagerExtraArgs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#controllermanager-flags) and [schedulerExtraArgs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#scheduler-flags) map in the `ClusterConfiguration` object during create and upgrade operations for generating control plane CAPI spec.
+
+**Testing:**
+
+* E2E tests will be required to test whether the flags are actually configured for k8s system components
```
chrisnegus, line 356:
```suggestion
* E2E tests will be required to test whether the flags are actually configured for Kubernetes system components
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+**Validations:**
+
+* Validate that the `bind-address` flag is in a proper format similar to how k8s does it [here](https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/pkg/proxy/apis/config/validation/validation.go#L274) for kube-proxy
```
chrisnegus, line 360:
```suggestion
* Validate that the `bind-address` flag is in a proper format similar to how Kubernetes does it [here](https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/pkg/proxy/apis/config/validation/validation.go#L274) for kube-proxy
```

designs/expose-metrics.md
```diff
@@ -0,0 +1,371 @@
...
+
+**Documentation:**
+
+We can add `controlPlaneConfiguration.controllerManagerExtraArgs.bindAddress` and `controlPlaneConfiguration.schedulerExtraArgs.bindAddress` as an optional configuration to our EKS-A docs
```
chrisnegus, line 364:
```suggestion
We can add `controlPlaneConfiguration.controllerManagerExtraArgs.bindAddress` and `controlPlaneConfiguration.schedulerExtraArgs.bindAddress` as an optional configuration to our EKS Anywhere docs
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "Expose metrics for EKS-A components"
```
chrisnegus, line 2:
```suggestion
title: "Expose metrics for EKS Anywhere components"
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+linkTitle: "Expose metrics"
+weight: 100
+date: 2024-04-06
+description: >
+  Expose metrics for EKS-A components
```
chrisnegus, line 7:
```suggestion
  Expose metrics for EKS Anywhere components
```

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+---
+
+Some Kubernetes system components like kube-controller-manager, kube-scheduler and kube-proxy expose metrics only on the localhost by default. In order to expose metrics for these components so that other monitoring systems like Prometheus can scrape them, you can deploy a proxy as a Daemonset on the host network of the nodes. The proxy pods also need to be configured with control plane tolerations so that they can be scheduled on the control plane nodes.
+
+To configure a proxy for exposing metrics on an EKS-A cluster, you can perform the following steps:
```
chrisnegus, line 12:
```suggestion
To configure a proxy for exposing metrics on an EKS Anywhere cluster, you can perform the following steps:
```

chrisnegus (commented):
Looks good. A lot of my comments were just that we spell out "EKS Anywhere" and "Kubernetes" in public docs.

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
sp1999: Yes, moved it to another sub-section

docs/content/en/docs/clustermgmt/observability/expose-metrics.md
sp1999: I have updated the docs to have the client permissions configuration steps in a separate subsection. I think it should be fine to continue with the current example in the context of that subsection.

PR #7988, opened 2024-04-15
Fixing registry mirror setup function return

pokearu:
*Description of changes:*
Looks like the setup method was returning early and not checking the conditions below for Docker Airgapped and OCINamespaces.


abhinavmpandey08 (approved):
/approve
/lgtm
/woof

PR #8001, opened 2024-04-17
Log in to public ECR for E2E test

chrisdoherty4:
Some tests are being rate limited when pulling from public ECR. This logs all tests into public ECR when running the CI so we avoid the 100 pull rate limit.


abhinavmpandey08 (approved):
/approve
/lgtm

PR #8007, opened 2024-04-18
Log test runner instance config in e2e

chrisdoherty4:
Its difficult to understand what the configuration when launhcing a test runner is. This logs the configuration giving insight into IPs among other information.


abhinavmpandey08 (approved):
/approve

PR #8018, opened 2024-04-19
Add check for nil eksaversion when setting etcd url

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
Related to https://github.com/aws/eks-anywhere/pull/7295

If a user has a management cluster with the new version but workload clusters created from pre 0.17, the eksaversion field is not set as bundlesRef was used then. Due to this for external etcd clusters, there would be a nil dereference, causing a panic in the controller. This will return an empty string if the cluster does not have the eksaversion set, which is usually set when a cluster is upgraded to a newer version.

*Testing (if applicable):*
unit testing and functional testing

### DIFF
--- a/pkg/clusterapi/etcd.go
+++ b/pkg/clusterapi/etcd.go
@@ -15,7 +15,7 @@ import (
 )
 
 // SetUbuntuConfigInEtcdCluster sets up the etcd config in EtcdadmCluster.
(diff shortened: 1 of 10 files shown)


vignesh-goutham (commented):
/lgtm

PR #8101, opened 2024-05-03
Change the cloudwatch namespace of quick e2e test metrics

d8660091:
*Issue #, if available:*

*Description of changes:* Quick e2e tests set are different than those in the nightly run pipeline, so the quick e2e test metrics should be separated. And the easiest way to separate them is through namespaces.

*Testing (if applicable):* Manually invoked the quick test run against the merging branch.


internal/test/e2e/cloudwatch.go
```diff
@@ -15,6 +16,8 @@ import (
 
 var svc *cloudwatch.CloudWatch
 
+const INTEGRATION_TEST_CLOUDWATCH_NAMESPACE_OVERRIDE = "INTEGRATION_TEST_CLOUDWATCH_NAMESPACE_OVERRIDE"
```
cxbrowne1207, line 19: How about?
```suggestion
const IntegrationTestCloudWatchNamespaceOverrideEnvVar = "INTEGRATION_TEST_CLOUDWATCH_NAMESPACE_OVERRIDE"
```

It seems that the linter does not like the use of ALL_CAPS in Go names. This also makes it clear that this constant holds the name of an env var

d8660091: Right, I commited this change

cxbrowne1207 (approved):

PR #8136, opened 2024-05-10
Add 1.30 Vsphere E2E tests

rahulbabu95:
*Issue #, if available:*
Add E2E tests for Vsphere provider for new K8s version 1.30. 

*Description of changes:*

*Testing (if applicable):*
Ran all the 1.30 quick tests against my local branch and these are the [results](https://tiny.amazon.com/17pdjtxm4/IsenLink) . Out of three failures 2 BR tests fail because of the disk size validation which should be fixed once the templates are fixed in CI. The TestVSphereKubernetes129To130StackedEtcdRedHatUpgrade test failed with `template /SDDC-Datacenter/vm/Templates/kubernetes-1-29-eks-10-redhat is missing tag eksdRelease:kubernetes-1-30-eks-3` which can be considered a flake because of a random race condition that happens in our controller when the template in Machine Config is updated but not actually reflected when the validations are being checked on the cluster object. The other TestVSphereKubernetes129To130RedHatUpgrade test passes which ensures that it is not related to any newly added framework function as they are being used by both the tests.


rahulbabu95:
Ran all the 1.30 quick tests against my local branch and these are the [results](https://tiny.amazon.com/17pdjtxm4/IsenLink). Out of three failures 2 BR tests fail because of the disk size validation which should be fixed once the templates are fixed in CI. The TestVSphereKubernetes129To130StackedEtcdRedHatUpgrade test failed with `template /SDDC-Datacenter/vm/Templates/kubernetes-1-29-eks-10-redhat is missing tag eksdRelease:kubernetes-1-30-eks-3` which can be considered a flake cos of the random race condition that happens in our controller when the Machine Config is updated but not actually reflected when the validations are being checked. The other TestVSphereKubernetes129To130RedHatUpgrade test passes which ensures that it is not related to any newly added framework function as they are being used by both the tests.

PR #8140, opened 2024-05-10
Add poweroff hardware cleanup step after Tinkerbell E2E tests

cxbrowne1207:
*Issue #, if available:*

*Description of changes:*
This PR makes changes to ensure that Tinkerbell hardware for a test are powered off after a run. A few changes were made as a result:
- Changed `--cleanup-vms` to `--cleanup-resources` to make it more generic (because tinkerbell does not use VMs).
- Changed the E2E test `Provider.CleanupVMs` method to `Provider.CleanupResources` 
- Move `powerOffHardware` out of `ValidateHardwareDecommissioned` so it does not run twice at the end of existing tests.
- Implemented the E2E test Tinkerbell `Provider.CleanupResources` method to power off hardware machines.

*Testing (if applicable):*


internal/test/bmc.go
```diff
@@ -0,0 +1,17 @@
+package test
```
jacobweinstock, line 1: I can understand why this might make sense to go here, but if we are to ever get to a point where we can make our e2e test tool a completely different Go module, remove the spaghetti dependency tree, and drastically reduce the imports of our actual product code, eksctl-anywhere, we should not continue to add to this package tree. I recommend that we try to stick with the package tree that starts with the top level `test` directory/package. The `internal/test/cleanup` package would be ok for now but all that clean up code should ideally live in the top level `test` directory/package as well.

Also, i see that this function probably went here to that it can be consumed in 2 locations. Duplicating this code where it's used as opposed to creating a new dependency might be a good thing here. This is not a lot of code and taking on the extra code over the dependency, to me, makes sense. "[A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)" - https://go-proverbs.github.io/

internal/test/cleanup/cleanup.go
```diff
@@ -189,3 +195,73 @@ func NutanixTestResources(clusterName, endpoint, port string, insecure, ignoreEr
 	}
 	return nil
 }
+
+// TinkerbellTestResources cleans up machines by powering them down.
+func TinkerbellTestResources(ctx context.Context, inventoryCSVFilePath string, ignoreErrors bool) error {
```
jacobweinstock, line 200: You moved other code from `vms` to `machines`. This function defined `resources`. Are resources different than machines?

internal/test/cleanup/cleanup.go
```diff
@@ -189,3 +195,73 @@ func NutanixTestResources(clusterName, endpoint, port string, insecure, ignoreEr
...
+
+	state, err := bmcClient.GetPowerState(ctx)
+	if err != nil {
+		state = "unknown"
+	}
+	if strings.Contains(strings.ToLower(state), "off") {
+		return nil
+	}
+
+	if _, err := bmcClient.SetPowerState(ctx, "off"); err != nil {
```
jacobweinstock, line 253: We talked internally about the importance of machines to be off (because of duplicate DHCP servers running) before each test run. The current model, which i know this PR is mostly just moving code around and not new code you've created, doesn't provide us any guarantees that the hardware is actually power off. Do we want to address this kind of guarantee in this PR?

test/framework/tinkerbell.go
```diff
@@ -146,8 +149,11 @@ func (t *Tinkerbell) WithProviderUpgrade(fillers ...api.TinkerbellFiller) Cluste
 	}
 }
 
-func (t *Tinkerbell) CleanupVMs(_ string) error {
-	return nil
+// CleanupMachines runs a clean up the Tinkerbell machines which simply powers them down.
+func (t *Tinkerbell) CleanupMachines(_ string) error {
+	ctx, done := context.WithTimeout(context.Background(), 2*time.Minute)
```
jacobweinstock, line 154: This context appears to be shared between all bmclib calls to all machines in each test run. This means that all machines have a combined duration of 2minutes for all `GetPowerState` and `SetPowerState` that run. This might not be enough time. BMC's can be notoriously slow. Also, if any call takes the length of the duration then all subsequent calls will fail as the context will have been canceled/timed out. I recommend adding the timeout in `powerOffHardware`.

cxbrowne1207: Ah, thanks. Good catch., I'll move it there as suggested.

internal/test/cleanup/cleanup.go
cxbrowne1207: You're right, I agree that this might be a good opportunity to introduce a mechanism to verify that the hardware is indeed powered off before tests commence. 

I believe we could add some retries on failure to address this. But even then, if the retries fail in that case we could raise some kind of alert that requires manual intervention. We currently don't fail the test in the cleanup stage, which I think is the most accurate way to guarantee and raise alarm to machines not powered off. I'd love to hear your thoughts on that and if that should change.

That said, I would like to keep this PR focused on the current task and maybe address the guarantee issue in another. Let me know if you think this works for you.

internal/test/bmc.go
cxbrowne1207: Thanks for calling this out.

Yeah, it went here so that it can be consumed in 2 locations. Now,  I agree with the principle that a little copying is better than a little dependency, especially since it's not a lot of code. I will go with duplicating this rather than introducing the dependency.

internal/test/cleanup/cleanup.go
cxbrowne1207: In this specific instance, `TinkerbellTestResources` is intended to handle machines. I named it this way to align with existing method of naming these cleanups. The term "resources" was used more broadly to encompass potential future expansions where we might manage other types of test resources, not just machines. However, I see how this could lead to confusion given our current usage. I'll rename this to `TinkerbellTestMachines` and we could always expand around it if needed.

internal/test/cleanup/cleanup.go
```diff
@@ -189,3 +195,86 @@ func NutanixTestResources(clusterName, endpoint, port string, insecure, ignoreEr
...
+	if err != nil {
+		return fmt.Errorf("failed to create hardware map from inventory csv: %v", err)
+	}
+
+	logger.Info("Powering off hardware: %+v", hardwarePool)
+	return powerOffHardwarePool(hardwarePool, ignoreErrors)
+}
+
+func powerOffHardwarePool(hardware map[string]*hardware.Machine, ignoreErrors bool) error {
+	ctx, done := context.WithTimeout(context.Background(), 2*time.Minute)
```
jacobweinstock, line 211: this gives bmclib 2 minutes to run operations on all machines. is that the intended behavior? There are many scenarios where one machine could take the whole time. I would recommend you create a separate 2 minute context per `powerOffHardware` call.

cxbrowne1207: No this was not intentional, it should have been in powerOffHardware. Thanks for the catch!

cmd/integration_test/cmd/run.go
```diff
@@ -88,7 +88,7 @@ func runE2E(ctx context.Context) error {
 	maxConcurrentTests := viper.GetInt(maxConcurrentTestsFlagName)
 	testsToSkip := viper.GetStringSlice(skipFlagName)
 	bundlesOverride := viper.GetBool(bundlesOverrideFlagName)
-	cleanupVms := viper.GetBool(cleanupVmsFlagName)
+	cleanupMachines := viper.GetBool(cleanupMachinesFlagName)
```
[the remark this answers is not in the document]
cxbrowne1207: Yes, I missed renaming a few of these.

cmd/integration_test/cmd/run.go
```diff
@@ -22,7 +22,7 @@ const (
 	maxConcurrentTestsFlagName = "max-concurrent-tests"
 	skipFlagName               = "skip"
 	bundlesOverrideFlagName    = "bundles-override"
-	cleanupVmsFlagName         = "cleanup-vms"
+	cleanupMachinesFlagName    = "cleanup-resources"
```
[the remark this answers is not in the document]
cxbrowne1207: Yes, I missed renaming a few of these.

jacobweinstock (approved):

PR #8226, opened 2024-05-29
Update generate cluster config command for Tinkerbell

rahulbabu95:
*Issue #, if available:*
Update generate command for Tinkerbell to populate Tinkerbell IP in the Datacenter config and OSImage URL in Machine config object as we have defaulted the OS from BR to Ubuntu. This change also adds a check when a modular K8s version is specified with a value above 1.28 and the OS family is set to BR.
   
*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/pkg/api/v1alpha1/tinkerbelldatacenterconfig.go
+++ b/pkg/api/v1alpha1/tinkerbelldatacenterconfig.go
@@ -22,7 +22,9 @@ func NewTinkerbellDatacenterConfigGenerate(clusterName string) *TinkerbellDatace
 		ObjectMeta: ObjectMeta{
 			Name: clusterName,
 		},
-		Spec: TinkerbellDatacenterConfigSpec{},
+		Spec: TinkerbellDatacenterConfigSpec{
+			TinkerbellIP: "",
+		},
 	}
 }
 
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig.go
@@ -27,6 +27,7 @@ func NewTinkerbellMachineConfigGenerate(name string, opts ...TinkerbellMachineCo
 		Spec: TinkerbellMachineConfigSpec{
 			HardwareSelector: HardwareSelector{},
 			OSFamily:         Ubuntu,
+			OSImageURL:       "",
 			Users: []UserConfiguration{
 				{
 					Name:              "ec2-user",
--- a/pkg/api/v1alpha1/tinkerbellmachineconfig_types.go
+++ b/pkg/api/v1alpha1/tinkerbellmachineconfig_types.go
@@ -18,7 +18,7 @@ type TinkerbellMachineConfigSpec struct {
 	// OSImageURL is a URL to the OS image used during provisioning. It must include
 	// the Kubernetes version(s). For example, a URL used for Kubernetes 1.27 could
 	// be http://localhost:8080/ubuntu-2204-1.27.tgz
-	OSImageURL          string               `json:"osImageURL,omitempty"`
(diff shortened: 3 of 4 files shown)


pkg/api/v1alpha1/tinkerbellmachineconfig_types.go
```diff
@@ -18,7 +18,7 @@ type TinkerbellMachineConfigSpec struct {
 	// OSImageURL is a URL to the OS image used during provisioning. It must include
 	// the Kubernetes version(s). For example, a URL used for Kubernetes 1.27 could
 	// be http://localhost:8080/ubuntu-2204-1.27.tgz
-	OSImageURL          string               `json:"osImageURL,omitempty"`
```
abhinavmpandey08, line 21: Some versions of bottlerocket are still supported so we shouldn't do this until we stop supporting them fully

pkg/api/v1alpha1/tinkerbelldatacenterconfig.go
```diff
@@ -22,7 +22,9 @@ func NewTinkerbellDatacenterConfigGenerate(clusterName string) *TinkerbellDatace
 		ObjectMeta: ObjectMeta{
 			Name: clusterName,
 		},
-		Spec: TinkerbellDatacenterConfigSpec{},
+		Spec: TinkerbellDatacenterConfigSpec{
+			TinkerbellIP: "",
```
abhinavmpandey08, line 26: isn't `TinkerbellIP` optional for workload clusters?

pkg/api/v1alpha1/tinkerbellmachineconfig.go
```diff
@@ -27,6 +27,7 @@ func NewTinkerbellMachineConfigGenerate(name string, opts ...TinkerbellMachineCo
 		Spec: TinkerbellMachineConfigSpec{
 			HardwareSelector: HardwareSelector{},
 			OSFamily:         Ubuntu,
+			OSImageURL:       "",
```
abhinavmpandey08, line 30: any reason why we are exposing this in the machine config instead of datacenter config? I would think most users would use OSImageURL there generally

rahulbabu95: So basically when we defaulted to ubuntu we missed adding OSImageURL field. For modular upgrades machine configs is expected in OSImageURL field. Also, this is more akin to how we set templated at machine config level in other providers. Once we introduced OSImageURL on machine config level the plan was to give preference to this field and eventually deprecate OSImageURL field at Datacenter config. But we haven't reached there yet!

pkg/api/v1alpha1/tinkerbellmachineconfig_types.go
rahulbabu95: Yes, they are still supported. But we already made that API change on data center config in this [PR](https://github.com/aws/eks-anywhere/commit/6394dadbefb138af9ffd8d20bc57a6a070d09515) . So this PR is merely trying to make it consistent. And don't think again toggling them back to optional makes sense here given we are eventually deprecating.

pkg/api/v1alpha1/tinkerbelldatacenterconfig.go
rahulbabu95: Doesn't look like it's an optional [field](https://github.com/aws/eks-anywhere/blame/5acb0dacb527185383444e9f85ec3194a5818004/pkg/api/v1alpha1/tinkerbelldatacenterconfig_types.go#L13) .? I think just that it has to be same as Management cluster for workload clusters .?

abhinavmpandey08 (approved):
/lgtm
/approve

PR #8231, opened 2024-05-30
Update CLI logging output to be cluster type aware:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
This removes the confusion in the CLI logs when creating a management cluster as the log lines say it's creating a workload cluster. This is a leak of implementation details about how clusters are created. While CAPI calls the initial cluster a workload cluster and then pivots its control plane objects to make it a permanent management cluster this is an implementation detail we shouldn't expose to users. 

*Testing (if applicable):*

### DIFF
--- a/pkg/workflows/management/create_install_eksa.go
+++ b/pkg/workflows/management/create_install_eksa.go
@@ -47,7 +47,7 @@ func (s *installEksaComponentsOnWorkloadTask) Run(ctx context.Context, commandCo
 		commandContext.ClusterSpec.Cluster.ClearTinkerbellIPAnnotation()
 	}
 
-	logger.Info("Installing EKS-A custom components on workload cluster")
+	logger.Info("Installing EKS-A custom components on the management cluster")
 
 	err := installEKSAComponents(ctx, commandContext, commandContext.WorkloadCluster)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/workflows/management/create_install_eksa.go
```diff
@@ -47,7 +47,7 @@ func (s *installEksaComponentsOnWorkloadTask) Run(ctx context.Context, commandCo
 		commandContext.ClusterSpec.Cluster.ClearTinkerbellIPAnnotation()
 	}
 
-	logger.Info("Installing EKS-A custom components on workload cluster")
+	logger.Info("Installing EKS-A custom components on the management cluster")
```
mitalipaygude, line 50: nit: initially the workload cluster in all these places meant wrt the kind cluster being the mgmt cluster vs the workload cluster getting created by the kind cluster.  

Looks fine to me if we want to make this change tho since by this time we already would've moved the capi components.

jacobweinstock: I don't believe a new customer coming to eksa for the first time is going to grasp this implementation detail. They have asked for a management cluster and are now seeing logs about a workload cluster.

jacobweinstock: Think of the cognitive load we're adding to customers just to understand this one log line. They have to understand quite a lot of implementation details of how we create clusters. I believe that's an unknown and unreasonable expectation.

mitalipaygude: Yeah it looks fine to me since for the customer can understand this better. When we use this statement for debugging the engineer can probably figure out which cluster we are talking about. LGTM!

mitalipaygude (approved):

PR #8238, opened 2024-06-01
Fetch latest E2E test AMI ID dynamically

abhay-krishna:
This PR adds the logic to dynamically fetch the latest AMI ID to use for E2E tests. We do this by calling the `DescribeImages` API with filters for the name, description and owner and then getting the latest image by creation time.

### DIFF
--- a/cmd/integration_test/build/script/create_infra_config.sh
+++ b/cmd/integration_test/build/script/create_infra_config.sh
@@ -17,11 +17,27 @@ set -e
 set -x
 set -o pipefail
 
+REPO_ROOT=$(git rev-parse --show-toplevel)
+source $REPO_ROOT/test/e2e/E2E_AMI_FILTER_VARS
+
+INTEGRATION_TEST_AMI_ID=$(aws ec2 describe-images \
+  --profile ${AWS_PROFILE} \
+  --owners ${AMI_OWNER_ID_FILTER} \
+  --filters "Name=name,Values=${AMI_NAME_FILTER}" "Name=description,Values=${AMI_DESCRIPTION_FILTER}" \
+  --query 'sort_by(Images, &CreationDate)[-1].[ImageId]' \
+  --output text
+)
+
+if [ -z "$INTEGRATION_TEST_AMI_ID" ]; then
+  echo "INTEGRATION_TEST_AMI_ID cannot be empty. Exiting"
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


scripts/e2e_test_docker.sh
```diff
@@ -39,36 +39,44 @@ BIN_FOLDER=$REPO_ROOT/bin
...
 EOF
 
+unset AWS_ROLE_ARN AWS_WEB_IDENTITY_TOKEN_FILE
+
 INTEGRATION_TEST_INFRA_CONFIG="/tmp/test-infra.yml"
 export T_TINKERBELL_S3_INVENTORY_CSV_KEY="inventory/den80/den80-hardware.csv"
 
+INTEGRATION_TEST_AMI_ID=$(aws ec2 describe-images \
+  --profile ${AWS_PROFILE} \
+  --owners 857151390494 \
```
g-gaston, line 65: can't account be dynamic?

g-gaston (commented):
are we gonna generate new amis automatically every week?

scripts/e2e_test_docker.sh
abhay-krishna: Since we only build it in the build account and share it to these accounts, we are hardcoding the owner here.

abhay-krishna:
> are we gonna generate new amis automatically every week?

The AMI generation will be triggered by us every week, but it will get used automatically in CI.

cmd/integration_test/build/script/create_infra_config.sh
```diff
@@ -17,11 +17,22 @@ set -e
 set -x
 set -o pipefail
 
+REPO_ROOT=$(git rev-parse --show-toplevel)
+source $REPO_ROOT/test/e2e/E2E_AMI_FILTER_VARS
+
+INTEGRATION_TEST_AMI_ID=$(aws ec2 describe-images \
```
g-gaston, line 23: nit: could check the content of this var before passing it to the config file? that way we can error out and make the error explicit if for whatever reason we can't find a valid AMI

g-gaston (commented):
rest looks good

cmd/integration_test/build/script/create_infra_config.sh
abhay-krishna: Added a check!

g-gaston (commented):
/lgtm

PR #8240, opened 2024-06-02
Marking packages commands as deprecated

pokearu:
*Issue https://github.com/aws/eks-anywhere-internal/issues/2270*

*Description of changes:*
As described in the issue we want to mark packages commands as deprecated. For release v0.20.0 the commands are still present and in the next release we would remove the commands themselves.

*Documentation added/planned:*
Will update the docs and call out that these commands are not supported.

### DIFF
--- a/cmd/eksctl-anywhere/cmd/upgradepackages.go
+++ b/cmd/eksctl-anywhere/cmd/upgradepackages.go
@@ -55,6 +55,7 @@ var upgradePackagesCommand = &cobra.Command{
 		}
 		return nil
 	},
+	Deprecated: "refer to `https://anywhere.eks.amazonaws.com/docs/packages/packagebundles` for upgrading",
 }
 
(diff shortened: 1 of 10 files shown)


cmd/eksctl-anywhere/cmd/upgradepackages.go
```diff
@@ -55,6 +55,7 @@ var upgradePackagesCommand = &cobra.Command{
 		}
 		return nil
 	},
+	Deprecated: "refer `https://anywhere.eks.amazonaws.com/docs/packages/packagebundles` for upgrading",
```
drewvanstone, line 58:
```suggestion
	Deprecated: "refer to `https://anywhere.eks.amazonaws.com/docs/packages/packagebundles` for upgrading",
```

drewvanstone (approved):

PR #8254, opened 2024-06-05
Handle cluster status reconcile autoscaler

sp1999:
*Issue #, if available:*
[#2231](https://github.com/aws/eks-anywhere-internal/issues/2231)

*Description of changes:*
This PR updates the cluster status reconciliation logic for `workersReady` condition when the cluster autoscaler is configured to scale the worker nodes. It also removes the `replicas` field from the md template for worker node groups which have autoscaling configured. E2E test has been added to verify that the cluster status becomes ready when autoscaler is configured to scale the worker nodes.

*Testing (if applicable):*
Ran the E2E test case locally in vSphere Dev environment and it passed:
```
make local-e2e
```
Ran the following commands to test the EKS-A cli and e2e test binaries:
```
make eks-a
make e2e-cross-platform
```
Ran the following commands to check for golangci lint and unit tests:
```
make lint
make unit-test
```
Ran `make build` in the docs folder to verify that docs site is built

### DIFF
--- a/docs/content/en/docs/clustermgmt/cluster-status.md
+++ b/docs/content/en/docs/clustermgmt/cluster-status.md
@@ -77,7 +77,10 @@ Conditions provide a high-level status report representing an assessment of clus
   * `DefaultCNIConfigured` - reports the configuration state of the default CNI specified in the cluster specifications. It will be marked as `True` once the default CNI has been successfully configured on the cluster. 
   However, if the EKS Anywhere default cilium CNI has been [configured to skip upgrades]({{< relref "../getting-started/optional/cni/#use-a-custom-cni" >}}) in the cluster specification, then this condition will be marked as `False` with the reason `SkipUpgradesForDefaultCNIConfigured`.
 
-  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready.
+  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the following conditions are met:
+    * For worker node groups with [autoscaling]({{< relref "../getting-started/optional/autoscaling" >}}) configured, number of worker nodes in that group lies between the minCount and maxCount number of worker nodes as defined in the cluster specification.
+    * For fixed worker node groups, number of worker nodes in that group matches the expected number of worker nodes in those groups as defined in the cluster specification.
+    * All the worker nodes are up to date and ready.
 
   * `Ready` - reports a summary of the following conditions: `ControlPlaneInitialized`, `ControlPlaneReady`, and `WorkersReady`. It indicates an overall operational state of the EKS Anywhere cluster. It will be marked `True` once the current state of the cluster has fully reached the desired state specified in the Cluster spec.
 
--- a/docs/content/en/docs/getting-started/baremetal/bare-spec.md
+++ b/docs/content/en/docs/getting-started/baremetal/bare-spec.md
@@ -185,7 +185,7 @@ You can omit `workerNodeGroupConfigurations` when creating Bare Metal clusters. 
 >**_NOTE:_** Empty `workerNodeGroupConfigurations` is not supported when Kubernetes version <= 1.21.
 
 ### workerNodeGroupConfigurations[*].count (optional)
-Number of worker nodes. (default: `1`) Optional if autoscalingConfiguration is used, in which case count will default to `autoscalingConfiguration.minCount`.
+Number of worker nodes. (default: `1`) It will be ignored if the [cluster autoscaler curated package]({{< relref "../../packages/cluster-autoscaler/addclauto" >}}) is installed and `autoscalingConfiguration` is used to specify the desired range of replicas.
 
 Refers to [troubleshooting machine health check remediation not allowed]({{< relref "../../troubleshooting/troubleshooting/#machine-health-check-shows-remediation-is-not-allowed" >}}) and choose a sufficient number to allow machine health check remediation.
 
--- a/docs/content/en/docs/getting-started/optional/autoscaling.md
+++ b/docs/content/en/docs/getting-started/optional/autoscaling.md
@@ -35,10 +35,9 @@ Configure an EKS Anywhere worker node group to be picked up by a Cluster Autosca
           machineGroupRef:
             kind: VSphereMachineConfig
             name: worker-machine-b
-          count: 1
 ```
 
-Note that if no `count` is specified for the worker node group it will default to the `autoscalingConfiguration.minCount` value.
+Note that if `count` is specified for the worker node group, it's value will be ignored during cluster creation as well as cluster upgrade. If only one of `minCount` or `maxCount` is specified, then the other will have a default value of `0` and `count` will have a default value of `minCount`.
 
 EKS Anywhere automatically applies the following annotations to your `MachineDeployment` objects for worker node groups with autoscaling enabled. The Cluster Autoscaler component uses these annotations to identify which node groups to autoscale. If a node group is not autoscaling as expected, check for these annotations on the `MachineDeployment` to troubleshoot.
 ```
--- a/pkg/controller/clusters/status.go
+++ b/pkg/controller/clusters/status.go
@@ -12,6 +12,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anywherev1 "github.com/aws/eks-anywhere/pkg/api/v1alpha1"
+	"github.com/aws/eks-anywhere/pkg/clusterapi"
 	"github.com/aws/eks-anywhere/pkg/controller"
 )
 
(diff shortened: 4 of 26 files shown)


docs/content/en/docs/clustermgmt/cluster-status.md
```diff
@@ -77,7 +77,7 @@ Conditions provide a high-level status report representing an assessment of clus
...
 
-  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready.
+  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready for worker node groups not having [autoscaling]({{< relref "../getting-started/optional/autoscaling" >}}) configuration as well as the number of worker nodes in the cluster lies between the minCount and maxCount number of replicas defined in the cluster specification for worker node groups having autoscaling configuration.
```
vivek-koppuru, line 80: We should say something like "once the following conditions are met:
1. For worker node groups with autoscaling configured, ...
2. For fixed worker node groups, ..."

docs/content/en/docs/getting-started/baremetal/bare-spec.md
```diff
@@ -185,7 +185,7 @@ You can omit `workerNodeGroupConfigurations` when creating Bare Metal clusters.
 >**_NOTE:_** Empty `workerNodeGroupConfigurations` is not supported when Kubernetes version <= 1.21.
 
 ### workerNodeGroupConfigurations[*].count (optional)
-Number of worker nodes. Optional if autoscalingConfiguration is used, in which case count will default to `autoscalingConfiguration.minCount`.
+Number of worker nodes. It will be ignored if the [cluster autoscaler curated package]({{< relref "../../packages/cluster-autoscaler/addclauto" >}}) is installed and autoscalingConfiguration is used to specify the desired range of replicas.
```
vivek-koppuru, line 188: We should still call this out as optional only if autoscalingConfiguration is used, because we will fail validation if not as it is required otherwise.
```suggestion
Number of worker nodes. It will be ignored if the [cluster autoscaler curated package]({{< relref "../../packages/cluster-autoscaler/addclauto" >}}) is installed and `autoscalingConfiguration` is used to specify the desired range of replicas.
```

docs/content/en/docs/getting-started/optional/autoscaling.md
```diff
@@ -35,10 +35,9 @@ Configure an EKS Anywhere worker node group to be picked up by a Cluster Autosca
           machineGroupRef:
             kind: VSphereMachineConfig
             name: worker-machine-b
-          count: 1
 ```
 
-Note that if no `count` is specified for the worker node group it will default to the `autoscalingConfiguration.minCount` value.
+Note that if `count` is specified for the worker node group, it will be ignored during cluster creation as well as cluster upgrade.
```
vivek-koppuru, line 40: We should talk about the defaulting logic that CAPI follows, which is set it to minCount or something right? And maybe even reference their docs talking about it if possible.

pkg/controller/clusters/status.go
```diff
@@ -215,6 +222,13 @@ func updateWorkersReadyCondition(cluster *anywherev1.Cluster, machineDeployments
 			return
 		}
 
+		// Skip updating the replicas for the machine deployments which have autoscaling configuration annotation
+		if md.ObjectMeta.Annotations != nil {
+			if _, ok := md.ObjectMeta.Annotations[clusterapi.NodeGroupMinSizeAnnotation]; ok {
```
vivek-koppuru, line 227: Should we check for the existence of either min or max? We shouldn't assume that both need to be specified in case we change the requirement in the future, unless autoscaler requires both in general on the MD object

pkg/controller/clusters/status.go
```diff
@@ -253,6 +267,18 @@ func updateWorkersReadyCondition(cluster *anywherev1.Cluster, machineDeployments
 		return
 	}
 
+	for _, md := range machineDeployments {
```
vivek-koppuru, line 270: Add a comment that we are going through all the ASG machineDeployments now. Also, we could probably store these ones in another map rather than having to loop through again, but I guess that increases the storage.

docs/content/en/docs/getting-started/baremetal/bare-spec.md
sp1999: It's an optional field even if autoscalingConfiguration is not used as it will default to `1` if not specified. We won't fail any validation if it is not specified.

docs/content/en/docs/getting-started/optional/autoscaling.md
sp1999: Agreed, I will update it

pkg/controller/clusters/status.go
sp1999: If both min and max are not specified, our controller considers it as autoscaling configuration not being set and does not add autoscaler annotations to the md object. So we can assume that atleast one of them must be specified. By default, capi defaults the min and max count to 0 if not specified. So if only min is specified, then max will be defaulted to 0 and min must be set to 0 as we have a validation which checks that min is less than or equal to max. In either case where only of them is specified, there will be annotations for both of them on the md object so it should be fine to check for existence on either of min or max.

pkg/controller/clusters/status.go
sp1999: I will add a comment for now.

pkg/controller/clusters/status.go
vivek-koppuru: It will be fine for us, but let's say we change that behavior in the future. Is it a requirement for autoscaler for both to exist on the MD object?

docs/content/en/docs/clustermgmt/cluster-status.md
```diff
@@ -77,7 +77,9 @@ Conditions provide a high-level status report representing an assessment of clus
...
 
-  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready.
+  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the following conditions are met:
+    * For worker node groups with [autoscaling]({{< relref "../getting-started/optional/autoscaling" >}}) configured, number of worker nodes in the cluster lies between the minCount and maxCount number of replicas as defined in the cluster specification.
+    * For fixed worker node groups, number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready.
```
vivek-koppuru, line 82:
```suggestion
    * For fixed worker node groups, number of worker nodes in that group matches the expected number of worker nodes in those groups as defined in the cluster specification.
    * All the worker nodes are up to date and ready.
```

docs/content/en/docs/clustermgmt/cluster-status.md
```diff
@@ -77,7 +77,9 @@ Conditions provide a high-level status report representing an assessment of clus
...
   However, if the EKS Anywhere default cilium CNI has been [configured to skip upgrades]({{< relref "../getting-started/optional/cni/#use-a-custom-cni" >}}) in the cluster specification, then this condition will be marked as `False` with the reason `SkipUpgradesForDefaultCNIConfigured`.
 
-  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the number of worker nodes in the cluster match the expected number of worker nodes as defined in the cluster specifications and all the worker nodes are up to date and ready.
+  * `WorkersReady` - reports that the condition of the current state of worker machines versus the desired state specified in the Cluster specification. This condition is marked `True` once the following conditions are met:
+    * For worker node groups with [autoscaling]({{< relref "../getting-started/optional/autoscaling" >}}) configured, number of worker nodes in the cluster lies between the minCount and maxCount number of replicas as defined in the cluster specification.
```
vivek-koppuru, line 81:
```suggestion
    * For worker node groups with [autoscaling]({{< relref "../getting-started/optional/autoscaling" >}}) configured, number of worker nodes in that group lies between the minCount and maxCount number of worker nodes as defined in the cluster specification.
```

docs/content/en/docs/getting-started/optional/autoscaling.md
```diff
@@ -35,10 +35,9 @@ Configure an EKS Anywhere worker node group to be picked up by a Cluster Autosca
           machineGroupRef:
             kind: VSphereMachineConfig
             name: worker-machine-b
-          count: 1
 ```
 
-Note that if no `count` is specified for the worker node group it will default to the `autoscalingConfiguration.minCount` value.
+Note that if `count` is specified for the worker node group, it's value will be ignored during cluster creation as well as cluster upgrade. If both `minCount` and `maxCount` are not specified, then autoscaling will not be configured for that worker node group and `count` will have a default value of `1`. If only one of `minCount` or `maxCount` is specified, then the other will have a default value of `0` and `count` will have a default value of `minCount`.
```
vivek-koppuru, line 40: If both aren't specified, do you mean if the `autoscalingConfiguration` field is specified with no fields underneath? The minCount and maxCount are pointers? And it's possible for minCount to be 0 if not specified and set the count to 0?

pkg/controller/clusters/status.go
sp1999: Yes, autoscaler validates that both annotations exist on the md object for that worker node group to be considered a scalable resource. 
Ref:- https://github.com/kubernetes/autoscaler/blob/83db225ce2c4b368a8ee30152691746ed9e6d5ae/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go#L119C1-L149C2

docs/content/en/docs/getting-started/optional/autoscaling.md
sp1999: If `autoscalingConfiguration` is specified as `null` in the cluster spec, then the asg struct will be a nil object and autoscaling won't be configured for that worker node group. If it's specified as an empty struct with curly brackets `{}`, then min count and max count will be defaulted to `0` and so count will also be defaulted to min count which is `0`.

sp1999: I will update the docs to remove that statement to make it less confusing.

vivek-koppuru (approved):

PR #8264, opened 2024-06-08
Shuffle hardware inventory for tinkerbell before reservation

rahulbabu95:
*Description of changes:*
Shuffle hardware inventory before reserving hardware for Tinkerbell E2E tests. As we run quick e2e more frequently the boot entries on the boot list get populated quickly leading to an error when there's no space left to add to that boot list. Ideally we should have an automation around removing the boot entries periodically on the BMCs but until then we should try reserving the hardware  in random order for quick test to not burden the boot entries on the first few hardware. Also, with randomness we reduce the likelihood of picking up an erroneous hardware in case during repetitive quick E2E runs.

*Testing (if applicable):*
Kicked of [run](https://tiny.amazon.com/10c5mxgd8/IsenLink) against my branch and verified that the hardware reserved for the test were different from the regular hardware (`eksa-ci01 to eksa-ci12`) that gets reserved at present. 

### DIFF
--- a/internal/test/e2e/run.go
+++ b/internal/test/e2e/run.go
@@ -2,7 +2,6 @@ package e2e
 
 import (
 	"fmt"
-	"math/rand"
 	"os"
 	"regexp"
 	"sort"
@@ -220,7 +219,11 @@ func RunTests(conf instanceRunConf, inventoryCatalogue map[string]*hardwareCatal
 		}
 		conf.Logger.Info("Shuffling hardware inventory for tinkerbell")
 		// shuffle hardware to introduce randomness during hardware reservation.
-		shuffleHardwareInventory(hardwareCatalogue)
+		// we do not want quick e2e runs to always pick the first few available hardware from the list and over-populate the boot entries
+		// this will quickly break the booting process as the hardware runs out of boot space to store these entries.
+		// randomly picking the hardware will distribute the boot entries across these hardware during each run
+		// ideally for long term we want a clear cleanup of the boot entries in the hardware
+		hardwareCatalogue.shuffleHardware()
 		err = reserveTinkerbellHardware(&conf, hardwareCatalogue)
 		if err != nil {
 			return "", nil, err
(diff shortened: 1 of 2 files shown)


internal/test/e2e/run.go
```diff
@@ -592,3 +596,10 @@ func logTinkerbellTestHardwareInfo(conf *instanceRunConf, action string) {
 	}
 	conf.Logger.V(1).Info(action+" hardware for TestRunner", "hardwarePool", strings.Join(hardwareInfo, ", "))
 }
+
+func shuffleHardwareInventory(invCatalogue *hardwareCatalogue) {
```
g-gaston, line 600: nit: why the use of inventory and catalogue? aren't they representing the same thing?

internal/test/e2e/run.go
```diff
@@ -217,6 +218,9 @@ func RunTests(conf instanceRunConf, inventoryCatalogue map[string]*hardwareCatal
 		} else {
 			hardwareCatalogue = inventoryCatalogue[nonAirgappedHardware]
 		}
+		conf.Logger.Info("Shuffling hardware inventory for tinkerbell")
+		// shuffle hardware to introduce randomness during hardware reservation.
```
g-gaston, line 222: nit: I would expand more on why randomness is desired. We don't do this to introduce randomness, we do this to avoid picking up the same machines on every run. Randomness is just the mechanism to achieve that goal.

g-gaston (commented):
overall lgtm, only nit comments

Have we planned already the work to clean the boot entries? I'm totally ok merging this, it's a good patch, but it doesn't guarantee the problem won't happen again. In fact if I'm understanding this correctly, it will 100% happen, it will just take longer. And it doesn't seem like an easy issue to diagnose.

internal/test/e2e/run.go
g-gaston, line 600: Why not make this a method in `hardwareCatalogue`? It's manipulating the internal extructure, it seems like a good idea to abstract that in a method instead of exposing it like this.

I fact, don't you need to use the mutex? If I'm not mistaken the `hardwareCatalogue` is shared between runner threads and all of them are going to try to call this method concurrently.

rahulbabu95: Good catch! I should have seen this.

rahulbabu95:
> overall lgtm, only nit comments
> 
> Have we planned already the work to clean the boot entries? I'm totally ok merging this, it's a good patch, but it doesn't guarantee the problem won't happen again. In fact if I'm understanding this correctly, it will 100% happen, it will just take longer. And it doesn't seem like an easy issue to diagnose.

I think long term, we were looking into a solution to automate the cleanup of hardware. Jacob has a runbook on how to do it manually and we still have to figure some nitbits there before it could be fully automated. I can create a issue on CI board to see if that has to be tracked.

g-gaston (commented):
/lgtm

PR #8310, opened 2024-06-12
Remove remaining 1.25 E2E tests and update API tests to latest Kube version

rahulbabu95:
Deprecate 1.25 E2E tests for the release and it's associated helper functions. Additionally there were a bunch of API tests that were never added for new Kubernetes version as currently we have a unified flow between CLI and controller. This change removes redundant API tests that are already tested in the CLI flow and bumps the Kubernetes version to the latest for those API tests that aren't tested in the CLI flow. If a single API test is being tested by two different CLI test like updating label and updating taints we still deprecate them as we are testing those flows. Going forward we should find a way to test the workflows that API tests are using through the CLI test.   

*Description of changes:*

*Testing (if applicable):*
Built E2E cross-binary to ensure e2e binary builds fine. 


test/e2e/SKIPPED_TESTS.yaml
```diff
@@ -82,32 +78,28 @@ skipped_tests:
 - TestTinkerbellKubernetes126To127Ubuntu2204Upgrade
 - TestTinkerbellKubernetes126Ubuntu2004To2204Upgrade
 - TestTinkerbellKubernetes127Ubuntu2004To2204Upgrade
-- TestTinkerbellKubernetes125UbuntuAddWorkerNodeGroupWithAPI
 - TestTinkerbellSingleNode125ManagementScaleupWorkloadWithAPI
```
cxbrowne1207, line 81: Do we need to remove this one as well?

test/e2e/cloudstack_test.go
```diff
@@ -3923,7 +3691,7 @@ func TestCloudStackMulticlusterWorkloadClusterNewCredentialsSecretGitHubFluxAPI(
 		api.CloudStackToConfigFiller(
 			api.WithCloudStackCredentialsRef("test-creds"),
 		),
-		cloudstack.WithRedhat9Kubernetes126(),
+		cloudstack.WithRedhat9Kubernetes130(),
```
ahreehong, line 3694: does this need to be a higher k8s version? the original test uses 1.26 here rather than 1.25

g-gaston (commented):
I'm not sure removing the API tests is the right thing to do here
Can we postpone that decision? The issue here is that they were using 1.25, can we just use 1.30 wherever those tests were using 1.25?

We can discuss the removal of tests later. I would like for Xu to finish his documentation of features and coverage before removing any tests.

test/e2e/cloudstack_test.go
rahulbabu95: Yeah, I checked on this with Cavaughn. So the intention of these tests is to update the CS credential on workload cluster and the kubernetes version used doesn't really matter so wanted to just bump it to latest.

ahreehong (approved):

PR #8325, opened 2024-06-14
Fix e2e tests creating old bundle workload clusters

g-gaston:

### DIFF
--- a/test/e2e/vsphere_test.go
+++ b/test/e2e/vsphere_test.go
@@ -4329,6 +4329,7 @@ func TestVSphereKubernetes127to128UpgradeFromLatestMinorReleaseBottleRocketAPI(t
 	runMulticlusterUpgradeFromReleaseFlowAPI(
 		test,
 		release,
+		wc.ClusterConfig.Cluster.Spec.KubernetesVersion,
 		v1alpha1.Kube128,
 		framework.Bottlerocket1,
 	)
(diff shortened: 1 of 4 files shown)


test/e2e/vsphere_test.go
```diff
@@ -4329,6 +4329,7 @@ func TestVSphereKubernetes127to128UpgradeFromLatestMinorReleaseBottleRocketAPI(t
 	runMulticlusterUpgradeFromReleaseFlowAPI(
 		test,
 		release,
+		wc.ClusterConfig.Cluster.Spec.KubernetesVersion,
```
d8660091, line 4332: is Kube127 here more readable here? or creating a variable called oldVersion?

d8660091: My motivation was the annoy to figure out what this value really is, so I had an idea maybe simplify it can be helpful. 

But taking another look, maybe it can be inferred rom test (the first parameter), like test.ManagementCluster.ClusterConfig.Cluster.Spec.KubernetesVersion.

g-gaston: I'm not sure i follow. What is annoying? The kube version is set in the workload cluster explicitly

g-gaston: when you say `the definition of this parameter out of the 5 parameter` are talking about what this argument is used for inside `runMulticlusterUpgradeFromReleaseFlowAPI`? if so, that's in the function doc comment I wrote

abhinavmpandey08 (approved):

test/e2e/vsphere_test.go
d8660091: > Can this parameter deduced from test?
I think the parameter oldVersio is not needed, it can be replaced with inside runMulticlusterUpgradeFromReleaseFlowAPI. 5 Parameters are too much, doc doesn't help much for this case.

PR #8339, opened 2024-06-17
Fix issue where incorrect flag is getting passed to helm commands

ahreehong:
*Issue #, if available:*

*Description of changes:*
Fix issue where incorrect additional flags were being passed to the helm upgrade command. Pass in empty additional flags if none are provided to prevent reusing the flags from a previous helm command. 

We might want to revisit the logic regarding passing in helm opts to the upgrade and uninstall helm commands. We currently need it due to the changing `NO_PROXY` values during the create/upgrade flow using proxy configs. 

*Testing (if applicable):*


abhinavmpandey08 (approved):
/lgtm
/approve
/woof

PR #8393, opened 2024-06-26
[Draft] Group docker simple flow tests to suite

d8660091:
*Issue #, if available:*

*Description of changes:*
1. Convert Docker SimpleFlow to subtests, removed duplicated code
2. Quick e2e test regexps are now passing as a new flag, to make it allow to pick an subtest inside the TestSuite

*Testing (if applicable):* Manually invoked the quick-e2e-tests with this code change 


test/e2e/docker_test.go
```diff
@@ -1480,3 +1434,16 @@ func TestDockerKubernetes130KubeletConfigurationSimpleFlow(t *testing.T) {
 	)
 	runKubeletConfigurationFlow(test)
 }
+
+func TestDockerKubernetesSimpleFlowSuite(t *testing.T) {
```
jaxesn, line 1438: I want to see a mock of this method that handles all OSs and Providers, you dont have to make it work out of the gate, but Id like to see it so we can iterate on the design here before rolling this out.

d8660091: Guillermo has made one showing how the suite handles OSs and Providers: https://github.com/aws/eks-anywhere/compare/main...g-gaston:eks-anywhere:poc-multi-test?expand=1

d8660091: I have a gut feeling, parameterize provider could make the e2e test harder to write, which should be as simple as a test yaml file + "eksa create cluster". I am OK with parameterize OS though

PR #8449, opened 2024-07-03
Set go version in govulncheck action for release

g-gaston:
## Description of changes
For some reason, we can't use `go 1.22` in the release/cli go.mod. It needs to have a patch version of `go mod tidy` will change it. I'm not sure why this is, since not specifying the patch is completely valid. I suspect one of our dependencies has set 1.22.0 in their go.mod, and that makes it a requirement for us, which is annoying.

The issue with selecting a particular patch is that, when a new version of go is required bc of a CVE, govulncheck will fail until the the go.mod is updated, it won't pull the latest version.

For now, we update the github action so we set the go version explicitly instead of reading from the go.mod file.

The drawback is the github action will also need to be updated when we bump to 1.23. But probably not a big deal.

### DIFF
--- a/.github/workflows/vulnerability.yml
+++ b/.github/workflows/vulnerability.yml
@@ -7,8 +7,9 @@ on:
       - main
   pull_request:
     paths-ignore:
-    - '**.yaml'
-    - '**.yml'
+    - 'cmd/integration_test/build/buildspecs/*.yaml'
+    - 'cmd/integration_test/build/buildspecs/*.yml'
+    - 'test/e2e/QUICK_TESTS.yaml'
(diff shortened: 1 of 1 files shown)


.github/workflows/vulnerability.yml
```diff
@@ -48,4 +49,4 @@ jobs:
           repo-checkout: false
           cache: false # cache will be already setup by previous step
           work-dir: release/cli
-          go-version-file: release/cli/go.mod
+          go-version-input: '1.22'
```
rahulbabu95, line 52: why are we hardcoding the version here?

PR #8493, opened 2024-07-16
Enable lb_class_only env var in tink stack kubevip

rahulbabu95:
*Description of changes:*
Enable `lb_class_only` env var on kube-vip so that it only manages IP for services with LoadBalancerClass set to `kube-vip.io/kube-vip-class` on the service. Without this env var, kube-vip is trying to manage IPs for any service requesting for a Load Balancer IP which isn't expected.  

### DIFF
--- a/pkg/providers/tinkerbell/stack/stack.go
+++ b/pkg/providers/tinkerbell/stack/stack.go
@@ -563,6 +563,12 @@ func (s *Installer) createValuesOverride(bundle releasev1alpha1.TinkerbellBundle
 						"name":  "prometheus_server",
 						"value": ":2213",
 					},
+					// The Tinkerbell stack needs a load balancer to work properly.
+					// We bundle Kube-vip in, as the load balancer, when we deploy the stack.
+					// We don't want this load balancer to be used by any other workloads.
+					// It allows us greater confidence in successful lifecycle events for the Tinkerbell stack, amongst other things.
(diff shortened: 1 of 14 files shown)


pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -563,6 +563,10 @@ func (s *Installer) createValuesOverride(bundle releasev1alpha1.TinkerbellBundle
 						"name":  "prometheus_server",
 						"value": ":2213",
 					},
+					{
+						"name":  "lb_class_only",
```
jacobweinstock, line 567: It would be useful and helpful for future us (and other EKSA contributors) to have a code comment here that explains the business logic for this. The why this is being set. You PR description alludes to what i think is the why but it doesn't quite hit it for me. This is the business logic from my perspective, feel free to disagree or add or subtract from it. 

```Go
// The Tinkerbell stack need a load balancer to work properly.
// We bundle Kube-vip in, as the load balancer, when we deploy the stack.
// We don't want this load balancer to be used by any other workloads.
// It allows us greater confidence in successful lifecycle events for the Tinkerbell stack, amongst other things.
// Also, the user should be free from Tinkerbell stack constraints
// and free to deploy a load balancer of their choosing and not be coupled to ours. 
```

jacobweinstock (approved):

PR #8555, opened 2024-07-30
Add Nutanix failure domain documentation

yannickstruyf3:
*Issue #, if available:*
N/A
*Description of changes:*
Update Nutanix documentation to include failure domain example.
*Testing (if applicable):*
N/A


docs/content/en/docs/getting-started/nutanix/nutanix-spec.md
```diff
@@ -251,6 +273,36 @@ __Example__:</br>
...
+### failureDomains[0].name
+The name of the failure domain.
+
+### failureDomains[0].cluster (required)
+Reference to the failure domain Prism Element cluster.
+
+### failureDomains[0].cluster.type (required)
+Type to identify the failure domain Prism Element cluster. (Permitted values: `name` or `uuid`)
+
+### failureDomains[0].cluster.name (`name` or `UUID` required)
```
chrisnegus, line 289: Do you mean: "failureDomains[0].cluster.name or failureDomains[0].cluster.uuid required"? Someone might think you mean this value could be set to `name` or `UUID`. It looks like this could only be set to `name`. Same question with other objects below that say: "(`name` or `UUID` required)". There are eight such objects here.

chrisnegus (commented):
I just have one query that applies to several objects.

docs/content/en/docs/getting-started/nutanix/nutanix-spec.md
yannickstruyf3: Thanks for the feedback @chrisnegus. 
For the cluster object in a failure domain the name or the uuid are required. What would your guidance be to make this more clear? For example should we use the full path: "`failureDomains[0].cluster.name` or `failureDomains[0].cluster.UUID` required"?

chrisnegus: Yeah, maybe. My first impression was that `failureDomains[0].cluster.type` could be set to a name or a UUID, which I don't believe is the case here. So being explicit might be good.

yannickstruyf3: I have updated the documentation to use the full paths of the spec attributes. Please take a look.

sp1999:
Can you update the hash [here](https://github.com/aws/eks-anywhere/blob/main/docs/content/en/docs/osmgmt/overview.md?plain=1#L42C188-L42C220) with `#imagename-imagename-or-imageuuid-required` to pass the docs presubmit?

yannickstruyf3:
Corrected the link. Presubmit passed now.

chrisnegus (approved):
/lgtm

PR #8568, opened 2024-08-05
Enable dhcrelay for baremetal provider

ahreehong:
*Issue #, if available:*

*Description of changes:*
Add dhcrelay to the bundle and configure tinkerbell stack to use dhcrelay instead of using smee in hostNetwork mode. 

The upstream chart uses dhcp relay to allow all services, including smee, to be put behind a load balancer. We had not enabled dhcp relay for EKS-A yet because we did not have the build changes all set up. Instead, we had continued to deploy smee as `hostNetwork: true`. We are now enabling dhcprelay so that the tink stack will use dhcp relay rather than deploying smee as `hostNetwork: true` mode.

*Testing (if applicable):*
Built the dhcp relay and dhcp relay init images locally and tested using bundles override.
Verified that create mgmt, create workload clusters, upgrade clusters, and delete clusters all work as expected


pkg/providers/tinkerbell/stack/testdata/expected_upgrade.yaml
```diff
@@ -40,6 +40,8 @@ stack:
   loadBalancerIP: 1.2.3.4
   relay:
     enabled: false
+    image: {}
+    initImage: {}
```
abhinavmpandey08, line 44: Can we add the images for these to the test bundle so the test verifies the actual values?

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -544,7 +544,7 @@ func (s *Installer) createValuesOverride(bundle releasev1alpha1.TinkerbellBundle
 				},
 				"additionalKernelArgs": bootEnv,
 			},
-			"hostNetwork":     true,
+			"hostNetwork":     false,
```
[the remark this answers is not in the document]
ahreehong: yeah, we can remove this line

abhinavmpandey08 (approved):
/lgtm

abhinavmpandey08 (approved):
/lgtm

mitalipaygude:
Can we update the testing section to describe how we tested the changes?

pkg/providers/tinkerbell/stack/stack.go
```diff
@@ -584,7 +583,9 @@ func (s *Installer) createValuesOverride(bundle releasev1alpha1.TinkerbellBundle
 				enabled: s.stackService,
 			},
 			relay: map[string]interface{}{
-				enabled: false,
+				enabled:     s.dhcpRelay,
+				image:       bundle.TinkerbellStack.Tink.TinkRelay,
```
mitalipaygude, line 587: Are TinkRelay and TinkrelayInit images added by default now or just when dhcp relay is enabled?

ahreehong: It is going to be enabled by default. This is also the default behavior of the upstream chart.

PR #8569, opened 2024-08-05
Curated packages tests using regional packages for vSphere

pokearu:
*Issue #, if available:*

*Description of changes:*
Today we have only one e2e test that uses the regional (subscriptions) package registries. We want to increase the testing since we hope to make regional the default. 

*Testing:*
```
./e2e.test -test.v -test.run "TestVSphereKubernetes129RegionalCuratedPackagesSimpleFlow"
```

### DIFF
--- a/test/framework/cluster.go
+++ b/test/framework/cluster.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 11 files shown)


test/framework/cluster.go
```diff
@@ -1172,6 +1173,20 @@ func (e *ClusterE2ETest) InstallCuratedPackageFile(packageFile, kubeconfig strin
 	})
 }
 
+// ValidateIfRegionalCuratedPackage checks if the current e2e test is a Regional Curated packages test and runs a validation.
+func (e *ClusterE2ETest) ValidateIfRegionalCuratedPackage() {
+	regionalPackagesRegex := `^.*RegionalCuratedPackages.*$`
+	if regexp.MustCompile(regionalPackagesRegex).MatchString(e.T.Name()) {
```
jhaanvi5, line 1179: Would it be better to use regex.Compile instead of MustCompile over here, MustCompile panics if the compilation fails: https://pkg.go.dev/regexp#MustCompile and if we panic in the middle of a test, we might risk losing output. In case of regex.Compile we can gracefully handle the error if any.

abhay-krishna: What you say makes sense, but we have in fact been using `MustCompile` throughout our E2E [code](https://github.com/search?q=repo%3Aaws%2Feks-anywhere+MustCompile+path%3A%2F%5Einternal%5C%2Ftest%5C%2Fe2e%5C%2F%2F&type=code) to check for the test regex. So ideally we should change all of them, or just let this one be `MustCompile` as well.

jhaanvi5: Right, make sense. We can address this in separate PR as well.

PR #8576, opened 2024-08-07
Run conformance for 1.29 and above in default mode

abhinavmpandey08:
*Description of changes:*
The 129 and 130 conformance tests are failing currently because we are using the `e2e-skip` flag along with `--mode` flag, which is disallowed. 
This is from the test logs
```
conformance.go:68: Error running k8s conformance tests: executing sonobuoy: Error: mode flag and e2e-focus/skip flags both provided and may cause unintended behavior
```
To get around it and get the conformance tests passing, we are going to run 129 and 130 in the default mode so the validation doesn't fail. We can change this behavior to run with `--mode=certified-conformance` once we stop using `e2e-skip` for these versions.

*Testing (if applicable):*
Ran and verified the tests locally.

### DIFF
--- a/pkg/executables/sonobuoy.go
+++ b/pkg/executables/sonobuoy.go
@@ -27,7 +27,6 @@ func (k *Sonobuoy) Run(ctx context.Context, contextName string, args ...string) 
 		"--context",
 		contextName,
 		"run",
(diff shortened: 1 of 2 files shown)


abhay-krishna (approved):
/approve

PR #7310, opened 2024-01-17
Bottlerocket for Baremetal is being deprecated:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*
Removing e2e related tests. Leaving 3 simple flow tests until Bottlerocket support is removed.

*Testing (if applicable):*

*Documentation added/planned (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #8643, opened 2024-08-20
Update skipped tests

sp1999:
*Description of changes:*

**Commit 1**
This commit removes non-existent tests from the SKIPPED_TESTS file. This is because those tests had been removed previously for various reasons. #5799 removed external etcd tests for tinkerbell as external etcd was not supported at the time for kubernetes version v1.27. #7310 removed bottlerocket tests for tinkerbell as bottlerocket is being deprecated for baremetal which included the workload cluster upgrade tests. #7659 temporarily skipped couple of E2E tests which were broken either due to feature changes or env issues which included the `TestTinkerbellAirgappedKubernetes129UbuntuProxyConfigFlow` test but the latest v1.30 test for this flow has been passing for a while now so it's safe to remove from the SKIPPED_TESTS.

**Commit 2**
This commit adds more simple flow and upgrade flow tinkerbell tests for kubernetes versions v1.28 and v1.29 to SKIPPED_TESTS as we already skip the tests for those tests for kubernetes versions v1.26 and v1.27. Those tests are passing on the latest kubernetes version v1.30 so it's safe to skip the tests for older versions.

**Commit 3**
This commit adds v1.31 UpgradeFromLatest tests to SKIPPED_TESTS and updates the outdated tests to the latest kubernetes version v1.30 as these tests won't pass until the new EKS-A release is out with the new kubernetes version.

**Commit 4**
This commit removes tinkerbell ubuntu simple flow tests for kubernetes versions 1.26 and 1.27 from SKIPPED_TESTS as we should run at least the simple tests for all Kubernetes version to catch any version specific regression.


test/e2e/SKIPPED_TESTS.yaml
```diff
@@ -68,11 +61,20 @@ skipped_tests:
 - TestTinkerbellKubernetes130RedHatSimpleFlow
 - TestTinkerbellKubernetes126UbuntuSimpleFlow
 - TestTinkerbellKubernetes127UbuntuSimpleFlow
+- TestTinkerbellKubernetes128UbuntuSimpleFlow
+- TestTinkerbellKubernetes129UbuntuSimpleFlow
```
rahulbabu95, line 65: I think we still run some older version simpleFlow/upgrade tests to catch any regression that could be specific to a kube version. For example, we found a regression on InPlace that only impacted 1.27 version and not others. Are we skipping this only to reduce the e2e runtime?

sp1999: Since we already skip the tests for oldest versions(1.26 and 1.27), I think it makes sense to skip it for 1.28 and 1.29 too with that reasoning. But if we are worried about catching any regressions with a particular kubernetes version, then it would also apply to those older versions which are already skipped. So I feel we should either skip all the older versions and just run the latest one or don't skip any of them at all.

rahulbabu95: I think we should try running at least the simple tests for all Kubernetes version to catch any version specific regression! So let's unskip the older two versions simple flow tests

abhay-krishna (commented):
/lgtm
/approve

PR #8656, opened 2024-08-23
Rollout new nodes for OSImageURL change on spec without changing K8s version

rahulbabu95:
*Issue #, if available:*
For Baremetal user might want to patch the nodes with updated OSImage while without upgrading the K8s version. Consequently user might also want to rollout nodes with any specific config change that doesn't involve a K8s version upgrade but warrants a node rollout. Ex. API server flags update. Issue: https://github.com/aws/eks-anywhere-internal/issues/2443

*Description of changes:*

The upgrade flow only rolls out node on Tinkerbell provider for a K8s version change or when there's scale up of Control Plane or Workers. This is primarily due to the fact that there are hardware requirements for  node rollout on Tinkerbell and if EKS-A misses any of those validations during the upgrade and doesn't set an error message on the corresponding object like KCP/MD, once the spec is applied CAPT would just get stuck waiting for additional hardware to be available for the rollout. This would be much later in the stack and would be inconvenient to debug. Currently we perform a bunch of hardware validations on the CLI and controller. This cover cases like RollingUpgrades, ModularUpgrades and Scaling operations on Control Plane and Worker Node groups. 

The existing validations are implemented in a common place for CLI and controller and this is achieved by using a common [interface](https://github.com/aws/eks-anywhere/blob/e0fafe07dfd1bb6ff98df5c09efb765fbd086110/pkg/providers/tinkerbell/assert.go#L275)  `ValidatableCluster`. As we open up more flags/features for ControlPlane/Worker Groups, the existing method of validating won't be scalable as we would require this interface to implement more customized methods depending on the field of interest. This change Introduces a new Hardware Validation method that `ExtraHardwareAvailableAssertionForNodeRollOut` that could be used to validate any node rollout between ControlPlane and Workers. As we open up more feature gates through our spec more validations could be added ControlPlane or Worker Specific changes from CLI. On the controller side, only time CAPI rollsout a new node is when the underlying machine template `Spec.MachineTemplate.InfrastructureRef` changes for KCP or MD object. This change adds validation for that condition. 

Tinkerbell reconciler overrides any generated template during an upgrade by using the `omitTinkerbellMachineTemplates` method which is called in every reconcile by fetching the existing machine templates from KCP and MD objects when the version on KCP and MD doesn't change. This has caused problems when users try to just update the `TinkerbellTemplateConfig` object on the cluster spec and expect a node to be upgraded. As this change adds hardware validations on controller for a node rollout, it is safe to remove this function and the controller would fail early. 

*Testing (if applicable):*
Tested upgrades for control plane and worker node groups separately using both CLI and controller workflows.

### DIFF
--- a/pkg/providers/tinkerbell/assert.go
+++ b/pkg/providers/tinkerbell/assert.go
@@ -231,7 +231,7 @@ func MinimumHardwareAvailableAssertionForCreate(catalogue *hardware.Catalogue) C
 
 		// Build a set of required hardware counts per machine group. minimumHardwareRequirements
 		// will account for the same selector being specified on different groups.
-		requirements := minimumHardwareRequirements{}
+		requirements := MinimumHardwareRequirements{}
 
 		err := requirements.Add(
 			spec.ControlPlaneMachineConfig().Spec.HardwareSelector,
@@ -380,7 +380,7 @@ func AssertionsForScaleUpDown(catalogue *hardware.Catalogue, current Validatable
 		}
 		// Build a set of required hardware counts per machine group. minimumHardwareRequirements
 		// will account for the same selector being specified on different groups.
-		requirements := minimumHardwareRequirements{}
+		requirements := MinimumHardwareRequirements{}
 
 		if current.ControlPlaneReplicaCount() != spec.Cluster.Spec.ControlPlaneConfiguration.Count {
 			if rollingUpgrade {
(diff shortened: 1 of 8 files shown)


Cajga:
This affect us at the moment. We want to do an important change in the config of the nodes. Is there any way how I can force a new rollout other than wait for the next Kubernetes version (we are on 1.30)?

vivek-koppuru (approved):

pokearu (approved):
/lgtm

rahulbabu95:
@Cajga right now, we do not support rollout without changing the K8s version. Only other way I can think of is to try and make CAPI force rollout a node instead: https://cluster-api.sigs.k8s.io/clusterctl/commands/alpha-rollout.html. You might have the eks-a cluster controller paused during that process: https://release-0-19.anywhere.eks.amazonaws.com/docs/clustermgmt/cluster-upgrades/baremetal-upgrades/#:~:text=During%20the%20workload%20cluster%20upgrade%20process%2C%20EKS%20Anywhere%20pauses%20the%20cluster%20controller%20reconciliation%20by%20adding%20the%20paused%20annotation%20anywhere.eks.amazonaws.com/paused%3A%20true%20to%20the%20EKS%20Anywhere%20cluster%2C%20provider%20datacenterconfig%20and%20machineconfig%20resources%2C%20before%20the%20components%20upgrade.

Cajga:
@rahulbabu95 , just to report it here that I successfully managed to roll out new images pausing the eks-a cluster controller and then use the `clusterctl alpha rollout restart` for the kubeadmcontrolplane and to the machinedeployment.

Thanks for the hint!

rahulbabu95:
@Cajga great, we also have a v0.20.5 release out which has this fix included. You will not have to manually rollout nodes going forward!

Cajga:
@rahulbabu95 great news. Thanks for the support.

PR #8759, opened 2024-09-13
Remove 1.26 conformance tests

sp1999:
*Issue #, if available:*
[#2399](https://github.com/aws/eks-anywhere-internal/issues/2399)

*Description of changes:*
Remove conformance E2E tests for kubernetes version 1.26

*Testing (if applicable):*
make build-all-test-binaries
make lint

### DIFF
--- a/pkg/controller/clientutil/kubernetes_test.go
+++ b/pkg/controller/clientutil/kubernetes_test.go
@@ -122,7 +122,7 @@ func TestKubeClientUpdate(t *testing.T) {
 	client := clientutil.NewKubeClient(cl)
 
 	updatedCluster := cluster.DeepCopy()
-	updatedCluster.Spec.KubernetesVersion = anywherev1.Kube126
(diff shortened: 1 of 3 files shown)


pkg/controller/clientutil/kubernetes_test.go
```diff
@@ -122,7 +122,7 @@ func TestKubeClientUpdate(t *testing.T) {
 	client := clientutil.NewKubeClient(cl)
 
 	updatedCluster := cluster.DeepCopy()
-	updatedCluster.Spec.KubernetesVersion = anywherev1.Kube126
+	updatedCluster.Spec.KubernetesVersion = anywherev1.Kube127
```
rahulbabu95, line 125: nit: Don't think we care about unit tests while deprecating a K8s version right. Like we have so many unit tests still referencing very old K8s version. But this shouldn't affect anything.

Issue #8787, opened 2024-09-17
BOTTLEROCKET_VERSION is null


saiteja313:
**What happened**:

```
export BOTTLEROCKET_VERSION=$(curl -sL https://raw.githubusercontent.com/aws/eks-anywhere-build-tooling/$BUILD_TOOLING_COMMIT/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES | yq ".$(echo $KUBEVERSION | tr '.' '-').$BOTTLEROCKET_IMAGE_FORMAT-release-version")

echo $BOTTLEROCKET_VERSION
```

https://anywhere.eks.amazonaws.com/docs/osmgmt/artifacts/#download-bottlerocket-node-images 

**What you expected to happen**:

`echo $BOTTLEROCKET_VERSION` should print bottlerocket image version. something wrong with the yq command.

**How to reproduce it (as minimally and precisely as possible)**:

Follow instruction from the below url and try to build OVA for any kubernetes version or specifically 1.28 or 1.30

https://anywhere.eks.amazonaws.com/docs/osmgmt/artifacts/#download-bottlerocket-node-images  

**Anything else we need to know?**:

Fix:
Broken: 
```
BUNDLE_MANIFEST_URL=$(curl -sL https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml | yq ".spec.releases[] | select(.version==\"$EKSA_RELEASE_VERSION\").bundleManifestUrl")
BUILD_TOOLING_COMMIT=$(curl -s $BUNDLE_MANIFEST_URL | yq ".spec.versionsBundles[0].eksD.gitCommit")
export BOTTLEROCKET_VERSION=$(curl -sL https://raw.githubusercontent.com/aws/eks-anywhere-build-tooling/$BUILD_TOOLING_COMMIT/projects/kubernetes-sigs/image-builder/BOTTLEROCKET_RELEASES | yq ".$(echo $KUBEVERSION | tr '.' '-').$BOTTLEROCKET_IMAGE_FORMAT-release-version")
```

Working:
```
BUNDLE_MANIFEST_URL=$(curl -sL https://anywhere-assets.eks.amazonaws.com/releases/eks-a/manifest.yaml | yq ".spec.releases[] | select(.version==\"$EKSA_RELEASE_VERSION\").bundleManifestUrl")
[...]

PR #8789, opened 2024-09-18
fixed BOTTLEROCKET_VERSION null value issue

saiteja313:
*Issue #, if available:*
https://github.com/aws/eks-anywhere/issues/8787 

*Description of changes:*
Bottlerocket ova version yq parsing is broken. 

*Testing (if applicable):*

tested with eksctl anywhere version v0.18.7

This is a doc update


vivek-koppuru (commented):
Why do we need to hardcode ami? I tried ova for example and it works for kubeversion 1-30

PR #8831, opened 2024-10-04
Update docs for Advanced Bare Metal cluster configuration:

jacobweinstock:
*Issue #, if available:*

*Description of changes:*

We changed to embedded images which necessitated the doc update.

*Testing (if applicable):*


docs/content/en/docs/getting-started/baremetal/bare-spec.md
```diff
@@ -577,294 +562,155 @@ spec:
...
+      INITRD_PATH: /boot/initrd
+      CMD_LINE: "root=/dev/sda3 ro"
+
+| Env variable | Type | Default Value | Required | Description |
+|--------------|------|---------------|----------|-------------|
+| BLOCK_DEVICE | string | "" | yes | The block device to mount. |
+| FS_TYPE | string | "" | yes | The filesystem type of the block device. |
+| KERNEL_PATH | string | "" | no | The path to the kernel to boot. When not supplied the kernel path will discovered from `/boot/grub/grub.cfg` on the `BLOCK_DEVICE`. |
+| INITRD_PATH | string | "" | yes | The path to the initial ramdisk to boot. |
+| CMD_LINE | string | "" | no | The command line to pass to the kernel. When not supplied the kernel cmdline parameters will discovered from `/boot/grub/grub.cfg` on the `BLOCK_DEVICE`. |
```
chrisnegus, line 666:
```suggestion
| CMD_LINE | string | "" | no | The command line to pass to the kernel. When not supplied, the kernel cmdline parameters will discovered from `/boot/grub/grub.cfg` on the `BLOCK_DEVICE`. |
```

docs/content/en/docs/getting-started/baremetal/bare-spec.md
```diff
@@ -577,294 +562,155 @@ spec:
...
-* volumes: The volume (directory) to mount into the container from the installed system.
-### version
-
-Matches the current version of the Tinkerbell template.
-
 ## Custom Tinkerbell action examples
-By creating your own custom Tinkerbell actions, you can add to or modify the installed operating system so those changes take effect when the installed system first starts (from a reboot or pivot).
-The following example shows how to add a .deb package (`openssl`) to an Ubuntu installation:
+
+The following example shows how to add a .deb package (`openssl`) to a Redhat installation:
```
chrisnegus, line 688: Probably "Red Hat", but see my note about different ways to refer to it.

chrisnegus (commented):
I don't know how to check the config examples, so I just did a copy edit. Looks good. Mostly small suggestions.

docs/content/en/docs/getting-started/baremetal/bare-spec.md
```diff
@@ -577,294 +562,149 @@ spec:
     version: "0.1"
 ```
 
-### Bottlerocket TinkerbellTemplateConfig example
+## TinkerbellTemplateConfig fields
 
-Pay special attention to the `BOOTCONFIG_CONTENTS` environment section below if you wish to set up console redirection for the kernel and systemd.
-If you are only using a direct attached monitor as your primary display device, no additional configuration is needed here.
-However, if you need all boot output to be shown via a server's serial console for example, extra configuration should be provided inside `BOOTCONFIG_CONTENTS`.
+Each Action has its own set of environment variables that are used to configure the Action. EKS Anywhere embeds a set of images into HookOS. This means that when these images are used in an Action they will not be pulled from any external registry. No external network communication is needed to use them. These Actions are prefixed with `127.0.0.1/embedded`.
```
2ez4szliu, line 567: does user still have the option to pull images from public ecr like in the example before? If yes do we want to add an example of pulling from public ecr?

jacobweinstock: Yes. Users can pull images from anywhere they want. I think if that wasn't clear that I should update the doc to make that more clear. In regard to providing an explicit example for pulling from ECR, I don't think we should. The whole idea of the embedded images is that pulling from ECR is not necessary. Also, there is not "latest" images in ECR for all the images. Each release we do has a new ECR location for images. Keeping these up to date in the docs is a lot of work and we have opted to not do this. This means that any exact reference provided in the docs is almost always out of date.

PR #8851, opened 2024-10-16
Skip generating AWS IAM Kubeconfig on cluster upgrade

jhaanvi5:
*Description of changes:*

This PR introduces a change to the write-cluster-config step of the upgrade workflow. Specifically, it adds a check to skip writing the AWS IAM Kubeconfig to disk during this step. During a cluster upgrade, the AWS IAM Kubeconfig is not updated. As a result, the step to write the Kubeconfig to disk can be safely skipped without any impact on the upgrade process.

*Testing:*
Built eksa binary locally and tested upgrade workflow.

### DIFF
--- a/pkg/workflows/workload/upgrade_test.go
+++ b/pkg/workflows/workload/upgrade_test.go
@@ -207,7 +207,7 @@ func TestUpgradeRunWithAWSIAMSuccess(t *testing.T) {
 	test.expectUpgradeWorkloadCluster(nil)
 	test.expectBuildClientFromKubeconfig(nil)
 	test.expectWriteWorkloadClusterConfig(nil)
-	test.expectAWSIAMAuthKubeconfig(nil)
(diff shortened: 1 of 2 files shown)


pkg/workflows/workload/upgrade_test.go
```diff
@@ -135,6 +135,11 @@ func (c *upgradeTestSetup) expectWriteWorkloadClusterConfig(err error) {
 	)
 }
 
+func (c *upgradeTestSetup) expectAWSIAMAuthKubeconfig(err error) {
```
vivek-koppuru, line 138: I am wondering if we call this "without" or "no" to make it clear that it is indeed not trying to call AWSIAMSuccess
```suggestion
func (c *upgradeTestSetup) expectWithoutAWSIAMAuthKubeconfig(err error) {
```

jhaanvi5: Makes sense, updated this.

PR #8963, opened 2024-11-02
Remove curated package list from concepts page

sp1999:
*Description of changes:*
This PR removes the curated package list from concepts page as it is already being referenced on the same page above as a hyperlink to the updated list in the package management section. It makes sense to have a single source of truth for the updated list of curated packages. The PR also fixes the link for the EKS Distro documentation page.

*Testing (if applicable):*
`make build`


abhay-krishna (approved):
/approve

PR #8967, opened 2024-11-02
Remove baremetal-branch flag from the integration test CLI

sp1999:
*Description of changes:*
This PR removes the `--baremetal-branch` flag from the integration test CLI as we no longer run tinkerbell tests simultaneously on multiple branches. This flag was introduced in [#3201](https://github.com/aws/eks-anywhere/pull/3201) more than two years ago to set the branch name for baremetal tests to run on. We now have additional mechanisms to prevent tests from being triggered from different branches at the same time so this flag doesn't add any value anymore. Since we started supporting n-1 EKS-A release branch too, we keep doing patch releases every 3 out of 4 weeks which requires us to run tests on release branches frequently. This change will reduce the headache of changing the secret value every time we do a patch release from main to release branch and then back to main. We also added periodic triggers for release pipelines to run tests on release branches every weekend at 7pm PST. But those pipelines don't have anything to handle changing the secret value before running the tests so this will also help in making sure that even tinkerbell tests run on the release branches every weekend. Overall, removing this flag adds a lot of value to our CI and release process.

*Testing (if applicable):*
```
make eks-a
make build-all-test-binaries
make lint
make unit-test
```

### DIFF
--- a/cmd/integration_test/build/buildspecs/cloudstack-test-eks-a-cli.yml
+++ b/cmd/integration_test/build/buildspecs/cloudstack-test-eks-a-cli.yml
@@ -108,7 +108,6 @@ phases:
         --cleanup-resources=true
         --test-report-folder=reports
         --branch-name=${BRANCH_NAME}
-        --baremetal-branch=${BAREMETAL_BRANCH}
   post_build:
     commands:
       - unset AWS_SDK_LOAD_CONFIG AWS_PROFILE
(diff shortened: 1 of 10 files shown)


abhay-krishna:
Also need to remove [this](https://github.com/aws/eks-anywhere/blob/main/scripts/e2e_test_docker.sh#L103) line.

abhay-krishna (approved):

